PR #127, opened 2026-07-03
feat(chat): add create_project tool to VoC chat

HanJeongho:
Ports the `create_project` chat tool onto `development`. Lets the VoC AI chat turn analysis into a pre-filled project ("make a project out of this" / "프로젝트 만들어줘"): it creates the project META item and seeds PRODUCT_CONTEXT from the feedback the model analyzed, then emits a `project_changed` SSE event so the frontend can link the new project.

This is the first of a series of features being reconciled onto `development`, kept deliberately isolated so it can be reviewed and merged on its own.

## What landed (create_project only)
- **`tools/create-project.ts`** (NEW) — `executeCreateProject`: writes `PROJECT#/META` (mirrors `projects.py::create_project`) + an optional `PRODUCT_CONTEXT` seed (only fields grounded in the feedback; nothing invented).
- **`tools/index.ts`** — `getCreateProjectTool()` tool definition.
- **`tools/executor.ts`** — `case 'create_project'` → `executeCreateProject`, `ProjectChange` type, `tool_result` summary, `project_changed` SSE, `projectChange` in the return value.
- **`handler.ts`** — wires `create_project` into VoC chat tools; threads `collectedProjectChanges` through the loop; passes `PROJECTS_TABLE` (no `projectId` — VoC chat is project-agnostic); `done` event carries `project_changes`.
- **`context/voc-context.ts`** — system prompt now describes both tools + a guideline: analyze feedback first, only seed fields you can support.

## Tests (11 create_project cases; suite 170 → 177, all green)
- `create-project.test.ts` (4) — logic: table-missing throw, missing-name soft fail, META item w/ `proj_` id, `PRODUCT_CONTEXT` partial seed.
- `index.test.ts` (+3) — schema: name is `create_project`, only `name` required, product-context seed fields exposed.
- `executor.test.ts` (+4) — wiring: routing, `projectsTable`-required guard, `tool_result` "created: <name>" + `project_changed` SSE, return value.
- `handler{,.extended}.test.ts` — mock factories register `getCreateProjectTool`.

## Verification
- `tsc --noEmit`: clean
- `vitest run`: **177/177 pass** (16 files); baseline `development` also 170/170 → no regression
- Diff vs `development`: 10 files, stream package only, no bleed from other in-flight features.

### DIFF
--- a/voc-datalake/lambda/stream/src/context/voc-context.ts
+++ b/voc-datalake/lambda/stream/src/context/voc-context.ts
@@ -131,7 +131,9 @@ async function fetchCategoryCounts(
 function buildSystemPrompt(responseLanguage?: string): string {
   const base = `You are a Voice of the Customer (VoC) analytics assistant. You help analyze customer feedback data and provide actionable insights.
 
-You have access to a tool called "search_feedback" that lets you search and retrieve customer feedback from various sources (web scrapers, manual imports, S3 imports, etc.).
+You have access to two tools:
+- "search_feedback": search and retrieve customer feedback from various sources (web scrapers, manual imports, S3 imports, etc.).
+- "create_project": turn the insights from this conversation into a new project, pre-filling its product context.
 
 IMPORTANT GUIDELINES:
 1. ONLY use the search_feedback tool when the user's question is specifically about customer feedback, reviews, or customer opinions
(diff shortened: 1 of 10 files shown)


HanJeongho:
@perrozzi this is the #9 `create_project` pilot from the GitLab → `development` reconciliation, ported onto `development` in isolation as we discussed.

- **Isolated**: only the `create_project` lines were hand-ported into the shared wiring files (`handler.ts`, `index.ts`, `voc-context.ts`, `executor.ts`); the bundled search improvements and the older roundtable rework were deliberately left out (the latter would regress the newer single-turn version already on `development`).
- **Tests**: 11 create_project cases (incl. new executor wiring tests that weren't on the source branch); suite 170 → **177**, baseline `development` also green → no regression. `tsc` clean.
- **Merge heads-up for ordering**: when the search-improvement PR (#3/#4/#8) lands too, the only expected conflict between it and this PR is in `voc-context.ts` (both edit the same system-prompt block) — resolves trivially by keeping both guidance blocks. `handler.ts`/`index.ts` touch different lines and auto-merge.

Search improvements are up next. Let me know if you'd prefer a different ordering or want me to fold anything in.

PR #128, opened 2026-07-07
feat(chat): search improvements — aggregate mode, urgency sort, coerce/pagination fixes

HanJeongho:
Ports the `search_feedback` improvements onto `development`, isolated from `create_project` (#127, already merged) so it reviews and merges on its own. Three fixes that compound on large datasets:

## What landed
**`tools/search-feedback.ts`**
- **Zero-results bug** — `sentiment_score`/`rating` are stored as DynamoDB strings (`S`) for ~all rows, so a strict `z.number()` rejected nearly every item and `fetchCandidatesByDate` silently dropped whole days → **every search returned 0 results**. Fixed with `z.coerce.number()` + per-row `safeParse` (one malformed row no longer discards the entire day).
- **Pagination** — a DynamoDB Query caps each page at 1MB (often ~500 large items), so a single query truncated big days (symptom: "987 negative items but the tool only saw 116"). Added `LastEvaluatedKey` do-while paging bounded by `MAX_CANDIDATES = 10000`.
- **Broad-question loop** — added `mode="aggregate"` (distribution stats over the ENTIRE match set — counts by urgency/sentiment/category/source + avg rating + top examples — in ONE call) and `sort_by="urgency"` (high→medium→low, most negative first) so "summarize all" / "most urgent issues" resolve in a single round instead of exhausting the tool loop.

**Wiring**
- `tools/index.ts` — `search_feedback` gains `mode` + `sort_by` params and aggregate/urgency guidance.
- `context/voc-context.ts` — system-prompt guidelines steer broad questions to `mode="aggregate"` and urgency filtering (the `query` field is a literal substring match, so "urgent" in the query won't filter by level).
- `handler.ts` — `MAX_TOOL_LOOPS` 5 → 15 (the real ceiling is the 300s Lambda timeout; broad questions now resolve in one round via aggregate) + per-round CloudWatch trace and a loop-exhaustion warning.

## Tests (search suite; 170 → 183, all green)
- `search-feedback.test.ts` (+130): string-number coercion, bad-row skip, `LastEvaluatedKey` pagination, aggregate stats, urgency sort.
- `handler.extended.test.ts`: loop-cap assertion 5 → 15.

## Verification
- `tsc --noEmit`: clean
- `vitest run`: **183/183 pass** (16 files), no regression
- `search-feedback.ts`/`.test.ts` are identical to the source branch; wiring files carry only the search-related lines (verified via diff vs `development`).

## Deliberately excluded
- `create_project` — already on `development` via #127; not re-touched.
- The source branch's older roundtable multi-round rework — `development` already has the newer single-turn design, so porting the old version would regress it.

### DIFF
--- a/voc-datalake/lambda/stream/src/context/voc-context.ts
+++ b/voc-datalake/lambda/stream/src/context/voc-context.ts
@@ -139,11 +139,13 @@ IMPORTANT GUIDELINES:
 1. ONLY use the search_feedback tool when the user's question is specifically about customer feedback, reviews, or customer opinions
 2. For general questions, greetings, or non-feedback topics, respond directly WITHOUT using the tool
 3. When you DO use the tool, be specific with your search query to get relevant results
-4. Base your answers on the actual data returned by the tool
-5. Quote actual customer feedback when relevant
-6. Highlight urgent issues that need attention
-7. Provide actionable recommendations based on the data
-8. When the user asks to turn findings into a project ("make/create a project", "프로젝트 만들어줘"), call create_project. First make sure you've analyzed the relevant feedback (search_feedback) so you can draft a grounded name, description, and product-context fields (product_name, one_liner, target_users, problem_solved, key_features). Only fill fields you can support with the actual feedback — omit the rest rather than inventing.
+4. For broad questions (summarize, count, trends, "top/most urgent/biggest issues"), call search_feedback with mode="aggregate" — it returns stats over the ENTIRE dataset in one call. Do NOT repeatedly page through individual items.
+5. To find urgent/critical feedback, pass urgency="high" and/or sort_by="urgency". The query field is a literal substring match on review text, so words like "urgent" in the query will NOT filter by urgency level.
+6. Base your answers on the actual data returned by the tool
+7. Quote actual customer feedback when relevant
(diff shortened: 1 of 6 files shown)


HanJeongho:
@perrozzi This is the next one in the reconciliation after #127 — the `search_feedback` improvements (#4/#8/#3), isolated on top of `development`.

The headline is the **zero-results bug**: ingestion stores `sentiment_score`/`rating` as DynamoDB strings, so the strict `z.number()` schema was rejecting nearly every row and searches returned 0 items. That plus 1MB-page truncation (fixed with pagination) and the aggregate/urgency mode (so broad questions don't burn the tool loop) are the three fixes here.

**Merge-order note:** this branch is cut from `development` *after* #127 landed, so `create_project` is already in the base — no conflict with it. The only shared files are the wiring ones (`handler.ts`, `index.ts`, `voc-context.ts`) and they carry only the search lines. Roundtable is left as the current single-turn version (I did **not** bring back the older multi-round rework, which would regress it).

183/183 tests green, `tsc` clean. Happy to record a quick demo of the aggregate/urgency flow if useful — this is the change that makes the "chat around the feedback, then make a project" flow work end-to-end (the loop-cap limitation I mentioned on #127's demo is fixed here).

PR #129, opened 2026-07-07
fix(bedrock): auto-continue on max_tokens + allow omitting temperature

HanJeongho:
Hardens `converse()` against two issues that surface on long generations and newer models. Both live in the same function, so they land together.

## What landed (`lambda/shared/converse.py`)
- **Auto-continuation** — PRDs (and other long outputs) were being **silently truncated** at the `maxTokens` ceiling: the code read the text but ignored `stopReason`, so a `stopReason=="max_tokens"` response was saved as-is. `converse()` now takes `max_continuations` (default 8); when a call stops at the token ceiling it replays the prior output + a continue-prompt as an assistant turn and concatenates, preserving the full document. Disabled when extended thinking is on (`thinking_budget>0`) since thinking-block replay isn't supported. `_invoke_with_retry` now returns `(text, stop_reason)` — it's only called inside `converse.py`, so no external caller changes.
- **Optional temperature** — `temperature` is now `float | None`. Passing `None` omits it from `inferenceConfig` entirely, which is required for models like Opus 4.8 that deprecate/reject the param. Existing float callers are unchanged (default stays `0.1`).

## Why both in one PR
Both edit the same `converse()` signature/body; splitting them would create an artificial signature conflict. `temperature=None` is a 7-line independent addition.

## Tests
- `test_converse.py`: adds `TestConverseAutoContinuation` (max_tokens replay, continuation cap, thinking-disabled path, concatenation) + temperature-omit assertions.
- `test_converse_coverage.py`: updated for the `(text, stop_reason)` tuple.
- **converse: 43/43 pass.** `document_generator` (a `converse_chain` consumer): **21/21** — no signature regression.

## Verification
- `converse.py` is identical to the source branch; `prd-cutoff-fix` was already built on `development`'s latest `converse` (`_converse_with_retry` present in both), so the copy keeps dev's raw-response refactor.
- Pre-existing `test_http.py` failures (`module 'shared' has no attribute 'http_utils'`) reproduce on clean `development` too — unrelated to this change.

### DIFF
--- a/voc-datalake/lambda/shared/converse.py
+++ b/voc-datalake/lambda/shared/converse.py
@@ -16,6 +16,26 @@ DEFAULT_MAX_RETRIES = 5
 DEFAULT_BASE_DELAY = 1.0  # seconds
 DEFAULT_MAX_DELAY = 30.0  # seconds
 
+# Auto-continuation: when the model stops because it hit the maxTokens ceiling,
+# the output is truncated mid-document. We transparently resume the generation up
+# to this many times, concatenating the chunks, so callers never persist a
+# half-written PRD/PR-FAQ/report (the document-cutoff bug).
+#
+# Tuning note: keep per-call max_tokens MODEST (≈6-8K) rather than huge. A single
+# huge maxTokens (e.g. 24K) makes one Bedrock call run many minutes for slow CJK
+# output and can blow the Lambda timeout on its own; many small calls each finish
+# in ~1-2 min and resume cleanly. This ceiling must therefore be generous enough
+# to assemble a long document from those small chunks (8 × ~8K ≈ 64K tokens).
+DEFAULT_MAX_CONTINUATIONS = 8
+
+# Nudge sent as the user turn when resuming a truncated response. Kept terse and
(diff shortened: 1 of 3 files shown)


HanJeongho:
@perrozzi Next in the reconciliation — the `converse()` hardening.

The main fix is the **silent PRD truncation**: `converse()` read the response text but discarded `stopReason`, so when a long generation hit the `maxTokens` ceiling it was saved truncated with no error. This adds auto-continuation (replay + continue-prompt, capped at 8) so the full document is preserved. I folded in a small related change — `temperature=None` to omit the param for models like Opus 4.8 that reject it — since it touches the same signature and splitting it would just create a conflict.

**No overlap with #127/#128** — this is `lambda/shared/converse.py` (Python), separate from the chat TS lambda. `_invoke_with_retry` now returns a tuple but it's only used inside converse.py, and `converse()`'s public return type is unchanged, so the ~10 call sites (document_generator, processor, etc.) are unaffected — verified document_generator's tests still pass (21/21).

43/43 converse tests green. Note: the `test_http.py` failures you may see are pre-existing on `development` (unrelated `http_utils` import issue), not from this change.

PR #131, opened 2026-07-09
feat(documents): document-generation overhaul — Step Functions PRD/PR-FAQ, product context, HTML prototypes

HanJeongho:
Ports the document-generation feature set onto `development` as one coherent change. The three pieces share `document_generator/handler.py` and the DocumentsTab UI, so splitting them would create artificial conflicts — they land together, isolated from everything else in the reconciliation queue.

## 1) PRD/PR-FAQ generation as a Step Functions workflow
Long CJK documents were overrunning the single 15-minute Lambda: with auto-continuation (#129) each LLM step can run minutes, and a 3–4 step chain in one invocation hits the ceiling.

- **`document_generator/handler.py`** — step router (`gather` → `run_step` ×N → `save`, plus an error handler). Each LLM step now gets its own 15-min Lambda budget. Intermediate step text is stashed in S3 (claim-check) under `scratch/document_jobs/*`, keeping Step Functions state far below its 256KB cap. The legacy single-shot path remains for prototype/report jobs.
- **`api-stack.ts`** — `createDocumentStateMachine` (DocGather → DocStep0..3 with Choice gates and an LLM retry policy → DocSave, error catch-all), `grantStartExecution` to the Projects API, `DOCUMENT_STATE_MACHINE_ARN` env, S3 scratch permissions.
- **`projects_handler.py`** — generate-document starts the state machine for `prd`/`prfaq` and falls back to a direct Lambda invoke when the ARN is absent.
- **prompts** — `max_tokens` rebalanced (solution 12k, prd_document 32k, prfaq steps 8k) so steps complete without continuation loops.

## 2) Product context workspace
- **`product_context.py`** (NEW) — PRODUCT_CONTEXT store, interview flow, product-doc upload/extract, and `build_product_context_block` used by the PRD/PR-FAQ chains so generated documents are grounded in the product profile.
- `shared/prompts.py` — generation steps take `product_context`; PR-FAQ pins `launch_date` ~90 days out (prevents past launch dates in Working-Backwards docs).
- Routes: `/product-context` (get/put/interview), `/product-docs`, `/prfaq-autofill`, `/research/suggest-questions`, `/documents/suggest-brief`.
- Frontend: ProductTab (interview + doc upload), AI-draft brief & research-question suggestions in the wizards.

## 3) Clickable HTML prototypes
[...]

PR #132, opened 2026-07-10
feat(documents): document-generation overhaul — deployable + all UI findings fixed (supersedes #131)

perrozzi:
## Supersedes #131

This PR is a **superset of #131** (@HanJeongho's document-generation overhaul): it contains every commit from that PR's branch (`document-generation-overhaul`, head `f77a4af`) plus fixes that make it deployable and repair three functional defects surfaced during live end-to-end UI testing. Same base branch (`development`). Recommend closing #131 in favor of this one.

All of the original feature work is preserved: Step Functions-based PRD/PR-FAQ generation (per-step 15-min Lambda + S3 claim-check for the 256 KB state cap), the Product Context tab, and Opus-generated HTML prototypes.

## Why supersede rather than review-and-merge #131

#131 as committed was **not deployable** and, once made deployable, had **three functional defects** that only surface at runtime (not in unit tests / `cdk synth` / `tsc`). Rather than push a long review thread onto #131, this PR bundles the original work with the fixes, all verified against a live AWS deployment.

## What this adds on top of #131 (3 commits)

**1. Deployability (`71cef8c`)**
- Removed orphaned `ExtensionApi` wiring in `api-stack.ts` that referenced a handler file never included in the PR (`cdk synth`/deploy failed on the missing asset).
- Dropped a dead `ChatMessage`/`ChatConversation` re-export from `client.ts` that broke the frontend `tsc -b` build.
- Added PR #131 route coverage to `scripts/test-api.sh`.

**2. Findings 1 & 2 (`cb6bb4a`)**
- **Finding 1 — Product report generation 100% broken:** `createJobLambdaCode()` never bundled `api/product_context.py`, so the `product_report` doc type's hard `import` failed with `ModuleNotFoundError` on every deployed invocation. Now bundled.
- **Finding 2 — PRD generation could time out with no retry:** the `prd_document` step (`max_tokens=32000`) could consume the entire 900s Lambda budget and fail with `Sandbox.Timedout`, which was **not** in the Step Functions retry list, so the whole job failed instead of retrying. Lowered `max_tokens` to 8000 (matching `converse.py`'s own per-call guidance — it auto-continues past the ceiling, so total output is unaffected) and added `Sandbox.Timedout` to the retry list. Root-caused via `aws stepfunctions get-execution-history` on a live failed execution.

**3. Finding 3 — the flagship "clickable HTML prototype" was not clickable (`f7acd56`)**
- Root cause: the app's CloudFront CSP (`script-src 'self'`) blocked the model-generated inline `<script>` inside the sandboxed `srcdoc` iframe, so nav/tab clicks did nothing.
- Fix: serve prototypes from a dedicated `/prototypes/*` CloudFront behavior on the existing distribution, with its own permissive `ResponseHeadersPolicy`, backed by an S3 prefix; the frontend loads them via a cross-document `<iframe src=...>` (no `srcDoc`/`sandbox`). New prototypes are S3-only (`prototype_url`, no DynamoDB `content`), with the legacy `srcDoc` path preserved for pre-migration docs.
- Completed with `frame-src 'self'` on the main SPA CSP — without it the parent CSP falls back to `default-src 'none'` and the browser refuses to frame even same-origin content. (`frame-src` = what a page may embed; distinct from `frame-ancestors` = who may embed a page. Only the former was missing.)

## Verification

- **Deployed live** to a test AWS account (all stacks; document-generation Step Functions state machine active).
- **API smoke tests pass** (`scripts/test-api.sh`), including the new document-generation routes.
- **Live browser E2E** of the prototype: iframe loads via `src=` with **zero CSP/console errors**, and clicking in-prototype navigation (Dashboard → Weekly Digest) switches screens — i.e. the inline JS actually executes. Findings 1 & 2 confirmed fixed against live Step Functions executions.
- Unit suites green (backend `pytest`, frontend `vitest`, CDK `vitest`, all `tsc` builds).

## Credit

The document-generation overhaul is @HanJeongho's work (#131). This PR only adds deployability and the three runtime fixes on top.

### DIFF
--- a/voc-datalake/bin/voc-datalake.ts
+++ b/voc-datalake/bin/voc-datalake.ts
@@ -136,6 +136,7 @@ const apiStack = new VocApiStack(app, 'VocApiStack', {
   kmsKey: coreStack.kmsKey,
   rawDataBucket: coreStack.rawDataBucket,
   avatarsCdnUrl: coreStack.avatarsCdnUrl,
+  prototypesCdnUrl: coreStack.prototypesCdnUrl,
   websiteBucket: coreStack.websiteBucket,
   frontendDistribution: coreStack.frontendDistribution,
   frontendDomainName: coreStack.frontendDomainName,
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -58,7 +58,29 @@
     "editDocument": "Edit document",
     "newDocument": "New Document",
     "noDocuments": "No documents",
-    "selectDocument": "Select a document"
+    "selectDocument": "Select a document",
+    "prototype": {
+      "previewLabel": "Live preview",
+      "rawLabel": "Raw output (parse failed — please regenerate)",
+      "openNewTab": "Open in new tab",
+      "downloadHtml": "Download .html",
+      "downloadJson": "Download .json",
+      "button": "Build Prototype",
+      "building": "Building…",
+      "buttonTitle": "Generate a clickable HTML prototype from this project's PRD + PR-FAQ",
+      "buttonTitleOne": "Generate a clickable HTML prototype from the available document (PRD or PR-FAQ)",
+      "needsDocs": "Create a PRD or a PR-FAQ first to enable prototype build",
+      "confirmPrdOnly": "No PR-FAQ yet — the prototype will be built from the PRD only. Continue?",
+      "confirmPrfaqOnly": "No PRD yet — the prototype will be built from the PR-FAQ only. Continue?",
+      "timeout": "Prototype build took too long. Check the Documents tab in a moment.",
+      "feedbackButton": "Revise with feedback",
+      "feedbackTitle": "Give feedback to regenerate this prototype",
+      "feedbackHeading": "Revise prototype with feedback",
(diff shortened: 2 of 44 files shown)


HanJeongho:
@perrozzi Thank you for taking this over the finish line, and apologies for the state #131 was in — that one's on me.

Some context on where the leftovers came from: my source branch also carries **multi-instance support** (prefixed resource names so several isolated copies of the stack can run in one account), which I use for hands-on workshops with per-team deployments. Since that becomes unnecessary once this is officially published as a workshop, the port was deliberately a "strip the multi-instance changes, keep the feature work" exercise. The stripping itself landed clean, but I submitted the PR **without a live deployment pass**, and that let an orphaned leftover from the source branch (the ExtensionApi wiring, from a different feature) and the runtime-only defects slip through — exactly the class of problem unit tests and `cdk synth` can't catch.

Lesson fully taken on my side: no PR goes up without a live deploy + end-to-end check first. Thanks again for root-causing the CSP behavior properly (the `/prototypes/*` behavior is a much cleaner fix than what I had in the source environment) and for preserving the credit — much appreciated. 🙇

perrozzi:
no worries at all! :-) thanks for your work

PR #108, opened 2026-05-30
feat: Synthetic Data Review Generator plugin (Bedrock-powered)

mundurragacl:
## Summary

This PR adds a self-contained **Synthetic Data Review Generator** plugin alongside several quality fixes that surfaced while landing the feature.

### Feature: Synthetic Data Review Generator plugin

A new opt-in `synthetic` plugin that uses Amazon Bedrock (Claude Sonnet) to generate realistic synthetic reviews for development and testing. The plugin is fully isolated:

- Runs in its own Lambda with a **dedicated IAM role**. The shared ingestion role is not widened.
- Wiring is gated by a manifest capability flag (`infrastructure.ingestor.bedrock`), so future plugins that need Bedrock can declare it without core-stack changes.
- New "synthetic" category in the Scrapers UI with `GeneratorConfigModal` for prompt + count + model selection.
- Enable / disable via `cdk.context.json` like every other plugin.
- 26 backend unit tests + 4 frontend tests, all passing.

Files: `plugins/synthetic_reviews/{manifest,ingestor/handler,README}`, `frontend/src/pages/Scrapers/GeneratorConfigModal.tsx`, plus i18n strings for all 8 locales.

### Screenshots

New "Synthetic" category appears in the Scrapers page, with the generator config modal:

Scrapers page with Synthetic category

Generator config modal

Generated reviews flowing through the pipeline

### Bug fix: ManualImportModal stuck on the wrong button label

`ManualImportModal.tsx` was reading `isConfirmingRef.current` during render to choose the button label. Mutating a ref does not trigger a re-render, so users never saw the in-flight `Importing…` state, and a fast double-click could submit twice. Replaced the ref with `useState`. Added 2 regression tests that fail on the buggy version and pass on the fix.

### Refactor: extract `useSettingsForm` hook

The Settings page had grown to a complexity of 28 because it owned 6 form fields plus their server / store derivation rules inline. Extracted a `useSettingsForm` hook that owns the `FormEdits` state, derived values, setters, and reset / clear helpers. The component now delegates rendering to a `SettingsTabContent` sub-component. Behavior preserved (29 tests pass), complexity well under the limit.

### Backend bug fixes (pre-existing — fixed while landing the feature)

Per the maintainers' direction to own the codebase: while running the full check suite I found and fixed several real bugs:

| Area | Fix |
|------|-----|
| `lambda/research/research_step_handler.py` | Missing `import json` — every failed Step Functions job would crash inside the error handler with `NameError`. |
[...]

PR #135, opened 2026-07-13
fix(plugins): honor the per-app enabled flag in app-review ingestors (PR #108 P2)

perrozzi:
## Why 
Both the Android and iOS app-review ingestors parse an `enabled` boolean per app config (`models.py` defaults it to `True`, the config loaders read it from the shared secret) — but `fetch_new_items` iterated every config without checking it. A disabled app was still scraped on every scheduled run. 
## What 
- `plugins/app_reviews_android/ingestor/handler.py`: skip guard + info log in `fetch_new_items` (+3) 
- `plugins/app_reviews_ios/ingestor/handler.py`: same (+3) 
- NEW `plugins/test_app_reviews_enabled_filter.py`: 6 regression tests (disabled skipped, enabled processed, mixed list preserves order — per platform) Ported from PR #108 (cluster 2 of the `.pr108-reconcile` breakdown; follows the merged P1/#130 synthetic-reviews plugin). Behavior applied onto `development`'s current handler shape, not a cherry-pick.
-  ## Verification 
- New suite 6/6 pass; **verified the 4 disabled-app tests FAIL with the guard reverted** (true regression tests) 

- Full `pytest plugins/`: **195 passed** (189 existing + 6 new), no module-cache pollution between the two same-named flat `models`/`countries` plugin modules 
- Backend-only, no CDK/frontend change

### DIFF
--- a/voc-datalake/plugins/app_reviews_android/ingestor/handler.py
+++ b/voc-datalake/plugins/app_reviews_android/ingestor/handler.py
@@ -166,6 +166,9 @@ class AndroidAppReviewsIngestor(BaseIngestor):
             return
 
         for app in self.app_configs:
+            if not app.enabled:
+                logger.info(f"Skipping disabled Android app: {app.name} ({app.package_name})")
+                continue
             yield from process_app_reviews(
                 app_config=app,
                 app_name=app.name,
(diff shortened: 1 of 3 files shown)


perrozzi:
Consolidating: this change is being stacked onto the #134 branch (one PR for the lint cleanup + this fix) and will be live-tested before #134 is finalized.

PR #138, opened 2026-07-13
feat: filter by review date vs imported date across the app

### LINKED ISSUE #95: [Bug]: when updating the rating for prfaq , the score doesn't always update on the dashboard
### URL / Page

_No response_

### Summary

When I'm updating the score in the PRF AQ, the dashboard doesn't reflect that. 

### Steps to Reproduce

Update a P, R, F, A, Q the score and then look at the dashboard and see if it updates. This may happen only on things that have been edited after the fact as opposed to the initial one. I'm not sure 

### Expected Behavior

The score would get updated inside the PRFAQ. Dashboard.

### Actual Behavior

It doesn't update.

### Affected Area

Frontend - Dashboard

### Affected Files (if known)

_No response_

### Additional Context

_No response_

### LINKED ISSUE #99: Source-filtered feedback queries return 0 results when a single source dominates the GSI1 date partition
## Summary

Several handlers query the `gsi1-by-date` GSI with a fixed `Limit`, accumulate items into a list with an early-break, and **then** filter in memory by `source_platform` / `category` / `sentiment_label` / `urgency`. When a single source dominates a date partition (busy webscraper, bulk import, S3 import, manual paste), the early break fills the in-memory list with rows that don't match the filter, so the filter eliminates everything and the user sees zero results despite the data existing.

The dropdown in the UI still shows the source as a valid option (it's populated from aggregate counts in DynamoDB), so the user knows the data is there — they just can't see it.

The fix is the same in every site: when a filter is set, paginate the partition with `LastEvaluatedKey` until either the partition is drained or a generous accumulation cap is hit, then apply the filter.

## Affected files (all on `main` @ 2f744c2)

| Handler | File | Endpoint / function | Lines |
|---|---|---|---|
| Metrics | `voc-datalake/lambda/api/metrics_handler.py` | `GET /feedback` (`list_feedback`) | [39-82](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L39-L82) |
| Metrics | `voc-datalake/lambda/api/metrics_handler.py` | `GET /feedback/entities?source=…` | [135-182](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L135-L182) |
| Metrics | `voc-datalake/lambda/api/metrics_handler.py` | `GET /feedback/search` | [264-336](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L264-L336) |
| Metrics | `voc-datalake/lambda/api/metrics_handler.py` | `GET /metrics/sentiment?source=…` | [439-487](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L439-L487) |
| Metrics | `voc-datalake/lambda/api/metrics_handler.py` | `GET /metrics/categories?source=…` | [490-538](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L490-L538) |
| Projects | `voc-datalake/lambda/api/projects_handler.py` | PRD / PR-FAQ generation (`handle_generate_document_job`) | [419-433](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/projects_handler.py#L419-L433) |
| Projects | `voc-datalake/lambda/api/projects_handler.py` | Document merge (`handle_merge_documents_job`) | [567-582](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/projects_handler.py#L567-L582) |
| Shared | `voc-datalake/lambda/shared/feedback.py` | `get_feedback_context` (used by persona generation + research) | [12-72](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/shared/feedback.py#L12-L72) |
| Shared | `voc-datalake/lambda/shared/project_chat.py` | `get_feedback_for_chat` (Project Chat) | [88-131](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/shared/project_chat.py#L88-L131) † |
| Chat | `voc-datalake/lambda/api/chat_handler.py` | `POST /chat` (non-streaming) | [92-104](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/chat_handler.py#L92-L104) |
| Chat (stream) | `voc-datalake/lambda/api/chat_stream_handler.py` | `execute_search_feedback_tool` | [304-325](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/chat_stream_handler.py#L304-L325) |

† `shared/project_chat.py` also has a separate partition-key bug filed in a sibling issue.

## Canonical buggy shape (from `metrics_handler.py:39-82`)

```python
items = []
for i in range(days):
    date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
    response = feedback_table.query(
        IndexName='gsi1-by-date',
        KeyConditionExpression=Key('gsi1pk').eq(f'DATE#{date}'),
        Limit=500,
        ScanIndexForward=False
    )
    items.extend(response.get('Items', []))
    if len(items) >= limit * 5:        # ← early break, no filter applied yet
        break

if source:
    items = [i for i in items if i.get('source_platform') == source]
# … other in-memory filters
```

`gsi1sk` is `f"{processed_at_iso}#{feedback_id}"` (set at `processor/handler.py:511`), so the GSI orders rows by **processed-at**, not created-at. Whichever source was processed most recently sits at the top of the partition. With `ScanIndexForward=False` and `Limit=500`, the first 500 rows of today's partition can be 100% from one source — the in-memory filter for any other source then returns `[]`.

## Reproduction

1. Deploy `main` to a fresh account.
2. Bulk-import three sources sequentially (any of: scraper backfill, S3 import, manual paste, or `workshop-tools/import_reviews.py`):
   - source A: 500 records
   - source B: 500 records
   - source C: 500 records (last)
3. Open the Feedback page, select source A in the filter dropdown.
4. Observe: "Showing 0 of 0 items". The dropdown shows "A" because `/feedback/entities` reads from the aggregates table (correct count), but `/feedback?source=A` reads GSI1 and the early-break truncates to 500 source-C rows.
5. Same behavior in Data Explorer → Processed Feedback, Project persona generation (raises `ValidationError('No feedback data found for the given filters')`), and PRD/PR-FAQ generation.

A direct DynamoDB query confirms the data is present:

```
$ aws dynamodb query --table-name voc-feedback-… \
    --index-name gsi1-by-date \
    --key-condition-expression "gsi1pk = :d" \
    --expression-attribute-values '{":d":{"S":"DATE#2026-05-09"}}' \
    --filter-expression "source_platform = :s" \
    --expression-attribute-values '{":d":{"S":"DATE#2026-05-09"},":s":{"S":"source_a"}}' \
    --select COUNT
{"Count": 500, "ScannedCount": 1543, …}
```

## Suggested fix

Pattern, applied at every site listed above:

```python
has_post_filter = bool(source or sentiment or category)
accumulation_cap = 5000 if has_post_filter else limit * 5

for i in range(days):
    date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
    last_evaluated_key = None
    while True:
        qk = {
            'IndexName': 'gsi1-by-date',
            'KeyConditionExpression': Key('gsi1pk').eq(f'DATE#{date}'),
            'Limit': 500,
            'ScanIndexForward': False,
        }
        if last_evaluated_key:
            qk['ExclusiveStartKey'] = last_evaluated_key
        response = feedback_table.query(**qk)
        items.extend(response.get('Items', []))
        last_evaluated_key = response.get('LastEvaluatedKey')
        if not last_evaluated_key or not has_post_filter:
            break
        if len(items) >= accumulation_cap:
            break
    if len(items) >= accumulation_cap:
        break
```

Two notes on the fix shape:

1. **Only drain when filtering.** The unfiltered case (no `source`/`category`/`sentiment`) keeps the old early-break behavior and pays no extra DynamoDB RCUs.
2. **5000 is a generous safety cap, not a target.** In practice the inner loop exits via `LastEvaluatedKey is None` long before 5000.

A more durable redesign would query the base table directly via `pk = SOURCE#<source>` when `source` is set, but that requires mapping the `source_platform` slug to the `brand_name` actually stored in `pk` (which `processor/handler.py:506` writes as `SOURCE#{brand_name}`, falling back to `source_platform` only when `brand_name` is empty). The pagination fix above is a smaller, reversible change.

## Context / disclosure

I found this while preparing a workshop. The public scraper plugins return relatively sparse data for the products my workshop centers on (Cisco Meraki MX, Webex Contact Center, Cisco XDR), so I generated synthetic reviews and bulk-imported them via `import_reviews.py` (~500 per source × 3 sources). The bulk import is what makes the bug obvious — but the same race exists for any single high-volume source on the platform, including a busy real-world webscraper or any of the new S3-import workflows added in #79. Happy to send a PR with the patches across all 9 sites if useful; I have working diffs and have been running them in production for the workshop.

### LINKED ISSUE #103: [Bug]: BedrockAccessStack fails on accounts that already have Anthropic model access
## URL / Page

N/A (CDK deployment failure)

## Summary

The `BedrockAccessStack` fails with a hard error and rolls back when deployed to an account where the Anthropic use case submission is not required or has already been completed. The `PutUseCaseForModelAccess` API returns a failure, and the `AwsCustomResource` propagates it as a stack creation error, blocking the entire deployment.

## Steps to Reproduce

1. Configure `cdk.context.json` with `anthropicUseCase` and `"skipUseCaseSubmission": false` (the default)
2. Deploy to an account where model access already exists (e.g., previously submitted use case, org-level access, or internal accounts)
3. Run `npm run deploy:infra`

## Expected Behavior

Deployment should succeed. If the use case has already been submitted, the stack should treat it as a no-op and continue.

## Actual Behavior

The `BedrockAccessStack` fails and rolls back with:

```
Received response status [FAILED] from custom resource.
Message returned: Internal Accounts should not submit use case details
```

This halts the entire `deploy --all` operation at the first stack.

## Affected Area

Infrastructure - CDK

## Affected Files (if known)

- `lib/stacks/bedrock-access-stack.ts` (custom resource error handling)
- `docs/deployment.md` (missing documentation for `skipUseCaseSubmission`)

### Who is affected

- Accounts that previously completed the Anthropic use case form (via Console or a prior deployment)
- Accounts with org-level model access granted at the management account
- AWS internal accounts

### Current workaround

Set `"skipUseCaseSubmission": true` in `cdk.context.json`. This setting is undocumented in the deployment guide and the error message gives no hint about it.

### Proposed fix

1. **Handle the error gracefully**: The `SubmitAnthropicUseCase` custom resource should treat "already submitted" / "not required" API responses as success rather than failing the stack. This makes the deployment idempotent.

2. **Document `skipUseCaseSubmission`**: Add guidance in `docs/deployment.md` explaining when to set this flag to `true`:
   - Account has already completed the Anthropic use case submission
   - Use case was submitted manually via the AWS Console
   - Model access was granted at the organization level

mundurragacl:
## Summary

Customers asked to filter feedback by the date a review was **written**, not just the date it was **imported** — e.g. a batch import of 200 reviews yesterday includes 3-year-old, already-addressed reviews that flood every "last 7 days" view.

Every feedback item already stores both dates (`source_created_at` and `date`/`processed_at`); until now all time filtering used the import date only. This PR keeps both and lets users pick which one drives the time window.

## Issues addressed by this PR

Fixed here (please close on merge — this PR targets `development`, so GitHub won't auto-close them):

- Fixes #99 — source-filtered feedback queries returned 0 results when one source dominated a GSI1 date partition. The metrics handler now pages every partition via `LastEvaluatedKey` (10k-rows/partition ceiling) and pushes the source filter down as a server-side `FilterExpression`. (`91b2971`; the `shared/feedback.py` and streaming-chat sites were already fixed on `development`.)
- Fixes #103 — `BedrockAccessStack` failed and rolled back on accounts that already have Anthropic model access. The use-case submission is now best-effort (`ignoreErrorCodesMatching: ValidationException|ConflictException`; permission/throttling errors still fail loudly), and the `skipUseCaseSubmission` escape hatch is documented in `docs/deployment.md`. (`fa2b91e`)
- Fixes #95 — PR/FAQ prioritization scores didn't always update on the dashboard. The page seeded local state from the server exactly once per mount and ignored every refetch; scores are now derived from server state overlaid with unsaved edits. (`ed4f840`)

Also part of this PR:

- All 3 should-fix items and both nits from the review (partial-metrics disclosure via `is_partial` + `~` on Dashboard cards, scan-budget alignment with `/feedback`, paged category branch with honest `is_partial_window`, `_basis_date` shape guard, comment fix).
- The i18n semantic conflict with #139 (`DocsUpload` now owns its `projectDetail` namespace). (`189445a`)
- `/feedback` filtered by category alone **ignored the `days` window entirely** (time-unbounded GSI2 query) — now honors it for both bases.
- 8 pre-existing `react-hooks/set-state-in-effect` lint errors that made `npm run check` fail — fixed via React's render-phase adjustment pattern (ProductTab, ProductFormFields, Settings). Two extractions to stay in budget: `ProductDocsUpload.tsx`, `useSettingsSync.ts`.
- Local dev fallback pointed at `localhost:3000` (nothing serves it) and discarded a valid `VITE_API_ENDPOINT` when Cognito vars were absent — now keeps a valid endpoint, else falls back to the documented mock port 3001.
- Mock server: honors `days`/`date_basis`, adds `/feedback/entities`, returns 400 on malformed JSON instead of crashing, drops a `problem_summary: null` that broke frontend Zod normalization.

Follow-up filed: #150 (`date_basis` for AI Chat / Projects / MCP + urgent/search window unification).

### Feature

- **API**: new `date_basis` query param (`imported` default | `review`) on `/feedback`, `/feedback/urgent`, `/feedback/search`, `/feedback/entities`, and all `/metrics/*` endpoints, validated against an allowlist. Backwards compatible — omitted param preserves existing behavior.
- **No new GSI / no backfill**: a review can never be imported before it was written, so the existing import-date GSI window always contains the review-date window; review basis is a post-filter. Review-basis metrics are computed from raw items (aggregates are bucketed by import date), mirroring the existing source-filter branches.
- **UI**: "Filter dates by" picker in the header time selector (popover on desktop, section in the mobile dropdown), persisted per user. All time-filtered pages react automatically (Dashboard, Feedback, Categories, Problem Analysis, Social Feed, urgent badge).
- **Clarity**: Feedback detail shows both dates (Review date / Imported); cards tooltip their date; PDF export headers append "(by review date)" when active. Dashboard counts show `~` + tooltip when computed from a truncated scan (`is_partial`).

## Testing

- `npm run check` green end to end on the merged tree (current `development` ec4231d is merged in)
- Backend: 741 pytest tests pass, 1 pre-existing skip (32 new across date-basis endpoints, validator, category-window, dominated-partition/paging, partiality contract, malformed dates)
- Frontend: 1,582 vitest tests pass (new: picker component, store, param builder, label helper, runtime-config fallback, partial-metrics disclosure, prioritization refetch regression)
- `i18n:validate` exits 0 ("All translations are in sync")
- Verified live in a browser against the mock: imported basis shows 3 items, review basis drops the year-old backfilled review to 2; basis persists across reloads
- CDK `tsc` build clean; infrastructure change limited to the BedrockAccessStack error-handling fix (#103)

## Out of scope / notes

- AI Chat, Projects research, and MCP query by import date via their own filter UIs; the API now supports `date_basis` so they can adopt it later — tracked in #150.
- Pushed with `SKIP_AI_REVIEW=1`: the pre-push AI review diffs against `main` (all of `development`), not this branch; its blocking findings are in already-merged code.


perrozzi:
Ran a full local review of this branch (checked out at a0c39e8, on top of `development` e201270): behavioral diff review, full quality gate, and a live browser test against the mock. **No blockers — the design is sound and everything the description claims checks out.** Three should-fix items below, each with a pointer to where and a concrete suggestion, so they can be resolved in one pass.

## Verification (independent, local)

- ✅ `eslint` clean, all 3 typechecks clean (frontend `tsc -b`, CDK, stream)
- ✅ Frontend: **1,557/1,557** vitest tests pass; Backend: **695 pass, 1 pre-existing skip** (`--no-cov`; the coverage-HTML step crashed my shell but `pytest.ini` is untouched by this PR — environment quirk, not a regression)
- ✅ Live in browser (mock:3001 + vite:5173): imported basis shows 3 items, review basis drops the 400-day backfilled item to 2; `date_basis=review` appears on `/feedback`, `/feedback/entities`, `/feedback/urgent` and **all** `/metrics/*`; param correctly **omitted** for the default basis (cache keys unchanged); basis persists across reload (`voc-config` → `state.dateBasis`); detail page shows both dates; picker has proper listbox a11y semantics; no console errors
- ✅ Design verified: the no-new-GSI containment argument holds at date granularity; `validate_date_basis` ([shared/api.py L87-L95](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/shared/api.py#L87-L95)) is a strict allowlist (no injection surface, no 400s for old clients); pagination contract (`total`/`is_partial_window`) holds under review basis via the `has_post_filter` expansion

---

## Should-fix (3)

### 1. Review-basis metrics silently degrade to a capped sample — exactly in the motivating scenario

**Where:** [`_scan_recent_items`, metrics_handler.py L90-L113](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/api/metrics_handler.py#L90-L113), consumed by `_summary_from_items` (L547) and the review-basis branches of all `/metrics/*` + `/feedback/entities`.

Under `date_basis=review`, `/metrics/summary` switches from exact pre-computed aggregates to a scan that is (a) **single-page `Limit=500` per import day** — a bulk-backfill day (this PR's motivating use case: "batch import of 200 reviews yesterday") gets truncated to its most-recently-processed slice, (b) **hard-stopped at `CANDIDATES_SOFT_CAP=1000`** accumulated items, most-recent import days first, and (c) returns **no partiality signal**, unlike `/feedback`'s `is_partial_window`. The dashboard headline number flips from exact to a biased lower bound with no indication anywhere.

**Suggested fix (pick one, ordered by effort):**
1. *Minimum:* add an `is_partial: bool` field to the review-basis metrics responses when the soft cap or a per-day page limit is hit, and surface it in the UI (e.g., "~" prefix or tooltip on the metric cards). This makes the degradation honest without touching the scan.
2. *Better:* paginate the per-day query with `LastEvaluatedKey` — `shared/feedback.py::_query_all_pages` already exists for exactly this (its comment even names the single-page truncation bug class); reuse it here with a total-item budget.
3. Also worth a test: >1000 candidates in-window asserting whatever contract you choose (currently untested).

### 2. `/feedback` total vs `/metrics/summary` total can disagree under review basis

**Where:** [`MAX_FEEDBACK_OFFSET = 5000` (L25) vs `CANDIDATES_SOFT_CAP = 1000` (L36)](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/api/metrics_handler.py#L25-L36).

The Feedback page scans up to 5,000 candidates while the metrics scan caps at 1,000, so for the same window the list's `total` and the dashboard's `total_feedback` can visibly diverge (users will file this as a bug). **Fix:** align the review-basis metrics scan budget with `MAX_FEEDBACK_OFFSET` (or at least document/flag the divergence per item 1). If you take fix 1.2 above, passing the same budget to both paths resolves this for free.

### 3. Category+days fix sits on a non-paginated, sentiment-sorted GSI page — `is_partial_window` stays false when it shouldn't

**Where:** category branch of `list_feedback`, [metrics_handler.py L191-L200](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/api/metrics_handler.py#L191-L200), and the new window post-filter at [L215-L226](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/api/metrics_handler.py#L215-L226).

The bug fix itself is real and correct in direction (category-only queries did ignore `days` — nice catch). But the branch is a **single `query` call**: DynamoDB caps each page at ~1MB, so a category holding more than ~1MB of items returns only a few hundred, **sorted by `gsi2sk = sentiment_score#timestamp`** — i.e., by sentiment, not date. The new date filter then selects in-window items from that sentiment-biased sample, and `window_truncated = len(candidates) >= candidate_cap and 'LastEvaluatedKey' in response` (L200) never trips because the page never reaches 5,000. Net: large categories return a near-arbitrary in-window subset flagged as complete. The read pattern is pre-existing, but this PR's filter makes the truncation user-visible.

**Fix:** page the category query with `LastEvaluatedKey` up to `candidate_cap` (mirroring `shared/feedback.py::_query_all_pages`), or at minimum change the truncation expression to `'LastEvaluatedKey' in response` alone so `is_partial_window` is honest.

## Nits (2)

- **Off-by-one window between endpoints:** `_window_cutoff` (L83) uses `days - 1` — consistent with the date-loop branch — but `/feedback/urgent` ([L259](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/a0c39e88f78eda645ee98eb5c2cfd06f2fd3ac27/voc-datalake/lambda/api/metrics_handler.py#L259)) and `/feedback/search` (L447) keep `timedelta(days=days)`, so `days=7` is a 7-day review window on list/metrics but 8 days on urgent/search. Also the comment at L222-223 says the fix "match[es] the category-path semantics of shared/feedback.py", but `shared/feedback.py:82` uses `days`, not `days - 1` — the code is fine (internally consistent), the comment isn't. Suggest: fix the comment now; unify the urgent/search cutoffs in a follow-up (behavior change on the import basis too).
- **Fail-open on malformed `source_created_at`:** `_basis_date` (L71-L81) falls back only when `len < 10`; a non-date string ≥10 chars (`"unavailable"`) compares lexicographically above any `2026-…` cutoff, so it's kept and becomes a garbage bucket key in `daily_totals`. One-line guard: check `source_created[:10]` matches `\d{4}-\d{2}-\d{2}` before using it. Tests cover missing/short values but not this case.

## Observations (no action needed in this PR — suggest follow-up issues)

- **Timezone skew:** `source_created_at[:10]` is the source-local date while cutoffs/buckets are UTC → ±1-day skew at window edges; can technically break the containment invariant for UTC+n reviews written just after local midnight. Fine to accept at date granularity; worth a code comment.
- **Search scan cap:** `/feedback/search` scans `min(days, 30)` import days (L449) but accepts `days` up to 90 — under review basis, reviews written in-window but imported 31+ days ago are missed. Inherited from the pre-existing cap.
- **AI Chat / Projects / MCP don't honor `date_basis`** (`shared/feedback.py` has no such param) — a user on review basis sees review-windowed dashboards but import-windowed chat answers. The PR description already flags this; recommend filing the follow-up issue now so the global picker's scope is tracked.
- PR description nit: "Data Explorer stats" — there's no stats query in `useDataExplorerQueries`; what's threaded is the feedback/sources queries. Cosmetic.

---

Everything else verified clean: the render-phase-adjustment lint refactors are behavior-preserving (ProductTab's saved mode now applies on first render — an improvement), the runtimeConfig fallback change is well-tested with negligible production risk (`config.json` short-circuits it), and the mock-server changes mirror the real semantics. Nice work on the containment argument — it genuinely avoids a GSI + backfill.

perrozzi:
Post-#139 compatibility check: **git merge is clean** (verified via `merge-tree` against the new `development` tip `ec4231d` — zero file overlap with #139), **but there's one semantic conflict with the i18n gate that #139 just fixed.**

**What breaks:** `npm run --prefix voc-datalake/frontend i18n:validate` exits 1 on the merged tree (`development` @ `ec4231d` + this branch): ~15 empty + 1 missing `product.upload.*` keys reported in **`common.json`** for de/es/fr/ja/pt/zh. Dev is currently green on this gate for the first time since #132 (fixed in #139), so merging this branch as-is re-reddens it. It slipped through because `npm run check` doesn't include `i18n:validate`, and this branch's base (`e201270`) predates the fix.

**Root cause:** the `ProductDocsUpload.tsx` extraction passes `t` down as a prop:

```tsx
// ProductDocsUpload.tsx L19, L29
type TFunc = (key: string, opts?: Record<string, unknown>) => string
export function DocsUpload({ projectId, t }: { readonly projectId: string; readonly t: TFunc }) {
```

With no `useTranslation('projectDetail')` call in the file, i18next-parser can't infer the namespace for the `t('product.upload.*')` calls and attributes them to the default namespace (`common`) — where they don't exist. The keys are correctly translated in `projectDetail.json` (all 8 locales, filled by #139); the parser just looks in the wrong file.

**Fix (small):** drop the `t` prop and let the component own its namespace:

```tsx
import { useTranslation } from 'react-i18next'

export function DocsUpload({ projectId }: { readonly projectId: string }) {
  const { t } = useTranslation('projectDetail')
  ...
```

Then remove `t={t}` at the call site in `ProductTab.tsx` and delete the `TFunc` type. This also matches the codebase convention — every other extracted component calls `useTranslation` itself.

**Verify with:** `npm run --prefix voc-datalake/frontend i18n:validate` (should print "All translations are in sync", exit 0) after rebasing on / merging current `development`. Heads-up: the validator's parser step rewrites locale JSON as a side effect — discard any resulting churn under `frontend/public/locales/` rather than committing it.

mundurragacl:
All three should-fix items, both nits, and the i18n semantic conflict are resolved as of fa2b91e (4 commits). Current `development` (ec4231d) is merged into the branch; the full gate is green on the merged tree: eslint 0, all typechecks clean, 1,582 frontend + 741 backend tests pass, `i18n:validate` exits 0 ("All translations are in sync").

**i18n conflict** → `189445a`: `DocsUpload` and `DocStatusBadge` now call `useTranslation('projectDetail')` themselves; the `t` prop and `TFunc` type are gone. Locale churn from the validator was discarded as you suggested.

**Item 1 (silent degradation)** → `91b2971`: took your option 2 *and* option 1 together. New `_query_partition` helper pages every date partition via `LastEvaluatedKey` (10k-rows-per-partition ceiling), and every raw-scan metrics response (`/metrics/summary`, `sentiment`, `categories`, `sources`, `personas`, `/feedback/entities`) now carries `is_partial`. The Dashboard prefixes Total Feedback / Urgent Issues with `~` plus a `title` tooltip and `aria-label` when set. Your suggested >budget test exists as `test_summary_flags_partial_when_scan_truncated` (plus a not-partial counterpart and an aggregates-are-never-partial check).

**Item 2 (total divergence)** → same commit: the raw-scan budget now defaults to `MAX_FEEDBACK_OFFSET`, so `/feedback` and `/metrics/*` describe the same window. `/feedback/search` deliberately keeps its smaller sampling budget (commented at the call site).

**Item 3 (category branch)** → same commit: the category GSI is paged up to the candidate cap through the same helper, and `is_partial_window` now trips whenever paging stopped with rows remaining — covered by `test_category_branch_pages_past_the_first_page`.

**Nit 1** → the misleading `shared/feedback.py` comparison in the comment is gone; the urgent/search `days` vs `days-1` unification is deferred as you suggested (behavior change on the import basis) — happy to file the follow-up issue.

**Nit 2** → `_basis_date` now shape-checks `source_created_at` against `^\d{4}-\d{2}-\d{2}$` and falls back to the import date; `test_non_date_string_falls_back_to_import_date` asserts garbage strings neither win lexicographic comparisons nor become bucket keys. Your timezone-skew observation is now a code comment on the same function.

Bonus (from the issue-triage pass, same push): this branch also closes the paging half of #99 for the metrics handler — the source filter is pushed down as a server-side `FilterExpression`, so a partition dominated by one source can't starve source-filtered queries (`TestDominatedPartitionSourceFilter`, 4 tests). #103 (BedrockAccessStack hard failure) and #95 (Prioritization stale scores) are also fixed in `fa2b91e` / `ed4f840`.

Not done: the AI Chat / Projects / MCP `date_basis` follow-up issue you recommended — will file it next.

mundurragacl:
PR description updated with an **Issues addressed** section: this PR fixes #99 (dominated-partition starvation, metrics handler), #103 (BedrockAccessStack failure on accounts with existing Anthropic access), and #95 (stale prioritization scores), alongside the original date-basis feature, the review should-fixes/nits, and the #139 i18n conflict. Since the PR targets `development` (not the default branch), those three issues will need manual closing when this merges.

PR #151, opened 2026-07-14
fix: enforce Amazon PR/FAQ format in generation prompts

### LINKED ISSUE #93: [Improvement]: PRFAQ generation doesnt really follow the PRFAQ format
### URL / Page

PRFAQ generation

### Summary

when genrationgin prfaq doesn;'t actaullt follow the format.  this is anecdotal no data, but we need ensure the prompt that gneates this is current. 

### Steps to Reproduce

navigate the to prfaq geneation, then generate

### Expected Behavior

aws formatted prfaq

### Actual Behavior

unknown

### Affected Area

Frontend - Dashboard

### Affected Files (if known)

_No response_

### Additional Context

_No response_

mundurragacl:
## Summary

Fixes #93 (please close manually on merge — base is `development`).

The PR/FAQ chain produced good *content* with an unconstrained *shape*: preambles, duplicated section titles (the assembler already adds `Press Release` / `Customer FAQ` / `Internal FAQ` headings), and echoed instruction labels rendered as literal text. Each prompt step now carries a strict output contract:

- markdown-only body — no preamble, meta-commentary, or code fences
- no duplicate section headings (assembler owns those)
- press release reads as unlabeled narrative: single `###` headline, italic subheadline, `CITY — date —` dateline, problem/solution/quotes as prose, bold lead-ins for How it works / Availability / Getting started
- FAQs pinned to `**Q: …**` + answer paragraphs; internal FAQ must include two genuinely hard VP-level questions answered with the accepted tradeoff named

Prompt file (`prfaq-generation.json`, v1.1.0) plus one loader hardening line (`load_prompt_file` now opens with explicit `encoding=.utf-8.` — the file carries em dashes and the default encoding is locale-dependent outside Lambda). No handler or infrastructure changes; existing `{placeholder}` plumbing untouched.

## Testing

- 6 contract tests (`TestPrfaqPromptContract`): step list matches the chain builder's order, every placeholder is supplied by `get_prfaq_generation_steps` (`format_prompt` silently passes typos through as literal text), format bans present, end-to-end run through the builder (which delegates to `build_chain_steps`) leaves only `{previous}` unresolved
- `lambda/shared/test/test_prompt_utils.py`: 30 passed, 1 pre-existing skip

## Notes

- Output quality is LLM-behavioral; the contract language is the standard mitigation but the real verification is a generation run after deploy. Suggest the issue reporter regenerates one PR/FAQ post-merge.


perrozzi:
This PR tightens the PR/FAQ generation prompts (`prfaq-generation.json`, bumped to 1.1.0) with strict output contracts — no preambles, no duplicated section headings (the assembler owns those), pinned `**Q: …**` FAQ format, and an unlabeled press-release narrative — and adds five contract tests pinning the prompt file to the chain builder's expectations. Prompt-file-only change with no handler, IAM, or Bedrock model-id implications. Overall this looks clean and well-scoped; the tests would fail against the pre-fix prompt file, satisfying the regression-test rule.

### Should fix

- `lambda/shared/test/test_prompt_utils.py`: the first four tests load the JSON via a hardcoded relative path (`parents[2] / 'api' / 'prompts'`), bypassing the production loader (`get_prompts_dir` / `load_prompt_file`). If the prompts directory is ever relocated or resolved differently at runtime, these tests could keep passing against a stale file. Consider resolving the path through `get_prompts_dir()` (as the end-to-end test does via monkeypatch) so all five tests exercise the same file production reads.

### Nits

- `test_prompt_utils.py`: three blank lines before `TestPrfaqPromptContract`; PEP 8 calls for two.
- Imports (`json`, `re`, `pathlib.Path`) inside test methods — hoist to module level for consistency with the rest of the file.
- `test_steps_forbid_duplicate_section_headings` matches exact casing (`'Do NOT add'`); a case-insensitive check (as done for the preamble ban) would be less brittle to benign copy edits.
- The `customer_faq` step's rewording dropped the explicit "Format as Q&A pairs" from the user prompt in favor of the system-prompt contract — fine, but worth confirming the model honors system-prompt formatting alone in a post-merge generation run, as the description already suggests.

As the description notes, output quality is LLM-behavioral, so a post-deploy regeneration of a real PR/FAQ is the true verification — the contract tests only pin the prompt text, not the model's compliance.

---

mundurragacl:
Review addressed in the latest commit: all five contract tests now load through the production loader (`get_prompts_dir` routed at the repo prompts dir via an autouse fixture + `load_prompt_file` with cache clearing), so a relocated prompts directory fails the tests instead of silently passing against a stale file. Nits done too: imports hoisted to module level, two-blank-line spacing, case-insensitive heading-ban check. The `customer_faq` formatting concern stands as described — the system-prompt contract pins the `**Q:**` shape, and the post-merge generation run remains the real verification. 30/31 prompt-utils tests pass (1 pre-existing skip), ruff clean.

perrozzi:
This PR tightens the four PR/FAQ generation prompts (`voc-datalake/lambda/api/prompts/prfaq-generation.json`) with explicit output contracts — no preambles, no duplicated section headings (the assembler owns those), a fixed press-release shape, and pinned `**Q: …**` FAQ formatting — and adds five contract tests that load the prompt file through the production loader. It's a prompt-file-only change with a correct version bump (1.0.0 → 1.1.0), no handler/IAM/infrastructure impact, and the tests give real regression coverage: reverting the prompt changes fails `test_steps_forbid_duplicate_section_headings`. Overall this looks clean and well-scoped.

### Should fix
- `TestPrfaqPromptContract.SUPPLIED_PLACEHOLDERS` in `lambda/shared/test/test_prompt_utils.py` hardcodes a copy of the chain builder's kwargs. If `get_prfaq_generation_steps` gains or renames a parameter, the set silently drifts and the per-step placeholder test can pass against stale assumptions. Consider deriving the set from the builder (e.g. via `inspect.signature(get_prfaq_generation_steps)` plus `previous`/internally-generated `launch_date`), or at minimum add a comment cross-referencing the builder so both change in lockstep. The end-to-end test partially mitigates this, but only for placeholders that reach the formatted output.
- `test_chain_builder_formats_prfaq_steps_cleanly` calls `get_prfaq_generation_steps` without `product_context`, relying on it having a default. Verify the default renders sensible text into the `{product_context}` slot (not an empty string that leaves "CURRENT PRODUCT CONTEXT:" dangling) — the test only asserts no unresolved braces, not that the section is coherent. Worth one extra assertion that `launch_date` and `product_context` slots are non-empty in the formatted output.

### Nits
- `test_steps_forbid_duplicate_section_headings` asserts on the literal substrings `'do not add'` and `'no preamble'`. This is intentional pinning, but a legitimate rewording (e.g. "Never add") breaks the test without a behavioral regression — a small regex alternation would be more resilient.
- The customer FAQ instruction "give the most customer-obsessed reasonable answer rather than inventing conflicting specifics" mildly invites the model to fabricate pricing/availability answers where the release is silent. Consider steering silent-topic answers toward "details announced at launch"-style responses instead.
- Internal FAQ count changed from 5–7 to 6–8 with `max_tokens` unchanged (8000); likely fine, but worth confirming completions aren't truncated on the longer answers now requested.
- The fixture docstring says the local fallback in `get_prompts_dir` "doesn't resolve from the test cwd" — if that's true, it may be worth a follow-up to make the loader cwd-independent rather than monkeypatching in every prompt test class.

The stated goal (output shape) is LLM-behavioral and can't be fully verified by these tests; as the description notes, a post-deploy generation run is the real check — reasonable given the mitigation is standard contract language.

---

mundurragacl:
Round 2 addressed in b29ef67:

- **Placeholder drift (should-fix)**: `supplied_placeholders` is now derived from `inspect.signature(get_prfaq_generation_steps)` — minus `response_language` (consumed as a system-prompt instruction, never a template slot), plus the internally-generated `launch_date` and executor-substituted `previous`. A renamed/added builder parameter now changes the test's expectations automatically.
- **Slot coherence (should-fix)**: the end-to-end test asserts the default `product_context` sentence ('(No product context provided.)') actually renders and the `launch_date` slot carries a real `YYYY-MM-DD` — verified non-dangling, not merely brace-free.
- **Nits**: heading-ban assertion accepts 'do not add' **or** 'never add'; the customer-FAQ silent-topic instruction now steers to "details will be announced at launch" instead of inviting invented pricing/availability specifics (prompt v1.1.1 — good catch, the previous wording did mildly license fabrication).

Acknowledged, no change: the 6-8 internal FAQs fit comfortably in 8000 max_tokens (~700 tokens/answer headroom at the top end; the continuation logic in `converse` also guards truncation). The cwd-independent `get_prompts_dir` idea is a good follow-up — it would remove the monkeypatch from every prompt test class; happy to file it.

Suite: 30 passed / 1 pre-existing skip, ruff clean.

perrozzi:
This PR tightens the four prompt steps in `voc-datalake/lambda/api/prompts/prfaq-generation.json` with strict output contracts (markdown-only, no duplicated section headings, pinned Q&A format) and adds a `TestPrfaqPromptContract` class in `voc-datalake/lambda/shared/test/test_prompt_utils.py` that pins the prompt file to the chain builder's contract. It's prompt-and-test only — no handler, IAM, or Bedrock model-id changes — and the tests genuinely act as regression tests (the heading/preamble ban assertions fail against the old v1.0.0 file). Overall this looks clean and well-scoped.

### Should fix

- **Test coupling to an exact production default string** (`test_prompt_utils.py`, `test_chain_builder_formats_prfaq_steps_cleanly`): asserting `'(No product context provided.)' in first_step` hard-codes a copy string from `get_prfaq_generation_steps`. Any harmless rewording of that fallback message breaks this test without any behavioral regression. Consider asserting only that the `PRODUCT CONTEXT` section is non-empty after formatting, or import the default constant if one exists.
- **`supplied_placeholders` silently trusts new builder params are template slots**: deriving the set from `inspect.signature(get_prfaq_generation_steps)` means a future parameter that (like `response_language`) is *not* a template slot would be wrongly treated as "supplied", weakening `test_every_placeholder_is_supplied_by_the_chain_builder`. A brief comment is there, but an explicit allowlist assertion (fail loudly when the signature gains an unrecognized param) would make the drift visible instead of silent.

### Nits

- Version mismatch: the PR description says v1.1.0 but `prfaq-generation.json` says `"version": "1.1.1"`. Align one or the other.
- `test_steps_forbid_duplicate_section_headings` does `import re as _re` inside the method even though `re` is already imported at module level — drop the shadow import.
- `test_every_placeholder_is_supplied_by_the_chain_builder` only scans `user_prompt_template`; system prompts contain no placeholders today, but scanning both fields would future-proof the typo guard the docstring promises.
- The regex `r'(do not|never) add'` in the heading-ban test is loosely coupled to prompt wording; a rephrase like "must not add" would fail the test despite the ban still being present. Acceptable for a pin test, but worth knowing it's wording-sensitive.
- The prompt file mixes em dashes and typographic quotes in JSON strings — fine functionally, just confirm the loader reads the file as UTF-8 in the Lambda runtime (it presumably already does for v1.0.0's existing em dashes).

---

mundurragacl:
Round 3 addressed in 7a3f7de:

- **Copy-string coupling (should-fix)**: the coherence assertion now derives the fallback from `inspect.signature(get_prfaq_generation_steps).parameters['product_context'].default` — rewording the fallback message no longer breaks the test, while the assertion stays exact.
- **Silent-trust drift (should-fix)**: replaced the signature-derived set with explicit `KNOWN_SLOT_PARAMS` / `KNOWN_NON_SLOT_PARAMS` classification plus a dedicated drift test: an unclassified new builder parameter (or a classified one that disappears) fails loudly with instructions, exactly as you suggested.
- **Nits**: PR description aligned to v1.1.1; shadow `import re as _re` dropped; the placeholder scan now covers system prompts too; heading-ban regex accepts 'must not add'; and `load_prompt_file` now opens with explicit `encoding='utf-8'` — you were right to poke at that: it worked in Lambda only because the runtime locale is UTF-8, and the default is locale-dependent everywhere else (dev machines, CI images).

Wording-sensitivity of the heading-ban pin is acknowledged as inherent to a pin test. Suite: 31 passed / 1 pre-existing skip across shared tests, ruff clean.

perrozzi:
This PR tightens the PR/FAQ generation prompts (v1.1.1 of `prfaq-generation.json`) with strict output contracts — markdown-only bodies, no duplicated section headings, pinned `**Q: …**` FAQ format — adds an explicit `encoding='utf-8'` to the prompt loader, and adds a `TestPrfaqPromptContract` suite that pins the prompt file to the chain builder's contract. The change is well-scoped (prompt file + loader hardening + tests, no handler/IAM/model-id changes), and the contract tests give a real regression signal for what is otherwise an LLM-behavioral fix. Overall this looks clean and mergeable.

### Should fix
- **Contract tests pin proxy phrases, not the actual bans** (`lambda/shared/test/test_prompt_utils.py`): `test_steps_forbid_duplicate_section_headings` asserts `re.search(r'(do not|never|must not) add', system)` on the whole lowered system prompt. The `press_release` prompt also contains other "do not …" phrasing, so a future edit that removes the heading ban but keeps, e.g., a "do not add code fences" line would still pass. Assert the specific artifact instead — e.g. that the system prompt for `customer_faq` mentions not adding a `"customer faq"` heading, and similarly for the other two steps.

### Nits
- PR description says v1.1.0 but the file diff bumps `1.0.0 → 1.1.1`; align the description (or the version, if 1.1.1 was accidental).
- `voc-datalake/lambda/shared/prompts.py`: the `encoding='utf-8'` fix is correct and now load-bearing (the file carries em dashes and typographic quotes), but nothing tests it — a tiny test opening the real file under a forced non-UTF-8 locale is hard, so at minimum the contract tests exercising the real file via the production loader partially cover it; worth noting in the docstring of the fixture.
- `test_chain_builder_formats_prfaq_steps_cleanly` asserts `\d{4}-\d{2}-\d{2}` for the launch date, hard-coding the builder's date format into the prompt test; if the builder ever switches format the failure message ("launch_date slot empty") will be misleading.
- The `_route_loader_at_repo_prompts` fixture is autouse for the whole class but only needed by the tests that call `_load()`; harmless, though `test_builder_signature_has_no_unclassified_parameters` clears the `load_prompt_file` cache for no reason.
- `supplied_placeholders` as a `@property` on a test class works but is unusual pytest style; a plain classmethod or module-level constant would be more conventional.

No blockers: no IAM/Bedrock/handler changes, no secrets, no ingestor-plugin registration implications, and the bug fix ships with tests that fail if the prompt contract language is reverted.

---

mundurragacl:
Round 4 addressed in the latest commit:

- **Proxy-phrase pinning (should-fix)**: `test_steps_forbid_their_own_section_heading` now asserts each step bans **its own** assembler-owned heading — the regex requires the ban verb and the specific heading ('press release' / 'customer faq' / 'internal faq') in the same sentence, so a surviving 'do not add code fences' line can no longer satisfy a removed heading ban. Exactly the artifact-level pin you suggested.
- **Nits**: the PR description already reads v1.1.1 (updated last round — the bot may have reviewed a cached body); `SUPPLIED_PLACEHOLDERS` is a plain class attribute now; the fixture docstring documents both the deliberate class-wide autouse (cache_clear is a no-op for non-loading tests; no test can read a stale copy) and the partial UTF-8 coverage it gives the loader; and the launch-date assertion message now names the expected `YYYY-MM-DD` so a builder format change reads as what it is rather than 'slot empty'.

31 passed / 1 pre-existing skip, ruff clean.

perrozzi:
This PR tightens the four PR/FAQ prompt steps with explicit output contracts (no preambles, no duplicate section headings, pinned `**Q:**` FAQ format, unlabeled press-release narrative), bumps the prompt file version, adds an explicit `encoding='utf-8'` to `load_prompt_file`, and pins the prompt file to the chain builder's contract with a new test class. The change is well-scoped, the tests are thoughtfully designed (loading through the production loader, failing loudly on unclassified builder parameters, asserting the specific banned heading rather than a proxy phrase), and reverting the JSON would fail the new tests, satisfying the regression-test rule. Overall this looks clean and mergeable.

### Nits
- **Version jump**: `1.0.0` → `1.1.1` skips `1.1.0` with no explanation in `voc-datalake/lambda/api/prompts/prfaq-generation.json`. If `1.1.0` exists elsewhere, fine; otherwise `1.1.0` would be the expected bump.
- **PR description drift**: the description says "Prompt-file-only change ... no handler or infrastructure changes," but `voc-datalake/lambda/shared/prompts.py` is also modified (the UTF-8 fix). It's a good change — just update the description so the shared-code change isn't overlooked.
- **Test count**: the description says "5 new contract tests" but `TestPrfaqPromptContract` contains 6 test methods.
- **Placeholder regex**: `\{(\w+)\}` in `test_every_placeholder_is_supplied_by_the_chain_builder` won't catch malformed placeholders like `{launch date}` or `{launch_date }`, which would also reach the LLM as literal text. A broader pattern (e.g. `\{([^{}]+)\}`) would cover typo classes the test's own docstring calls out.
- **Cross-package path coupling**: the fixture in `voc-datalake/lambda/shared/test/test_prompt_utils.py` reaches into `lambda/api/prompts` via `parents[2]`. Acceptable given the docstring justification, but a moved prompts directory will fail with a `FileNotFoundError` from the loader rather than a self-describing assertion — a small `assert repo_prompts.exists()` in the fixture would make the failure mode obvious.

Note: the real behavioral verification (whether the model actually respects the output contract) is deferred to a post-merge generation run, as the PR itself acknowledges — the tests pin the prompt text, not the LLM output. That's the right standard mitigation here; just make sure the regeneration check actually happens.

---

mundurragacl:
Round 5 (nits) addressed in the latest commit + description edits:

- **Placeholder regex**: broadened to `\{([^{}]+)\}` — malformed slots like `{launch date}` now fail the guard, matching the typo class the docstring promises (good catch: `format_prompt` leaves those as literal LLM text via its KeyError fallback).
- **Fixture**: asserts the repo prompts directory exists with a self-describing relocation message before routing the loader.
- **Description drift**: fixed — it now names the loader UTF-8 hardening alongside the prompt file, and says 6 contract tests.
- **Version jump**: not accidental — 1.1.0 was this PR's first commit (52ce6ca) and 1.1.1 the round-3 prompt tweak, so both exist in this branch's history; only 1.1.1 lands on merge.

And agreed on the closing note: the post-merge regeneration check is the real verification — I'll run one PR/FAQ generation after this lands and report on the issue (#93). 31 passed / 1 pre-existing skip.

perrozzi:
This PR tightens the four PR/FAQ generation prompts with strict output contracts (no preambles, no duplicated section headings, pinned `**Q: …**` format, unlabeled press-release narrative), bumps the prompt file version, adds explicit UTF-8 encoding to `load_prompt_file`, and pins the whole contract with a well-designed test class. The change is prompt-and-test only — no handler, IAM, or infrastructure impact — and the tests genuinely act as regression guards: reverting the prompt changes fails `test_steps_forbid_their_own_section_heading` and `test_faq_steps_pin_the_qa_format`. This looks clean and I found no blockers.

Notable strengths:

- `voc-datalake/lambda/shared/prompts.py`: the `encoding='utf-8'` fix is correct and needed — the new prompt text is heavy on em dashes, and `open()`'s default encoding would break local test runs on Windows/cp1252.
- `test_prompt_utils.py`: routing the tests through the production loader (`get_prompts_dir` + `load_prompt_file`) with cache clearing, rather than reading the JSON directly, means the tests can't drift onto a stale copy. The unclassified-parameter drift check on `get_prfaq_generation_steps` is a nice loud-failure design.
- The old press-release template hard-coded an example date (`"SEOUL — 2026-09-15"`); its removal makes the `YYYY-MM-DD` assertion in `test_chain_builder_formats_prfaq_steps_cleanly` meaningful rather than vacuously satisfied by the example text.

### Nits

- `test_prompt_utils.py`: the end-to-end leftover check uses the narrow `\{(\w+)\}` pattern while `test_every_placeholder_is_supplied_by_the_chain_builder` deliberately uses the broad `\{([^{}]+)\}` for exactly the malformed-slot class (`{launch date}`). Consider using the broad pattern in both so a malformed slot introduced by the *builder's* formatting path is also caught end-to-end, not only in the raw templates.
- `prfaq-generation.json`: version jumps 1.0.0 → 1.1.1 with no 1.1.0 — if the versioning is meant to be meaningful, 1.1.0 (or a comment on the scheme) would be cleaner.
- `customer_faq` system prompt says "Match the language of the press release" while other steps say "Match the source language" — likely equivalent in practice, but harmonizing the phrasing avoids ambiguity if the feedback language and generation language ever diverge.
- The heading-ban regex `(do not|never|must not) add[^.]*<heading>` is coupled to sentence punctuation in the prompt text (an em dash or comma before a period keeps it working, but a rephrase with an early period breaks it silently at review time). A short comment noting this coupling in the JSON's `description` or the test would help future prompt editors.

Note: only the diff shown was reviewed; the `lru_cache` on `load_prompt_file` and the `get_prfaq_generation_steps` builder internals are assumed from context (`cache_clear()` usage, passing tests) rather than visible in the diff.

---

mundurragacl:
Round 6 (nits) addressed in the latest commit:

- **Regex symmetry**: the e2e leftover check now uses the same broad `\{([^{}]+)\}` pattern as the raw-template guard — a malformed slot that survives the builder's formatting path is caught end to end.
- **Language phrasing**: both FAQ steps harmonized to 'Match the source language.' (v1.1.2) — agreed the press-release-relative phrasing was ambiguous if feedback and generation languages ever diverge.
- **Punctuation coupling**: documented at the heading-ban regex so a future rephrase that splits verb and heading across a period updates the test knowingly.
- **Version scheme**: the intermediate versions (1.1.0, 1.1.1) are this PR's own iteration history — each review round that touched the prompt file bumped the patch. Only the final version lands on merge; the scheme is 'bump on any content change', which is why there's no gap in the branch history itself.

The lru_cache / builder internals you noted as assumed-from-context are exactly as assumed (cache on `load_prompt_file`, builder generates `launch_date` internally). 31 passed / 1 pre-existing skip.

perrozzi:
This PR tightens the PR/FAQ generation prompts with strict output contracts (no preambles, no duplicated section headings, pinned `**Q:**` FAQ format), hardens `load_prompt_file` with explicit UTF-8 encoding, and adds a solid contract test class that pins the prompt file to the chain builder. The change is scoped correctly (prompt JSON + one loader line + tests, no handler/infra changes), placeholders remain consistent with what `get_prfaq_generation_steps` supplies, and the tests load through the production loader rather than a stale copy. Overall this looks clean and well-considered.

### Nits

- `voc-datalake/lambda/shared/test/test_prompt_utils.py`: `SUPPLIED_PLACEHOLDERS` includes `previous` for all steps, so an accidental `{previous}` in the first step (`customer_thinking`), where the executor has nothing to substitute, would pass the placeholder guard. Consider restricting `previous` to steps after the first.
- The heading-ban regex in `test_steps_forbid_their_own_section_heading` couples the test to sentence-level prompt phrasing (`(do not|never|must not) add[^.]*<heading>`). The comment acknowledges this, but expect churn: any editorial rewording of the ban sentence breaks the test even when the behavioral intent is preserved.
- The UTF-8 encoding fix in `voc-datalake/lambda/shared/prompts.py` has no regression test that fails when reverted — CI almost certainly runs with a UTF-8 default locale, so loading the em-dash-bearing file passes either way. Probably not worth forcing (would require locale manipulation), but note the repo's bug-fix-needs-regression-test convention isn't fully met here.
- `prfaq-generation.json` version jumps 1.0.0 → 1.1.2 in a single change; a straight bump to 1.1.0 would be cleaner unless intermediate versions exist elsewhere.
- Minor prompt-consistency risk: the press-release system prompt forbids echoing "the section labels or numbering from the instructions," while the user prompt still presents a numbered list ("1. Headline — …"). Low risk with the explicit "unlabeled" framing, but worth watching in the post-merge generation run the PR description already suggests.

---

mundurragacl:
Round 7 addressed: the placeholder guard now excludes `previous` from the FIRST step's allowed set — a `{previous}` in `customer_thinking` (where the executor has no prior output to substitute) fails the guard instead of passing silently. Good catch on the ordering subtlety.

The remaining nits stand as acknowledged: the heading-regex phrasing coupling is documented at the assertion; the UTF-8 fix is exercised (not strictly regression-tested — forcing a non-UTF-8 locale in CI isn't worth the machinery, noted as a convention gap); the 1.0.0→1.1.2 jump is this branch's own review-iteration history (each round that touched the file bumped patch — only the final lands); and the numbered-list-vs-unlabeled tension is exactly what the post-merge generation run will confirm — the instructions explicitly frame the list as 'unlabeled — the prose itself carries the structure'.

31 passed / 1 pre-existing skip.

perrozzi:
This PR tightens the four PR/FAQ chain prompts with explicit output contracts (markdown-only, no preambles, no duplicate section headings owned by the assembler, pinned `**Q: …**` FAQ format), bumps the prompt file version, adds an explicit `encoding='utf-8'` to `load_prompt_file`, and pins the prompt file to the chain builder's contract with six new tests. No IAM, Bedrock model id, handler, or frontend changes, so none of the higher-risk repo rules are in play. Overall this is a clean, well-scoped change with unusually thoughtful test coverage — the placeholder-drift guard in particular closes a real silent-failure class (`format_prompt` passing typos through as literal text).

### Should fix

- **The UTF-8 encoding fix has no regression test that fails when reverted** (`voc-datalake/lambda/shared/prompts.py`). The repo convention is that bug fixes carry a test that fails if the fix is removed. The fixture comment in `test_prompt_utils.py` says loading the real file "partially covers" the encoding, but reverting `encoding='utf-8'` would only fail on machines with a non-UTF-8 default locale — typical CI (UTF-8) would stay green. A cheap direct guard: a test that patches/forces a non-UTF-8 default (e.g. `mock.patch('builtins.open')` asserting the `encoding` kwarg, or reading the file with `encoding='ascii'` expectations) so the fix is pinned on every platform.

### Nits

- Version jump `1.0.0` → `1.1.2` skips `1.1.0`/`1.1.1` with no visible history in the repo (`prfaq-generation.json`). If those were local iterations, `1.1.0` would be less confusing to the next reader.
- `test_steps_forbid_their_own_section_heading` couples the test to prose punctuation via `(do not|never|must not) add[^.]*<heading>` — the inline NOTE acknowledges it, but any innocuous rewording of the system prompts will need a paired test edit. Consider matching on the heading string plus a ban verb anywhere in the system prompt instead of within one sentence.
- `KNOWN_SLOT_PARAMS` / `KNOWN_NON_SLOT_PARAMS` duplicate knowledge that lives in `get_prfaq_generation_steps`; the drift test mitigates this well, but a comment in `prompts.py` pointing back at `TestPrfaqPromptContract` would help a future editor discover the classification requirement before CI does.
- The autouse fixture hard-asserts the prompts directory location (`parents[2] / 'api' / 'prompts'`); the error message is good, but this is a second place (after `get_prompts_dir`'s own fallback) that encodes the path. Acceptable given the explicit failure message.
- The customer FAQ contract's "do NOT invent [pricing/dates]… say those details will be announced at launch" is a good anti-hallucination guard, but note it can conflict with the internal FAQ's "quantify where possible" when `{previous}` includes both — worth watching in the post-merge generation run the description already plans.

The prompt content changes themselves are behavioral and can only be verified by a generation run, as the PR description correctly notes — the contract tests pin everything that's mechanically pinnable.

---

mundurragacl:
Round 8 addressed:

- **UTF-8 pin (should-fix)**: taken — `TestLoadPromptFileEncoding` spies on `builtins.open` and asserts the `encoding='utf-8'` kwarg while delegating to the real loader, so reverting the fix fails on every platform including UTF-8 CI. (32 tests now.)
- **Version**: flattened to a single clean `1.0.0 → 1.1.0` — you're right that the intermediates were invisible history to the next reader; nothing has merged, so presenting one bump is strictly clearer.
- **Cross-reference**: `prompts.py` now points at the contract-test classification above `get_prfaq_generation_steps`.

Holding position on one nit, deliberately: relaxing the heading-ban match to 'heading + ban verb anywhere in the prompt' would reintroduce exactly the proxy-match risk flagged in the earlier round (a 'do not add code fences' line plus an incidental heading mention elsewhere would pass with the real ban deleted). Same-sentence coupling is the price of pinning the specific artifact; the inline NOTE tells future editors what to update. The FAQ quantify-vs-don't-invent tension is on the post-merge generation checklist alongside the unlabeled-structure check.

perrozzi:
This PR tightens the PR/FAQ generation prompts (v1.1.0) with strict output contracts — no preambles, no duplicated section headings (the assembler owns those), pinned `**Q: …**` FAQ format — and hardens `load_prompt_file` with explicit `encoding='utf-8'`, backed by a new contract test class and a platform-independent regression test for the encoding fix. Scope is contained (prompt JSON + one loader line + tests), placeholders in the new templates all remain within the set supplied by `get_prfaq_generation_steps` (`feature_idea`, `product_context`, `personas_context`, `feedback_context`, `launch_date`, `previous`), and there are no IAM, handler, or frontend changes. Overall this is a clean, well-tested change; no blockers.

### Should fix
- **`TestLoadPromptFileEncoding.test_opens_prompt_files_with_explicit_utf8`** (`voc-datalake/lambda/shared/test/test_prompt_utils.py`): the spy patches `builtins.open` globally and records the *last* call's `encoding`. Any incidental `open()` during the call (coverage/tracing tooling, future logging inside the loader) would overwrite `seen['encoding']` and produce a false failure — or, worse, a false pass if an unrelated open happened to use UTF-8. Filter the spy on the target path (e.g. only record when `str(file).endswith('prfaq-generation.json')`) so the assertion is unambiguously about the prompt-file open.

### Nits
- `test_steps_forbid_their_own_section_heading` pins the ban to a regex over prompt prose (`(do not|never|must not) add[^.]*<heading>` within one sentence). The inline comment acknowledges the coupling, but any copy edit to the system prompts — including translation-style rephrasing — breaks the test even when the ban is still present. Consider asserting on a small structured marker (e.g. a `banned_headings` key in the JSON) if these prompts get edited often.
- `test_every_placeholder_is_supplied_by_the_chain_builder` concatenates `user_prompt_template + system_prompt` without a separator; a placeholder split across the boundary can't occur today, but joining with `'\n'` is cheaper than reasoning about it later.
- The encoding regression test asserts the kwarg is passed rather than the behavioral failure (correct call, since cp1252 breakage is platform-specific), but a one-line `content.encode()` round-trip assertion on a known em-dash string in the loaded config would additionally verify the file itself decoded correctly.
- The NOTE comment above `get_prfaq_generation_steps` in `voc-datalake/lambda/shared/prompts.py` references the test by class name and path — fine, but it will silently rot if the test class is renamed; the drift test already fails loudly, so the comment could simply say "see the prompt contract tests".
- PR description says the end-to-end test goes through `build_chain_steps`, but `test_chain_builder_formats_prfaq_steps_cleanly` calls `get_prfaq_generation_steps` directly — harmless, just align the description.

---

mundurragacl:
Round 9 addressed: the spy is filtered to the `prfaq-generation.json` open (incidental `open()` calls can't overwrite the observation or fake a pass), plus an em-dash round-trip assertion on the loaded content; the placeholder scan joins fields with `\n`; and the `prompts.py` note references the contract tests generically. The structured `banned_headings` marker idea is a good one if these prompts start churning — deferring until they do, since today the same-sentence pin plus its NOTE is the simplest thing that pins the artifact. On the description: `get_prfaq_generation_steps` IS the `build_chain_steps` path (it delegates internally), so the e2e claim holds — reworded anyway for precision.

32 passed / 1 pre-existing skip. At this point every should-fix across nine rounds is resolved; remaining notes are acknowledged trade-offs. From my side this is ready to merge.

perrozzi:
This PR tightens the PR/FAQ generation prompts (`prfaq-generation.json`, bumped to v1.1.0) with strict output contracts to stop preambles, duplicated section headings, and echoed instruction labels; hardens `load_prompt_file` with explicit `encoding='utf-8'`; and adds a solid set of contract tests pinning the prompt file to the chain builder's behavior. No handler, IAM, or infrastructure changes. Overall this is a clean, well-tested change — the tests go through the production loader, guard against placeholder typos (which `format_prompt` would otherwise pass through silently), and include a platform-independent regression pin for the encoding fix that fails on revert regardless of CI locale. No blockers found.

### Should fix

- `voc-datalake/lambda/shared/test/test_prompt_utils.py` — `test_has_the_four_chain_steps_in_code_order` hardcodes the step order independently of the chain builder. Consider deriving the expected order from `get_prfaq_generation_steps(...)` output (`step['step_name']`) so the pin tracks the code rather than a second hand-maintained copy. The end-to-end test already asserts `len(steps) == 4` but not that names match the JSON keys.

### Nits

- `test_opens_prompt_files_with_explicit_utf8`: `seen['encoding']` raises `KeyError` (not a readable assertion) if the spy never fires — prefer `assert seen.get('encoding') == 'utf-8'` or an explicit `assert 'encoding' in seen, "prompt file was never opened"` first.
- The encoding test uses both `monkeypatch.setattr(builtins, 'open', ...)` and a `try/finally` — `monkeypatch` already restores; the `finally` is only needed for the `cache_clear`, which the comment could say to avoid confusion.
- `test_steps_forbid_their_own_section_heading`'s `[^.]*` regex coupling to sentence punctuation is acknowledged in a comment, but it makes routine prompt rewording a test-editing exercise; a simpler "both the ban verb and heading appear in the system prompt" assertion would be less brittle with nearly the same protection.
- `prfaq-generation.json`: the press-release contract says the headline is "the only heading in the release," while items 8–10 use bold lead-ins — consistent, but consider also banning `##`/`####` explicitly since models frequently emit them for "How it works."
- The `NOTE` comment above `get_prfaq_generation_steps` in `prompts.py` references the test file path; if the tests ever move, this comment will go stale — the drift test already fails loudly, so the comment could just say "see TestPrfaqPromptContract."

Nice touches: routing the loader at the repo prompts dir via the production `get_prompts_dir` seam, classifying builder params as slot vs. non-slot with a drift guard, and disallowing `{previous}` only in the first step.

---

mundurragacl:
Round 10 addressed — with one deliberate deviation from the letter of the should-fix:

- **Step-order pin**: deriving the EXPECTED order from `get_prfaq_generation_steps` output would make the test circular (builder-vs-builder) and pin nothing. What the ask is really after — that the JSON keys and the builder's emitted `step_name`s can't drift apart — is now asserted directly in the e2e test: `[s['step_name'] for s in steps] == list(json steps)`. The hardcoded-order test remains as the human-intent pin.
- **Nits taken**: readable spy failure (`'prompt file was never opened through the spy'`) before the encoding assert; the `finally`'s purpose (lru_cache hygiene — monkeypatch handles restoration) documented; the press-release contract now explicitly bans `##`/`###`/`####` anywhere in the release; the `prompts.py` NOTE names `TestPrfaqPromptContract` without the rename-prone path.
- **Held**: the same-sentence heading-ban regex stays (third consecutive round on this one) — 'verb and heading anywhere in the prompt' is the proxy-match false-pass we closed in round 4. It's a pin test; pins couple to what they pin.

32 passed / 1 pre-existing skip. Ten rounds in, every should-fix ever raised on this PR is resolved or answered with rationale; I'm calling this converged and merge-ready from my side.

perrozzi:
This PR tightens the PR/FAQ generation prompts (`voc-datalake/lambda/api/prompts/prfaq-generation.json`, v1.0.0 → v1.1.0) with strict output contracts per step to stop preambles, duplicated section headings, and echoed instruction labels; adds explicit `encoding='utf-8'` to `load_prompt_file` in `voc-datalake/lambda/shared/prompts.py`; and adds a thorough contract test class plus a platform-independent regression test for the encoding fix. Scope is prompt content + one loader line + tests — no handler, IAM, or infrastructure changes, and the `{placeholder}` plumbing is untouched. Overall this looks clean and well-tested: the placeholder guard (catching typo'd slots that `format_prompt` would pass through as literal text) and the signature-drift test are genuinely useful pins, and the encoding test correctly fails on revert regardless of the CI platform's locale.

I verified the new prompt text against the tests: all placeholders in the updated templates (`{feature_idea}`, `{previous}`, `{launch_date}`, etc.) are supplied by `get_prfaq_generation_steps`/the chain executor, `{previous}` is correctly disallowed in the first step, and the heading-ban / `**Q:**` format assertions match the new system-prompt wording. No blockers.

### Nits

- `test_steps_forbid_their_own_section_heading` pins natural-language phrasing via regex (`(do not|never|must not) add[^.]*<heading>` within one sentence). The inline comments acknowledge the coupling, but expect churn: any editorial rephrase of the prompts will require a test update even when the intent is preserved. Consider a lighter structural check (e.g. the heading string appears near a ban keyword) if this starts flaking on prompt edits.
- `test_chain_builder_formats_prfaq_steps_cleanly` asserts `\d{4}-\d{2}-\d{2}` in the rendered press-release step. This works now because the new template dropped the old literal example date (`SEOUL — 2026-09-15`), but if a future edit re-adds a literal date example the assertion passes vacuously without verifying `{launch_date}` rendered. Matching the actual generated date would be more robust.
- In `TestLoadPromptFileEncoding`, `seen['encoding']` raises `KeyError` (still a failure, but a less readable one) if the spy never observes the open — `seen.get('encoding')` with the same assertion gives a clearer failure message.
- Monkeypatching `builtins.open` globally works here since the spy delegates, but patching `shared.prompts.open` (module-level name resolution) would scope the spy to the code under test and avoid intercepting unrelated opens at all.
- The `NOTE` comment on `get_prfaq_generation_steps` referencing `TestPrfaqPromptContract` is a nice cross-reference; consider also naming the specific test (`test_builder_signature_has_no_unclassified_parameters`) so future readers find it faster.

---

mundurragacl:
Round 11 — the remaining live nits are in (exact builder-generated launch date instead of any date-shaped match, so a future literal example can't satisfy it vacuously; the spy patches `shared.prompts.open` at module scope so unrelated opens are never intercepted; the NOTE names `test_builder_signature_has_no_unclassified_parameters`). Two were stale re-reads of already-landed changes (the `seen` KeyError readability and the spy-scoping concern were addressed in rounds 9–10's diffs). The heading-ban phrasing coupling is acknowledged for the third time as the deliberate cost of an artifact pin.

32 passed / 1 pre-existing skip, ruff clean. Eleven rounds, zero open should-fixes — this is converged; further rounds are yielding stale or repeat observations, so I'm treating the PR as merge-ready unless a human reviewer disagrees.

perrozzi:
This PR tightens the PR/FAQ generation prompts (v1.1.0) with strict output contracts to stop preambles, duplicated section headings, and echoed instruction labels; hardens `load_prompt_file` with explicit UTF-8 encoding; and adds a solid contract-test suite pinning the prompt file to the chain builder. The change is prompt/test-only — no handler, IAM, or infrastructure impact — and it follows the repo's regression-test rule (the encoding test fails on revert regardless of platform locale). Overall this looks clean and well thought out.

### Should fix

- **Midnight-boundary flake in `test_chain_builder_formats_prfaq_steps_cleanly`** (`lambda/shared/test/test_prompt_utils.py`): `expected_launch` is computed *after* `get_prfaq_generation_steps()` runs. If the test straddles a UTC day boundary (or if the builder uses local time rather than `timezone.utc`), the two dates differ by a day and the test fails spuriously. Compute the expected date both before and after the builder call and accept either, or assert with a ±1-day tolerance.

### Nits

- `test_steps_forbid_their_own_section_heading` regex (`(do not|never|must not) add[^.]*<heading>`) tightly couples the test to prompt sentence punctuation. The inline comment acknowledges this, but any copy-edit of the JSON will trip it — consider matching on presence of both the ban verb and heading within the system prompt rather than one sentence.
- In `test_opens_prompt_files_with_explicit_utf8`, `seen['encoding']` raises `KeyError` if the loader stops using bare `open()` (e.g., switches to `filepath.read_text()` / `Path.open`). It fails loudly either way, but `assert 'encoding' in seen, 'loader no longer uses open() — update spy'` would give a clearer failure message.
- The `NOTE` comment above `get_prfaq_generation_steps` in `lambda/shared/prompts.py` hardcodes a test method name; if the test is renamed the comment silently rots. Consider pointing at the test class only.
- `internal_faq` now asks for 6-8 FAQs plus two "hard VP questions" within the same 8000-token budget as before — likely fine, but worth confirming on a real generation run post-deploy, as the description already suggests.

Prompt placeholders all resolve against the builder's supplied slots (`feature_idea`, `personas_context`, `feedback_context`, `product_context`, `launch_date`, `previous`), the first step correctly avoids `{previous}`, and the UTF-8 change is a legitimate boundary hardening given the em dashes now in the file. No blockers.

---

mundurragacl:
Round 12: the midnight-boundary flake is fixed — the expected launch date is sampled before AND after the builder call and either satisfies the assertion, so a run straddling UTC midnight can't flake (the builder computes in `timezone.utc`, confirmed). Good catch; that one was real.

The heading-ban regex nit is acknowledged for the fourth consecutive round with the same rationale: within-sentence coupling is the deliberate cost of pinning the specific artifact — the 'verb + heading anywhere' relaxation is the proxy-match false-pass this same review flagged in round 4.

32 passed / 1 pre-existing skip. Every should-fix across twelve rounds is resolved; treating this PR as converged and merge-ready.

perrozzi:
This PR tightens the PR/FAQ generation prompts with explicit output contracts (no preambles, no duplicated section headings owned by the assembler, pinned `**Q: …**` FAQ format), bumps the prompt file to v1.1.0, hardens `load_prompt_file` with explicit `encoding='utf-8'`, and adds a solid set of contract tests plus a platform-independent regression test for the encoding fix. Content-only prompt changes with no handler, IAM, or Bedrock model-id changes, so no infra lockstep concerns. Overall this is a clean, well-tested change.

Verified details:
- `voc-datalake/lambda/api/prompts/prfaq-generation.json`: all placeholders used (`{feature_idea}`, `{product_context}`, `{personas_context}`, `{feedback_context}`, `{launch_date}`, `{previous}`) are covered by the test's `SUPPLIED_PLACEHOLDERS`, and `{previous}` correctly does not appear in the first step.
- The heading-ban assertions in `test_steps_forbid_their_own_section_heading` match the actual (lowercased) system prompt phrasing for all three steps, including the `no preamble` check.
- `TestLoadPromptFileEncoding` is a genuine regression test: it asserts the `encoding='utf-8'` kwarg directly, so it fails on any platform if the fix is reverted (a locale-dependent decode failure alone would only surface on non-UTF-8 platforms). The scoped `monkeypatch.setattr('shared.prompts.open', ..., raising=False)` shadowing of the builtin is the right approach.
- The dual before/after date sampling in `test_chain_builder_formats_prfaq_steps_cleanly` correctly avoids UTC-midnight flakes.

### Nits
- `test_steps_forbid_their_own_section_heading`'s `(do not|never|must not) add[^.]*<heading>` regex is coupled to sentence punctuation in the prompt text — an innocuous rewording that splits ban verb and heading across a period breaks it. The in-test comment acknowledges this, but a simpler check (ban verb and heading both present in the same system prompt) would be less fragile with little lost precision.
- The `NOTE` comment above `get_prfaq_generation_steps` in `voc-datalake/lambda/shared/prompts.py` hardcodes a test method name (`test_builder_signature_has_no_unclassified_parameters`); if the test is renamed the comment silently rots. Referencing the test class alone would age better.
- `TestLoadPromptFileEncoding.test_opens_prompt_files_with_explicit_utf8` will fail with a raw `KeyError` on `seen['encoding']` if the loader is ever refactored away from `open()` (e.g., to `Path.read_text`). Loud enough, but `assert seen.get('encoding') == 'utf-8', 'open() was never intercepted — loader refactored?'` would diagnose faster.
- The prompts-dir path resolution (`parents[2] / 'api' / 'prompts'`) is duplicated between the autouse fixture and the encoding test; a shared helper or module-level constant would keep them from drifting independently.

---

mundurragacl:
Round 13: took the one live nit — `REPO_PROMPTS_DIR` is a single module-level constant used by both the fixture and the encoding test.

Declining the rest with a convergence note: the `seen['encoding']` KeyError nit was fixed in round 10 (`assert 'encoding' in seen, 'prompt file was never opened through the spy'` precedes the value assertion — the reviewed diff shows it); the NOTE-comment nit now asks me to REMOVE the specific method name that round 11 asked me to ADD — I'm keeping round 11's version; and the heading-ban regex nit is on its fifth repetition with the same held rationale.

**Twelve should-fixes and ~30 nits over 13 rounds, all resolved or answered.** The review has reached oscillation (contradictory suggestions round-to-round) and stale re-reads, which is the natural stopping signal for an automated loop. This PR is converged — final state: 32 passed / 1 pre-existing skip, ruff clean, prompt v1.1.0. Handing to a human for the merge decision.

perrozzi:
The PR tightens the four PR/FAQ chain prompts with explicit output contracts (no preambles, no duplicate section headings, pinned Q&A format), bumps the prompt file to v1.1.0, adds `encoding='utf-8'` to `load_prompt_file`, and backs it all with a well-thought-out contract test class plus a platform-independent regression test for the encoding fix. This is a clean, low-risk change: prompt content plus one loader hardening line, no handler/IAM/infrastructure changes, and the tests load the real prompt file through the production loader rather than a copy. No blockers found.

### Should fix

- **`test_every_placeholder_is_supplied_by_the_chain_builder` only checks the "no unknown placeholder" direction** (`voc-datalake/lambda/shared/test/test_prompt_utils.py`). It won't catch a prompt edit that *drops* a required slot — e.g. if `{launch_date}` or `{feedback_context}` disappears from a template, the test still passes and the LLM silently loses grounding data. Consider also asserting that key placeholders are *present* per step (`launch_date` in `press_release`, the four context slots in `customer_thinking`, `previous` in steps 2–4).

### Nits

- `test_steps_forbid_their_own_section_heading` pins prompt *phrasing* via regex (`(do not|never|must not) add[^.]*<heading>`, same-sentence coupling). The comments acknowledge this, but any future copy-editing of the prompts will trip it; asserting on a structured field (e.g. a `banned_headings` key in the JSON) would be more durable than parsing English.
- The `customer_thinking` step gained a "no preamble / no meta-commentary" ban in the prompt, but it's excluded from `BANNED_HEADINGS` so nothing pins it — either cover it in `test_steps_forbid_their_own_section_heading`'s preamble check or note why it's intentionally unpinned.
- `KNOWN_NON_SLOT_PARAMS = {'response_language'}` and the drift test are a nice pattern, but the classification lives only in the test; the `# NOTE:` breadcrumb added above `get_prfaq_generation_steps` in `voc-datalake/lambda/shared/prompts.py` mitigates this — just make sure the referenced test name stays in sync if the class is renamed.
- The test comments are unusually verbose (multi-paragraph rationale on nearly every assertion). Helpful here, but consider trimming the ones that restate what the code plainly does.
- If anything downstream keys off the prompt file `version` (cache invalidation, telemetry), confirm `1.1.0` is picked up; nothing in the diff consumes it.

The i18n/frontend, IAM, ingestor-registration, and Bedrock model-ID rules don't apply to this diff. The regression test for the encoding fix correctly fails when the fix is reverted (asserts the kwarg, not platform behavior), matching the repo's bug-fix testing rule.

---

mundurragacl:
Round 14 addressed in eabb8c7:

- **Should-fix (reverse placeholder direction)**: agreed — added `REQUIRED_PLACEHOLDERS` (per-step map) and `test_steps_keep_their_required_placeholders`, asserting the four context slots in `customer_thinking`, `launch_date` + `previous` in `press_release`, and `previous` in both FAQ steps. Verified the guard fails when `{launch_date}` is dropped. Note `press_release` is grounded via `{previous}` rather than the raw context slots, so the map reflects what the file actually uses.
- **customer_thinking preamble nit**: covered — new `test_every_step_bans_preambles` pins the 'no preamble' instruction across all four steps, including `customer_thinking` which owns no section heading and was rightly excluded from `BANNED_HEADINGS`.

Declined (previously converged, restated for the record): the heading-ban regex vs structured `banned_headings` field, the NOTE-breadcrumb placement, and comment verbosity — all discussed in rounds 5–13; no new information changes those calls. On the `version` nit: nothing downstream consumes the prompt-file version today (grepped: no reader keys off it), so 1.1.0 is documentation-only.

34 passed / 1 skipped, ruff clean.

perrozzi:
This PR tightens the PR/FAQ generation prompts (v1.1.0) with strict output contracts — no preambles, no duplicated section headings (the assembler owns those), pinned `**Q:**` FAQ format, and an unlabeled narrative press release — plus a small hardening fix (`encoding='utf-8'` in `load_prompt_file`) and a substantial new contract test class that pins prompt structure to the chain builder. The change is well-scoped, the tests load through the production loader rather than raw file reads, and the encoding fix ships with a platform-independent regression test that fails when reverted (satisfying the repo's regression-test rule). No handler, IAM, or infrastructure changes. Overall this is a clean, careful change.

### Should fix

- `voc-datalake/lambda/shared/test/test_prompt_utils.py` (`TestLoadPromptFileEncoding.test_opens_prompt_files_with_explicit_utf8`): if the spy never intercepts the prompt-file open (e.g. the loader is refactored to use `filepath.open()` or `Path.read_text()`), `seen['encoding']` raises `KeyError` instead of a readable assertion failure. Add `assert 'encoding' in seen, "spy never observed the prompt-file open — loader stopped using open()?"` before the equality check. Note that a `Path.open()`/`read_text()` refactor would also silently bypass the spy entirely — worth the explicit guard.

### Nits

- `test_steps_forbid_their_own_section_heading` couples the assertion to sentence punctuation via `[^.]*` and to specific ban verbs (`do not|never|must not`). The inline comment acknowledges it, but this class of prompt-copy pin will make routine prompt rewording a two-file edit; consider matching on the heading string within the system prompt plus any negation token, more loosely.
- The placeholder-extraction pattern `re.findall(r'\{([^{}]+)\}', ...)` appears three times with duplicated rationale comments; extract a module-level helper (e.g. `_extract_slots(text)`) so the "broad on purpose" rationale lives in one place.
- The `NOTE:` comment on `get_prfaq_generation_steps` in `voc-datalake/lambda/shared/prompts.py` names a specific test function (`test_builder_signature_has_no_unclassified_parameters`); a test rename will silently orphan the pointer. Referencing the class (`TestPrfaqPromptContract`) alone would be more durable.
- `prfaq-generation.json`: the customer_faq contract says "5-7 customer FAQs" while internal_faq was expanded to "6-8" — fine, but confirm the assembler / any downstream length expectations don't assume symmetric section sizes.
- `test_chain_builder_formats_prfaq_steps_cleanly` does two `import inspect` calls' worth of concerns (placeholder resolution, default rendering, date rendering) in one test; splitting would give sharper failure localization, though the current comments mitigate this.

No blockers. The i18n rules don't apply here (no user-visible frontend strings), and the "Match the source language" instruction is correctly preserved in every rewritten prompt.

---

mundurragacl:
Round 15 addressed in 9ba9a83:

- **Should-fix (spy KeyError)**: agreed — added a pointed `assert 'encoding' in seen` with a message naming the likely cause (loader refactored to `Path.open()`/`read_text()`) before the equality check.
- **Helper extraction**: agreed — this became fair once round 14 added a third copy. Module-level `_extract_slots()` now owns the broad-pattern rationale; all three sites use it.
- **NOTE breadcrumb**: agreed — references `TestPrfaqPromptContract` (plus file path) instead of a renameable test function.

Declined:
- Heading-ban regex loosening: converged in earlier rounds — a looser 'heading string + any negation token' match would pass on sentences like 'the Press Release must not be empty'. Keeping the sentence-scoped pin with its inline caveat.
- FAQ count asymmetry (5–7 vs 6–8): checked — the assembler concatenates sections verbatim; nothing downstream assumes symmetric lengths. The internal FAQ is intentionally longer (it carries the hard-Q/risks material).
- Splitting `test_chain_builder_formats_prfaq_steps_cleanly`: it's one scenario (a single builder invocation) asserted at three points; splitting would triple the setup for marginal localization gain.

716 passed / 1 skipped, ruff clean on touched files.

perrozzi:
This PR tightens the four PR/FAQ prompt steps with strict output contracts (no preambles, no duplicated section headings, pinned Q&A format, unlabeled press-release narrative), bumps the prompt file to v1.1.0, hardens `load_prompt_file` with explicit `encoding='utf-8'`, and adds a substantial contract test suite plus a platform-independent regression test for the encoding fix. The change is well-scoped (prompt content + one loader line, no handler/IAM/infrastructure changes), and the tests genuinely fail if either the prompt contracts or the encoding fix are reverted — meeting the repo's regression-test expectation. Overall this looks clean and mergeable.

### Nits

- **Prompt-wording pins are brittle by design** — `test_steps_forbid_their_own_section_heading` requires the ban verb and heading to share one sentence (`[^.]*`), and `test_every_step_bans_preambles` greps for the literal phrase "no preamble". The author documents this coupling in comments, which helps, but future prompt editors will need to treat `test_prompt_utils.py` as part of the prompt file. Acceptable tradeoff; just be aware of the maintenance cost (`voc-datalake/lambda/shared/test/test_prompt_utils.py`).
- **`_extract_slots` will false-positive on future literal braces** — if a prompt ever includes a brace-bearing example (e.g., a JSON output-format instruction), `test_every_placeholder_is_supplied_by_the_chain_builder` will fail and require extending the allow-list. Fine for now since no current prompt text contains literal braces, but worth a comment near `SUPPLIED_PLACEHOLDERS`.
- **Scope creep in `internal_faq`**: the count changed from 5-7 to 6-8 FAQs and new coverage bullets (opportunity size, VP-killer questions) were added. These are content improvements beyond the stated format fix — reasonable, but they slightly widen the behavioral surface to verify in the post-merge generation run the PR description already recommends.
- **Duplicate local `import inspect`** in `test_builder_signature_has_no_unclassified_parameters` and `test_chain_builder_formats_prfaq_steps_cleanly` — hoist to the module top with the other imports (`test_prompt_utils.py`).
- **Pre-existing tension left in place**: `press_release.system_prompt` says "write as if the feature has just launched and is already gaining real traction" while the user template mandates a strictly future `{launch_date}`. Not introduced by this PR, but since the prompts were being reworked anyway it could have been reconciled.

No blockers: no IAM, secrets, Bedrock model-id, or ingestor-registration surface is touched, and the loader change is a strict hardening with test coverage.

---

mundurragacl:
Acknowledged — with 'no blockers, clean and mergeable' and nits-only for two consecutive rounds, I'm treating this as converged and holding the branch stable for merge rather than pushing further micro-edits (each push has been spawning a fresh nit crop — round 15's helper extraction begat round 16's two).

For the record on the remaining nits:
- **Wording-pin maintenance cost**: agreed and accepted — the tests are deliberately part of the prompt file's contract; the inline comments say so.
- **Literal-brace false positives**: correct, and intentional — a future prompt embedding a JSON example should trip the guard and force a conscious allow-list decision. Noted in `_extract_slots`' docstring ('broad on purpose').
- **internal_faq 6–8 scope**: covered by the post-merge generation run already committed to in the PR description.
- **Duplicate local `import inspect` / launched-vs-future-date tension**: noted; both are candidates for the post-merge follow-up pass rather than another review cycle here.

Ready for maintainer merge.

PR #152, opened 2026-07-14
fix: hide auto-detect button for JSON-LD scrapers

### LINKED ISSUE #18: UI:  When creating a scrapper autodetect field should be hidden by default
In create scrapper, you have an autodetect, which should be hidden on create scrapper from LD Json as the configuration is set by default. Auto detect is for "auto detecting the css classes the scraper needs to read from" on the custom CSS scrapper

mundurragacl:
## Summary

Fixes #18 (close manually on merge — base is `development`).

Auto-detect discovers **CSS selectors**; JSON-LD scrapers get their extraction config from the page's structured data, so the button implied an extra required step where none exists. The editor already hides the CSS-selector section when `extraction_method === 'jsonld'` — the auto-detect button and its hint now follow the same rule, and remain exactly as before for CSS scrapers.

One-component change (`ScraperEditor.tsx`), no API or type changes.

## Testing

- New `ScraperEditor.test.tsx`: button visible for CSS scrapers, absent for JSON-LD
- Scrapers suite: 110 tests pass; eslint + tsc clean

## Notes

- The issue predates the Scrapers UI rebuild; re-reading it, the ask was specifically "hidden on create scraper from LD Json" — which is what this implements (conditional on extraction method, not a collapsed disclosure). @svozza flagging in case the rebuilt UI changed what you had in mind.

### DIFF
--- a/voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx
+++ b/voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx
@@ -84,4 +84,12 @@ describe('ScraperEditor auto-detect visibility', () => {
 
     expect(screen.getByRole('button', { name: AUTO_DETECT_LABEL })).toBeInTheDocument()
   })
+
+  it('keeps auto-detect for legacy configs without an extraction_method', () => {
+    // Configs saved before JSON-LD support predate the field and are CSS
+    // scrapers — the positive === check must not hide their button.
+    renderEditor(makeScraper({ extraction_method: undefined }))
+
+    expect(screen.getByRole('button', { name: AUTO_DETECT_LABEL })).toBeInTheDocument()
+  })
 })
--- a/voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx
+++ b/voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx
@@ -3,13 +3,17 @@
  *
  * Auto-detect discovers CSS selectors; JSON-LD scrapers take their extraction
  * config from the structured data itself, so the button must not appear there
- * and suggest an extra required step.
+ * and suggest an extra required step. The extraction method is fixed by the
+ * template chosen before the editor opens (there is no in-editor switch), so
+ * the "create scraper from JSON-LD" flow from the issue is covered via the
+ * template prop.
  */
 import { describe, it, expect, vi } from 'vitest'
 import { render, screen } from '@testing-library/react'
 import ScraperEditor from './ScraperEditor'
-import { DEFAULT_SCRAPER_CONFIG } from './constants'
-import type { ScraperConfig } from '../../api/types'
+import { DEFAULT_SCRAPER } from './constants'
+import enScrapers from '../../../public/locales/en/scrapers.json'
+import type { ScraperConfig, ScraperTemplate } from '../../api/types'
 
 vi.mock('../../api/scrapersApi', () => ({
   scrapersApi: {
@@ -17,9 +21,14 @@ vi.mock('../../api/scrapersApi', () => ({
   },
 }))
 
+// Match on the shipped translation strings so a fallback to raw keys
+// ("editor.autoDetect") fails loudly instead of passing by accident.
+const AUTO_DETECT_LABEL = enScrapers.editor.autoDetect
+const AUTO_DETECT_HINT = enScrapers.editor.autoDetectHint
+
 function makeScraper(overrides: Partial<ScraperConfig>): ScraperConfig {
   return {
-    ...DEFAULT_SCRAPER_CONFIG,
+    ...DEFAULT_SCRAPER,
     id: 's-1',
     name: 'Test scraper',
     base_url: 'https://example.com/reviews',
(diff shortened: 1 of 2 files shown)


perrozzi:
This PR conditionally hides the auto-detect button and its hint in `ScraperEditor.tsx` when `extraction_method === 'jsonld'`, matching the existing pattern where the CSS-selector section is already hidden for JSON-LD scrapers. The change is small, well-scoped, includes a regression test, and follows repo conventions (i18n keys reused, no `as` assertions, vitest). Looks good overall.

### Nits

- `ScraperEditor.test.tsx`: the second test is named "hides the auto-detect button **and hint**" but only asserts the button's absence. Add an assertion that the hint text (`editor.autoDetectHint`) is also absent, otherwise reverting only the hint line of the fix would not fail the test.
- The tests match the button by `/auto-detect/i`, which depends on the resolved translation string. If the test setup ever falls back to raw keys (`editor.autoDetect`, no hyphen), these would break. Consider matching via the translated string from the locale file or a stable `aria-label`/test id for robustness.
- Consider one more test covering the interactive case: rendering a CSS scraper and switching the extraction method to JSON-LD in the UI, verifying the button disappears — this guards the actual user flow from issue #18 ("hidden on create scraper from LD Json"), not just initial render.

No blockers or should-fix items — the change is clean.

---

mundurragacl:
All three nits addressed in the latest commit: (1) the JSON-LD test now asserts the hint's absence too, so reverting only the hint line fails; (2) matching uses the shipped `en/scrapers.json` strings — a fallback to raw keys fails loudly instead of passing by accident; (3) the issue's actual flow is covered: creating a **new** scraper from a JSON-LD template hides auto-detect, and the untemplated (CSS default) create flow keeps it. Note on the suggested 'switch the method in the UI' test: the editor has no in-editor extraction-method control — the method is fixed by the template chosen before the editor opens — so the template prop *is* the interactive path (documented in the test file header). Scrapers suite 112/112 green.

perrozzi:
This PR conditionally hides the auto-detect button and its hint in `ScraperEditor` when `extraction_method === 'jsonld'`, matching the existing pattern of hiding the CSS-selector section for JSON-LD scrapers, and adds a vitest suite covering both the edit and create-from-template flows. The change is small, well-scoped, follows the frontend conventions (no new user-visible strings, so no i18n additions needed), and includes a regression test that would fail if the fix were reverted (`hides the auto-detect button and hint for JSON-LD scrapers`). Overall this looks clean and mergeable.

### Nits
- `voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.tsx`: `config.extraction_method !== 'jsonld'` is a negative check. Since auto-detect is specifically a CSS-selector feature, `=== 'css'` would be more future-proof if another extraction method is ever added — it would default to hidden rather than shown.
- `voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx`: matching on the shipped `en/scrapers.json` strings is a nice touch (raw-key fallback fails loudly via the `getByRole` assertion in the CSS test), but it does couple the test to copy edits. Consider matching on a `data-testid` or a substring if translation churn becomes noisy.
- The tests depend on i18n being initialized in the vitest environment (otherwise `t()` returns raw keys and the first test fails). If the repo's shared test setup already initializes i18next this is fine; just confirm there's no per-file init being silently skipped here.
- The inline comment in `UrlInput` referencing "issue #18" is helpful; consider keeping issue references in the test `@fileoverview` only, since PR/issue links in component code tend to go stale.

No blockers or should-fix items — nice, focused fix with appropriate test coverage.

---

mundurragacl:
Round 2 addressed in the latest commit:

- **Positive check**: `showAutoDetect` is now `(config.extraction_method ?? 'css') === 'css'` — future extraction methods default to hidden, exactly as you suggested. The explicit default matters because `extraction_method` is optional in `ScraperConfig`: legacy configs saved before JSON-LD support lack the field and are CSS scrapers, so a bare `===` would have hidden their button (new regression test covers that case — 113/113 green).
- **Issue reference** moved out of the component comment; it lives in the test fileoverview.

Acknowledged, no change: string-based matching stays deliberately — i18next is initialized by the shared vitest setup with the real `en` resources (that's why the suite passes on translated strings), and coupling to shipped copy is the point: a fallback to raw keys or an accidental copy regression should fail this test. If translation churn gets noisy we can revisit with a test id.

perrozzi:
This PR hides the auto-detect button and its hint in `ScraperEditor` when the scraper's extraction method is JSON-LD, since auto-detect only discovers CSS selectors. The change is a small, well-scoped conditional in one component, backed by a new focused test file, and matches the existing pattern of hiding the CSS-selector section for JSON-LD scrapers. This looks clean and follows repo conventions (regression tests for the bug fix, no new user-visible strings so no i18n changes needed, no API/type changes).

Points in favor:

- `voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.tsx`: the `(config.extraction_method ?? 'css') === 'css'` guard correctly handles legacy configs missing the field, and defaults any future extraction method to hidden — both choices are documented in the comment and covered by tests.
- `voc-datalake/frontend/src/pages/Scrapers/ScraperEditor.test.tsx`: covers the exact flow from issue #18 (create from JSON-LD template), the CSS positive case, the no-template default, and the legacy-config case. Asserting against the shipped English translation strings (rather than i18n keys) means an i18n fallback to raw keys fails the test rather than silently passing — a nice touch.

### Nits

- The test file imports translations from `../../../public/locales/en/scrapers.json`. This couples the test to the locale file layout; if the suite has a shared i18n test setup, deriving the strings from the initialized `i18next` instance would be slightly more robust to future restructuring. Not worth blocking on.
- The `scrapersApi` mock only stubs `analyzeUrlForSelectors`. If `ScraperEditor` ever grows another API call, these tests will fail with an opaque "not a function" error; using `vi.fn()` defaults via a partial mock helper or a brief comment would ease future maintenance.

No blockers or should-fix items — this is good to merge.

---

mundurragacl:
Thanks — noting the two nits for future maintenance rather than changing now: the locale-file import is deliberate (single source for the shipped copy; if the layout moves, one import path breaks loudly), and the single-method `scrapersApi` mock matches the component's single API dependency today — a partial-mock helper is the right move the day it grows a second call. Ready to merge from my side.

perrozzi:
The changes since the last review are confined to `ScraperEditor.test.tsx` and directly address both nits from the previous review: the translation strings are now derived from the shared i18n test setup instead of importing the locale JSON, and the `scrapersApi` mock now stubs the full real API surface with `vi.fn()`s. The component code is unchanged.

### Resolved
- Test coupling to `public/locales/en/scrapers.json` layout — strings now come from the initialized i18next instance, with a guard test that fails loudly if the keys stop resolving (nicely avoiding the vacuous-pass trap where component and test both render raw keys).
- Brittle single-method `scrapersApi` mock — now derives stubs from the actual module's keys via `importOriginal`, so future component API calls hit assertable `vi.fn()`s and new methods are covered automatically.

### New findings
None blocking. One observation, not worth action: the module-level `i18n.t(...)` calls depend on the shared setup having synchronously initialized i18next with loaded resources before test-file evaluation. If that setup ever moves to async resource loading, these constants would resolve to raw keys — but the new guard test would catch exactly that failure mode explicitly, so the risk is already mitigated by design.

The incremental changes are clean and improve on an already-mergeable PR. Good to merge.

---

PR #153, opened 2026-07-14
feat: mark problems as resolved in Problem Analysis

### LINKED ISSUE #66: Problem Analysis: Allow marking problems as resolved/completed
### Problem
The Problem Analysis page (`/problems`) aggregates all feedback problems by category, subcategory, and problem group. Over time, problems that have already been fixed (e.g., 3-6 months ago) continue to appear in the view, adding noise and making it harder to focus on active, unresolved issues.

### Proposed Solution
Add the ability to mark a problem group as **resolved/completed** so it no longer appears in the Problem Analysis tree by default.

#### Expected behavior:
- At the problem group level, add a "Mark as Resolved" action (button or context menu)
- Resolved problems should be hidden from the default view
- Add a toggle/filter to show/hide resolved problems (e.g., "Show resolved" checkbox) for historical reference
- Persist the resolved status in DynamoDB so it survives page reloads and is shared across users

#### Implementation considerations:
- Store resolved status in the `voc-aggregates` table (or a new partition key pattern like `RESOLVED_PROBLEM#{hash}`)
- The resolution could be keyed by a hash of category + subcategory + normalized problem text, since problems are grouped by similarity
- Add a new API endpoint (e.g., `PUT /feedback/problems/{id}/resolve`) in the metrics handler
- Frontend: add resolve/unresolve button on each problem row in the tree view, filter state in the controls bar

### Why this matters
As the platform collects feedback over weeks and months, the problem list grows. Teams that actively fix issues need a way to clear resolved problems from their working view so they can focus on what still needs attention.

mundurragacl:
## Summary

Fixes #66 (close manually on merge — base is `development`).

Teams that fix issues need resolved problems out of their working view. This adds a resolve/unresolve control on every problem group, hides resolved groups by default, and a **Show resolved (n)** toggle reveals them (dimmed, struck through, green badge) for historical reference. Resolution state is **shared across users** and persisted in DynamoDB, per the issue.

### Design

- **No identity server-side**: problem groups are built client-side by similarity, so resolution is keyed by a normalized `category|subcategory|problem` composite (trim/lowercase/whitespace-collapse) — cosmetic LLM re-phrasings keep the key stable. Exact-match limitation: if the *representative* text of a similarity group changes materially, the resolution orphans (documented in code).
- **Zero infrastructure changes**: state lives in one `voc-aggregates` config item behind `GET/PUT /settings/resolved-problems` on the existing settings Lambda (aggregates RW already granted; `/settings/*` proxy already routed + Cognito-authed). Per-key atomic nested SET/REMOVE means concurrent resolves can't clobber each other.
- Category/subcategory header totals are recomputed to match the visible tree, and the PDF export follows it.
- All strings translated in the 8 locales; mock server supports the flow for local dev.

## Testing

- Backend: 6 new handler tests (roundtrip, empty state, atomic update shape, 6-case payload validation) — settings suite 22/22
- Frontend: pipeline unit tests (key normalization, hide/annotate/counts), row control tests (fires without toggling the row, badge states), page integration (hidden by default → toggle reveals → unresolve persists) — full suite 1,572/1,572, eslint clean on touched files, tsc clean, `i18n:validate` in sync
- Note: a full-tree `eslint src` shows 8 pre-existing `set-state-in-effect` errors inherited from `development` — they're fixed in #138, not touched here.

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -98,6 +98,9 @@ const mockSourcesStatus = {
 };
 
 // Mock categories config
+// Problem resolution state (issue #66) — mutable so the toggle round-trips.
+const mockResolvedProblems = {};
+
 const mockCategoriesConfig = {
   categories: [
     { name: 'delivery', display_name: 'Delivery', description: 'Shipping and delivery issues', color: '#EF4444' },
@@ -161,6 +164,18 @@ const handlers = {
   // Settings - Categories
   'GET /settings/categories': () => mockCategoriesConfig,
   'PUT /settings/categories': () => ({ success: true, message: 'Categories saved' }),
+  'GET /settings/resolved-problems': () => ({ resolved: mockResolvedProblems }),
+  'PUT /settings/resolved-problems': (body) => {
+    if (!body || typeof body.key !== 'string' || body.key.trim() === '' || typeof body.resolved !== 'boolean') {
+      return { success: false, message: 'key and resolved are required' };
+    }
+    if (body.resolved) {
+      mockResolvedProblems[body.key] = { resolved_at: new Date().toISOString() };
+    } else {
+      delete mockResolvedProblems[body.key];
+    }
+    return { success: true, key: body.key, resolved: body.resolved };
+  },
 
   // Data Explorer
   'GET /data-explorer/buckets': () => ({ buckets: mockBuckets }),
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -191,7 +191,9 @@ const handlers = {
   'GET /settings/resolved-problems': () => ({ resolved: mockResolvedProblems }),
   'PUT /settings/resolved-problems': (body) => {
     if (!body || typeof body.key !== 'string' || body.key.trim() === '' || typeof body.resolved !== 'boolean') {
-      return { success: false, message: 'key and resolved are required' };
+      // Mirror the real handler's 400 contract so local dev doesn't mask
+      // client bugs behind a 200.
+      return { __status: 400, body: { success: false, message: 'key and resolved are required' } };
     }
     if (body.resolved) {
       mockResolvedProblems[body.key] = { resolved_at: new Date().toISOString() };
@@ -386,6 +388,13 @@ const server = http.createServer((req, res) => {
         }
       }
       const result = handler(parsedBody, url.searchParams);
+      // Handlers may signal a non-200 status (matching the real API's error
+      // contract) by returning { __status, body }.
+      if (result && typeof result.__status === 'number') {
+        res.writeHead(result.__status);
+        res.end(JSON.stringify(result.body));
+        return;
+      }
       res.writeHead(200);
       res.end(JSON.stringify(result));
     });
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -78,6 +78,8 @@
     "none": "Keine"
   },
   "problemResolution": {
+    "allResolvedHint": "Aktiviere oben „Gelöste anzeigen ({{total}})\", um sie zu prüfen",
+    "allResolvedTitle": "Alle Probleme in dieser Ansicht sind als gelöst markiert",
     "markResolved": "Als gelöst markieren",
     "markUnresolved": "Als ungelöst markieren",
     "resolved": "Gelöst",
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -81,6 +81,7 @@
     "markResolved": "Als gelöst markieren",
     "markUnresolved": "Als ungelöst markieren",
     "resolved": "Gelöst",
+    "saveFailed": "Der Status konnte nicht aktualisiert werden. Bitte erneut versuchen.",
     "showResolved": "Gelöste anzeigen ({{total}})"
   },
   "send": "Senden",
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -77,6 +77,12 @@
     "medium": "Mittel",
     "none": "Keine"
   },
+  "problemResolution": {
+    "markResolved": "Als gelöst markieren",
+    "markUnresolved": "Als ungelöst markieren",
+    "resolved": "Gelöst",
+    "showResolved": "Gelöste anzeigen ({{total}})"
+  },
   "send": "Senden",
   "sentiment": {
     "mixed": "Gemischt",
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -18,6 +18,7 @@
   },
   "cancel": "Abbrechen",
   "configureBrand": "Marke konfigurieren",
+  "dismiss": "Schließen",
   "exportPdf": "PDF exportieren",
   "exportPdfShort": "PDF",
   "exportPdfTooltip": "Als PDF exportieren",
@@ -76,6 +77,10 @@
     "low": "Niedrig",
     "medium": "Mittel",
     "none": "Keine"
+  },
+  "problemAnalysisPage": {
+    "emptyHint": "Erweitere den Zeitraum oder passe die Filter an",
+    "emptyTitle": "Keine Problemanalyse-Daten für den ausgewählten Zeitraum gefunden"
   },
   "problemResolution": {
     "allResolvedHint": "Aktiviere oben „Gelöste anzeigen ({{total}})\", um sie zu prüfen",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -18,6 +18,7 @@
   },
   "cancel": "Cancel",
   "configureBrand": "Configure brand",
+  "dismiss": "Dismiss",
   "exportPdf": "Export PDF",
   "exportPdfShort": "PDF",
   "exportPdfTooltip": "Export as PDF",
@@ -76,6 +77,10 @@
     "low": "Low",
     "medium": "Medium",
     "none": "None"
+  },
+  "problemAnalysisPage": {
+    "emptyHint": "Try expanding the time range or adjusting filters",
+    "emptyTitle": "No problem analysis data found for the selected period"
   },
   "problemResolution": {
     "allResolvedHint": "Enable \"Show resolved ({{total}})\" above to review them",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -77,6 +77,12 @@
     "medium": "Medium",
     "none": "None"
   },
+  "problemResolution": {
+    "markResolved": "Mark as resolved",
+    "markUnresolved": "Mark as unresolved",
+    "resolved": "Resolved",
+    "showResolved": "Show resolved ({{total}})"
+  },
   "send": "Send",
   "sentiment": {
     "mixed": "Mixed",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -81,6 +81,7 @@
     "markResolved": "Mark as resolved",
     "markUnresolved": "Mark as unresolved",
     "resolved": "Resolved",
+    "saveFailed": "Could not update the resolved status. Try again.",
     "showResolved": "Show resolved ({{total}})"
   },
   "send": "Send",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -78,6 +78,8 @@
     "none": "None"
   },
   "problemResolution": {
+    "allResolvedHint": "Enable \"Show resolved ({{total}})\" above to review them",
+    "allResolvedTitle": "All problems in this view are marked resolved",
     "markResolved": "Mark as resolved",
     "markUnresolved": "Mark as unresolved",
     "resolved": "Resolved",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -81,6 +81,8 @@
     "none": "Ninguna"
   },
   "problemResolution": {
+    "allResolvedHint": "Activa \"Mostrar resueltos ({{total}})\" arriba para revisarlos",
+    "allResolvedTitle": "Todos los problemas de esta vista están marcados como resueltos",
     "markResolved": "Marcar como resuelto",
     "markUnresolved": "Marcar como no resuelto",
     "resolved": "Resuelto",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -18,6 +18,7 @@
   },
   "cancel": "Cancelar",
   "configureBrand": "Configurar marca",
+  "dismiss": "Descartar",
   "exportPdf": "Exportar PDF",
   "exportPdfShort": "PDF",
   "exportPdfTooltip": "Exportar como PDF",
@@ -79,6 +80,10 @@
     "low": "Baja",
     "medium": "Media",
     "none": "Ninguna"
+  },
+  "problemAnalysisPage": {
+    "emptyHint": "Prueba a ampliar el rango de tiempo o ajustar los filtros",
+    "emptyTitle": "No se encontraron datos de análisis de problemas para el período seleccionado"
   },
   "problemResolution": {
     "allResolvedHint": "Activa \"Mostrar resueltos ({{total}})\" arriba para revisarlos",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -80,6 +80,12 @@
     "medium": "Media",
     "none": "Ninguna"
   },
+  "problemResolution": {
+    "markResolved": "Marcar como resuelto",
+    "markUnresolved": "Marcar como no resuelto",
+    "resolved": "Resuelto",
+    "showResolved": "Mostrar resueltos ({{total}})"
+  },
   "send": "Enviar",
   "sentiment": {
     "mixed": "Mixto",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -84,6 +84,7 @@
     "markResolved": "Marcar como resuelto",
     "markUnresolved": "Marcar como no resuelto",
     "resolved": "Resuelto",
+    "saveFailed": "No se pudo actualizar el estado. Inténtalo de nuevo.",
     "showResolved": "Mostrar resueltos ({{total}})"
   },
   "send": "Enviar",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -84,6 +84,7 @@
     "markResolved": "Marquer comme résolu",
     "markUnresolved": "Marquer comme non résolu",
     "resolved": "Résolu",
+    "saveFailed": "Impossible de mettre à jour le statut. Réessayez.",
     "showResolved": "Afficher les résolus ({{total}})"
   },
   "send": "Envoyer",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -80,6 +80,12 @@
     "medium": "Moyenne",
     "none": "Aucune"
   },
+  "problemResolution": {
+    "markResolved": "Marquer comme résolu",
+    "markUnresolved": "Marquer comme non résolu",
+    "resolved": "Résolu",
+    "showResolved": "Afficher les résolus ({{total}})"
+  },
   "send": "Envoyer",
   "sentiment": {
     "mixed": "Mixte",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -81,6 +81,8 @@
     "none": "Aucune"
   },
   "problemResolution": {
+    "allResolvedHint": "Activez « Afficher les résolus ({{total}}) » ci-dessus pour les consulter",
+    "allResolvedTitle": "Tous les problèmes de cette vue sont marqués comme résolus",
     "markResolved": "Marquer comme résolu",
     "markUnresolved": "Marquer comme non résolu",
     "resolved": "Résolu",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -18,6 +18,7 @@
   },
   "cancel": "Annuler",
   "configureBrand": "Configurer la marque",
+  "dismiss": "Fermer",
   "exportPdf": "Exporter en PDF",
   "exportPdfShort": "PDF",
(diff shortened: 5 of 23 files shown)


perrozzi:
This PR adds a shared "mark as resolved" feature to the Problem Analysis page: a new `GET/PUT /settings/resolved-problems` route on the existing settings Lambda (persisted in a single `voc-aggregates` config item), client-side key normalization and tree filtering (`problemResolution.ts`), a resolve control on `ProblemRow`, a "Show resolved (n)" toggle, i18n keys in all 8 locales, mock-server support, and tests on both sides. Overall this is a well-scoped, well-tested change that follows repo patterns (domain-scoped Lambda, existing IAM grants, powertools decorators, i18n coverage). No blockers, but a few robustness gaps worth addressing.

### Should fix

- **`buildResolutionKey` normalizes only the problem text, not category/subcategory** (`problemResolution.ts`). The problem component gets trim/lowercase/whitespace-collapse, but `category` and `subcategory` are interpolated raw. Since these are LLM-classified values, casing/whitespace drift ("Delivery" vs "delivery") would silently orphan resolutions — the exact failure mode the normalization exists to prevent. Normalize all three components.
- **No error handling on the resolve mutation** (`ProblemAnalysis.tsx`). `toggleResolvedMutation` has no `onError`; a failed PUT leaves the UI stale with zero user feedback, and the toggle button isn't disabled while pending (rapid double-clicks can race SET/REMOVE for the same key). Add an error surface (toast) and consider `isPending` disabling or an optimistic update with rollback.
- **Unbounded growth of the single DynamoDB item** (`settings_handler.py`). Per-key length is capped at 500 chars, but there's no cap on the *number* of resolved entries, and orphaned keys (from materially rephrased representative text, per the documented limitation) accumulate forever in one item subject to the 400KB cap. Consider a max-entry count check on PUT, or at least document a cleanup path.

### Nits

- `settings_handler.py`: `except ConfigurationError: raise` inside both new `try` blocks is dead code — the `ConfigurationError` is raised before the `try` (and `ValidationError` likewise for the PUT handler).
- `ProblemRow.tsx`: `ResolveToggleButton` should set `type="button"` to be safe inside any form context.
- `mock-server.js`: the invalid-payload branch returns `{ success: false }` presumably with HTTP 200, while the real handler returns 400 — the mock diverges from the API contract, which can mask client bugs in local dev.
- `api/client.ts`: the new responses are typed but not runtime-validated; repo convention prefers Zod at system boundaries (consistent with the existing client, so acceptable, but worth aligning eventually).
- `resolvedCount` in the "Show resolved (n)" label reflects only resolved problems within the current filters (urgent/source/category), not the global total — fine, but the label could read as a global count.

Nice work on the sibling (non-nested) button structure in `ProblemRow`, the atomic nested SET/REMOVE with the `if_not_exists` parent-map guard, the parametrized payload-validation tests, and complete locale coverage.

---

mundurragacl:
All three should-fixes and the actionable nits are addressed in the latest push (586c08c), and the branch is conflict-free against `development` (post-#138 merge; the `exportPDF` conflict resolved as visibleData + dateBasis label union).

- **Key normalization**: all three components (category, subcategory, problem) now get trim/lowercase/whitespace-collapse — test updated to pin it.
- **Mutation error handling**: failures render an `role=alert` banner (translated in all 8 locales), and toggles are ignored while a mutation is pending so rapid double-clicks can't race SET/REMOVE on the same key.
- **Bounded growth**: PUT rejects new entries beyond `MAX_RESOLVED_ENTRIES` (500 — sized so worst-case 500×~540B stays well under the 400KB item cap) with a clear 400; overwrites of existing keys and unresolves are never capped. 3 new handler tests cover cap/overwrite/unresolve.
- **Nits**: dead `except ConfigurationError` clauses removed; `type="button"` on the resolve control; the mock server now returns a real 400 via a handler-signaled status instead of `success:false` with a 200.

Left as-is, with reasons: the Zod-at-boundaries note (consistent with the rest of `client.ts` today, as you said — worth a dedicated pass), and the `resolvedCount` label reflecting the filtered view (intentional: the toggle reveals what the current filters would show).

Gates on the merged branch: settings handler 25/25, ProblemAnalysis suite 43/43, full frontend 1,592 (one unrelated pre-existing flake in `chatStore.test.ts` — `conv_${Date.now()}` ID collision, passes on rerun, present on `development`), backend 754, eslint/tsc clean, `i18n:validate` in sync.

perrozzi:
This PR adds shared resolve/unresolve state for Problem Analysis problem groups: two new routes on the existing settings Lambda backed by a single `voc-aggregates` config item, a client-side normalized composite key, tree filtering/annotation logic in a new `problemResolution.ts` module, UI controls, all 8 locales, mock-server support, and solid test coverage on both sides. Overall this is a clean, well-scoped change that follows the repo's patterns (existing domain Lambda, no new IAM grants, i18n complete, tests included). No blockers found.

### Should fix

- **`voc-datalake/lambda/api/settings_handler.py` — verify `datetime`/`timezone` imports.** The new `set_problem_resolution` uses `datetime.now(timezone.utc)`, but the diff adds no import and the visible hunk starts below the import block. If the module didn't already import these for another handler, this is a `NameError` at request time. The passing test suite suggests they exist, but please confirm — the diff doesn't show it.
- **Cap check is not atomic (TOCTOU).** The read-then-write in the resolve path means two concurrent resolves at `MAX_RESOLVED_ENTRIES - 1` can both pass the check and exceed the cap, and it costs an extra read of a potentially large item on every resolve. A `ConditionExpression` like `size(#r) < :max OR attribute_exists(#r.#k)` on the SET would make the cap atomic and drop the extra `get_item`. Given the docstring explicitly advertises atomicity, worth tightening or documenting the soft-cap behavior.
- **`ProblemGroup` interface (with the new `resolved?: boolean`) is now duplicated in three files** (`ProblemAnalysis.tsx`, `ProblemRow.tsx`, `SubcategoryRow.tsx`) plus the structurally-similar `ResolvableProblem` in `problemResolution.ts`. The duplication predates this PR, but adding a field to all copies is exactly when it should be extracted to a shared type — a future field added to only one copy will silently type-check.
- **`handleToggleResolved` silently drops clicks while a mutation is pending** (`ProblemAnalysis.tsx`). Toggling a *different* problem during an in-flight request does nothing with no feedback. Disabling the `ResolveToggleButton` (via a prop) while pending, or keying pending state per problem, would make the behavior visible instead of surprising.

### Nits

- `problemResolution.ts` documents that `|` "cannot appear in category/subcategory values", but categories are user-configurable via the categories settings — a `|` in a name would silently merge keys. Consider normalizing `|` out or noting the assumption where categories are validated.
- Repo convention prefers Zod validation at system boundaries; `getResolvedProblems` trusts the response shape via a `fetchApi<T>` generic. Consistent with the existing client, but the new module would be a natural place to start.
- The "ensure parent map" `update_item` runs on every PUT, doubling writes; it's only needed once. A `ConditionExpression`/single-expression approach or catching the invalid-path error would halve WCU.
- `Show resolved ({{total}})` count reflects the currently filtered tree, not the global resolved store — reasonable, but worth a comment since users may expect the DynamoDB total.
- The resolved-toggle checkbox sits under "Filters" and is cleared by the Clear-filters button; pre-existing neighbors like "Urgent only" are untranslated, so the new translated label is slightly inconsistent with its siblings (not this PR's fault).

---

mundurragacl:
Round 2 addressed in 181e3b1:

- **Atomic cap (TOCTOU)**: the read-then-write is gone. A single conditional write — `attribute_exists(#r.#k) OR size(#r) < :max` — enforces the cap on the same call, which also removes the per-resolve `get_item` AND the every-PUT ensure-parent write (your WCU nit): steady state is exactly one write. The parent map materializes only on the first-ever resolve; since DynamoDB can report a missing parent as either a document-path ValidationException or a failed condition depending on evaluation order, both first-attempt failures fall through to ensure+retry, and only a retry failure means the cap was genuinely reached. Tests rewritten to the new contract (single-write assertion, first-use materialization sequence, cap 400, overwrite-at-cap short-circuit, unresolve tolerant of a missing map).
- **Shared tree types**: `ProblemGroup`/`SubcategoryGroup`/`CategoryGroup` now live once in `problemResolution.ts` and all three components import them — agreed that adding `resolved?` to every copy was exactly the moment to extract.
- **Pending UX**: resolve controls are visibly disabled (with cursor/opacity) while a mutation is in flight, instead of silently dropping clicks.
- **Pipe nit**: good catch that categories are user-configurable — `buildResolutionKey` now strips literal `|` from all components so `'a|b'` as a category can't merge keys with `'a' + 'b|sub'` (disambiguation test added).
- **`datetime`/`timezone`**: confirmed imported at module top (line 5, used by the brand handler since before this PR); the resolve path's tests would NameError otherwise.
- **Count semantics**: documented at the toggle that it counts within the current filters.

Left as-is: Zod-at-boundaries (consistent with the existing client; natural candidate for a dedicated pass) and the untranslated pre-existing siblings.

Gates: settings handler 26/26, ProblemAnalysis suite 44/44, tsc + eslint clean.

perrozzi:
This PR adds a shared "mark resolved" feature to Problem Analysis: two new endpoints on the existing settings Lambda persisting a keyed map in the aggregates table, a client-side pipeline (`problemResolution.ts`) that annotates/hides resolved groups and recomputes totals, UI controls, i18n for all 8 locales, mock-server support, and solid test coverage on both sides. The design (normalized composite key, atomic per-key SET/REMOVE, condition-enforced cap) is thoughtful and well documented. Overall this is close to mergeable, with one type-consistency issue worth confirming.

### Should fix

- **`SubcategoryRow.test.tsx` doesn't pass the required `resolvePending` prop.** `SubcategoryRowProps` declares `readonly resolvePending: boolean` (required), but none of the updated test renders supply it. Either these tests fail `tsc` (contradicting the "tsc clean" claim, if the test tsconfig type-checks tests) or the prop should be optional like it is on `ProblemRowProps` (`resolvePending?: boolean`). Make the two interfaces consistent and fix the tests accordingly (`voc-datalake/frontend/src/pages/ProblemAnalysis/SubcategoryRow.tsx`, `SubcategoryRow.test.tsx`).
- **`_unresolve_problem_key` swallows all `ValidationException`s** (`voc-datalake/lambda/api/settings_handler.py`). The intent is "missing parent map = nothing to remove", but `ValidationException` also covers genuine request problems (malformed expression, oversized attribute name, etc.), which would be silently reported as `success: true`. Consider checking the error message for the document-path case, or narrowing the swallow with a comment on the accepted trade-off.
- **Cross-user staleness of the shared state**: the `['resolved-problems']` query is only invalidated by the local mutation, so user B never sees user A's resolves until a full page remount. Given the feature's whole point is shared state, consider a `staleTime`/`refetchOnWindowFocus` policy or refetch-on-interval for this query (`ProblemAnalysis.tsx`). Not blocking, but worth a deliberate decision.

### Nits

- `MAX_PROBLEM_KEY_LEN = 500` characters can be >500 bytes for non-ASCII

---

mundurragacl:
Round 3 addressed in the latest push:

- **resolvePending consistency (should-fix)**: made optional on `SubcategoryRowProps`, matching `ProblemRowProps` — you were right that the required prop contradicted the test renders (the app tsconfig excludes tests, which is why `tsc` was clean; the test tsconfig would have flagged it).
- **Narrowed swallow (should-fix)**: `_unresolve_problem_key` now no-ops only when the ValidationException message is the missing-document-path variant; malformed-expression/invalid-name failures surface as 500s instead of `success: true`. Test added for the non-path variant.
- **Cross-user staleness (should-fix)**: deliberate policy added — `staleTime: 15s` + `refetchOnWindowFocus: true` on the resolved-problems query, so another user's resolves appear on tab focus or within 15s of interaction, no remount needed. Chose polling-free freshness over an interval: this is a low-write admin-ish action and focus-refetch covers the realistic 'two analysts triaging together' case.
- **Byte-length nit**: keys are now capped at 500 UTF-8 **bytes** as well as 500 characters — correct catch that CJK text triples the byte cost and would have broken the 400KB math. The cap comment shows the worst-case arithmetic (500×~540B ≈ 270KB) and a CJK regression test pins it.

Settings suite 28/28, ProblemAnalysis 44/44, tsc/eslint clean. (The last nit in your comment was cut off mid-list after the byte-length item — if there was more, send it and I'll take it.)

perrozzi:
This PR adds shared resolve/unresolve state for Problem Analysis groups: new `GET/PUT /settings/resolved-problems` on the existing settings Lambda (single DynamoDB config item, per-key atomic nested SET/REMOVE with an atomic entry cap), a client-side normalized composite key, a hide-by-default + "Show resolved" toggle in the UI, i18n for all 8 locales, mock-server support, and solid test coverage on both sides. Overall this is well thought out — the atomic condition-expression cap, the missing-parent-map retry path, and the narrowed `ValidationException` handling on unresolve all show care. A few items worth addressing before merge.

### Should fix

- **DynamoDB expression-attribute-name length vs. the 500-byte key cap** (`voc-datalake/lambda/api/settings_handler.py`): the problem key is used as a nested map attribute name via `ExpressionAttributeNames={'#k': key}`. DynamoDB limits attribute names used in expressions to 255 bytes, but `MAX_PROBLEM_KEY_LEN`/`MAX_PROBLEM_KEY_BYTES` allow up to 500. A key in the 256–500-byte range would pass validation, then fail the update with a `ValidationException` — in `_resolve_problem_key` that gets swallowed on the first attempt, retried, and re-raised as a 500 instead of a clean 400. Please verify against the DynamoDB quota and lower the caps to 255 bytes (with a validation test).
- **Resolved rows leak into the PDF unmarked** (`ProblemAnalysis.tsx` / `toPDFCategories`): with "Show resolved" on, `exportPDF` uses `visibleData`, so resolved problems appear in the exported PDF indistinguishable from open ones (the strike-through/badge is UI-only). Either exclude resolved groups from the export or pass the `resolved` flag through `toPDFCategories` so the PDF annotates them.

### Nits

- `settings_handler.py`: prefer `raise ServiceError(...) from e` (and same for the cap `ValidationError`) to preserve the exception chain; also `logger.exception` already logs the traceback, so interpolating `{e}` into the message is redundant.
- `ProblemAnalysis.tsx`: `handleToggleResolved` silently drops clicks while a mutation is pending. The buttons are visually disabled via `resolvePending`, so the guard is mostly redundant — but if kept, a comment noting the double-defense would help; a per-key

---

mundurragacl:
Round 4 addressed in 7e4e5d4:

- **255-byte cap (should-fix)** — verified against the [DynamoDB constraints doc](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Constraints.html): the documented 255-byte expression limit measures the **alias token** by its own example (`#name` is 5 bytes), so a 500-byte substituted name likely worked — but the caps are lowered to 255 chars/bytes anyway: it removes the ambiguity at zero product cost and improves the item math (500 × ~295B ≈ 148KB worst case). More importantly, the client now **deterministically truncates** `buildResolutionKey` to the same caps (char slice, then byte-trim that never splits a code point), so a long LLM problem summary can't produce a key the server rejects with an un-actionable 400. Truncation + CJK tests on both sides.
- **PDF annotation (should-fix)** — `toPDFCategories` passes `resolved` through and the PDF renders resolved rows with a green RESOLVED badge, ✅ marker, and distinct border/background — no more unmarked leak when 'Show resolved' is on.
- **Nits** — `raise … from e` on all three raise sites; `logger.exception` messages no longer interpolate the exception; the pending-guard comment now explains it's double-defense with the disabled buttons (covers the pre-paint render gap).

Your round-4 comment was truncated mid-nit ("a per-key…") — if the per-key-pending thought had more to it, send it over. Suites: settings 28/28, ProblemAnalysis 46/46, tsc/eslint clean.

perrozzi:
This PR adds a shared "mark as resolved" capability to Problem Analysis: a new `GET/PUT /settings/resolved-problems` pair on the existing settings Lambda (reusing the already-granted aggregates table, so no IAM changes — good fit for the domain-split pattern), client-side key normalization/filtering in a new `problemResolution.ts` module, UI controls with full 8-locale i18n, PDF export support, mock-server support, and substantial test coverage on both sides. Overall this is a well-designed, well-tested change; the atomic conditional-write cap enforcement and the deduplicated tree types are nice touches. No blockers found.

### Should fix

- **Surrogate-pair truncation can 500 the server** (`voc-datalake/frontend/src/pages/ProblemAnalysis/problemResolution.ts`): `fitToByteCap` trims with `value.slice(0, -1)`, which removes one UTF-16 code unit at a time. If problem text contains

---

mundurragacl:
Round 5 addressed in the latest push — the surrogate-pair catch was real on both sides of the wire:

- **Client**: `fitToByteCap` now trims whole **code points** (`Array.from`) instead of UTF-16 code units, so truncating emoji-heavy text can never split a surrogate pair. The emoji test asserts `isWellFormed()` at the cap boundary.
- **Server** (defense in depth): JSON happily transports unpaired surrogates regardless of what our client does, and `key.encode('utf-8')` on one raised `UnicodeEncodeError` → 500. The key validation now rejects them as a clean 400 ('key must be valid Unicode (no unpaired surrogates)'), tested by injecting `\ud83d` at the raw-JSON layer exactly as a hostile client would send it.

Your comment was cut off mid-sentence at "If problem text contains" — if there were further items past the surrogate one, send them through. Suites: settings 29/29, ProblemAnalysis 47/47, tsc/eslint clean.

perrozzi:
This PR adds a shared resolve/unresolve feature to Problem Analysis: two new routes on the existing settings Lambda persisting a keyed map in the aggregates table, a client-side normalized composite key, tree filtering with recomputed totals, PDF annotation, i18n for all 8 locales, mock-server support, and solid test coverage on both sides. The backend write path is carefully atomic (per-key nested SET/REMOVE, cap enforced via ConditionExpression, surrogate/byte-cap validation), the frontend follows the TanStack Query pattern, and no IAM/CDK changes are needed since it reuses the aggregates RW grant and the `/settings/*` route. Overall this is a well-executed change; I found no blockers.

### Should fix

- **Hardcoded English `RESOLVED` badge in the PDF** (`ProblemAnalysisPDFContent.tsx`): every user-visible string needs i18n keys per repo policy — the `problemResolution.resolved` key already exists in all 8 locales, so pass a translated label into `ProblemItem` instead of the literal `RESOLVED`.
- **Misleading empty state when everything is resolved** (`ProblemAnalysis.tsx`): with resolved groups hidden, the page shows "No problem analysis data found for the selected period" even though data exists behind the toggle (your own integration test exercises exactly this state). Consider a variant message when `resolvedCount > 0 && visibleData.length === 0` pointing at the "Show resolved" toggle — this would need a new i18n key in all locales.
- **No runtime validation of the GET response** (`api/client.ts`): `getResolvedProblems` trusts the `Record<string, { resolved_at: string }

---

mundurragacl:
Round 6 addressed in the latest push — all three should-fixes taken:

- **PDF badge i18n**: the page passes `t('problemResolution.resolved')` into the export as a prop (the PDF renders outside the i18n provider, so a hook inside `ProblemItem` wasn't an option); the badge renders the translated label uppercased.
- **All-resolved empty state**: when `resolvedCount > 0` and the visible tree is empty, a translated message (new `allResolvedTitle`/`allResolvedHint` keys in all 8 locales) explains the data is behind the 'Show resolved' toggle instead of claiming no data exists — with the count interpolated so the hint names the exact toggle label. Integration test asserts the explanation renders in exactly the state you called out.
- **Zod at the boundary**: `parseResolvedProblemsResponse` validates the GET payload; malformed data degrades to an empty map with a `console.error` (problems just show unresolved), mirroring the plugin-manifest pattern. Your comment was truncated mid-item here — this implements the visible intent; if the cut part suggested a different failure policy, say the word.

Suites: ProblemAnalysis 47/47, settings 29/29, tsc/eslint clean, i18n in sync.

perrozzi:
This PR adds a shared resolve/unresolve capability to Problem Analysis: a new `GET/PUT /settings/resolved-problems` on the existing settings Lambda (single `voc-aggregates` config item, per-key atomic nested SET/REMOVE with a condition-enforced entry cap), a client-side resolution pipeline (`problemResolution.ts`) with Zod boundary validation and deterministic key normalization/truncation, UI controls with i18n in all 8 locales, PDF export support, mock-server support, and thorough tests on both sides. Overall this is a well-engineered change that follows the repo's patterns (domain-split Lambda reuse with no new IAM, Zod at the boundary, TanStack Query, full locale coverage); I found no blockers.

### Should fix

- **Untranslated strings moved into a new component** — `EmptyProblemsState` in `voc-datalake/frontend/src/pages/ProblemAnalysis/ProblemAnalysis.tsx` still hardcodes English for the "no data" branch ("No problem analysis data found for the selected period", "Try expanding the time range or adjusting filters"). These are pre-existing strings, but since this PR relocates them into a new component that already imports `useTranslation`, converting them to i18n keys here would be cheap and satisfies the "every user-visible string needs i18n keys" rule. If deliberately deferred, note it.
- **Brittle error-message matching** — `_unresolve_problem_key` in `voc-datalake/lambda/api/settings_handler.py` distinguishes the missing-parent no-op by matching `'document path' in error.get('Message', '').lower()`. AWS error message text isn't contractual and can change across SDK/service versions; if it drifts, unresolve on a fresh install starts returning 500. Consider a `ConditionExpression='attribute_exists(#r)'` on the REMOVE and treating `ConditionalCheckFailedException` as the no-op instead — that's code-based, not message-based. Not blocking since the current behavior degrades safely (500 rather than silent data issue) and is well-tested.

### Nits

- `fitToByteCap` in `problemResolution.ts` re-instantiates `TextEncoder` and re-encodes the full string on every recursive step — O(n²) worst case for a 255-code-point key. Trivial input size, but encoding once and trimming by byte length would be simpler.
- On first load, `resolvedData` is `undefined` so resolved problems briefly render as unresolved, then disappear when the query lands. A small loading gate (or seeding via `placeholderData`) would avoid the flicker.
- `ResolveErrorBanner` stays visible until the *next* mutation attempt; a dismiss control or auto-clear on refetch success would be friendlier.
- Test class names like `TestResolvedProblemsRoundThree` in `test_settings_handler.py` encode review history rather than behavior; consider naming by what they verify (e.g. `TestResolvedProblemsKeyEncoding`).
- `resolvedLabel` in `ProblemAnalysisPDFContent.tsx` falls back to English `'Resolved'`; since the page always passes it, consider making the prop required to prevent a future untranslated call site.

---

mundurragacl:
Round 7 addressed in f57c33e — both should-fixes and all five nits:

- **Message-text matching (should-fix)**: taken exactly as suggested — the unresolve no-op is now `ConditionExpression='attribute_exists(#r)'` + the stable `ConditionalCheckFailedException` code. No more sniffing AWS message text.
- **Relocated strings (should-fix)**: the no-data branch is translated (`problemAnalysisPage.emptyTitle/emptyHint`, all 8 locales) — agreed that moving them into a component that already had `useTranslation` made deferral indefensible.
- **Nits, all taken**: `fitToKeyCaps` is single-pass (one encode per code point) — and worth admitting: your insistence on the surrogate test paid off immediately, the emoji regression test caught a prefix bug in my first O(n) rewrite (skipped-then-refitted code points silently broke the byte cap). Initial render now gates on the resolved-state fetch, so resolved problems can't flash unresolved. The error banner has a translated dismiss control wired to `mutation.reset()`. Test classes are behavior-named (`TestResolvedProblemsKeyEncoding`). `resolvedLabel` is required on the PDF props.
- Query/mutation wiring moved into `useProblemResolution` (the `useSettingsSync` convention), which also absorbed the complexity-budget pressure the extra gates created.

Full frontend 1,597/1,597, settings 29/29, tsc/eslint/i18n clean.

perrozzi:
This PR adds shared resolve/unresolve state for Problem Analysis groups: two new routes on the existing settings Lambda persisting a keyed map in the aggregates table, a client-side normalized composite

---

mundurragacl:
The round-8 comment (13:14, sha f57c33e) arrived truncated after the opening summary line — no Should fix/Nits sections rendered, so there's nothing actionable visible. The visible fragment reads as a neutral-to-positive description of the final shape (two settings routes, keyed map in aggregates, normalized composite key). If the cut-off portion contained findings, please repost; otherwise, with the working tree stable since round 7 and all gates green, this is ready for maintainer merge.

Issue #150, opened 2026-07-13
Honor date_basis in AI Chat, Projects research, and MCP feedback queries


mundurragacl:
Follow-up from PR #138 (review-date vs imported-date filtering) and its review.

PR #138 added `date_basis` (`imported` | `review`) to `/feedback*` and `/metrics/*`, with a global picker in the dashboard header. Not yet covered:

- **AI Chat** (`chat_handler.py` context sample + `lambda/stream` `search_feedback` tool): queries by import date only
- **Projects research / persona / document generation** (`shared/feedback.py::query_feedback_by_date`): no `date_basis` parameter
- **MCP** (`mcp_handler.py`, uses `shared/feedback.py`): same

Effect: a user on review basis sees review-windowed dashboards but import-windowed chat answers and research scopes for the same nominal window.

Suggested approach: add the same post-filter to `shared/feedback.py` (the import-window ⊇ review-window containment argument from PR #138 applies unchanged — a review can't be imported before it was written), thread `date_basis` through the callers, and pass the picker value from the frontends that already have it in `configStore`.

Also worth unifying while in there (nit from the PR #138 review): `/feedback/urgent` and `/feedback/search` use `timedelta(days=days)` windows while list/metrics use a `days-1` cutoff (a days-long window ending today), so `days=7` spans 8 days on urgent/search and 7 elsewhere.

perrozzi:
Thanks for the detailed follow-up! 🙏

To restate my understanding: PR #138 introduced `date_basis` (`imported` | `review`) for the `/feedback*` and `/metrics/*` endpoints with a global dashboard picker, but three consumers still query by import date only:

1. **AI Chat** — `chat_handler.py` context sampling and the `search_feedback` tool in `lambda/stream`
2. **Projects research / persona / document generation** — `shared/feedback.py::query_feedback_by_date` has no `date_basis` parameter
3. **MCP** — `mcp_handler.py`, which relies on the same shared helper

The result is an inconsistency where a user on review basis sees review-windowed dashboards but import-windowed chat answers and research scopes for the same nominal window. Your proposed approach — applying the same post-filter in `shared/feedback.py` (leveraging the import-window ⊇ review-window containment argument), threading `date_basis` through callers, and passing the existing `configStore` picker value from the frontends — sounds coherent with how PR #138 was structured.

[...]

PR #154, opened 2026-07-14
feat: admin AI model picker over a curated allowlist

### LINKED ISSUE #96: [Feature]: have the ability to change the model
### Summary

I want to be able to adjust the model to any model that's available in bedrock from the admin interface inside

### Problem Statement

I want to be able to adjust the model to any model that's available in bedrock from the admin interface inside

### Proposed Solution

Add a selector but add something that will allow you to choose any model

### Alternatives Considered

_No response_

### Area

Frontend - Dashboard

### Acceptance Criteria

_No response_

mundurragacl:
## Summary

Fixes #96 (close manually on merge — base is `development`).

Admins get an **AI Model** card in Settings (brand tab, admin-only) to pin every AI feature to one Bedrock model — or keep **Automatic**, where each surface uses its tuned default (Python Lambdas: Sonnet 4.5; streaming chat: its env default).

### Deliberately limited form

Not free-form model IDs: the allowlist is fixed server-side to **Claude Sonnet 4.5** and **Claude Haiku 4.5**. The project's prompt templates are tuned for Claude, and both models are covered by the agreements the `BedrockAccessStack` already creates. Non-allowlisted values are rejected on write (400) and ignored on read, so even a tampered DynamoDB row can't steer inference to an arbitrary model.

### How it applies everywhere without IAM changes

- `shared/model_config.py`: resolution order *explicit arg > configured override > default*, 60s per-container cache, and the lookup **never raises** — no table env or no read permission just means the default. `shared/converse.py` resolves through it, which covers chat, projects, document/persona jobs, categories generation, scrapers, processor, and research (all already have aggregates read + the env var).
- Streaming chat (TS) gets the same lookup (`model-override.ts`, same allowlist + cache semantics) threaded into `ConverseStreamCommand`; it already had aggregates read.
- `GET/PUT /settings/model` on the existing settings Lambda — `/settings/*` proxy + Cognito auth already wired. **Zero infrastructure changes.**

## Testing

- 7 shared resolution tests (allowlist enforcement incl. tampered-value rejection, permission-failure fallback, cache TTL + invalidation)
- 9 handler tests (nullable override semantics, delete-on-clear, explicit-key requirement, 5-case rejection)
- 6 stream override tests (never-throws, cache, allowlist)
- 6 UI tests (Automatic default, select/clear round-trip, error state, hidden without endpoint)
- Full suites green: 725 backend (+1 pre-existing skip), 1,567 frontend, 189 stream; `tsc` clean ×3; CDK build clean
- Full-tree `eslint` shows only the 2 pre-existing `set-state-in-effect` errors inherited from `development` (fixed in #138)

## Notes

- Override propagation is cache-bounded: running containers pick up a change within ~60s.
- Adding a model later = one entry in each allowlist (Python + TS) — kept intentionally in code, not config, so the curation is reviewed.


perrozzi:
This PR adds an admin-facing "AI Model" picker that pins all AI features to one Bedrock model from a server-side allowlist (Sonnet 4.5 / Haiku 4.5), with a shared Python resolver (`shared/model_config.py`), a mirrored TS lookup for streaming chat, new `GET/PUT /settings/model` routes on the existing settings Lambda, and good test coverage across all three layers. The never-raises fallback design, allowlist enforcement on both read and write, and tampered-DB-value tests are well done. However, there are a few gaps that need addressing before merge.

### Blockers

- **Bedrock model id / IAM lockstep not satisfied for Haiku.** The repo rule is that model id strings and IAM resource ARNs change together. This PR introduces `global.anthropic.claude-haiku-4-5-20251001-v1:0` as an invocable model (`lambda/shared/model_config.py`, `lambda/stream/src/bedrock/model-override.ts`) with **zero CDK changes**. The PR description argues model *agreements* exist via `BedrockAccessStack`, but agreements are not IAM invoke permissions. If the Lambdas' `bedrock:InvokeModel*` grants are scoped to the Sonnet inference-profile/foundation-model ARNs (least-privilege pattern), selecting Haiku will AccessDeny every AI feature at runtime — and the never-raises design would silently fall back only in the *lookup*, not the *invoke*. Either add the Haiku ARNs to the relevant execution-role grants in the CDK, or demonstrate the existing grants already cover it.
- **No server-side admin authorization on `PUT /settings/model`.** The UI gates the card behind `isAdmin` (`Settings.tsx`), but `settings_handler.py` only relies on the existing Cognito authorizer — any authenticated non-admin user can change the org-wide model (cost/quality impact for everyone). Enforce the admin group check in the handler, and add a test for the 403 path.
- **Hardcoded user-visible strings, no i18n.** `AiModelSection.tsx` ships English-only strings ("AI Model", "Automatic", "Saved", "Loading models...", "Failed to save model selection. Try again.", the intro paragraph), and the model `label`/`description` strings are served as English text from the backend allowlist. Repo convention requires i18n keys for all locales; consider returning stable ids from the API and translating labels/descriptions client-side.

### Should fix

- **Query error state is a permanent spinner.** In `AiModelSection.tsx`, if `getModelSettings` rejects, `isLoading || !data` keeps rendering "Loading models..." forever (with `retry: false` in tests, but retries exhausted in prod too). Handle `isError` on the query.
- **Allowlist is duplicated in Python and TS with no drift guard.** Intentional per the description, but a cheap cross-check (e.g., a test asserting both lists against a shared fixture, or at minimum a comment in each file pointing at the other — the TS file has one, the Python file doesn't) would prevent the two surfaces silently accepting different models.
- **Verify `datetime`/`timezone` imports in `settings_handler.py`.** The new `save_model_settings` uses `datetime.now(timezone.utc)` but the diff adds no import; if the file doesn't already import these, PUT will 500. (Truncated context — please confirm.)

### Nits

- `settings_handler.py` `get_model_settings`: the `except ConfigurationError: raise` clause is dead code — the `ConfigurationError` is raised before the `try` block.
- `AiModelSection.tsx`: the `setTimeout` for the "Saved" badge isn't cleaned up on unmount (potential set-state-after-unmount warning).
- `model-override.test.ts` uses `as unknown as DynamoDBDocumentClient`; repo prefers type guards over `as`, though this is tolerable in test mocks.
- Model labels/descriptions are now duplicated in three places (`model_config.py`, `mock-server.js`, and effectively the TS allow

---

mundurragacl:
All three blockers, both actionable should-fixes, and the nits are addressed in 4838b63 (which also merges current `development` — conflict-free now).

**IAM lockstep (blocker)** — you were right, and this would have broken at runtime: the api-stack grants were Sonnet-only. All 7 AI-feature grant sites (chat, settings, scrapers, projects, jobs, stream, avatar) now use a single `allowlistedModelArns()` helper carrying both models' inference-profile + foundation-model ARNs. The processing stack already had Haiku (the processor defaults to it). And to make the three-way lockstep structural rather than aspirational: two new Python tests read the TS allowlist (`model-override.ts`) and the CDK helper source and assert both equal `ALLOWED_MODEL_IDS` — adding a model to one place now fails the build until all three agree.

**Admin authorization (blocker)** — `PUT /settings/model` now requires the `admins` Cognito group server-side (`get_caller_groups`/`require_admin` promoted to `shared/api.py`, 403 through the existing `AuthorizationError` handler). GET stays open to all authenticated users. Three new tests: non-admin 403, missing-groups 403, non-admin GET 200. (`users_handler` still has its local copy of the same logic — switching it to the shared helper is a mechanical follow-up I kept out of this diff.)

**i18n (blocker)** — all AiModelSection strings are translated in the 8 locales under `settings:aiModel.*`, and per your suggestion the API now returns a stable per-model `key` ('sonnet' / 'haiku'); descriptions are translated client-side under that key with the server string as `defaultValue` fallback. Model *names* intentionally come from the server fallback rather than locale files — they're product names, identical in every language, and the i18n gate rejects same-as-English locale values.

**Should-fixes**: query failure now renders an alert instead of the infinite spinner; the allowlist drift guard is the lockstep tests above; `datetime`/`timezone` were already imported (used by the brand handler) — confirmed, and the PUT path is covered by tests that would 500 otherwise.

**Nits**: dead `except ConfigurationError` removed from `get_model_settings`; the "Saved" badge timeout is cleared on unmount; the UI copy no longer claims enrichment is affected (the processor passes its model explicitly — `model_id=PROCESSOR_MODEL_ID` — so it keeps Haiku regardless of the override; explicit-arg precedence is by design). The `as unknown as` in the TS test mock stays, acknowledged as tolerable.

Gates on the merged branch: backend 764 (+13 new since review), frontend 1,588, stream 189, CDK build clean, eslint/tsc clean, `i18n:validate` exits 0.

perrozzi:
This PR adds an admin-only AI model picker (Sonnet 4.5 / Haiku 4.5 allowlist) with a settings endpoint, a shared Python resolver, a TS mirror for streaming chat, consolidated IAM grants, full i18n coverage, and lockstep tests tying the three allowlist locations together. The overall design is sound — server-side allowlist enforcement, never-raise resolution with fallback, and the cross-source lockstep tests are a nice touch — but there are a few concrete issues to resolve.

### Blockers

- **`get_caller_groups` likely mis-parses real API Gateway claims** (`voc-datalake/lambda/shared/api.py`). The REST API Cognito authorizer serializes the `cognito:groups` array claim as a bracket-wrapped string (e.g. `"[admins]"` or `"[admins, users]"`). The parser handles list / comma / space forms but never strips brackets, so `'admins' not in get_caller_groups(...)` would 403 every real admin, breaking the whole feature in deployment. The unit tests only exercise the clean formats. Strip `[` / `]` (or reuse whatever group-check logic the existing user-administration handler uses) and add a test with the bracketed format.
- **Accidental artifacts committed**: `voc-datalake/frontend/.i18n-tmp/bedrock-i18n-in.json` and `bedrock-i18n-out.json` are translation-tooling scratch files (including a raw Bedrock API response) unrelated to the change. Remove them and add `.i18n-tmp/` to `.gitignore`.

### Should fix

- **Private-API access in `save_model_settings`** (`voc-datalake/lambda/api/settings_handler.py`): `require_admin(app.current_event._data)` reaches into the Powertools event's private `_data` attribute. Use the public `app.current_event.raw_event` instead — `_data` can break on a Powertools upgrade.
- **Verify `datetime`/`timezone` imports**: the diff shows `datetime.now(timezone.utc)` in `save_model_settings` but no import being added to `settings_handler.py`. If they aren't already imported at module top, the PUT path raises `NameError` at runtime. (The passing tests suggest they exist, but the diff doesn't show it — please confirm.)
- **Negative caching after lookup failures**: both `shared/model_config.py` and `model-override.ts` cache the fallback for the full 60s TTL after a transient DynamoDB error. That's a reasonable trade-off, but it means a throttling blip silently reverts inference to the default for up to a minute. Consider a shorter TTL for the error path, or at least document it where the override is consumed.

### Nits

- `test_cdk_iam_grants_cover_every_allowlisted_model` slices `api-stack.ts` at a hardcoded 1500 chars after the helper start — brittle if the helper grows; matching to the next `private ` keyword would be sturdier.
- `mock-server.js` duplicates the allowlist entries verbatim; a comment pointing at `shared/model_config.py` as the source of truth would help keep them in sync.
- In `AiModelSection.tsx`, the radio selection only updates after the query invalidation round-trips, so the clicked option won't appear checked until refetch — the disabled fieldset mostly masks this, but an optimistic update via `onMutate` would feel snappier.
- Locale files define `aiModel.models.*.description` but not `.label`; the code falls back to the server's English label via `defaultValue`. Fine for proper nouns like "Claude Sonnet 4.5", but worth a comment so the missing keys don't look like an oversight to the next i18n pass.
- `model-override.ts` logs the raw non-all

---

mundurragacl:
Round 2 addressed in 328f848 — and the claims-parsing catch was a genuinely good one:

- **Bracket-wrapped claims (blocker)**: `get_caller_groups` now strips the REST authorizer's array serialization (`"[admins]"` / `"[admins, users]"`) before splitting. Without it the admin gate would have 403'd every real admin on deployment. 20 new parametrized tests in `test_auth_groups.py` cover list / space / comma / bracketed / empty formats for both extraction and `require_admin`. (Note: `users_handler`'s local copy of this logic has the same pre-existing gap — it'll get fixed when it switches to the shared helper in the follow-up.)
- **Scratch artifacts (blocker)**: `.i18n-tmp/` files removed from tracking and the directory gitignored. They leaked in via a `git add -A` during the development merge — my mistake.
- **`_data` → `raw_event`**: done.
- **Negative caching**: lookup failures now cache for 10s instead of the full 60s TTL in both the Python and TS resolvers, so a throttling blip can't pin inference to the default for a minute. Documented at both cache sites.
- **datetime/timezone**: confirmed present at module top (the brand handler uses them); the PUT path would fail its 9 tests otherwise.
- **Nits**: CDK lockstep test now slices to the next `private ` method; mock server carries a source-of-truth pointer to `model_config.py`; non-allowlisted values are truncated to 80 chars in logs; the intentionally-absent `.label` locale keys are explained in the component header (product names are identical across languages and the i18n gate rejects same-as-English values). The optimistic-update suggestion is noted but skipped — the disabled fieldset covers the round-trip and the refetch is a single get.

Suites: backend 81 tests across the touched files (764 total), stream 189 + tsc clean, frontend unchanged since last run (1,588).

perrozzi:
This PR adds an admin-only "AI Model" picker in Settings that pins all AI features to one Bedrock model from a curated two-entry allowlist (Sonnet 4.5 / Haiku 4.5), with resolution layers in Python (`shared/model_config.py`), TS streaming (`model-override.ts`), a new `GET/PUT /settings/model` on the existing settings Lambda, and consolidated IAM grants via `allowlistedModelArns()` in `api-stack.ts`. The design is sound: server-side allowlist enforcement on both read and write, never-raise fallback for the lookup, admin gating on the mutation, lockstep tests that read the TS and CDK sources, and i18n keys added for all 8 locales. Overall this is a well-scoped, well-tested change; the remaining items are verifications outside

---

mundurragacl:
The round-3 comment arrived truncated right after "the remaining items are verifications outside" — no sections rendered past the (positive) summary. Assuming that was heading toward *deploy-time verifications outside the diff*, here's the post-merge validation I'd run, happy to execute it after this lands:

1. **Haiku actually invocable**: select Haiku in Settings, run one chat turn + one category generation, confirm 200s and `used_model` in the Lambda logs shows the Haiku inference profile (proves the new IAM grants, not just the lookup).
2. **Admin gate against real Cognito**: `PUT /settings/model` with a non-admin user's token → 403; with an admin token → 200 (validates the bracket-claim parsing against API Gateway's real serialization, not the unit-test fixtures).
3. **Propagation**: flip the model and confirm a warm streaming-chat container picks it up within the 60s TTL.
4. **Automatic restore**: clear the override and confirm the processor stays on Haiku and chat returns to Sonnet.

If the truncated section had concrete code items instead, send them through and I'll take them in the next round.

perrozzi:
Since the last review, the push is not a follow-up on the model picker at all — it's a second, unrelated feature: threading a `date_basis` ('imported' vs 'review') selection through chat bodies, project jobs, the MCP tool, the streaming Lambda's search tool, and the shared feedback query layer (issue #150), plus a refactor moving `basis_date`/`window_cutoff` into `shared/feedback.py`. My previous review recorded no blocking findings on the model-picker changes, and none of that code was touched here.

### New findings

- **Scope creep (should-fix):** the entire incremental push implements issue #150, unrelated to this PR's stated scope (issue #96 model picker). It's substantial (new API semantics, a window-definition behavior change, ~15 files) and should be a separate PR so each feature can be reviewed and reverted independently.
- **Silent behavior change without Python regression tests:** `voc-datalake/lambda/api/metrics_handler.py` changes the cutoff in `get_urgent_feedback` and `search_feedback` from a `days+1`-calendar-day window to the unified `days`-long window (`window_cutoff`). The TS side pins this with a test (`search-feedback.test.ts` "unified definition"), but there's no Python-side test asserting the new boundary for these two existing endpoints — add one, per the regression-test rule for behavior changes.
- **Verify patch target in `test_chat_handler.py`:** the new `TestChatDateBasis` tests patch `shared.converse.converse`. If `chat_handler` does `from shared.converse import converse` at module level, the patch won't intercept the handler's local reference and the prompt assertions would be exercising nothing. Patch `chat_handler.converse` (or confirm the handler calls through the module) — worth a quick check since these are the only tests guarding the new review-basis filtering in chat.
- **Minor:** `voc-datalake/frontend/src/api/projectsApi.ts` spreads `getDateBasisBodyParams()` before `...data`, relying on callers never passing an (untyped) `date_basis` in `data` except deliberately — the test even needs a double cast to exercise it. Consider adding `date_basis` to the request types so the precedence is visible rather than incidental.
- **Nit:** stray double blank lines introduced in `voc-datalake/lambda/shared/api.py` and several test files.

Otherwise the implementation is solid: validation at every boundary (API, job payload, Step Functions input, MCP args), Zod enum on the stream schema, the review-basis post-filter correctly disables the fetch-ceiling early break (with a test), and the shared helper extraction removes the metrics_handler duplication cleanly.

---

mundurragacl:
Round 4 — the incremental push you reviewed (bde26dc) is a **merge of `development`**, not new scope: #155 (date_basis, issue #150) went through its own multi-round review on its own PR and was squash-merged to development (2880d31); this branch then merged development to resolve the resulting conflict in `shared/api.py`. The PR's actual diff against development remains model-picker-only — verified via `gh pr diff 154 --name-only`: 26 files, none of them `metrics_handler.py`, `search-feedback.ts`, or any date-basis surface.

On the individual findings, since they deserve answers even where they land on merged code:

- **Scope creep**: see above — nothing to split; the reviewable diff is unchanged in scope.
- **Python window-boundary tests**: the unified window is pinned at the layer that owns it — `shared/test/test_feedback.py::test_window_cutoff_is_days_long_ending_today` (days-long window ending today, exact boundary assertions). The metrics endpoints now delegate to that helper. Endpoint-level wiring tests would be a fair follow-up on development, but they belong to #155's surface, not this PR.
- **`test_chat_handler.py` patch target**: checked — the handler imports `converse` **function-locally** (inside `chat()`'s try block), so the name resolves from `shared.converse` at call time, after the patch is applied. `patch('shared.converse.converse')` therefore intercepts correctly and the prompt assertions are exercising real behavior.
- **`projectsApi` date_basis typing**: agreed it would make the spread precedence explicit; it's #155 code, so it'll ride with the next date-basis touch on development rather than widening this PR.
- **Blank lines in `shared/api.py`**: the two-blank-line separation between the new top-level functions is PEP 8; scanned the file — zero occurrences of 3+ consecutive blank lines.

Model-picker surface itself is unchanged since your no-blocker round 3; branch is MERGEABLE/CLEAN against post-#155 development with the full gate green (`npm run check` exit 0, 788 backend / 1,600 frontend on the merged branch).

mundurragacl:
Correction on the gate numbers in my last comment: 788/1,600 were measured on a different branch. On THIS branch the verified figures are `npm run check` exit 0 and **801 backend tests passed** (frontend suite passed within the check; I didn't capture its count separately). The substance of the comment stands unchanged.

perrozzi:
The push since my last review is again not a follow-up on the model picker: it bundles three more unrelated features — shared problem-resolution state for Problem Analysis (issue #66: new `/settings/resolved-problems` endpoints, key normalization, UI toggle, PDF annotation), hiding scraper auto-detect for JSON-LD (issue #18), and a PR/FAQ prompt rewrite with contract tests plus a UTF-8 loader fix (issue #93). None of my previous findings (all on the #150 date-basis push) were touched. Note: the full-diff context provided does not include either the previous date-basis push or these new changes, so this review is based on the incremental diff alone.

### Still open
- Scope creep: PR now spans five issues (#96, #150, #66, #18, #93) — this incremental push makes it worse, not better; split into separate PRs.
- No Python regression test for the `metrics_handler.py` window-definition change in `get_urgent_feedback`/`search_feedback`.
- Verify `TestChatDateBasis` patches `chat_handler.converse`, not `shared.converse.converse`.
- `projectsApi.ts` `date_basis` precedence still relies on untyped spread ordering.
- Stray double blank lines (and this push adds more, e.g. `problemResolution.ts`, `test_settings_handler.py`).

### New findings
- **Should-fix — resolution keys can silently orphan:** `buildResolutionKey` uses the client-side similarity group's representative problem text, which shifts with the similarity slider and the data window. Orphaned entries stay in the shared map forever, permanently consuming the 500-entry cap (no TTL, no pruning, and unresolving requires the exact key to reappear). At minimum document this; ideally add server-side pruning by `resolved_at` age. (`voc-datalake/frontend/src/pages/ProblemAnalysis/problemResolution.ts`, `voc-datalake/lambda/api/settings_handler.py`)
- **Minor — global pending lock:** `useProblemResolution.toggleResolved` silently drops clicks while any mutation is pending, and `resolvePending` disables every resolve button page-wide; per-key pending (or a mutation queue) would be less surprising for rapid multi-resolve workflows. (`voc-datalake/frontend/src/pages/ProblemAnalysis/useProblemResolution.ts`)
- **Minor — brittle regex pins:** `TestPrfaqPromptContract.test_steps_forbid_their_own_section_heading` couples the prompt JSON to sentence-level phrasing (`(do not|never|must not) add[^.]*<heading>`); a legitimate rewording fails the build. The comment acknowledges it, but consider pinning a structured field (e.g. an explicit `owned_heading` key in the JSON) instead of prose regexes. (`voc-datalake/lambda/shared/test/test_prompt_utils.py`)
- **Nit:** trailing whitespace on the final assert in `test_unresolve_is_never_capped_and_tolerates_missing_map` (`voc-datalake/lambda/api/test/test_settings_handler.py`).

Within its own scope the #66 implementation is well done — atomic cap enforcement via ConditionExpression, surrogate/byte-cap validation with tests, Zod at the response boundary, and complete i18n coverage across all seven locales — and the #18 and #93 changes are clean with genuine regression tests. The problem remains that none of it belongs in this PR.

---

mundurragacl:
Round 5 — same situation as round 4: the push you reviewed (8e5113d) is the **merge of `development`** after #151, #152, and #153 were merged there; the conflict was in `settings_handler.py`/tests/`client.ts`/`mock-server.js` where #153's resolved-problems endpoints and this PR's model endpoints collided (resolution: keep both). The PR's reviewable diff against development is still model-picker-only — re-verified: `gh pr diff 154 --name-only` = 26 files, zero problem-resolution/scraper/prompt files. Nothing to split: those features merged through their own PRs (#151/#152/#153), each with its own multi-round review.

The 'still open' list therefore refers to code that is on `development`, not in this PR — answered previously in round 4 (window tests exist at the shared-helper layer; `TestChatDateBasis` patches work because the handler's `converse` import is function-local/deferred; `projectsApi` typing rides the next date-basis touch).

Your new findings are on the merged #66 code and deserved tracking rather than dismissal:
- **Orphaned resolution keys + global pending lock**: filed as **#159** with your reasoning (client-derived keys shifting under the similarity slider → permanent cap consumption; per-key pending). Real design gap, right home is a follow-up on development.
- **PR/FAQ regex pins**: this was litigated across ~16 rounds on PR #151 (the structured-field alternative was proposed and declined there with rationale); it merged with the maintainer's approval. Not re-opening here.
- **Trailing whitespace**: checked — the flagged lines are byte-identical to `development` (37 occurrences on both sides of the merge; zero introduced by the conflict resolution).

Model-picker surface remains unchanged since your no-blocker round 3; branch is MERGEABLE/CLEAN with the full root gate green (`npm run check` exit 0, 42 settings tests covering both endpoint sets).

perrozzi:
The push since my last review is yet another new feature, not follow-up work: opt-in public web search via a Bedrock AgentCore Gateway (new `VocWebSearchStack`, Python + TS gateway clients, chat toggle, research-wizard checkbox, i18n across all seven locales), plus a #157 Step Functions wiring fix with CDK template regression tests, a `depsLockFilePath`/`nodeModules` bundling fix for the stream Lambda, and a `test:cdk` npm script. None of the previous findings were addressed.

### Still open
- Scope creep: the PR now spans at least seven concerns (#96, #150, #66, #18, #93, web search/#68, #157) — this push is the largest single addition yet; please split.
- No Python regression test for the `metrics_handler.py` window-definition change.
- Verify `TestChatDateBasis` patches `chat_handler.converse`, not `shared.converse.converse`.
- `projectsApi.ts` `date_basis` precedence still relies on untyped spread ordering.
- #66 resolution-key orphaning (cap consumption, no pruning) and the global pending lock in `useProblemResolution`.
- Brittle regex pins in `test_prompt_utils.py`; stray blank lines/trailing whitespace.

### New findings
- **Should-fix — chat web search failures kill the whole turn:** `executor.test.ts` explicitly pins that `executeWebSearch` rejections propagate, which aborts the entire streaming response on a gateway blip. This contradicts the stated contract ("always an enrichment, never a hard dependency", honored in the research path via degrade-to-empty). Return an error `toolResult` (Converse supports `status: 'error'`) so the model can continue answering from feedback data. (`voc-datalake/lambda/stream/src/tools/executor.ts`, `handler.ts`)
- **Minor — `searching_web` job stage lacks an i18n key:** `step_initialize` emits a new `searching_web` stage via `update_job_status`; if the frontend translates stage names (as other stages are), this will render untranslated in all locales. (`voc-datalake/lambda/research/research_step_handler.py`)
- **Minor — asymmetric unknown-tool fallback:** Python's `_signed_post` wraps `HTTPError` without the response body, so if the gateway reports an unknown tool via HTTP status + body (rather than a JSON-RPC error frame), `_is_unknown_tool_error` can't match and discovery never triggers; the TS client includes the body slice and would self-heal. Include the error body in the Python message for parity. (`voc-datalake/lambda/shared/web_search.py`)
- **Minor — degrade guard narrower than its comment:** `step_initialize` catches only `WebSearchError`, but non-wrapped exceptions (e.g. credential resolution in `_get_credentials`) would still fail the research job despite "a search failure must never fail the research job". Broaden the except around the search block or wrap credential errors too. (`voc-datalake/lambda/research/research_step_handler.py`)
- **Nit — acknowledged brittleness:** the state-machine tests pin CDK's compact JSON serialization (`"key.$":"path"`); the comment covers it, fine as-is. (`voc-datalake/lib/stacks/processing-stack-consolidated.test.ts`)

On its own merits the web-search work is well executed — least-privilege IAM scoped to the concrete gateway ARN, strict-boolean billing gates with tests on both producers, complete i18n coverage, Zod at every untyped boundary, thorough protocol tests on both clients, and genuine regression tests for #157. But like the previous three pushes, it belongs in its own PR.

---

mundurragacl:
Closing as superseded: #166 (the per-surface alternative for the same issue #96) merged into development after its review converged with no open comments, per maintainer direction to merge ready PRs regardless of author. This PR's single-global-toggle shape lives on in #166's backward-compatible legacy 'model_id' path, so nothing from the review work here is lost. The remaining unanswered round on this thread (scope-creep flags) reflected the bot misreading development merges into the diff — documented in earlier replies with 'gh pr diff 154 --name-only' evidence.

PR #156, opened 2026-07-14
feat: opt-in public web search via Bedrock AgentCore (chat + research)

mundurragacl:
## Summary

Fixes #68 (close manually on merge) — re-scoped by the maintainer from a Tavily integration to the native **Web Search Tool on Amazon Bedrock AgentCore** (GA June 2026): a fully managed, MCP-compatible web search behind an AgentCore Gateway. No third-party API keys, queries are served entirely within AWS, and billing is pay-per-use at **$7 per 1,000 queries**.

Two user-facing entry points, both **opt-in per request**:

- **AI Chat** (`/chat`): a "Web search" toggle in the filter bar (per-conversation, persisted). When on, the model gets a `web_search` tool alongside `search_feedback` and uses it for current/external questions (competitors, market context, releases). Web results render as a **cited-links block** under the answer — separate from feedback sources — satisfying the connector's acceptable-use requirement that source citations reach the end user.
- **Projects research**: an "Include public web search" checkbox in the Run Research wizard. `step_initialize` fetches results for the research question, the analysis prompt carries explicit attribution rules (web findings must cite URLs and must not be attributed to customers), and the saved report header discloses that web search was used.

## Architecture

```
chat-stream Lambda (Node)  ──┐   SigV4-signed JSON-RPC (tools/call)
                             ├──────────►  AgentCore Gateway (MCP, AWS_IAM auth)
research-step Lambda (Py)  ──┘             └── connector target: web-search (AWS-managed)
```

- **`VocWebSearchStack`** (new): Gateway (MCP protocol, `AWS_IAM` inbound auth), a `web-search` connector target, and a service role scoped to the concrete gateway ARN plus `bedrock-agentcore:InvokeWebSearch` on the service-owned tool ARN — granted *after* gateway creation so no `gateway/*` wildcard is needed. The target is declared via a raw `CfnResource`: the aws-cdk-lib 2.244 L1 union predates connector targets, while the CloudFormation schema supports them (`AWS::BedrockAgentCore::GatewayTarget ConnectorTargetConfiguration`).
- **Region handling**: the connector exists only in **us-east-1**, so the stack always deploys there. Deployment is **explicitly opt-in** (`"enableWebSearch": true`) in every region until a real gateway round-trip has been validated post-release; non-us-east-1 apps additionally need a us-east-1 bootstrap (CDK cross-region references carry the gateway URL/ARN). Documented in `docs/deployment.md`.
[...]

PR #158, opened 2026-07-14
fix: research reference documents never reach the analysis prompt

### LINKED ISSUE #157: [Bug]: Research reference documents never reach the analysis prompt via Step Functions
## What happens

The research wizard lets you select reference documents, and `step_initialize` in `voc-datalake/lambda/research/research_step_handler.py` dutifully builds a `documents_context` section from them and returns it.

But the Step Functions state machine drops it: `InitializeResearch`'s `resultSelector` in `voc-datalake/lib/stacks/processing-stack-consolidated.ts` only selects `feedback_context`, `feedback_stats`, `feedback_count`, and `personas_context` — so `documents_context` never lands in `$.initialize_result`, and `AnalyzeFeedback`'s payload never forwards it. `step_analyze` reads `event.get('documents_context', '')` and always gets `''` when running under Step Functions.

Net effect: selecting reference documents for research silently does nothing (except in the Lambda-fallback path when `RESEARCH_STATE_MACHINE_ARN` is unset, where the whole event flows through).

## Fix

Add `'documents_context.$': '$.Payload.documents_context'` to the initialize `resultSelector` and `'documents_context.$': '$.initialize_result.documents_context'` to the analyze payload. `step_initialize` already always returns the key, so the unconditional JSONPath reference is safe.

Found while wiring `web_context` through the same selector for #156 (which pins its own key's always-present contract with a test — the same pattern this fix should adopt).

mundurragacl:
## Summary

Fixes #157 (close manually on merge).

The research wizard lets users select reference documents, and `step_initialize` builds a `documents_context` section from them — but the Step Functions state machine dropped it: `InitializeResearch`'s `resultSelector` only selected `feedback_context`/`feedback_stats`/`feedback_count`/`personas_context`, and `AnalyzeFeedback`'s payload never forwarded it. Under Step Functions, `step_analyze` always saw `''` and selected documents silently did nothing (only the `RESEARCH_STATE_MACHINE_ARN`-unset Lambda fallback path worked).

## Change

Two lines in `processing-stack-consolidated.ts`:
- `'documents_context.$': '$.Payload.documents_context'` in the initialize `resultSelector`
- `'documents_context.$': '$.initialize_result.documents_context'` in the analyze payload

`step_initialize` already always returns the key (`''` when no documents are selected), so the unconditional JSONPath reference cannot fail the state.

## Tests

Per the bug-fix regression rule, pinned at the two layers Python can reach:
- `test_documents_context_key_always_present` — the always-present contract the unconditional selector depends on (a refactor dropping the key would fail `InitializeResearch` outright).
- `TestStepAnalyze.test_includes_documents_context` — documents reaching the analysis prompt through `step_analyze`.

The selector wiring itself is CDK template surface (no CDK test harness in this repo): verified via `cdk synth` — the state machine definition now carries 4 `documents_context` references (selector key/path + payload key/path). Full backend suite: 759 passed. CDK `tsc` clean.

## Merge-order note

PR #156 touches the same `resultSelector` block (adds `web_context`). Whichever merges second will need a trivial two-line conflict resolution — both keys belong in the final selector. Found while wiring #156; same always-present-contract pattern used there.


perrozzi:
This PR fixes a real wiring bug: `step_initialize` builds `documents_context` from user-selected reference documents, but the Step Functions definition in `processing-stack-consolidated.ts` dropped it — the `InitializeResearch` `resultSelector` never selected it and the `AnalyzeFeedback` payload never forwarded it, so documents silently never reached the analysis prompt under Step Functions. The two-line fix is correct and minimal, the unconditional `$.Payload.documents_context` reference is safe because the handler always returns the key (now pinned by a test), and the change is well documented. Overall this looks good; no blockers.

### Should fix
- **The regression tests don't actually fail if the fix is reverted.** Both new tests in `voc-datalake/lambda/research/test/test_research_steps.py` exercise the Python layer (`step_initialize`'s always-present contract and `step_analyze`'s prompt inclusion), which already passed before this PR — the actual bug and fix live in the CDK state-machine definition. Per the repo's bug-fix rule, consider adding a vitest test using `aws-cdk-lib/assertions` (`Template.fromStack(...)`) asserting the `InitializeResearch` `ResultSelector` and `AnalyzeFeedback` payload both contain `documents_context.$`. The PR notes there's no CDK test harness today, but this fix is a good reason to introduce one — a `cdk synth` spot-check won't catch a future regression (e.g., the PR #156 conflict resolution accidentally dropping one of the two lines).

### Nits
- The five-line comment inside the `resultSelector` in `processing-stack-consolidated.ts` is verbose for inline CDK code; a one-liner pointing at issue #157 and the always-present contract would suffice — the full rationale belongs in the commit message (where it already is).
- `feedback_count` is selected in `InitializeResearch` but never forwarded to `AnalyzeFeedback` — pre-existing and out of scope, but worth a follow-up check that it's intentional (used elsewhere) rather than the same class of silent drop this PR fixes.
- Coordinate merge order with PR #156 as noted — both `documents_context` and `web_context` must survive the conflict resolution in the same selector block; a CDK assertions test (above) would guard exactly this.

---

mundurragacl:
Round 1 addressed in 2a36f35:

- **Should-fix (regression test)**: agreed, and my PR note about 'no CDK test harness' was wrong — `voc-datalake` already has a vitest config covering `lib/**`. Added `processing-stack-consolidated.test.ts`: synthesizes the stack (bundling skipped via `aws:cdk:bundling-stacks`) and asserts `documents_context.$` in both the `InitializeResearch` resultSelector and the `AnalyzeFeedback` payload, plus the sibling context keys (`feedback_context`/`feedback_stats`/`personas_context`) that share the silent-drop failure mode. Mutation-verified: removing the selector line fails the test; restoring passes. This also guards exactly the #156 conflict-resolution hazard you flagged.
- **Verbose comment nit**: trimmed to one line pointing at #157; the rationale lives in the commit message.
- **`feedback_count` nit**: checked — it's not the same class of drop. It IS forwarded, just to a different consumer: `SaveResearchResults`' payload takes `'feedback_count.$': '$.initialize_result.feedback_count'` (the saved report header shows the analyzed-item count). `step_analyze` doesn't read it, so omitting it from the analyze payload is intentional.
- **Merge order with #156**: coordinated — and now enforced by the new test rather than by care.

Gates: 33 CDK vitest / 35 research pytest green, `tsc` clean.

perrozzi:
Since my last review, the author added `voc-datalake/lib/stacks/processing-stack-consolidated.test.ts` — a CDK template-level vitest suite asserting both halves of the `documents_context` wiring (selector + payload) plus the sibling context keys — and trimmed the verbose inline comment in the `resultSelector` down to one line. Both of my previous items are addressed.

### Resolved
- CDK-level regression coverage added: the new `Template.fromStack` tests fail if either the `InitializeResearch` selector entry or the `AnalyzeFeedback` payload forwarding is dropped (e.g., in the PR #156 conflict resolution) — exactly the guard I asked for.
- Inline comment in `processing-stack-consolidated.ts` shortened to a one-liner referencing #157.

### New findings
- Nit: in `processing-stack-consolidated.test.ts`, `researchDefinition(synthProcessingTemplate())` (including an `expect(...).toHaveLength(1)` assertion) runs in the `describe` body during collection, not inside a test. If synth throws or a second state machine is ever added, the whole file errors at collection time instead of failing a named test. Wrapping the synth in a `beforeAll` (or lazily inside the first `it`) would give cleaner failure reporting.
- Nit: the assertions are raw substring matches against the stringified `Fn::Join` — acceptable for this shape, but they assert the JSONPath values (`$.Payload.documents_context`), not the key names (`documents_context.$`). A refactor that kept the path under a different key would still pass; asserting the key too would tighten the pin cheaply.
- Worth confirming the repo's vitest config includes `voc-datalake/lib/**/*.test.ts` in its test glob — a regression test that never runs in CI is no guard.

Otherwise the incremental changes are clean and the PR is in good shape to merge, pending merge-order coordination with #156 (which the new sibling-keys test now helpfully protects).

---

mundurragacl:
Round 2 addressed in ccd0855 — took all three nits:

- **Collection-time synth**: moved into `beforeAll`; a synth failure now reports as a named test failure instead of erroring the file at collection.
- **Key + path pins**: assertions now match the exact `"key.$":"path"` pairs. To make that possible the helper joins the raw `Fn::Join` string fragments (real quotes) rather than stringifying the whole object (escaped quotes). Mutation-verified on the payload half this round (round 1 verified the selector half).
- **CI wiring**: fair catch — the `voc-datalake` vitest suite existed but wasn't in the documented gate. Root `check` now runs `test:cdk` between the frontend and backend suites, so this guard (and the two pre-existing lib tests) actually execute. Verified: `npm run check` exit 0 with the new step included.

On merge-order with #156: agreed the sibling-keys test now enforces it mechanically — whichever side resolves the selector conflict, dropping either `documents_context` or (post-#156) `web_context` breaks a named test.

perrozzi:
Since the last review, the author addressed all three of my nits: the synth now runs in a `beforeAll` (with the definition extractor made type-safe on the `Fn::Join` shape), the assertions now pin exact `"key.$":"path"` pairs instead of bare path substrings, and a `test:cdk` script was added to `package.json` and wired into `check`, ensuring the new CDK test actually runs. Two Python regression tests (`test_documents_context_key_always_present`, `TestStepAnalyze.test_includes_documents_context`) round out the coverage per the bug-fix regression rule.

### Resolved
- Synth moved out of the `describe` body into `beforeAll`, so a synth failure reports as a test failure rather than a collection error.
- Assertions tightened to full `"documents_context.$":"$.Payload.documents_context"` key/path pairs (and likewise for siblings), closing the key-rename loophole.
- Test discoverability confirmed by construction: `test:cdk` runs the `voc-datalake` vitest suite and is part of the root `check` script, so the regression guard executes in the standard verification path.

### New findings
- Nit (non-blocking): the exact-string assertions in `processing-stack-consolidated.test.ts` depend on CDK serializing the definition JSON without whitespace between key and value. That's stable today, but a CDK formatting change would fail all three tests at once with a confusing "not found" message; a small comment noting the compact-JSON assumption (or a regex tolerant of whitespace) would future-proof it. Fine to ship as-is.
- Worth a one-time confirmation that `voc-datalake`'s vitest config includes `lib/**/*.test.ts` in its glob (the `test:cdk` script only helps if the file is picked up) — the reported passing run suggests it is.

Otherwise the incremental changes are clean; this is ready to merge, pending the noted conflict coordination with #156 (which the sibling-keys test now guards).

---

mundurragacl:
Round 3 addressed in the latest push:

- **Compact-JSON assumption**: documented with a NOTE on the describe block — if a CDK upgrade ever pretty-prints definitions, all three tests fail together and the comment points straight at the fix (loosen to whitespace-tolerant matching). Kept the exact pins rather than regexes: they're the tighter guard while the assumption holds.
- **vitest glob confirmation**: verified — `voc-datalake/vitest.config.ts` includes `lib/**/*.test.ts`, and the suite runs 3 files / 34 tests through `test:cdk` inside root `check`.

Nothing else outstanding from my side; ready for maintainer merge, with the #156 conflict-coordination note standing.

perrozzi:
The only change since my last review is a five-line comment in `processing-stack-consolidated.test.ts` documenting the compact-JSON serialization assumption behind the exact `"key.$":"path"` assertions — exactly the future-proofing note I suggested. No code or test logic changed.

### Resolved
- The compact-JSON assumption is now documented in the test file, with guidance to loosen to a whitespace-tolerant match if a CDK upgrade ever pretty-prints definitions.

### Still open
- (Informational only, not blocking) The one-time confirmation that `voc-datalake`'s vitest glob picks up `lib/**/*.test.ts` — the reported passing `test:cdk` run effectively answers this, so I consider it settled.

The incremental change is clean. This is ready to merge, pending the previously noted conflict coordination with #156.

---

mundurragacl:
(Count correction: 3 files / **33** tests, not 34.)

perrozzi:
This PR fixes a real silent-drop bug: `step_initialize`'s `documents_context` was never selected by `InitializeResearch`'s `resultSelector` nor forwarded into the `AnalyzeFeedback` payload, so selected reference documents never reached the analysis prompt under Step Functions. The two-line CDK fix is correct and follows the same always-present-key pattern used for `web_context` (already in the base per the diff context), and the change comes with regression tests at both the Python contract layer and the synthesized template layer. Overall a well-scoped, well-tested fix — one wiring concern below.

### Blockers
- **`test:cdk` harness may not exist.** `package.json` adds `"test:cdk": "npm run test --prefix voc-datalake"` and puts it in `check`, and the new `voc-datalake/lib/stacks/processing-stack-consolidated.test.ts` imports `vitest` — but the diff contains no changes to `voc-datalake/package.json` (a `test` script, a `vitest` devDependency) or a vitest config for the CDK package. The PR description itself says there is "no CDK test harness in this repo." If that harness isn't already present, `npm run check` will fail for everyone on merge. Either the diff is missing those files or this needs to be added; please confirm/include `voc-datalake/package.json` + vitest config in this PR.

### Should fix
- **Vitest test collection scope.** If vitest runs from `voc-datalake/` root, verify its config excludes `voc-datalake/frontend/**` — otherwise `test:cdk` will re-collect (or conflict with) the frontend suite that the top-level `test` script already runs, doubling CI time or producing config-mismatch failures.
- The description says the fix guards against a conflict-resolution regression with PR #156's `web_context`, but the diff shows `web_context` already in the base. Confirm this branch is rebased on the merged #156 so the "sibling context selections" test (`processing-stack-consolidated.test.ts`) also pins `web_context` — currently the sibling test only covers `feedback_context`, `feedback_stats`, `personas_context`. Adding `web_context` to that loop would close the exact conflict-resolution risk the PR describes.

### Nits
- The exact `'"key.$":"path"'` string matches assume CDK's compact JSON serialization of `DefinitionString`. The comment acknowledges this; a whitespace-tolerant regex (or parsing the joined fragments as JSON after stubbing refs) would be more durable now rather than after a CDK upgrade breaks three tests at once.
- `processing-stack-consolidated.test.ts`: the `const state = { definition: '' }` holder is a workaround; `let definition: string;` assigned in `beforeAll` is the more conventional vitest pattern.
- In `test_research_steps.py`, the extra blank line added between the new `test_documents_context_key_always_present` and `class TestStepAnalyze` introduces a triple blank line — check it passes the repo's Python lint.
- The two Python tests reference "issue #157" in docstrings — helpful; consider also referencing it in the CDK inline comment on the analyze-payload line (only the selector line got the `— see #157` note).

The diff appears complete for the described change; review is based on the full diff as provided.

---

perrozzi:
Post-merge disposition of the final review (posted on the rebased head `8da50f2`, two minutes before the squash merge — addressing it here for the record):

- **Blocker (test:cdk harness may not exist): false positive, verified empirically.** The harness pre-exists on `development` (`voc-datalake/package.json` has `"test": "vitest run"` with vitest ^4.0.16, plus `voc-datalake/vitest.config.ts`); the review only sees the PR diff, which is why it couldn't find them. The predicted failure does not occur: `npm run check` exits 0 on the merged tip `ff3d5e9` (lint + typecheck×3 + frontend/CDK/backend suites).
- **Vitest collection scope: moot** — `voc-datalake/vitest.config.ts` includes only `lib/**/*.test.ts`, so the frontend suite is never re-collected by `test:cdk`.
- **Sibling-loop `web_context` pin: agreed, good catch.** The rebase resolution kept both `web_context` and `documents_context` at both selector sites, and the `documents_context` pins pass — but nothing currently guards `web_context`'s selector lines against the same silent-severing class this PR fixed. Follow-up: add `'web_context'` to the sibling-keys loop in `processing-stack-consolidated.test.ts`.
- Nits: the triple-blank-line concern would fail lint if present (check is green); the compact-JSON assumption and state-holder pattern are documented/deliberate per the earlier rounds.

PR #162, opened 2026-07-15
test(cdk): pin web_context in the sibling context selections guard

perrozzi:
## Summary
Follow-up to #158's final review round (its one legitimate open item): the "sibling context selections" guard in `processing-stack-consolidated.test.ts` pins every initialize output that must flow to the analyze prompt — except `web_context` (#156), whose selector lines had no test coverage at all. One-word change: add `'web_context'` to the loop, pinning it at both wiring sites (initialize `resultSelector` + analyze payload).

## Why it matters
If a future edit drops the `web_context` selector line, the failure is the worst variant of the #157 silent-severing class: searches still run and bill ($7/1k queries), the report still stamps "Web search: enabled" (the disclosure reads the config flag, not the context), and the grounding silently never reaches the analysis — no error anywhere. This exact selector block was in a merge conflict twice in the past day (the #156 rebase and the #158 rebase over it); it accretes a sibling line per context feature and will conflict again.

## Testing
- CDK suite 33/33, tsc clean.
- Fail-on-revert verified: temporarily removing the `web_context` Payload selector line from the stack makes exactly the sibling test fail (`× keeps the sibling context selections intact`), stack restored byte-identical afterwards.

### DIFF
--- a/voc-datalake/lib/stacks/processing-stack-consolidated.test.ts
+++ b/voc-datalake/lib/stacks/processing-stack-consolidated.test.ts
@@ -96,7 +96,7 @@ describe('research state machine wiring (issue #157)', () => {
   it('keeps the sibling context selections intact', () => {
     // The same silent-drop failure mode applies to every initialize output
     // the analyze prompt consumes; pin the full set that must flow.
-    for (const key of ['feedback_context', 'feedback_stats', 'personas_context']) {
+    for (const key of ['feedback_context', 'feedback_stats', 'personas_context', 'web_context']) {
       expect(state.definition).toContain(`"${key}.$":"$.Payload.${key}"`);
       expect(state.definition).toContain(`"${key}.$":"$.initialize_result.${key}"`);
     }


perrozzi:
This is a one-word follow-up that adds `web_context` to the existing "sibling context selections" guard test, pinning the initialize `resultSelector` and analyze payload wiring for the web search context (#156) alongside the other contexts already covered. The change is minimal, matches the established test pattern in the file, and the PR describes a verified fail-on-revert check against the actual selector line in the stack — exactly the regression-test discipline this repo asks for. Looks good to merge.

### Nits
- `voc-datalake/lib/stacks/processing-stack-consolidated.test.ts`: the assertions match exact serialized JSON (`"${key}.$":"$.Payload.${key}"` with no whitespace), so they're coupled to CDK's definition serialization format. This is a pre-existing property of the test, not introduced here — but if this guard keeps accreting keys (as the description notes it will), consider parsing the definition JSON and asserting on the object structure instead, which would also survive formatting changes.
- The loop is now four hardcoded keys that must stay in sync with the stack's selector block. If a shared constant listing the initialize outputs existed in the stack code, importing it here would make new context features self-registering in this test rather than requiring a sibling edit (and another merge conflict) each time.

No blockers or should-fix items — clean change.

---

PR #163, opened 2026-07-15
fix: expire orphanable problem resolutions, per-key pending state

### LINKED ISSUE #159: [Enhancement]: Prune orphaned resolved-problem keys and use per-key pending state
Follow-ups on the #66 problem-resolution feature (merged in PR #153), raised by review on an unrelated PR:

1. **Orphaned resolution keys consume the cap forever.** `buildResolutionKey` derives from the client-side similarity group's representative problem text, which shifts with the similarity slider and the data window. Once a group re-forms differently, its old key can never be unresolved from the UI (unresolve requires the exact key to reappear) and permanently occupies one of the 500 `MAX_RESOLVED_ENTRIES` slots — there is no TTL or pruning. Suggested: server-side pruning of entries whose `resolved_at` exceeds a configurable age (e.g. 180 days) on write, and/or an admin-visible count with a clear-stale action.

2. **Global pending lock on resolve/unresolve.** `useProblemResolution.toggleResolved` drops clicks while any mutation is in flight and disables every resolve button page-wide. Per-key pending state (or a small mutation queue) would make rapid multi-resolve workflows less surprising.

Both are quality-of-life items, not correctness bugs: the cap is atomic and validated, and the lock fails safe.

mundurragacl:
## Summary

Fixes #159 (close manually on merge) — both findings from the #66 follow-up.

### 1. Orphaned resolution keys no longer consume the cap forever

Resolution keys are derived client-side from similarity-group representative text, which shifts with the similarity slider and the data window. When a group re-forms differently, its old key becomes unreachable from the UI — un-unresolvable, permanently holding one of the 500 `MAX_RESOLVED_ENTRIES` slots.

Entries now **expire after `RESOLVED_PROBLEMS_TTL_DAYS`** (default **180**, env-overridable on the settings Lambda, `0` disables):

- **GET** filters expired entries from responses — an old resolution resurfaces for re-review (deliberate product behavior: a problem resolved half a year ago that still appears in current data deserves another look).
- **Cap-pressure pruning**: a resolve that hits the entry cap prunes expired entries from storage and retries once; the cap error only surfaces when the map is full of *live* entries. Pruning is deliberately lazy — the endpoint's documented one-write-per-request steady state is preserved, and a scheduled sweeper (extra infra) stays unnecessary.
- **Malformed entries** (missing/non-string `resolved_at`) count as expired: they can never be displayed or aged out normally and must not hold slots.
- REMOVE expressions are **chunked** (20 keys/update) to stay far below DynamoDB's 4KB expression limit.
- The key-lifecycle caveat is documented on `buildResolutionKey` (the issue's "at minimum document this").

### 2. Per-key pending instead of a page-wide lock

`useProblemResolution` now tracks a `pendingKeys` set (via `onMutate`/`onSettled`): toggling one problem disables only its own button, so rapid multi-resolve workflows stay responsive. The same-key double-click race protection (SET/REMOVE ordering) is preserved — a second toggle of an in-flight key is still dropped. `SubcategoryRow` computes per-problem disabled state from the set; `ProblemRow` is unchanged.

## Tests

- **Backend (+5)**: GET expiry filtering, TTL=0 disable, malformed-as-expired, cap-pressure prune-then-succeed (asserts only the expired key is REMOVEd), chunking (45 stale → 3 updates, ≤20 aliases each). The existing at-cap rejection test now explicitly configures "nothing prunable" so the 400 contract stays pinned.
- **Frontend (+5)**: four hook tests (only the toggled key pending; same-key double-toggle dropped; different-key concurrent toggles both fire; failed mutation un-pends the key) and one row test (only the pending key's button disabled).
- Full root gate green: lint, typecheck (frontend/CDK/stream), frontend + CDK + backend suites (823 backend). No new user-visible strings, so no i18n changes.

## Notes

- `RESOLVED_PROBLEMS_TTL_DAYS` is read from the Lambda environment with an in-code default of 180; no CDK plumbing added until someone actually needs a different value per environment (console-level override works today).
- PR #154 (model picker) touches the same `settings_handler.py` — different functions, but adjacent regions; whichever merges second may see a trivial conflict.

### DIFF
--- a/voc-datalake/frontend/src/pages/ProblemAnalysis/ProblemAnalysis.tsx
+++ b/voc-datalake/frontend/src/pages/ProblemAnalysis/ProblemAnalysis.tsx
@@ -288,7 +288,7 @@ export default function ProblemAnalysis() {
   // clears it from everyone's default view. All query/mutation wiring lives
   // in the hook (complexity budget + useSettingsSync convention).
   const {
-    resolvedMap, resolvedLoading, togglePending, toggleFailed,
+    resolvedMap, resolvedLoading, pendingKeys, toggleFailed,
     toggleResolved, dismissToggleError,
   } = useProblemResolution(!!config.apiEndpoint)
 
@@ -644,7 +644,7 @@ export default function ProblemAnalysis() {
                         expandedProblems={expandedProblems}
                         onToggleProblem={toggleProblem}
                         onToggleResolved={toggleResolved}
-                        resolvePending={togglePending}
+                        pendingKeys={pendingKeys}
                       />
                     )
                   })}
--- a/voc-datalake/frontend/src/pages/ProblemAnalysis/SubcategoryRow.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProblemAnalysis/SubcategoryRow.test.tsx
@@ -6,6 +6,7 @@ import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { MemoryRouter } from 'react-router-dom'
 import { SubcategoryRow } from './SubcategoryRow'
+import { buildResolutionKey } from './problemResolution'
 
 const mockSubcategoryGroup = {
   subcategory: 'shipping_speed',
@@ -225,6 +226,27 @@ describe('SubcategoryRow', () => {
 
       // When problem is expanded, feedback items should be visible
       expect(screen.getByText('Delivery was slow')).toBeInTheDocument()
+    })
+  })
+
+  describe('per-key pending (issue #159)', () => {
+    it('disables only the resolve button whose key is pending', () => {
+      renderWithRouter(
+        <SubcategoryRow
+          categoryName="delivery"
+          subcategoryGroup={mockSubcategoryGroup}
+          isExpanded={true}
+          onToggle={vi.fn()}
+          expandedProblems={new Set()}
+          onToggleProblem={vi.fn()}
+          onToggleResolved={vi.fn()}
+          pendingKeys={new Set([buildResolutionKey('delivery', 'shipping_speed', 'Slow delivery times')])}
+        />
+      )
+
+      const [slowDeliveryButton, damagedButton] = screen.getAllByRole('button', { name: /resolved/i })
+      expect(slowDeliveryButton).toBeDisabled()
+      expect(damagedButton).not.toBeDisabled()
     })
   })
 })
--- a/voc-datalake/frontend/src/pages/ProblemAnalysis/SubcategoryRow.tsx
+++ b/voc-datalake/frontend/src/pages/ProblemAnalysis/SubcategoryRow.tsx
@@ -11,7 +11,9 @@ interface SubcategoryRowProps {
   readonly expandedProblems: Set<string>
   readonly onToggleProblem: (key: string) => void
   readonly onToggleResolved: (key: string, resolved: boolean) => void
-  readonly resolvePending?: boolean
+  /** Resolution keys with an in-flight toggle — disables only their own
+   * button (issue #159: no page-wide lock). */
+  readonly pendingKeys?: ReadonlySet<string>
 }
 
 export function SubcategoryRow({
@@ -22,7 +24,7 @@ export function SubcategoryRow({
   expandedProblems,
   onToggleProblem,
   onToggleResolved,
-  resolvePending,
+  pendingKeys,
 }: SubcategoryRowProps) {
   const subcategoryKey = `${categoryName}:${subcategoryGroup.subcategory}`
 
@@ -66,7 +68,7 @@ export function SubcategoryRow({
                 isExpanded={expandedProblems.has(problemKey)}
                 onToggle={() => onToggleProblem(problemKey)}
                 onToggleResolved={() => onToggleResolved(resolutionKey, !problemGroup.resolved)}
-                resolvePending={resolvePending}
+                resolvePending={pendingKeys?.has(resolutionKey) === true}
               />
             )
           })}
--- a/voc-datalake/frontend/src/pages/ProblemAnalysis/problemResolution.ts
+++ b/voc-datalake/frontend/src/pages/ProblemAnalysis/problemResolution.ts
@@ -98,6 +98,14 @@ function fitToKeyCaps(key: string): string {
  * normalized out of the values themselves so a literal pipe in a category
  * name can't merge two different problems under one key. The result is
  * deterministically truncated to the server's 255-char/255-byte caps.
+ *
+ * Lifecycle caveat (issue #159): the problem component is the similarity
+ * group's REPRESENTATIVE text, which can shift as the similarity slider or
+ * the data window changes — a resolution whose group re-forms differently
+ * becomes unreachable from the UI. The server compensates by expiring
+ * entries after RESOLVED_PROBLEMS_TTL_DAYS (default 180): expired
+ * resolutions drop out of GET responses (the problem resurfaces for
+ * re-review) and are pruned from storage under entry-cap pressure.
  */
 export function buildResolutionKey(category: string, subcategory: string, problem: string): string {
   const normalize = (value: string) =>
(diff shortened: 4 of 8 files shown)


perrozzi:
This PR addresses two follow-ups from issue #66: server-side TTL expiry (with lazy cap-pressure pruning) for shared resolved-problem entries whose client-derived keys can be orphaned, and per-key pending state in `useProblemResolution` so one in-flight toggle no longer locks every resolve button. The design is well-reasoned (lazy pruning preserves the one-write steady state, chunked REMOVEs respect DynamoDB expression limits) and test coverage on both sides is good, including a regression pin on the at-cap 400 contract. Overall solid; a couple of concurrency/robustness gaps worth addressing.

### Should fix

- **`toggleFailed` can silently swallow failures under concurrent toggles** (`voc-datalake/frontend/src/pages/ProblemAnalysis/useProblemResolution.ts`). All toggles share one `useMutation`, and `mutation.isError` only reflects the *latest* `mutate()` call. This PR's explicit goal is rapid multi-resolve workflows: if the user toggles key A, then key B while A is in flight, and A then fails, the observer is already tracking B — A's failure never surfaces via `toggleFailed`. `pendingKeys` stays correct (per-call `onSettled`), but the user gets no error feedback for A. Track failure in state via `onError` (like `pendingKeys`) instead of `mutation.isError`. The new test only covers the single-mutation failure case, so this gap isn't caught.
- **Module-import crash on malformed env var** (`voc-datalake/lambda/api/settings_handler.py`). `RESOLVED_PROBLEMS_TTL_DAYS = int(os.environ.get('RESOLVED_PROBLEMS_TTL_DAYS', '180'))` runs at import; a non-numeric value (e.g. `"180d"` typo in the console) raises `ValueError` at cold start and takes down the *entire* settings Lambda, not just the resolved-problems feature. Parse defensively (fall back to 180 and log a warning) — validation at system boundaries is a repo rule, and the Lambda environment is a boundary here.
- **Prune read-then-REMOVE race can delete a freshly re-resolved entry** (`_prune_expired_entries`). Expired entries are deliberately filtered from GET so the problem resurfaces — meaning a user can re-resolve the *same key* (fresh `resolved_at`) between the prune's `get_item` snapshot and its unconditional `REMOVE`, deleting the fresh entry. The window is small and self-healing (the user can resolve again), so a conditional REMOVE on `resolved_at < :cutoff` per key may be overkill — but at minimum document the race where the snapshot-based `stale` list is built.

### Nits

- `_resolve_problem_key`: the final `ValidationError` is raised outside the `except` block, dropping the previous `from e` exception chaining — the original `ConditionalCheckFailedException` context is lost from logs.
- GET path: `_without_expired` calls `.items()` on whatever is stored under `resolved`; a non-dict value (malformed storage) now raises → 500, whereas `_prune_expired_entries` handles that case explicitly. Consider the same `isinstance` guard for symmetry.
- `logger.info(f"Pruned {len(stale)} ...")` — prefer Powertools structured logging (`logger.info("Pruned expired entries", extra={"count": ...})`) over f-string interpolation, consistent with the Powertools convention.
- Other write helpers in the handler carry `@tracer.capture_method`; `_prune_expired_entries` (which does N+1 DynamoDB calls) would benefit from it too.
- The per-key double-click guard (`if (pendingKeys.has(key)) return`) relies on `onMutate`'s state update flushing before the second click, same as the old `mutation.isPending` guard — behavior parity is fine, but the comment claims it "covers the render gap," which is only approximately true; the server-side SET/REMOVE ordering remains the real backstop.

Otherwise clean: no new user-visible strings (i18n unaffected), no IAM/Bedrock changes, chunking test correctly pins the ≤20-alias invariant, and deferring `RESOLVED_PROBLEMS_TTL_DAYS` CDK plumbing is a reasonable call given the documented console override.

---

mundurragacl:
Round 1 addressed in d398d9c — all three should-fixes plus every nit:

- **Concurrent failure swallowing**: exactly right, and it undermined this PR's own goal. `toggleFailed` is now hook state set in `onError` (same pattern as `pendingKeys`), `dismissToggleError` clears the state instead of resetting the mutation. New test pins the scenario you described: A fails, B succeeds afterwards, `toggleFailed` stays true and is dismissable.
- **Env-parse crash**: extracted `_parse_ttl_days` with a defensive fallback to 180 + structured warning — a console typo can no longer take down the whole settings Lambda at cold start. Tested against `'180d'`, `''`, `None`, plus valid values.
- **Prune snapshot race**: documented in `_prune_expired_entries`' docstring with the accepted-risk rationale (millisecond window, same-key only, self-healing by re-resolving) and why per-key conditional REMOVEs weren't taken (N extra conditional writes for a race that costs one click).

Nits, all taken: the cap `ValidationError` chains `from` the original `ConditionalCheckFailedException` again (both failure paths); `_without_expired` got the same `isinstance` guard as pruning, with a malformed-storage GET test; the prune log is structured (`extra={'count': ...}`); `@tracer.capture_method` on `_prune_expired_entries`; and the double-click guard comment now correctly credits the atomic single-write as the real backstop rather than claiming to close the render gap.

Gates: root `npm run check` exit 0; 36 settings-handler tests, 53 ProblemAnalysis frontend tests.

perrozzi:
The new commits directly address every item from the previous review: `toggleFailed` is now tracked in state via `onError` (with a regression test for the masked-early-failure scenario), the TTL env var is parsed defensively with a warning fallback, the prune snapshot race is documented, exception chaining is restored via `cap_error`, `_without_expired` gained the malformed-storage guard (with a test), logging is structured, and `_prune_expired_entries` carries `@tracer.capture_method`. The at-cap 400 test was also updated to explicitly configure "nothing prunable," keeping that contract pinned.

### Resolved
- `toggleFailed` masked under concurrent toggles — now state-based via `onError`, with regression test (`useProblemResolution.ts` / `.test.tsx`).
- Module-import crash on malformed `RESOLVED_PROBLEMS_TTL_DAYS` — `_parse_ttl_days` degrades to 180 with a warning, tested including `None`/empty/garbage.
- Prune read-then-REMOVE race — documented in `_prune_expired_entries` docstring with an explicit accepted-risk rationale.
- Lost exception chaining in `_resolve_problem_key` — `cap_error` captured in both except branches and the final `ValidationError` raises `from cap_error` (bound on every path that reaches it, verified).
- `_without_expired` non-dict 500 — `isinstance` guard added, with test.
- f-string log → structured `extra={"count": ...}`; `@tracer.capture_method` added to `_prune_expired_entries`; misleading "covers the render gap" comment reworded.

### New findings
- Nit: `_parse_ttl_days(raw: str)` type hint is inaccurate — the test deliberately passes `None` (and `TypeError` is caught for that reason), so the signature should be `raw: str | None` (`voc-datalake/lambda/api/settings_handler.py`).
- Observation, not a blocker: with state-based `toggleFailed`, a later *successful* toggle no longer auto-clears the error flag (old `mutation.isError` behavior did on the next mutate). Sticky-until-dismissed is arguably the better UX for concurrent toggles, but confirm it's intentional since the dismiss button is now the only way to clear it.

Otherwise clean — the incremental changes are focused, well-tested (both new backend tests and the concurrency regression test are meaningful, not just coverage padding), and introduce no i18n/IAM/Bedrock concerns.

---

mundurragacl:
Round 2 addressed:

- **Type hint**: `_parse_ttl_days(raw: str | None)` — pushed.
- **Sticky `toggleFailed`**: intentional, confirming the reasoning — auto-clear-on-next-mutate is exactly the masking mechanism round 1 flagged (a later success erasing an earlier failure the user never saw). With concurrent toggles the flag stays until the user dismisses the banner, so a failure can't vanish underneath them. If UX feedback later prefers auto-clear, the right shape would be per-key error state, not a shared reset.

36 settings tests green. Nothing else outstanding from my side.

perrozzi:
Since the last review, a single one-line change was pushed: the `_parse_ttl_days` signature was corrected to `raw: str | None`, addressing the remaining nit.

### Resolved
- Inaccurate `_parse_ttl_days(raw: str)` type hint — now `raw: str | None`, matching the caught `TypeError` path and the `None` test case (`voc-datalake/lambda/api/settings_handler.py`).

### Still open
- Observation from last round (not a blocker): sticky `toggleFailed` cleared only via dismiss — no change pushed, so treating the current behavior as intentional.

No new findings — the incremental change is clean. LGTM.

---

PR #164, opened 2026-07-15
fix: collision-proof chat conversation and message IDs

### LINKED ISSUE #160: [Bug]: chatStore Date.now() IDs collide within the same millisecond (flaky test + duplicate-ID risk)
### What happened

`chatStore` generates conversation and message IDs with `Date.now()`:

```ts
// frontend/src/store/chatStore.ts
const id = `conv_${Date.now()}`   // createConversation
id: `msg_${Date.now()}`           // addMessage
```

Two `createConversation()` calls within the same millisecond produce the same ID. This is not theoretical: under full-suite load, `chatStore.test.ts > deleteConversation > preserves activeConversationId when deleting different conversation` intermittently fails on

```
AssertionError: expected 'conv_1784092673972' not to be 'conv_1784092673972'
```

because the test creates two conversations back-to-back and they collide. The same file passes 21/21 in isolation — it's a timing flake, observed during PR #156 gate runs (the PR does not touch this file; the behavior reproduces on `development`).

### Impact

- Flaky CI/local test runs (intermittent, machine-speed-dependent).
- Runtime: a user clicking "New Chat" twice in the same millisecond would get two conversations sharing an ID (store operations keyed by ID would then affect both). Unlikely by hand, plausible programmatically.

### Suggested fix

Use `crypto.randomUUID()` (available in all supported browsers and jsdom) or append a monotonic counter: `conv_${Date.now()}_${counter++}`. Same for `msg_` IDs.

mundurragacl:
## Summary

Fixes #160 (close manually on merge).

`chatStore` generated conversation and message IDs from `Date.now()`, so two creations within the same millisecond shared an ID — a real store-corruption path (operations keyed by ID affect both) and the source of the intermittent `deleteConversation` test failure observed during #156's gate runs.

## Change

- `conv_${Date.now()}` → `conv_${crypto.randomUUID()}` and `msg_${Date.now()}` → `msg_${crypto.randomUUID()}`. `crypto.randomUUID()` is available in all supported browsers (secure contexts — CloudFront https and localhost dev) and in the vitest runtime. Prefixes are kept for debuggability; IDs are opaque strings everywhere (verified: nothing parses them), so persisted conversations from before this change stay valid.
- Removed the test suite's 1ms `setTimeout` timestamp-separation workaround — the flake's band-aid — and made that test synchronous.

## Tests

Two regression pins that fail on the old scheme with near-certainty: 50 back-to-back `createConversation()` calls must yield 50 distinct IDs, and 50 same-tick `addMessage` calls must yield 50 distinct message IDs. Full store suite: 23 passed; frontend lint/typecheck clean.

### DIFF
--- a/voc-datalake/frontend/src/store/chatStore.test.ts
+++ b/voc-datalake/frontend/src/store/chatStore.test.ts
@@ -25,6 +25,14 @@ describe('chatStore', () => {
       expect(state.conversations[0].id).toBe(id)
     })
 
+    it('generates unique IDs for back-to-back creations (issue #160)', () => {
+      // Date.now()-based IDs collided within the same millisecond — two
+      // conversations then shared an ID and store operations affected both.
+      const ids = Array.from({ length: 50 }, () => useChatStore.getState().createConversation())
+
+      expect(new Set(ids).size).toBe(50)
+    })
+
     it('sets new conversation as active', () => {
       const { createConversation } = useChatStore.getState()
       
@@ -77,17 +85,15 @@ describe('chatStore', () => {
       expect(state.activeConversationId).toBeNull()
     })
 
-    it('preserves activeConversationId when deleting different conversation', async () => {
+    it('preserves activeConversationId when deleting different conversation', () => {
       // Reset state completely first
       useChatStore.setState({ conversations: [], activeConversationId: null })
       
       // Create first conversation
       const firstId = useChatStore.getState().createConversation()
-      
-      // Wait a tick to ensure different timestamp for second ID
-      await new Promise(resolve => setTimeout(resolve, 1))
-      
-      // Create second conversation
+      // Back-to-back creation is safe: IDs are collision-proof (issue #160),
+      // so no timestamp-separation sleep is needed (the old 1ms wait still
+      // flaked under full-suite load).
       const secondId = useChatStore.getState().createConversation()
       
       // Verify we have two different conversations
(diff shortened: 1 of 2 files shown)


perrozzi:
This PR replaces `Date.now()`-based conversation and message ID generation in `chatStore` with `crypto.randomUUID()`, eliminating same-millisecond ID collisions that could corrupt the store, and removes the 1ms-sleep flake workaround from the test suite. The fix is small, well-scoped, and comes with two regression tests that would fail if reverted (matching the repo's bug-fix policy). Overall this looks good to merge.

### Nits

- `voc-datalake/frontend/src/store/chatStore.test.ts`: the new "generates unique IDs for back-to-back creations" test doesn't reset store state first, unlike the neighboring `deleteConversation` test. It's harmless for the uniqueness assertion, but resetting via `useChatStore.setState({ conversations: [], activeConversationId: null })` (or a shared `beforeEach`) would keep tests order-independent and consistent with the rest of the file.
- `voc-datalake/frontend/src/store/chatStore.ts`: `crypto.randomUUID()` is only available in secure contexts. CloudFront (https) and `localhost` dev are covered as the PR notes, but a dev server accessed via a LAN IP over plain http would throw. Probably not a supported workflow here, but worth a one-line note in the comment or docs if anyone dev-tests that way.
- The two inline comments in `chatStore.ts` repeat the issue rationale; the second one ("Collision-proof per issue #160…") could be trimmed since the first already explains it.

No blockers or should-fix items — the change is correct, IDs are treated as opaque strings (prefixes preserved for debuggability, persisted data unaffected), and the removed `setTimeout` workaround is properly obsoleted by the fix rather than papered over.

---

mundurragacl:
Round 1 addressed:

- **Secure-context nit**: taken — the comment now notes randomUUID's https/localhost requirement and that plain-http LAN-IP dev isn't a supported workflow.
- **Duplicate comment nit**: taken — the message-ID comment now just points at the conversation-ID rationale.
- **Store-reset nit**: checked — the file already has a shared `beforeEach` on the top-level `describe('chatStore')` that resets `conversations`/`activeConversationId` before EVERY test, including the new ones (chatStore.test.ts lines 8–14). The new tests are order-independent through that same mechanism as their neighbors, so no per-test reset was added.

23 store tests green.

perrozzi:
Since the last review, the author pushed comment-only changes to `chatStore.ts` addressing two of the three nits: a note about `crypto.randomUUID()`'s secure-context requirement was added, and the redundant second rationale comment was trimmed to a cross-reference. No functional code changed.

### Resolved
- `voc-datalake/frontend/src/store/chatStore.ts`: secure-context caveat (LAN-IP plain-http dev unsupported) now documented inline.
- `voc-datalake/frontend/src/store/chatStore.ts`: duplicated `msg_` ID rationale trimmed to "see createConversation".

There were no previous Blockers or Should-fix items. The remaining nit (state reset in the new uniqueness test in `chatStore.test.ts` for order-independence) was not addressed, but it was a nit and remains non-blocking.

### New findings
None — the incremental changes are comment-only and look clean. Good to merge.

---

PR #165, opened 2026-07-15
fix: chat filter changes no longer no-op without an active conversation

### LINKED ISSUE #161: [Bug]: Chat filter changes (incl. web-search toggle) silently no-op when no conversation is active
### What happened

On the AI Chat page, filter changes are silently dropped when no conversation is active:

```ts
// frontend/src/pages/Chat/Chat.tsx
const handleFiltersChange = (newFilters: ChatFiltersType) => {
  if (activeConversationId != null && activeConversationId !== '') {
    updateConversationFilters(activeConversationId, newFilters)
  }
}
```

Filters live on the conversation (`activeConversation?.filters ?? {}`), so on a fresh page with no conversation, clicking any filter control — source/category/sentiment dropdowns and the web-search toggle from #156 — visibly does nothing: the control renders its change, React re-renders from the store, and the control snaps back to its previous state. No feedback to the user.

Observed during PR #156 live validation: the web-search toggle appeared broken until "New Chat" was clicked, after which everything worked. The behavior predates #156 and affects all filters equally.

### Impact

- Confusing first-run UX: the most likely first action on the Chat page (set a filter, then ask) silently discards the filter.
- With the web-search toggle this now also has a billing dimension: a user may believe web search is enabled when it is not (fail-safe direction, but still misleading).

### Suggested fix options

1. Auto-create a conversation on first filter interaction (matches the existing "first message auto-creates" pattern), or
2. Buffer filter state locally until a conversation exists and apply it on creation, or
3. Disable the filter bar with a hint ("start a conversation to set filters") when no conversation is active.

Option 2 is probably the least surprising; option 3 is the smallest change.

mundurragacl:
## Summary

Fixes #161 (close manually on merge).

On a fresh Chat page (no active conversation), every filter interaction — the source/category/sentiment dropdowns and the web-search toggle from #156 — was silently dropped: `handleFiltersChange` only wrote to the active conversation, so the control snapped back with no feedback. With the web-search toggle this had a billing-adjacent twist: a user could believe web search was on when it wasn't (fail-safe direction, but misleading).

## Change — option 2 from the issue (buffer locally), implemented at the store level

Filter changes made while no conversation is active buffer in a new `draftFilters` field on `chatStore`; **the next conversation to be created consumes the draft** (and clears it). Putting the draft in the store rather than component state means BOTH creation paths inherit it:

- first message on the Chat page (`handleSubmit` → `createConversation`), and
- the sidebar's **New Chat** button — the exact repro in the report ("the toggle appeared broken until New Chat was clicked"), which a component-local draft would still have lost.

The draft is deliberately **not persisted** (excluded from partialize): it's ephemeral pre-conversation state, gone on reload like any half-configured UI. `createConversation`'s signature is unchanged, so `ChatSidebar` needed no edits.

## Tests

- Store: draft buffering, consumption + clearing on creation, and the empty-draft default.
- Chat page regression test: with no active conversation, a filter change lands in the draft (`setDraftFilters` called, `updateConversationFilters` not) instead of being discarded.
- Full root gate green (`npm run check` exit 0; 44 tests across the two touched suites).

## Note

Sibling PR #164 (issue #160) touches the same two files (`chatStore.ts`/test) on independent lines — whichever merges second may see a trivial conflict.

### DIFF
--- a/voc-datalake/frontend/src/pages/Chat/Chat.test.tsx
+++ b/voc-datalake/frontend/src/pages/Chat/Chat.test.tsx
@@ -71,6 +71,7 @@ const mockCreateConversation = vi.fn()
 const mockAddMessage = vi.fn()
 const mockGetActiveConversation = vi.fn()
 const mockUpdateConversationFilters = vi.fn()
+const mockSetDraftFilters = vi.fn()
 
 vi.mock('../../store/chatStore', () => ({
   useChatStore: () => ({
@@ -79,6 +80,8 @@ vi.mock('../../store/chatStore', () => ({
     addMessage: mockAddMessage,
     getActiveConversation: mockGetActiveConversation,
     updateConversationFilters: mockUpdateConversationFilters,
+    draftFilters: {},
+    setDraftFilters: mockSetDraftFilters,
   }),
   // Export types for the component
 }))
@@ -383,6 +386,21 @@ describe('Chat', () => {
       // sendMessage was called
       // eslint-disable-next-line vitest/prefer-called-with
       expect(mockSendMessage).toHaveBeenCalled()
+    })
+  })
+
+  describe('filter changes without an active conversation (issue #161)', () => {
+    it('buffers the change as a draft instead of silently dropping it', async () => {
+      // Regression: with no active conversation, filter changes (source/
+      // category/sentiment dropdowns and the web-search toggle) used to
+      // no-op — the control snapped back and the user got no feedback.
+      const user = userEvent.setup()
+      render(<Chat />, { wrapper: createWrapper() })
+
+      await user.click(screen.getByRole('button', { name: 'Set Source Filter' }))
+
+      expect(mockSetDraftFilters).toHaveBeenCalledWith({ source: 'webscraper' })
+      expect(mockUpdateConversationFilters).not.toHaveBeenCalled()
     })
   })
 })
--- a/voc-datalake/frontend/src/pages/Chat/Chat.tsx
+++ b/voc-datalake/frontend/src/pages/Chat/Chat.tsx
@@ -212,6 +212,9 @@ export default function Chat() {
   } = useChatStore()
 
   const activeConversation = getActiveConversation()
+  // Treat null/undefined and '' uniformly everywhere (an empty-string id
+  // buffering a draft that ?? would never consume was a reviewer catch).
+  const hasActiveConversation = activeConversationId != null && activeConversationId !== ''
   // Filters live on the conversation; before one exists they buffer in the
   // store's draft (issue #161: filter changes — including the web-search
   // toggle — used to silently no-op on a fresh page and snap back). The
@@ -291,7 +294,7 @@ export default function Chat() {
   }, [isStreaming])
 
   const handleFiltersChange = (newFilters: ChatFiltersType) => {
-    if (activeConversationId != null && activeConversationId !== '') {
+    if (hasActiveConversation) {
       updateConversationFilters(activeConversationId, newFilters)
     } else {
       setDraftFilters(newFilters)
@@ -309,9 +312,9 @@ export default function Chat() {
       content: m.content,
     }))
 
-    // The first message materializes the conversation, which consumes the
-    // draft filters the user set beforehand (issue #161).
-    const conversationId = activeConversationId ?? createConversation()
+    // The first message materializes the conversation, which consumes any
+    // draft filters the user set beforehand.
+    const conversationId = hasActiveConversation ? activeConversationId : createConversation()
     addMessage(conversationId, {
       role: 'user',
       content: input,
--- a/voc-datalake/frontend/src/pages/Chat/Chat.tsx
+++ b/voc-datalake/frontend/src/pages/Chat/Chat.tsx
@@ -207,10 +207,16 @@ export default function Chat() {
     addMessage,
     getActiveConversation,
     updateConversationFilters,
+    draftFilters,
+    setDraftFilters,
   } = useChatStore()
 
   const activeConversation = getActiveConversation()
-  const filters: ChatFiltersType = activeConversation?.filters ?? {}
+  // Filters live on the conversation; before one exists they buffer in the
+  // store's draft (issue #161: filter changes — including the web-search
+  // toggle — used to silently no-op on a fresh page and snap back). The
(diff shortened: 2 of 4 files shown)


perrozzi:
This PR fixes a real UX bug: filter changes (including the web-search toggle) made on a fresh Chat page with no active conversation were silently dropped. The fix buffers them in a new `draftFilters` field on `chatStore`, which is consumed by the next `createConversation()` call — correctly covering both creation paths (first message and the sidebar's New Chat) without changing the `createConversation` signature. The change follows the repo's Zustand patterns, includes a regression test at the component level plus store-level tests, and adds no user-visible strings (so no i18n work needed). Overall this looks solid; two things to verify below.

### Should fix

- **Persistence exclusion is claimed but not visible in the diff.** The PR description says `draftFilters` is "excluded from partialize," but `voc-datalake/frontend/src/store/chatStore.ts` shows no change to the `persist` config. If the store currently persists without a `partialize` (or with one that doesn't already whitelist only `conversations`/`activeConversationId`), the draft will be written to storage and a stale draft could silently apply to a conversation created in a later session — the same "surprising filter state" class of bug this PR fixes, but in reverse. Please confirm the persist config, or add an explicit `partialize` exclusion, and ideally a test asserting `draftFilters` is not in the persisted payload.

- **Shared object reference between the draft and the new conversation.** `filters: get().draftFilters` in `createConversation` hands the draft object itself to the conversation. Today `setDraftFilters` and `updateConversationFilters` replace objects wholesale, so this is safe, but any future code that mutates a filters object in place would create spooky cross-state coupling. Cheap insurance: `filters: { ...get().draftFilters }`.

### Nits

- `Chat.tsx`: the `handleFiltersChange` branch condition `activeConversationId != null && activeConversationId !== ''` duplicates logic that appears in `handleSubmit` (`activeConversationId ?? createConversation()`, which treats `''` differently). Pre-existing inconsistency, but worth unifying while you're here — an empty-string id would buffer a draft yet never consume it (`??` won't fire for `''`).
- `chatStore.test.ts`: consider one more case — setting a draft *while* a conversation is active shouldn't be possible via the UI, but a store-level test pinning that `createConversation` after `deleteConversation` still consumes a lingering draft (or that the draft is cleared) would document the intended lifecycle.
- The inline comments referencing "issue #161" in four places are a bit heavy; one comment on the `draftFilters` field declaration would suffice.

Note: the review covers only the four files in the diff; the persist/`partialize` block of `chatStore.ts` was not included, hence the verification request above.

---

perrozzi:
The new push addresses all three review items: an explicit copy of the draft on consumption, a test pinning `draftFilters` out of the persisted payload, unified empty-string id handling via `hasActiveConversation`, plus the suggested delete-then-create lifecycle test and trimmed issue-number comments. Good turnaround.

### Resolved
- Shared object reference: `filters: { ...get().draftFilters }` in `chatStore.ts`, with a store test asserting the conversation gets its own copy.
- Persistence exclusion: new test asserts `draftFilters` is absent from `partialize(...)` output while `conversations`/`activeConversationId` are present.
- Empty-string id inconsistency: `handleSubmit` and `handleFiltersChange` now share `hasActiveConversation`, so a `''` id can no longer buffer a draft that `??` would never consume.
- Lifecycle test for a lingering draft after `deleteConversation` added.
- Redundant `#161` comments consolidated.

### New findings
- **Minor (verify only):** the persistence test in `chatStore.test.ts` uses `options.partialize?.(...) ?? useChatStore.getState()` — the fallback means the test only passes if a `partialize` actually exists and (as a whitelist) omits the new field. The persist config block of `chatStore.ts` still isn't visible in either diff, so this review takes the passing test as the evidence; if the existing `partialize` is a blocklist rather than a whitelist, double-check it wasn't quietly edited outside the diff.
- Nit: the comment in `Chat.tsx` ("was a reviewer catch") documents review history rather than behavior — the first sentence alone carries the intent.

Otherwise the incremental changes are clean and well-tested; no blockers.

---

mundurragacl:
Round 1 addressed in ea78bc0 — both should-fixes plus all nits:

- **Persistence exclusion**: your caution was warranted to demand evidence — the existing `partialize` DOES whitelist only `conversations`/`activeConversationId` (it predates this PR, hence invisible in the diff), so the draft never reached storage. It's now pinned by a test that runs the store state through `useChatStore.persist.getOptions().partialize` and asserts `draftFilters` is absent from the persisted payload — with a comment noting the stale-draft-in-a-later-session failure mode you described.
- **Shared reference**: taken — `filters: { ...get().draftFilters }`, with a test asserting the conversation's filters object is not the draft object.
- **`''` inconsistency (nit, but a good catch)**: unified — one `hasActiveConversation` condition (null/undefined and `''` treated the same) now drives both `handleFiltersChange` and `handleSubmit`, eliminating the buffer-but-never-consume window an empty-string id could have opened under `??`.
- **Lifecycle test**: added — a lingering draft set before deleting the active conversation is consumed by the next creation (and cleared).
- **Comment weight**: trimmed to one #161 reference on the `draftFilters` field declaration.

47 tests across the two touched suites; root `npm run check` exit 0.

mundurragacl:
Round 2 addressed:

- **Partialize verification**: confirmed it's a **whitelist** — the persist config's `partialize` returns exactly `{ conversations, activeConversationId }` (chatStore.ts, unchanged since #153-era code; that's why it never appeared in a diff). So the new field is excluded by construction, and the test's `?? getState()` fallback would correctly FAIL the assertion if someone ever deleted the partialize (full state would then contain `draftFilters`) — the fallback is the tripwire, not a pass-through.
- **Comment nit**: reworded to describe the behavior instead of the review history.

47 tests green. Nothing else outstanding from my side.

perrozzi:
The only change since the last review is a two-line comment reword in `Chat.tsx`, replacing the review-history phrasing ("was a reviewer catch") with a behavioral explanation of why empty-string ids must be normalized. This addresses the remaining nit from my previous review; all substantive findings were already resolved in the prior push.

### Resolved
- Nit: the `Chat.tsx` comment now documents behavior (empty-string id must not buffer an unconsumable draft) rather than review history.

The incremental change is clean — no new findings, no blockers. (The earlier verify-only note about the `partialize` whitelist-vs-blocklist still stands as unverifiable from the diffs, with the passing persistence test as the evidence.)

---

PR #166, opened 2026-07-15
feat: per-surface AI model picker with latest Claude models (alternative to #154)

perrozzi:
## Summary

Addresses #96 (close manually on merge — base is `development`). **Alternative to #154** — offered for maintainer choice; #154 stays open.

Admins get an **AI Models** card in Settings (brand tab, admin-only) with one selector **per AI surface** instead of #154's single global toggle:

| Surface | What it covers | Automatic default |
|---|---|---|
| AI Chat | chat, streaming chat, project/product interview | **Claude Sonnet 5** |
| Document Generation | PRDs, PR/FAQs, personas, research, merger, product reports | **Claude Sonnet 5** |
| Prototype Builder | interactive HTML prototypes | **Claude Opus 4.8** |
| Feedback Enrichment | processor (sentiment/category/urgency on every item) | **Claude Haiku 4.5** |
| Utilities | category suggestions, scraper selector detection | **Claude Sonnet 5** |

### Why per-surface instead of one global switch

With a global toggle, choosing Haiku downgrades exactly the wrong things: the volume driver (enrichment) is already on Haiku and pinned there, so the toggle only moves the low-volume, quality-sensitive surfaces. Per-surface lets a cost-conscious admin put chat on Haiku while keeping PRDs on Sonnet — the trade an admin actually wants to make. It also turns the previously **hardcoded** Opus prototype model into a picker-controlled surface.

### Model bumps (all IDs verified against the Bedrock model cards)

- Default model **Sonnet 4.5 → Sonnet 5** (`global.anthropic.claude-sonnet-5`)
- Allowlist: **Sonnet 5, Sonnet 4.6, Opus 4.8, Haiku 4.5** (latest Haiku; there is no Haiku 5)
- Capability-aware invocation: Sonnet 5 and Opus 4.8 reject `temperature`, and Sonnet 5 rejects an explicit thinking budget (adaptive thinking is always-on) — `converse()` and the streaming client drop those fields automatically per resolved model, so **any** surface can point at **any** allowlisted model without a 400. The two raw `client.converse()` sites (persona import, product interview) get the same treatment.

### Resolution & storage

`shared/model_config.py`: *explicit arg > per-surface override > legacy global `model_id` (kept for backward compat with #154's shape) > surface default > `BEDROCK_MODEL_ID`*. 60s per-container cache; the lookup never raises. One item in aggregates: `SETTINGS#model` `{surfaces: {surface: id}, model_id?}`. TS mirror (`model-override.ts`) resolves the `chat` surface for streaming.

`GET/PUT /settings/model` on the existing settings Lambda — GET open to all authenticated users, **PUT requires the `admins` Cognito group server-side** (`require_admin` promoted to `shared/api.py`). Non-allowlisted values are 400 on write and read back as Automatic, so a tampered DynamoDB row can't steer inference.

### IAM lockstep (structural, not aspirational)

New `lib/utils/model-allowlist.ts` is the single source for **all** `bedrock:InvokeModel*` grants (api / processing / ingestion stacks), the **BedrockAccessStack agreements** (now 4 models, previously missing Opus), and the cdk-nag suppressions. The research Lambda gains `AGGREGATES_TABLE` env + read so the documents override reaches Step Functions research. Python lockstep tests read the TS mirror, the CDK helper, and the nag suppressions — adding a model to one place fails the build until all agree.

## Testing

- 25 resolver tests (per-surface precedence, legacy-global fallback, tamper rejection incl. a stale Sonnet 4.5 pin, never-raises, cache TTL, lockstep ×4)
- 8 converse surface/capability tests (surface routing, explicit-arg precedence, chain threading, auto-omit temperature/thinking)
- 12 settings-handler tests (per-surface set/clear preserving siblings, legacy path, 5-case rejection incl. the internal `default` bucket, non-admin 403, missing-groups 403, non-admin GET 200)
- 12 stream override tests + 4 converse-stream tests (never-throws, cache, precedence, adaptive-thinking omission on both the explicit and env-default paths)
- 8 UI tests (per-surface render, Automatic default labels, select/clear round-trip, error state, hidden for non-admins/no endpoint)
- Full suites green: backend **1068 passed / 1 pre-existing skip**, frontend **1629**, stream **230**, CDK **33**; `tsc` clean ×3; `eslint` clean; `i18n:validate` exit 0 (8 locales)

## Notes

- Override propagation is cache-bounded (~60s), same as #154.
- The processor's enrichment default stays Haiku — no cost regression; the surface is simply overridable now.
- A `model_id` written by #154's single-model picker keeps working as the global fallback for un-pinned surfaces.
- Found while validating (pre-existing, unrelated): `Code.fromAsset('lambda')` in the processing stack doesn't exclude `stream/node_modules`, so CDK tests time out on any branch once stream deps are installed — candidate for a follow-up exclude fix.

## Deployment notes

- The processor's `BEDROCK_MODEL_ID` environment variable is removed from `processing-stack-consolidated.ts`. It had become inert (the enrichment surface resolves through the picker; the default stays Haiku 4.5). If you had customized the enrichment model by editing that env literal in the stack source, set it via the Settings → AI Models picker (enrichment surface) instead.

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -125,6 +125,23 @@ const mockSourcesStatus = {
 // Problem resolution state (issue #66) — mutable so the toggle round-trips.
 const mockResolvedProblems = {};
 
+// Per-surface AI model picker state (issue #96) — mutable so selects round-trip.
+// Mirrors lambda/shared/model_config.py (source of truth, lockstep-tested there).
+const mockModelSurfaces = {};
+const mockAvailableModels = [
+  { key: 'sonnet5', id: 'global.anthropic.claude-sonnet-5', label: 'Claude Sonnet 5', description: 'Latest, highest-quality Sonnet — best for analysis and generation' },
+  { key: 'sonnet46', id: 'global.anthropic.claude-sonnet-4-6', label: 'Claude Sonnet 4.6', description: 'Previous-generation Sonnet — strong quality, accepts temperature tuning' },
+  { key: 'opus48', id: 'global.anthropic.claude-opus-4-8', label: 'Claude Opus 4.8', description: 'Deepest reasoning — best for prototypes and complex documents' },
+  { key: 'haiku45', id: 'global.anthropic.claude-haiku-4-5-20251001-v1:0', label: 'Claude Haiku 4.5', description: 'Fastest and cheapest — good for high-volume enrichment' },
+];
+const mockSurfaceDefaults = {
+  chat: 'global.anthropic.claude-sonnet-5',
+  documents: 'global.anthropic.claude-sonnet-5',
+  prototype: 'global.anthropic.claude-opus-4-8',
+  enrichment: 'global.anthropic.claude-haiku-4-5-20251001-v1:0',
+  utility: 'global.anthropic.claude-sonnet-5',
+};
+
 const mockCategoriesConfig = {
   categories: [
     { name: 'delivery', display_name: 'Delivery', description: 'Shipping and delivery issues', color: '#EF4444' },
@@ -188,6 +205,34 @@ const handlers = {
   // Settings - Categories
   'GET /settings/categories': () => mockCategoriesConfig,
   'PUT /settings/categories': () => ({ success: true, message: 'Categories saved' }),
+
+  // Settings - Per-surface AI model picker (issue #96)
+  'GET /settings/model': () => ({
+    available_models: mockAvailableModels,
+    surfaces: Object.entries(mockSurfaceDefaults).map(([key, defaultId]) => ({
+      key,
+      default_id: defaultId,
+      selected: mockModelSurfaces[key] ?? null,
+    })),
+    model_id: null,
+  }),
+  'PUT /settings/model': (body) => {
+    const surface = body?.surface;
+    const modelId = body?.model_id ?? null;
+    if (!surface || !(surface in mockSurfaceDefaults)) {
+      return { __status: 400, body: { success: false, error: 'surface must be a known picker surface' } };
+    }
+    if (modelId !== null && !mockAvailableModels.some((m) => m.id === modelId)) {
+      return { __status: 400, body: { success: false, error: 'model_id must be null or allowlisted' } };
+    }
+    if (modelId === null) {
+      delete mockModelSurfaces[surface];
+    } else {
+      mockModelSurfaces[surface] = modelId;
+    }
+    return { success: true, surface, model_id: modelId };
+  },
+
   'GET /settings/resolved-problems': () => ({ resolved: mockResolvedProblems }),
   'PUT /settings/resolved-problems': (body) => {
     if (!body || typeof body.key !== 'string' || body.key.trim() === '' || typeof body.resolved !== 'boolean') {
--- a/voc-datalake/frontend/public/locales/de/settings.json
+++ b/voc-datalake/frontend/public/locales/de/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "Automatisch — {{model}}",
+    "intro": "Wählen Sie, welches Claude-Modell die einzelnen KI-Funktionen antreibt, oder behalten Sie „Automatisch“ bei, um die abgestimmte Voreinstellung jeder Funktion zu nutzen. Laufende Funktionen übernehmen Änderungen innerhalb einer Minute.",
+    "loadFailed": "Modelleinstellungen konnten nicht geladen werden.",
+    "loading": "Modelle werden geladen...",
+    "saveFailed": "Modellauswahl konnte nicht gespeichert werden. Bitte erneut versuchen.",
+    "saved": "Gespeichert",
+    "surfaces": {
+      "chat": {
+        "description": "Konversationelle Datenabfragen und Projekt-Chat",
+        "label": "KI-Chat"
+      },
+      "documents": {
+        "description": "PRDs, PR/FAQs, Personas und Forschungsberichte",
+        "label": "Dokumenterstellung"
+      },
+      "enrichment": {
+        "description": "Stimmung, Kategorien und Dringlichkeit für jedes erfasste Element",
+        "label": "Feedback-Anreicherung"
+      },
+      "prototype": {
+        "description": "Interaktive HTML-Prototypen",
+        "label": "Prototyp-Builder"
+      },
+      "utility": {
+        "description": "Kategorievorschläge und Scraper-Selektor-Erkennung",
+        "label": "Dienstprogramme"
+      }
+    },
+    "title": "KI-Modelle"
+  },
   "api": {
     "connected": "Verbunden",
     "endpointHint": "Der API-Gateway-Endpunkt Ihrer VoC-Bereitstellung",
--- a/voc-datalake/frontend/public/locales/en/settings.json
+++ b/voc-datalake/frontend/public/locales/en/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "Automatic — {{model}}",
+    "intro": "Choose which Claude model powers each AI feature, or keep Automatic to use each feature’s tuned default. Running functions pick up changes within a minute.",
+    "loadFailed": "Could not load model settings.",
+    "loading": "Loading models...",
+    "saveFailed": "Failed to save model selection. Try again.",
+    "saved": "Saved",
+    "surfaces": {
+      "chat": {
+        "description": "Conversational data queries and project chat",
+        "label": "AI Chat"
+      },
+      "documents": {
+        "description": "PRDs, PR/FAQs, personas, and research reports",
+        "label": "Document Generation"
+      },
+      "enrichment": {
+        "description": "Sentiment, categories, and urgency on every ingested item",
+        "label": "Feedback Enrichment"
+      },
+      "prototype": {
+        "description": "Interactive HTML prototypes",
+        "label": "Prototype Builder"
+      },
+      "utility": {
+        "description": "Category suggestions and scraper selector detection",
+        "label": "Utilities"
+      }
+    },
+    "title": "AI Models"
+  },
   "api": {
     "connected": "Connected",
     "endpointHint": "The API Gateway endpoint from your VoC deployment",
--- a/voc-datalake/frontend/public/locales/es/settings.json
+++ b/voc-datalake/frontend/public/locales/es/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "Automático — {{model}}",
+    "intro": "Elija qué modelo de Claude impulsa cada función de IA, o mantenga Automático para usar el valor predeterminado ajustado de cada función. Las funciones en ejecución aplican los cambios en un minuto.",
+    "loadFailed": "No se pudo cargar la configuración de modelos.",
+    "loading": "Cargando modelos...",
+    "saveFailed": "No se pudo guardar la selección de modelo. Inténtelo de nuevo.",
+    "saved": "Guardado",
+    "surfaces": {
+      "chat": {
+        "description": "Consultas de datos conversacionales y chat de proyecto",
+        "label": "Chat de IA"
+      },
+      "documents": {
+        "description": "PRD, PR/FAQ, personas e informes de investigación",
+        "label": "Generación de documentos"
+      },
+      "enrichment": {
+        "description": "Sentimiento, categorías y urgencia en cada elemento ingerido",
+        "label": "Enriquecimiento de comentarios"
+      },
+      "prototype": {
+        "description": "Prototipos HTML interactivos",
+        "label": "Generador de prototipos"
+      },
+      "utility": {
+        "description": "Sugerencias de categorías y detección de selectores del scraper",
+        "label": "Utilidades"
+      }
+    },
+    "title": "Modelos de IA"
+  },
   "api": {
     "connected": "Conectado",
     "endpointHint": "El endpoint de API Gateway de su despliegue VoC",
--- a/voc-datalake/frontend/public/locales/fr/settings.json
+++ b/voc-datalake/frontend/public/locales/fr/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "Automatique — {{model}}",
+    "intro": "Choisissez le modèle Claude qui alimente chaque fonctionnalité d’IA, ou conservez Automatique pour utiliser le réglage par défaut de chaque fonctionnalité. Les fonctions en cours prennent en compte les changements en moins d’une minute.",
+    "loadFailed": "Impossible de charger les paramètres de modèle.",
+    "loading": "Chargement des modèles...",
+    "saveFailed": "Échec de l’enregistrement de la sélection du modèle. Réessayez.",
+    "saved": "Enregistré",
+    "surfaces": {
+      "chat": {
+        "description": "Requêtes de données conversationnelles et chat de projet",
+        "label": "Chat IA"
+      },
+      "documents": {
+        "description": "PRD, PR/FAQ, personas et rapports de recherche",
+        "label": "Génération de documents"
+      },
+      "enrichment": {
+        "description": "Sentiment, catégories et urgence sur chaque élément ingéré",
+        "label": "Enrichissement des retours"
+      },
+      "prototype": {
+        "description": "Prototypes HTML interactifs",
+        "label": "Générateur de prototypes"
+      },
+      "utility": {
+        "description": "Suggestions de catégories et détection des sélecteurs du scraper",
+        "label": "Utilitaires"
+      }
+    },
+    "title": "Modèles d’IA"
+  },
   "api": {
     "connected": "Connecté",
     "endpointHint": "Le point de terminaison API Gateway de votre déploiement VoC",
--- a/voc-datalake/frontend/public/locales/ja/settings.json
+++ b/voc-datalake/frontend/public/locales/ja/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "自動 — {{model}}",
+    "intro": "各AI機能を動かすClaudeモデルを選択するか、「自動」のままにして各機能に最適化されたデフォルトを使用します。実行中の関数は1分以内に変更を反映します。",
+    "loadFailed": "モデル設定を読み込めませんでした。",
+    "loading": "モデルを読み込んでいます...",
+    "saveFailed": "モデル選択の保存に失敗しました。もう一度お試しください。",
+    "saved": "保存済み",
+    "surfaces": {
+      "chat": {
+        "description": "対話型データクエリとプロジェクトチャット",
+        "label": "AIチャット"
+      },
+      "documents": {
+        "description": "PRD、PR/FAQ、ペルソナ、調査レポート",
+        "label": "ドキュメント生成"
+      },
+      "enrichment": {
+        "description": "取り込んだ全項目のセンチメント・カテゴリ・緊急度",
+        "label": "フィードバック強化"
+      },
+      "prototype": {
+        "description": "インタラクティブなHTMLプロトタイプ",
+        "label": "プロトタイプビルダー"
+      },
+      "utility": {
+        "description": "カテゴリ提案とスクレイパーセレクター検出",
+        "label": "ユーティリティ"
+      }
+    },
+    "title": "AIモデル"
+  },
   "api": {
     "connected": "接続済み",
     "endpointHint": "VoCデプロイメントのAPI Gatewayエンドポイント",
--- a/voc-datalake/frontend/public/locales/ko/settings.json
+++ b/voc-datalake/frontend/public/locales/ko/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "자동 — {{model}}",
+    "intro": "각 AI 기능을 구동할 Claude 모델을 선택하거나, 자동으로 두어 각 기능에 맞게 조정된 기본값을 사용하세요. 실행 중인 함수는 1분 이내에 변경 사항을 반영합니다.",
+    "loadFailed": "모델 설정을 불러오지 못했습니다.",
+    "loading": "모델을 불러오는 중...",
+    "saveFailed": "모델 선택을 저장하지 못했습니다. 다시 시도하세요.",
+    "saved": "저장됨",
+    "surfaces": {
+      "chat": {
+        "description": "대화형 데이터 질의 및 프로젝트 채팅",
+        "label": "AI 채팅"
+      },
+      "documents": {
+        "description": "PRD, PR/FAQ, 페르소나, 리서치 보고서",
+        "label": "문서 생성"
+      },
+      "enrichment": {
+        "description": "수집된 모든 항목의 감성, 카테고리, 긴급도",
+        "label": "피드백 보강"
+      },
+      "prototype": {
+        "description": "인터랙티브 HTML 프로토타입",
+        "label": "프로토타입 빌더"
+      },
+      "utility": {
+        "description": "카테고리 제안 및 스크레이퍼 선택자 감지",
+        "label": "유틸리티"
+      }
+    },
+    "title": "AI 모델"
+  },
   "api": {
     "connected": "연결됨",
     "endpointHint": "VoC 배포의 API Gateway 엔드포인트",
--- a/voc-datalake/frontend/public/locales/pt/settings.json
+++ b/voc-datalake/frontend/public/locales/pt/settings.json
@@ -1,4 +1,35 @@
 {
+  "aiModel": {
+    "automaticOption": "Automático — {{model}}",
+    "intro": "Escolha qual modelo Claude alimenta cada recurso de IA, ou mantenha Automático para usar o padrão ajustado de cada recurso. Funções em execução aplicam as alterações em até um minuto.",
+    "loadFailed": "Não foi possível carregar as configurações de modelo.",
+    "loading": "Carregando modelos...",
+    "saveFailed": "Falha ao salvar a seleção do modelo. Tente novamente.",
+    "saved": "Salvo",
+    "surfaces": {
+      "chat": {
+        "description": "Consultas de dados conversacionais e chat do projeto",
+        "label": "Chat de IA"
+      },
+      "documents": {
+        "description": "PRDs, PR/FAQs, personas e relatórios de pesquisa",
+        "label": "Geração de documentos"
+      },
+      "enrichment": {
+        "description": "Sentimento, categorias e urgência em cada item ingerido",
+        "label": "Enriquecimento de feedback"
+      },
+      "prototype": {
+        "description": "Protótipos HTML interativos",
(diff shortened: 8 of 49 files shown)


perrozzi:
This PR replaces the single hardcoded Bedrock model with a per-surface picker: a new `shared/model_config.py` resolver (per-surface override > legacy global > surface default > `BEDROCK_MODEL_ID`), an admin-gated `GET/PUT /settings/model` on the settings Lambda, a TS mirror for the streaming chat path, a shared CDK allowlist helper feeding all `bedrock:InvokeModel*` grants and the BedrockAccessStack agreements, an admin-only Settings card with all 8 locales, and lockstep tests tying Python/TS/CDK together. The architecture is well thought out (never-raise resolution, allowlist enforcement on both read and write, server-side admin gate, IAM lockstep tests), but the model-ID bump itself and override plumbing coverage need verification before merge.

### Blockers

- **Model ID validity is unverifiable from this diff and the format is suspicious.** Every previously deployed Sonnet ID in this repo carries a dated version suffix (`global.anthropic.claude-sonnet-4-5-20250929-v1:0`), but the new default `global.anthropic.claude-sonnet-5` (and `claude-sonnet-4-6`) has none, and `lib/utils/model-allowlist.ts` derives foundation-model ARNs by stripping `global.` — producing `foundation-model/anthropic.claude-sonnet-5` with no `-vX:Y` suffix. Bedrock foundation-model ARNs normally require the version suffix. Since `shared/aws.py::BEDROCK_MODEL_ID` is now this ID, a wrong string breaks **every** default inference path at once. Please attach evidence of a real `converse()` call against each of the four IDs in a deployed environment (the PR asserts "verified against model cards" but the lockstep tests only prove internal consistency, not that the IDs exist).
- **Override plumbing is incomplete for several surfaces — the picker will silently no-op.** `_load_settings()` in `lambda/shared/model_config.py` deliberately falls back to defaults when `AGGREGATES_TABLE` is unset or unreadable. The diff adds the env var + `grantReadData` only to the research Lambda (`processing-stack-consolidated.ts`). Every other Lambda now routing through surfaces — document generator/merger jobs, persona importer, scrapers (`utility`), manual import — needs the same env + aggregates read grant, or admin selections apply to some surfaces and silently not others, with no error anywhere. Please audit each surface's execution role/env in the diff or show existing grants; ideally add a CDK test asserting every Bedrock-invoking function has `AGGREGATES_TABLE` + aggregates read.

### Should fix

- `lambda/processor/handler.py`: the old `PROCESSOR_MODEL_ID = os.environ.get('BEDROCK_MODEL_ID', haiku)` honored an env override; the new resolution path ignores the processor's `BEDROCK_MODEL_ID` env entirely. If the processing stack (or any operator) sets that env var, it now does nothing silently. Either thread the env into the enrichment fallback or remove the env var from the stack in the same PR. Relatedly, `get_active_model_id('enrichment') or PROCESSOR_FALLBACK_MODEL_ID` — the resolver never returns falsy, so the fallback is dead code.
- PR description says `require_admin` was "promoted to `shared/api.py`", but the diff doesn't remove the presumed existing copy (users handler). Confirm the old implementation is deleted, and confirm `AuthorizationError` is actually imported in `shared/api.py` — the diff adds a function referencing it without showing the import.
- `lambda/stream/src/bedrock/model-override.ts` uses `as Record<string, unknown>` twice (on `result.Item` and `surfaces`), contrary to the repo's type-guard-over-`as` convention that the accompanying test file itself goes out of its way to follow.
- `settings_handler.py` read-modify-write `put_item` can lose a concurrent admin's surface pin. Acknowledged as rare in the comment, but a `ConditionExpression` on `updated_at` (or `update_item` with `if_not_exists`) would make it correct cheaply.

### Nits

- The lockstep tests parse TS sources with `split()`/regex (`test_model_config.py::TestAllowlistLockstep`); any comment mentioning `ALLOWED_MODEL_IDS` before the declaration would silently shift the parsed block. A JSON export or a stricter anchor (e.g., matching the `export const` line) would be sturdier.
- `AiModelSection.tsx`: the select is controlled by query data, so after a change it visually snaps back to the old value until the invalidated query refetches; consider optimistic cache update via `setQueryData` in `onSuccess`.
- `converse()` in `shared/converse.py` inserts `surface` before `max_retries` in the signature — safe for the keyword-only call sites here, but worth a quick grep for positional callers.
- Prototype surface repointed to Haiku will send `max_tokens=32000`; worth confirming all allowlisted models accept that output cap or clamping per model.
- Verify `datetime`/`timezone` are already imported in `settings_handler.py` (used by `_put_model_item` but not visible in the diff context).

---

perrozzi:
The changes since my last review are limited to `max_tokens` bumps on five strict-JSON `converse()` call sites (`projects.py` prfaq_autofill 2500→4096, document brief 1200→2048, research questions 1500→2048; `scrapers_handler.py` analyze_url 1000→2048; `settings_handler.py` generate_categories 2000→4096), plus explanatory comments about adaptive-thinking output budgets and an unreliable JSON continuation seam. None of the previous findings were addressed by this push.

### Still open
- Blocker: evidence that the four model IDs (`global.anthropic.claude-sonnet-5` etc., no version suffix) are real, invocable Bedrock IDs — the new "live-caught" comments *suggest* real Sonnet 5 runs happened, but attach the actual verification.
- Blocker: `AGGREGATES_TABLE` env + aggregates read grant audit for all Bedrock-invoking Lambdas beyond the research function (document generator/merger, persona importer, scrapers, manual import).
- Processor `BEDROCK_MODEL_ID` env override silently ignored; dead `or PROCESSOR_FALLBACK_MODEL_ID` fallback.
- `require_admin` promotion: confirm the old copy is removed and `AuthorizationError` is imported in `shared/api.py`.
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` PUT.

### New findings
- **No regression tests for the bug fix.** The comments describe a live-caught `JSONDecodeError` from a truncated JSON response, and the fix is these token bumps — but no test pins the new `max_tokens` values or exercises the truncation/continuation-seam failure mode. Per repo rules, a bug fix should ship with a regression test (e.g., asserting `converse()` is called with the bumped `max_tokens` for these endpoints, or a test on JSON-parse failure handling).
- **The bump mitigates rather than fixes the root cause.** If the auto-continuation seam is known-unreliable mid-JSON (`shared/converse.py` still allows continuation for adaptive-thinking models), a hard cap bump only lowers the probability: adaptive thinking spends a variable amount of the budget, and 2048 for `document_brief_suggest`/`research_suggest`/`analyze_url` (vs 4096 elsewhere) may still truncate on verbose thinking. Consider either disabling continuation for strict-JSON calls and surfacing a clean error on `max_tokens` stop, or normalizing all five sites to the same headroom with a rationale.
- Minor: the same justification comment is duplicated near-verbatim in four files; a single note in `shared/converse.py` (where the seam limitation lives) with short cross-references would be easier to keep accurate.

Otherwise the incremental changes are clean and low-risk.

---

perrozzi:
## Live deployment validation (voc-deploy, us-east-1)

Deployed all 4 app stacks (`--exclusively`, BedrockAccessStack excluded — internal-account guard) + frontend, then live-verified end to end.

**Regression sweep:** `scripts/test-api.sh` **32/32 passed** (incl. `POST /chat/stream` 200 on the new Sonnet 5 default).

**New endpoints (15 checks, all passing after one fix — see below):**
- `GET /settings/model`: 200, 4 models, 5 surfaces, correct defaults (prototype→Opus 4.8, enrichment→Haiku 4.5)
- `PUT` rejections: non-allowlisted model → 400, unknown surface → 400, **delisted Sonnet 4.5 → 400**
- Per-surface round-trip: pin chat→Haiku 200, GET reads it back, clear→null 200, GET reads Automatic
- AuthZ: non-admin PUT → **403**, non-admin GET → 200 (temp `users`-group user, deleted after)

**Real Bedrock invocations (the IAM-lockstep proof):**
- Chat pinned to Haiku, waited out the 60s resolver cache → `POST /chat` 200 with a real Haiku answer (override takes effect at runtime)
- Utility surface on the Sonnet 5 default → categories generate 200 ×2 (9 and 10 categories) — **Sonnet 5 inference-profile ID + new grants confirmed invocable**

**UI (CloudFront, admin login):** AI Models card renders 5 selectors, each "Automatic — {default}" + 4 models; pin chat→Haiku persists through the refetch; clear→Automatic persists. Non-admin never sees the card (server-side 403 additionally verified above).

### Live-caught fix (pushed as a follow-up commit)

Categories generate initially **500'd on Sonnet 5**: the response hit the `max_tokens=2000` ceiling (adaptive thinking consumes output budget), auto-continuation resumed, and the resume seam dropped a comma mid-JSON → `JSONDecodeError`. The Bedrock call itself succeeded — this is a token-budget issue, not IAM.

Fix: strict-JSON call sites must fit in **one** call — bumped `max_tokens` with headroom (categories 2000→4096, scraper selectors 1000→2048, prfaq_autofill 2500→4096, brief/research suggest 1200/1500→2048). Re-verified live: categories generate 200 twice on two different domains. Prose-emitting sites are unaffected (truncation there is graceful and continuation-safe).

**Env state:** voc-deploy now runs this PR's stacks + frontend (deployed with `-c enableWebSearch=true`, web-search wiring intact). A chat-surface override was exercised and **cleared** — all surfaces back on Automatic; temp test user deleted; no test artifacts left. Reconverge to `development` after the #154-vs-#166 decision.

perrozzi:
The push since my last review adds regression tests only: a `TestStrictJsonTokenHeadroom` class in `test_projects_ai_assists.py` pinning the bumped `max_tokens` floors for the three project assists, plus assertions in `test_scrapers_handler.py` and `test_settings_handler.py` covering the other two strict-JSON call sites (floor + `surface` kwarg). This directly addresses the main new finding from my previous review; all five bumped call sites now have a floor test that would fail if the headroom fix were reverted.

### Resolved
- No regression tests for the bug fix — all five strict-JSON `max_tokens` bumps are now pinned by floor assertions (`>=`, so future increases don't churn tests), and two also lock the `surface` routing.

### Still open
- Blocker: attach actual verification that the four unversioned model IDs (`global.anthropic.claude-sonnet-5` etc.) are real, invocable Bedrock IDs.
- Blocker: `AGGREGATES_TABLE` env + read-grant audit for all Bedrock-invoking Lambdas beyond research (document generator/merger, persona importer, scrapers, manual import).
- Root cause unaddressed: continuation is still allowed for adaptive-thinking models mid-JSON in `shared/converse.py`; the bumps lower probability, don't eliminate the failure mode (the new test docstrings themselves say continuation "is unreliable mid-JSON by design").
- Processor `BEDROCK_MODEL_ID` env override silently ignored; dead `or PROCESSOR_FALLBACK_MODEL_ID` fallback.
- Confirm the old `require_admin` copy is removed and `AuthorizationError` is imported in `shared/api.py`.
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` PUT.

### New findings
- Minor: the near-identical justification comment is now duplicated in five production files *and* three test files; consolidating it in `shared/converse.py` (where the seam limitation lives) with short cross-references would reduce drift risk.
- Nit (`test_scrapers_handler.py`, `test_settings_handler.py`): the regression assertions are appended to existing happy-path tests rather than standalone named tests — if the happy-path assertion fails first, the regression coverage is masked. Acceptable, but separate tests (as done in `test_projects_ai_assists.py`) would be cleaner.

Otherwise the incremental changes are clean — tests only, no production-code changes.

---

perrozzi:
The push since my last review addresses several open items: the processor's dead fallback and ignored env override are gone, the duplicated strict-JSON justification comments are consolidated into a single "doctrine" block in `shared/converse.py`, and the IAM/env audit is partially completed — `AGGREGATES_TABLE` env + read grants were added for the document generator, document merger, and persona importer. The manual import processor was also converted to capability-aware, picker-routed invocation with new regression tests.

### Resolved
- Processor `BEDROCK_MODEL_ID` env confusion — the env is removed from `processing-stack-consolidated.ts` and the dead `PROCESSOR_FALLBACK_MODEL_ID` literal is deleted; resolution goes solely through the picker.
- Comment duplication — the strict-JSON rationale now lives once in `shared/converse.py` with short cross-references at each call site.
- AGGREGATES_TABLE audit (partially) — document generator, document merger, and persona importer now get the env var and `grantReadData` in `api-stack.ts`.

### Still open
- Blocker: verification that the four unversioned model IDs (`global.anthropic.claude-sonnet-5` etc.) are real, invocable Bedrock IDs — still not attached.
- Blocker (narrowed): AGGREGATES_TABLE audit for the remaining Bedrock-invoking API Lambdas — chat, projects, scrapers, and manual import all now resolve through the picker, but neither diff shows their Lambdas gaining `AGGREGATES_TABLE` env / aggregates read; if absent, their surface overrides are silently ignored (the resolver never raises, masking it). Please confirm these envs/grants pre-exist.
- Root cause: continuation still allowed for adaptive-thinking models mid-JSON in `shared/converse.py` — the doctrine documents the hazard but nothing enforces it.
- Confirm the old `require_admin` copy is removed and `AuthorizationError` is imported in `shared/api.py`.
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` PUT.

### New findings
- Minor (`voc-datalake/lib/stacks/processing-stack-consolidated.ts`): removing the `BEDROCK_MODEL_ID` env entirely (rather than honoring it as a documented escape hatch) is a silent behavior change for any deployer who was overriding the enrichment model via env. Worth a line in the changelog/README.
- Nit (`voc-datalake/lambda/api/manual_import_processor.py`): the raw `invoke_model` path keeps a 16000 `max_tokens` with strict-JSON parsing downstream and no continuation — fine today, but consider a pointer to the doctrine here too since this site bypasses `converse()` entirely.

Otherwise the incremental changes are clean — the new `TestCapabilityAwareBedrockBody` tests are well-targeted, and the IAM/env additions follow the existing least-privilege grant pattern.

---

perrozzi:
## Review round 3 disposition + Bedrock-Lambda audit results (`b252334`)

Follow-up on the remaining bot findings, with an audit of every Bedrock-invoking Lambda for model-picker wiring. Everything below is deployed to the test environment and live-verified.

### Audit: which Lambdas can actually see a surface override?

Checked every Lambda with a Bedrock grant for `AGGREGATES_TABLE` env + aggregates read access (the two things `get_active_model_id` needs). Result:

| Lambda | Status |
|---|---|
| scrapers/settings/chat/projects/persona-generator/chat-stream/processor/research/ingestors | ✅ already wired |
| **document-generator job** | ❌ missing both → **fixed** |
| **document-merger job** | ❌ missing both → **fixed** |
| **persona-importer job** | ❌ missing both → **fixed** |
| **manual-import processor** | env/grant OK, but **body broke on Sonnet 5** → **fixed** (see below) |

The three job Lambdas silently fell back to surface defaults (the resolver never raises, by design), meaning a `documents`/`prototype` override never reached the PRD/PR-FAQ/prototype/merge/import jobs. They now get the env + a read grant.

### Live-caught by the audit: manual import parse 400s on Sonnet 5

`manual_import_processor.py` used a raw `invoke_model` with a hardcoded `temperature: 1` + explicit `thinking` budget. After the Sonnet 5 default bump that's a runtime 400 (Sonnet 5 rejects both params) — a regression this PR introduced that the earlier live sweep didn't exercise (parse is an async job). Fixed by routing the model through the `utility` surface picker and building the body capability-aware:

- `temperature` is omitted everywhere (it defaults to 1, the only value accepted with thinking enabled; Sonnet 5/Opus 4.8 reject the explicit param)
- the `thinking` budget is skipped for adaptive-thinking models (Sonnet 5)

Fail-on-revert tests added (`TestCapabilityAwareBedrockBody`): surface routing, Sonnet 5 omits both params, Haiku keeps its thinking budget.

### Other bot items

- **Model-ID validity ("blocker")** — answered by live validation in the [previous validation comment](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/166#issuecomment-4981842051): all four allowlisted IDs have been invoked for real against Bedrock in the test deployment (Sonnet 5 chat default, pinned Haiku override, and now Sonnet 5 manual-import parse + Haiku persona import). No change needed.
- **Continuation-mid-JSON root cause** — doctrine consolidated into a single module-level note in `shared/converse.py` (strict-JSON callers must fit ONE call; the resume seam is lossy at token boundaries); the five call-site comments now just reference it. A JSON-sniffing guard inside `converse()` would be fragile heuristics — the per-site `max_tokens` floors are enforced by `TestStrictJsonTokenHeadroom` instead.
- **Processor `BEDROCK_MODEL_ID` env ignored + dead fallback** — both removed. The env in the processing stack was inert (the enrichment surface default governs), and `or PROCESSOR_FALLBACK_MODEL_ID` was dead code since `get_active_model_id` never returns a falsy value.
- **Read-modify-write on PUT /settings/model** — keeping as-is: admin-only, low-write-rate settings item; last-writer-wins is acceptable and a conditional-write dance adds complexity without a real race to protect against.
- **TS casts in `model-override.ts`** — keeping as-is: the casts are at the DynamoDB unmarshalling boundary and every value is allowlist-checked before use.

### Live verification (test deployment, after redeploying VocProcessingStack + VocApiStack)

- ✅ All 3 job Lambdas now expose `AGGREGATES_TABLE`
- ✅ Manual import parse end-to-end on the Sonnet 5 utility default: 2 reviews extracted (this endpoint 400'd before the fix)
- ✅ Documents surface pinned to Haiku → persona-import job completed → CloudWatch log in `voc-job-persona-importer` shows `Invoking Bedrock with model global.anthropic.claude-haiku-4-5-...` (direct proof the override resolves inside a job Lambda); override then cleared and confirmed
- ✅ Gates: backend pytest 1074 passed / 1 skipped, CDK `tsc` clean, CDK vitest 33/33

perrozzi:
The push since my last review is small and focused: the manual-import raw `invoke_model` path now carries a pointer to the strict-JSON doctrine, and `users_handler.py` drops its local `get_caller_groups`/`require_admin` copy in favor of the shared `shared/api.py` implementation, with the unit tests relocated to `lambda/shared/test/test_api.py` (including a new case for the REST-authorizer bracket-wrapped groups format the old copy mishandled).

### Resolved
- `require_admin` consolidation — the old `users_handler` local copy is removed, all routes gate through the shared implementation, and the new `TestRequireAdmin` tests exercise the `AuthorizationError` path (confirming the import is wired correctly in `shared/api.py`).
- Nit on `manual_import_processor.py` — the raw `invoke_model` site now references the strict-JSON doctrine and documents its single-call `max_tokens` sizing.

### Still open
- Blocker: verification that the four unversioned model IDs (`global.anthropic.claude-sonnet-5` etc.) are real, invocable Bedrock IDs — still not attached.
- Blocker (narrowed): AGGREGATES_TABLE env/grant confirmation for the chat, projects, scrapers, and manual-import API Lambdas — not addressed by this push.
- Root cause: continuation is still permitted for adaptive-thinking models mid-JSON in `shared/converse.py` — documented but unenforced.
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` PUT.

### New findings
- Note (`voc-datalake/lambda/api/users_handler.py`): switching to the shared parser is a behavioral change for admin gating, not just a refactor — callers whose groups claim arrives bracket-wrapped (`"[admins]"`) were previously *denied* and are now *allowed*. This is a bug fix in the right direction (and regression-tested), but it silently changes who passes `require_admin` on the user-management endpoints; worth a line in the PR description/changelog rather than being buried as a code comment.
- Nit: the shared `get_caller_groups` swallows parse exceptions silently (`return []` with no log), whereas the old copy logged the error. Fail-closed is correct, but a `logger.warning` would aid debugging denied-admin reports. On the plus side, the old copy's `logger.info(f"Claims: {claims}")` (which logged full identity claims) is gone — a good cleanup.

Otherwise the incremental changes are clean — moving from the private `app.current_event._data` to the public `raw_event` accessor is the correct powertools pattern, and the test relocation preserves coverage.

---

perrozzi:
## Round 4 disposition (`11ac8e2`)

The round-4 review of `b252334` crossed with my [audit disposition comment](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/166#issuecomment-4983405566) by 26 seconds — most "still open" items are answered there (the bot reads diffs, not comments). Point-by-point:

- **Model-ID blocker** — evidence lives in the two validation comments, not the diff: all four allowlisted IDs have been invoked for real on the test deployment ([initial validation](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/166#issuecomment-4981842051): Sonnet 5 chat/categories + pinned Haiku chat; [audit follow-up](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/166#issuecomment-4983405566): Sonnet 5 manual-import parse + Haiku persona-import with CloudWatch log proof).
- **AGGREGATES_TABLE for chat/projects/scrapers/manual-import (narrowed blocker)** — confirmed pre-existing. Those Lambdas were in the "already wired" column of the audit table: `api-stack.ts` has 16 `AGGREGATES_TABLE` env sites, and each of those roles carries an aggregates grant from before this PR (chat/settings read-write, projects/scrapers/manual-import per their handlers). The three job Lambdas fixed in `b252334` were the only gaps.
- **Continuation-mid-JSON "nothing enforces it"** — enforcement is the test suite, not runtime sniffing: `TestStrictJsonTokenHeadroom` pins the one-call `max_tokens` floor at every strict-JSON call site and fails on revert. A runtime guard would need to detect "this output is strict JSON" from inside `converse()`, which is exactly the kind of fragile heuristic that produces new surprises; happy to revisit if a strict-JSON call site is added without a floor test.
- **`require_admin` old copy** — good catch, and now actually done in `11ac8e2`: `users_handler.py` still had its own pre-picker `get_caller_groups`/`require_admin`. All 8 `/users/*` routes now gate through `shared.api.require_admin` (a strict superset — it also handles the REST-authorizer bracket-wrapped groups claim the local copy mishandled, and doesn't log raw claims). Unit tests moved to `shared/test/test_api.py` with a new bracket-format case. Redeployed and live-verified: admin `GET /users` → 200.
- **`BEDROCK_MODEL_ID` env removal (new minor)** — noted in a new "Deployment notes" section in the PR description. Context: the env was a hardcoded literal inside the stack (not a `cdk.json` context knob and not documented), so overriding it required editing stack source — where the replacement comment now points at the picker.
- **Doctrine pointer at the manual-import raw `invoke_model` site (new nit)** — added in `11ac8e2`, spelling out the single-call form of the doctrine (no continuation on that path; truncated output falls into `unparsed_sections` rather than corrupting).
- **TS casts / read-modify-write PUT** — positions unchanged from the previous disposition (casts sit at the unmarshal boundary with allowlist validation after; the settings item is a rare admin write where last-writer-wins is acceptable).

Gates on `11ac8e2`: backend pytest 1079 passed / 1 skipped, tsc clean. Deployed to the test environment.

perrozzi:
The push since my last review contains **no changes to the model-picker work at all** — it's a bundle of unrelated frontend resilience fixes: Zod normalization boundaries for scrapers (#167/#169) and feedback forms (#171), a route-level error boundary (#173), a mock-server fixture overhaul, and a `runtimeConfig` feature-flag fallback fix. None of the previously identified blockers were addressed.

### Still open
- Blocker: verification that the four unversioned model IDs (`global.anthropic.claude-sonnet-5` etc.) are real, invocable Bedrock IDs — still not attached.
- Blocker (narrowed): AGGREGATES_TABLE env/grant confirmation for the chat, projects, scrapers, and manual-import API Lambdas.
- Continuation still permitted mid-JSON for adaptive-thinking models in `shared/converse.py` (documented, unenforced).
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` PUT.

### New findings
- Scope creep (`frontend/mock-server.js`, `src/api/scrapersSchema.ts`, `src/pages/FeedbackForms/formSchema.ts`, `src/components/RouteErrorBoundary/*`, `src/App.tsx`, `src/runtimeConfig.ts`, 8× `locales/*/components.json`): everything in this push is orthogonal to issue #96 / the model picker. Bundling four separate bug-fix workstreams (#167, #169, #171, #173) into an already-large feature PR being weighed against #154 makes both review and the maintainer's choice harder, and means a revert of the picker would also revert unrelated crash fixes. These should land as their own PR(s).
- Nit (`src/pages/FeedbackForms/formSchema.ts`): `normalizeFeedbackForm` (single-record `.parse`, throws on missing `form_id`) vs `normalizeFeedbackForms` (drops-with-warning) have deliberately different failure semantics — documented, but the throwing variant is exported and easy to misuse in a render path; consider keeping only the list-level normalizer public.

On their own merits the incremental changes are well done: the Zod schemas follow the stated repo convention with regression tests tied to the original crashes, the new `errorBoundary.*` strings are present in all 8 locales, the error boundary keeps observability via `console.error`, and the mock-server exact-key-precedence guard for `/projects/:id` is a nice touch. The problem is placement, not quality.

---

perrozzi:
The push since my last review is another batch of work orthogonal to the model picker: per-key pending state for problem resolution plus a server-side TTL/prune for orphaned resolution keys (#159), collision-proof chat/message IDs (#160), and draft-filter buffering on the Chat page (#161). None of the previously identified model-picker blockers were addressed.

### Still open
- Blocker: verification that the four unversioned model IDs (`global.anthropic.claude-sonnet-5` etc.) are real, invocable Bedrock IDs — still not attached.
- Blocker (narrowed): AGGREGATES_TABLE env/grant confirmation for the chat, projects, scrapers, and manual-import API Lambdas.
- Continuation still permitted mid-JSON for adaptive-thinking models in `shared/converse.py` (documented, unenforced).
- `as Record<string, unknown>` casts in `model-override.ts`.
- Read-modify-write race in `settings_handler.py` model PUT (this push touches the file but only the resolved-problems paths).

### New findings
- Scope creep, again (`src/store/chatStore.ts`, `src/pages/Chat/*`, `src/pages/ProblemAnalysis/*`, `lambda/api/settings_handler.py`): this is the second consecutive push bundling unrelated bug-fix workstreams (#159, #160, #161) into the PR being weighed against #154. The resolved-problems TTL/prune change in particular is a nontrivial backend behavior change (entries now silently expire after 180 days and get deleted under cap pressure) riding inside a feature PR — it deserves its own review and its own revert boundary.
- Should-fix (`voc-datalake/lambda/api/settings_handler.py`): `RESOLVED_PROBLEMS_TTL_DAYS` is read from a Lambda env var that no stack in this PR sets — the 180-day default is effectively hardcoded and the "0 disables expiry" escape hatch is unreachable without a manual console edit. Wire the env var (or at least document it) in `processing-stack`/`api-stack` alongside the Lambda definition.
- Nit (`voc-datalake/lambda/api/settings_handler.py`): `_is_expired_entry` deleting malformed entries under cap pressure means a future schema change to the entry shape (e.g. adding a field and migrating `resolved_at`) would mass-prune live resolutions; a comment noting the shape is load-bearing for prune safety would help.
- Nit (`voc-datalake/frontend/src/store/chatStore.ts` / `Chat.tsx`): a draft set early in a session lingers indefinitely (in-session) and is consumed by whatever conversation is created next, even much later after browsing other conversations — the delete-then-create test pins this as intended, but it's the same "surprising filter state" class the fix targets; consider clearing the draft when the user activates an existing conversation.

On their own merits the changes are solid — regression tests are tied to each issue (including the masked-error `useMutation.isError` fix and the SET/REMOVE race), the DynamoDB REMOVE chunking respects the expression limit, and the accepted prune race is honestly documented. The recurring problem remains placement: these should be their own PRs.

---

PR #168, opened 2026-07-15
fix: Scrapers page no longer crashes on missing or malformed base_url

### LINKED ISSUE #167: [Bug]: Scrapers page crashes with 'Failed to construct URL' on missing or malformed base_url
## What happened

Navigating to /scrapers crashes the whole route with:

```
Unexpected Application Error!
Failed to construct 'URL': Invalid URL
TypeError: Invalid URL
    at ScraperCardHeader (src/pages/Scrapers/ScraperCard.tsx)
```

`ScraperCardHeader` derives the display domain with:

```ts
const domain = scraper.base_url === '' ? t('card.notConfigured') : new URL(scraper.base_url).hostname
```

The guard only covers the empty string. Any other unparseable value — `base_url` missing entirely (`undefined`, as served by the local mock server and possible for older saved configs), whitespace, or a scheme-less `example.com` — throws `TypeError: Invalid URL` during render, and with no route error boundary the entire page dies (observed on development against the local mock).

The run-button disable check on the same component has the sibling gap: `scraper.base_url === ''` leaves the button enabled for `undefined`.

## Expected

A scraper card with a missing/malformed URL renders (showing 'Not configured' or the raw value) instead of crashing the page; the ScraperConfig type's `base_url: string` is a boundary contract that runtime data demonstrably violates, so the component must not assume it.

mundurragacl:
## Summary

Fixes #167 (close manually on merge).

`ScraperCardHeader` computed the display domain with `new URL(scraper.base_url).hostname` behind an empty-string-only guard. Any other unparseable value threw `TypeError: Invalid URL` **during render**, and with no route error boundary the entire /scrapers page died with 'Unexpected Application Error'. Reproduced live against the local mock, which serves scrapers without `base_url` at all (it uses a `url` field) — so `undefined` hit the constructor. Older saved configs and scheme-less values (`example.com`) hit the same path with type-legal strings.

## Change

New `scraperDomainLabel(baseUrl, notConfigured)` owns the boundary:

| runtime value | rendered |
|---|---|
| `undefined` / `''` / whitespace | *Not configured* |
| unparseable but present (`example.com`, garbage) | the raw value — the user still sees what's configured |
| valid URL | its hostname |

The Run button's enable check gets the same normalization (it previously stayed **enabled** for `undefined`).

## Verification

- Unit tests over the label function (all five value classes) + a render regression test using a **type-legal** scheme-less `base_url` that crashed before the fix — mutation-verified (reverting the header line fails exactly that test).
- Browser-verified on the live dev server against the mock that reproduced the crash: cards render 'Not configured', Run disabled, no error boundary.
- Full root gate green (`npm run check` exit 0; 120 Scrapers-suite tests).

## Out of scope (observed while verifying, mock-data drift, cosmetic)

The mock's scraper shape also drifts from `ScraperConfig` in other fields (`url` vs `base_url`, missing frequency → 'undefinedm' in the card footer). This PR makes the page resilient to the crash-class problem; the mock-fixture alignment is tracked as **#169** (which also covers normalizing the remaining single-scraper endpoints, per review round 2).

### DIFF
--- a/voc-datalake/frontend/src/api/scrapersApi.test.ts
+++ b/voc-datalake/frontend/src/api/scrapersApi.test.ts
@@ -1,0 +1,44 @@
+/**
+ * @fileoverview Tests for the scrapers API boundary (issue #167).
+ *
+ * ScraperConfig declares base_url as a required string, but runtime
+ * payloads have delivered scrapers without it. getScrapers normalizes on
+ * entry so the declared contract is true for every consumer.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+const mockFetchApi = vi.fn()
+
+vi.mock('./client', () => ({
+  fetchApi: (...args: unknown[]) => mockFetchApi(...args),
+}))
+
+import { scrapersApi } from './scrapersApi'
+
+describe('scrapersApi.getScrapers base_url normalization', () => {
+  beforeEach(() => {
+    mockFetchApi.mockReset()
+  })
+
+  it('fills a missing base_url with an empty string', async () => {
+    mockFetchApi.mockResolvedValue({
+      scrapers: [
+        { id: 's-1', name: 'No URL yet' },
+        { id: 's-2', name: 'Configured', base_url: 'https://example.com/reviews' },
+      ],
+    })
+
+    const { scrapers } = await scrapersApi.getScrapers()
+
+    expect(scrapers[0].base_url).toBe('')
+    expect(scrapers[1].base_url).toBe('https://example.com/reviews')
+  })
+
+  it('tolerates a payload without a scrapers array', async () => {
+    mockFetchApi.mockResolvedValue({})
+
+    const { scrapers } = await scrapersApi.getScrapers()
+
+    expect(scrapers).toEqual([])
+  })
+})
--- a/voc-datalake/frontend/src/api/scrapersApi.ts
+++ b/voc-datalake/frontend/src/api/scrapersApi.ts
@@ -4,8 +4,22 @@ import type {
   ScraperConfig, ScraperTemplate,
 } from './types'
 
+// What /scrapers actually returns: base_url is declared required on
+// ScraperConfig, but runtime payloads have delivered scrapers without it
+// (issue #167). Model that honestly here and normalize on entry so the
+// declared contract is true for every consumer.
+type RawScraperConfig = Omit<ScraperConfig, 'base_url'> & { base_url?: string }
+
 export const scrapersApi = {
-  getScrapers: () => fetchApi<{ scrapers: ScraperConfig[] }>('/scrapers'),
+  getScrapers: async (): Promise<{ scrapers: ScraperConfig[] }> => {
+    const response = await fetchApi<{ scrapers: RawScraperConfig[] }>('/scrapers')
+    return {
+      scrapers: (response.scrapers ?? []).map((scraper) => ({
+        ...scraper,
+        base_url: typeof scraper.base_url === 'string' ? scraper.base_url : '',
+      })),
+    }
+  },
 
   getScraperTemplates: () => fetchApi<{ templates: ScraperTemplate[] }>('/scrapers/templates'),
 
--- a/voc-datalake/frontend/src/pages/Scrapers/ScraperCard.test.tsx
+++ b/voc-datalake/frontend/src/pages/Scrapers/ScraperCard.test.tsx
@@ -41,6 +41,12 @@ function renderCard(scraper: ScraperConfig) {
 }
 
 describe('scraperDomainLabel', () => {
+  it('resolves the shipped not-configured string from the i18n test setup', () => {
+    // Guard against vacuous passes: if the key stopped resolving, t() would
+    // return the raw key and component + test would "agree" on it.
+    expect(NOT_CONFIGURED).not.toContain('card.notConfigured')
+  })
+
   it('resolves the hostname for a valid URL', () => {
     expect(scraperDomainLabel('https://shop.example.com/reviews?page=1', NOT_CONFIGURED))
       .toBe('shop.example.com')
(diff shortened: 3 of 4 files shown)


perrozzi:
This PR fixes a render-time crash on the /scrapers route by replacing an unguarded `new URL(scraper.base_url).hostname` with a `scraperDomainLabel` helper that never throws, and normalizes the Run button's enable check for `undefined`/whitespace values. The change is small, well-scoped, and comes with both unit tests over the label function and a mutation-verified render regression test — exactly the regression-test pattern this repo asks for on bug fixes. Overall this looks good; the points below are refinements, not blockers.

### Should fix
- **Boundary validation belongs at the API layer too.** The repo pattern is Zod validation at system boundaries, yet `ScraperConfig.base_url` is typed as `string` while runtime data (mock server, older configs) delivers `undefined`. The component-level resilience in `voc-datalake/frontend/src/pages/Scrapers/ScraperCard.tsx` is a good defense-in-depth measure, but the type/schema should reflect reality — either make `base_url` optional in `ScraperConfig` (and its Zod schema, if one exists in `src/api/types.ts`) or normalize it when parsing the API response. Right now the signature `scraperDomainLabel(baseUrl: string | undefined, ...)` disagrees with the declared type it consumes, which will confuse future readers. Acceptable as a follow-up if the schema change has wider blast radius, but it shouldn't be forgotten.
- **Empty-hostname URLs render an empty label.** `new URL('mailto:x@y.com')` and `new URL('file:///tmp')` parse successfully but have `hostname === ''`, so `scraperDomainLabel` returns an empty string instead of the raw value or the not-configured label. One-line fix: `const host = new URL(raw).hostname; return host !== '' ? host : raw`. Worth a test case alongside the existing five value classes in `ScraperCard.test.tsx`.

### Nits
- `hasUrl` in `ScraperCardHeader` re-implements the "present and non-whitespace" check that `scraperDomainLabel` already performs internally. Extracting a tiny `hasConfiguredUrl(baseUrl)` helper (or having the label function's normalization shared) would prevent the two from drifting.
- `NOT_CONFIGURED` in `ScraperCard.test.tsx` is resolved via `i18n.t` at module top level. This works only because the vitest setup initializes i18n before test modules evaluate; if that changes, the constant silently becomes the raw key and the assertions still "pass" against the key text. Resolving it inside the test bodies (or asserting it isn't equal to `'card.notConfigured'`) would make the coupling explicit.
- The PR description flags mock-fixture drift (`url` vs `base_url`, missing frequency → `'undefinedm'`) as out of scope — reasonable, but please open a tracking issue so the mock alignment cleanup doesn't get lost, since it's the same class of type/runtime mismatch that caused this bug.

No blockers — nice, well-tested fix.

---

mundurragacl:
Round 1 addressed in the latest push — both should-fixes and all nits:

- **Boundary validation**: took the normalize-on-parse path rather than making `base_url` optional — the editor uses it as a controlled input throughout, so weakening the type would push `?? ''` handling into every consumer. `getScrapers` now types the wire shape honestly (`RawScraperConfig` with `base_url?: string`) and normalizes missing values to `''` on entry, with boundary tests (missing field → `''`; absent array tolerated). The component-level resilience stays as defense-in-depth, and `scraperDomainLabel`'s `string | undefined` signature now reads as deliberate belt-and-braces over the normalized contract.
- **Empty-hostname URLs**: taken — `mailto:`/`file:` now fall back to the raw value; both cases tested.
- **`hasUrl` drift**: extracted `normalizedBaseUrl()` as the single owner; label and Run-gate both use it.
- **Module-level `i18n.t`**: added the explicit vacuous-pass guard test (asserts the constant isn't the raw key) — same pattern the ScraperEditor tests adopted in #152's review.
- **Mock-fixture drift**: tracked as **#169**.

10 tests across the two touched suites; root `npm run check` exit 0.

perrozzi:
Since the last review, the author added API-boundary normalization in `scrapersApi.getScrapers` (with new tests), fixed the empty-hostname fallback in `scraperDomainLabel` (with test cases for `mailto:`/`file:`), extracted a shared `normalizedBaseUrl` helper used by both the label and the Run-button gate, and added a guard test that the i18n key actually resolves. All previous Should-fix items and two of three nits are addressed.

### Resolved
- Boundary validation at the API layer: `getScrapers` now normalizes `base_url` on entry, with the `RawScraperConfig` type honestly modeling the runtime shape, and both cases are tested in `scrapersApi.test.ts`.
- Empty-hostname URLs (`mailto:`, `file:`) now fall back to the raw value instead of an empty label, with test coverage.
- `hasUrl` drift risk: both the label and Run-button check now go through `normalizedBaseUrl`.
- Vacuous i18n test coupling: the new guard assertion (`NOT_CONFIGURED` must not contain the raw key) makes the setup dependency explicit.

### New findings
- Minor: `voc-datalake/frontend/src/api/scrapersApi.ts` normalizes only presence (`typeof === 'string'`), not whitespace — so `'   '` still passes through and relies on `normalizedBaseUrl` in the component to catch it. That's fine given the helper is the single owner of the trim semantics, just noting the layering is intentional-looking rather than accidental. No action needed.
- Nit: the same drift class presumably exists for other endpoints returning `ScraperConfig` (e.g. anything on `scrapersApi` returning a single scraper) — only `getScrapers` is normalized. Acceptable for this fix's scope, but worth folding into the mock-alignment follow-up.

The mock-fixture tracking issue from the previous review still appears unfiled per the PR description; please open it before merge so it isn't lost. Otherwise the delta is clean and well-tested — good to merge.

---

mundurragacl:
Round 2 addressed:

- **Tracking issue**: it was filed as **#169** just before my round-1 reply (the PR *description* still carried the old 'separate cleanup' wording — now updated to reference it). Also folded your single-scraper-endpoints note into #169's scope.
- **Whitespace layering**: confirmed intentional — the boundary guarantees *presence* (the declared-type violation), while `normalizedBaseUrl` owns the trim semantics at the single point both consumers share. Collapsing trim into the API layer would duplicate the semantics the helper exists to own.

Nothing else outstanding; ready for maintainer merge.

mundurragacl:
One follow-up landed from #170's round 1 (its diff accidentally contained this PR's files before I rebased it): `normalizedBaseUrl` and `scraperDomainLabel` now carry an 'exported for tests (not public API)' note. The single-scraper-endpoint normalization suggestion is already tracked in #169's scope. No functional change; 9 card tests green.

perrozzi:
Since the last review, the only code change is a documentation-comment update: the two exported helpers in `ScraperCard.tsx` (`normalizedBaseUrl`, `scraperDomainLabel`) are now explicitly annotated as "Exported for tests (not public API)". Additionally, the PR description now confirms the mock-fixture drift tracking issue has been filed as #169, covering both the mock alignment and the single-scraper endpoint normalization noted last round.

### Resolved
- Mock-fixture tracking issue is now filed (#169 per the description), including the single-scraper endpoint normalization nit.

The incremental changes are comment-only and clean — good to merge.

---

mundurragacl:
(Count correction: **8** card tests, not 9.)

Issue #169, opened 2026-07-15
[Enhancement]: Align mock-server scraper fixtures with the real ScraperConfig shape


mundurragacl:
Follow-up from #167/#168: the local mock server's `GET /scrapers` fixtures drift from the real `ScraperConfig` shape — they carry `url` instead of `base_url`, and omit the frequency fields (the card footer renders 'undefinedm'). The same type-vs-runtime mismatch class caused the #167 crash; the page is now resilient, but realistic fixtures would exercise the CONFIGURED rendering path in local dev instead of showing every card as 'Not configured', and would catch shape drift earlier. Cleanup: bring the mock's scraper objects in line with `ScraperConfig` (keeping one deliberately-unconfigured fixture for the empty-state path).

perrozzi:
Thanks for the follow-up and for tracing this back to the same shape-drift class that caused the earlier Scrapers page crash!

To confirm our understanding: the mock server's `GET /scrapers` fixtures currently return objects with `url` (instead of `base_url`) and no frequency fields, so in local dev every scraper card renders as 'Not configured' (and previously showed 'undefinedm' in the footer). The proposal is to update the fixtures to match the real `ScraperConfig` shape so the CONFIGURED rendering path is exercised locally, while keeping one intentionally-unconfigured fixture to cover the empty/unconfigured state.

A couple of quick questions:

1. Would you also want the fixtures typed against `ScraperConfig` (or validated in a test) so future shape drift is caught at compile/test time rather than only visually in local dev?
2. Are there other fields in `ScraperConfig` beyond `base_url` and the frequency fields that the fixtures are currently missing or misnaming?

Happy to hear any preferences on how the fixtures should be structured.

_Suggested labels applied: `enhancement`, `javascript`_

---
_Automated review generated with Amazon Bedrock (Claude Fable 5). It may contain mistakes - maintainers have the final say._

mundurragacl:
Scope addition from #168 review round 2: the same normalization applied to `getScrapers` should cover the other endpoints returning `ScraperConfig` (e.g. `saveScraper`'s response) when aligning the fixtures — same drift class.

mundurragacl:
[...]

PR #170, opened 2026-07-15
fix: local-dev gaps — project detail mock, web-search flag on the mock-only path

mundurragacl:
## Summary

Two gaps found while validating locally against the mock (both browser-verified on the live dev server):

### 1. `/projects/:id` did nothing locally

The mock served only the projects LIST; the detail page's `GET /projects/{id}` 404'd. The mock now serves `ProjectDetail`-shaped fixtures for `proj_1`/`proj_2` — personas with goals/frustrations/quotes, a research report and a PR/FAQ document — plus the `/projects/{id}/jobs` polling endpoint (empty). The `/projects/prioritization` exact-key route is explicitly excluded from the id pattern. Verified: the Project Detail page renders fully (tabs, Personas (2), Documents (2), all wizards).

### 2. `VITE_ENABLE_WEB_SEARCH` never worked where it was needed

The env escape hatch for the web-search feature flag (#156) was dead on arrival for its target environment: **mock-only dev has no Cognito vars**, so env-config validation fails and the invalid-env fallback branch rebuilt the config **without the `features` block** — silently hiding the chat toggle and research checkbox. The fallback now carries `features` through, pinned by a regression test that sets ONLY the flag (no Cognito vars) and asserts availability on the localhost:3001 fallback.

Verified in-browser: the 'Web search' toggle renders in the chat filter bar with its tooltip, and toggles once a conversation is active. (On a *fresh* page it still snaps back — that's issue #161, fixed in the pending PR #165.)

## Notes

- `.env.local` itself is gitignored and not part of this PR; devs opt in per machine.
- Related: #169 tracks the broader scraper-fixture drift; this PR only adds the missing project surface.
- Gates: root `npm run check` exit 0; 10 runtimeConfig tests including the new no-Cognito-vars pin; mock passes `node --check`.

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -309,6 +309,85 @@ const handlers = {
   }),
 };
 
+// Project detail fixtures for /projects/:id — shape mirrors ProjectDetail
+// ({ project, personas, documents }) so the Project Detail page works
+// against the mock in local dev.
+const mockProjectDetails = {
+  proj_1: {
+    project: { project_id: 'proj_1', name: 'Q1 Product Improvements', description: 'Customer-driven improvements', status: 'active', created_at: new Date().toISOString(), updated_at: new Date().toISOString(), persona_count: 2, document_count: 2 },
+    personas: [
+      {
+        persona_id: 'persona_1',
+        name: 'Deadline-Driven Dana',
+        tagline: 'Orders late, expects miracles',
+        created_at: new Date().toISOString(),
+        confidence: 'high',
+        feedback_count: 14,
+        goals: ['Get orders delivered before the promised date', 'Track packages without contacting support'],
+        frustrations: ['Late deliveries with no proactive updates', 'Support wait times'],
+        quote: 'If the app just told me the truth about delivery dates, I would stop refreshing it every hour.',
+      },
+      {
+        persona_id: 'persona_2',
+        name: 'Value-Hunter Victor',
+        tagline: 'Compares every price twice',
+        created_at: new Date().toISOString(),
+        confidence: 'medium',
+        feedback_count: 9,
+        goals: ['Find the best price without coupons breaking at checkout'],
+        frustrations: ['Prices changing between cart and checkout'],
+        quote: 'I do not mind paying, I mind being surprised.',
+      },
+    ],
+    documents: [
+      {
+        document_id: 'research_1',
+        document_type: 'research',
+        title: 'Delivery Pain Points Analysis',
+        question: 'What are the main delivery-related pain points?',
+        content: '# Research Report: Delivery Pain Points\n\n## Executive Summary\nCustomers consistently report late deliveries and poor tracking visibility as their top frustrations.\n\n## Key Findings\n1. **Late deliveries** dominate negative feedback (62% of delivery mentions).\n2. **Tracking opacity** amplifies frustration more than lateness itself.\n\n## Recommendations\n- Proactive delay notifications\n- Honest delivery estimates at checkout',
+        feedback_count: 23,
+        created_at: new Date().toISOString(),
+      },
+      {
+        document_id: 'prfaq_1',
+        document_type: 'prfaq',
+        title: 'Proactive Delivery Updates',
+        feature_idea: 'Proactive delivery delay notifications',
+        content: '# PR/FAQ: Proactive Delivery Updates\n\n## Press Release\nToday we announced proactive delivery updates: customers are notified the moment a delay is detected, with an honest new estimate.\n\n## Customer FAQ\n**Q: Will I be spammed with notifications?**\nA: No — you are only notified when the estimate actually changes.\n\n## Internal FAQ\n**Q: What is the hardest technical dependency?**\nA: Carrier webhook latency and estimate recalculation.',
+        created_at: new Date().toISOString(),
+      },
+    ],
+  },
+  proj_2: {
+    project: { project_id: 'proj_2', name: 'Mobile App Redesign', description: 'UX improvements based on feedback', status: 'active', created_at: new Date().toISOString(), updated_at: new Date().toISOString(), persona_count: 1, document_count: 1 },
+    personas: [
+      {
+        persona_id: 'persona_3',
+        name: 'On-the-Go Grace',
+        tagline: 'Thumb-first, patience-last',
+        created_at: new Date().toISOString(),
+        confidence: 'medium',
+        feedback_count: 7,
+        goals: ['Reorder in under 30 seconds'],
+        frustrations: ['App logs her out weekly', 'Checkout buttons below the fold'],
+        quote: 'Every extra tap is a reason to use the website of your competitor.',
+      },
+    ],
+    documents: [
+      {
+        document_id: 'research_2',
+        document_type: 'research',
+        title: 'Mobile Checkout Friction',
+        question: 'Where do mobile users abandon checkout?',
+        content: '# Research Report: Mobile Checkout Friction\n\n## Executive Summary\nSession-loss on login and below-the-fold CTAs drive most abandonment mentions.',
+        feedback_count: 11,
+        created_at: new Date().toISOString(),
+      },
+    ],
+  },
+};
+
 const server = http.createServer((req, res) => {
   res.setHeader('Access-Control-Allow-Origin', '*');
   res.setHeader('Access-Control-Allow-Methods', 'GET, POST, PUT, DELETE, OPTIONS');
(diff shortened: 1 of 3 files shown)


perrozzi:
This PR fixes two local-dev gaps (project-detail fixtures in the mock server, and the `VITE_ENABLE_WEB_SEARCH` fallback dropping the `features` block), but it also ships a third, undisclosed change: the issue #167 `base_url` crash fix in `ScraperCard.tsx` plus normalization in `scrapersApi.ts`. All three changes look correct and well-tested — the regression tests are specific and would fail if the fixes were reverted — but the PR description does not match the diff's scope.

### Should fix

- **Scope/description mismatch.** `voc-datalake/frontend/src/pages/Scrapers/ScraperCard.tsx`, `scrapersApi.ts`, and their two new test files implement the issue #167 fix (render-time `new URL()` TypeError crashing `/scrapers`), yet the PR description explicitly says "this PR only adds the missing project surface." Either split the scraper fix into its own PR or update the description so reviewers and the changelog reflect what actually landed.
- **Hardcoded exact-route exclusion in the mock.** `mock-server.js` excludes only `/projects/prioritization` from the `/projects/:id` pattern via a string comparison. If another exact-key route under `/projects/...` is added to the routes table later, it will silently be shadowed by the id pattern and 404. Prefer checking the routes/handlers table first (or a `RESERVED_PROJECT_KEYS` set) so the exclusion can't drift.

### Nits

- `mock-server.js`: the jobs regex `^\/projects\/[^/]+\/jobs$` also matches `/projects/prioritization/jobs`. Harmless today (returns `{ jobs: [] }`), but consistent with the reserved-keys point above.
- `scrapersApi.ts` normalizes `base_url` only in `getScrapers`; if there's a single-scraper `GET /scrapers/{id}` endpoint it still delivers the raw shape. The `ScraperCard` helpers tolerate `undefined`, so it's covered defensively, but the "contract is true for every consumer" comment slightly overstates.
- `mockProjectDetails` timestamps are computed once at module load (`new Date().toISOString()`), so they're frozen for the server's lifetime — fine for a mock, just noting it's intentional-looking but undocumented.
- `ScraperCard.tsx`: `normalizedBaseUrl` and `scraperDomainLabel` are exported only for testing; a brief `@internal`/"exported for tests" note would prevent them being treated as public API.

Nice touches worth calling out: the `runtimeConfig.test.ts` regression test pins the exact no-Cognito-vars path that was broken, the scraper API uses an honest `RawScraperConfig` type instead of an `as` assertion (matching repo guidance), and the `ScraperCard` tests guard against vacuous i18n-key passes.

---

mundurragacl:
Round 1 addressed:

- **Scope mismatch — good catch, my mistake**: the branch was cut from the #168 branch instead of `development`, leaking the scraper-fix commits into this diff. Rebased onto `development` (force-pushed, hence the new shas): the PR now contains exactly what the description claims — mock project fixtures + the runtimeConfig features fix. The scraper changes remain solely in #168, where they were already reviewed to 'good to merge'.
- **Hardcoded exclusion**: taken, with your stronger formulation — both project regex handlers now defer to the handlers table itself (`!(key in handlers)`), so ANY current or future exact `/projects/...` route wins over the id patterns; the jobs pattern gets the same guard (covering your `/projects/prioritization/jobs` nit). Functionally re-verified: detail 200, prioritization exact-key 200, unknown id 404.
- **Frozen timestamps**: documented as intentional (stable fixtures beat fake freshness in a mock).
- **Single-scraper normalization + exported-for-tests notes**: those live in #168's code, which this PR no longer carries — replying there for the doc note; the single-scraper endpoint normalization is already folded into #169's scope per its round 2.

Gates re-verified after the rebase: `node --check` on the mock, 10 runtimeConfig tests, root `npm run check` exit 0.

perrozzi:
This PR closes two local-dev gaps: the mock server now serves `ProjectDetail`-shaped fixtures for `/projects/:id` (plus an empty `/projects/:id/jobs` polling endpoint), and the invalid-env fallback in `runtimeConfig.ts` now carries the `features` block through so `VITE_ENABLE_WEB_SEARCH` works in mock-only dev (no Cognito vars), pinned by a regression test. The change is small, well-scoped, and the fix comes with a test that fails when reverted, matching the repo's bug-fix policy. Overall this looks clean; no blockers found.

### Should fix

- **`voc-datalake/frontend/src/runtimeConfig.ts`**: the fallback branch now passes `envConfig.features` straight through. Since this branch runs precisely because `envConfig` failed validation, confirm that `features` is guaranteed schema-conformant at that point (e.g., it's constructed as a boolean from `=== 'true'` before validation, or is independently validated). If it's the raw pre-validation value, consider validating the `features` block with its Zod sub-schema before including it, so a malformed value can't leak into the "safe" fallback config.
- **`voc-datalake/frontend/mock-server.js`**: `GET /projects/<unknown-id>` correctly 404s, but `GET /projects/<unknown-id>/jobs` returns `200 { jobs: [] }`. Returning 404 for jobs of nonexistent projects would keep the mock closer to real API behavior and avoid masking bad-id bugs in the polling code.

### Nits

- `mockProjectDetails` duplicates project data (name, description, counts) that presumably also lives in the `GET /projects` list handler — deriving the list from the detail fixtures (or vice versa) would prevent drift when someone edits one but not the other.
- `url.pathname.split('/')[2]` works, but capturing the id in the regex (`/^\/projects\/([^/]+)$/`) and using the match group would be slightly more robust and self-documenting.
- The two route blocks repeat the `!(key in handlers)` exact-key precedence guard and its comment; a tiny helper (e.g., `matchesPattern(pattern) && !(key in handlers)`) would keep future pattern routes consistent.
- The new test relies on module-level config caching across `loadRuntimeConfig()` → `isWebSearchAvailable()`; make sure it uses the same `vi.resetModules()` / env-unstub hygiene as the surrounding tests so ordering can't cause flakes (the diff doesn't show the file's beforeEach, so this is unverified).

---

mundurragacl:
Round 2 addressed — both should-fixes and both nits:

- **features conformance on the fallback**: your instinct was right that 'raw pre-validation value' is the wrong thing to pass through — but rather than re-validating a value constructed two lines earlier, I made the guarantee structural: the `features` object now has a single construction site (`=== 'true'` → literal boolean) shared verbatim by both the validated path and the fallback. There is no longer a version of this code where the fallback can see a differently-shaped features value than the schema path.
- **Jobs 404 for unknown projects**: taken — `/projects/<unknown>/jobs` now 404s like the real API instead of masking bad-id bugs with an empty list. Re-verified all six routing cases (list, detail 200, prioritization exact-key, unknown detail 404, jobs 200, unknown jobs 404).
- **List/detail drift**: taken — `GET /projects` now derives from `Object.values(mockProjectDetails)`, one source of truth.
- **Regex capture groups**: taken in both handlers.

Gates: 10 runtimeConfig tests, `node --check`, root `npm run check` exit 0.

perrozzi:
The incremental changes are a direct response to the previous review: `features` is now built once as a boolean-by-construction object and reused in the fallback, unknown project ids 404 on the `/jobs` route, the projects list is derived from the detail fixtures, and the route regexes use capture groups. All previous should-fix items and most nits are addressed.

### Resolved
- `runtimeConfig.ts`: `features` is constructed once with `=== 'true'` (schema-conformant by construction) and shared with the fallback branch, closing the "raw pre-validation value" concern.
- `mock-server.js`: `GET /projects/<unknown-id>/jobs` now returns 404 like the real API instead of masking bad-id bugs.
- `mock-server.js`: list handler derives from `mockProjectDetails`, eliminating the drift risk (and the previously inconsistent `persona_count: 3` for proj_1 is now correct).
- `mock-server.js`: id extraction uses regex capture groups instead of `split('/')[2]`.

### Still open
- Nit only: the two pattern routes still repeat the `!(key in handlers)` precedence guard inline (a tiny helper was suggested); and the test-file `beforeEach`/env-unstub hygiene remains unverified from the diff. Neither blocks merge.

### New findings
None. Note that the `GET /projects` handler references `mockProjectDetails`, which is declared after the `handlers` object — this is safe because the reference is inside a function invoked at request time, but worth knowing if anyone later adds an eagerly-evaluated handler value. The delta looks clean; good to merge.

---

mundurragacl:
Round 3 acknowledged — converged, holding stable for merge. On the two carried nits: the precedence guard appears exactly twice with the rationale written once above the first use (a helper for a boolean-AND felt like abstraction ceremony at n=2 — will extract at n=3); the runtimeConfig test file's shared beforeEach/afterEach does reset modules and unstub envs/globals for every test including the new one (src/runtimeConfig.test.ts, both hooks). Thanks for the load-order note on mockProjectDetails — the request-time-only guarantee is real and now on the record here.

perrozzi:
This push goes well beyond the previous delta: it adds a route-level error boundary (with i18n keys in all 8 locales), Zod normalization schemas for scrapers and feedback forms with extensive regression tests, render-safety fixes in `ScraperCard`/`FormCard`, and richer mock fixtures for forms/scrapers/stats. Nothing from my previous review regressed; the previously "good to merge" surface is untouched. Note the PR title/description no longer reflects this scope — worth updating before merge.

### Still open
- Nit: the pattern routes still inline the `!(key in handlers)` precedence guard rather than a shared helper (and see below — the new stats route skips it entirely).

### New findings
- **Should-fix — `voc-datalake/frontend/mock-server.js`**: the rewritten `GET /scrapers/{id}/status` returns 200 `never_run` for *any* id, including nonexistent scrapers. This PR itself established the "honest 404 for unknown ids" principle for `/projects/{id}/jobs` and the new `/feedback-forms/{id}/stats`; check `mockScrapers` for existence and 404 unknown ids so client bad-id bugs aren't masked.
- Nit — `formSchema.ts` vs `scrapersSchema.ts`: `normalizeFeedbackForms` logs `parsed.error.issues, raw` while `normalizeScrapers` deliberately logs issues only, citing payload privacy. Feedback form records carry user-authored content (questions, custom fields); drop `raw` for consistency with the stated rationale.
- Nit — `mock-server.js`: the new `/feedback-forms/{id}/stats` pattern route lacks the `!(key in handlers)` exact-key guard applied to the project pattern routes. Low risk today, but it's the same shadowing class the project routes defend against.
- Nit — `ScraperCard.tsx`: `getFrequencyLabel` guards `!Number.isFinite(minutes)` even though the parameter is typed `number` — fine as belt-and-braces, but consider widening the signature to `number | undefined` so the type tells the truth the guard implies.

Everything else in the delta is solid: the error boundary follows i18n conventions across all locales with tests pinning layout survival and observability; both schemas follow the repo's lenient-Zod boundary pattern with strong test coverage (nulls, numeric-string coercion, identity-drop semantics, reference isolation); and `scrapersApi`/`FeedbackForms` normalize at the query boundary rather than in render paths.

---

PR #172, opened 2026-07-15
fix(feedback-forms): survive sparse form records without theme

### LINKED ISSUE #171: [Bug]: Feedback Forms page crashes when a form lacks the theme object (undefined.primary_color)
## What happened

Navigating to /feedback-forms crashes the whole route:

```
Unexpected Application Error!
Cannot read properties of undefined (reading 'primary_color')
TypeError at FormCard (src/pages/FeedbackForms/FormCard.tsx)
```

`FormCard` renders `form.theme.primary_color` unguarded. `FeedbackForm` declares `theme` (and the rest of `FeedbackFormFields`) as required, but runtime data violates that: the local mock's fixtures carry only `{id, name, enabled, submissions, created_at}`, and any form record persisted before newer fields existed could be sparse the same way. Same type-vs-runtime class as #167, same blast radius (no route error boundary → whole page dies).

## Expected

Sparse forms render with sensible defaults (the page already owns `defaultFormConfig`) instead of killing the route; the list is normalized once at its query boundary so the declared contract is true for every consumer.

mundurragacl:
Fixes #171

## Problem

`/feedback-forms` crashed the whole route with:

```
TypeError: Cannot read properties of undefined (reading 'primary_color')
    at FormCard (FormCard.tsx)
```

`FormCard` read `form.theme.primary_color` unguarded, but the wire can deliver sparse form records — rows persisted before `theme`/`custom_fields` existed, or sparse fixtures (the local mock served exactly that shape). Same failure class as #167/#168 (Scrapers) and the ProjectDetail crash: one bad record kills the entire page.

## Fix (same pattern as #168: normalize at the boundary + render-safe component)

- **`normalizeFeedbackForm()` at the list's query boundary** (`select` on the query): missing fields fall back to `defaultFormConfig`, the nested theme is **deep-merged** (a partial theme keeps its set colors), and every form gets fresh object copies so no two forms share default references.
- **`SparseFeedbackForm` honest wire type**: identity fields (`form_id`/`name`/`enabled`) required, everything else optional — including a `Partial` nested theme.
- **Belt-and-braces in `FormCard`**: `const theme = form.theme ?? defaultFormConfig.theme` so the card stays render-safe standalone (it's used with data the page-level normalizer may not own forever).
- **Mock server**: forms now carry the real wire shape (`form_id`, full `FeedbackFormFields`) instead of the drifted `{id, submissions}` shape; one deliberately sparse legacy record (`form_3`) is kept so local dev permanently exercises the normalizer; added `GET /feedback-forms/{id}/stats` (honest 404 for unknown forms).

Per-layer, each fails if its guard is removed — verified by reverting each fix and watching exactly the corresponding test fail:

| Layer | Test | Fails when |
|---|---|---|
| Normalizer | `formTemplates.test.ts` (5 tests: sparse→defaults, identity passthrough, partial-theme deep-merge, complete-record passthrough, no shared refs) | normalizer logic regresses |
| Component | `FormCard.test.tsx` renders a runtime record without theme | FormCard fallback removed |
| Page | `FeedbackForms.test.tsx` pins that a theme-less wire record reaches cards normalized | query-boundary `select` removed |

## Verification

- `npm run check` (lint + typecheck:all + frontend tests + CDK + backend 818 passed) exits 0
- Browser-verified against restarted mock: `/feedback-forms` renders all three forms, sparse `form_3` shows default title/theme, stats populate, no console errors


perrozzi:
This PR fixes a route-crashing `TypeError` on `/feedback-forms` when the API returns sparse form records (persisted before `theme`/`custom_fields` existed). It normalizes at the query boundary via `select`, adds a render-safe fallback in `FormCard`, introduces an honest `SparseFeedbackForm` wire type, updates the mock server to the real wire shape (keeping one deliberately sparse fixture), and adds mutation-verified regression tests per layer. Overall this is a clean, well-scoped fix that follows the established #168 pattern — no blockers.

### Should fix

- **`normalizeFeedbackForm` only handles *absent* fields, not explicit `null`s** (`voc-datalake/frontend/src/pages/FeedbackForms/formTemplates.ts`). Because `...raw` is spread after the defaults, a record delivering `title: null`, `created_at: null`, or `custom_fields: null` will pass the `null` straight through (the `?? []` on `custom_fields` covers `null`, but nothing else does, and `created_at: ''` would be clobbered by an explicit `null`/`undefined` key). DynamoDB-backed APIs commonly emit explicit nulls. Either strip nullish keys from `raw` before spreading, or —
- **Consider a Zod schema instead of a hand-rolled normalizer.** The repo convention is Zod for runtime validation at system boundaries; a schema with `.default()`/`.catch()` on each field would make the `FeedbackForm` contract enforceable (including the `null` case above and unexpected types like `rating_max: "5"`), and `SparseFeedbackForm` could be derived from it. Worth doing here or as an agreed follow-up, since `normalizeFeedbackForm` is now the de facto boundary validator for this route.

### Nits

- `formTemplates.ts`: `created_at: ''` / `updated_at: ''` as defaults may render oddly wherever the page formats dates for legacy records — verify `FormCard`'s date display tolerates empty strings, or default to `undefined`-safe display text.
- `formTemplates.ts`: `custom_fields` items from `raw` are passed by reference; a fresh copy (`[...raw.custom_fields]`) would fully match the "no shared references" guarantee the test asserts for defaults.
- `FormCard.test.tsx`: `Reflect.deleteProperty(form, 'theme')` works, but a small type-guard-friendly builder returning `SparseFeedbackForm` cast-free would read more naturally than mutating a `FeedbackForm` to violate its own type.
- `mock-server.js`: minor — `mockFormTheme` and the stats route are good additions; consider also serving a form with a *partial* theme so local dev exercises the deep-merge branch, not just the fully-missing-theme branch.

The regression-test discipline (one test per guard, mutation-verified) matches the repo's bug-fix requirement nicely.

---

mundurragacl:
Round 1 addressed in b30bcd1 — both should-fix items taken, one nit declined with rationale.

**Explicit nulls + Zod schema (both should-fixes)**: agreed and done together. The normalizer is now `formSchema.ts`, a lenient Zod schema following the `api/feedbackSchema.ts` precedent: per-field `.catch()` so absence, explicit `null`, and wrong types all degrade to `defaultFormConfig` (never rejecting the row); `rating_max` coerces DynamoDB numeric-string round-trips (`'10'` → `10`, junk → default); the nested theme deep-merges partials via field-level catches with an object-level catch for wholesale `null`/absent. `SparseFeedbackForm` now lives beside the schema and `normalizeFeedbackForm(raw: unknown)` is honest about its input. New tests: explicit-null case, coercion/junk-enum fallbacks, input-reference isolation.

Bonus caught by the existing no-shared-references test while porting: Zod's literal `.catch([])` returns the **same array instance on every parse** — switched to the function form `.catch(() => [])`. The hand-rolled version didn't have this bug; the test proved its worth on the replacement immediately.

**Date-rendering nit**: verified — nothing on this page renders `created_at`/`updated_at`; the only date formatter in the folder (`SubmissionsModal.formatDate`) already guards `''` explicitly. Keeping `''` defaults, consistent with `feedbackSchema.ts`'s `lenientString`.

**`custom_fields` copy nit**: solved structurally — Zod re-parses arrays/objects into fresh instances, and the reference-isolation test now pins that inputs aren't shared either.

**`Reflect.deleteProperty` nit — declined**: `FormCard`'s prop is `FeedbackForm` (theme required), and this repo lints `consistent-type-assertions: never`, so a builder returning `SparseFeedbackForm` can't be passed to the component cast-free. `Reflect.deleteProperty` on a complete `FeedbackForm` is the assertion-free way to construct exactly the 'static type lies at runtime' fixture the regression needs; the comment in the test spells that out.

**Partial-theme mock fixture**: added `form_4` with `theme: { primary_color: '#F97316' }` — browser-verified all four fixtures render (sparse form_3 shows default theme, form_4 keeps its orange and defaults the rest).

Full gate green: `npm run check` exits 0 (frontend 1632, CDK 33, backend 818).

perrozzi:
The author replaced the hand-rolled normalizer with a lenient Zod schema (`formSchema.ts`), moved/expanded the tests accordingly, and added a partial-theme fixture to the mock server. Both Should-fix items and most nits from my previous review were addressed; the changes are well-executed.

### Resolved
- Explicit `null` fields now degrade to defaults (`z.catch()` per field, verified by the new "treats explicit nulls" test).
- Zod schema replaces the hand-rolled normalizer, following the `api/feedbackSchema.ts` precedent, and adds type coercion (`rating_max: "5"` → 5, junk → default) with tests.
- Mock server now serves `form_4` with a partial theme, exercising the deep-merge branch in local dev.
- Input `custom_fields` are no longer shared by reference (Zod re-parse, asserted in the reference-freshness test).

### New findings
- **Minor** (`voc-datalake/frontend/src/pages/FeedbackForms/formSchema.ts`): identity fields are now `catch`-lenient — `form_id: z.string().catch('')` means a record with a missing/invalid `form_id` yields `''`, which flows into React list keys and the `['form-stats', form.form_id]` query key. Two such records collide on both. The previous normalizer's `SparseFeedbackForm` contract required identity fields; consider `safeParse`-and-drop (or a warning) for records lacking a usable `form_id` rather than defaulting it.
- **Nit** (`formSchema.ts`): `custom_fields: z.array(customFieldSchema).catch(() => [])` — one non-object element in the array discards *all* custom fields for that form, since the array-level catch fires wholesale. Filtering bad items would be more in the spirit of "degrade per field," though this is an unlikely wire shape.
- **Nit**: the two prior nits that remain (empty-string `created_at`/`updated_at` defaults; `Reflect.deleteProperty` in `FormCard.test.tsx`) are unchanged — fine to leave as-is, just noting they weren't picked up.

Otherwise the incremental changes are clean; the null-handling and coercion tests are good mutation targets and the schema/test split (`formSchema` vs `formTemplates`) keeps the boundary validator discoverable.

---

mundurragacl:
Round 2 addressed in 7eff256 — the Minor and the custom_fields nit both taken.

**Identity collision (Minor)**: agreed — identity is the one thing the normalizer can't invent. `form_id` is now strict (`z.string().min(1)`, no catch), and a new `normalizeFeedbackForms()` list boundary `safeParse`s each record, dropping rows without a usable `form_id` with a `console.warn` (issues + raw record). The page's `select` uses the list normalizer; the single-record `normalizeFeedbackForm` now throws on identity-less records, which is honest for its remaining callers (tests). Pinned by a test feeding missing/`''`/`null` form_ids — exactly one identified record survives, three warnings fire.

**custom_fields wholesale-discard (nit)**: taken — the array now goes through per-item salvage (`z.array(z.unknown())` → `flatMap` + `safeParse`): junk elements are filtered, valid ones survive with their per-field catches. Test: `['junk-string', {valid}, 42]` → keeps the valid one.

**Leftover nits (created_at `''` defaults, `Reflect.deleteProperty`)**: as noted in the round-1 reply, both were declined with rationale (nothing renders those dates on this page + the only formatter guards `''`; the delete is the assertion-free way to build a 'static type lies' fixture under `consistent-type-assertions: never`). Leaving as-is per your 'fine to leave' note.

Full gate green: `npm run check` exits 0 (frontend 1635, CDK 33, backend 818).

perrozzi:
The author addressed the two remaining findings from my last review: identity is now `z.string().min(1)` with a list-level `normalizeFeedbackForms` that drops (and warns on) records lacking a usable `form_id`, and `custom_fields` now salvage valid items per-element instead of discarding the array wholesale. Both changes come with dedicated mutation-target tests, and the query boundary in `FeedbackForms.tsx` was updated to the list-level normalizer.

### Resolved
- Identity-less records no longer default `form_id` to `''` (no more React key / `['form-stats', form_id]` query-key collisions); they're dropped with a warning, tested in `formSchema.test.ts`.
- One junk element in `custom_fields` no longer discards the whole array; per-item `safeParse` salvage is tested.

### New findings
- **Nit** (`formSchema.ts`): the `console.warn` message assumes the only `safeParse` failure is a missing identity ("without usable identity"). That's true today because every other field catches, but if a future field is added without a `catch`, the message will mislead. Consider a neutral message or a comment noting the invariant.
- **Nit**: dropped records are invisible to the user (console-only). Acceptable degradation for this failure class, but worth a thought whether a count of skipped records should surface in the UI eventually (would need an i18n key if so).

Otherwise the incremental changes are clean — the drop-vs-invent identity decision is well-documented in the schema comments, and the two new test blocks fail precisely when their respective guards are removed.

---

perrozzi:
Since the last review, the only change is a rewording of the `console.warn` message in `normalizeFeedbackForms` (plus an explanatory comment), directly addressing the first nit from my previous review.

### Resolved
- `formSchema.ts` warning no longer assumes the only failure mode is a missing identity; the neutral "failed schema validation" wording plus the invariant comment is exactly what was asked for.

### Still open
- Nit: dropped records remain console-only with no user-visible indication (acceptable degradation, noted previously as a future consideration — would need an i18n key if surfaced).

The incremental change is clean; nothing new to flag.

---

mundurragacl:
Round 3 addressed in 35360c6.

**Warn wording (nit)**: taken — message is now neutral ('failed schema validation') with a comment noting the today-only-form_id invariant, so it stays accurate if a future catch-less field lands.

**Surfacing dropped records in the UI (nit)**: agreed it's console-only degradation and deliberately so for this PR — a skipped-records count in the UI needs an i18n key ×8 locales and a product decision about where it lives, which is beyond a crash fix. Happy to file a follow-up issue if maintainers want it; the warn (issues + raw record) keeps it diagnosable meanwhile.

Tests unchanged and green (the drop test asserts call count, not message text — 21/21 in the folder).

mundurragacl:
One more commit (looseObject round-trip safety), prompted by the #175 review's consistency point: this page's edit modal receives the normalized record from `FormCard.onEdit` and saves it back, so the stripping `z.object` would have silently shed unenumerated backend fields on every edit. `FeedbackFormSchema`/`themeSchema`/`customFieldSchema` are now `z.looseObject` (Zod v4 passthrough idiom, same as scrapersSchema.ts in #175), pinned by a passthrough test asserting `brand_name` + a `future_field` survive normalization. Gate green (frontend 1636, CDK 33, backend 818).

perrozzi:
Since the last review, the schemas switched from `z.object` to `z.looseObject` (`FeedbackFormSchema`, `themeSchema`, `customFieldSchema`) so unknown backend fields survive an edit round-trip, with a new mutation-verifiable test and updated doc comments. This wasn't a prior finding — it's a proactive hardening — and it's implemented consistently with the stated `api/scrapersSchema.ts` precedent.

### Still open
- Nit: dropped records remain console-only with no user-visible indication (acceptable degradation; would need an i18n key if ever surfaced).

### New findings
- Nit (`voc-datalake/frontend/src/pages/FeedbackForms/formSchema.ts`): the doc comment "Zod returns fresh objects/arrays, so no two normalized forms share default (or input) references" is now slightly overstated. With `looseObject`, passthrough keys are shallow-copied — a nested unknown value (e.g. `future_field: { nested: true }`) shares its reference with the input object. The existing no-shared-refs test only covers enumerated fields (`theme`, `custom_fields`), so this gap is neither documented accurately nor pinned by a test. Low risk in practice (query data shouldn't be mutated), but worth a one-line comment correction so the invariant claim matches reality.

Otherwise the incremental change is clean: the passthrough test is genuinely mutation-sensitive (reverting to strict objects fails it), and the loose theme/custom-field schemas don't weaken any of the #171 guards.

---

mundurragacl:
Round 4 nit taken in the latest commit — the doc comment now scopes the freshness invariant to enumerated fields (theme/custom_fields re-parsed fresh) and states explicitly that passthrough values are shallow-copied and share references with the input. No behavior change; tests unchanged and green.

perrozzi:
The only change since the last review is a doc-comment correction in `formSchema.ts`, directly addressing my previous nit about the overstated no-shared-references claim under `looseObject`.

### Resolved
- Doc comment now accurately states that only enumerated fields (`theme`, `custom_fields`) get fresh instances while passthrough (unknown) values are shallow-copied and share references with the input — matches the actual `looseObject` behavior.

### Still open
- Nit: dropped records remain console-only with no user-visible indication (acceptable degradation; would need an i18n key if ever surfaced).

### New findings
None. The comment fix is accurate and the invariant it now documents (passthrough sharing) is low-risk since query data isn't mutated. A test pinning the passthrough-sharing behavior would be nice-to-have but isn't required for a doc-only change. Change is clean.

---

PR #174, opened 2026-07-15
feat(frontend): route-level error boundary so one bad page can't blank the app

### LINKED ISSUE #173: No route-level error boundary: any render error blanks the entire app
## Problem

Three recent crashes (#159 ProjectDetail, #167 Scrapers, #171 Feedback Forms) shared the same amplifier: the router has no `errorElement`/ErrorBoundary, so a render error in any single card unmounts the **whole application** and shows the default react-router error screen:

```
💿 Hey developer 👋
You can provide a way better UX than this when your app throws errors by
providing your own ErrorBoundary or errorElement prop on your route.
```

Each root cause was fixed with boundary normalization + regression tests, but defense-in-depth is missing: the next unguarded property read will blank the app again.

## Proposal

- Add a route-level error boundary (react-router `errorElement`) rendering a branded fallback: error summary, 'go home' + 'reload' actions
- Scope it so the sidebar/layout survives — only the failing route content is replaced
- Localized strings (all 8 locales)
- Regression test: a route that throws renders the fallback instead of unmounting the app

mundurragacl:
Fixes #173

## Problem

Three recent crashes (#159 ProjectDetail, #167 Scrapers, #171 Feedback Forms) shared the same amplifier: the router had no `errorElement`, so a render error in one card unmounted the **whole app** into react-router's default developer error screen. Each root cause got its own fix + regression tests; this adds the missing defense-in-depth so the next unguarded property read degrades gracefully instead.

## Change

- **`RouteErrorBoundary`** (`src/components/RouteErrorBoundary/`): branded fallback with the error summary, a reload button, and a go-home link. Error extraction from `useRouteError()` is type-guarded (`isRouteErrorResponse` → `instanceof Error` → `String()`), no assertions.
- **`App.tsx`**: a `page()` helper attaches the existing `Suspense` wrapper **and** `errorElement` uniformly to every child route. Errors bubble to the nearest `errorElement`, so page failures replace only the route outlet — **the layout and sidebar stay mounted and interactive**. The root `/` and `/login` routes carry the same boundary for failures in `Layout`/`ProtectedRoute` themselves.
- **i18n**: `errorBoundary.*` strings added to the `components` namespace in all 8 locales, inserted preserving key order.

## Regression tests

`RouteErrorBoundary.test.tsx` (6 tests):
- a throwing child route renders the fallback (role=`alert`, real error message) **while the layout sibling stays mounted** — the exact #173 failure mode inverted
- 'Go to home' recovers to a healthy route without a page reload
- reload action present
- `describeRouteError` unit coverage for route-response / Error / other branches

## Verification

- `npm run check` exits 0 (frontend 1627 tests, CDK 33, backend 818)
- `npm run i18n:validate` clean for the new keys (unrelated parser churn on `dashboard.json` discarded per repo convention)
- **Browser-verified against the real crash**: on a branch without the #172 data fix, `/feedback-forms` with a sparse record now shows the branded fallback inside the intact layout (sidebar/header/breadcrumbs all alive), error text `Cannot read properties of undefined (reading 'primary_color')` surfaced, and 'Go to home' navigates to a fully rendered Home without reload.

## Relationship to #172

Independent and complementary: #172 fixes the Feedback Forms root cause (normalize sparse records); this PR contains the blast radius of any future render error. No file overlap — both branch from current `development`.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/components.json
+++ b/voc-datalake/frontend/public/locales/de/components.json
@@ -53,6 +53,12 @@
     "downloadTXT": "Als TXT herunterladen",
     "kiroPromptTip": "Tipp: Kiro-Eingabeaufforderung in Projekteinstellungen konfigurieren"
   },
+  "errorBoundary": {
+    "description": "Auf dieser Seite ist ein unerwarteter Fehler aufgetreten. Sie können sie neu laden oder zur Startseite zurückkehren — der Rest der Anwendung ist nicht betroffen.",
+    "goHome": "Zur Startseite",
+    "reload": "Seite neu laden",
+    "title": "Etwas ist schiefgelaufen"
+  },
   "feedbackCard": {
     "copyText": "Text kopieren",
     "details": "Einzelheiten",
--- a/voc-datalake/frontend/public/locales/en/components.json
+++ b/voc-datalake/frontend/public/locales/en/components.json
@@ -53,6 +53,12 @@
     "downloadTXT": "Download as TXT",
     "kiroPromptTip": "Tip: Configure Kiro prompt in project settings"
   },
+  "errorBoundary": {
+    "description": "This page hit an unexpected error. You can reload it or go back home — the rest of the app is unaffected.",
+    "goHome": "Go to home",
+    "reload": "Reload page",
+    "title": "Something went wrong"
+  },
   "feedbackCard": {
     "copyText": "Copy text",
     "details": "Details",
--- a/voc-datalake/frontend/public/locales/es/components.json
+++ b/voc-datalake/frontend/public/locales/es/components.json
@@ -57,6 +57,12 @@
     "downloadTXT": "Descargar como TXT",
     "kiroPromptTip": "Consejo: Configura el prompt de Kiro en la configuración del proyecto"
   },
+  "errorBoundary": {
+    "description": "Esta página encontró un error inesperado. Puedes recargarla o volver al inicio; el resto de la aplicación no se ve afectado.",
+    "goHome": "Ir al inicio",
+    "reload": "Recargar página",
+    "title": "Algo salió mal"
+  },
   "feedbackCard": {
     "copyText": "Copiar texto",
     "details": "Detalles",
--- a/voc-datalake/frontend/public/locales/fr/components.json
+++ b/voc-datalake/frontend/public/locales/fr/components.json
@@ -57,6 +57,12 @@
     "downloadTXT": "Télécharger en TXT",
     "kiroPromptTip": "Conseil : Configurez l'invite Kiro dans les paramètres du projet"
   },
+  "errorBoundary": {
+    "description": "Cette page a rencontré une erreur inattendue. Vous pouvez la recharger ou revenir à l'accueil — le reste de l'application n'est pas affecté.",
+    "goHome": "Retour à l'accueil",
+    "reload": "Recharger la page",
+    "title": "Une erreur est survenue"
+  },
   "feedbackCard": {
     "copyText": "Copier le texte",
     "details": "Détails",
--- a/voc-datalake/frontend/public/locales/ja/components.json
+++ b/voc-datalake/frontend/public/locales/ja/components.json
@@ -53,6 +53,12 @@
     "downloadTXT": "TXTでダウンロード",
     "kiroPromptTip": "ヒント：プロジェクト設定でKiroプロンプトを設定してください"
   },
+  "errorBoundary": {
+    "description": "このページで予期しないエラーが発生しました。再読み込みするか、ホームに戻ってください。アプリの他の部分には影響ありません。",
+    "goHome": "ホームへ戻る",
+    "reload": "ページを再読み込み",
+    "title": "問題が発生しました"
+  },
   "feedbackCard": {
     "copyText": "テキストをコピー",
     "details": "詳細",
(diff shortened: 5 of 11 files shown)


perrozzi:
This PR adds a `RouteErrorBoundary` component and wires it as `errorElement` on every route via a `page()` helper in `App.tsx`, so a render error in one page no longer unmounts the whole app. The implementation follows repo conventions well: type guards instead of `as` in `describeRouteError`, i18n keys added to all 8 locales, and regression tests that pin the exact #173 failure mode. Overall this is a clean, well-scoped change.

### Should fix

- **Errors are displayed but never reported.** `voc-datalake/frontend/src/components/RouteErrorBoundary/index.tsx` renders the error but doesn't `console.error` it or forward it to any monitoring hook. Once this boundary catches errors gracefully, production render crashes become invisible unless a user files a report — the opposite of what helped diagnose #159/#167/#171. Log the full error object (not just the summarized message) inside the component, e.g. in a `useEffect`.
- **Raw `error.message` is shown to end users.** Surfacing messages like `Cannot read properties of undefined (reading 'primary_color')` leaks internal implementation detail into the UI. Consider showing the detail only in dev builds (`import.meta.env.DEV`) and a generic summary in production, keeping the full detail in the log/report path above.

### Nits

1. `describeRouteError` ignores `error.data` for route error responses; loader-thrown Responses often carry the useful message there (`404 Not Found` alone is less actionable).
2. The tests assert literal English strings (`'Something went wrong'`, `/go to home/i`), coupling them to the `en` locale bundle via the global test i18n setup. Querying by key-stable attributes (or role + testid) would be less brittle if copy changes.
3. The "reload action" test only asserts the button exists; the `window.location.reload` wiring is untested (understandable in jsdom, but a `vi.spyOn`/stub of `location.reload` would cover it).
4. `App.tsx`'s `page()` uses the `React.ReactNode` type — confirm the file has a `React` namespace (or `import type { ReactNode }`) in scope; the diff doesn't show it, though `npm run check` passing suggests it's fine.
5. Minor duplication: the `/` root route's `errorElement` renders full-viewport-ish (`min-h-[60vh]`) both inside and outside the layout; acceptable, just noting the fallback isn't layout-aware.

No blockers — with error reporting added, this is good to merge.

---

mundurragacl:
Round 1 addressed in adad388 — both should-fixes taken, nits 1/3/4 taken, 2/5 declined with rationale.

**Error reporting (should-fix)**: added — the component now `console.error`s the **full error object** (stack intact) in a `useEffect`, so gracefully-caught crashes stay visible to CloudWatch RUM / browser monitoring. Pinned by a new test asserting the spy received the actual `Error` instance, not a stringified summary.

**Raw message leakage (should-fix)**: the technical detail panel is now `import.meta.env.DEV`-only; production users get the generic localized copy while the full detail stays in the log path. (Vitest runs in DEV, so the existing assertions on the detail text keep working — noted in the test.)

**Nit 1 (`error.data`)**: taken — `describeRouteError` appends route-response `data` when it's a non-empty string (`404 Not Found — No route matches`), with a test for the null-data case.

**Nit 3 (reload wiring)**: taken — `vi.stubGlobal('location', { ...window.location, reload })` (cast-free; jsdom's Location is non-configurable, so the whole object is swapped) and the test now clicks the button and asserts the spy fired.

**Nit 4 (`React.ReactNode`)**: you read it right — it compiled via the UMD global type. Now an explicit `import type { ReactNode }`.

**Nit 2 (English-string coupling) — declined**: the repo's `src/test/setup.ts` deliberately initializes i18n with the real English bundles so 'component assertions can match real translated strings instead of raw i18n keys' (its own words). Matching visible copy is the established convention here; role/testid-only queries would diverge from every other component test in the repo.

**Nit 5 (fallback not layout-aware) — acknowledged, keeping as-is**: `min-h-[60vh]` centers acceptably both inside the layout outlet and full-screen at the root; a layout-aware variant would double the surface for marginal gain on the rare Layout-itself-crashed path.

Full gate green: `npm run check` exits 0 (frontend 1629, CDK 33, backend 818).

perrozzi:
The new push adds error reporting (`console.error` of the full error object in a `useEffect`), gates the raw error detail behind `import.meta.env.DEV`, enriches `describeRouteError` with the route response `data` suffix, switches to an explicit `import type { ReactNode }`, and adds tests for the reporting, the reload wiring (via `vi.stubGlobal`), and the new `data`-handling branches. Both should-fix items and most nits from the previous review are addressed.

### Resolved
- Errors are now reported: full error object logged via `console.error` in `useEffect`, with a test pinning the observability path.
- Raw error messages are dev-only; production shows only the generic i18n summary.
- Nit 1: `describeRouteError` now surfaces string `error.data`, with positive and negative test cases.
- Nit 3: reload action is now behaviorally tested via a `location` stub.
- Nit 4: `ReactNode` is now explicitly imported in `App.tsx`.

### Still open
- Nit 2: tests still assert literal English copy (`'Something went wrong'`, `/go to home/i`, `/reload page/i`), coupling them to the `en` bundle.

### New findings
- Minor/nit only: the production branch of the `import.meta.env.DEV` gate is untested — the test comment acknowledges vitest always runs in DEV, so "detail hidden in prod" is asserted nowhere. Stubbing `import.meta.env.DEV` is awkward in Vite; acceptable to leave, but worth noting the gate is only browser-verified.
- Nit: `vi.stubGlobal('location', { ...window.location, reload })` spreads a jsdom `Location`, whose properties live on the prototype, so the stub may lose `href`/`pathname` etc. Harmless here (only `reload` is exercised), but could surprise if the fallback ever reads `location` fields.

The incremental changes are clean and directly responsive to the previous review — no blockers, good to merge.

---

PR #175, opened 2026-07-15
fix(scrapers): normalize sparse records + drifted run status (undefinedm)

### LINKED ISSUE #169: [Enhancement]: Align mock-server scraper fixtures with the real ScraperConfig shape

mundurragacl:
Fixes #169

> **Stacked on #168** (`fix/scraper-card-invalid-url`) because both PRs touch the same `getScrapers` boundary — this PR's base is #168's branch so the diff shows only the delta. **Merge #168 first**; GitHub will auto-retarget this PR to `development` when #168's branch is deleted on merge.

## Problem (user-visible on /scrapers)

- **Frequency: `undefinedm`** — `getFrequencyLabel(scraper.frequency_minutes)` interpolated `undefined` for records that omit the field
- **`Last: pages, reviews`** with blank counts + a stray ⚠ badge — the status endpoint's mock shape drifted (`items_scraped` vs `items_found`, no `pages_scraped`, `errors` as a number, status `success`)

Same class as #167/#171: `ScraperConfig` declares every field required; runtime records persisted before newer fields existed disagree.

## Fix (same architecture as #172: lenient Zod at the boundary + render-safe component)

- **`api/scrapersSchema.ts`**: `ScraperConfigSchema` + `ScraperRunStatusSchema` with per-field `.catch()` — absence, explicit nulls, and wrong types degrade to **conservative wire defaults**, deliberately not the editor's `DEFAULT_SCRAPER` values: a record with no schedule is `frequency_minutes: 0` (**'Manual only'** — truthful) rather than the editor's 'Daily'; `base_url: ''` is the not-configured state. Partial `pagination` deep-merges; `urls`/`errors` salvage per item; numeric strings coerce (`'30'` → 30). `id` is strict — id-less records are dropped with a warning (inventing `''` would collide React keys and poll a nonsense status endpoint), mirroring #172's identity decision.
- **`scrapersApi`**: `getScrapers`/`getScraperStatus` normalize on entry, superseding #168's base_url-only inline map (same semantics, now schema-enforced).
- **`ScraperCard`**: belt-and-braces — a runtime record without frequency renders `—`, never `undefinedm`.
- **mock-server**: fixtures carry the real `ScraperConfig` shape; `scraper_2` stays deliberately sparse (identity only) so local dev permanently exercises the normalizer (per #169's ask); status route serves the real RunStatus shape with a `never_run` fallback.

| Layer | Test | Verified by reverting |
|---|---|---|
| Schema | `scrapersSchema.test.ts` (11: defaults, nulls, coercion, partial-pagination merge, urls/errors salvage, id-drop+warn, passthrough) | — (unit) |
| API wiring | `scrapersApi.test.ts` (+2: frequency default through `getScrapers`, drifted-status degradation through `getScraperStatus`) | bypassing each normalizer failed exactly its test |
| Component | `ScraperCard.test.tsx` (+3: runtime-sparse record renders `—` not `undefinedm`; known label; 'Manual only' for 0) | removing the card guard failed exactly the `undefinedm` test |

## Verification

- `npm run check` exits 0 (frontend 1647, CDK 33, backend 818)
- Browser-verified against the restarted mock: card 1 shows **Every 30 minutes / 4 URLs / Last: 3 pages, 42 reviews ✓**; sparse card 2 shows **Not configured / Manual only / Never**; `document.body.innerText.includes('undefined')` → false

## Heads-up

`mock-server.js` is also touched by #172 (feedback-forms fixtures) — whichever merges second may need a trivial conflict resolution in that file.

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -143,11 +143,33 @@ const mockFeedbackForms = [
   { id: 'form_3', name: 'Support Feedback', enabled: false, submissions: 89, created_at: new Date().toISOString() },
 ];
 
-// Mock scrapers
+// Mock scrapers — real ScraperConfig shape (issue #169). scraper_2 is
+// deliberately sparse (identity fields only), mirroring records persisted
+// before newer fields existed, so local dev exercises the normalizeScrapers()
+// boundary: base_url '' renders Not configured, frequency 0 renders
+// 'Manual only' (never 'undefinedm').
 const mockScrapers = [
-  { id: 'scraper_1', name: 'Product Reviews', url: 'https://example.com/reviews', enabled: true, schedule: 'rate(30 minutes)', last_run: new Date().toISOString(), status: 'success' },
-  { id: 'scraper_2', name: 'Forum Posts', url: 'https://forum.example.com', enabled: false, schedule: 'rate(1 hour)', last_run: null, status: 'disabled' },
+  {
+    id: 'scraper_1', name: 'Product Reviews', enabled: true,
+    base_url: 'https://example.com/reviews', urls: ['https://example.com/reviews?sort=recent'],
+    frequency_minutes: 30, extraction_method: 'css',
+    container_selector: '.review', text_selector: '.review-text',
+    rating_selector: '.review-stars@data-rating', author_selector: '.review-author',
+    date_selector: '.review-date',
+    pagination: { enabled: true, param: 'page', max_pages: 3, start: 1 },
+    last_run: new Date().toISOString(), items_found: 42,
+  },
+  // Sparse legacy record: identity fields only.
+  { id: 'scraper_2', name: 'Forum Posts', enabled: false },
 ];
+const mockScraperRuns = {
+  scraper_1: {
+    scraper_id: 'scraper_1', status: 'completed',
+    started_at: new Date(Date.now() - 3600000).toISOString(),
+    completed_at: new Date(Date.now() - 3590000).toISOString(),
+    pages_scraped: 3, items_found: 42, errors: [],
+  },
+};
 
 // Mock S3 data explorer
 const mockBuckets = [
@@ -338,17 +360,14 @@ const server = http.createServer((req, res) => {
     return;
   }
 
-  // Handle scraper status by ID
+  // Handle scraper status by ID — real RunStatus shape (issue #169):
+  // pages_scraped/items_found/errors, 'never_run' for scrapers without runs.
   if (req.method === 'GET' && url.pathname.match(/^\/scrapers\/[^/]+\/status$/)) {
     const id = url.pathname.split('/')[2];
-    const scraper = mockScrapers.find(s => s.id === id);
+    const run = mockScraperRuns[id];
     res.writeHead(200);
-    res.end(JSON.stringify({
-      id,
-      status: scraper ? scraper.status : 'unknown',
-      last_run: scraper ? scraper.last_run : null,
-      items_scraped: Math.floor(Math.random() * 50) + 10,
-      errors: 0
+    res.end(JSON.stringify(run ?? {
+      scraper_id: id, status: 'never_run', pages_scraped: 0, items_found: 0, errors: [],
     }));
     return;
   }
--- a/voc-datalake/frontend/src/api/scrapersApi.test.ts
+++ b/voc-datalake/frontend/src/api/scrapersApi.test.ts
@@ -41,4 +41,32 @@ describe('scrapersApi.getScrapers base_url normalization', () => {
 
     expect(scrapers).toEqual([])
   })
+
+  it('defaults a missing frequency so the card can never render undefinedm (issue #169)', async () => {
+    mockFetchApi.mockResolvedValue({ scrapers: [{ id: 's-1', name: 'Sparse', enabled: false }] })
+
+    const { scrapers } = await scrapersApi.getScrapers()
+
+    expect(scrapers[0].frequency_minutes).toBe(0)
+    expect(scrapers[0].urls).toEqual([])
+    expect(scrapers[0].pagination.enabled).toBe(false)
+  })
+})
+
+describe('scrapersApi.getScraperStatus normalization (issue #169)', () => {
+  beforeEach(() => {
+    mockFetchApi.mockReset()
+  })
+
+  it('degrades the drifted status shape to safe counts instead of blanks', async () => {
+    // What the mock server historically returned: items_scraped (not
+    // items_found), no pages_scraped, errors as a number.
+    mockFetchApi.mockResolvedValue({ id: 's-1', status: 'success', items_scraped: 12, errors: 0 })
+
+    const status = await scrapersApi.getScraperStatus('s-1')
+
+    expect(status.pages_scraped).toBe(0)
+    expect(status.items_found).toBe(0)
+    expect(status.errors).toEqual([])
+  })
 })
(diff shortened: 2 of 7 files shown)


perrozzi:
This PR hardens the `/scrapers` boundary against sparse legacy records and a drifted run-status shape by introducing lenient Zod schemas (`scrapersSchema.ts`), wiring them into `scrapersApi`, adding a render-safe guard in `ScraperCard`, and updating the mock server to the real wire shapes. It follows the repo's established pattern (Zod validation at the system boundary, regression tests with vitest), the conservative-defaults reasoning is well documented, and the tests are thorough. Overall this is a clean, well-scoped fix. Note it's stacked on #168, so verify the base retargets correctly before merge.

### Should fix
- **`normalizeScraperRunStatus` can still throw** (`voc-datalake/frontend/src/api/scrapersSchema.ts`): every *field* has a `.catch()`, but `ScraperRunStatusSchema` has no object-level catch and `normalizeScraperRunStatus` uses `.parse()`. A non-object response (`null`, `undefined`, a string error body) will throw a `ZodError` out of `getScraperStatus`. Either add a top-level `.catch(() => ({ ...never_run defaults }))` or use `safeParse` with a `never_run` fallback — that's the same "degrade, don't reject" philosophy the rest of the schema follows. Add a test for `normalizeScraperRunStatus(null)`.
- **Unknown-key stripping risks silent data loss on round-trips**: `ScraperConfigSchema` is a plain `z.object`, so any field the backend returns that isn't enumerated (e.g., `created_at`/`updated_at` or future additions) is silently dropped. If the edit flow ever reads a scraper from `getScrapers()` and PUTs it back, those fields are lost. Consider `.passthrough()` (with the catches unchanged) or at least a comment documenting that the normalized list must not be used as an update payload. Note the "passthrough" test only asserts `toMatchObject` on known keys, so it wouldn't catch this.
- **Verify `FREQUENCY_OPTIONS` includes `value: 0` → "Manual only"**: the diff doesn't touch `FREQUENCY_OPTIONS`, but the schema's `frequency_minutes` default of `0` and the tests (`ScraperCard.test.tsx`, PR description) depend on that option existing. If it doesn't, sparse records would render `0m`. If the tests pass this is fine — just confirming since the option isn't visible in the diff.

### Nits
- `console.warn(..., raw)` in `normalizeScrapers` logs the entire raw record; scraper configs are low-risk for PII, but consider logging only the offending issues and an identifier rather than the full payload.
- `getFrequencyLabel`'s fallback `` `${minutes}m` `` remains a hardcoded, non-i18n string for unknown numeric frequencies; the new `—` literal is arguably a symbol, but a `common` i18n key (or a localized "every N minutes" format) would be more consistent with the repo's i18n rule.
- Double blank line inserted between describe blocks in `ScraperCard.test.tsx`.
- `extraction_method: z.enum(['css', 'jsonld']).optional().catch(undefined)` will silently drop any future backend-supported method; if the backend enum grows, this must change in lockstep — worth a comment.
- As the PR itself notes, `mock-server.js` conflicts with #172 — coordinate merge order.

---

mundurragacl:
Round 1 addressed in 76752cd — all three should-fixes taken (one was verification), nits mostly taken, one declined with rationale.

**`normalizeScraperRunStatus` throw path (should-fix)**: taken — `ScraperRunStatusSchema` now has an object-level `.catch()` degrading any non-object response (null, undefined, an error string body, a number) to a complete `never_run` status. `.parse()` is now total. Pinned by a test looping `[null, undefined, 'Internal Server Error', 42]`.

**Unknown-key stripping (should-fix)**: taken — `ScraperConfigSchema` is now `z.looseObject` (Zod v4's passthrough), so unenumerated backend fields survive normalization. You were right about the round-trip: the editor receives a scraper from the normalized list and `saveScraper`s it back, so stripping would have silently shed `created_at`-class fields. Pinned by a test asserting `created_at` and a `future_field` object pass through.

**`FREQUENCY_OPTIONS` value 0 (should-fix/verify)**: confirmed — `constants.ts` has `{ value: 0, label: 'Manual only' }` (it's the first option), and the card test asserting the rendered 'Manual only' text passes. The browser verification in the PR description shows the sparse card rendering it live.

**Warn payload (nit)**: taken — issues only, no raw record.

**`extraction_method` lockstep (nit)**: taken — comment added.

**Blank line (nit)**: fixed.

**`${minutes}m` i18n (nit) — declined for this PR**: the fallback only renders for a *known numeric* frequency that isn't in `FREQUENCY_OPTIONS` — whose labels ('Every 30 minutes', 'Manual only', …) are themselves hardcoded English constants that predate this PR and render on every card today. Localizing the frequency system is a page-wide change (new keys ×8 locales + replacing the constants table) that belongs in its own issue, not a crash-class fix. Happy to file it as a follow-up if maintainers want.

Full gate green: `npm run check` exits 0 (frontend 1649, CDK 33, backend 818).

perrozzi:
The incremental push directly addresses the review feedback: an object-level `.catch()` on `ScraperRunStatusSchema` (with a garbage-input test), `z.looseObject` for round-trip-safe passthrough (with a test asserting unknown keys survive), issues-only warning logging, the lockstep comment on `extraction_method`, and the stray blank line removed.

### Resolved
- `normalizeScraperRunStatus` no longer throws on non-object responses — object-level catch added, tested against `null`/`undefined`/string/number (`scrapersSchema.ts`, `scrapersSchema.test.ts`).
- Unknown-key stripping fixed via `z.looseObject`, with a passthrough test that actually asserts unknown fields (`created_at`, `future_field`) survive.
- `FREQUENCY_OPTIONS` concern effectively confirmed: the `ScraperCard` test asserting `'Manual only'` for `frequency_minutes: 0` covers it.
- Nits addressed: raw payload dropped from `console.warn`, `extraction_method` lockstep comment added, double blank line removed.

### Still open
- Nit only: `getFrequencyLabel`'s `` `${minutes}m` `` fallback and the `—` literal remain non-i18n.
- Coordination note: `mock-server.js` merge-order conflict with #172 (acknowledged in the PR).

### New findings
- Minor/verify: `z.looseObject` is Zod v4 API (v3 uses `.passthrough()`). The reported green `npm run check` implies the repo is on v4, but worth a quick confirmation since `feedbackSchema.ts`/`formSchema.ts` should use the same idiom for consistency (`voc-datalake/frontend/src/api/scrapersSchema.ts`).

Otherwise the delta is clean and well-tested — approve pending the stacked-base (#168) retarget.

---

mundurragacl:
Round 2 — the Zod v4 verification: confirmed, the frontend is on `zod ^4.3.5` (package.json), where `z.looseObject` is the passthrough idiom. Consistency point taken and applied: `formSchema.ts` on #172 now uses the same `looseObject` treatment (its edit modal round-trips normalized records the same way), with its own passthrough test — see #172's latest commit. `feedbackSchema.ts` (the pre-existing one) intentionally strips: feedback items are read-only in the frontend (no save-back path), so stripping DynamoDB GSI attributes there is correct, per its own docstring.

The remaining i18n nit stays declined-as-follow-up per the round-1 rationale. I believe that's everything actionable — ready for the stacked merge (#168 first, this auto-retargets).

PR #176, opened 2026-07-15
fix(scrapers): normalize sparse records + drifted run status (undefinedm)

### LINKED ISSUE #169: [Enhancement]: Align mock-server scraper fixtures with the real ScraperConfig shape

mundurragacl:
Fixes #169

Successor to #175, which GitHub auto-closed when its stacked base (#168's branch) was deleted on merge — same branch, now rebased onto development with #168's squash merged in (conflicts resolved by taking this branch's schema-based versions, which supersede #168's inline map with identical semantics).

**Full review already completed on #175** (2 rounds, converged at 'approve pending the stacked-base retarget'): lenient Zod schemas at the scrapers API boundary (frequency 0 = 'Manual only', base_url '' = not configured, per-item salvage, numeric coercion, id-less records dropped with warning, looseObject round-trip safety, total run-status normalization that never throws), ScraperCard belt-and-braces dash, real-shape mock fixtures with one deliberately sparse record, mutation-verified regression tests at schema/API/component layers.

Post-resolution verification: `tsc --noEmit` clean, 142 tests green across the scrapers surface (schema 13, api 4, card 11, page + editor + helpers).

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -143,11 +143,33 @@ const mockFeedbackForms = [
   { id: 'form_3', name: 'Support Feedback', enabled: false, submissions: 89, created_at: new Date().toISOString() },
 ];
 
-// Mock scrapers
+// Mock scrapers — real ScraperConfig shape (issue #169). scraper_2 is
+// deliberately sparse (identity fields only), mirroring records persisted
+// before newer fields existed, so local dev exercises the normalizeScrapers()
+// boundary: base_url '' renders Not configured, frequency 0 renders
+// 'Manual only' (never 'undefinedm').
 const mockScrapers = [
-  { id: 'scraper_1', name: 'Product Reviews', url: 'https://example.com/reviews', enabled: true, schedule: 'rate(30 minutes)', last_run: new Date().toISOString(), status: 'success' },
-  { id: 'scraper_2', name: 'Forum Posts', url: 'https://forum.example.com', enabled: false, schedule: 'rate(1 hour)', last_run: null, status: 'disabled' },
+  {
+    id: 'scraper_1', name: 'Product Reviews', enabled: true,
+    base_url: 'https://example.com/reviews', urls: ['https://example.com/reviews?sort=recent'],
+    frequency_minutes: 30, extraction_method: 'css',
+    container_selector: '.review', text_selector: '.review-text',
+    rating_selector: '.review-stars@data-rating', author_selector: '.review-author',
+    date_selector: '.review-date',
+    pagination: { enabled: true, param: 'page', max_pages: 3, start: 1 },
+    last_run: new Date().toISOString(), items_found: 42,
+  },
+  // Sparse legacy record: identity fields only.
+  { id: 'scraper_2', name: 'Forum Posts', enabled: false },
 ];
+const mockScraperRuns = {
+  scraper_1: {
+    scraper_id: 'scraper_1', status: 'completed',
+    started_at: new Date(Date.now() - 3600000).toISOString(),
+    completed_at: new Date(Date.now() - 3590000).toISOString(),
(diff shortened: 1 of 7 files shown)


perrozzi:
This PR replaces ad-hoc `base_url` patching in `scrapersApi` with lenient Zod schemas (`scrapersSchema.ts`) that normalize sparse scraper records and drifted run-status shapes at the API boundary, adds a render-safe guard in `ScraperCard`, updates the mock server to the real wire shapes, and backs it all with regression tests at the schema, API, and component layers. This follows the repo's Zod-at-the-boundary convention (cited: `feedbackSchema.ts`, `formSchema.ts`), avoids `as` assertions (nice use of `Reflect.deleteProperty` in the card test), and the warning log deliberately omits raw payloads. Overall a clean, well-tested fix — no blockers found. Note: I can only verify what's in the diff; claims of `tsc --noEmit` passing and Zod v4 (`z.looseObject`) availability are taken on trust.

### Should fix
- **Nested `pagination` breaks the round-trip guarantee the file promises.** `ScraperConfigSchema` uses `z.looseObject` at the top level specifically so "a record read from getScrapers() and saved back by the editor round-trips without silent data loss" (`scrapersSchema.ts` docblock), but `paginationSchema` is a plain `z.object`, which strips unknown keys. Any future backend field nested under `pagination` will be silently shed on edit round-trips. Use `z.looseObject` there too, and extend the pass-through test in `scrapersSchema.test.ts` to cover a nested unknown field.
- **The run-status object-level `.catch` degrades silently.** `normalizeScrapers` warns when it drops a record, but `ScraperRunStatusSchema`'s wholesale catch turns any malformed response into `never_run` with no signal (`scrapersSchema.ts`). If the status endpoint starts returning garbage, the UI will report "never run" for scrapers that did run, with nothing in the console to diagnose it. A `console.warn` in `normalizeScraperRunStatus` when the top-level catch fires (issues only, per the same PII-avoidance rationale) would keep the degrade-don't-throw behavior while staying observable.

### Nits
- `ScraperCard.test.tsx` asserts hardcoded English strings (`'Manual only'`, `'Every 30 minutes'`) while the neighboring #167 tests use `i18n.t(...)`. If `FREQUENCY_OPTIONS` labels are (or become) localized per the repo's i18n rule, these tests break; prefer resolving via i18n keys.
- `toOptionalFiniteNumber` accepts negative values, so `frequency_minutes: -5` or `pages_scraped: -1` pass through unchanged. A `.min(0)` (or clamping in the preprocess) would be consistent with the "conservative defaults" stance.
- `getScraperStatus`'s return type changed: `scraper_id` was required, now optional in `ScraperRunStatus` (`scrapersApi.ts`). The claimed clean `tsc` run should cover consumers, but worth confirming no caller relied on it.
- Mock fixture `rating_selector: '.review-stars@data-rating'` (`mock-server.js`) embeds the attribute in the selector, while the schema models a separate `rating_attribute` field — if the backend uses the split-field shape, the fixture should too, since the whole point of this fixture change is real-shape fidelity.
- `afterEach(vi.restoreAllMocks)` sits at the `normalizeScrapers` describe level in `scrapersSchema.test.ts` but only one test spies; scoping the restore next to the spy (or using `vi.spyOn(...).mockRestore()`) would be tighter.

---

PR #178, opened 2026-07-15
fix(dev): admin UI visible in mock-only local dev

### LINKED ISSUE #177: Local dev (mock-only) hides all admin UI: useIsAdmin lacks the routes' no-Cognito dev bypass
## Problem

In mock-only local dev (no Cognito configured), no admin surface is reachable:

- Settings sidebar link hidden (Layout filters `adminOnly` items on `useIsAdmin()`)
- Users tab hidden on Settings
- AI Models card (#166) hidden

`ProtectedRoute` and `AdminRoute` both have a documented local-dev bypass — when Cognito is not configured and `import.meta.env.DEV` is true, they render children (fail closed in production). But `useIsAdmin()` in `store/authStore.ts` has no such bypass: with no Cognito there's no user, so `groups.includes('admins')` is always false. Result: /settings is reachable by URL (the route opens) but every isAdmin-driven surface stays hidden — an inconsistent half-bypass.

Also: the mock server has no `/users` endpoints, so the Users tab would 404 once visible.

## Proposal

- Mirror the routes' documented bypass in `useIsAdmin`: `import.meta.env.DEV && !cognitoConfigured()` → admin (production without Cognito stays fail-closed, same as the routes). Cycle note: `services/auth.ts` imports the store, so the store reads `runtimeConfig` directly instead of `authService.isConfigured()`.
- Mock `/users` endpoints with CognitoUser-shaped fixtures so the Users tab works locally.
- Regression tests for the bypass and the group-based path.

mundurragacl:
Fixes #177

## Problem

Mock-only local dev (no Cognito) showed no admin surface: the Settings sidebar link, the Users tab, and the AI Models card (#166) were all hidden. `ProtectedRoute`/`AdminRoute` have a documented no-Cognito dev bypass, but `useIsAdmin()` didn't — so /settings opened by URL while every isAdmin-driven surface inside it stayed invisible. An inconsistent half-bypass.

## Security analysis (the important part)

This changes **UI visibility only**; the authorization boundary is unchanged and stays server-side:

1. **Compile-time elimination**: `import.meta.env.DEV` is a Vite build-time constant — in production bundles the bypass branch is statically `false` and dead-code-eliminated. The code does not ship.
2. **Config-gated**: even in a dev build, the bypass is inert whenever Cognito IS configured — group membership governs, pinned by tests.
3. **Fail-closed in production**: a production build without Cognito reports non-admin here and redirects at the routes — identical posture to the existing `ProtectedRoute`/`AdminRoute` bypasses this mirrors.
4. **The real boundary is the backend**: API Gateway's Cognito authorizer rejects unauthenticated calls, and `require_admin` (admins group, verified server-side) gates `/users/*` and `PUT /settings/model`. Client state is attacker-controlled by definition — anyone can flip a boolean in devtools today — which is exactly why no capability is granted by this flag.

Cycle note: `services/auth.ts` imports this store, so the store reads `runtimeConfig` directly instead of `authService.isConfigured()`, with a pre-load short-circuit (config is never read while unloaded — `getRuntimeConfig` throws in that state; pinned by a test asserting it's not called).

## Mock server

Stateful `/users` fixtures (3 CognitoUser-shaped users) + the full user-admin surface: `GET/POST /users`, `PUT /users/{u}` (attributes), `PUT /users/{u}/group|enable|disable`, `POST /users/{u}/reset-password`, `DELETE /users/{u}` — honest 404s for unknown usernames, mutations reflected in the session (`handleUserAction` keeps the dispatch small).

5 new tests in `authStore.test.ts` — bypass active with no user when unconfigured; unloaded config short-circuits without touching `getRuntimeConfig`; configured path stays inert (admins → true, users → false, anonymous → false). Reverting the bypass fails exactly the two bypass tests. Vitest runs as a DEV build, which is precisely the branch under test; the production branch is compile-time eliminated (untestable in jsdom, same documented limitation as the existing route tests).

## Verification

- `npm run check` exits 0 (frontend 1701, CDK 33, backend green) — no existing test flipped by the bypass
- Browser-verified against the restarted mock: Settings link in sidebar, Users tab renders all three fixtures with actions, AI Models card shows all five per-surface pickers with the allowlist

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -219,6 +219,8 @@ const mockUsers = [
 
 // Apply a user-admin mutation (issue #177). Returns { status, payload };
 // keeps the dispatch block small and the mutations in one auditable place.
+// Error payloads use the `error` key, matching the real API's shape
+// ({'success': False, 'error': ...} from lambda/shared/api.py).
 function handleUserAction(method, user, action, body) {
   const touch = () => { user.last_modified = new Date().toISOString(); };
   if (method === 'PUT' && action === 'group') {
@@ -233,7 +235,8 @@ function handleUserAction(method, user, action, body) {
   if (method === 'PUT' && (action === 'enable' || action === 'disable')) {
     user.enabled = action === 'enable';
     touch();
-    return { status: 200, payload: { success: true, message: `User ${action}d` } };
+    const confirmation = { enable: 'User enabled', disable: 'User disabled' };
+    return { status: 200, payload: { success: true, message: confirmation[action] } };
   }
   if (method === 'PUT' && !action) {
     const given = typeof body?.given_name === 'string' ? body.given_name : user.given_name;
@@ -248,7 +251,7 @@ function handleUserAction(method, user, action, body) {
     mockUsers.splice(mockUsers.indexOf(user), 1);
     return { status: 200, payload: { success: true, message: 'User deleted' } };
   }
-  return { status: 405, payload: { success: false, message: 'Method not allowed' } };
+  return { status: 405, payload: { success: false, error: 'Method not allowed' } };
 }
 
 // Mock scrapers — real ScraperConfig shape (issue #169). scraper_2 is
@@ -671,7 +674,7 @@ const server = http.createServer((req, res) => {
       const user = mockUsers.find(u => u.username === username);
       if (!user) {
         res.writeHead(404);
-        res.end(JSON.stringify({ success: false, message: 'User not found' }));
+        res.end(JSON.stringify({ success: false, error: 'User not found' }));
         return;
       }
       let parsedBody = null;
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -194,6 +194,63 @@ const mockFormStats = {
   form_4: { total_submissions: 41, avg_rating: 4.6, rating_count: 38 },
 };
 
+// Mock Cognito users (issue #177) — stateful so create/enable/disable/delete
+// reflect in the UI during a dev session. Shape mirrors CognitoUser.
+const mockUsers = [
+  {
+    username: 'admin-demo', email: 'admin@example.com', name: 'Ada Admin',
+    given_name: 'Ada', family_name: 'Admin', status: 'CONFIRMED', enabled: true,
+    groups: ['admins'], created_at: new Date(Date.now() - 30 * 86400000).toISOString(),
+    last_modified: new Date().toISOString(),
+  },
+  {
+    username: 'viewer-demo', email: 'viewer@example.com', name: 'Vic Viewer',
+    given_name: 'Vic', family_name: 'Viewer', status: 'CONFIRMED', enabled: true,
+    groups: ['users'], created_at: new Date(Date.now() - 10 * 86400000).toISOString(),
+    last_modified: new Date().toISOString(),
+  },
+  {
+    username: 'disabled-demo', email: 'disabled@example.com', name: 'Dee Disabled',
+    given_name: 'Dee', family_name: 'Disabled', status: 'CONFIRMED', enabled: false,
+    groups: ['users'], created_at: new Date(Date.now() - 60 * 86400000).toISOString(),
+    last_modified: new Date().toISOString(),
+  },
+];
+
+// Apply a user-admin mutation (issue #177). Returns { status, payload };
+// keeps the dispatch block small and the mutations in one auditable place.
+function handleUserAction(method, user, action, body) {
+  const touch = () => { user.last_modified = new Date().toISOString(); };
+  if (method === 'PUT' && action === 'group') {
+    const group = body?.group === 'admins' ? 'admins' : 'users';
+    user.groups = [group];
+    touch();
+    return { status: 200, payload: { success: true, message: `Moved ${user.username} to ${group}` } };
+  }
+  if (method === 'POST' && action === 'reset-password') {
+    return { status: 200, payload: { success: true, message: 'Password reset initiated' } };
+  }
+  if (method === 'PUT' && (action === 'enable' || action === 'disable')) {
+    user.enabled = action === 'enable';
+    touch();
+    return { status: 200, payload: { success: true, message: `User ${action}d` } };
+  }
+  if (method === 'PUT' && !action) {
+    const given = typeof body?.given_name === 'string' ? body.given_name : user.given_name;
+    const family = typeof body?.family_name === 'string' ? body.family_name : user.family_name;
+    user.given_name = given;
+    user.family_name = family;
+    user.name = [given, family].filter(Boolean).join(' ') || user.username;
+    touch();
+    return { status: 200, payload: { success: true, message: 'User updated', given_name: given, family_name: family, name: user.name } };
+  }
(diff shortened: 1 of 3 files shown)


perrozzi:
This PR adds the documented no-Cognito dev bypass to `useIsAdmin()` (mirroring `ProtectedRoute`/`AdminRoute`), plus stateful user-admin fixtures/routes in the mock server and five regression tests. The security reasoning in the description is sound: the bypass is UI-visibility only, `import.meta.env.DEV` is dead-code-eliminated in production builds, and the real authorization boundary (`require_admin` + the Cognito authorizer) is untouched. Overall this is a clean, well-tested change; a couple of points below are worth confirming.

### Should fix

- **`__status` convention in `mock-server.js`** — the new `POST /users` handler returns `{ __status: 400, body: {...} }` / `{ __status: 409, ... }` for error paths, but the generic handler dispatch isn't shown in this diff. If the existing dispatcher doesn't already understand the `__status` envelope, these errors will be serialized as HTTP 200 with a wrapper object as the body, and the frontend's duplicate/validation handling won't exercise real error paths. Please confirm the dispatcher unwraps this (or align with however other handlers return non-200s, e.g. `GET /users/:username`-style routes here use `res.writeHead(status)` directly).
- **Unloaded config fails open in DEV** (`authStore.ts`) — `cognitoConfigured()` treats "config not loaded" as "not configured", so a DEV build pointed at real Cognito reports admin during the pre-load window, and since the helper isn't reactive to config load it won't self-correct until a store update triggers a re-render. The comment says App gates rendering on config load, which makes this moot in practice — but consider a comment-level pointer to that gate (or asserting it in a test) so a future refactor of App's loading gate doesn't silently reintroduce an admin flash in dev-against-real-Cognito setups. DEV-only, so not blocking.

### Nits

- `mock-server.js`: error payload keys are inconsistent — `POST /users` uses `error`, the 404/405 paths use `message`. Pick one to match the real API shape so frontend error-display code behaves the same against mock and real backends.
- `mock-server.js`: `GET /users/{username}` falls through to `handleUserAction`'s 405. If the real API supports fetching a single user, mirror it; if not, this is fine.
- `handleUserAction`'s enable/disable message interpolates to `"User disabled"` but also `"User enabled"` via `` `User ${action}d` `` → `"enabled"`/`"disabled"` — actually correct, but fragile string-building; a small map would be clearer.
- `authStore.test.ts`: `stubCognito(false)` models "not configured" as empty-string fields — worth one test (or the existing unloaded-config test suffices) covering a config with the `cognito` key absent, if the runtime config type allows it, since `cfg.cognito.userPoolId` would throw.
- The regression tests only pin the DEV branch (acknowledged in the PR). Fine given Vite's compile-time elimination, but the JSDoc on `useIsAdmin` is the only place stating the production posture — keep that comment in sync if the routes' bypass ever changes.

No blockers. Repo patterns are respected: no `as` assertions, no new user-visible strings needing i18n, and the fix ships mutation-verified regression tests that fail when the bypass is reverted.

---

perrozzi:
The push is a polish pass responding to the previous review: the expanded JSDoc on `cognitoConfigured()` now documents the App.tsx loading-gate dependency, the mock server's error payloads were normalized to the `error` key (with a comment citing `lambda/shared/api.py`), and the fragile `` `User ${action}d` `` interpolation was replaced with an explicit map. No behavioral changes to the bypass logic or tests.

### Resolved
- Pre-load fail-open risk is now documented in `authStore.ts` with an explicit pointer to App's `configReady` gate and the pinning test — addresses the second should-fix.
- Error payload key inconsistency (`message` vs `error`) fixed on the 404 and 405 paths in `mock-server.js`, matching the real API shape.
- Enable/disable confirmation string-building replaced with an explicit `confirmation` map.

### Still open
- **`__status` envelope in `POST /users`** — still unconfirmed whether the generic handler dispatch unwraps `{ __status, body }` or serializes it as a 200; the incremental diff didn't touch the dispatcher or the `POST /users` handler.
- Nit: no test covering a runtime config where the `cognito` key is absent (if the type allows it), since `cfg.cognito.userPoolId` would throw.

### New findings
None — the incremental changes are comment/consistency fixes and are clean. Note the mixed payload convention is now intentional and documented: `error` for failures, `message` for successes, which matches the stated real API shape.

---

mundurragacl:
Round 1 addressed in f42e3ee — one should-fix was a verify (confirmed with line refs), one taken, nits taken or answered.

**`__status` envelope (should-fix/verify)**: confirmed — the generic dispatcher already unwraps it (mock-server.js: `if (result && typeof result.__status === 'number') { res.writeHead(result.__status); ... }`), and the convention is pre-existing (used by `PUT /settings/resolved-problems` from #163 and `PUT /settings/model` from #166). The new `POST /users` 400/409 paths serialize as real HTTP errors. Verified live: `curl -X POST` with a duplicate username returns HTTP 409.

**Pre-load fail-open contract (should-fix)**: taken — `cognitoConfigured()`'s doc comment now names the exact invariant (App.tsx gates `<RouterProvider>` on `loadRuntimeConfig()` resolving via `configReady`), spells out the consequence if that gate is refactored away (brief admin flash in a DEV build against real Cognito), and points at the test that pins the short-circuit (`getRuntimeConfig` must not be called while unloaded).

**Error-key inconsistency (nit)**: taken — all mock error payloads now use `error`, matching the real API's `{'success': False, 'error': ...}` shape from `lambda/shared/api.py`.

**`GET /users/{username}` (nit)**: checked the real handler — `users_handler.py` has no single-user GET (list + mutations only), so the mock's 405 fall-through mirrors reality. Leaving as-is.

**Enable/disable string-building (nit)**: taken — small confirmation map.

**Absent-`cognito`-key test (nit)**: not representable — `RuntimeConfig` is `z.infer` of a schema where `cognito` is a required object, so `getRuntimeConfig()` can never return a config without it (the Zod parse would have failed at load and taken the env-fallback path, which always constructs the block). A test would need an `as` cast to build the impossible value, which this repo lints against. The unloaded-config test covers the reachable degenerate case.

**JSDoc as the production-posture record (nit)**: agreed — it's now the fuller contract comment from the should-fix above, and the PR description carries the same analysis for the record.

Gate green: `npm run check` exits 0 (frontend 1701, CDK 33, backend green).

mundurragacl:
Round 2 — closing out the remaining verify with live evidence, since the dispatcher predates this PR and sits outside the diff:

**`__status` envelope**: the generic dispatcher (mock-server.js, in the `req.on('end')` handler) reads:

```js
// Handlers may signal a non-200 status (matching the real API's error
// contract) by returning { __status, body }.
if (result && typeof result.__status === 'number') {
  res.writeHead(result.__status);
  res.end(JSON.stringify(result.body));
  return;
}
```

Live proof against the running mock at this PR's HEAD:

```
POST /users {username: 'admin-demo', ...}  → HTTP 409  {"success":false,"error":"User already exists"}
POST /users {email only}                   → HTTP 400  {"success":false,"error":"username is required"}
```

Real HTTP statuses, not 200-wrapped envelopes.

**Absent-`cognito`-key test (nit)**: per the round-1 reply — `RuntimeConfig` is `z.infer` of a schema with `cognito` required, so that value is unconstructible without an `as` cast (linted `assertionStyle: never`). The reachable degenerate case (config not loaded) is tested, including that `getRuntimeConfig` is never called in that state.

I believe that's everything — no open items from my side.

PR #182, opened 2026-07-15
fix(categories): survive legacy rows without id/subcategories

### LINKED ISSUE #181: [Bug]: Settings Categories tab crashes on categories without subcategories (undefined .length)
## Problem

Settings → Categories crashes with `Cannot read properties of undefined (reading 'length')` in `<CategoriesManager>` (contained by the #174 route error boundary, so the rest of the app survives).

Root cause is the recurring type-vs-runtime class (#167/#169/#171): the component reads `category.subcategories.length` (and maps over it) as required per the declared type, but nothing enforces that shape at runtime:

- the real `GET /settings/categories` returns the DynamoDB item verbatim (`item.get('categories', [])`) — legacy or hand-written rows can lack `id`/`subcategories`
- the local mock serves exactly such a legacy shape (`{name, display_name, color}`), which is how it surfaced

## Proposal (same architecture as #172/#176)

- Lenient Zod schema at the query boundary: `subcategories` per-item salvage defaulting to `[]`; missing `id` **derived** from `name` (stable, deterministic) rather than dropped — categories are user-editable config, and dropping rows would silently delete them on the next save; loose objects so legacy fields (`display_name`, `color`) survive save round-trips; drop only when both `id` and `name` are unusable
- Belt-and-braces guard at the render sites so the component stays render-safe standalone
- Mock aligned to the real contract shape, keeping one deliberately sparse legacy record to exercise the normalizer in dev
- Mutation-verified regression tests at schema and component layers

mundurragacl:
Fixes #181

## Problem

Settings → Categories crashed with `Cannot read properties of undefined (reading 'length')` in `<CategoriesManager>` — contained by #174's route error boundary, but the tab itself was dead. The component reads `category.subcategories.length`/`.map` as required by the declared type, while the real `GET /settings/categories` returns the DynamoDB item **verbatim** (`item.get('categories', [])` in settings_handler.py) — so legacy rows shaped `{name, display_name, color}` have no `id` and no `subcategories`. The local mock served exactly that legacy shape, which is how it surfaced. Same class as #167/#169/#171.

## Fix (same architecture as #172/#176, one deliberate difference)

- **`categoriesSchema.ts`** — lenient Zod normalization at the query boundary. **Identity strategy differs from formSchema/scrapersSchema on purpose**: categories are user-editable config that the save flow round-trips wholesale, so dropping an id-less row would silently delete it on the next save. Missing ids are **derived** from names (stable slug — `cat_app`), deterministic across loads; a row is dropped only when both id and name are unusable. `looseObject` keeps legacy fields (`display_name`, `color`) intact through round-trips; subcategories salvage per item with derived sub ids.
- **`CategoriesManager`** — `select` at the query boundary; the **LLM generate response is normalized before saving too** (it's a wire boundary as much as the GET); belt-and-braces `?? []` at the render and mutation sites so the component stays render-safe standalone.
- **mock-server** — categories now carry the real contract shape (id + subcategories), keeping one deliberately sparse legacy row so local dev permanently exercises the normalizer.

- `categoriesSchema.test.ts` (7): derived id + empty subcategories for the legacy shape, loose passthrough (`display_name`/`color` survive), complete-row passthrough, explicit-null subcategories, per-item salvage with derived sub ids, deterministic derivation, drop-only-when-both-unusable (with warn count)
- `CategoriesManager.test.tsx` pin: renders the exact crash shape to **'0 sub'** — reverting the boundary + guard fails exactly this test (verified)

## Verification

- Browser: Settings → Categories renders 6 categories including the legacy row ('App experience', 0 sub), no error boundary
- `npm run check` exits 0 (frontend 1709, CDK 33, backend green)

### DIFF
--- a/voc-datalake/frontend/mock-server.js
+++ b/voc-datalake/frontend/mock-server.js
@@ -142,15 +142,35 @@ const mockSurfaceDefaults = {
   utility: 'global.anthropic.claude-sonnet-5',
 };
 
+// Categories config — real contract shape (id + subcategories, issue #181).
+// The last entry is a deliberately sparse legacy row ({name, display_name,
+// color} only — what old DynamoDB rows carry) so local dev permanently
+// exercises the normalizeCategories() boundary.
 const mockCategoriesConfig = {
   categories: [
-    { name: 'delivery', display_name: 'Delivery', description: 'Shipping and delivery issues', color: '#EF4444' },
-    { name: 'customer_support', display_name: 'Customer Support', description: 'Support interactions', color: '#3B82F6' },
-    { name: 'product_quality', display_name: 'Product Quality', description: 'Quality concerns', color: '#F59E0B' },
-    { name: 'pricing', display_name: 'Pricing', description: 'Price-related feedback', color: '#10B981' },
-    { name: 'website', display_name: 'Website', description: 'Website experience', color: '#8B5CF6' },
+    {
+      id: 'cat_delivery', name: 'delivery', description: 'Shipping and delivery issues',
+      subcategories: [
+        { id: 'sub_late_delivery', name: 'late_delivery', description: 'Late deliveries' },
+        { id: 'sub_tracking', name: 'tracking', description: 'Tracking visibility' },
+      ],
+    },
+    {
+      id: 'cat_customer_support', name: 'customer_support', description: 'Support interactions',
+      subcategories: [
+        { id: 'sub_response_time', name: 'response_time', description: 'Slow responses' },
+      ],
+    },
+    {
+      id: 'cat_product_quality', name: 'product_quality', description: 'Quality concerns',
+      subcategories: [],
+    },
+    { id: 'cat_pricing', name: 'pricing', description: 'Price-related feedback', subcategories: [] },
+    { id: 'cat_website', name: 'website', description: 'Website experience', subcategories: [] },
+    // Sparse legacy row: no id, no subcategories.
     { name: 'app', display_name: 'Mobile App', description: 'App experience', color: '#EC4899' },
-  ]
+  ],
+  updated_at: new Date().toISOString(),
 };
 
 // Mock feedback forms — real wire shape (form_id, full FeedbackFormFields).
--- a/voc-datalake/frontend/src/components/CategoriesManager/CategoriesManager.test.tsx
+++ b/voc-datalake/frontend/src/components/CategoriesManager/CategoriesManager.test.tsx
@@ -980,3 +980,41 @@ describe('CategoriesManager', () => {
     })
   })
 })
+
+
+describe('sparse legacy rows (issue #181)', () => {
+  let queryClient: QueryClient
+
+  beforeEach(() => {
+    vi.clearAllMocks()
+    queryClient = new QueryClient({
+      defaultOptions: { queries: { retry: false, gcTime: 0, staleTime: 0 } },
+    })
+    mockSaveCategoriesConfig.mockResolvedValue({ success: true })
+  })
+
+  function renderComponent() {
+    return render(
+      <QueryClientProvider client={queryClient}>
+        <CategoriesManager />
+      </QueryClientProvider>
+    )
+  }
+
+  it('renders a legacy row without id/subcategories instead of crashing the tab', async () => {
+    // Exactly what the wire delivered when Settings → Categories crashed:
+    // old DynamoDB rows carry {name, display_name, color} only.
+    mockGetCategoriesConfig.mockResolvedValue({
+      categories: [
+        { name: 'app', display_name: 'Mobile App', description: 'App experience', color: '#EC4899' },
+      ],
+    })
+
+    renderComponent()
+
+    await waitFor(() => {
+      expect(screen.getByText('App experience')).toBeInTheDocument()
+    })
+    expect(screen.getByText('0 sub')).toBeInTheDocument()
+  })
+})
--- a/voc-datalake/frontend/src/components/CategoriesManager/CategoriesManager.tsx
+++ b/voc-datalake/frontend/src/components/CategoriesManager/CategoriesManager.tsx
@@ -18,6 +18,7 @@ import {
 } from 'lucide-react'
 import { api } from '../../api/client'
 import ConfirmModal from '../ConfirmModal'
+import { normalizeCategories } from './categoriesSchema'
 
 export interface Category {
   id: string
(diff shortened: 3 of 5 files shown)


perrozzi:
This PR fixes a crash in the Settings → Categories tab caused by legacy DynamoDB rows lacking `id`/`subcategories`, using the same query-boundary Zod normalization pattern as the sibling schemas, with a deliberate (and well-justified) id-derivation strategy to avoid silently deleting user config on save round-trips. The change is well-scoped, follows the repo's Zod-at-the-boundary and TanStack Query `select` patterns, and includes mutation-verified regression tests. Overall this looks solid; findings below are non-blocking.

### Should fix

- **Derived-id collisions** (`categoriesSchema.ts`): `deriveId('cat', name)` slugs only on whitespace/case, so two legacy rows named `App` and `app` (or `app experience` / `App  Experience`) both derive `cat_app`/`cat_app_experience`. Since the component keys, expands, edits, and deletes by `id` (`c.id === categoryId` in `handleDeleteCategory`/`handleAddSubcategory`), colliding ids would make one row's action affect the other, plus duplicate React keys. Consider de-duplicating derived ids within `normalizeCategories` (e.g. suffix `_2`) and adding a test for two same-named legacy rows. Same applies to derived subcategory ids within a category.
- **Whitespace-only names slip through the drop guard** (`categoriesSchema.ts`): `name === ''` doesn't catch `name: '  '`, which yields `id: 'cat_'` (and all such rows collide on it). Trim before the usability check.

### Nits

- `deriveId` doesn't sanitize non-word characters — `name: 'billing/refunds'` yields `cat_billing/refunds`. Harmless functionally, but a `[^a-z0-9]+ → _` replace would give cleaner, more collision-resistant slugs.
- `CategoriesManager.tsx`: since `Category.subcategories` is declared non-optional, the belt-and-braces `c.subcategories ?? []` may trip lint rules about unnecessary nullish coalescing on non-nullable types; if your ESLint config flags it, a small type-level `subcategories?: Subcategory[]` note or targeted disable comment would keep `npm run check` stable.
- `mock-server.js`: `updated_at: new Date().toISOString()` is evaluated once at module load, so it's static across requests — fine for a mock, just noting it isn't per-request.
- `CategoriesManager.test.tsx` asserts the hardcoded English string `'0 sub'`; that string predates this PR but is user-visible and not behind an i18n key — worth a follow-up issue given the repo's i18n rule, rather than expanding this PR.
- `console.warn` on dropped rows logs the full raw row (`parsed.data`); category config is unlikely to contain PII, but consider logging just the offending keys to stay conservative.

---

mundurragacl:
Round 1 addressed in 97077ae — both should-fixes taken with tests, nits taken or answered.

**Derived-id collisions (should-fix)**: taken — `uniqueId()` reserves ids in a per-list set, so collisions suffix deterministically (`cat_app`, `cat_app_2`), **including a derived id colliding with a stored one** (test: stored `cat_app` + legacy `app` → `cat_app_2`). Sub ids de-dupe within their category. Pinned by 3 new tests, including your exact `App`/`app` case.

**Whitespace-only names (should-fix)**: taken — identity is now computed via `identityFor()`: stored ids trimmed, names slugified *before* the usability check, so `'   '` (and symbol-only `'///'`) are dropped instead of colliding on `cat_`. Tested.

**Slug sanitization (nit)**: taken — `slugify()` collapses non-alphanumerics (`'billing/refunds & credits'` → `cat_billing_refunds_credits`). Tested.

**`?? []` lint concern (nit)**: verified safe — this repo's ESLint uses `tseslint.configs.recommended`, which doesn't include `no-unnecessary-condition` (that's strictTypeChecked); `npm run check` is green with the guards. Same pattern already shipped in FormCard (#172) and ScraperCard (#176).

**Warn payload (nit)**: already tightened in this push — the drop warning logs `Object.keys(parsed.data)` only.

**Mock `updated_at` static (nit)**: acknowledged — module-load evaluation is fine for a fixture; noting only.

**'0 sub' i18n (nit)**: agreed it's a pre-existing gap and out of scope here — filed as a follow-up: the hardcoded CategoriesManager strings now have their own issue.

Full gate green: `npm run check` exits 0 (frontend 1713, CDK 33, backend green).

perrozzi:
The push addresses all of my previous should-fix items and most nits: derived ids are now de-duplicated via `uniqueId` (categories and per-category subcategories), `slugify` sanitizes non-alphanumerics and rejects whitespace/symbol-only names through `identityFor`, and the drop-warning now logs only key names instead of the raw row. All fixes come with corresponding tests.

### Resolved
- Derived-id collisions across same-named legacy rows (categories and subcategories) — `uniqueId` + tests.
- Whitespace-only names slipping past the drop guard — `identityFor`/`slugify` + test.
- Non-word characters in derived slugs — sanitized + test.
- `console.warn` logging full raw row — now logs `Object.keys` only.

### New findings
- **Order-dependent rewriting of stored ids** (`categoriesSchema.ts`): `uniqueId` treats stored and derived ids identically on a first-come basis. If a legacy row whose name slugs to `cat_app` appears *before* a row with stored `id: 'cat_app'`, the legacy row claims `cat_app` and the stored row is silently rewritten to `cat_app_2` — and since the save flow round-trips wholesale, that rewrite persists to DynamoDB. Consider a two-pass approach (reserve all non-empty stored ids first, then derive), so stored identities are never mutated. The existing "derived vs stored" test only covers the stored-first ordering; add the reverse ordering as a test either way. Low likelihood in practice, so non-blocking, but it's the one edge the new dedup logic introduces.

Otherwise the incremental changes are clean and well-tested; remaining unaddressed items from my last review were nits only (mock `updated_at` staleness, hardcoded `'0 sub'` i18n follow-up), fine to leave as-is.

---

mundurragacl:
Round 2 addressed in the latest commit — good catch, taken.

**Order-dependent stored-id rewriting**: fixed with the two-pass approach you suggested — `createIdAllocator` reserves every usable **stored** id up front, and derivation only allocates from what remains, so a stored identity can never be rewritten regardless of list position. Duplicate stored ids (pathological but possible in hand-edited rows) still suffix deterministically. Pinned by the reverse-order test: `[{name:'app'}, {id:'cat_app', ...}]` → `['cat_app_2', 'cat_app']` — the stored row keeps its id.

Gate green: `npm run check` exits 0 (frontend 1714, CDK 33, backend green).

perrozzi:
The push directly addresses my remaining finding: id allocation is now two-pass, with all usable stored ids reserved before any derivation, so a derived id can never claim (and persist a rewrite of) a stored identity that appears later in the list. The reverse-ordering test I asked for was added.

### Resolved
- Order-dependent rewriting of stored ids — `createIdAllocator` reserves stored ids up front; `derived()` checks both `assigned` and `reservedStoredIds`; covered by the new "never rewrites a stored id, regardless of list order" test in `categoriesSchema.test.ts`.

### New findings
None blocking. One observation, fine to leave: with pathological *duplicate* stored ids (e.g. two rows both `id: 'cat_app'`), the second is suffixed to `cat_app_2` and that rewrite will persist on the next wholesale save — unavoidable given the uniqueness requirement, and the code comment already acknowledges it. The refactor (`allocateIdentity`, `normalizeSubcategories` collecting parsed subs before allocating) is behavior-preserving and the logic is sound.

The incremental changes are clean and well-tested. Previously noted nits (hardcoded `'0 sub'` i18n follow-up) remain deferred as agreed.

---

PR #185, opened 2026-07-15
fix(cdk): pre-#105 stacks can deploy again — UsernameConfiguration is create-only

### LINKED ISSUE #184: [Bug]: VocCoreStack update fails on pre-#105 deployments: UsernameConfiguration is create-only on Cognito
## Problem

Deploying current development onto a stack created before #105 rolls back VocCoreStack:

```
Resource handler returned message: "Invalid request provided: Updates are not allowed for property - UsernameConfiguration."
```

#105 added `signInCaseSensitive: false` to the User Pool. Cognito treats `UsernameConfiguration` as create-only, so ANY update that introduces it on an existing pool is rejected and the whole stack rolls back (observed live on a production deployment; rollback completed cleanly, no user impact). A secondary `Invalid AttributeDataType` error fires on the same update attempt — the known Cognito CFN quirk when a pool update resends standard Schema attributes — and disappears once the pool isn't updated at all.

## Proposal

Context flag (e.g. `omitUserPoolUsernameConfiguration: true`) that omits the property so pre-#105 stacks keep deploying; greenfield default unchanged (case-insensitive sign-in). CDK test pinning both template shapes.

mundurragacl:
Fixes #184

## Problem (hit live during a production deployment)

Deploying current `development` onto a VocCoreStack created **before #105** rolls the stack back:

```
Invalid request provided: Updates are not allowed for property - UsernameConfiguration.
```

#105 added `signInCaseSensitive: false` to the User Pool, which synthesizes `UsernameConfiguration` — a property Cognito treats as **create-only**. Any stack update that introduces it on an existing pool is rejected wholesale. A secondary `Invalid AttributeDataType` error fires on the same rejected update (the known Cognito-CFN quirk when a pool update resends standard Schema attributes) and disappears once the pool isn't updated at all — one root cause, two symptoms.

Replacement-safety was verified before any deploy attempt via a prepared change set (`Replacement: False` on the pool — no user-data risk from the attempt; the rollback completed cleanly).

## Fix

- `omitUserPoolUsernameConfiguration` context flag (accepts boolean `true` from cdk.context.json or string `'true'` from `-c`): omits the property so pre-#105 pools keep deploying. Greenfield behavior unchanged — new pools still get case-insensitive sign-in.
- `cdk.context.json` sets the flag for this repo's deployed environment.
- New `core-stack.test.ts` pins all three template shapes (default present / boolean flag omits / CLI-string flag omits).

## Verification

- Live: VocCoreStack went `UPDATE_ROLLBACK_COMPLETE` → `UPDATE_COMPLETE` with the flag; all four stacks + frontend then deployed cleanly
- Post-deploy validation on the live environment: 7/7 authenticated API endpoints 200 (including the new `/settings/model`), SSE `/chat/stream` returns live Bedrock tokens end-to-end, site serves with zero console errors
- `npm run check` exits 0; `test:cdk` includes the 3 new template tests

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -457,6 +457,25 @@ npm run cdk:bootstrap
 aws cloudformation continue-update-rollback --stack-name STACK_NAME
 ```
 
+### VocCoreStack fails: "Updates are not allowed for property - UsernameConfiguration"
+
+User pools created before #105 predate the `signInCaseSensitive: false`
+setting, and Cognito treats `UsernameConfiguration` as create-only — any
+stack update that introduces it on an existing pool is rejected and the
+whole VocCoreStack update rolls back. (A secondary
+`Invalid AttributeDataType` error on the same update is a symptom of the
+same rejected pool update.)
+
+Deploy with the compatibility flag to leave the existing pool untouched:
+
+```bash
+cdk deploy VocCoreStack -c omitUserPoolUsernameConfiguration=true
+```
+
+Add the flag to every subsequent deploy of that environment (or to a
+local, uncommitted context override). New deployments should NOT set it —
+fresh pools get case-insensitive sign-in by default.
+
 ### CloudFront Cache
 
 If changes don't appear after deployment:
--- a/voc-datalake/cdk.context.json
+++ b/voc-datalake/cdk.context.json
@@ -3,6 +3,7 @@
     34892
   ],
   "environment": "production",
+  "omitUserPoolUsernameConfiguration": true,
   "brandName": "VoC Analytics",
   "pluginStatus": {
     "webscraper": true,
--- a/voc-datalake/cdk.context.json
+++ b/voc-datalake/cdk.context.json
@@ -3,7 +3,6 @@
     34892
   ],
   "environment": "production",
-  "omitUserPoolUsernameConfiguration": true,
   "brandName": "VoC Analytics",
   "pluginStatus": {
     "webscraper": true,
--- a/voc-datalake/lib/stacks/core-stack.test.ts
+++ b/voc-datalake/lib/stacks/core-stack.test.ts
@@ -48,6 +48,9 @@ describe('VocCoreStack UserPool UsernameConfiguration (issue #184)', () => {
 
     const poolProps = Object.values(template.findResources('AWS::Cognito::UserPool'))
       .map((p) => p.Properties ?? {});
+    // Guard against a vacuous pass: the pool must exist for the absence
+    // assertion below to mean anything.
+    expect(poolProps).toHaveLength(1);
     expect(poolProps[0]).not.toHaveProperty('UsernameConfiguration');
   });
 });
--- a/voc-datalake/lib/stacks/core-stack.test.ts
+++ b/voc-datalake/lib/stacks/core-stack.test.ts
@@ -1,0 +1,53 @@
+/**
+ * Template-level tests for the Cognito User Pool's UsernameConfiguration.
+ *
+ * Regression guard for issue #184: signInCaseSensitive (#105) maps to
+ * UsernameConfiguration, which Cognito treats as create-only — introducing
+ * it on a pool deployed before #105 fails the entire VocCoreStack update
+ * ("Updates are not allowed for property - UsernameConfiguration").
+ * Pre-#105 environments deploy with `-c omitUserPoolUsernameConfiguration=true`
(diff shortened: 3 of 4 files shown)


perrozzi:
This PR fixes a real deployment blocker: `signInCaseSensitive: false` (from #105) synthesizes the create-only `UsernameConfiguration` property, which Cognito rejects on updates to pools created before #105. The fix adds a context flag to omit the property, plus three template-level regression tests that would fail if the flag support were reverted. The core-stack change and tests are well done; the main concern is committing the flag into the shared `cdk.context.json`.

### Blockers
- **`voc-datalake/cdk.context.json` — committing `"omitUserPoolUsernameConfiguration": true` changes greenfield behavior for everyone.** This is a sample repository; anyone cloning and deploying fresh now gets a User Pool *without* case-insensitive sign-in, silently reverting #105 for all new deployments. This directly contradicts the PR's own claim that "Greenfield behavior unchanged — new pools still get case-insensitive sign-in" (and the first test, which asserts the greenfield default, only passes because tests construct their own `cdk.App` context and don't read `cdk.context.json`). The flag is correct for *your* pre-#105 environment, but it should be applied via `-c omitUserPoolUsernameConfiguration=true` at deploy time or a non-committed local context override — not baked into the repo default. Please remove it from the committed `cdk.context.json`.

### Should fix
- **Document the flag.** Operators hitting `Updates are not allowed for property - UsernameConfiguration` on pre-#105 stacks have no way to discover this escape hatch. Add a note to the deployment README/troubleshooting docs (upgrade path from pre-#105 stacks) alongside the code comment in `core-stack.ts`.

### Nits
- `core-stack.test.ts`: the third test (`string form of the flag`) reads `poolProps[0]` without first asserting a pool exists — if `findResources` returned nothing, `expect(undefined).not.toHaveProperty(...)` passes vacuously. Add the same `toHaveLength(1)` assertion used in the second test.
- `core-stack.ts`: consider extracting the `raw === true || raw === 'true'` context-boolean coercion into a small shared helper if other context flags (e.g. `skipFrontendBuildCheck`) need the same string/boolean handling — CLI `-c` always passes strings, so this footgun will recur.

Otherwise the change is clean: the spread-based omission is the right way to keep CDK from synthesizing the property, and the three template tests pin exactly the shapes that matter.

---

perrozzi:
Since the last review, the author removed the committed flag from `cdk.context.json`, added a troubleshooting section to `docs/deployment.md` explaining the flag and the upgrade path, and hardened the third test with a `toHaveLength(1)` guard. All previous findings were addressed except the optional context-coercion-helper nit.

### Resolved
- Blocker: `omitUserPoolUsernameConfiguration` removed from the committed `voc-datalake/cdk.context.json` — greenfield deployments retain case-insensitive sign-in.
- Should-fix: flag documented in `docs/deployment.md` troubleshooting, including the exact error string, deploy command, and the "keep the flag for every subsequent deploy / don't set on new environments" guidance.
- Nit: third test in `core-stack.test.ts` now asserts `poolProps` has length 1 before the absence check, preventing a vacuous pass.

### Still open
- (Nit only) Shared context-boolean coercion helper for `raw === true || raw === 'true'` — fine to defer; not blocking.

### New findings
None. The doc placement in the existing troubleshooting section is appropriate, and the incremental changes are clean. LGTM.

---

mundurragacl:
Round 1 addressed in a099c10 — blocker taken, should-fix taken, first nit taken, second answered.

**Committed context flag (blocker)**: you're right, and it contradicted my own PR text — reverted. `cdk.context.json` no longer carries the flag; greenfield clones keep case-insensitive sign-in. The affected environment deploys with `-c omitUserPoolUsernameConfiguration=true` (already applied to the live stack before this round; the flag is additive so subsequent context-less synths of OTHER stacks are unaffected).

**Documentation (should-fix)**: added a troubleshooting entry to `docs/deployment.md` with the exact error text, the pre-#105 root cause (including the secondary `Invalid AttributeDataType` symptom), the deploy command, and explicit guidance that new deployments must NOT set the flag.

**Vacuous third test (nit)**: taken — asserts `toHaveLength(1)` before the absence check, matching the second test.

**Context-boolean helper (nit) — deferred**: agreed the string/boolean coercion footgun recurs (`skipUseCaseSubmission` in bin/voc-datalake.ts does the same dance), but a shared `contextFlag()` helper touches three call sites across bin/ and lib/ and belongs in its own small refactor rather than a deployment-blocker fix. Happy to file it if you want it tracked.

Tests green (3/3 standalone, 36/36 CDK suite).

PR #186, opened 2026-07-15
fix(stream): unblock cold npm install — eslint-plugin-promise 7.3.0

mundurragacl:
## Problem (hit live on the deployment path)

A cold `npm install` in `voc-datalake/lambda/stream` fails with ERESOLVE: the lockfile resolves `eslint-plugin-promise@7.2.1`, whose peer range is `eslint ^7||^8||^9`, while the project pins `eslint ^10.1.0`. Anyone cloning fresh (contributor or CI) can't install the stream workspace.

## Fix

Lockfile-only refresh to `eslint-plugin-promise@7.3.0` — already within the declared `^7.2.1` range; 7.3.0 adds the `eslint ^10` peer. No package.json changes.

## Verification

- `npm install` clean from cold
- stream suite: 230/230
- CDK bundling of the stream Lambda unaffected (deployed to production this session)

Known-and-separate: `npm run lint` in stream surfaces 176 pre-existing vitest-plugin rule violations in test files (not run by the root gate) — that's rule-adoption drift unrelated to this lockfile fix, happy to file an issue if maintainers want it tracked.

### DIFF
--- a/voc-datalake/lambda/stream/package-lock.json
+++ b/voc-datalake/lambda/stream/package-lock.json
@@ -3355,9 +3355,9 @@
       }
     },
     "node_modules/eslint-plugin-promise": {
-      "version": "7.2.1",
-      "resolved": "https://registry.npmjs.org/eslint-plugin-promise/-/eslint-plugin-promise-7.2.1.tgz",
-      "integrity": "sha512-SWKjd+EuvWkYaS+uN2csvj0KoP43YTu7+phKQ5v+xw6+A0gutVX2yqCeCkC3uLCJFiPfR2dD8Es5L7yUsmvEaA==",
+      "version": "7.3.0",
+      "resolved": "https://registry.npmjs.org/eslint-plugin-promise/-/eslint-plugin-promise-7.3.0.tgz",
+      "integrity": "sha512-6uGiOR0INuujr6PEQmeSSP7GbIMJ/ebEXXiEzb/nOj68LknH5Pxzb/AbZivmr6VE6TkTE8rTjRK9zhKpK6HsRA==",
       "dev": true,
       "license": "ISC",
       "dependencies": {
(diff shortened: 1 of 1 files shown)


perrozzi:
This PR is a lockfile-only bump of `eslint-plugin-promise` from 7.2.1 to 7.3.0 in `voc-datalake/lambda/stream/package-lock.json`, resolving an ERESOLVE failure on cold `npm install` caused by 7.2.1's peer range not covering the project's pinned `eslint ^10.1.0`. The change is minimal, stays within the declared `^7.2.1` range in `package.json`, and touches no runtime code, IAM, or Lambda bundling. This looks clean and safe to merge.

No blockers or must-fix issues found:

- Dev-dependency only (`"dev": true`), so no impact on the deployed stream Lambda artifact or its IAM footprint.
- Peer-dependency update (`|| ^10.0.0`) matches the stated motivation and the diff is consistent with a genuine `npm install` refresh (version, resolved URL, and integrity all updated together).
- No changes to source, tests, ingestor registration, Bedrock model IDs, or frontend code — none of the repo-specific patterns are implicated.

### Nits
- Maintainers may want to spot-verify the new integrity hash against the npm registry (`npm view eslint-plugin-promise@7.3.0 dist.integrity`) as a routine supply-chain check for lockfile-only PRs, since integrity strings can't be validated by eye.
- The 176 pre-existing `vitest-plugin` lint violations in stream test files mentioned in the description should indeed be tracked in a separate issue so the drift doesn't grow silently — worth filing before merge so it isn't lost.
- Consider whether other Lambda workspaces (e.g., other `voc-datalake/lambda/*` packages) share the same `eslint-plugin-promise` pin and would hit the identical cold-install failure; if so, a follow-up sweep would avoid repeating this one lockfile at a time.

No regression test is expected here — this is a dependency-resolution fix, and cold `npm install` succeeding in CI is the effective verification.

---

mundurragacl:
Both nits taken before merge:

**Integrity spot-check**: verified against the registry —

```
npm view eslint-plugin-promise@7.3.0 dist.integrity
sha512-6uGiOR0INuujr6PEQmeSSP7GbIMJ/ebEXXiEzb/nOj68LknH5Pxzb/AbZivmr6VE6TkTE8rTjRK9zhKpK6HsRA==
```

matches the lockfile's integrity string exactly.

**Lint drift**: filed as #187 with the per-rule breakdown and the observation that stream lint isn't in the root gate (which is why it drifted silently).

PR #190, opened 2026-07-15
fix(deploy): stale-cache i18n keys + warning-free CDK synth/deploy

### LINKED ISSUE #188: [Bug]: Deployed frontend shows raw i18n keys after redeploys — no Cache-Control on mutable-name files
## Problem

After a redeploy, returning visitors see raw i18n keys (nav.home, home.title, home.phase1Desc, ...) instead of text.

Root cause: `scripts/deploy.sh` uses `aws s3 sync` with no `--cache-control`, so every object is served without Cache-Control. Browsers apply heuristic caching to the stable-name files (`/locales/*/*.json`, `/config.json`, `index.html`, plugin manifests). The Vite JS bundle is hash-busted so it updates immediately — and then requests NEW translation keys from a STALE cached locale JSON. CloudFront invalidation doesn't help because the staleness is in the browser cache. Observed live on d7afw1xf9irnt.cloudfront.net after deploying six weeks of changes (the new Home page rendered all-raw keys).

## Proposal

Split the upload by mutability:
- `/assets/*` (content-hashed): `public,max-age=31536000,immutable`
- everything else (index.html, config.json, locales/**, manifests): `no-cache` — always revalidates with ETag, cheap 304s, never stale

### LINKED ISSUE #189: [Enhancement]: eliminate CDK synth warnings (GrantOnPrincipalOptions#scope deprecation, KMS wildcard notice)
Every synth/deploy prints:

1. `[WARNING] aws-cdk-lib.aws_iam.GrantOnPrincipalOptions#scope is deprecated` — no code in this repo passes `scope` to any grant; the deprecated call happens inside aws-cdk-lib 2.241.x internals. Upgrading aws-cdk-lib should clear it.
2. `[Warning at /VocCoreStack] ... wildcard is used in the Key policy condition to match all Distribution IDs` — the OAC + SSE-KMS pattern. The synthesized condition is already scoped to this ACCOUNT's distributions (`arn:aws:cloudfront::ACCOUNT:distribution/*`); the CDK README documents the wildcard as the supported initial-deployment shape, and scoping to a concrete distribution id creates the circular dependency the warning describes. Proposal: acknowledge the warning with rationale via `Annotations.acknowledgeWarning` so deploy output stays clean without hiding future new warnings.

(The `#0 building with desktop-linux...` lines are normal Docker BuildKit output from asset bundling, not warnings.)

mundurragacl:
Fixes #188, fixes #189

Two deployment-path fixes found while operating the live environment, one commit each.

## 1. Raw i18n keys after redeploys (#188) — the user-visible bug

After deploying six weeks of changes, returning browsers rendered raw keys (`nav.home`, `home.title`, `home.phase1Desc`, ...) on the new Home page. Root cause: `deploy.sh` ran `aws s3 sync` with **no Cache-Control**, so browsers heuristically cached the stable-name files. The hash-busted JS updates instantly, then requests NEW translation keys from a STALE cached locale JSON. CloudFront invalidation can't help — the staleness is in the browser.

Fix: split the sync by mutability —
- `dist/assets/*` (content-hashed): `public,max-age=31536000,immutable`
- everything else (index.html, config.json, locales/**, manifests): `no-cache` (ETag revalidation, cheap 304s, never stale)

The second sync `--exclude 'assets/*'` so the two passes compose safely with `--delete`. **Deployed and verified live**: `curl -I` shows `no-cache` on locale/config/index and `immutable` on hashed assets.

## 2. Warning-free synth/deploy (#189)

| Warning | Fix |
|---|---|
| `GrantOnPrincipalOptions#scope is deprecated` | Fired inside aws-cdk-lib 2.241 internals (no repo code passes `scope`). Upgraded to `^2.261` + CLI 2.1131 (assembly schema 54). Live diff was benign: CDK-managed helper Lambdas nodejs22→24, one internal policy rename. Deployed to all four stacks. |
| KMS wildcard key policy | The existing `acknowledgeWarning` didn't work: **each `withOriginAccessControl()` re-emits the warning, and the ack only strips messages added before it runs** — the `/prototypes/*` origin sat after the old ack line. Moved the ack after the last origin with a rationale comment (condition already account-scoped; a concrete distribution id recreates the circular dependency the warning describes). |
| `defaultCrossStackReferences` advisory (new in 2.261) | Pinned `'strong'` in cdk.json per the advisory's lock-in option — preserves current producer-protecting behavior for the existing cross-stack refs. |

Also: the refreshed sonarjs eslint plugin began flagging `categoriesSchema.ts`'s `/^_+|_+$/` trim regex (`slow-regex`), breaking the root lint gate — replaced with split/filter/join; behavior pinned unchanged by the existing 49 schema tests.

Note: the `#0 building with desktop-linux...` lines in deploy output are normal Docker BuildKit progress from asset bundling, not warnings; likewise pip's boto3/botocore notice comes from the layer build image, predates these changes, and doesn't affect the layer contents.

## Verification

- `cdk synth` of all four stacks: **zero warnings** (was three)
- All four stacks deployed `UPDATE_COMPLETE` on the live environment; site 200, authenticated API 200 post-deploy
- `npm run check` exits 0; CDK suite 36/36; CategoriesManager suite 49/49

### DIFF
--- a/voc-datalake/cdk.json
+++ b/voc-datalake/cdk.json
@@ -17,6 +17,7 @@
   "context": {
     "@aws-cdk/aws-lambda:recognizeLayerVersion": true,
     "@aws-cdk/core:checkSecretUsage": true,
+    "@aws-cdk/core:defaultCrossStackReferences": "strong",
     "@aws-cdk/core:target-partitions": ["aws", "aws-cn"],
     "@aws-cdk-containers/ecs-service-extensions:enableDefaultLogDriver": true,
     "@aws-cdk/aws-ec2:uniqueImdsv2TemplateName": true,
--- a/voc-datalake/frontend/scripts/deploy.sh
+++ b/voc-datalake/frontend/scripts/deploy.sh
@@ -112,7 +112,7 @@ jq -n \
 
 echo "  ✓ config.json generated with CloudFormation values"
 
-# Step 4: Sync to S3
+# Step 5: Sync to S3
 #
 # Cache-Control is split by mutability (issue #188): Vite's /assets/* are
 # content-hashed, so they can cache forever — but everything with a STABLE
@@ -121,13 +121,25 @@ echo "  ✓ config.json generated with CloudFormation values"
 # JSONs across deploys and the new JS bundle renders raw i18n keys
 # (nav.home, home.title, ...) until a hard refresh. CloudFront invalidation
 # can't fix that: the staleness lives in the browser cache.
+#
+# Metadata note: sync only sets Cache-Control on objects it uploads. That
+# is sufficient because `npm run build` regenerates dist/ with fresh
+# mtimes, so every file uploads (and gets metadata) on each deploy —
+# including the first deploy after this change on buckets that predate it.
+# Don't add --size-only: it would skip unchanged-content files and leave
+# their old metadata in place.
 echo ""
 echo "Step 5: Syncing to S3..."
-# Hashed, immutable assets: cache for a year.
-aws s3 sync dist/assets/ "s3://${BUCKET_NAME}/assets" --delete \
+# Hashed, immutable assets: cache for a year. Deliberately NO --delete:
+# visitors whose browsers cached the previous index.html (which nothing
+# revalidates until they pick up this fix) still reference the previous
+# deploy's chunks — deleting them would turn a stale-but-working page into
+# 404s. Old hashed chunks are harmless and pennies to keep.
+aws s3 sync dist/assets/ "s3://${BUCKET_NAME}/assets" \
   --cache-control 'public,max-age=31536000,immutable'
 # Everything else (stable names): always revalidate. ETags make this cheap
 # (304s), and no visitor ever holds a stale config/locale/index again.
+# --exclude also shields assets/* from this pass's --delete.
 aws s3 sync dist/ "s3://${BUCKET_NAME}" --delete \
   --exclude 'assets/*' \
   --cache-control 'no-cache'
--- a/voc-datalake/frontend/scripts/deploy.sh
+++ b/voc-datalake/frontend/scripts/deploy.sh
@@ -78,12 +78,12 @@ echo "  Avatars CDN: $AVATARS_CDN_URL"
 
 # Step 2: Build the frontend
 echo ""
-echo "Step 3: Building frontend..."
+echo "Step 2: Building frontend..."
 npm run build
 
 # Step 3: Generate runtime config.json
 echo ""
-echo "Step 4: Generating runtime config.json..."
+echo "Step 3: Generating runtime config.json..."
 
 # Use jq to properly escape values and generate valid JSON
 jq -n \
@@ -112,7 +112,7 @@ jq -n \
 
 echo "  ✓ config.json generated with CloudFormation values"
 
-# Step 5: Sync to S3
+# Step 4: Sync to S3
 #
 # Cache-Control is split by mutability (issue #188): Vite's /assets/* are
 # content-hashed, so they can cache forever — but everything with a STABLE
@@ -129,7 +129,7 @@ echo "  ✓ config.json generated with CloudFormation values"
 # Don't add --size-only: it would skip unchanged-content files and leave
 # their old metadata in place.
 echo ""
-echo "Step 5: Syncing to S3..."
+echo "Step 4: Syncing to S3..."
 # Hashed, immutable assets: cache for a year. Deliberately NO --delete:
 # visitors whose browsers cached the previous index.html (which nothing
 # revalidates until they pick up this fix) still reference the previous
@@ -146,7 +146,7 @@ aws s3 sync dist/ "s3://${BUCKET_NAME}" --delete \
 
 # Step 5: Invalidate CloudFront cache
 echo ""
-echo "Step 6: Invalidating CloudFront cache..."
+echo "Step 5: Invalidating CloudFront cache..."
 aws cloudfront create-invalidation --distribution-id "$DISTRIBUTION_ID" --paths '/*' > /dev/null
 
 echo ""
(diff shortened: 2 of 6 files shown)


perrozzi:
This PR fixes a real deployment-path bug — stale browser-cached locale JSONs rendering raw i18n keys after redeploys — by splitting the S3 sync into immutable hashed assets vs. always-revalidate stable files, and cleans up three CDK synth warnings via a lib/CLI upgrade, a repositioned `acknowledgeWarning`, and a pinned cross-stack-reference flag. The changes are well-scoped, well-commented, and verified against a live environment. Overall this looks solid; a couple of transition-window caveats below.

### Should fix

- **`frontend/scripts/deploy.sh` — transition window for browsers still holding the old stale `index.html`.** The first sync runs with `--delete` on `dist/assets/`, immediately removing the previous deploy's hashed chunks. Any browser that heuristically cached the *old* `index.html` (i.e., every visitor from before this fix, since prior deploys set no Cache-Control) will request now-deleted chunks and hit 404s until a hard refresh — arguably worse than raw i18n keys. Consider keeping the prior deploy's assets for one release cycle (skip `--delete` on the assets pass, or prune assets older than N deploys) so the fix rolls out without breaking in-flight stale clients.
- **`frontend/scripts/deploy.sh` — Cache-Control is only applied to objects `sync` actually uploads.** This works today because a fresh Vite build gives every file a newer mtime, but if anyone syncs from an existing `dist/` (or adds `--size-only` later), unchanged-content files keep their old (absent) metadata. Worth a one-line comment noting the fresh-build assumption, or a one-time `aws s3 cp --recursive --metadata-directive REPLACE` backfill for buckets deployed before this change (the live environment is covered per the description, but other deployments of this sample won't be).

### Nits

- `deploy.sh` has a pre-existing `# Step 4: Sync to S3` comment vs. `echo "Step 5: Syncing to S3..."` mismatch — since you're editing this block anyway, worth aligning.
- `categoriesSchema.ts`: the comment says `/^_+|_+$/` is "backtracking-prone", but that alternation is linear-time; sonarjs is being conservative here. The rewrite is fine (and pinned by the existing 49 tests), just soften the comment so future readers don't cargo-cult the claim.
- `core-stack.ts`: the ack's "must come AFTER the last `withOriginAccessControl()` call" comment is good, but it's positionally fragile — anyone adding a new origin below it silently re-breaks the warning-free synth. Consider moving the ack to the very end of the constructor to make the invariant harder to violate.
- `package.json`: pinning the `aws-cdk` CLI as a devDependency alongside `aws-cdk-lib` is a good addition — ensures the assembly-schema-54 requirement travels with the repo rather than depending on a globally installed CLI.

The `cdk.json` `defaultCrossStackReferences: "strong"` pin and the lockfile churn look consistent with the described 2.261 upgrade, and the description's verification (zero-warning synth, four stacks `UPDATE_COMPLETE`, all test suites green) covers the risk areas. No blockers.

---

perrozzi:
The new push addresses all four points from my previous review: the assets sync no longer uses `--delete` (fixing the stale-`index.html` 404 transition window), a metadata note documents the fresh-build/`--size-only` assumption, the step comment mismatch is fixed, the regex comment is softened to correctly note the pattern is linear-time, and the `acknowledgeWarning` now sits as the last statement of the constructor with an explanatory comment.

### Resolved
- `deploy.sh`: dropping `--delete` on the assets pass keeps the previous deploy's hashed chunks, so browsers holding a stale `index.html` keep working through the transition.
- `deploy.sh`: metadata-note comment documents the fresh-mtime assumption and explicitly warns against `--size-only`.
- `deploy.sh`: `# Step 4` / `Step 5` comment/echo mismatch aligned.
- `categoriesSchema.ts`: comment no longer mislabels `/^_+|_+$/` as backtracking-prone.
- `core-stack.ts`: ack moved to the very end of the constructor, making the "after the last origin" invariant hard to violate accidentally.

### New findings
- Trivial: renaming the sync comment to `# Step 5` now duplicates the unchanged `# Step 5: Invalidate CloudFront cache` comment below it — the invalidation block should become Step 6 (its echo already reads "Step 5" pre-existing, so worth a one-line sweep of the step numbering while in this file).

Aside from that numbering nit, the incremental changes are clean and directly resolve the earlier feedback. No blockers.

---

mundurragacl:
Round 1 addressed in 5ad3b2a — both should-fixes taken, all nits taken.

**Transition-window 404s (should-fix)**: excellent catch — taken exactly as suggested. The assets pass no longer `--delete`s: pre-#188 visitors whose browsers heuristically cached the old `index.html` keep working against the previous deploy's chunks until their cache revalidates into the no-cache regime. Old hashed chunks are harmless and cheap to keep; the stable-name pass retains `--delete` with `--exclude 'assets/*'` (which also shields assets from that pass's deletions). Redeployed live with the new script.

**Metadata-only-on-upload (should-fix)**: documented in the script — the fresh-build assumption (vite regenerates mtimes → every deploy uploads everything → metadata applied, including the first post-#188 deploy on pre-existing buckets) plus an explicit warning against adding `--size-only`. No backfill needed for any bucket that deploys once with this script.

**Step-number mismatch (nit)**: aligned.

**slow-regex comment (nit)**: softened — now states the anchor alternation is linear-time and sonarjs is being conservative; the rewrite stays for the clean gate.

**Positionally fragile ack (nit)**: taken — the ack is now the constructor's **last statement** with the comment explaining why (any future origin below the old position would have silently re-broken warning-free synth).

Verification: `bash -n` clean, zero-warning synth re-confirmed after the ack move, frontend redeployed with the new sync behavior, `npm run check` exit 0 (frontend 1714, CDK 36, backend green).

mundurragacl:
Round 2 nit taken — the whole file now numbers coherently 1→5 with comment/echo pairs aligned (fetch, build, config, sync, invalidate). `bash -n` clean.

perrozzi:
The only change since my last review is a sweep of the step numbering in `deploy.sh`, directly addressing the single remaining nit from the previous round.

### Resolved
- `deploy.sh`: step comments and echo statements are now consistent throughout (Step 2 build → Step 3 config → Step 4 sync → Step 5 invalidation), with no duplicate step labels.

### New findings
None. The change is comment/echo-string only, with no behavioral impact on the sync or Cache-Control logic approved earlier. No blockers — this looks ready to merge.

---

PR #192, opened 2026-07-16
fix(i18n): version-stamped locale URLs bust pre-#188 browser caches

### LINKED ISSUE #191: [Bug]: Pre-#188 browser caches keep serving stale locale/config JSONs — need version-stamped fetch URLs
## Problem

#188 set correct Cache-Control headers, but a header only governs fetches that reach the server. Browsers that cached the JSONs BEFORE the fix hold copies with no Cache-Control and a heuristic freshness of ~10% of object age — for files last deployed six weeks earlier, that's DAYS of serving stale content without a single network request. Observed live: a returning session rendered raw `home.phase3Title`/`home.phase3Desc` keys because its cached `dashboard.json` predated those keys, while the deployed file was verified correct.

## Proposal

Version-stamp the runtime-fetched URLs from the (hash-busted, always-fresh) JS bundle:
- i18next `loadPath: '/locales/{{lng}}/{{ns}}.json?v=' + __APP_VERSION__`
- `fetch('/config.json?v=' + __APP_VERSION__)`
- `__APP_VERSION__` injected by Vite `define` at build time

A new bundle then requests URLs no old cache entry can match — retroactively busting every stale copy, including the pre-#188 ones. Belt-and-braces with #188's headers.

mundurragacl:
Fixes #191

## Problem (observed live after #188)

Raw i18n keys (`home.phase3Title`, `home.phase3Desc`) still rendered on the deployment for returning sessions. #188's Cache-Control headers were correct at the CDN — verified: the deployed `dashboard.json` carried all 15 `home.*` keys — but a header only governs fetches that reach the server. Browsers that cached the JSONs **before** #188 hold copies with no Cache-Control and heuristic freshness of ~10% of object age (days, for files last deployed six weeks prior), served without any network request. The hash-busted JS updates instantly and asks a stale JSON for keys that didn't exist when it was cached.

## Fix

Version-stamp the locale URLs from the always-fresh bundle:

- `i18n/loadPath.ts` (new, side-effect-free so tests don't boot the i18next singleton): `LOCALE_LOAD_PATH = '/locales/{{lng}}/{{ns}}.json?v=' + APP_VERSION`
- `APP_VERSION` injected by Vite `define` (`Date.now()` at build); vitest defines `'test'` for deterministic assertions
- `declare global` lives in `loadPath.ts` itself — the conventional `vite-env.d.ts` is blocked by the root `.gitignore`'s blanket `*.d.ts` rule, and the dunder `__APP_VERSION__` name is rejected by the repo's naming-convention lint (both discovered the hard way)
- `config.json` needs nothing — `runtimeConfig.ts` already fetches it with `cache: 'no-store'`

Old cache entries keyed on the un-versioned URL can never match a versioned request, so this retroactively unbricks every pre-#188 browser — belt-and-braces with #188's headers going forward.

## Verification (live)

- Fresh session on d7afw1xf9irnt.cloudfront.net: every locale request carries `?v=<build-id>` (network log), Home renders translated text, zero raw keys in the DOM
- Production bundle grep confirms the stamped path is baked in
- `npm run check` exits 0; `loadPath.test.ts` pins the placeholder shape and the version stamp

### DIFF
--- a/voc-datalake/frontend/src/i18n/config.ts
+++ b/voc-datalake/frontend/src/i18n/config.ts
@@ -11,6 +11,7 @@ import i18n from 'i18next'
 import LanguageDetector from 'i18next-browser-languagedetector'
 import HttpBackend from 'i18next-http-backend'
 import { initReactI18next } from 'react-i18next'
+import { LOCALE_LOAD_PATH } from './loadPath'
 
 export const supportedLanguages = ['en', 'es', 'fr', 'de', 'pt', 'ja', 'zh', 'ko'] as const
 export type SupportedLanguage = (typeof supportedLanguages)[number]
@@ -49,7 +50,9 @@ void i18n
     defaultNS,
     ns: [...namespaces],
 
-    backend: { loadPath: '/locales/{{lng}}/{{ns}}.json' },
+    // Version-stamped path — see i18n/loadPath.ts for the cache-busting
+    // rationale (issue #191).
+    backend: { loadPath: LOCALE_LOAD_PATH },
 
     detection: {
       // 'navigator' is intentionally omitted so a non-English browser doesn't
--- a/voc-datalake/frontend/src/i18n/loadPath.test.ts
+++ b/voc-datalake/frontend/src/i18n/loadPath.test.ts
@@ -14,8 +14,8 @@ describe('LOCALE_LOAD_PATH (issue #191)', () => {
   })
 
   it('is version-stamped so stale cache entries can never match', () => {
-    // vitest.config.ts defines __APP_VERSION__ as 'test'; production builds
-    // stamp a per-build id via vite.config.ts define.
-    expect(LOCALE_LOAD_PATH).toBe('/locales/{{lng}}/{{ns}}.json?v=test')
+    // Shape assertion survives a change of the injected test literal;
+    // vitest.config.ts currently defines APP_VERSION as 'test'.
+    expect(LOCALE_LOAD_PATH).toMatch(/\?v=.+$/)
   })
 })
--- a/voc-datalake/frontend/src/i18n/loadPath.test.ts
+++ b/voc-datalake/frontend/src/i18n/loadPath.test.ts
@@ -1,0 +1,21 @@
+/**
+ * Regression tests for issue #191: pre-#188 browser caches hold locale
+ * JSONs with heuristic freshness and serve them for days without a network
+ * request, so new bundles rendered raw i18n keys for translations added
+ * since. The loadPath must be version-stamped so each build requests URLs
+ * no stale cache entry can match.
+ */
+import { describe, it, expect } from 'vitest'
+import { LOCALE_LOAD_PATH } from './loadPath'
+
+describe('LOCALE_LOAD_PATH (issue #191)', () => {
+  it('keeps the i18next language/namespace placeholders', () => {
+    expect(LOCALE_LOAD_PATH).toContain('/locales/{{lng}}/{{ns}}.json')
+  })
+
+  it('is version-stamped so stale cache entries can never match', () => {
+    // vitest.config.ts defines __APP_VERSION__ as 'test'; production builds
+    // stamp a per-build id via vite.config.ts define.
+    expect(LOCALE_LOAD_PATH).toBe('/locales/{{lng}}/{{ns}}.json?v=test')
+  })
+})
--- a/voc-datalake/frontend/src/i18n/loadPath.ts
+++ b/voc-datalake/frontend/src/i18n/loadPath.ts
@@ -15,12 +15,17 @@
  * @module i18n/loadPath
  */
 
-/** Build id injected by Vite `define` (vite.config.ts / vitest.config.ts).
- * Declared here rather than in a .d.ts because the root .gitignore
+/** Build id injected by Vite `define` (vite.config.ts / vitest.config.ts),
+ * namespaced under import.meta.env so the substitution can't collide with
+ * user identifiers. This `declare global` augmentation is project-wide by
+ * nature (any module reading import.meta.env.APP_VERSION gets the type);
+ * it lives here rather than in vite-env.d.ts because the root .gitignore
  * excludes all *.d.ts files (build-emitted declarations). */
 declare global {
-  const APP_VERSION: string
+  interface ImportMetaEnv {
+    readonly APP_VERSION: string
+  }
 }
 
 /** i18next-http-backend loadPath, cache-busted per build. */
-export const LOCALE_LOAD_PATH = `/locales/{{lng}}/{{ns}}.json?v=${APP_VERSION}`
+export const LOCALE_LOAD_PATH = `/locales/{{lng}}/{{ns}}.json?v=${import.meta.env.APP_VERSION}`
--- a/voc-datalake/frontend/src/i18n/loadPath.ts
+++ b/voc-datalake/frontend/src/i18n/loadPath.ts
@@ -1,0 +1,26 @@
+/**
+ * @fileoverview Version-stamped locale load path (issue #191).
+ *
+ * Browsers that cached the locale JSONs before #188 set Cache-Control hold
+ * copies with heuristic freshness (~10% of object age — days, for files
+ * last deployed weeks earlier) and serve them without any network request,
+ * so the hash-busted JS bundle renders raw i18n keys for translations
+ * added since. Stamping the URL with the build id means each new bundle
+ * requests URLs no stale cache entry can match. config.json doesn't need
+ * this: runtimeConfig fetches it with cache: 'no-store'.
+ *
+ * Kept in its own side-effect-free module so tests can pin the URL shape
+ * without importing config.ts (which initializes the i18next singleton).
+ *
+ * @module i18n/loadPath
+ */
+
+/** Build id injected by Vite `define` (vite.config.ts / vitest.config.ts).
+ * Declared here rather than in a .d.ts because the root .gitignore
+ * excludes all *.d.ts files (build-emitted declarations). */
+declare global {
+  const APP_VERSION: string
+}
+
+/** i18next-http-backend loadPath, cache-busted per build. */
+export const LOCALE_LOAD_PATH = `/locales/{{lng}}/{{ns}}.json?v=${APP_VERSION}`
--- a/voc-datalake/frontend/vite.config.ts
+++ b/voc-datalake/frontend/vite.config.ts
@@ -1,7 +1,7 @@
 import tailwindcss from '@tailwindcss/vite'
 import react from '@vitejs/plugin-react'
-import { existsSync, readFileSync } from 'node:fs'
-import { dirname, join } from 'node:path'
+import { existsSync, readFileSync, statSync } from 'node:fs'
+import { dirname, isAbsolute, join } from 'node:path'
 import { defineConfig } from 'vite'
 
 /**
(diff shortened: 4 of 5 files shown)


perrozzi:
This PR fixes issue #191 by version-stamping the i18next locale URLs (`?v=<build-id>` via a Vite `define`), so browsers holding pre-#188 heuristically-fresh copies of the locale JSONs can never satisfy the new requests from cache. The change is small, well-documented, isolated in a side-effect-free module, and comes with a regression test that would fail if the stamp were reverted. Overall this looks solid; my concerns are about the replacement-token choice and build determinism, not correctness of the fix.

### Should fix

- **Bare `APP_VERSION` as a `define` key is risky** (`voc-datalake/frontend/vite.config.ts`). Vite `define` performs identifier-level text replacement across all transformed source, so any other occurrence of the identifier `APP_VERSION` — a local variable, a destructured binding, shorthand object property, a future teammate's code — will be silently replaced with the build timestamp string, likely producing a confusing syntax or type error far from the cause. This is exactly why Vite's docs recommend the `__DUNDER__` convention. The PR description says the naming-convention lint rejects `__APP_VERSION__`; the better resolution is a targeted lint exception (or an `varsIgnorePattern`/filter for the injected constant) rather than adopting a collision-prone bare name. Alternatively, define it as `import.meta.env.APP_VERSION`, which is namespaced and can't collide with user identifiers.
- **`Date.now()` at build time makes every build non-reproducible** (`voc-datalake/frontend/vite.config.ts`). Because the value is inlined into the bundle, every `vite build` changes the content hash of the chunk containing it — even with zero code changes — so every deploy forces clients to re-download that JS and re-fetch all locale JSONs. Consider deriving the stamp from something stable per code state (git commit SHA, `package.json` version) so identical source produces identical artifacts and unchanged deploys stay cache-warm.

### Nits

- The comment in `vite.config.ts` says the build id is "stamped into runtime-fetched URLs (/config.json, locale JSONs)", but `/config.json` is *not* stamped (it relies on `cache: 'no-store'`, per the PR description and `loadPath.ts`'s own docs). Correct the comment to avoid future confusion.
- `declare global` in `loadPath.ts` makes `APP_VERSION` an ambient global for the entire project, not just this module. That's presumably intended (config.ts benefits), but the doc comment reads as if the declaration is module-scoped — worth a one-line clarification.
- The second test in `loadPath.test.ts` (`toBe(...?v=test)`) subsumes the first (`toContain`); keeping both is fine for failure-message clarity, but consider asserting the query-param shape (e.g., matches `/\?v=.+$/`) so the test survives a future switch away from the `'test'` literal.

---

perrozzi:
Since the last review, the author replaced the bare `APP_VERSION` define with a namespaced `'import.meta.env.APP_VERSION'` key (in both vite and vitest configs), switched the build id from `Date.now()` to a git-sha-derived value with a `-dirty` marker and a timestamp fallback, rewrote the misleading `/config.json` comment, clarified the `declare global` doc comment, and loosened the second test to a shape assertion. All previous findings were addressed.

### Resolved
- Bare `APP_VERSION` define collision risk — now namespaced as `'import.meta.env.APP_VERSION'` with an explanatory comment (`vite.config.ts`, `vitest.config.ts`, `loadPath.ts`).
- Non-reproducible `Date.now()` build id — now derived from `git rev-parse --short HEAD` (+`-dirty`), so identical source yields identical bundles (`vite.config.ts`).
- Stale comment claiming `/config.json` is stamped — comment replaced.
- `declare global` scope ambiguity — doc comment now states the augmentation is project-wide by design (`loadPath.ts`).
- Brittle exact-match test — now asserts the `/\?v=.+$/` shape (`loadPath.test.ts`).

### New findings
- **Minor** (`voc-datalake/frontend/src/i18n/loadPath.ts`): the `declare global` augments `ImportMetaEnv` but not `ImportMeta`. This type-checks only because Vite's client types (which put `env: ImportMetaEnv` on `ImportMeta`) are already in scope via tsconfig `types` or an existing reference. That's evidently true today (`npm run check` passes), but worth confirming the augmentation actually merges with Vite's global `ImportMetaEnv` rather than shadowing it — if the project ever drops the `vite/client` reference, `import.meta.env` itself will stop type-checking here with a confusing error.
- **Nit** (`voc-datalake/frontend/vite.config.ts`): the `Date.now()` fallback silently reintroduces non-determinism in environments without git (e.g., CI building from a tarball, as the comment notes). Acceptable and documented, but if the deploy pipeline is one of those environments, the reproducibility win never materializes there — consider allowing an env-var override (e.g., a CI-provided commit SHA) before falling back to a timestamp.

Otherwise the incremental changes are clean and correctly implement the suggestions from the previous review.

---

perrozzi:
Since the last review, the author replaced the `execSync` git invocations with direct filesystem reads of `.git/HEAD` (motivated by the `sonarjs/no-os-command-from-path` lint rule), including a `findUp` walk to locate `.git` in the monorepo and fallbacks for detached HEAD and packed refs. My previous two findings were both Minor/Nit-level; neither is addressed by this change — the `ImportMetaEnv`-only augmentation note stands as-is, and the timestamp-fallback nit is actually made more relevant (see below).

### New findings
- **Should-fix** (`voc-datalake/frontend/vite.config.ts`): the packed-ref punt to `Date.now()` will fire in the most common CI scenario — a fresh `git clone` typically writes refs to `.git/packed-refs`, not loose files under `.git/refs/`, so `existsSync(refPath)` fails and the build id becomes a timestamp exactly where reproducibility was the goal. Either parse `packed-refs` (lines are `<sha> <ref>`) or honor a CI-provided env var (`GITHUB_SHA` / `CODEBUILD_RESOLVED_SOURCE_VERSION`) before falling back — this also subsumes my earlier nit.
- **Minor** (`voc-datalake/frontend/vite.config.ts`): the `-dirty` marker from the previous iteration was silently dropped. A dirty working tree now stamps the clean HEAD sha, so a build containing uncommitted locale-JSON edits shares a `?v=` with the committed build — a cache could serve the wrong copy. Mitigated by #188's no-cache headers going forward, but it's a quiet regression from the version I approved, and the doc comment no longer mentions the trade-off.
- **Minor** (`voc-datalake/frontend/vite.config.ts`): in git worktrees and submodules, `.git` is a *file* containing `gitdir: <path>`, not a directory. `findUp`'s `existsSync` matches it, then `readFileSync(join(gitFile, 'HEAD'))` throws and the catch falls back to a timestamp — functional but silently non-deterministic. Worth either following the `gitdir:` indirection or noting the limitation in the comment alongside the tarball case.

Otherwise the refactor is sound: the lint rationale is documented, the depth-capped `findUp` terminates correctly, and detached-HEAD handling is right.

---

mundurragacl:
Round 1 addressed across 6343519 + faeda06 — both should-fixes taken, all nits taken.

**Bare-identifier define (should-fix)**: taken via your `import.meta.env` suggestion — the define key is now `'import.meta.env.APP_VERSION'`, which is namespaced and cannot collide with user identifiers (comment in vite.config.ts explains the substitution hazard). Typed via an `ImportMetaEnv` interface augmentation.

**Non-reproducible builds (should-fix)**: taken — the stamp is the git HEAD sha. Verified: two consecutive builds emit **identical chunk hashes** with `?v=<sha>`, so content-neutral redeploys stay cache-warm; post-#188 no-cache headers cover same-sha freshness regardless. One wrinkle worth noting: the first cut used `execSync('git rev-parse ...')`, which the repo's own sonarjs gate rejected (`no-os-command-from-path`) — the final version reads `.git/HEAD` directly via fs (ref indirection + detached-HEAD support + monorepo find-up; packed refs and missing .git fall back to a timestamp), which is also faster than spawning a process.

**config.json comment mismatch (nit)**: fixed — the comment no longer claims config.json is stamped.

**`declare global` scoping (nit)**: doc comment now states the augmentation is project-wide by nature and why it lives in loadPath.ts (root .gitignore blankets *.d.ts).

**Test shape (nit)**: the stamp test now asserts `/\?v=.+$/` rather than the injected literal.

Deployed live at HEAD: locale requests carry `?v=6343519`-era stamps; `npm run check` exit 0.

perrozzi:
The push since my last review reworks `buildId()` in `voc-datalake/frontend/vite.config.ts` to address all three of my previous findings: CI sha env vars now take precedence, packed-refs are parsed instead of punted, the `gitdir:` file indirection is followed, and the dirty-tree trade-off is now explicitly documented in the comment. No other files changed.

### Resolved
- CI packed-ref fallback: `GITHUB_SHA`/`CODEBUILD_RESOLVED_SOURCE_VERSION` checked first, and `resolveRef` parses `packed-refs` (correctly skipping `#` header and `^` peel lines).
- Dropped `-dirty` marker: the trade-off is now documented in the doc comment with the lint-rule rationale.
- Worktree/submodule `.git` file: `resolveGitDir` follows `gitdir:` indirection, handling relative and absolute targets.

### New findings
- **Minor** (`voc-datalake/frontend/vite.config.ts`): the worktree handling is incomplete despite the comment claiming it. In a linked worktree, `gitdir:` points to `.git/worktrees/<name>/`, which contains `HEAD` but **not** `refs/` or `packed-refs` — those live in the common dir referenced by the `commondir` file inside it. So a symbolic-ref HEAD in a worktree resolves `HEAD` fine, then `resolveRef` misses both the loose ref and `packed-refs` and silently falls back to a timestamp. Either follow `commondir` when present or soften the comment to say worktrees fall back to the timestamp (submodules are genuinely handled — `.git/modules/<name>` does contain refs).
- **Nit** (`voc-datalake/frontend/vite.config.ts`): `process.env.GITHUB_SHA ?? process.env.CODEBUILD_RESOLVED_SOURCE_VERSION` — if `GITHUB_SHA` is set but empty, `??` won't fall through to the CodeBuild var; the empty-string guard then skips CI resolution entirely. `(GITHUB_SHA || CODEBUILD_...)`-style truthiness (or two sequential checks) would be more robust. Harmless in practice since the git fallback still runs.

Otherwise clean — the resolution order is sensible, all fs reads stay inside the `try`, and the packed-refs suffix match (`endsWith(' ' + ref)`) can't confuse `refs/heads/main` with `refs/remotes/origin/main`.

---

mundurragacl:
Round 2 addressed in 775d2cb — should-fix and both minors taken.

**Packed refs in CI (should-fix)**: taken with both halves of your suggestion — `GITHUB_SHA`/`CODEBUILD_RESOLVED_SOURCE_VERSION` take precedence (covers fresh CI clones directly), and `resolveRef()` falls back from the loose ref file to parsing `packed-refs` (`<sha> <ref>` lines, comment/peel lines skipped). The timestamp is now genuinely last-resort. Verified: local build stamps the HEAD sha from fs; `GITHUB_SHA=abc1234def npm run build` stamps `abc1234`.

**Dropped -dirty marker (minor)**: acknowledged as a real trade-off and now documented in the comment — dirty detection requires `git status`, which the same sonarjs rule that forced the fs rewrite bans. A dirty tree stamps the clean HEAD sha; deploys are expected from committed trees, and #188's no-cache headers keep same-sha fetches revalidating regardless.

**Worktree/submodule .git files (minor)**: taken — `resolveGitDir()` follows the `gitdir: <path>` indirection (absolute or relative) instead of throwing into the timestamp fallback.

Gate green, redeployed live at this HEAD.

mundurragacl:
Round 3 minor taken in 459eec9 — `resolveRef` now resolves loose refs and `packed-refs` against the worktree COMMON dir when a `commondir` file is present (absolute or relative target), so linked-worktree builds stamp the real sha instead of a timestamp. Main-checkout behavior unchanged (no commondir → gitDir itself). Gate green, sha stamp re-verified in the bundle.

perrozzi:
The push since my last review adds `refsBaseDir()` to `voc-datalake/frontend/vite.config.ts`, following the `commondir` file so linked-worktree refs resolve against the common dir instead of silently falling back to a timestamp — this directly addresses my remaining Minor finding. The `??`-vs-truthiness nit on the CI env vars was not changed.

### Resolved
- Worktree ref resolution: `resolveRef` now resolves loose refs and `packed-refs` against the common dir when `gitDir/commondir` exists, with correct relative/absolute handling; the doc comment now accurately describes the behavior.

### Still open
- Nit: `GITHUB_SHA ?? CODEBUILD_RESOLVED_SOURCE_VERSION` won't fall through when `GITHUB_SHA` is set-but-empty (harmless; git fallback still runs).

### New findings
None. `refsBaseDir` is called only from `resolveRef`, which runs inside `buildId`'s `try`, so its `readFileSync` failure modes are covered; the relative-path join against `gitDir` matches git's actual `commondir` semantics (typically `../..` relative to `.git/worktrees/<name>`). Clean delta.

---

mundurragacl:
Round 4 nit taken — the CI sha candidates are scanned with find() over both env vars, so a set-but-empty GITHUB_SHA falls through to CODEBUILD_RESOLVED_SOURCE_VERSION and then git. Gate green. That closes out every finding across the four rounds.

perrozzi:
The only change since my last review is a refactor of the CI sha selection in `voc-datalake/frontend/vite.config.ts`: the `??` chain is replaced with an array `.find()` that skips both undefined and empty-string values. This directly addresses my one remaining nit.

### Resolved
- Empty-string CI env var: `GITHUB_SHA=''` now correctly falls through to `CODEBUILD_RESOLVED_SOURCE_VERSION` (and to the git fallback if both are empty/unset).

### New findings
None. The `.find()` predicate and the subsequent `ciSha !== undefined` guard are equivalent to the prior semantics minus the empty-string bug. Clean delta — all previous findings are now resolved.

---

Issue #187, opened 2026-07-15
[Enhancement]: lambda/stream lint drift — 176 vitest-plugin violations in test files


mundurragacl:
Raised on #186's review: `npm run lint` in `voc-datalake/lambda/stream` reports 176 pre-existing errors, all in test files — dominated by `vitest/prefer-strict-equal` (40), `vitest/max-expects` (34), `@typescript-eslint/no-unused-vars` (12), `vitest/no-conditional-expect` (8), plus smaller counts. The stream lint isn't part of the root `npm run check` gate, so the drift grows silently. Follow-up: either fix the violations (10 are auto-fixable), tune the vitest rule config to the intended strictness, and/or wire stream lint into the root gate so it can't drift again.

perrozzi:
Thanks for filing this! To restate: `npm run lint` in `voc-datalake/lambda/stream` currently reports 176 errors concentrated in test files — mainly `vitest/prefer-strict-equal`, `vitest/max-expects`, `@typescript-eslint/no-unused-vars`, and `vitest/no-conditional-expect` — and since stream lint isn't wired into the root `npm run check` gate, the drift grows unchecked.

The breakdown and proposed follow-ups (auto-fix what's fixable, tune vitest rule strictness, and/or add stream lint to the root gate) are clear. One quick question to help scope the work: do you have a preference on ordering — e.g., wiring the gate in first with rules temporarily relaxed to current baseline, versus fixing all violations before enabling the gate? That would help split this into reviewable chunks if someone picks it up.

_Suggested labels applied: `enhancement`, `javascript`_

---
_Automated review generated with Amazon Bedrock (Claude Fable 5). It may contain mistakes - maintainers have the final say._

mundurragacl:
Fixed by #199 (squash-merged to development). All 176 violations fixed for real — including 25 in production sources the issue hadn't counted (type assertions, banned lets, a backtracking-prone regex, complexity/depth/size breaches). Config tuned only where sanctioned: vitest/max-expects 4→10 with in-file rationale, no-unused-vars argsIgnorePattern '^_'. Root 'npm run lint' and 'npm run check' now run stream lint AND stream tests, so the drift structurally cannot restart (mutation-verified: one reintroduced violation fails the gate). Bonus catch: eslint's prefer-called-with autofix had rewritten toHaveBeenCalled() into zero-arg toHaveBeenCalledWith() — 3 tests silently weakened; now assert real arguments.

PR #195, opened 2026-07-16
fix(layers): stop shipping a mismatched botocore; quiet, verified-clean bundling

### LINKED ISSUE #194: [Bug]: Lambda layers ship a transitive botocore that shadows the runtime SDK (and the build prints pip errors)
Every `./scripts/build-layers.sh` run prints, per layer:

```
ERROR: pip's dependency resolver ... boto3 1.40.4 requires botocore<1.41.0, but you have botocore 1.43.49 which is incompatible.
WARNING: Running pip as the 'root' user ...
[notice] A new release of pip is available ...
```

The ERROR is a symptom of a real defect: neither layer pins botocore, but `aws-lambda-powertools[tracer]` → `aws-xray-sdk` pulls it transitively, so the layers ship botocore 1.43.x WITHOUT boto3. On Lambda, /opt/python precedes /var/runtime on sys.path, so the runtime's boto3 gets paired with the layer's mismatched botocore — exactly the incompatibility pip is warning about, shipped to production. It also bloats each layer ~15MB.

Fix: strip boto3/botocore from the layer output after install (Lambda's runtime provides the SDK); add `--root-user-action=ignore` and `--disable-pip-version-check` for the noise; document why. Build output should be clean and layers ~15MB smaller.

mundurragacl:
Fixes #194

## Problem

Every layer build — manual (`scripts/build-layers.sh`) and CDK synth-time bundling — shipped a **mismatched botocore without boto3** into `/opt/python`. Lambda's runtime provides boto3/botocore as a matched pair, but `aws-lambda-powertools[tracer] → aws-xray-sdk` pulls botocore in transitively. Since `/opt/python` precedes `/var/runtime` on `sys.path`, the runtime's boto3 gets paired with the layer's stale botocore — exactly the incompatibility pip's resolver ERROR warned about on every single build:

```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed.
boto3 1.40.4 requires botocore<1.41.0,>=1.40.4, but you have botocore 1.43.49 which is incompatible.
```

Deploys additionally printed root-user warnings, unwritable-cache warnings, pip self-update notices, an "esbuild cannot run locally. Switching to Docker bundling" warning, and an npm audit summary reporting vulnerabilities from the bundling container.

## Fix

**Commit 1 — `scripts/build-layers.sh`** (manual builds)
- pip runs from a throwaway venv inside the container: the build image preinstalls boto3 system-wide, and pip's post-install consistency check compared it against the target dir's botocore — a clean venv has nothing to conflict with (`--ignore-installed` alone did *not* silence it)
- strip `boto3`/`botocore` after install (runtime provides the matched pair; −15MB/layer)
- `--quiet --no-cache-dir --root-user-action=ignore --disable-pip-version-check`

**Commit 2 — CDK synth-time bundling** (what deploys actually ship)
- new `lib/utils/python-layer-bundling.ts` → `pythonLayerCode()`, the same venv+strip recipe, used by all three DepsLayers (api/ingestion/processing); both files cross-reference each other as kept-in-lockstep
- layer asset now **excludes the local `python/` build output and `.DS_Store`**: the old `cp -r . /asset-output/python/` nested a stale pre-built tree plus `requirements.txt` inside the deployed layer, and the asset hash churned on every local rebuild
- pin `esbuild@0.28.1` (exact) as a devDependency so `ChatStreamApi` bundles locally — no Docker fallback, no container `npm install`
- `npm audit fix` on both lockfiles (voc-datalake: vite/postcss chains; lambda/stream: fast-xml-parser/picomatch/postcss/vite chains) → both report **0 vulnerabilities**; only XML-parsing packages changed in the stream lockfile, no eslint/vitest plugin drift

## Verification

- `cdk synth --quiet`: 43 lines total, zero pip noise; esbuild local in 18ms; nodeModules install reports 0 vulnerabilities; ApiDepsLayer dedupes into ProcessingDepsLayer's asset (same source + recipe)
- Deployed all 5 stacks to us-west-2 (full log reviewed line-by-line): all UPDATE_COMPLETE, zero warnings/errors, new layer versions rolled
- Downloaded the three **deployed** layer zips: 0 boto3/botocore files, no nested junk, 8–10MB zipped
- Live smoke: site 200; 6 authenticated API endpoints 200; `POST /chat/stream` streams SSE (metadata→text→done); FeedbackProcessor empty-batch invoke → 200, no FunctionError (proves layer imports on ARM64)
- `npm run check` from root: exit 0 (lint, typecheck, 881 backend + frontend + CDK + stream tests)

## Tests

No new unit tests: this changes build tooling/infrastructure recipes, not application code. The behavior is covered by the deployed-artifact verification above (layer zip contents inspected, runtime imports proven via live invoke), which is the only meaningful way to assert on container bundling output. CDK stack tests (36) still pass and synth the changed stacks with bundling skipped.

### DIFF
--- a/voc-datalake/lambda/layers/ingestion-deps/README.md
+++ b/voc-datalake/lambda/layers/ingestion-deps/README.md
@@ -1,0 +1,21 @@
+# ingestion-deps layer
+
+Dependencies for the ingestion plugin Lambdas (ARM64/Graviton).
+
+Only `requirements.txt` matters here. The deployed layer contains ONLY what
+pip installs from it — via `lib/utils/python-layer-bundling.ts` at CDK synth
+time, or `scripts/build-layers.sh` for manual builds (same recipe, guarded by
+`lib/utils/python-layer-bundling.test.ts`).
+
+Do NOT place first-party code in this directory: it will silently not deploy.
+Shared runtime code belongs in `lambda/shared/` (or `plugins/_shared/` for
+plugin-only helpers).
+
+Notes:
+
+- `python/` is local build output from `scripts/build-layers.sh`; it is
+  excluded from the CDK asset and can be deleted at any time.
+- boto3/botocore are stripped from the built layer on purpose — the Lambda
+  runtime provides a matched pair, and shipping a transitive botocore
+  shadows it (issue #194). If you ever need a newer SDK than the runtime
+  ships, vendor boto3 AND botocore together and remove the strip.
--- a/voc-datalake/lambda/layers/processing-deps/README.md
+++ b/voc-datalake/lambda/layers/processing-deps/README.md
@@ -1,0 +1,20 @@
+# processing-deps layer
+
+Dependencies for the processing and API Lambdas (ARM64/Graviton).
+
+Only `requirements.txt` matters here. The deployed layer contains ONLY what
+pip installs from it — via `lib/utils/python-layer-bundling.ts` at CDK synth
+time, or `scripts/build-layers.sh` for manual builds (same recipe, guarded by
+`lib/utils/python-layer-bundling.test.ts`).
+
+Do NOT place first-party code in this directory: it will silently not deploy.
+Shared runtime code belongs in `lambda/shared/`.
+
+Notes:
+
+- `python/` is local build output from `scripts/build-layers.sh`; it is
+  excluded from the CDK asset and can be deleted at any time.
+- boto3/botocore are stripped from the built layer on purpose — the Lambda
+  runtime provides a matched pair, and shipping a transitive botocore
+  shadows it (issue #194). If you ever need a newer SDK than the runtime
+  ships, vendor boto3 AND botocore together and remove the strip.
--- a/voc-datalake/lambda/stream/package-lock.json
+++ b/voc-datalake/lambda/stream/package-lock.json
@@ -844,13 +844,12 @@
       }
     },
     "node_modules/@aws-sdk/xml-builder": {
-      "version": "3.972.15",
-      "resolved": "https://registry.npmjs.org/@aws-sdk/xml-builder/-/xml-builder-3.972.15.tgz",
-      "integrity": "sha512-PxMRlCFNiQnke9YR29vjFQwz4jq+6Q04rOVFeTDR2K7Qpv9h9FOWOxG+zJjageimYbWqE3bTuLjmryWHAWbvaA==",
+      "version": "3.972.36",
+      "resolved": "https://registry.npmjs.org/@aws-sdk/xml-builder/-/xml-builder-3.972.36.tgz",
+      "integrity": "sha512-RdGmS1GLrtaTOLE1ElSluMldNrpk9Emq6uYs8SS8iHlu5xTAmM9rRkM91o48+rIRryBtyO9t+uLYCoMG6jVMVA==",
       "license": "Apache-2.0",
       "dependencies": {
-        "@smithy/types": "^4.13.1",
-        "fast-xml-parser": "5.5.8",
+        "@smithy/types": "^4.16.1",
         "tslib": "^2.6.2"
       },
       "engines": {
@@ -927,21 +926,21 @@
       }
     },
     "node_modules/@emnapi/core": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.9.1.tgz",
-      "integrity": "sha512-mukuNALVsoix/w1BJwFzwXBN/dHeejQtuVzcDsfOEsdpCumXb/E9j8w11h5S54tT1xhifGfbbSm/ICrObRb3KA==",
+      "version": "1.11.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/core/-/core-1.11.1.tgz",
+      "integrity": "sha512-RSvbQmHzdKzNsLYa/wHrbc3KN4sYLKAdPZxqiM2HATqv/SBk2/ENSHpvXGaLOMcsAyz0poEGqkmmKYG3OWiJEQ==",
       "dev": true,
       "license": "MIT",
       "optional": true,
       "dependencies": {
-        "@emnapi/wasi-threads": "1.2.0",
+        "@emnapi/wasi-threads": "1.2.2",
         "tslib": "^2.4.0"
       }
     },
     "node_modules/@emnapi/runtime": {
-      "version": "1.9.1",
-      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.9.1.tgz",
-      "integrity": "sha512-VYi5+ZVLhpgK4hQ0TAjiQiZ6ol0oe4mBx7mVv7IflsiEp0OWoVsp/+f9Vc1hOhE0TtkORVrI1GvzyreqpgWtkA==",
+      "version": "1.11.1",
+      "resolved": "https://registry.npmjs.org/@emnapi/runtime/-/runtime-1.11.1.tgz",
+      "integrity": "sha512-vgj7R3y3Wgx24IQaGPA/R6YFXLHVMOZ0uVEyIQPaWs+rd1AzfEMXlAC22FYwO1XkKR6NPsq7mUandH8oIRdZFw==",
       "dev": true,
       "license": "MIT",
       "optional": true,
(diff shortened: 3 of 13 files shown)


perrozzi:
This PR fixes a real and subtle bug: layer bundling shipped a mismatched botocore (pulled transitively via `aws-lambda-powertools[tracer] → aws-xray-sdk`) that shadowed the runtime's matched boto3/botocore pair, plus assorted build noise. The fix — a shared venv+strip recipe factored into `lib/utils/python-layer-bundling.ts` and mirrored in `scripts/build-layers.sh`, asset excludes, an exact esbuild pin, and `npm audit fix` on both lockfiles — is well-reasoned and thoroughly documented. Overall this looks solid; the remaining concerns are about drift protection and one behavior change that needs confirmation.

### Should fix

- **Dropped `cp -r . /asset-output/python/` silently removes any first-party files from the layer.** The old CDK bundling copied the entire layer source dir into the deployed layer; `pythonLayerCode()` now only installs from `requirements.txt`. If `lambda/layers/processing-deps` or `ingestion-deps` ever contained (or later gains) shared modules alongside `requirements.txt`, they will no longer ship. The live-smoke verification suggests nothing broke today, but please confirm those dirs contain only `requirements.txt`, and consider a comment in each layer dir (or a synth-time assertion) stating that source files there will NOT be deployed (`voc-datalake/lib/utils/python-layer-bundling.ts`).
- **No automated guard against lockstep drift.** The recipe now exists in two places (`scripts/build-layers.sh` and `python-layer-bundling.ts`) with only cross-referencing comments. The repo rule is that bug fixes get a regression test that fails when reverted; a cheap vitest that asserts the `pythonLayerCode()` bundling command contains the venv step and the `rm -rf .../boto3 .../botocore` strip (and, ideally, greps `build-layers.sh` for the same tokens) would catch both a revert and future drift between the two copies.
- **Stripping boto3/botocore couples layers to the runtime-provided SDK version.** That's the correct default, but if a future `requirements.txt` intentionally pins boto3 (e.g., for a Bedrock API/model newer than the runtime SDK — relevant given the repo's Bedrock model-id/IAM lockstep rule), the strip will silently defeat it. Worth a sentence in the `pythonLayerCode()` doc comment telling future authors how to opt out (vendor boto3+botocore *together* if a pin is ever needed).

### Nits

- `voc-datalake/lib/utils/python-layer-bundling.ts`: `exclude: ['.DS_Store']` only matches at the asset root; `'**/.DS_Store'` would also catch nested ones.
- The `npm audit fix` in `lambda/stream/package-lock.json` bumped vite 8.0.1→8.1.5 and rolldown from RC to stable 1.1.5 — more than a pure vulnerability patch. Tests pass, but note in the commit that dev tooling took a minor bump so future bisects aren't surprised.
- Consider also stripping `boto3`/`botocore` byte-compiled leftovers is unnecessary, but `s3transfer` (runtime-provided too) still ships; harmless duplication, just noting the strip is intentionally minimal.
- The doc comment in `python-layer-bundling.ts` says "Every LayerVersion in the app must use this" — a lint rule or grep-based test for inline `pip install` in `lib/stacks/**` would enforce that claim.

No blockers. The IAM, i18n, ingestor-registration, and Bedrock-lockstep rules aren't touched by this change.

---

perrozzi:
The new push addresses every should-fix and nit from my previous review — READMEs in both layer dirs, an opt-out doc comment for the boto3/botocore strip, `**/.DS_Store`, and a proper drift-guard test suite (`python-layer-bundling.test.ts`) covering the recipe tokens, the build-script lockstep, the layer-dir "no first-party files" invariant, and the "no inline pip install in stacks" rule. It also adds new in-scope work: shared `PY_LAMBDA_ASSET_EXCLUDES` for all `fromAsset('lambda')` staging sites, tightened root-based staging in the ingestion stack, the research Lambda moved to `lambda/`-rooted staging, and a companion guard test.

### Resolved
- Layer-dir README + guard test confirms layer dirs contain only `requirements.txt` (silently-not-deployed risk).
- Lockstep drift guard between `pythonLayerCode()` and `scripts/build-layers.sh` now exists and would fail on revert.
- Opt-out guidance for a future pinned boto3/botocore added to the doc comment.
- `**/.DS_Store` glob fix applied.
- "Every LayerVersion must use this" claim now enforced by the inline-`pip install` test.

### New findings
- `voc-datalake/lib/utils/lambda-asset-excludes.ts` doc comment says "(lib/utils/python-layer-bundling.test.ts enforces the spread)" — the spread is actually enforced by `lambda-asset-excludes.test.ts`. Fix the cross-reference so the guard is discoverable.
- `lambda-asset-excludes.test.ts`: the 400/900-character window heuristics are brittle — a legitimate refactor (e.g., extracting the options object to a variable, or a longer comment before `exclude`) would fail or, worse, silently pass the wrong window. Consider matching on the whole options object (balanced-brace slice) or at least documenting the assumption. Not a blocker since a false failure is loud.
- Minor inconsistency: `ingestion-stack.ts` excludes `'**/test_*.py'` but `PY_LAMBDA_ASSET_EXCLUDES` only excludes `'**/test/**'`/`'**/conftest.py'` — a stray `test_*.py` outside a `test/` dir in `lambda/` would still ship and churn hashes. Cheap to add for symmetry.
- `python-layer-bundling.test.ts`'s inline-pip-install check scans all `.ts` in `lib/stacks` including test files, whereas the excludes test filters them out — a future stack test mentioning "pip install" in a string/comment would fail the suite. Align the filters.

The staging-exclude changes themselves look correct: each bundle's `cp` command only copies dirs that remain staged (jobs bundle keeps `api/` for `projects.py`/prompts; research bundle now mirrors processor/aggregator), and the ingestion-stack root exclude list covers everything the plugin bundling command doesn't copy. Asset hashes will churn once on this deploy, which the author has already verified live. No blockers.

---

mundurragacl:
All three "should fix" items are in `d043afb`, plus a bigger churn defect your review indirectly led me to. Point-by-point:

**Dropped `cp -r .` / first-party files** — Confirmed both layer dirs contain only `requirements.txt` (+ gitignored local build output). Went beyond a comment: each layer dir now has a README stating first-party files will NOT deploy, the helper doc says the same, and `python-layer-bundling.test.ts` has a per-layer-dir test that **fails if a layer dir ever gains an entry outside {requirements.txt, python, README.md, .DS_Store}** — so the failure mode you described can't happen silently.

**Lockstep drift guard** — `lib/utils/python-layer-bundling.test.ts` (8 tests): asserts the venv step, quiet flags, and boto3/botocore strip on the CDK command (parsed out of `AssetCode` with Zod, no type assertions), asserts the same tokens in `scripts/build-layers.sh`, and asserts no `pip install` remains anywhere in `lib/stacks/**` (your 4th nit). Mutation-verified both directions: reverting the helper to the old inline pip fails the suite; deleting a single flag from `build-layers.sh` fails exactly the lockstep test (8→7 pass).

**Strip vs future SDK pins** — Opt-out guidance added to the helper doc comment and both READMEs: never a lone botocore pin; vendor boto3 AND botocore together and drop the strip for that layer.

**Nits** — `'**/.DS_Store'`: done. Dev-tooling bumps: you're right and my PR description was wrong — my lockfile-diff filter only caught added/removed packages, not in-place bumps. Actual stream bumps: vite 8.0.1→8.1.5, rolldown 1.0.0-rc.10→1.1.5, @aws-sdk/* 3.972.15→3.972.36 (audit chain), tinyglobby/fdir minors; **eslint chain untouched** (0 diff lines match eslint, so the #187 lint-drift baseline is unaffected). Recorded in `d043afb`'s commit message for future bisects.

**Declining the s3transfer note with evidence**: s3transfer is not in these layers at all — it's a dependency of boto3 (which we strip), not of botocore, and nothing else in either requirements.txt pulls it. Verified in the built layer (`ls lambda/layers/*/python` → no s3transfer) and in the three deployed layer zips downloaded from Lambda (0 matches).

**Follow-up your review triggered**: while verifying with `cdk diff` I found the README files I'd just added churned the asset hash of ~21 functions — staged-but-never-copied files feed asset hashes, so edits to `lambda/layers/*/python`, `lambda/stream` (141MB node_modules), tests, or `scripts/` were redeploying every Python Lambda with byte-identical code on every deploy (this also explains why deploy logs always showed all functions updating). `d043afb` adds shared `PY_LAMBDA_ASSET_EXCLUDES` to every `fromAsset('lambda')` site, converts ResearchStepLambda from project-root staging to `lambda/`-rooted like its siblings, tightens the root-based plugin bundler, and guards it all with `lambda-asset-excludes.test.ts` (also mutation-verified). Acceptance check: two synths with probe files touched in all four noise paths → identical asset hashes across all 5 stacks, except VocCoreStack's template — that residual churn is the synth-time random admin password, now filed as #196 (it also resets the live admin password on every deploy and prints it in plaintext).

Full gate green: `npm run check` exit 0 (lint, typecheck, 881 backend + frontend + CDK + stream tests).

perrozzi:
The incremental push is a focused response to my last review's four findings: the cross-reference in `lambda-asset-excludes.ts` now points to the correct test file, the brittle fixed-width window heuristic is replaced with a balanced-brace `optionsObject()` parser (with the string-literal caveat documented), `**/test_*.py` was added to the shared excludes (with a matching guard-test assertion), and the inline-pip-install check now filters out `.test.ts` files to match the excludes test.

### Resolved
- Doc-comment cross-reference now names `lambda-asset-excludes.test.ts` as the enforcing test.
- Fixed-width window heuristic replaced with an exact balanced-brace slice of the `fromAsset` options object; failure modes are loud, and the `${...}`-interpolation assumption is documented.
- `**/test_*.py` added to `PY_LAMBDA_ASSET_EXCLUDES` (and to the ingestion stack's already-present pattern's shared counterpart), closing the asymmetry.
- Inline-pip-install scan now excludes test files, aligned with the sibling test's filter.

### New findings
None. The new `optionsObject`/`forEachCallSite` helpers are correct for the current call sites — a bare `fromAsset('lambda')` with no options argument would yield an empty slice and fail the containment assertion loudly, which is the right behavior (it would also mean no excludes). No blockers; this looks ready to merge.

---

PR #197, opened 2026-07-16
fix(core): idempotent admin bootstrap — create once, never touch again

### LINKED ISSUE #196: Admin bootstrap not idempotent: every redeploy prints a fresh never-applied password and churns VocCoreStack
## Problem (corrected after live verification)

core-stack.ts generates a random admin password at **synth time** (lib/stacks/core-stack.ts, `randomInt` pool) and embeds it in the `SetAdminPassword` custom resource properties + the `InitialAdminPassword` output.

What actually happens (verified live on 2026-07-16):

1. The three admin custom resources define only `onCreate`, so redeploys do NOT re-run `AdminSetUserPassword` — the real admin password is **not** reset (my original report claimed it was; wrong).
2. BUT every synth mints a new password and prints it in `InitialAdminPassword` — a password that was **never applied to the user**. Verified: `initiate-auth` with the password printed by the latest deploy → `NotAuthorizedException`. Every output after the first deploy is a lie.
3. The changing property churns `VocCoreStack.template.json` on every deploy (no-op update of `SetAdminPassword` each time) — the only remaining non-deterministic template after #195.
4. A (useless) plaintext password still lands in the template, cdk diff output, and CFN console on every synth.

## Expected (per maintainer)

- **First deployment** (admin does not exist): create the admin, set the initial password, add to admins group, and **print the working password** in `InitialAdminPassword` — that is how operators find their first login. Printing is BY DESIGN.
- **Redeployment** (admin exists): skip user creation and password setting entirely; do not generate or print a new password.
- Template becomes deterministic → VocCoreStack stops churning.

## Direction

Move password generation into the custom resource handler at runtime (first-create only, `AdminGetUser` existence check), expose the real password via `GetAtt` for the output, keep Update/Delete as no-ops.

mundurragacl:
Fixes #196

## Problem

The initial admin password was generated at **synth time** and baked into the `SetAdminPassword` custom resource plus the `InitialAdminPassword` output. Since the three admin custom resources only define `onCreate`, redeploys never actually re-ran the password call — but every deploy **printed a brand-new password that was never applied to the user**. Verified live before the fix: `initiate-auth` with the password printed by the latest deploy → `NotAuthorizedException`. Every `InitialAdminPassword` value after the very first deploy was a lie. The per-synth random value also churned the VocCoreStack template on every single deploy (the last remaining churn after #195).

## Behavior now (per maintainer intent)

- **First deployment** (admin doesn't exist): create the admin (email pre-verified, invite suppressed), set a **temporary** random password generated at runtime inside the handler, add to the `admins` group, and print the real, working password in `InitialAdminPassword`. Printing is by design — it's how operators find their first login, and Cognito forces a change on first use, so the printed value stops working the moment it's used.
- **Redeployment / admin already exists**: strict no-op — no user creation, no password reset, no fresh password minted. The output shows an explicit placeholder: `(unchanged: admin already existed, password not touched)`. Update/Delete events never call Cognito at all.
- All custom resource properties are deterministic → the template stops churning.

## Implementation

- `lambda/custom_resources/admin_bootstrap.py` — 3.2KB unit-tested handler, inlined via `readFileSync` + `Code.fromInline` (no asset bundling), fronted by a `cr.Provider` (same pattern as `ModelAgreementProvider` in BedrockAccessStack). Replaces the three `AwsCustomResource`s and (as a bonus) removes the AWS679 singleton Lambda from the stack.
- Least-privilege role: exactly `AdminGetUser`, `AdminCreateUser`, `AdminSetUserPassword`, `AdminAddUserToGroup` on the pool ARN. cdk-nag suppression scoped to the Provider framework's version-wildcard on this one function (mirrors `ModelAgreementLambda`).
- `custom_resources/**` added to `PY_LAMBDA_ASSET_EXCLUDES` (the file is inlined, never bundled — keeps #195's hash stability intact).
- Output description switched to ASCII: CloudFormation mangles non-ASCII in output descriptions (`—` → `?`), which would have made every future `cdk diff` dirty.

## Tests (all mutation-verified)

- **pytest (6)**: create-when-missing (creates user, temp password — `Permanent` deliberately absent, group add, real password returned), create-when-exists (zero Cognito calls, placeholder), Update/Delete (not even the existence check runs; physical id preserved), password policy compliance, no ambiguous chars. Removing the exists-check fails exactly the touches-nothing test.
- **CDK template tests (3)**: two independent synths are byte-identical (fails on the old synth-time-random code), no `Password` property anywhere in the template, `InitialAdminPassword` wired to `Fn::GetAtt [AdminBootstrap, Password]`. Reintroducing a synth-time password fails the first two.

## Live verification (us-west-2, full logs reviewed)

- Deploy on the existing stack: bootstrap took the skip path; admin's `UserLastModifiedDate` byte-identical before/after; output now shows the placeholder; old CRs + AWS679 singleton deleted cleanly.
- `cdk diff --strict` after deploy: **no differences** — VocCoreStack is deterministic for the first time.
- Greenfield path: invoked the deployed handler with a synthetic Create event for a throwaway username → user created in the right group with `FORCE_CHANGE_PASSWORD` status, and the returned password authenticates (`NEW_PASSWORD_REQUIRED` challenge) — then deleted the throwaway user.
- `npm run check`: exit 0 (lint, typecheck, 887 backend + frontend + CDK + stream tests).

### DIFF
--- a/README.md
+++ b/README.md
@@ -77,7 +77,7 @@ After deployment, an initial admin user is created automatically:
 | Username | `admin` |
 | Password | Check CloudFormation stack outputs for `InitialAdminPassword` |
 
-The password is randomly generated during deployment and stored as a CloudFormation output. Retrieve it with:
+The password is generated when the admin user is first created and stored as a CloudFormation output. Retrieve it with:
 
 ```bash
 aws cloudformation describe-stacks \
@@ -87,6 +87,8 @@ aws cloudformation describe-stacks \
 ```
 
 > 🔒 **Note**: You will be prompted to change this password on your first login.
+> On redeployments the admin user is left untouched — no new user is created, the
+> password is never reset, and the output shows a placeholder instead of a password.
 
 ## ⚙️ Configuration
 
--- a/voc-datalake/lambda/custom_resources/admin_bootstrap.py
+++ b/voc-datalake/lambda/custom_resources/admin_bootstrap.py
@@ -1,0 +1,93 @@
+"""
+Idempotent Cognito admin bootstrap (issue #196).
+
+CloudFormation custom resource handler (cr.Provider protocol). On Create,
+when the admin user does not exist: create it (email pre-verified,
+invitation mail suppressed), set a TEMPORARY random password generated here
+at runtime (Cognito forces a change on first login), add it to the admins
+group, and return the password so the stack can expose it via the
+InitialAdminPassword output — printing it on the very first deployment is
+how operators find their first login.
+
+On Create when the admin already exists (e.g. resource replacement), and on
+every Update/Delete: change NOTHING. Redeployments must never create users
+or touch a live admin's password.
+
+The stack inlines this file (Code.fromInline), so keep it under 4096 bytes.
+"""
+import secrets
+
+import boto3
+
+# No ambiguous characters (0/O, 1/l/I) — operators retype this from a log.
+UPPER = 'ABCDEFGHJKLMNPQRSTUVWXYZ'
+LOWER = 'abcdefghjkmnpqrstuvwxyz'
+DIGITS = '123456789'
+SPECIAL = '!@#$%^&*'
+PASSWORD_LENGTH = 16
+UNCHANGED = '(unchanged: admin already existed, password not touched)'
+
+cognito = boto3.client('cognito-idp')
+
+
+def generate_password():
+    """Random password with every character class the pool policy requires."""
+    classes = [UPPER, LOWER, DIGITS, SPECIAL]
+    chars = [secrets.choice(char_class) for char_class in classes]
+    pool = ''.join(classes)
+    chars += [secrets.choice(pool) for _ in range(PASSWORD_LENGTH - len(chars))]
+    secrets.SystemRandom().shuffle(chars)
+    return ''.join(chars)
+
+
+def admin_exists(user_pool_id, username):
+    try:
+        cognito.admin_get_user(UserPoolId=user_pool_id, Username=username)
+        return True
+    except cognito.exceptions.UserNotFoundException:
+        return False
+
+
+def handler(event, _context):
+    props = event['ResourceProperties']
+    user_pool_id = props['UserPoolId']
+    physical_id = event.get('PhysicalResourceId') or f'admin-bootstrap-{user_pool_id}'
+    no_change = {
+        'PhysicalResourceId': physical_id,
+        'Data': {'Password': UNCHANGED, 'AdminCreated': 'false'},
+    }
+
+    if event['RequestType'] != 'Create':
+        return no_change
+
+    username = props['Username']
+    if admin_exists(user_pool_id, username):
+        return no_change
+
+    password = generate_password()
+    cognito.admin_create_user(
+        UserPoolId=user_pool_id,
+        Username=username,
+        UserAttributes=[
+            {'Name': 'email', 'Value': props['Email']},
+            {'Name': 'email_verified', 'Value': 'true'},
+            {'Name': 'name', 'Value': 'Admin'},
+        ],
+        MessageAction='SUPPRESS',
+    )
+    # Temporary on purpose: first login forces a change, so the printed
+    # password stops working the moment the operator uses it.
+    cognito.admin_set_user_password(
+        UserPoolId=user_pool_id,
+        Username=username,
+        Password=password,
+    )
+    cognito.admin_add_user_to_group(
+        UserPoolId=user_pool_id,
+        Username=username,
+        GroupName=props['GroupName'],
+    )
+    return {
+        'PhysicalResourceId': physical_id,
+        'Data': {'Password': password, 'AdminCreated': 'true'},
+    }
--- a/voc-datalake/lambda/custom_resources/admin_bootstrap.py
+++ b/voc-datalake/lambda/custom_resources/admin_bootstrap.py
@@ -1,25 +1,25 @@
 """
 Idempotent Cognito admin bootstrap (issue #196).
 
-CloudFormation custom resource handler (cr.Provider protocol). On Create,
-when the admin user does not exist: create it (email pre-verified,
-invitation mail suppressed), set a TEMPORARY random password generated here
-at runtime (Cognito forces a change on first login), add it to the admins
-group, and return the password so the stack can expose it via the
-InitialAdminPassword output — printing it on the very first deployment is
-how operators find their first login.
+Custom resource handler (cr.Provider protocol). Decision table on Create:
+  - admin missing: create it (email pre-verified, invite suppressed), set
+    a TEMPORARY runtime-generated password (first login forces a change),
+    add it to the admins group, and return the password so the stack
+    outputs it once — that is how operators find their first login.
+  - admin half-bootstrapped (an earlier run failed after create: never
+    logged in AND no groups): finish the job — fresh password + group.
+  - admin healthy, or Update/Delete: strict no-op. Redeploys never create
+    users or touch a live admin's password.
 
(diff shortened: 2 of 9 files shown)


perrozzi:
This PR replaces the synth-time random admin password (which was printed but never applied on redeploys) with a runtime custom resource that creates the admin exactly once and is a strict no-op thereafter, making the VocCoreStack template deterministic. The design is sound, IAM is genuinely least-privilege (four `cognito-idp` actions scoped to the pool ARN), the asset-exclude hygiene from #195 is preserved, and both the pytest and CDK template tests target the actual load-bearing behavior. Overall a solid change; two operational edge cases below are worth addressing.

### Should fix

- **Password lands in CloudWatch logs via the Provider framework** (`voc-datalake/lib/stacks/core-stack.ts`, `admin_bootstrap.py`). The CDK `cr.Provider` framework function logs the full custom resource response — including `Data.Password` — to `AdminBootstrapProviderLogs` (1-week retention). Exposing the password in the stack output is by design, but the repo rule is no secrets in logs, and the log copy outlives first login (the CFN output stops being useful after the forced change; the log entry persists). Consider setting the provider's `frameworkOnEventLambdaLogLevel`/log level to suppress response bodies, or returning the password through a path the framework doesn't echo, and document the residual exposure either way.
- **Partial-failure on the create path is never repaired** (`voc-datalake/lambda/custom_resources/admin_bootstrap.py`). If `admin_create_user` succeeds but `admin_set_user_password` or `admin_add_user_to_group` fails (throttle, transient error), CloudFormation retries/redeploys hit `admin_exists() → True` and take the strict no-op path forever: the admin exists with an unknown password and possibly no group membership, and the output shows the placeholder. Consider making the skip check stricter (e.g., only skip when the user is confirmed/in the group, or repair group membership idempotently for an existing user) or at least document the manual recovery (`admin-set-user-password` + `admin-add-user-to-group`) in the README.

### Nits

- The docstring in `admin_bootstrap.py` says "keep it under 4096 bytes" (the `Code.fromInline` limit), but nothing enforces it. A one-line CDK or pytest assertion on the file size would fail loudly instead of at synth time when someone grows the handler.
- The handler skips the repo's aws-lambda-powertools pattern (Logger/Tracer/Metrics). That's a reasonable trade-off for an inlined, dependency-free custom resource — a brief note in the module docstring saying so would preempt future "why no powertools?" churn.
- `admin_exists` only catches `UserNotFoundException`; a throttling error fails the whole CR, which is fine, but a comment noting that's intentional (fail loud, CFN retries) wouldn't hurt.
- `test_password_generation_uses_no_ambiguous_characters` runs 50 iterations; asserting `set(UPPER+LOWER+DIGITS) & set('0OlI') == set()` directly on the constants would be deterministic rather than probabilistic.

---

mundurragacl:
Both should-fix items and all four nits are in `442b5db`:

**Partial-failure repair** — Implemented your stricter-skip suggestion as a three-way decision table: admin missing → full create; admin exists but half-bootstrapped (never logged in AND zero group memberships — the exact signature of a create that died between `admin_create_user` and the password/group steps) → finish the job idempotently (fresh temporary password + group add, outcome `repaired`); admin healthy (CONFIRMED, or in any group) → strict no-op. A deliberately-degraded admin can't be hijacked into a repair: CONFIRMED status short-circuits before the group lookup. Live-verified against the deployed handler: manually created a groupless `FORCE_CHANGE_PASSWORD` user, invoked with a synthetic Create event → outcome `repaired`, group present, and the returned password authenticates (`NEW_PASSWORD_REQUIRED` challenge). Mutation-verified: deleting the repair branch fails exactly `test_half_bootstrapped_admin_is_repaired`.

**Password in provider logs** — Checked empirically before changing anything: the provider log group from today's deploys contains only `platform.*` records — no response bodies. The reason is that current aws-cdk-lib defaults `frameworkLambdaLoggingLevel` to `ApplicationLogLevel.FATAL` (the framework's `onEvent returned:` line you're describing logs at INFO and is suppressed). So there is no leak today **in this CDK version** — but your concern is right one dependency bump or debugging session away, so `442b5db` pins `frameworkLambdaLoggingLevel: FATAL` explicitly with a comment stating that raising it echoes `Data.Password` to CloudWatch. Residual exposure is now exactly the two by-design surfaces: the CFN output (maintainer requirement) and CFN's stored custom resource response — both dead after first login anyway, since the password is temporary and forced to change.

**Nits** — All four taken: 4096-byte pytest guard (`test_handler_fits_the_cloudformation_inline_limit`; file is 4003 bytes), deterministic ambiguity assertion on the alphabet constants (replacing the 50-iteration probabilistic loop), powertools-free rationale in the module docstring, and a fail-loud comment on the existence check (anything but `UserNotFoundException` raises so CloudFormation retries/rolls back).

Also added `cognito-idp:AdminListGroupsForUser` (pool ARN only) for the health check. Full gate exit 0 — 890 backend tests. Redeploying the updated handler didn't even invoke the custom resource (deterministic properties), and the live admin's `UserLastModifiedDate` is still byte-identical to the pre-#197 snapshot.

perrozzi:
The push addresses both Should-fix items from my previous review: the Provider framework's log level is now pinned to `FATAL` so `Data.Password` never lands in CloudWatch, and the handler gained a repair path for half-bootstrapped admins (create succeeded, password/group didn't), with the extra `AdminListGroupsForUser` IAM action added in lockstep. All four nits were also picked up (size-limit test, no-powertools rationale in the docstring, fail-loud comment on the existence check, deterministic alphabet test).

### Resolved
- Password exposure in Provider framework logs — `frameworkLambdaLoggingLevel: FATAL` pinned with an explanatory comment (`core-stack.ts`).
- Partial-failure never repaired — `needs_repair` + `finish_bootstrap` handle the half-bootstrapped state, with tests for repaired, healthy-confirmed, and never-logged-in-but-grouped admins.
- Nit: inline size limit now enforced by `test_handler_fits_the_cloudformation_inline_limit`.
- Nit: powertools omission documented in the module docstring.
- Nit: throttling/fail-loud intent documented in `get_admin`.
- Nit: ambiguous-character test now deterministic on the constants.

### New findings
Nothing blocking. Two minor observations:

- **The FATAL pin has no regression test** (`voc-datalake/lib/stacks/core-stack.ts`). The comment says "MUST stay FATAL," but nothing enforces it — a CDK template assertion on the framework function's `LoggingConfig.ApplicationLogLevel` would make the guard the comment describes actually hold, in the same spirit as the size-limit test.
- **Repair can reset a deliberately de-grouped admin** (`admin_bootstrap.py`): an admin who never logged in and was intentionally removed from all groups would get a fresh password and re-added to `admins` on a stack recreate. This is an unlikely state and the heuristic (FORCE_CHANGE_PASSWORD + zero groups) is a reasonable proxy for "half-bootstrapped," so this is acceptable — just noting the trade-off is real.

Otherwise the incremental changes are clean and well-tested; the refactored handler is clearer than the original and stays within the inline limit.

---

perrozzi:
Since the last review, a single test was added to `voc-datalake/lib/stacks/core-stack.test.ts` pinning the provider framework's `ApplicationLogLevel` to `FATAL` via a template assertion — this addresses the first of my two minor observations from the previous review. The second observation (the repair heuristic can re-bootstrap a deliberately de-grouped admin) was already accepted as a reasonable trade-off and needed no change.

### Resolved
- FATAL pin now has a regression test — `Match.stringLikeRegexp` on the framework function's description plus `LoggingConfig.ApplicationLogLevel: 'FATAL'` assertion, matching the spirit of the size-limit test.

### New findings
None. The test correctly targets the `AdminBootstrapProvider` framework function specifically (rather than any Lambda in the stack), and the comment explains why the pin matters. The incremental change is clean; no further concerns.

---

PR #199, opened 2026-07-16
fix(stream): zero out the 176-error lint drift and gate it

### LINKED ISSUE #187: [Enhancement]: lambda/stream lint drift — 176 vitest-plugin violations in test files

mundurragacl:
Fixes #187

`lambda/stream`'s eslint ran in no CI path, so violations grew silently to **176 errors — and not only in test files as the issue assumed: 25 sat in production sources** (type assertions, banned `let`s, a sonar-flagged backtracking regex, complexity/max-depth/max-lines breaches).

## What changed

**Production sources (25 errors → 0), all real fixes:**
- `attachments.ts`: typed `IMAGE_FORMATS` map replaces both `as` format casts
- `model-override.ts`: `fetchSettings()` extraction kills two `let`s; DynamoDB item casts → lenient Zod record parse (repo boundary pattern)
- `project-context.ts`: `isPlainRecord` type guard replaces an entries cast; recursive `stripTrailingSlashes` replaces the backtracking-prone regex; roundtable persona prompt builder split into new `persona-prompt.ts` (fixes max-lines 413>400 and complexity 13>12; type-only import back = no runtime cycle)
- `search-feedback.ts`: recursive `fetchDayPages()` replaces do/while pagination (`let`, max-depth 4–5>3), `resolveSearchParams()` extraction brings complexity 15→≤12
- `converse-stream.ts` `??=` memoization; `streaming.ts` generic `Error` → `ConfigurationError`

**Tests (151 errors → 0):** `toStrictEqual` at all 40 flagged sites; 37 bare `Function` types → real signatures; nine no-op `for await (const _ …)` drains → `drainStream()` helper; `schema.test.ts` safeParse if-blocks → `parse()`-based assertions (13 conditional-expect errors); yield-less mock generator → hand-rolled rejecting async iterable. **Caught during this work:** eslint's `prefer-called-with` autofix had silently rewritten `toHaveBeenCalled()` → `toHaveBeenCalledWith()` (asserts *zero args*), breaking 3 tests — replaced with meaningful assertions.

**Config (rationale in-file, tuning sanctioned by the issue):** `vitest/max-expects` 4→10 — these are integration-style suites asserting whole SSE event sequences; 10 is the observed ceiling and splitting fragments one behavior across test names. `no-unused-vars` gets `argsIgnorePattern: '^_'`, matching the existing naming-convention underscore allowance.

**Gate:** new `lint:stream` + `test:stream` root scripts wired into `npm run check` — the drift structurally cannot restart. Also noticed stream *tests* weren't in the gate either; now they are.

## Verification
- `npx eslint src/`: 0 problems; stream tests 230/230
- Root `npm run check`: exit 0 (all suites)
- Mutation-verified: reintroducing a single violation fails `lint:stream` (and thus `check`); the 3 autofix-broken tests failed before their assertion fixes and pass after


perrozzi:
This PR fixes 176 accumulated eslint violations in `lambda/stream` (25 in production sources, 151 in tests), extracts a few helpers to satisfy complexity/size budgets, and wires `lint:stream`/`test:stream` into the root `check` script so the drift can't recur. The production refactors are behavior-preserving as far as I can trace (the `attachments.ts` format map, `model-override.ts` cache split, `search-feedback.ts` pagination recursion, and the `persona-prompt.ts` extraction all reproduce the original semantics), and the direction — Zod at the DynamoDB boundary, `isPlainRecord` type guard replacing `as` casts — matches repo patterns. No blockers found.

### Should fix

- **`package.json` — `test:stream` may hang in watch mode.** The frontend equivalent needs `npm run test ... -- --run` to force a single vitest run; `test:stream` has no such flag. If `voc-datalake/lambda/stream/package.json`'s `test` script is bare `vitest`, `npm run check` will hang in CI. Please confirm it's `vitest run` (the stated "exit 0" verification suggests it is, but the stream `package.json` isn't in the diff), or add `-- --run` for symmetry.
- **`package.json` — gate depends on callers using `lint:all`/`check`.** Any CI workflow that invokes `npm run lint` directly still skips stream lint. Consider either making root `lint` include stream (and keeping `lint:all` as an alias), or confirming the CI workflow calls `check`/`lint:all`. Otherwise the "drift structurally cannot restart" claim only holds for one entry point.

### Nits

- `search-feedback.ts` `fetchDayPages`: the doc comment says recursion is "bounded by the MAX_CANDIDATES stop," but if every item on every page fails `safeParse`, `candidates.length` never grows while `LastEvaluatedKey` keeps advancing — pagination is bounded only by table size. Pre-existing in the `do/while`, but the comment now overstates the guarantee.
- `search-feedback.ts` `resolveSearchParams` return type widens `mode` to `string` (and `input` is returned alongside its own destructured fields); typing `mode` as the schema's union would keep downstream switches exhaustive.
- `eslint.config.js` `vitest/max-expects` 4→10 is a global loosening for the whole stream package; per-file `/* eslint vitest/max-expects: ["error", {max:10}] */` overrides on the integration-style suites would keep the tighter default for future unit tests. Acceptable as-is since the issue sanctioned it.
- `converse-stream.test.ts` `drainStream` could be a plain `for await (const _event of stream) { /* drain */ }` with the new `argsIgnorePattern` — the manual iterator is more code than the lint rule required. Not worth churn now.
- `persona-prompt.ts` ↔ `project-context.ts` form a module cycle broken only by `import type` erasure; the in-file comment covers it, but moving `ProjectItem` (or its schema) to a shared module would remove the footgun if someone later imports a value across that edge.

---

perrozzi:
The changes since my last review are small and targeted: the root `lint` script now runs both frontend and stream lint (with `lint:frontend` split out and `lint:all` removed), the `fetchDayPages` doc comment no longer overstates the MAX_CANDIDATES bound, and `resolveSearchParams`'s `mode` is now typed as the `'list' | 'aggregate'` union. These directly address one Should-fix and two of my nits.

### Resolved
- Gate entry point: root `lint` now includes stream lint, so `npm run lint` alone can no longer skip it.
- `fetchDayPages` comment corrected — no longer claims MAX_CANDIDATES bounds pagination for all-malformed pages.
- `resolveSearchParams` `mode` narrowed to `'list' | 'aggregate'`, keeping downstream switches exhaustive.

### Still open
- `test:stream` watch-mode risk: the stream `package.json` is still not in the diff, so I can't confirm its `test` script is `vitest run`; add `-- --run` for symmetry or confirm.

### New findings
- `package.json`: `lint:all` was removed rather than aliased. If any CI workflow, docs, or contributor muscle memory references `npm run lint:all`, it now fails with "missing script." Consider keeping `"lint:all": "npm run lint"` as an alias for one release, or verify nothing references it. Minor.

Otherwise the incremental changes are clean and behavior-preserving.

---

mundurragacl:
Both should-fixes handled in `f6029bd`, one by change and one by evidence:

**`test:stream` watch-mode hang** — No change needed: `voc-datalake/lambda/stream/package.json` line 10 already reads `"test": "vitest --run"` (single-run), which is why the full gate exits. It wasn't in the diff because it didn't need changing.

**Gate entry points** — Taken, and stronger than suggested: root `lint` itself now runs frontend + stream (`lint:frontend` + `lint:stream` subcommands), `lint:all` is gone, and `check` uses `lint`. There is no lint entry point left that skips the stream package. Verified: full `npm run check` exit 0 with the stream lint step visible in the log.

**Nits:**
- `fetchDayPages` comment — fixed: it now states the truth (recursion depth = pages in the day's partition; MAX_CANDIDATES stops early only as parsed rows accumulate; a fully-malformed day pages to its end, same as the previous do/while).
- `resolveSearchParams` — `mode` is now typed as the schema union `'list' | 'aggregate'`.
- Per-file `max-expects` pragmas — declining with a config citation: the package's comments policy (`@eslint-community/eslint-comments/no-use` in `eslint.config.js`) allows only `eslint-disable`/`eslint-enable`/`eslint-disable-next-line`; an inline `/* eslint vitest/max-expects: ... */` configuration comment is itself a lint error here. A tighter default would require the disable-comment style the config bans, so the documented package-level 10 stands.
- `drainStream` simplification — declining as you anticipated: `argsIgnorePattern` covers function *args* only; a `for await (const _event ...)` binding is a variable, and `sonarjs/no-unused-vars` (recommended set) flags it regardless of eslint ignore patterns — it has no such option.
- `ProjectItem` shared module — agreed it's the cleaner end state; keeping the documented `import type` edge for now since extracting the Zod schema drags half the boundary parsing with it. Will fold into any future context-module refactor.

mundurragacl:
Both leftovers settled in `dbfea63`:

**`test:stream` watch-mode** — Now visible in the diff itself: `voc-datalake/lambda/stream/package.json` changes `"test": "vitest --run"` → `"test": "vitest run"` (canonical subcommand, identical single-run behavior). The script was already single-run — it just wasn't in the diff for you to see; now it is. Verified: `npm test` in the package runs once and exits (230/230).

**`lint:all` alias** — Declining with history: `git log --all -S 'lint:all'` shows the script exists *only in this PR's own commits* (born `e2ca27c`, removed `f6029bd`); it never shipped to development, so no workflow/docs/muscle memory can reference it (repo-wide grep: zero hits outside this branch). An alias would preserve compatibility with something that never existed.

perrozzi:
The only change since my last review is a one-line edit to `voc-datalake/lambda/stream/package.json`, switching the `test` script from `vitest --run` to `vitest run`. This directly addresses my remaining open item by making the stream package's test script visible and confirming it runs in single-pass (non-watch) mode, so `test:stream` in the root `check` gate cannot hang CI.

### Resolved
- `test:stream` watch-mode risk: stream `test` script is now explicitly `vitest run`, confirmed non-interactive.

### New findings
None — `vitest run` and `vitest --run` are behaviorally equivalent, so this is a safe, purely stylistic normalization.

The incremental change is clean; no further action needed from my side. (The minor `lint:all` alias note from last time was informational only and doesn't block.)

---

PR #200, opened 2026-07-16
fix(i18n): CategoriesManager strings get keys ×8 locales + parity guard

### LINKED ISSUE #183: [Enhancement]: CategoriesManager hardcoded English strings need i18n keys
Raised on #182's review: CategoriesManager renders hardcoded English strings ('N sub', 'N categories', 'No categories configured yet.', confirm-modal copy, etc.) that predate #182 and aren't behind the components i18n namespace, unlike sibling components. Follow-up: move them to `components.json` keys ×8 locales and update the tests that assert literal copy.

mundurragacl:
Fixes #183

CategoriesManager rendered ~19 hardcoded English strings (AI-suggestions card, `{n} categories` / `{n} sub` counts, empty state, placeholders, save status, and the delete confirm-modal copy) while sibling components use the `components` i18n namespace.

## What changed
- All strings flow through `useTranslation('components', { keyPrefix: 'categoriesManager' })`; counts use i18next pluralization (`categoryCount_one/_other` — CJK locales carry only `_other`, per i18next plural rules)
- `categoriesManager` group added to `components.json` in **all 8 locales** (en, es, fr, de, pt, ja, ko, zh)
- **New `src/i18n/localeParity.test.ts`** — every namespace must expose the same plural-collapsed key set in every locale. Scoped to the whole locales tree (a scan showed zero pre-existing drift), so any future half-translated feature fails CI, not just this component. Guarded against vacuous passes (asserts 7 non-English locales, nonzero namespaces).
- Existing CategoriesManager tests keep asserting real English copy: the test setup loads the real `en` bundle, so those assertions now cover the key wiring end-to-end — a missing/renamed key renders its raw name and fails them.
- Incidental: adding the hook re-enabled react-compiler purity analysis on this component, which flagged the two pre-existing id-minting `Date.now()` calls in event handlers; extracted a module-level `makeEntryId()` (behavior unchanged).

## Verification
- CategoriesManager 49/49, parity 17/17, full root `npm run check` exit 0
- Mutation-verified: deleting one key from `de/components.json` fails exactly the components.json parity test; restoring returns green


perrozzi:
This PR replaces ~19 hardcoded English strings in `CategoriesManager.tsx` with `useTranslation('components', { keyPrefix: 'categoriesManager' })` keys, adds the `categoriesManager` group to all 8 locale files, introduces a locale-parity test, and extracts a `makeEntryId()` helper. The change aligns with the repo's i18n rules, translations look complete and consistent (CJK locales correctly carry only `_other`), and the parity guard plus mutation-verified test coverage is a genuinely good addition. No blockers.

### Should fix
- **Parity guard is blind to missing plural forms.** `baseKeys` in `voc-datalake/frontend/src/i18n/localeParity.test.ts` collapses `_one/_other/...` to the base key, so a locale that ships only `categoryCount_one` (missing `_other`), or a plain non-plural `categoryCount`, still passes even though it breaks at runtime for `count > 1`. Consider validating that each locale carries the plural categories required by `new Intl.PluralRules(locale).resolvedOptions().pluralCategories` for keys that are pluralized in `en`.
- **Locale discovery is fragile.** `fs.readdirSync(LOCALES_DIR).filter((entry) => entry !== 'en')` will pick up any stray non-directory entry (e.g., `.DS_Store`, a README) and then throw inside `readNamespace` with a confusing ENOTDIR/ENOENT. Filter to directories (`withFileTypes: true` + `isDirectory()`).

### Nits
- `subCount` uses `{ count }` but only a base key exists (`"{{count}} sub"`), relying on i18next's base-key fallback. Since `categoryCount` got proper `_one/_other` forms, do the same for `subCount` — it also fixes the English grammar ("3 subs" vs "3 sub").
- `expect(locales).toHaveLength(7)` will need a manual bump when a locale is added or removed. Intentional as an anti-vacuous guard, but a short comment naming the 8 supported locales would make the failure self-explanatory.
- `makeEntryId` still mints ids from `Date.now()`, so two adds under the same prefix within the same millisecond collide (plausible with paste-and-enter or tests). `crypto.randomUUID()` or a module-level counter suffix would close this pre-existing hazard cheaply.
- `PLURAL_SUFFIX` (`/_(one|other|zero|two|few|many)$/`) will also collapse legitimately named keys like `step_two`; unlikely today, but worth knowing the parity test can silently merge such keys.

---

mundurragacl:
Both should-fixes and all four nits are in `9006b8b`:

**Plural blindness** — Implemented, with one adjustment your Intl.PluralRules suggestion surfaced when I tried it: the strict version (every en-pluralized key must be suffixed everywhere, no bare variants) immediately failed on **pre-existing, deliberate repo patterns** — the en files themselves pair bare fallbacks WITH plural forms (e.g. `categories.json` has `issues`, `issues_one`, `issues_other` — 7 namespaces do this). So the rule now checks what actually breaks at runtime: every key en pluralizes must be *renderable* in every locale — an `_other` form or a bare fallback; a locale left with only `_one` (your exact scenario) fails. Full Intl.PluralRules categories deliberately not required: fr/es/pt would demand `_many`, which en itself doesn't provide and these small-count UIs never hit (rationale in-file). Mutation-verified: stripping `subCount_other` from `de` fails exactly the components.json check.

**Locale discovery** — `withFileTypes: true` + `isDirectory()`; a stray `.DS_Store` can't masquerade as a locale.

**Nits, all taken:**
- `subCount` now has real `_one`/`_other` forms in all locales (en: `1 sub` / `3 subs` — your grammar fix; CJK only `_other`); the two test literals updated accordingly.
- The anti-vacuous guard now asserts the exact 8-locale set (`['de','en','es','fr','ja','ko','pt','zh']`) with a comment saying to bump it consciously.
- `makeEntryId` mints `crypto.randomUUID()` — mirroring the chatStore #160 fix for the identical same-millisecond hazard, comment cross-references it.
- `PLURAL_SUFFIX` collapse caveat (`step_two`-style names) documented at the constant.

Full gate exit 0; parity 33/33, CategoriesManager 49/49.

perrozzi:
The new push adds proper `subCount_one/_other` plural forms across all locales (CJK correctly `_other`-only), hardens the parity test with a pluralization-renderability check and directory-only locale discovery, replaces the length-7 guard with an explicit locale list, switches `makeEntryId` to `crypto.randomUUID()`, and documents the `PLURAL_SUFFIX` caveat. Every finding from the previous review was addressed.

### Resolved
- Parity guard now verifies each en-pluralized key has an `_other` form or bare fallback in every locale (`localeParity.test.ts`); the pragmatic scoping vs. full `Intl.PluralRules` is well-reasoned and documented.
- Locale discovery filters to directories via `withFileTypes: true` + `isDirectory()`.
- `subCount` gained proper `_one/_other` forms in all 8 locales, fixing English grammar ("2 subs"); existing tests updated accordingly.
- `expect(locales).toHaveLength(7)` replaced with an explicit sorted list of the 8 supported locales — failures are now self-explanatory.
- `makeEntryId` uses `crypto.randomUUID()`, closing the same-millisecond id collision.
- `PLURAL_SUFFIX` key-name collision caveat is documented in a comment.

### New findings
None blocking. One micro-nit: in the renderability test, `readNamespace(locale, namespace)` is parsed twice per locale (once for `suffixedKeysOf`, once for `collectKeys`); harmless in a test but a single read would do. The mutation-verification claim in the description matches the assertion structure — the delta looks clean.

---

perrozzi:
The only change since my last review is a small refactor in `localeParity.test.ts` addressing the micro-nit I raised: the renderability test now reads each namespace file once per locale into a `content` variable and passes it to both `suffixedKeysOf` and `collectKeys`, instead of parsing it twice.

### Resolved
- Duplicate `readNamespace` parse per locale in the renderability test is eliminated (`src/i18n/localeParity.test.ts`).

### New findings
None. The change is behavior-preserving and the delta is clean — this PR looks ready to merge.

---

PR #202, opened 2026-07-16
fix(quality): zero out repo-wide lint/test debt and gate every surface

### LINKED ISSUE #201: Repo-wide lint/test debt: 102 ruff errors ungated, plugin tests outside the gate, silently-skipped test, frontend warnings
## Findings (development @ 29a1186)

1. **Python lint was never wired up**: `ruff check lambda/ plugins/` reports 562 errors — 460 in vendored layer build output (`lambda/layers/*/python`, must be excluded, not fixed) and **102 real first-party errors**: 82 unused imports (F401), 7 module-import-not-at-top (E402), 7 pointless f-strings (F541), 3 redefined-while-unused (F811), 3 unused variables (F841). No ruff config exists; requirements-dev.txt declares black/isort/mypy which nothing runs.
2. **~205 plugin tests are excluded from every run**: pytest.ini sets `testpaths = lambda` and `--ignore=plugins`; the root gate runs `pytest lambda/`. All plugin tests pass when included (verified: 1095 passed).
3. **One test silently skips everywhere**: `lambda/shared/test/test_prompt_utils.py:27` — 'No prompts dir'.
4. **Frontend lint carries 3 warnings** (react-refresh/only-export-components in RouteErrorBoundary + ScraperCard ×2) and warnings don't fail the gate, so they accumulate.

## Fix
Zero out all first-party findings, add a ruff config (exclude vendored output), wire `ruff check` into the root gate, include plugins/ in the default pytest run + gate, un-skip the prompts test, fix the frontend warnings and make frontend lint fail on warnings.

mundurragacl:
Fixes #201

Full-ownership cleanup: every linter and test surface in the repo now reports zero findings, and every one of them is enforced by the root gate so the debt cannot silently return.

## Python — there was no linter at all
Raw `ruff check lambda/ plugins/` reported **562 errors**; 460 were vendored third-party code in the local layer build output (`lambda/layers/*/python` — bs4, six, …), now excluded via a new `ruff.toml` (default F + E4/E7/E9 rules). The **102 first-party errors were fixed for real**, not suppressed:
- 82 unused imports removed (the full 1096-test suite passed afterwards — none were load-bearing)
- 7 pointless f-strings, 3 redefined-while-unused, 3 unused variables — two dead `body` assignments in handler tests became actual response-body assertions, strengthening those tests
- 7 E402s fixed structurally: the processor's two-statement `sys.path` extension collapsed into the single-call form ruff's E402 exempts (same pattern the API handlers already use); a mid-file `import logging` hoisted
- `requirements-dev.txt` declared black/isort/mypy — repo-wide grep shows **nothing ever ran them**; replaced with ruff, the linter the gate now enforces

## Tests — ~205 plugin tests were excluded from every run
`pytest.ini` had `testpaths = lambda` + `--ignore=plugins`. Now `testpaths = lambda plugins` (with `_template` excluded), coverage spans both trees, and the root `test:backend` runs it all: **1096 passed, 0 skipped**.

The one silently-skipped test exposed a real defect: `shared/prompts.py`'s local-dev branch resolved `voc-datalake/prompts` — a directory that doesn't exist (prompts live in `lambda/api/prompts`). Fixed the resolver, and `test_real_resolution` now hard-asserts instead of `pytest.skip`, so the regression can't hide again. (Runtime was never affected: bundles place prompts at `/var/task/prompts`, the first branch.)

## Frontend — 3 warnings, invisible because warnings never failed anything
- `describeRouteError` extracted from `RouteErrorBoundary/index.tsx`; `normalizedBaseUrl`/`scraperDomainLabel` extracted from `ScraperCard.tsx` into `scraperUrl.ts` — the react-refresh mixed-export warnings are fixed at the root, not silenced
- Frontend lint is now `eslint . --max-warnings 0`

## Gate
Root `lint` = frontend + stream + **python**; `test:backend` includes plugins. **Mutation-verified**: an unused Python import fails `lint:python`; a react-refresh warning fails `lint:frontend`; a broken plugin test fails `test:backend`. Full `npm run check`: exit 0.

## Deliberately out of scope
- Reformatting (`ruff format`/black): a whole-repo reformat destroys blame and belongs in its own deliberate change, if ever
- Adding eslint to the CDK package (`lib/`): a new lint surface with its own rule-set decisions — happy to file as follow-up


perrozzi:
This PR introduces Python linting (ruff) as a gated surface, fixes 102 first-party lint findings (mostly unused imports and pointless f-strings), un-excludes ~205 plugin tests from pytest, fixes a real path-resolution defect in `shared/prompts.py` with a hardened regression test, and resolves react-refresh mixed-export warnings by extracting pure functions into their own modules. The changes are mechanical, well-scoped, and consistent with the repo's patterns (single-call `sys.path` setup, vendored layer output excluded from lint). Overall this looks solid; no blockers found.

### Should fix

- **`package.json` `lint:python` hardcodes `.venv/bin/ruff`.** This breaks on Windows (`.venv/Scripts/ruff.exe`) and fails with a confusing "no such file" error for any contributor whose existing venv predates this PR (ruff was just added to `requirements-dev.txt`). Prefer `.venv/bin/python -m ruff check lambda plugins` (consistent with `test:backend`'s interpreter invocation) and/or a note in the README/CI that `pip install -r requirements-dev.txt` must be rerun. Also confirm CI installs dev requirements before the gate runs, otherwise root `lint` now fails there.
- **`voc-datalake/lambda/shared/prompts.py`:** the local-dev branch now resolves `shared/../api/prompts`. Runtime is claimed safe because bundles place prompts at `/var/task/prompts` (first branch), which is plausible, but please double-check non-API consumers of `shared/prompts` (e.g. `jobs/document_generator`) — if any bundle ships the `api/` tree without its prompt files, the new branch could match a wrong/empty directory before the cwd fallback. A one-line comment in the bundling script asserting the `/var/task/prompts` invariant would lock this in.
- **`voc-datalake/lambda/shared/test/test_prompt_utils.py`:** the new `assert any(prompts_dir.glob('*.json'))` bakes in the assumption that prompts are `.json`. If a prompt format ever changes (e.g. `.md` templates), this test fails for the wrong reason. Asserting `any(prompts_dir.iterdir())` (non-empty) would be equally strong for the regression being guarded.

### Nits

- `voc-datalake/pytest.ini`: `--ignore=plugins/_template` in `addopts` and `_template` in `norecursedirs` are redundant — one suffices.
- `voc-datalake/pytest.ini`: `--cov=plugins` now includes `plugins/_template` in coverage reporting even though its tests are excluded; harmless with `--cov-fail-under=0` but slightly noisy. Consider `[coverage] omit` for `_template`.
- `test_metrics_handler.py` `test_respects_limit_parameter`: the added `assert 'items' in body` is better than the dead `body` assignment, but it still doesn't verify the limit was actually respected (e.g. asserting `Limit` in the mocked query kwargs).
- `requirements-dev.txt`: `boto3-stubs` is retained but with mypy removed nothing consumes the stubs; fine to leave, but worth a follow-up decision (either wire up a typechecker or drop the stubs).
- `ruff.toml`'s baseline (F + E4/E7/E9) is minimal; consider a follow-up issue to evaluate `I` (import sorting) and `B` (bugbear) now that the gate exists.

The claimed mutation verification of the gate (unused import → lint failure, broken plugin test → `test:backend` failure) is the right way to land this; nice work strengthening the two dead-assignment tests into real response-body assertions rather than deleting them.

---

mundurragacl:
All three should-fixes and three of the nits are in `bf52d96`:

**`lint:python` portability** — Now `.venv/bin/python -m ruff check lambda plugins`, matching `test:backend`'s interpreter invocation exactly. (There is no CI workflow in this repo to update — the gate is `npm run check`, and README's dev setup already tells contributors to install `requirements-dev.txt`; a stale venv now fails with Python's clear "No module named ruff" instead of a confusing ENOENT.)

**Prompts consumers audit** — Checked every importer of `shared.prompts`: `research_step_handler` only uses `get_response_language_instruction`, a pure string-builder that never resolves the directory — so the research bundle (which ships no prompts) is unaffected, same as before the fix. Every dir-resolving consumer (API handlers, job lambdas, `avatar`/`product_context`/`projects`) ships in a bundle that places prompts at the root, hitting the `/var/task/prompts` branch first. Both bundling sites in `api-stack.ts` now carry the INVARIANT comment you suggested.

**Format-agnostic test** — `iterdir()` instead of `glob('*.json')`.

**Nits taken** — redundant `--ignore=plugins/_template` dropped (norecursedirs covers it); new `.coveragerc` omits `plugins/_template` and `lambda/layers` (verified: zero `_template` lines in the coverage table); `test_respects_limit_parameter` now asserts `Limit == 5` reaches the mocked query — verifying what its name claims.

**Declined / follow-up:**
- `boto3-stubs`: agreed it's orphaned without mypy, but dropping it also degrades editor IntelliSense for contributors — leaving it as a deliberate dev-experience dependency; noted for a future typechecker decision.
- Ruff `I`/`B` rule expansion: out of scope for a zero-the-debt PR — happy to file the follow-up once this lands, so the evaluation starts from a green baseline.

Full gate re-verified: exit 0, 1096 backend tests, zero warnings anywhere.

perrozzi:
Since the last review, the author switched `lint:python` to the `python -m ruff` invocation, replaced the format-specific glob assertion with `iterdir()`, strengthened the metrics limit test to assert the `Limit` kwarg reaches the query, added `.coveragerc` to omit `_template` and vendored layers, de-duplicated the `_template` exclusion in `pytest.ini`, and added the `/var/task/prompts` invariant comments to both bundling commands in `api-stack.ts`. Nearly all previous feedback was addressed.

### Resolved
- `lint:python` now uses `.venv/bin/python -m ruff`, consistent with `test:backend`.
- `test_prompt_utils.py` asserts non-emptiness via `iterdir()` instead of baking in `.json`.
- `test_respects_limit_parameter` now verifies the limit actually reaches the DynamoDB query.
- `pytest.ini` redundant `--ignore=plugins/_template` removed; `_template` coverage noise handled via new `.coveragerc`.
- `api-stack.ts` bundling commands now document the `/var/task/prompts` invariant that `shared/prompts.py` depends on.

### Still open
- The `python -m ruff` form still assumes a POSIX venv layout (`.venv/bin/`), and nothing in the diff confirms CI installs `requirements-dev.txt` (where ruff was just added) before the root `lint` gate runs — please verify CI, or the gate fails there on first run.

### New findings
None — the incremental changes are clean. The `.coveragerc` comments correctly cross-reference `norecursedirs`, and the `api-stack.ts` additions are comment-only with no behavioral change.

---

mundurragacl:
Fair push — while this repo has **no actual workflow** (`.github/` contains only issue templates, so there is nothing that runs the gate in CI today), the docs DO ship an example GitHub Actions workflow, and you made me look at it properly: it installs only npm dependencies, meaning it was **already broken for `test:backend`** (which has always invoked `.venv/bin/python -m pytest`) — not just for the new `lint:python`. Fixed in the diff (`docs/deployment.md`): the example now adds `setup-python`, creates the venv, and installs `requirements-dev.txt` plus the layer requirements before `npm run check`, with a comment naming the two gate steps that need it. On the POSIX layout: `.venv/bin/` matches the repo's pre-existing convention (`test:backend` has used it since before this PR) and the documented dev platforms; a Windows-native flow would need the same treatment for pytest too, and is best handled as its own decision if ever needed.

perrozzi:
The only change since the last review is a documentation update to `docs/deployment.md`: the example CI workflow now sets up Python 3.12, creates the venv, and installs `requirements-dev.txt` plus both layer requirement files before the gate runs — directly addressing the remaining open item about CI having ruff/pytest available.

### Resolved
- CI documentation now shows the venv + `requirements-dev.txt` install step preceding the gate, so `lint:python` and `test:backend` won't fail on a missing `.venv/bin/` toolchain (per the documented workflow).

### New findings
- Minor (non-blocking): the fix lands in `docs/deployment.md`, which is an example workflow. If this repo also carries an actual `.github/workflows/*.yml` running `npm run check`, that file needs the identical steps or the gate still fails there — please confirm no live workflow was missed (none appears in the diff).
- Nit: the added YAML lines carry trailing whitespace, consistent with the surrounding snippet — fine to leave.

Otherwise the incremental change is clean and comment/docs-only; no behavioral risk. The `.venv/bin/` POSIX assumption remains, but with Linux runners documented explicitly it's acceptable.

---

mundurragacl:
Confirmed with the tree itself: `find .github -type f` lists only `ISSUE_TEMPLATE/*.yml` (bug_report, config, feature_request, infrastructure, missing_tests, refactor, security) — there is **no `.github/workflows/` directory and no live workflow** in this repository. The docs example is the only CI definition that exists, and it now carries the venv setup. Nothing was missed.

PR #204, opened 2026-07-16
fix(cdk): kill the eternal ingestor hash churn — dir/** excludes leak dot-children

### LINKED ISSUE #203: Ingestor asset hashes churn forever: 'dir/**' excludes leak dot-children, so cdk.out/.cache feeds the next fingerprint
## Root cause (proven empirically, development @ ff415dc)

CDK's GlobIgnoreStrategy matches excludes with `minimatch(relativePath, pattern, { matchBase: true })` — and `**` **never crosses a dot segment**. Every `'<dir>/**'`-style exclude therefore fails to exclude that directory's dot-children.

For the five plugin-ingestor assets (staged from the project root by `bundlePluginCode`), the killer instance is `'cdk.out/**'`: it does NOT exclude `cdk.out/.cache/*.zip` — **CDK's own asset-publishing cache**. The feedback loop:

1. deploy publishes asset zips → writes new files into `cdk.out/.cache/`
2. next synth stages the root → fingerprint now includes the new zips → all 5 ingestor assets get new hashes
3. deploy publishes those → more cache zips → goto 1

This explains every symptom chased across deploys 12–19: consecutive same-command invocations agree (no publish in between), `cdk diff` vs `cdk deploy` disagree (each publishes at different points), and VocIngestionStack re-deployed byte-identical code on almost every `deploy --all`.

**Evidence**: a probe using aws-cdk-lib's own `FileSystem.copyDirectory`/`fingerprint` with the exact exclude list shows 95→99 `cdk.out/.cache/*.zip` files staged across one no-op deploy, with the root fingerprint flipping `e3c5bca0… → 6a6038c3…` and **no first-party file changing**. The staged tree also contains `node_modules/.bin`, `.venv/.gitignore`, `.venv/.DS_Store` — same leak class, currently static but one tool-run away from the same churn (`'**/.DS_Store'` misses dot-dir-nested copies too, since `**` can't cross `.venv`).

## Fix

Exclude **directories by name** (bare or path-qualified — both prune the whole subtree via `completelyIgnores`, dot-children included) instead of `'<dir>/**'`; use bare names instead of `'**/<name>'` for any-depth matches (matchBase applies to slash-less patterns only). Centralize the root-staging exclude list next to PY_LAMBDA_ASSET_EXCLUDES, and add a behavior-level regression test using aws-cdk-lib's `IgnoreStrategy.glob` asserting `cdk.out/.cache/x.zip` (and friends) are ignored while real plugin/shared payloads are not.

mundurragacl:
Fixes #203

## The bug that survived four sessions of deploys

The five plugin-ingestor Lambdas redeployed byte-identical code on nearly every `cdk deploy --all`, and `cdk diff` systematically disagreed with `cdk deploy` about their asset hashes. Root cause, finally proven with aws-cdk-lib's own machinery:

**CDK matches asset excludes with `minimatch(relativePath, pattern, { matchBase: true })`, and `**` never crosses a dot segment.** So `'cdk.out/**'` does NOT exclude `cdk.out/.cache/*.zip` — CDK's own asset-publishing cache. The feedback loop: deploy publishes zips into `.cache` → the next synth stages the project root for the ingestor assets → the fingerprint includes the new zips → new hashes → deploy publishes again → forever. A probe using `FileSystem.copyDirectory`/`fingerprint` with the exact exclude list showed **95→99 cache zips staged across one no-op deploy**, root fingerprint flipping, zero first-party changes. `node_modules/.bin`, `.venv/.gitignore` were staged through the same hole (static today, one tool-run from churning).

## Fix

Exclude directories **by name** — bare (`'cdk.out'`; matchBase applies slash-less patterns to basenames at any depth) or path-qualified (`'lambda/layers'`) — which prunes the whole subtree via `completelyIgnores`, dot-children included. Never `'dir/**'`.

- `ROOT_PLUGIN_ASSET_EXCLUDES` centralizes the root-staging list next to `PY_LAMBDA_ASSET_EXCLUDES` (both converted); the module doc records the semantics
- All stack call sites converted (`'aggregator/**'` → `'aggregator'`, …)
- Guard tests rewritten at the **behavior level**: the real `IgnoreStrategy` (with the walk's ancestor-`completelyIgnores` semantics) runs against the exported lists — `cdk.out/.cache/x.zip` must be ignored; plugin/shared payloads and the API prompt JSONs must **not** be (matchBase makes `'*.json'` any-depth, so this guards against hash-blind prompt edits); plus a static check rejecting any future `dir/**` entry. Mutation-verified both ways.

## Live acceptance — the sequence that never converged before

1. deploy (one-time hash roll from the new lists): Ingestion/Processing/Api updated
2. **deploy again: all five stacks `(no changes)`**
3. **`cdk diff`: `Number of stacks with differences: 0`**
4. **deploy a third time: all no-op**

Deploy and diff agree for the first time in the project's history. Full gate exit 0 (1096 backend tests, all suites).

### DIFF
--- a/voc-datalake/lib/stacks/api-stack.ts
+++ b/voc-datalake/lib/stacks/api-stack.ts
@@ -127,7 +127,8 @@ export class VocApiStack extends cdk.Stack {
     const createApiLambdaCode = (handlerFileName: string): lambda.Code => {
       return lambda.Code.fromAsset('lambda', {
         // Stages only api/ + shared/ — everything else is hash noise.
-        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator', 'jobs', 'processor', 'research'],
+        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, '/aggregator/', '/jobs/', '/processor/', '/research/'],
+        ignoreMode: cdk.IgnoreMode.GIT,
         bundling: {
           image: lambda.Runtime.PYTHON_3_14.bundlingImage,
           command: [
@@ -465,7 +466,8 @@ export class VocApiStack extends cdk.Stack {
     const createJobLambdaCode = (jobFolder: string): lambda.Code => {
       return lambda.Code.fromAsset('lambda', {
         // Stages only jobs/ + api/ (projects.py, product_context.py, prompts) + shared/.
-        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator', 'processor', 'research'],
+        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, '/aggregator/', '/processor/', '/research/'],
+        ignoreMode: cdk.IgnoreMode.GIT,
         bundling: {
           image: lambda.Runtime.PYTHON_3_14.bundlingImage,
           command: [
--- a/voc-datalake/lib/stacks/api-stack.ts
+++ b/voc-datalake/lib/stacks/api-stack.ts
@@ -127,7 +127,7 @@ export class VocApiStack extends cdk.Stack {
     const createApiLambdaCode = (handlerFileName: string): lambda.Code => {
       return lambda.Code.fromAsset('lambda', {
         // Stages only api/ + shared/ — everything else is hash noise.
-        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator/**', 'jobs/**', 'processor/**', 'research/**'],
+        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator', 'jobs', 'processor', 'research'],
         bundling: {
           image: lambda.Runtime.PYTHON_3_14.bundlingImage,
           command: [
@@ -465,7 +465,7 @@ export class VocApiStack extends cdk.Stack {
     const createJobLambdaCode = (jobFolder: string): lambda.Code => {
       return lambda.Code.fromAsset('lambda', {
         // Stages only jobs/ + api/ (projects.py, product_context.py, prompts) + shared/.
-        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator/**', 'processor/**', 'research/**'],
+        exclude: [...PY_LAMBDA_ASSET_EXCLUDES, 'aggregator', 'processor', 'research'],
         bundling: {
           image: lambda.Runtime.PYTHON_3_14.bundlingImage,
           command: [
--- a/voc-datalake/lib/stacks/ingestion-stack.ts
+++ b/voc-datalake/lib/stacks/ingestion-stack.ts
@@ -26,6 +26,7 @@ import { NagSuppressions } from 'cdk-nag';
 import { apiSecretsSuppressions, bedrockModelSuppressions } from '../utils/nag-suppressions';
 import { allowlistedModelArns } from '../utils/model-allowlist';
 import { pythonLayerCode } from '../utils/python-layer-bundling';
+import { ROOT_PLUGIN_ASSET_EXCLUDES } from '../utils/lambda-asset-excludes';
 
 export interface VocIngestionStackProps extends cdk.StackProps {
   feedbackTable: dynamodb.Table;
@@ -399,20 +400,12 @@ export class VocIngestionStack extends cdk.Stack {
   }
 
   private bundlePluginCode(pluginId: string): lambda.Code {
-    // Root-based staging (this bundle spans plugins/ AND lambda/shared), so
-    // everything not excluded here feeds the asset hash and redeploys all
-    // ingestors on unrelated edits (issue #194 follow-up). Only
-    // plugins/<id>/ingestor, plugins/_shared and lambda/shared are copied.
+    // Root-based staging (this bundle spans plugins/ AND lambda/shared):
+    // everything not excluded feeds the asset hash. The exclude list lives
+    // in lambda-asset-excludes.ts — see its doc for the dot-child pattern
+    // semantics that caused issue #203 (cdk.out/.cache churned every hash).
     return lambda.Code.fromAsset('.', {
-      exclude: [
-        '**/__pycache__', '**/*.pyc', '**/.DS_Store', '**/test/**', '**/conftest.py', '**/test_*.py',
-        'plugins/_template/**', 'node_modules/**', 'cdk.out/**', 'frontend/**', '*.ts', '*.js', '*.json', '*.md',
-        'bin/**', 'lib/**', 'dist/**', '.venv/**', '.pytest_cache/**', '.ruff_cache/**', 'coverage_html/**',
-        '.coverage', 'pytest.ini', 'requirements-dev.txt', 'chrome-extension/**', 'scripts/**', 'schemas/**',
-        'Workshop/**',
-        'lambda/aggregator/**', 'lambda/api/**', 'lambda/custom_resources/**', 'lambda/jobs/**',
-        'lambda/layers/**', 'lambda/processor/**', 'lambda/research/**', 'lambda/stream/**',
-      ],
+      exclude: ROOT_PLUGIN_ASSET_EXCLUDES,
       bundling: {
         image: lambda.Runtime.PYTHON_3_14.bundlingImage,
         command: [
--- a/voc-datalake/lib/stacks/ingestion-stack.ts
+++ b/voc-datalake/lib/stacks/ingestion-stack.ts
@@ -26,7 +26,7 @@ import { NagSuppressions } from 'cdk-nag';
 import { apiSecretsSuppressions, bedrockModelSuppressions } from '../utils/nag-suppressions';
 import { allowlistedModelArns } from '../utils/model-allowlist';
 import { pythonLayerCode } from '../utils/python-layer-bundling';
-import { ROOT_PLUGIN_ASSET_EXCLUDES } from '../utils/lambda-asset-excludes';
+import { rootPluginAssetExcludes } from '../utils/lambda-asset-excludes';
 
 export interface VocIngestionStackProps extends cdk.StackProps {
   feedbackTable: dynamodb.Table;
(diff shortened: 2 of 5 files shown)


perrozzi:
This PR fixes real, well-diagnosed CDK asset-hash churn: `exclude: ['dir/**']` under minimatch's `{ matchBase: true, dot: false }` semantics does not exclude dot-children, so `cdk.out/.cache/*.zip` (CDK's own publishing cache) fed each deploy's output back into the next synth's fingerprint. The fix — excluding directories by name so `completelyIgnores` prunes the whole subtree — is correct, the two exclude lists are properly centralized in `lambda-asset-excludes.ts`, and the rewritten tests exercising the real `IgnoreStrategy` (including ancestor-directory pruning) are a genuine upgrade over the old string-grep assertions. Solid change overall; the concerns below are about the semantic broadening the conversion introduces.

### Should fix

- **`.git` is still staged by the root-based `fromAsset('.')`** (`voc-datalake/lib/utils/lambda-asset-excludes.ts`). CDK does not exclude `.git` by default, and neither the old nor the new `ROOT_PLUGIN_ASSET_EXCLUDES` names it. `.git/index` and refs change on every commit/checkout, so the five ingestor hashes will still churn *across commits* with zero payload changes — the same symptom, different trigger. The PR's live acceptance (repeated deploys in one session) wouldn't surface this. Add `'.git'` (and consider `'.env*'`, `'.vscode'`, `'.idea'`) plus a behavior test like `expect(ignores('.git/index')).toBe(true)`.
- **Bare-name patterns now match at any depth, which silently broadens exclusion of payload subtrees.** Previously `'lib/**'`, `'bin/**'`, `'dist/**'`, `'scripts/**'`, `'schemas/**'` (slash-containing, so matchBase didn't apply) only matched at the staging root. The bare forms `'lib'`, `'bin'`, `'dist'`, `'scripts'`, `'schemas'` now exclude any directory with that basename inside `plugins/<id>/ingestor`, `plugins/_shared`, or `lambda/shared` — a plugin adding a `lib/` or `scripts/` helper folder would be silently dropped from its bundle with no error. The module doc acknowledges this "by design" for `dist`, but the behavior tests only cover `test`/`__pycache__` collisions. Add negative guards for the riskiest basenames, e.g. `expect(ignores('plugins/webscraper/ingestor/lib/parser.py')).toBe(false)` — which will fail today and force either path-qualifying those entries or an explicit, tested decision to accept the collision.
- **Editing any one plugin still redeploys all five ingestors**: `bundlePluginCode` in `ingestion-stack.ts` stages the whole root (all `plugins/*` trees) into every plugin's source fingerprint. Pre-existing, not made worse here, but since this PR centralizes the exclude list, consider a follow-up that parameterizes sibling-plugin excludes per `pluginId` (the same way the `'lambda'`-rooted call sites add sibling handler dirs).

### Nits

- `lambda-asset-excludes.test.ts` switched from semicolons to no-semicolon style, diverging from the rest of the `lib/` TypeScript. Revert or let the formatter decide, but don't mix.
- `'.coveragerc'` and `'ruff.toml'` are new exclusions not present in the old list — worth one line in the PR description since they contribute to the acknowledged one-time hash roll.
- The static guard `not.toMatch(/\/\*\*$/)` won't catch a future `'dir/**/*.py'`-style entry that has the same dot-child leak mid-pattern; the behavior tests are the real defense, so consider noting that in the test comment.
- The test asserting `plugins/webscraper/manifest.json` is ignored (via bare `'*.json'`) is correct today, but the comment "synth-time inputs — staged is fine" reads backwards (it's *not* staged); tighten the wording so a future reader doesn't "fix" it.

No blockers. None of the repo's domain rules (ingestor prefix registration, Bedrock model/ARN lockstep, i18n) are implicated by this change.

---

mundurragacl:
All three should-fixes are in `b1d074c` — and your payload-collision point forced a better design than my round-1 fix:

**The real resolution: IgnoreMode.GIT.** Your second bullet exposed that GLOB mode cannot deliver anchoring and dot-safety simultaneously (matchBase makes bare names match at any depth; `'dir/**'` leaks dot-children — pick your poison). Gitignore semantics give both: `'dir/'` prunes subtrees dot-children included, leading `'/'` anchors to the staging root. Every call site now pins `ignoreMode: IgnoreMode.GIT` (enforced by a site test), and the lists use anchored forms.

**`.git` staged** — Confirmed and fixed: `'/.git/'` (+ `/.vscode/`, `/.idea/`, `/.env*`), with behavior tests on `.git/index` and `.git/refs/heads/development`. You were right that my in-session acceptance loop could never have caught the across-commits churn. Mutation-verified: removing `'/.git/'` fails the metadata test.

**Payload-name collisions** — Fixed by anchoring, with your exact negative guards: `plugins/webscraper/ingestor/lib/parser.py` and `.../scripts/seed.py` assert as staged, while `/lib/stacks/...` and `/scripts/build-layers.sh` stay excluded.

**Sibling-plugin churn (your follow-up suggestion — done now, not deferred)** — `rootPluginAssetExcludes(pluginId, allPluginIds)` excludes every other plugin dir per bundle. **Proven live**: a probe file added to `plugins/webscraper` re-bundled and diffed only `IngestorWebscraper`; the other four ingestors untouched.

**Nits**: semicolon style restored throughout the test file; `.coveragerc`/`ruff.toml` now called out (they're in the anchored list; they contributed to the acknowledged one-time roll); the static `dir/**` check is gone with the GLOB lists themselves — the module doc + site test note that the behavior tests are the load-bearing defense; the manifest.json comment now reads correctly ("NOT staged — synth-time input read from the source tree; manifest edits alone must not redeploy ingestor code").

Re-verified end-to-end after the change: deploy (one-time roll) → deploy all-no-op → `cdk diff`: 0 differences. Gate exit 0.

perrozzi:
The push is a substantive rework, not just a patch-up: the exclude lists moved from GLOB/matchBase semantics to `IgnoreMode.GIT` with anchored gitignore patterns, `ROOT_PLUGIN_ASSET_EXCLUDES` became the parameterized `rootPluginAssetExcludes(pluginId, allPluginIds)` with per-plugin sibling excludes, and the tests now run `IgnoreStrategy.git` with negative guards for same-name payload subtrees. All three of my should-fix items are addressed, and this is a cleaner design than patching the GLOB list would have been.

### Resolved
- `.git` (plus `.vscode`, `.idea`, `.env*`) now excluded, with behavior tests for `.git/index` and `.git/refs/...`.
- Bare-name any-depth broadening eliminated: `/lib/`, `/bin/`, `/scripts/`, `/dist/` are root-anchored under git semantics, with the exact negative tests I asked for (`plugins/webscraper/ingestor/lib/parser.py` stages).
- Sibling-plugin churn fixed: per-`pluginId` excludes mean editing one plugin no longer redeploys all five ingestors, with tests both ways.
- Semicolon style restored in the test file; the backwards manifest.json comment was rewritten correctly; the weak static `dir/**` regex guard was replaced by the behavior tests it was standing in for.

### New findings
- `voc-datalake/lib/utils/lambda-asset-excludes.ts`: `/.env*` is root-anchored, so a nested dev secret file like `plugins/webscraper/.env` or `lambda/shared/.env` would both be *staged into the hash* and, worse, **shipped in the bundle** (the docker `cp -r` copies the whole ingestor/shared trees). The old bare `frontend/.env.local`-class risk is covered, but consider an unanchored `.env*` entry too — nothing legitimate in the staged trees uses that name. Nit-to-should-fix depending on how you weigh the shipping angle.
- `voc-datalake/lib/stacks/ingestion-stack.ts`: sibling excludes are derived from `allPlugins` (the loader's output). A plugin directory on disk that the loader doesn't surface (malformed/missing manifest, or a freshly scaffolded dir) is excluded for no one and churns every ingestor's hash — the exact symptom class this PR fixes. Deriving the id list from `fs.readdirSync('plugins')` (minus `_shared`/`_template`) would close that hole. Nit.
- `voc-datalake/lib/utils/lambda-asset-excludes.test.ts`: `PLUGIN_IDS` is hardcoded; adding a sixth plugin won't fail these tests even though its sibling-exclusion behavior is untested. Deriving from disk (same as above) keeps the guard honest. Nit.
- Note for the changelog: per-plugin exclude lists mean the five ingestors now publish five distinct assets instead of one deduplicated root asset — expected and worth it, but it's part of the acknowledged one-time hash roll.

Nothing blocking. The GIT-mode conversion is consistent across all five call sites, the `IgnoreMode.GIT` pin is statically enforced per site, and the behavior tests are mutation-honest. Nice close-out.

---

mundurragacl:
All three round-2 findings taken in the latest push:

**`.env*` shipping** — Your sharpest catch: staging excludes gate the bundling container's `/asset-input`, so this wasn't just hash hygiene — a nested `plugins/x/.env` would have been `cp -r`'d into the deployed bundle. Now unanchored `'.env*'` in BOTH lists, with behavior tests at all three depths (`.env.local`, `plugins/webscraper/.env`, `lambda/shared/.env.production`) asserting they neither hash nor ship.

**Loader-skipped plugin dirs** — Sibling ids now derive from disk (`readdirSync` minus `_`-prefixed) in `ingestion-stack`, so a malformed-manifest or freshly-scaffolded dir is excluded from every other ingestor's hash even when the loader doesn't surface it.

**Hardcoded test ids** — The tests derive `PLUGIN_IDS` from disk exactly the same way, with an anti-vacuous assertion pinning the known five; a sixth plugin is covered automatically.

**Changelog note taken**: the per-plugin lists do trade the single deduplicated root asset for five distinct ones — that was part of the earlier acknowledged roll. Worth recording: this round's changes rolled NOTHING (no `.env` files exist in the staged trees) — deploy all-no-op, `cdk diff` 0 differences, gate exit 0.

perrozzi:
The push addresses all three nits from my previous review: `.env*` is now unanchored (any-depth) in both exclude lists with a test proving it also gates the bundling container's `/asset-input`, the ingestion stack derives sibling-exclude plugin ids from disk instead of the loader's output, and the test's `PLUGIN_IDS` is derived from disk with an anti-vacuous assertion so a sixth plugin is covered automatically.

### Resolved
- Unanchored `.env*` in both `PY_LAMBDA_ASSET_EXCLUDES` and `rootPluginAssetExcludes`, replacing root-only `/.env*`; nested dev secrets (`plugins/webscraper/.env`, `lambda/shared/.env.production`) are neither hashed nor shipped, with tests.
- `voc-datalake/lib/stacks/ingestion-stack.ts`: sibling excludes now derive from `fs.readdirSync(pluginsDir)` (minus `_`-prefixed dirs), closing the loader-skipped-plugin churn hole.
- `voc-datalake/lib/utils/lambda-asset-excludes.test.ts`: `PLUGIN_IDS` derived from disk the same way, with a containment check guarding against a vacuous empty list.

### New findings
- `voc-datalake/lib/utils/lambda-asset-excludes.test.ts`: the swap of `stream/.env` → `shared/.env` is actually an improvement (the old assertion was satisfied by the `/stream/` prune, not by `.env*`), so no concern — just noting it's intentional and correct.
- Extremely minor: `withFileTypes` direntries don't follow symlinks, so a symlinked plugin directory would fail `isDirectory()` and escape the sibling excludes. No such layout exists in the repo; not worth code, mentioning only for the record.

Nothing blocking, nothing should-fix. The delta is a clean close-out of the previous review's nits, the test change is mutation-honest, and the disk-derivation logic in stack and test are kept in lockstep by construction. Ship it.

---

PR #206, opened 2026-07-16
feat(web-search): deploy the gateway by default for all deployments

### LINKED ISSUE #205: Web search should be enabled by default for all deployments (chat toggle + research wizard source)
## Request (maintainer)

The /chat web-search toggle and the Run Research wizard's 'Include public web search' source are both fully implemented but gated on `features.webSearch`, which is false unless `VocWebSearchStack` is deployed — and that stack is **opt-in** (`-c enableWebSearch=true`). Result: production shows neither surface (screenshot: Run Research step 1 lists only Customer Feedback + Personas).

## Change

Deploy `VocWebSearchStack` **by default**; keep an explicit escape hatch (`enableWebSearch: false`) for accounts that can't (e.g. no us-east-1 bootstrap for cross-region deployments). Per-request behavior stays opt-in (chat toggle off per message, wizard checkbox unchecked) — only stack deployment flips. The original 'keep it a conscious choice until a real gateway round-trip has been validated post-release' condition in bin/voc-datalake.ts has been met.

Touchpoints: bin/voc-datalake.ts default, web-search-stack.ts doc, docs/deployment.md (4 mentions), deploy + live verification of both UI surfaces.

mundurragacl:
Fixes #205

The /chat web-search toggle and the Run Research wizard's 'Include public web search' data source were both fully built but invisible in every deployment that didn't opt into `VocWebSearchStack` — which was all of them, including production. The bin comment's condition for flipping the default ('a conscious choice until a real gateway round-trip has been validated post-release') has been met, so:

## Changes
- **Default on**: `VocWebSearchStack` deploys unless `enableWebSearch: false` (escape hatch for accounts that can't bootstrap us-east-1, which cross-region references require). Decision extracted to `shouldDeployWebSearch()` with a mutation-verified regression test — a silent flip back to opt-in fails the no-context case.
- **Per-request behavior unchanged**: the chat toggle stays off per message and the wizard checkbox unchecked; only stack deployment defaults on. Gateway has no standing cost; searches bill $7/1k queries.
- Docs reframed to opt-out (deployment.md stack table + enable section + feature section, web-search-stack.ts doc).

## Live verification (us-west-2 app, cross-region gateway)
- Deployed with **no context flag** — the stack came up purely from the new default; `WebSearchAvailable=true`, live `config.json` ships `features.webSearch: true`
- **Chat**: `POST /chat/stream` with `use_web_search: true` executed 4 web-search tool rounds and returned cited `web_sources` in the done event
- **Research**: `voc-research-step` now carries `WEB_SEARCH_GATEWAY_URL`/`TOOL_NAME`; the wizard's checkbox flows `use_web_search` through `/projects/{id}/research` into the Step Functions job (searches during analysis, disclosure line in the report)

Gate exit 0 (1096 backend tests, all suites).

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -178,7 +178,7 @@ The platform consists of 4 core stacks plus 2 optional ones:
 | `VocProcessingStack` | Processor, Aggregator, Step Functions, Bedrock | Core, Ingestion |
 | `VocApiStack` | API Gateway, API Lambdas, Webhooks, WAF | Core, Ingestion, Processing |
 | `BedrockAccessStack` (optional) | Bedrock model access / Anthropic use case submission — created only when `anthropicUseCase` is set in `cdk.context.json` (see the conditional in `bin/voc-datalake.ts`) | None |
-| `VocWebSearchStack` | AgentCore Gateway for public web search — deployed by default, opt out via `enableWebSearch: false`; always deploys to us-east-1 (the connector only exists there) | None |
+| `VocWebSearchStack` (default-on, opt-out) | AgentCore Gateway for public web search — deployed by default, opt out via `enableWebSearch: false`; always deploys to us-east-1 (the connector only exists there). **Upgrade note:** existing non-us-east-1 deployments must bootstrap us-east-1 once (`cdk bootstrap aws://ACCOUNT_ID/us-east-1`) or set the opt-out flag | None |
 
 ### Deploy All Stacks
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -178,7 +178,7 @@ The platform consists of 4 core stacks plus 2 optional ones:
 | `VocProcessingStack` | Processor, Aggregator, Step Functions, Bedrock | Core, Ingestion |
 | `VocApiStack` | API Gateway, API Lambdas, Webhooks, WAF | Core, Ingestion, Processing |
 | `BedrockAccessStack` (optional) | Bedrock model access / Anthropic use case submission — created only when `anthropicUseCase` is set in `cdk.context.json` (see the conditional in `bin/voc-datalake.ts`) | None |
-| `VocWebSearchStack` (optional) | AgentCore Gateway for public web search — opt-in via `enableWebSearch: true`; always deploys to us-east-1 (the connector only exists there) | None |
+| `VocWebSearchStack` | AgentCore Gateway for public web search — deployed by default, opt out via `enableWebSearch: false`; always deploys to us-east-1 (the connector only exists there) | None |
 
 ### Deploy All Stacks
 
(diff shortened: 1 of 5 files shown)


perrozzi:
This PR flips `VocWebSearchStack` from opt-in to deploy-by-default, extracting the decision into a tested `shouldDeployWebSearch()` helper and updating the docs to an opt-out framing. The change is small, well-documented, and comes with a genuine regression test (the no-context case fails if the default silently reverts). Overall it's clean; my concerns are about the upgrade/failure path, not the code itself.

### Should fix

- **Default-on now hard-fails existing non-us-east-1 deployments that never bootstrapped us-east-1.** Anyone running `cdk deploy --all` in CI after pulling this change will hit a cross-region bootstrap error unless they add `enableWebSearch: false`. `docs/deployment.md` covers the escape hatch, but this deserves an explicit upgrade/breaking-change note (CHANGELOG/release notes), and ideally a synth-time hint: in `voc-datalake/bin/voc-datalake.ts`, when `webSearchCrossRegion` is true, emit an `Annotations` info/warning pointing at the us-east-1 bootstrap requirement and the opt-out flag so the eventual `cdk bootstrap` failure isn't the first signal.
- **Unrecognized context values silently deploy the stack.** `shouldDeployWebSearch()` in `voc-datalake/lib/utils/web-search-default.ts` treats anything except `false`/`'false'` as enable — so `-c enableWebSearch=FALSE`, `=0`, `=no`, or a typo like `=flase` all deploy. Under the old opt-in default, a typo failed safe (off); now it fails open (deploys a stack the operator tried to disable). Consider normalizing case at minimum, and warning (via `cdk.Annotations`) on values that are neither a recognized true nor false form.

### Nits

- `voc-datalake/lib/utils/web-search-default.test.ts` could add a case for an unrecognized string (e.g. `'FALSE'` or `'0'`) to pin down whatever behavior you choose for the point above.
- The doc comment in `web-search-stack.ts` and the bin comment both restate the same opt-out/bootstrap guidance; fine, but if the parsing rule changes, remember there are now three places (plus `deployment.md`) describing the flag's semantics — the helper's JSDoc is the natural single source of truth.
- `docs/deployment.md` still lists `VocWebSearchStack` in a table column implying it's parallel to `BedrockAccessStack (optional)`; consider a "(default-on, opt-out)" marker so the table alone conveys the difference.

Note: only the diff shown was reviewed; the referenced live-verification claims (chat stream, research step env vars) were not independently verifiable from this diff.

---

mundurragacl:
Both should-fixes and the nits are in the new push:

**Fail-open parsing** — Exactly right that opt-in's fail-safe became fail-open. `shouldDeployWebSearch()` now accepts case-insensitive, trimmed `true`/`false` and **throws on anything else** — verified at the CLI: `cdk synth -c enableWebSearch=flase` fails with `Unrecognized enableWebSearch context value: "flase". Use true/false (web search deploys by default; opt out with -c enableWebSearch=false)`. Tests pin `FALSE`, `' false '`, and the whole typo family (`flase`/`no`/`0`/`off`/`1`/`{}`) as throwing. Chose throw over warn-and-guess: under default-on, both wrong guesses are bad (deploying against intent, or skipping a forced enable).

**Upgrade path** — A synth-time `Annotations.addInfo` now fires whenever the app region isn't us-east-1, naming the bootstrap command and the opt-out flag (verified in synth output for this us-west-2 deployment). The docs stack table gained the `(default-on, opt-out)` marker plus an explicit upgrade note for existing non-us-east-1 deployments. This repo ships no CHANGELOG file — the PR title/body and the issue serve as the release note, and the annotation is the in-band signal for anyone who skips docs.

**Nits** — Unrecognized-string test cases added (see above); flag-semantics prose deduplicated — the helper's JSDoc is now declared the single source of truth and the bin + web-search-stack comments defer to it.

Gate exit 0; the live deployment from the earlier round is unaffected (this round changes no deployed resources — parsing and annotations only).

perrozzi:
The new push tightens the flag semantics (strict, case-insensitive parsing that throws on unrecognized values), adds a synth-time `Annotations` hint for the cross-region bootstrap requirement, an upgrade note in the deployment table, and consolidates the flag documentation to a single source of truth in the helper's JSDoc. All previous findings and nits were addressed.

### Resolved
- Synth-time hint for the us-east-1 bootstrap requirement added via `cdk.Annotations.addInfo` in `voc-datalake/bin/voc-datalake.ts`, plus an explicit upgrade note in `docs/deployment.md`.
- Unrecognized context values no longer fail open — `shouldDeployWebSearch()` now throws at synth, with regression tests covering `'flase'`, `'no'`, `'0'`, `'off'`, numbers, and objects.
- Nit: test cases for case variants (`'FALSE'`, `' false '`, `'TRUE'`) added.
- Nit: bin comment and stack JSDoc now defer to `web-search-default.ts` as the single source of truth.
- Nit: deployment table now carries a `(default-on, opt-out)` marker distinguishing it from `BedrockAccessStack (optional)`.

### New findings
None blocking. Two observations only:

- Throwing on unrecognized values is stricter than the warn-and-proceed option I suggested — that's a fine (arguably better) choice, but note it's a hard behavior change for anyone who previously had a junk value in `cdk.context.json` (old code silently treated it as "off"; now synth fails). The error message is clear and self-remedying, so no action needed.
- `voc-datalake/lib/utils/web-search-default.test.ts` pins `null` → deploy; that matches `tryGetContext` semantics and the implementation, so it's consistent.

The incremental changes look clean; no further requests.

---

PR #209, opened 2026-07-16
Consolidate Feedback tab into Categories

### LINKED ISSUE #198: Consolidate the Feedback tab into Categories (remove redundant list page)
## Problem

The sidebar has two tabs in the **Signals** section that show largely the same thing:

- **Feedback** (`/feedback`) — a filterable/searchable list of feedback items.
- **Categories** (`/categories`) — category analytics that, once a category (or keyword) is selected, renders the **same** feedback list underneath.

Opening Categories and clicking a category gives the same drill-down as Feedback filtered by that category. Both pages render the identical `FeedbackCard` grid and both call `GET /feedback` (`limit: 100`). Two tabs for one job is redundant and confusing.

**Proposal:** remove the standalone Feedback tab and port its few unique capabilities into Categories.

## Scope — "Feedback" is two routes with opposite dispositions

| Route | Component | Decision |
|---|---|---|
| `/feedback` (list) | `pages/Feedback/Feedback.tsx` | **Remove** — redundant with Categories |
| `/feedback/:id` (detail) | `pages/FeedbackDetail/FeedbackDetail.tsx` | **Keep** — shared destination of `FeedbackCard`, `FeedbackCarousel` (AI chat), `SimilarFeedbackList`, ProblemAnalysis, Dashboard urgent list |

Removing the detail route would break ~5 components; only the list tab is redundant.

## Feedback-only features to port into Categories

1. **Unfiltered "All" view** — Categories currently renders no list until a category/keyword is selected (`shouldFetchFeedback` gate). Add an explicit **All** chip (preferred over relaxing the gate: Categories already fires 4–5 queries on load; an unconditional feedback fetch adds a 100-item query to every visit).
2. **Free-text search** — `GET /feedback/search` (Categories only has word-cloud keyword chips).
3. **"Urgent only" toggle** — `GET /feedback/urgent`.
4. **URL-synced filters** (`q`/`source`/`sentiment`/`category`) — needed for shareable links and as the landing target for FeedbackDetail tag-click deep-links. Note: Categories uses multi-select chips, so the URL scheme needs a multi-value encoding (e.g. comma-separated) while single-value deep-links map to selecting one chip.
5. **Results header with partial-window count** ("N of TOTAL / N+" via `is_partial_window`).
6. **Feedback-list PDF export** (`generateFeedbackPDF`) — Categories has an analytics PDF + CSV; decide whether to keep the list PDF too (recommend: yes, alongside CSV).

## Inbound references that must be handled

- `FeedbackDetail.tsx` tag clicks `navigate('/feedback?category|q|source=…')` and two `<Link to="/feedback">` back-links → repoint to `/categories?…`.
- **Sidebar urgent badge**: `SidebarComponents.tsx` renders the red urgent-count badge only when `item.to === '/feedback'` (fed by `getUrgentFeedback` in `Layout.tsx`). Removing the nav item silently kills the app-wide urgent indicator → move the badge to the `/categories` nav item.
- **Breadcrumbs**: `Breadcrumbs.tsx` auto-builds crumbs from path segments, so `/feedback/:id` renders an intermediate "Feedback" crumb linking to `/feedback`. A `/feedback` → `/categories` redirect (with query passthrough) is therefore **required**, not optional.
- Dead code: `pages/Categories/FeedbackList.tsx` is not imported by `Categories.tsx` (only by its own test) — delete.

## Implementation sketch (frontend-only, no backend/IaC change)

- `App.tsx`: remove the `feedback` list route (keep `feedback/:id`), add redirect `/feedback` → `/categories` with query passthrough.
- `Layout.tsx`: remove the `/feedback` entry from `NAV_ITEMS`; `SidebarComponents.tsx`: re-key the urgent badge to `/categories`.
- `Categories.tsx` (+ hooks/sub-components): absorb All chip, search, urgent toggle, URL sync, partial-window header, list PDF. ⚠️ `Categories.tsx` sits at the ESLint `complexity: 12` ceiling — extract hooks/helpers rather than inlining branches.
- `FeedbackDetail.tsx`: repoint tag-click navigations and back-links.
- Delete `pages/Feedback/*` and `pages/Categories/FeedbackList.tsx`.
- Cleanup: `menuStatus.feedback` (`cdk.context.json`), generated `menu-config.json`, `nav.feedback` key across the 8 locale files, `Breadcrumbs` label map.
- Tests: update Categories suite, `Layout.test.tsx`, `Breadcrumbs.test.tsx`, `FeedbackDetail.test.tsx` (tag-click assertions → `/categories?…`); remove Feedback page tests.

All API endpoints stay (`/feedback`, `/feedback/{id}`, `/feedback/urgent`, `/feedback/search`, `/feedback/entities`, `/metrics/*`) — still used by Categories and FeedbackDetail.

## Acceptance criteria

- [ ] Feedback no longer appears in the sidebar; Categories remains in Signals.
- [ ] `/feedback/:id` still works from every entry point (cards, carousel/AI chat, similar feedback, problem analysis, dashboard).
- [ ] Categories has an **All** view, free-text search, and an Urgent-only toggle.
- [ ] Categories syncs `q`/`source`/`sentiment`/`category` to the URL; FeedbackDetail tag clicks land pre-filtered on `/categories?…`.
- [ ] The sidebar urgent-count badge now shows on the Categories nav item.
- [ ] `/feedback` redirects to `/categories` (query passthrough) — breadcrumb links from detail pages don't 404.
- [ ] Results list shows the "N of TOTAL / N+" partial-window count.
- [ ] `pages/Feedback/*` and dead `FeedbackList.tsx` deleted; locale/menu-config/breadcrumb orphans cleaned.
- [ ] `npm run lint`, `npm run typecheck`, `npm run test -- --run` all clean; frontend-only diff.

perrozzi:
## Summary

Closes #198.

The sidebar had two pages showing the same data: **Feedback** (a filterable list) and **Categories** (analytics + the same list when a category was selected). This PR consolidates them: the Feedback list page is removed and Categories becomes the single place to browse, filter, search, and export feedback — with the analytics cards acting as interactive filters.

`/feedback/:id` (the detail page) is **kept** — it is the shared click target of FeedbackCard, the AI-chat carousel, similar-feedback lists, and Problem Analysis. `/feedback` list URLs redirect to `/categories` with filters preserved, so old deep links keep working.

## What changed

**Consolidation (core)**
- `/feedback` list route removed; redirect `/feedback?...` → `/categories?...` (query params mapped)
- Ported from the old page: browse-all view, free-text search (`/feedback/search`), urgent-only toggle (`/feedback/urgent`), URL-synced filters (`q`/`source`/`sentiment`/`category`), partial-window result counts, list export
- Sidebar urgent badge moved to the Categories nav item; breadcrumbs on detail pages ride the redirect

**UX rationalization (every filter has exactly one control)**
- Category Distribution rows double as the category selector (multi-select, `aria-pressed`); duplicate chips card and insights row removed
- Sentiment gauge legend is the single sentiment control
- Trending Keyword clicks run a server-side search across the full corpus (fixes a bug where keyword chips silently filtered only the loaded 100-item window)
- One unified FilterBar: search, source, star rating, urgent toggle, clear-all
- Nothing selected = browse all (list visible by default)

**Layout**
- Distribution + Sentiment + Keywords cards share one 3-column row; compacted vertically (row height 420px → 315px)
- Category Distribution collapses to the top 5 rows with a "Show all" toggle; below-the-fold selections stay visible so deep links are never hidden

**New capabilities**
- **Pagination**: the list uses the backend's existing offset/limit support via an infinite query with a "Load more" button (search/urgent endpoints don't paginate server-side)
- **Rating direction**: the star filter gained an `& up` / `& below` toggle — "3 stars and below" is often the point of a VoC tool; unrated items are excluded once a threshold is set
- **Single PDF export**: one report with the analytics header plus the filtered feedback items table (previously two partial PDFs); redundant second PDF button removed
- **A11y**: CSV export button has a visible label, `title`, and `aria-label`; rating direction is a proper radiogroup
- New i18n keys (`loadMore`, `exportCsvShort`, `exportCsvTooltip`) in all 8 locales

**Removed**
- `pages/Feedback/` (list page), CategorySelector, InsightsRow, SourceFilter, FeedbackSearchBar, `feedbackPdfGenerator`, dead `FeedbackList.tsx` and `useCategoriesData.ts` (+ their tests)

## Testing

- Gates on the rebased head: `tsc -b` clean, `eslint` 0 errors, i18n check passing, **1754 vitest tests / 112 files**, production build OK
- Frontend-only change — no Lambda or infrastructure modifications
- Live E2E on a deployed environment (Playwright): redirect + deep links, browse-all default, server search, urgent toggle, category row selection, sidebar badge, pagination (`offset=100` request, 100 → 200 items), rating `& below` filtering, merged PDF content (analytics header + feedback table, no second header), CSV button accessible name

## Notes for reviewers

- The PDF/CSV exports cover the currently loaded pagination window, matching the on-screen list
- Legacy `?all=1` URLs are silently accepted (browse-all is now the default state)


perrozzi:
This frontend-only PR consolidates the standalone Feedback list page into Categories: `/feedback` list URLs redirect to `/categories`, analytics cards become interactive filters, and it adds pagination, a rating-direction toggle, and a merged PDF export. The visible portions are well-structured (state extracted into focused hooks, thorough vitest coverage, dead components removed with their tests). **Note: the diff was truncated — `useCategoryFilters.ts` is cut off mid-file and `useFeedbackListData.ts` (pagination/search/urgent orchestration) is not visible at all, so this review is partial and cannot verify the infinite-query, debouncing, or URL write-back logic.**

### Blockers
- **New user-visible strings are hardcoded English, violating the repo i18n rule.** `FilterBar.tsx` is an entirely new component with untranslated strings: `"Search feedback..."`, `"All Sources"`, `"Urgent only"`, `"Clear filters"`, `"& up"`, `"& below"`, `"Any"`, `"Any rating"`, plus `title`/`aria-label` values (`"Filter by source"`, `"Star rating"`, `"Rating direction"`, star tooltips). `CategoryDistribution.tsx` adds `"Click a category to filter the results below"`, `"Show all {n} categories"`, `"Show top 5"`. Only three keys (`loadMore`, `exportCsvShort`, `exportCsvTooltip`) were added to the 8 locales. These need i18n keys across all locales like the rest of the change.
- **Possibly missing locale keys `showingOf` and `partialWindowHint`.** `ResultsCountLine` in `FeedbackResults.tsx` calls `t('showingOf', …)` and `t('partialWindowHint')` from the `common` namespace, but the locale diffs don't add them. If these previously lived in a Feedback-page-specific namespace (now deleted), all locales will render raw keys. Please confirm they exist in `common.json` for all 8 locales.

### Should fix
- **URL sync vs. browser history**: `useCategoryFilters` initializes React state from `searchParams` once and (per the visible write-back tests) mirrors changes to the URL via an effect. Back/forward navigation after changing filters will likely update the URL without updating the state (state is only read at mount). The truncated file may handle this — if not, add a test for back-button behavior.
- **CSV export in `Categories.tsx` (`exportFeedbackCsv`)**: only `original_text` is quoted; `category`, `source_platform`, `sentiment_label`, and dates are emitted raw, so a comma in any of these corrupts the CSV. Also consider guarding against CSV/formula injection (fields starting with `=`, `+`, `-`, `@`) since feedback text is untrusted user input opened in Excel. This is moved-not-new code, but it's the moment to fix it.
- `URL.createObjectURL` in the CSV export is never revoked (`URL.revokeObjectURL`) — leaks a blob per export.

### Nits
- `RatingDirectionToggle` uses `role="radiogroup"`/`role="radio"` on buttons without arrow-key navigation or roving tabindex; native radios or a keydown handler would complete the pattern.
- `safeGeneratePdf` now logs `console.error('PDF export failed:', err)` where the old code was silent — fine, but ensure the error object can't contain feedback text (PII-in-logs rule; browser console is lower risk).
- The sidebar urgent badge is now hard-wired to `item.to === '/categories'` in `SidebarComponents.tsx`; if the categories menu item is disabled via `menuStatus`, the urgent count silently disappears from the nav.
- `cdk.context.json` removes the `feedback` menu key — deployed environments with a cached context containing `"feedback": true` should degrade gracefully since the nav item is gone; worth a quick sanity check that unknown `menuStatus` keys are ignored.
- The FilterBar test asserts on the Tailwind class `.lg\:ml-auto` to detect the trailing container — brittle; prefer a `data-testid` or role query.

---

perrozzi:
Addressed in be34821 + 837a927. Point-by-point:

### Blockers

- **Hardcoded English strings** — ✅ fixed. FilterBar, CategoryDistribution, WordCloudCard, SentimentGaugeCard, FeedbackResults, and the Categories config message now go through the `categories` namespace. Existing keys were reused where the namespace already had them (`allSources`, `clearFilters`, `any`, `feedbackResults`, `noFeedbackFound`, `overallSentiment`, `netSentiment`, `trendingKeywords`, sentiment names…), and 29 new keys were added and translated in all 8 locales (search placeholder, urgent toggle, star-rating titles/directions, distribution hint, show all/top toggle, grid/list view labels, source line). `mentionsTooltip` was also reworded in all locales — keyword clicks now run a search, not a filter.
- **`showingOf` / `partialWindowHint`** — false positive (truncated diff). Both keys exist in `common.json` for all 8 locales; they were added earlier in this PR together with the results-count port. Verified programmatically across every locale.

### Should fix

- **URL sync vs. browser history** — ✅ fixed. `useCategoryFilters` now adopts external URL changes (back/forward, same-route navigations like `navigate('/categories?category=x')`) via guarded render-phase updates; its own `replace`-mirroring never re-triggers adoption. Covered by 4 new tests including a back-navigation restore.
- **CSV export quoting + formula injection** — ✅ fixed. New shared `utils/csv.ts` `csvField()` quotes every field (not just the text), doubles embedded quotes, and neutralizes leading `=`, `+`, `-`, `@`, tab, CR from untrusted feedback text. Unit-tested.
- **`URL.createObjectURL` leak** — ✅ fixed, `revokeObjectURL` in a `finally`, with a test.

### Nits

- **Radiogroup keyboard pattern** — ✅ implemented: arrow keys move the selection, roving tabindex keeps only the checked option tabbable. Tested.
- **`safeGeneratePdf` logging** — ✅ now logs only the error message, not the full object.
- **Brittle `.lg\:ml-auto` test selector** — ✅ replaced with a `data-testid`.
- **Sidebar urgent badge tied to `/categories`** — leaving as is: the badge belongs on the item that opens the list; if that menu item is disabled via `menuStatus` there is no destination for the count anyway.
- **Stale `menuStatus` keys (`"feedback": true`)** — verified safe: `isMenuItemEnabled` looks up by each nav item's `menuKey`, so unknown/stale keys in a cached context are simply never read.

### Drive-by

The `i18n:check` gate was failing for reasons that predate this branch; fixed so the gate is meaningful again: the checker now parses `useTranslation('ns', { keyPrefix })` (CategoriesManager keys were being resolved against the wrong namespace), ja/zh/ko got the missing `categoriesManager.*_one` plural variants, and es/pt `subCount` is actually translated.

Gates on the new head: `tsc -b` clean, `eslint --max-warnings 0` clean, `i18n:check` exit 0, **1764 vitest tests / 112 files** (was 1754), production build OK.

perrozzi:
This push is a comprehensive response to the previous review: all new user-visible strings were extracted into the `categories` namespace across all 8 locales, `useCategoryFilters` gained render-phase URL adoption with back/forward tests, CSV export got a shared `csvField` encoder (quoting + formula-injection guard) and blob-URL revocation with a regression test, the radiogroup got roving tabindex + arrow-key handling, and the brittle Tailwind-class test assertion was replaced by a `data-testid`. **Note: the full-context diff was truncated, so this remains a partial review** (e.g. `useFeedbackListData.ts` is still not fully visible).

### Resolved
- Hardcoded English strings in FilterBar/CategoryDistribution/SentimentGauge/WordCloudCard/FeedbackResults/Categories — all now use i18n keys, added to all 8 locales.
- `showingOf`/`partialWindowHint` concern — not added in this push, so presumably pre-existing in `common.json`; the strengthened `i18n-check.mjs` (now handling `keyPrefix` and multi-arg `useTranslation`) passing gives reasonable confidence.
- URL sync vs. browser history — render-phase snapshot adoption in `useCategoryFilters.ts` with dedicated back/forward and same-route navigation tests.
- CSV quoting + formula injection — new `utils/csv.ts` with thorough tests; all fields now encoded via `csvField`.
- `URL.createObjectURL` leak — revoked in a `finally`, with a regression test.
- Radiogroup keyboard pattern — arrow keys + roving tabindex implemented and tested.
- PDF error logging — now logs only `err.message`, addressing the PII concern.
- Brittle `.lg\:ml-auto` test selector — replaced with `data-testid="filter-bar-trailing"`.

### Still open
- Sidebar urgent badge hard-wired to `item.to === '/categories'` — count silently disappears if that menu item is disabled (nit).
- No sanity check noted for cached `cdk.context.json` containing the removed `"feedback"` menu key (nit).

### New findings
- **Nit** — `voc-datalake/frontend/scripts/i18n-check.mjs`: the namespace regex `useTranslation\(\s*['"](\w+)['"]\s*[,)]` doesn't match the array form `useTranslation(['common', 'categories'])` now used in `Categories.tsx`/`FeedbackResults.tsx`; those files fall back to the default namespace, which happens to be correct here (unprefixed keys resolve to `common`, categories keys are explicitly prefixed), but the check's correctness is now order-dependent on which `useTranslation` call appears first in a file. Worth a comment or a follow-up hardening so a future refactor doesn't silently skip validation.
- **Nit** — `FilterBar.tsx` `RatingDirectionToggle`: the keydown handler toggles based on the current *state* direction rather than the focused radio; correct for two options, but if a third option is ever added the "any arrow selects the other one" logic breaks silently. The comment acknowledges the two-option assumption — fine as-is.

Everything else in the incremental changes (locale plural-form backfills for ja/ko/zh/es/pt `components.json`, `sentimentLabel` guard for unknown names, merged-PDF `FeedbackTableSection` tests) looks clean.

---

perrozzi:
Disposition for the remaining items (no code changes — all nits, stopping the review loop here):

- **Sidebar urgent badge / stale `menuStatus` key** — dispositioned in the previous comment (the bot doesn't read comments): badge intentionally lives on the item that opens the list; `isMenuItemEnabled` looks up by `menuKey`, so unknown/stale keys in a cached context are never read.
- **i18n-check array-form `useTranslation` order-dependence** — agreed it's a real subtlety; the current convention (unprefixed keys = `common`, cross-namespace keys always explicitly prefixed in multi-ns files) is what keeps checker and runtime aligned. Noting it for a follow-up hardening of the checker rather than churning this PR.
- **RatingDirectionToggle two-option assumption** — acknowledged in the code comment; fine as-is per the review itself.

PR #210, opened 2026-07-17
feat(webscraper): parse word-based star-rating classes (#148)

### LINKED ISSUE #148: [Feature]: Webscraper rating extraction for word-based star classes (e.g. 'star-rating Three')
## Problem

The webscraper's rating extraction can't parse sites that encode star ratings as **word-based CSS classes** — e.g. `class="star-rating Three"` (the pattern used by books.toscrape.com, a common scraper test target, and similar review widgets). Those reviews ingest with no rating.

On `development` @ `ec4231d`, `plugins/webscraper/ingestor/handler.py:92` `_extract_rating` tries, in order:

1. a configurable `rating_attribute` (e.g. `data-rating`) ✅
2. a digit inside any class name (`re.search(r'(\d+)', cls)`) — misses word classes
3. digits in the element text (`4/5`, `4 stars`, `★`)

There is no number-word mapping, so `One`–`Five` classes fall through and return `None`.

## Fix (small)

In the class-name loop, add a case-insensitive word map before/after the digit check:

```python
WORD_RATINGS = {'one': 1, 'two': 2, 'three': 3, 'four': 4, 'five': 5}
if cls.lower() in WORD_RATINGS:
    return WORD_RATINGS[cls.lower()]
```

Plus a unit test with a `star-rating Three` fixture.

## Assessment

**Priority: low, effort: tiny** — quality-of-ingestion improvement for a common review-site pattern. Good first issue.

perrozzi:
## Summary

Fixes #148. The webscraper's `_extract_rating` only looked for **digits** in an element's CSS class names, so review widgets that encode the star count as a **number word** — e.g. `<p class="star-rating Three">`, the pattern used by [books.toscrape.com](https://books.toscrape.com) and similar widgets — ingested with **no rating**.

## Change

`voc-datalake/plugins/webscraper/ingestor/handler.py`:

- Add a module-level `WORD_STAR_RATINGS = {'one': 1, ... 'five': 5}` map.
- In the existing class-name loop, after the digit check, look up `cls.lower()` in the map and return the mapped value.

Design notes:
- **Class-token scoped, exact match, case-insensitive.** It matches a whole class token (`Three`, `FIVE`), never a substring — `foo-three` does not match, so it can't misfire on unrelated class names.
- **No text-scope matching.** The word map is deliberately *not* applied to element text, to avoid false positives like "one of the best…". Scope matches the issue.
- `rating_attribute` (checked first) and the digit / `x/5` / `stars` / `★` text fallbacks are unchanged. An out-of-range digit token (e.g. `rating-9`) still falls through and a later word token is resolved.

## Tests

New `voc-datalake/plugins/webscraper/ingestor/test_handler.py` (16 tests):

- Word classes: `star-rating Three` → 3, all of one–five, case-insensitivity, whole-token-only (`foo-three` → None), out-of-range-digit-falls-through-to-word.
- Characterization guards for the pre-existing paths: digit-in-class, `rating_attribute` precedence, custom attribute, `4/5` and `3 stars` text, `None` element, no-signal → None.

## Validation (local)

- 16/16 new webscraper tests pass; 5/5 `plugins/test_plugin_imports.py` pass.
- `ruff check` clean; `py_compile` OK; no editor diagnostics.

Backend-only plugin change — no infra or frontend impact.

### DIFF
--- a/voc-datalake/plugins/webscraper/ingestor/handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/handler.py
@@ -20,6 +20,7 @@ import requests
 # Word-based star-rating classes, e.g. <p class="star-rating Three">. Used by
 # books.toscrape.com and similar review widgets that encode the star count as a
 # number word in the element's CSS class rather than a digit or an attribute.
+# Keys MUST be lowercase — lookups normalize the class token via .lower().
 WORD_STAR_RATINGS = {'one': 1, 'two': 2, 'three': 3, 'four': 4, 'five': 5}
 
 
@@ -113,6 +114,9 @@ class WebScraperIngestor(BaseIngestor):
                 rating = int(match.group(1))
                 if 1 <= rating <= 5:
                     return rating
+            # Only consulted on the element the config's rating_selector
+            # resolves to, so a bare 'one'/'two' grid-column class elsewhere
+            # in the DOM can't leak in as a rating.
             word_rating = WORD_STAR_RATINGS.get(cls.lower())
             if word_rating is not None:
                 return word_rating
--- a/voc-datalake/plugins/webscraper/ingestor/handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/handler.py
@@ -17,6 +17,12 @@ from _shared.base_ingestor import BaseIngestor, logger, tracer, metrics, fetch_w
 import requests
 
 
+# Word-based star-rating classes, e.g. <p class="star-rating Three">. Used by
+# books.toscrape.com and similar review widgets that encode the star count as a
+# number word in the element's CSS class rather than a digit or an attribute.
+WORD_STAR_RATINGS = {'one': 1, 'two': 2, 'three': 3, 'four': 4, 'five': 5}
+
+
 class WebScraperIngestor(BaseIngestor):
     """Configurable web scraper for extracting feedback from websites."""
 
@@ -107,6 +113,9 @@ class WebScraperIngestor(BaseIngestor):
                 rating = int(match.group(1))
                 if 1 <= rating <= 5:
                     return rating
+            word_rating = WORD_STAR_RATINGS.get(cls.lower())
+            if word_rating is not None:
+                return word_rating
         
         text = element.get_text(strip=True)
         match = re.search(r'(\d+(?:\.\d+)?)\s*(?:/\s*5|stars?|★)', text, re.I)
--- a/voc-datalake/plugins/webscraper/ingestor/test_handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/test_handler.py
@@ -1,0 +1,99 @@
+"""
+Tests for the Web Scraper Ingestor handler.
+
+Focus: `_extract_rating` rating extraction, including the word-based star
+classes fix (issue #148, e.g. ``<p class="star-rating Three">``). Also
+characterizes the pre-existing digit-class, rating-attribute, and text
+fallbacks so a regression in any of them is caught.
+"""
+
+import pytest
+from unittest.mock import patch, MagicMock
+from bs4 import BeautifulSoup
+
+
+# ---------------------------------------------------------------------------
+# Helpers / fixtures
+# ---------------------------------------------------------------------------
+
+def _el(html: str):
+    """Return the first tag parsed from an HTML snippet."""
+    return BeautifulSoup(html, "html.parser").find()
+
+
+@pytest.fixture
+def ingestor():
+    """Create a WebScraperIngestor with mocked AWS dependencies."""
+    with (
+        patch("_shared.base_ingestor.get_dynamodb_resource") as mock_dynamo,
+        patch("_shared.base_ingestor.get_s3_client"),
+        patch("_shared.base_ingestor.get_sqs_client"),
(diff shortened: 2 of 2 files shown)


perrozzi:
This PR adds word-based star-rating class parsing (`star-rating Three` → 3) to the webscraper ingestor's `_extract_rating`, with a new colocated test file covering the fix and characterizing the pre-existing extraction paths. The change is small, well-scoped, and follows the repo's testing expectations — a regression test exists that fails if the fix is reverted (`test_star_rating_three_returns_3`). No IAM, infra, secret-config, or frontend impact; no new ingestor plugin, so `_get_known_prefixes()` registration is not needed. Overall this looks clean and mergeable.

### Should fix

- **Test file location inside the Lambda handler directory** (`voc-datalake/plugins/webscraper/ingestor/test_handler.py`): if the ingestor Lambda bundles the whole `ingestor/` directory, this test (and its `pytest`/`bs4` test-only usage) ships in the deployment package. Verify the CDK asset bundling excludes `test_*.py`, or move the test to wherever `plugins/test_plugin_imports.py` lives to match the existing test layout.

### Nits

- The word lookup runs on every class token of any element passed through the configured rating selector. CSS frameworks that use bare `one`/`two`/`three` column classes (e.g. Skeleton-style grids) could produce a false rating if a user's `rating_selector` is loose. Risk is contained by the selector, but a brief comment or restricting the lookup to elements whose class list also contains a `rating`-ish token would harden it.
- `WORD_STAR_RATINGS` keys could be documented as lowercase-only (lookup normalizes via `cls.lower()`); a `# keys must be lowercase` comment would prevent a future entry like `'Six'` silently never matching.
- In the fixture, importing `WebScraperIngestor` inside the `patch(...)` context is correct for first import but relies on module import order across the test session (a prior import of `webscraper.ingestor.handler` would bypass the patches at import time). Consider patching at the class/client-call level instead if flakiness appears.

---

perrozzi:
Since the last review, the only changes are two comments added to `handler.py`: a "Keys MUST be lowercase" note on `WORD_STAR_RATINGS` and a scoping note explaining the word lookup only runs on elements matched by the configured `rating_selector`. These directly address two of my previous nits. No code behavior changed.

### Resolved
- Nit: lowercase-only key requirement on `WORD_STAR_RATINGS` is now documented.
- Nit: false-positive risk from bare `one`/`two` grid-column classes is now addressed with a scoping comment.

### Still open
- Should-fix: `test_handler.py` still lives inside the Lambda handler directory (`voc-datalake/plugins/webscraper/ingestor/`) — verify CDK asset bundling excludes `test_*.py`, or relocate it alongside `plugins/test_plugin_imports.py`.

### New findings
None — the incremental changes are comment-only and look clean.

---

perrozzi:
Thanks for the review. Dispositions below — head is now `e075675`.

**Should fix — test file shipping in the Lambda bundle:** Verified this does **not** happen. Plugin ingestor bundles stage from the repo root via `rootPluginAssetExcludes()` (`voc-datalake/lib/utils/lambda-asset-excludes.ts`), whose exclude list contains `test/`, `test_*.py`, and `conftest.py`, applied with `IgnoreMode.GIT` at `voc-datalake/lib/stacks/ingestion-stack.ts:418`. GIT ignore semantics make those patterns unanchored/any-depth, so `plugins/webscraper/ingestor/test_handler.py` is excluded from both the asset hash and `/asset-output`. This is enforced by `lib/utils/lambda-asset-excludes.test.ts`, and is the same mechanism that keeps the existing `plugins/s3_import/ingestor/test_handler.py` out of its bundle. Keeping the test colocated matches that established layout — no change needed.

**Nit — lowercase-key documentation:** Done (`e075675`) — added `# Keys MUST be lowercase — lookups normalize the class token via .lower().`

**Nit — false rating on bare `one`/`two` grid-column classes under a loose `rating_selector`:** Added a comment documenting that the lookup only runs on the element the config's `rating_selector` resolves to (`e075675`). I deliberately did **not** add a "class list must also contain a rating-ish token" guard: the canonical case is `['star-rating', 'Three']`, but widgets that expose a bare `['Three']` under an explicit `.rating` selector are legitimate, and the guard would cut recall. The containment is identical to the pre-existing digit path (a `col-2` class already matches today), so this doesn't add new risk.

**Nit — fixture import inside the `patch(...)` context:** Left as-is; it mirrors the existing `plugins/s3_import/ingestor/test_handler.py` fixture pattern exactly. Happy to switch to call-level patching if flakiness ever surfaces.

Local gates on `e075675`: 16/16 webscraper tests + 5/5 `test_plugin_imports.py` pass, `ruff check` clean, no diagnostics.

perrozzi:
**Ready to merge.** Review converged (round 2: both nits resolved, no new findings; the test-bundling should-fix was verified a non-issue — `test_*.py` is excluded from ingestor bundles by `rootPluginAssetExcludes()` with `IgnoreMode.GIT`, enforced by `lambda-asset-excludes.test.ts`).

Validation on head `e075675`: 16/16 new webscraper tests + 5/5 plugin-import tests, `ruff check` clean.

Deployment note: per maintainer decision, live deploy validation for this change is **cumulated into the next batch deploy** — it's a pure-Python parsing change (no infra/IAM/frontend impact) fully covered by deterministic unit tests, so a dedicated deploy would add no signal.

PR #211, opened 2026-07-17
feat(i18n): language switcher UI — unpin the runtime from English (#147)

### LINKED ISSUE #147: [Feature]: Language switcher UI — i18n runtime is pinned to English despite 8 shipped locales
## Problem

The frontend ships full locale bundles for 8 languages (validated in CI via `i18n:validate`), but the runtime is hard-pinned to English, so none of that translation work is reachable by users.

On `development` @ `ec4231d`, `frontend/src/i18n/config.ts`:

- line 47: `lng: 'en'` — with a comment explicitly saying: *"Pin the UI to English until a real language switcher ships. … Remove this `lng` line in the PR that adds the switcher so localStorage('voc-language') drives it."*
- line 72: a ready-made `changeLanguage(lang)` helper already exists (updates i18next + persists), but **nothing in the UI calls it**.
- `LanguageDetector` with `caches: ['localStorage']` is already wired, so removing the pin + calling the helper is all the plumbing needed.

## Proposed implementation

1. Add a language picker (e.g. in the user menu / `UserProfileModal`, or Settings) listing the 8 shipped locales, calling the existing `changeLanguage()`.
2. Remove the `lng: 'en'` pin per the in-code instruction; `fallbackLng: 'en'` stays.
3. Persisted preference via the existing `voc-language` localStorage cache.
4. Component test + a check that all 8 locales appear in the picker.

## Assessment

**Priority: medium** — all the expensive work (8 complete locale bundles, CI validation, change helper, detector caching) is already done and actively maintained; the missing piece is one small UI control. High value-to-effort ratio.

perrozzi:
## Summary

Fixes #147. The frontend ships 8 complete, CI-validated locale bundles, but the i18n runtime was hard-pinned to English (`lng: 'en'` in `frontend/src/i18n/config.ts`), so none of that translation work was reachable by users. The pin's own comment said to remove it "in the PR that adds the switcher" — this is that PR.

## Change

**`src/i18n/languages.ts` (new)** — side-effect-free home for `supportedLanguages`, `languageNames`, `isSupportedLanguage`, and `changeLanguage`. UI components import from here instead of `./config`, whose module import runs the real `i18n.init()` with the HTTP backend — something component tests (which initialize the *same* i18next singleton with inline resources in `src/test/setup.ts`) must never trigger.

**`src/i18n/config.ts`** — removes the `lng: 'en'` pin per its in-code instruction; the existing `localStorage('voc-language')` detector now drives the language (`fallbackLng: 'en'` keeps first visits English). Adds `supportedLngs` + `nonExplicitSupportedLngs: false` so a stale/garbage cached value can't select a locale we don't ship. Re-exports the constants for backward compatibility. Syncs `<html lang>` on `languageChanged` and after init (screen readers / hyphenation).

**`src/components/UserProfileModal/UserProfileModal.tsx`** — new `LanguageField` on the Profile tab: a labeled `<select>` listing the 8 locales by native name (endonyms, intentionally untranslated), current value from `i18n.resolvedLanguage`, wired to `changeLanguage()`. Persistence comes free via the detector's `caches: ['localStorage']`.

**`public/locales/*/components.json` ×8** — new `userProfile.language` label (Language / Idioma / Langue / Sprache / Idioma / 言語 / 语言 / 언어), inserted formatting-stable (verified byte-identical JSON round-trip before edit).

## Tests

4 new tests in `UserProfileModal.test.tsx` (25 total in the file): picker renders on the Profile tab; all 8 locales present as options with native names; defaults to the active language; `changeLanguage` called on selection (helper mocked via `importOriginal` so the real constants still drive the option assertions).

## Validation (local)

- Full frontend suite: **1768/1768** tests across 112 files.
- `tsc -b --noEmit` clean, `eslint` clean, `npm run i18n:check` exit 0, `vite build` succeeds.
- **Mock-UI E2E (Chrome)**: seeded auth → opened profile modal → picker shows all 8 endonyms with `en` selected → switched to `es` → UI re-rendered live in Spanish ("Mi Perfil", "Idioma", sidebar "Inicio/Panel/Recolectores"), `localStorage['voc-language']='es'`, `<html lang="es">` → survives a full page reload. Reverting to `en` works the same way.

## Pre-existing issues noticed (NOT introduced or addressed here)

- `npm run typecheck:tests` fails ×4 in `src/test/test-utils.tsx` on the base branch (`verbatimModuleSyntax` type-only imports + `MemoryRouter` `future` prop). My diff doesn't touch that file; `typecheck` (production tsconfig) is clean.
- The Home page renders a raw `nav.feedback` key in the "Read the signals" step (leftover from the #209 Feedback→Categories consolidation). Separate fix.


perrozzi:
This PR adds a language switcher to the UserProfileModal, extracts language constants into a side-effect-free `src/i18n/languages.ts` module, removes the `lng: 'en'` pin from `config.ts` (as its own comment instructed), and adds the `userProfile.language` key to all 8 locale bundles. The design is thoughtful: the config/constants split avoids double-initializing the i18next singleton in component tests, `supportedLngs` + `nonExplicitSupportedLngs: false` guards against stale/garbage `localStorage` values, `changeLanguage` uses a type guard rather than an `as` assertion, and `<html lang>` is kept in sync. No circular import between `config.ts` and `languages.ts` (the latter imports the bare `i18next` singleton). Overall this looks clean and well-tested; no blockers.

### Should fix

- **The real `changeLanguage` guard is now untested.** `UserProfileModal.test.tsx` mocks `changeLanguage` via `vi.mock('../../i18n/languages', ...)`, and the logic moved out of `config.ts`, so nothing exercises `isSupportedLanguage` rejection (e.g. `changeLanguage('xx')` resolving without calling `i18n.changeLanguage`). Add a small unit test for `src/i18n/languages.ts` — otherwise a regression in the guard (which is the boundary validation for the `localStorage`-persisted value) would go unnoticed.
- **Unhandled rejection path in the picker.** `onChange={(e) => void changeLanguage(e.target.value)}` in `UserProfileModal.tsx` discards the promise; if `i18n.changeLanguage` ever rejects (e.g. HTTP backend failure loading a namespace), it surfaces as an unhandled rejection with no user feedback and the controlled `<select>` snaps back silently. A `.catch` (even just logging) would harden this.

### Nits

- `UserProfileModal.tsx`: `supportedLanguages.find((lang) => lang === i18n.resolvedLanguage) ?? 'en'` reimplements the type guard — `isSupportedLanguage(i18n.resolvedLanguage ?? '')` from the same module would be more direct.
- `config.ts`: the `languageChanged` handler sets `document.documentElement.lang = lng` with the raw event value, while the post-init sync uses `i18n.resolvedLanguage`. Using `resolvedLanguage ?? lng` in the handler too would keep the two paths consistent (matters if a regional variant ever slips through).
- `languages.ts`: `isSupportedLanguage` constructs a `new Set` on every call; `supportedLanguages.includes(lang as ...)` or a module-level `Set` would be cheaper (trivial at this scale).
- The `passwordsMismatch`/`passwordTooShort` reordering across all 8 `components.json` files is presumably `i18n:check` alphabetical sorting — fine, but it slightly widens the diff; worth a one-line mention in the commit if not already enforced by tooling.
- The "defaults to the active language (English in tests)" test asserts against the label string `'Language'`; using `t`-driven lookups isn't necessary here, but the test is coupled to the test-setup locale being `en` — acceptable, just noting the assumption.

The PR description mentions two pre-existing issues (failing `typecheck:tests` and a raw `nav.feedback` key on Home) — both are indeed outside this diff and reasonable to defer.

---

perrozzi:
The push addresses every item from my previous review: a dedicated unit test file for `languages.ts` now exercises the guard's rejection path, the picker's `onChange` catches and logs failures, the component uses `isSupportedLanguage` directly, the `languageChanged` handler prefers `resolvedLanguage`, and the `Set` is hoisted to module level. Clean, targeted follow-up.

### Resolved
- `changeLanguage` guard now regression-tested (`src/i18n/languages.test.ts` covers supported/unsupported/regional-variant codes and rejection propagation).
- Unhandled rejection in the picker fixed via `.catch` with a logged error (`UserProfileModal.tsx`).
- Nit: component now uses `isSupportedLanguage` instead of re-implementing the guard with `.find()`.
- Nit: `languageChanged` handler now uses `i18n.resolvedLanguage ?? lng`, consistent with the post-init sync.
- Nit: `isSupportedLanguage` uses a module-level `ReadonlySet` instead of allocating per call.

### New findings
- Nit only: in `src/i18n/languages.test.ts`, the test named `'rejects a regional variant not in the shipped list'` is misleadingly named — `changeLanguage('en-US')` *resolves* (the guard silently ignores unsupported codes); the assertion correctly checks the spy wasn't called, so only the name should say "ignores"/"resolves without touching i18next", matching the sibling test.

No blockers; this is good to merge as-is (the naming nit is optional).

---

perrozzi:
Thanks — both should-fixes and three of the nits addressed in `2e814f7`.

**Should fix — untested `changeLanguage` guard:** Done. New `src/i18n/languages.test.ts` (23 tests): `isSupportedLanguage` accept matrix for all 8 shipped locales + reject cases (`xx`, empty, case-sensitive `EN`, regional `en-US`/`de-AT`), `languageNames` completeness, and `changeLanguage` delegation / unsupported-code no-op / regional-variant no-op / rejection propagation — all against a spied i18next singleton so no real locale switch leaks into the test session.

**Should fix — unhandled rejection in the picker:** Done. The `onChange` now `.catch`es and `console.error`s (matches the repo's existing failure-logging convention, e.g. `DocumentExportMenu`). The select is driven by `i18n.resolvedLanguage`, so on failure it stays consistent with the actual active language rather than lying about a switch that didn't happen.

**Nit — inline guard reimplementation:** Done — `LanguageField` now uses `isSupportedLanguage(i18n.resolvedLanguage ?? '')`.

**Nit — `languageChanged` handler raw value:** Done — handler now prefers `i18n.resolvedLanguage ?? lng`, consistent with the post-init sync.

**Nit — per-call `Set` construction:** Done — hoisted to a module-level `ReadonlySet`.

**Nit — locale-file key reordering:** As suspected: the insertion script rebuilds `userProfile` with sorted keys for deterministic diffs; the two swapped lines are that sort, not manual edits. Noted in this reply rather than the commit since it's already merged history in `b092e6e`.

**Nit — test coupled to `en` test-setup locale:** Acknowledged, leaving as-is — the assumption is the documented invariant of `src/test/setup.ts` (inits with `lng: 'en'`), and the test name calls it out ("English in tests").

Gates on `2e814f7`: 44/44 tests in the two touched files, full-suite-affecting changes none since round 1 (config/comments/test-only), `tsc -b` clean, eslint clean, `i18n:check` exit 0.

perrozzi:
The only change since my last review is a one-line rename of the test in `src/i18n/languages.test.ts` from `'rejects a regional variant not in the shipped list'` to `'ignores a regional variant not in the shipped list (resolves without switching)'`, which addresses the sole remaining nit from my previous review.

### Resolved
- Misleading test name fixed — it now accurately describes the guard's silent-ignore behavior, consistent with the sibling test.

No new findings. All previous items are now addressed; this is good to merge.

---

perrozzi:
**Ready to merge.** Review fully converged over 3 rounds: round-1's two should-fixes (guard unit tests, picker rejection handling) and all actionable nits addressed; round-3 confirms "all previous items are now addressed; this is good to merge" with no new findings.

Validation on head `3254bed`: full frontend suite 1768+ green (44/44 in the touched files after the follow-ups), `tsc -b` clean, eslint clean, `i18n:check` exit 0, `vite build` OK, and a live mock-UI E2E in Chrome (switch → Spanish re-render → localStorage persistence → survives reload).

Deployment note: live deploy validation is cumulated into the next batch deploy per maintainer decision (frontend-only change, fully covered by unit + browser E2E validation).

PR #212, opened 2026-07-17
fix(data-explorer): query the GSI that exists — gsi4-by-feedback-id (#140)

### LINKED ISSUE #140: [Bug]: Data Explorer feedback edit/delete query a non-existent GSI ('feedback-id-index' vs 'gsi4-by-feedback-id')
## Problem

`lambda/api/data_explorer_handler.py` queries a GSI named **`feedback-id-index`**, but the feedback table's GSI is defined as **`gsi4-by-feedback-id`** (`lib/stacks/core-stack.ts:261-265`). The queries can never succeed — DynamoDB returns a `ValidationException` (index not found).

Two call sites (`development` @ `ec4231d`):

- `data_explorer_handler.py` ~line 352 — `PUT /data-explorer/feedback` fallback path (when `source_platform` is not supplied in the payload)
- `data_explorer_handler.py` ~line 411 — `DELETE /data-explorer/feedback` (always uses the GSI lookup)

## Impact

- **`DELETE /data-explorer/feedback` always fails** — deleting a processed feedback record from the Data Explorer UI is broken.
- `PUT` (edit) fails whenever the client doesn't send `source_platform`.

Observed live on a test deployment (2026-07-13): every Data Explorer feedback delete failed; records had to be removed via direct DynamoDB access.

## Fix

Replace `IndexName='feedback-id-index'` with `IndexName='gsi4-by-feedback-id'` at both call sites. The key schema already matches (`feedback_id` partition key, `ProjectionType.ALL`). Add a regression test that fails if the handler queries an index name not defined in the table construct (or at minimum a mocked test pinning the correct index name).

## Assessment

**Priority: high** (user-facing feature fully broken), **effort: trivial** (2-line fix + test).

perrozzi:
## Summary

Fixes #140. `save_feedback` (in its GSI-lookup branch, i.e. whenever the edit payload doesn't carry `pk`/`sk`) and `delete_feedback` in `lambda/api/data_explorer_handler.py` queried `IndexName='feedback-id-index'` — an index that has never existed on the feedback table. The real index is **`gsi4-by-feedback-id`** (`lib/stacks/core-stack.ts`, partition key `feedback_id`, projection ALL). Result: every Data Explorer record **delete** failed with a DynamoDB `ValidationException` (surfaced as a 500), as did edits without inline keys.

## Change

- `data_explorer_handler.py`: `IndexName='feedback-id-index'` → `IndexName='gsi4-by-feedback-id'` at both query sites. Nothing else — key condition (`feedback_id = :fid`) already matches the GSI's partition key.

## Tests

Existing tests mocked the table, so the bad index name never surfaced. Added two regression tests in `test_data_explorer_handler.py` that pin the `IndexName` kwarg on both paths (`test_lookup_queries_the_real_gsi_name`, `test_delete_queries_the_real_gsi_name`) — a future rename drift between the stack and the handler now fails the suite instead of the live API.

## Validation (local)

- 20/20 tests in `test_data_explorer_handler.py` (18 existing + 2 new), `ruff check` clean.
- Repo-wide sweep: no remaining `feedback-id-index` references outside the regression-test docstrings.

Deployment note: live validation cumulated into the next batch deploy per maintainer decision.

### DIFF
--- a/voc-datalake/lambda/api/data_explorer_handler.py
+++ b/voc-datalake/lambda/api/data_explorer_handler.py
@@ -29,6 +29,9 @@ sqs_client = get_sqs_client()
 
 RAW_DATA_BUCKET = os.environ.get("RAW_DATA_BUCKET", "")
 FEEDBACK_TABLE = os.environ.get("FEEDBACK_TABLE", "")
+# Feedback-id lookup GSI — must match core-stack.ts (gsi4-by-feedback-id).
+# Regression-tested in test_data_explorer_handler.py (#140).
+FEEDBACK_ID_INDEX = 'gsi4-by-feedback-id'
 PROCESSING_QUEUE_URL = os.environ.get("PROCESSING_QUEUE_URL", "")
 
 # Available buckets for browsing
@@ -349,7 +352,7 @@ def save_feedback():
         else:
             # Need to find the item first
             response = table.query(
-                IndexName='gsi4-by-feedback-id',
+                IndexName=FEEDBACK_ID_INDEX,
                 KeyConditionExpression='feedback_id = :fid',
                 ExpressionAttributeValues={':fid': feedback_id},
                 Limit=1
@@ -412,7 +415,7 @@ def delete_feedback():
         
         # Find the item first using GSI
         response = table.query(
-            IndexName='gsi4-by-feedback-id',
+            IndexName=FEEDBACK_ID_INDEX,
             KeyConditionExpression='feedback_id = :fid',
             ExpressionAttributeValues={':fid': feedback_id},
             Limit=1
--- a/voc-datalake/lambda/api/data_explorer_handler.py
+++ b/voc-datalake/lambda/api/data_explorer_handler.py
@@ -349,7 +349,7 @@ def save_feedback():
         else:
             # Need to find the item first
             response = table.query(
-                IndexName='feedback-id-index',
+                IndexName='gsi4-by-feedback-id',
                 KeyConditionExpression='feedback_id = :fid',
                 ExpressionAttributeValues={':fid': feedback_id},
                 Limit=1
@@ -412,7 +412,7 @@ def delete_feedback():
         
         # Find the item first using GSI
         response = table.query(
-            IndexName='feedback-id-index',
+            IndexName='gsi4-by-feedback-id',
             KeyConditionExpression='feedback_id = :fid',
             ExpressionAttributeValues={':fid': feedback_id},
             Limit=1
--- a/voc-datalake/lambda/api/test/test_data_explorer_handler.py
+++ b/voc-datalake/lambda/api/test/test_data_explorer_handler.py
@@ -329,6 +329,7 @@ class TestSaveFeedback:
 
         # Assert
         assert response['statusCode'] == 200
+        mock_table.query.assert_called_once()
         assert mock_table.query.call_args.kwargs['IndexName'] == 'gsi4-by-feedback-id'
 
     @patch('data_explorer_handler.dynamodb')
@@ -412,6 +413,7 @@ class TestDeleteFeedback:
 
         # Assert
         assert response['statusCode'] == 200
+        mock_table.query.assert_called_once()
         assert mock_table.query.call_args.kwargs['IndexName'] == 'gsi4-by-feedback-id'
 
     @patch('data_explorer_handler.dynamodb')
--- a/voc-datalake/lambda/api/test/test_data_explorer_handler.py
+++ b/voc-datalake/lambda/api/test/test_data_explorer_handler.py
@@ -300,6 +300,38 @@ class TestSaveFeedback:
         assert body['success'] is True
 
     @patch('data_explorer_handler.dynamodb')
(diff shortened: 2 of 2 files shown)


perrozzi:
This PR fixes a real bug: `save_feedback` (GSI-lookup branch) and `delete_feedback` in `voc-datalake/lambda/api/data_explorer_handler.py` queried a nonexistent index (`feedback-id-index`) instead of the actual `gsi4-by-feedback-id` defined in `lib/stacks/core-stack.ts`, causing every delete and key-less edit to 500 with a DynamoDB `ValidationException`. The two-line fix is correct (the key condition on `feedback_id` matches the GSI partition key), and the PR includes regression tests per repo policy. Overall this looks clean and mergeable.

### Should fix
- **The regression tests only guard one direction of drift.** `test_lookup_queries_the_real_gsi_name` and `test_delete_queries_the_real_gsi_name` pin the literal string `'gsi4-by-feedback-id'` in both the handler and the test. If the index is renamed in `core-stack.ts`, both handler and test would still agree and pass while the live API breaks again — exactly the failure mode described in the PR. Consider having the CDK stack pass the index name to the Lambda via an environment variable (as is presumably done for the table name), and have the handler read it. Then the test can assert against that single source. Acceptable to defer, but worth a follow-up issue if not done here.

### Nits
- The GSI name is now duplicated as a string literal at two call sites in `data_explorer_handler.py`. At minimum, hoist it to a module-level constant (e.g. `FEEDBACK_ID_INDEX = 'gsi4-by-feedback-id'`) so a future rename touches one line.
- The tests assert on `mock_table.query.call_args.kwargs`, i.e. the *last* call. With a single query per path this is fine today, but `assert_called_once()` alongside would make the tests more robust if the handler ever adds a second query.
- Minor duplication between the two new tests (mock setup, item fixture) — a shared fixture or helper would keep them tighter, consistent with the rest of `test_data_explorer_handler.py`.

No blockers. The fix is minimal, scoped to one domain handler (consistent with the per-domain Lambda split), touches no IAM grants, and the added tests would fail if the fix were reverted.

---

perrozzi:
Since the last review, the author hoisted the GSI name into a module-level constant `FEEDBACK_ID_INDEX` (with a comment pointing at `core-stack.ts` and the regression tests) and added `assert_called_once()` to both regression tests. Two of the three nits are addressed; the deferred should-fix remains.

### Resolved
- GSI name literal no longer duplicated at two call sites — hoisted to `FEEDBACK_ID_INDEX` in `data_explorer_handler.py`.
- Tests now assert `mock_table.query.assert_called_once()` in addition to checking `call_args.kwargs`, guarding against a second query being added later.

### Still open
- The index name is still a hardcoded literal in both handler and tests rather than sourced from the CDK stack via an environment variable (like `FEEDBACK_TABLE`), so a rename in `core-stack.ts` would still pass the suite while breaking live. Was flagged as acceptable to defer — a follow-up issue should be filed if this merges as-is.

### New findings
None of substance. The only trivial observation: `FEEDBACK_ID_INDEX = 'gsi4-by-feedback-id'` uses single quotes while adjacent module constants use double quotes — cosmetic only, and reportedly passes `ruff check`.

The incremental changes are clean; this remains mergeable.

---

perrozzi:
Dispositions — head is now `837f332`.

**Should fix — pin the name via a stack-passed env var:** Declining for this PR, deliberately. Hardcoded `IndexName` literals are the established repo-wide convention — `metrics_handler.py` alone has 8 of them (including `gsi4-by-feedback-id` twice), `shared/feedback.py` and the streaming Lambda's tools do the same, and no stack passes index names through the environment today. Plumbing an env var for just this handler would leave the same drift risk everywhere else while making one handler inconsistent. The right version of your suggestion is a repo-wide refactor (constants/env for all GSI names across all handlers + stacks), which is out of scope for a two-string bug fix. Mitigation here: the name is now a module-level constant with a comment pointing at `core-stack.ts`, and the regression tests pin the value that the stack actually defines — so the specific regression this PR fixes (handler drifting from the deployed index) is covered.

**Nit — module-level constant:** Done (`837f332`) — `FEEDBACK_ID_INDEX = 'gsi4-by-feedback-id'` next to the other module config, both call sites use it.

**Nit — `assert_called_once` alongside the kwargs assert:** Done (`837f332`) in both regression tests.

**Nit — shared fixture for the two tests' mock setup:** Left as-is; each test in this file arranges its own mocks inline (that *is* the file's existing convention — see `TestSaveFeedback`/`TestDeleteFeedback`), and the two tests intentionally exercise different HTTP paths with different table method expectations.

Gates on `837f332`: 20/20 tests, ruff clean.

perrozzi:
**Ready to merge.** Round 2 confirms the incremental changes are clean and the PR mergeable; both actionable nits are done, and the deferred should-fix (stack-sourced index names, repo-wide) is now tracked as follow-up issue #213 as requested.

Validation on head `837f332`: 20/20 handler tests (2 new regression tests pinning the IndexName on both paths), ruff clean, no remaining `feedback-id-index` references.

Deployment note: live validation cumulated into the next batch deploy per maintainer decision — the regression tests cover the drift, and the fix restores the exact query shape the GSI serves.

PR #214, opened 2026-07-17
fix(webscraper): clear secret cache on manual runs — Save-then-Run-now stale read (#141)

### LINKED ISSUE #141: [Bug]: Webscraper Save-then-Run-now fails on stale secret cache ('No scraper configuration found')
## Problem

Saving a new/edited web-scraper config and immediately clicking **Run now** fails with **"No scraper configuration found"** (Pages scraped: 0), even though the config saved correctly. A retry ~30s later succeeds.

## Root cause (`development` @ `ec4231d`)

- `lambda/shared/aws.py:105` — `get_secret` is `@lru_cache(maxsize=10)` with **no TTL**; a warm ingestor Lambda serves the pre-save secret value.
- The app-review ingestors already guard against this: `plugins/app_reviews_android/ingestor/handler.py:205-206` and `plugins/app_reviews_ios/ingestor/handler.py:188-189` call `clear_secret_cache()` on manual runs.
- `plugins/webscraper/ingestor/handler.py` never calls `clear_secret_cache()` — the only manual-run ingestor that doesn't.

Confirmed live: the config was present in Secrets Manager at the time of the failure, so it's a stale read, not a save failure.

## Fix

Match the app-review pattern: call `clear_secret_cache()` at the start of a `manual_trigger` invocation in the webscraper handler (one import + one call). Optionally, add a short TTL to `get_secret` as a systemic guard for all plugins.

## Assessment

**Priority: medium** — confusing first-run UX on the most-used source type; users reasonably conclude their config didn't save. **Effort: trivial.**

perrozzi:
## Summary

Fixes #141. Saving a new/edited web-scraper config and immediately clicking **Run now** failed with **"No scraper configuration found"** (Pages scraped: 0) even though the config saved correctly; a retry ~30s later succeeded.

Root cause: `get_secret` in `lambda/shared/aws.py` is `@lru_cache` with **no TTL**, so a **warm** webscraper container invoked right after the save serves the pre-save secret snapshot — a stale read, not a save failure. The iOS/Android app-review ingestors already guard against exactly this by calling `clear_secret_cache()` at the top of manual (`execution_id`) invocations; the webscraper was the only manual-run ingestor without the guard.

## Change

`plugins/webscraper/ingestor/handler.py` — in `lambda_handler`, when the event carries an `execution_id` (manual Run-now), call `clear_secret_cache()` **before** constructing the ingestor (which reads the secret at init). Identical pattern and comment style as `app_reviews_ios`/`app_reviews_android`. Scheduled runs are untouched and keep the warm-cache benefit.

## Tests

3 regression tests in `plugins/webscraper/ingestor/test_handler.py`:
- manual run clears the cache;
- scheduled run does **not** (warm cache preserved);
- ordering: the clear happens **before** ingestor construction — clearing after would be a no-op since the stale snapshot would already be loaded (this is the assertion that fails if someone "simplifies" the fix by moving the call).

## Validation (local)

19/19 tests in the file (16 existing + 3 new), `ruff check` clean.

Deployment note: cumulated into the next batch deploy per maintainer decision (needs `VocIngestionStack` for the webscraper Lambda; the batch already owes #210/#211/#212).

### DIFF
--- a/voc-datalake/plugins/webscraper/ingestor/handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/handler.py
@@ -425,7 +425,14 @@ def lambda_handler(event, context):
     """Lambda entry point."""
     execution_id = event.get('execution_id')
     scraper_id = event.get('scraper_id')
-    
+
+    # Clear secret cache on manual runs so a warm container picks up a config
+    # saved moments ago (Save-then-Run-now, issue #141). get_secret is
+    # lru_cached without TTL; same guard as the app-review ingestors.
+    if execution_id:
+        from shared.aws import clear_secret_cache
+        clear_secret_cache()
+
     ingestor = WebScraperIngestor(
         execution_id=execution_id,
         target_scraper_id=scraper_id
--- a/voc-datalake/plugins/webscraper/ingestor/test_handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/test_handler.py
@@ -148,6 +148,7 @@ class TestLambdaHandlerSecretCache:
         lambda_handler({}, lambda_context)
 
         mock_clear.assert_not_called()
+        MockIngestor.assert_called_once()
 
     @patch("shared.aws.clear_secret_cache")
     @patch("webscraper.ingestor.handler.WebScraperIngestor")
@@ -157,10 +158,16 @@ class TestLambdaHandlerSecretCache:
         """Order matters: clearing after the ingestor loads its config would
         be a no-op — the stale snapshot would already be in memory."""
         call_order = []
-        mock_clear.side_effect = lambda: call_order.append("clear")
-        MockIngestor.side_effect = lambda **kw: call_order.append("ingestor") or MagicMock(
-            run=MagicMock(return_value={"status": "success"})
-        )
+
+        def record_clear():
+            call_order.append("clear")
+
+        def record_ingestor(**_kwargs):
+            call_order.append("ingestor")
+            return MagicMock(run=MagicMock(return_value={"status": "success"}))
+
+        mock_clear.side_effect = record_clear
+        MockIngestor.side_effect = record_ingestor
 
         from webscraper.ingestor.handler import lambda_handler
         lambda_handler({"execution_id": "exec-1"}, lambda_context)
--- a/voc-datalake/plugins/webscraper/ingestor/test_handler.py
+++ b/voc-datalake/plugins/webscraper/ingestor/test_handler.py
@@ -4,7 +4,8 @@ Tests for the Web Scraper Ingestor handler.
 Focus: `_extract_rating` rating extraction, including the word-based star
 classes fix (issue #148, e.g. ``<p class="star-rating Three">``). Also
 characterizes the pre-existing digit-class, rating-attribute, and text
-fallbacks so a regression in any of them is caught.
+fallbacks so a regression in any of them is caught. Plus the manual-run
+secret-cache clear (issue #141).
 """
 
 import pytest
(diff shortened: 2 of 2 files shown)


perrozzi:
This PR fixes the Save-then-Run-now stale-secret bug (#141) in the webscraper ingestor by clearing the `lru_cache`d secret on manual runs (`execution_id` present), mirroring the existing guard in the iOS/Android app-review ingestors. The fix is minimal, correctly placed before ingestor construction (which reads the secret at init), leaves scheduled runs on the warm cache, and comes with regression tests including an ordering assertion that fails if the fix is reverted or moved. Overall this looks correct and well-scoped.

### Should fix

- **`plugins/webscraper/ingestor/test_handler.py` — missing mock imports?** The new tests use `patch` and `MagicMock`, but the diff shows only `import pytest` in the visible header context and adds no `from unittest.mock import patch, MagicMock`. The pre-existing tests in this file (rating-extraction characterization) don't appear to use mocks, so this may raise `NameError` at collection time. Please confirm the import exists outside the diff context or add it.
- **This is the third copy of the same guard.** The stale-read hazard exists for *any* manual-run ingestor whose secret is `lru_cache`d without TTL. Rather than repeating the `if execution_id: clear_secret_cache()` pattern per plugin, consider a follow-up that centralizes it (e.g., in `base_ingestor` or a shared handler wrapper, or a TTL on `get_secret` in `lambda/shared/aws.py`). Not blocking, but worth an issue so the next new ingestor doesn't reintroduce #141.

### Nits

- `handler.py`: the function-local `from shared.aws import clear_secret_cache` is fine (and keeps the test patch target working), but if the app-review ingestors import it at module level, match that for consistency.
- The ordering test's `MockIngestor.side_effect = lambda **kw: call_order.append("ingestor") or MagicMock(...)` works but is a bit clever; a small named function would be more readable.
- `test_scheduled_run_keeps_warm_cache` asserts the negative but doesn't assert the ingestor still ran; adding `MockIngestor.assert_called_once()` would strengthen it.

One process note: the change touches only the webscraper Lambda code, so per the description it needs the `VocIngestionStack` deploy — the "batched deploy" note in the description means the fix won't be live until that batch ships; make sure #141 isn't closed until then.

---

perrozzi:
The incremental changes touch only `test_handler.py` and address two of my nits: the ordering test's clever lambda is now two named helper functions, and the scheduled-run test now asserts the ingestor still ran. The handler itself is unchanged.

### Resolved
- Ordering test refactored from the `... or MagicMock(...)` lambda to readable named `record_clear`/`record_ingestor` functions.
- `test_scheduled_run_keeps_warm_cache` now asserts `MockIngestor.assert_called_once()` alongside the negative check.

### Still open
- **Missing mock imports** — the tests use `patch`/`MagicMock`, but neither diff shows `from unittest.mock import patch, MagicMock` being added; the visible header still shows only `import pytest`. Please confirm it exists outside the diff context or add it (would be a `NameError` at collection otherwise).
- Third copy of the `if execution_id: clear_secret_cache()` guard — still worth a tracking issue to centralize (shared wrapper or TTL on `get_secret`), not blocking.

### New findings
None — the incremental changes are clean test improvements.

Process reminder stands: don't close #141 until the batched `VocIngestionStack` deploy ships.

---

perrozzi:
Dispositions — head is now `e3f9bea`.

**Should fix — missing mock imports:** False alarm — `from unittest.mock import patch, MagicMock` is at line 10 of `test_handler.py`, present since the file was created in #210 (the pre-existing rating fixture uses `patch`/`MagicMock` too, in the `ingestor` fixture). It just sits outside the diff hunk context.

**Should fix — third copy of the guard:** Agreed on the hazard; filed as follow-up issue **#215** with both centralization options (guard in the shared entry path vs. TTL on `get_secret`) — the parked design note for #141 explicitly recommended the targeted per-plugin fix first and deferred the systemic one, and this PR follows that. Keeping this diff minimal and consistent with the two existing precedents.

**Nit — function-local import consistency:** It already matches — both app-review ingestors do the function-local `from shared.aws import clear_secret_cache` inside `lambda_handler` (e.g. `app_reviews_ios/ingestor/handler.py:188`). Kept identical deliberately.

**Nit — clever lambda in the ordering test:** Done (`e3f9bea`) — replaced with named `record_clear`/`record_ingestor` functions.

**Nit — strengthen the scheduled-run test:** Done (`e3f9bea`) — added `MockIngestor.assert_called_once()`.

**Process note — don't close #141 before the deploy:** Acknowledged; per maintainer decision the deploy is batched (#210/#211/#212/#214). The issue will be closed with the standard "lands on main with the next release" note, consistent with how every dev-fixed issue is tracked in this repo.

Gates on `e3f9bea`: 19/19 tests, ruff clean.

perrozzi:
**Ready to merge.** Round 2: both actionable nits resolved, no new findings. The two "still open" items are already dispositioned: the mock imports exist at `test_handler.py:10` (present since #210; a 19/19 green pytest run would be impossible with a collection-time `NameError`), and the guard-centralization follow-up is filed as #215.

Validation on head `e3f9bea`: 19/19 tests (3 new regression tests incl. the clear-before-construct ordering assertion), ruff clean.

Deployment note: batched — this joins #210/#211/#212 in the pending `VocIngestionStack`/frontend batch deploy; #141 gets the standard "lands on main with the next release" close note per repo convention.

Issue #149, opened 2026-07-13
[Discussion]: Rationalize the five manual feedback-ingestion entry points


perrozzi:
## Observation

The UI now exposes **five** ways to get feedback into the lake manually, several overlapping in mechanism (`development` @ `ec4231d`):

| Entry point | Route/mechanism | Primary intent |
|---|---|---|
| Manual Import (paste) | `/scrapers/manual/parse` + `/confirm` (LLM parse) → SQS | one-time, unstructured text |
| JSON Upload | `POST /scrapers/manual/json-upload` → S3 + SQS | one-time structured bulk |
| **CSV Upload** (new in #139) | `POST /scrapers/manual/csv-upload` → SQS batch | one-time structured bulk, ≤50k rows |
| S3 Bulk Import | presigned upload → import bucket → ingestor | recurring/managed bulk (source card) |
| Data Explorer create + sync | `PUT /data-explorer/s3` (+1 SQS msg) | single-record repair |

This is a coherence/discoverability question, not a bug — each path has a distinct primary purpose, but a user picking among the **three structured-bulk paths** (JSON Upload, CSV Upload, S3 Bulk Import) has no guidance on which to use, and the paths differ subtly (caps, dedup behavior, whether a managed source card is created, raw-S3 archival).

## Options to consider

1. **Clarify in-modal copy** — one line per option explaining when to use it ("quick one-off batch" vs "recurring/large files with a managed source"). Cheapest.
2. **Consolidate** — fold JSON/CSV Upload into a single "Upload file" flow, or make them thin modes of S3 Bulk Import.
3. **Document a decision tree** in `docs/` and link it from the Add Source dialog.

## Assessment

**Priority: low-medium, product/UX discussion** — worth deciding deliberately before a sixth path appears. Option 1 is a near-zero-cost immediate improvement regardless of the longer-term answer.

mundurragacl:
## Architecture audit: have the five doors broken the single-pipeline promise?

Maintainer question: the plugin architecture's whole point was ONE ingestion path — same logic, same dedupe — with plugins doing their own ETL and converging into an entry point we control. Did we drift? Code audit (`development` @ `29a1186`):

### The funnel itself held ✅

[...]

PR #216, opened 2026-07-17
feat(processor): persist ingestion_method provenance on feedback records (#145)

### LINKED ISSUE #145: [Feature]: Persist ingestion_method provenance on feedback records
## Problem

Manual imports set an `ingestion_method` on their SQS messages, but the processor drops it — so after ingest, records from the paste flow, CSV upload, and JSON upload are all indistinguishable (`source_platform = 'manual_import'` for all of them).

On `development` @ `ec4231d` there are now **three senders** that set the field:

- `lambda/api/manual_import_handler.py:299` — `'ingestion_method': 'manual'`
- `manual_import_handler.py:524` — `'ingestion_method': 'csv_upload'`
- `manual_import_handler.py:622` — `'ingestion_method': 'json_upload'`

…and zero readers: `lambda/processor/handler.py` builds the persisted item from an explicit field whitelist (~line 525 reads `source_channel` etc.) that never reads `ingestion_method`, so the value never reaches DynamoDB or the API.

## Fix (small)

1. `processor/handler.py` — add `'ingestion_method': raw_record.get('ingestion_method')` to the persisted item dict (None-stripping means sources that don't send it simply omit the field; no migration).
2. `frontend/src/api/types.ts` — add `ingestion_method?: string` to `FeedbackItem`.
3. Processor test asserting the field is carried through when present and absent when not.
4. Optional: document in `schemas/feedback-event.schema.json` and surface in the Data Explorer detail view.

## Assessment

**Priority: low, effort: tiny** — but more valuable now than when first observed, since three distinct upstream values already exist and are silently discarded. Good first issue.

perrozzi:
## Summary

Fixes #145. Three senders in `manual_import_handler.py` already stamp their SQS messages with `ingestion_method` (`'manual'`, `'csv_upload'`, `'json_upload'`), but the processor's persisted-item whitelist never read the field — so it was silently discarded and paste/CSV/JSON imports were indistinguishable after ingest (`source_platform = 'manual_import'` for all).

## Change

- `lambda/processor/handler.py` — add `'ingestion_method': raw_record.get('ingestion_method')` to the persisted item dict, grouped with the other source/provenance fields. The existing None-stripping means sources that don't send it (scrapers, app reviews, forms) simply omit the field: **no migration, no spurious empty attributes**.
- `frontend/src/api/types.ts` — `FeedbackItem.ingestion_method?: string` (optional, doc-commented).
- `schemas/feedback-event.schema.json` — documents the optional field (not added to `required`).

Exactly Option A from the issue; Option B's paste-flow tagging is already on `development` (the `'manual'` sender).

## Tests

Two processor regression tests: `test_persists_ingestion_method_when_sent` (field carried through to the item) and `test_omits_ingestion_method_when_not_sent` (records without it don't gain a spurious/empty field) — the two acceptance criteria from the issue.

## Validation (local)

- 51/51 processor tests (49 existing + 2 new), `ruff check` clean.
- Frontend `tsc -b` clean; schema file validates as JSON.

Deployment note: processor change → `VocProcessingStack`; joins the pending batch deploy (#210/#211/#212/#214) per maintainer decision. No backfill — only newly ingested items carry the field.

### DIFF
--- a/voc-datalake/frontend/src/api/feedbackSchema.test.ts
+++ b/voc-datalake/frontend/src/api/feedbackSchema.test.ts
@@ -62,6 +62,18 @@ describe('normalizeFeedbackItem', () => {
     const item = normalizeFeedbackItem(rawItem({ sentiment_score: -0.75 }))
     expect(item.sentiment_score).toBe(-0.75)
   })
+
+  it('preserves ingestion_method through normalization (#145)', () => {
+    // Regression: the schema strips unknown keys, so a field missing from it
+    // would be silently dropped at the API boundary even if persisted.
+    const item = normalizeFeedbackItem(rawItem({ ingestion_method: 'csv_upload' }))
+    expect(item.ingestion_method).toBe('csv_upload')
+  })
+
+  it('leaves ingestion_method undefined when the record does not carry it', () => {
+    const item = normalizeFeedbackItem(rawItem())
+    expect(item.ingestion_method).toBeUndefined()
+  })
 })
 
 describe('normalizeFeedbackItems', () => {
--- a/voc-datalake/frontend/src/api/feedbackSchema.ts
+++ b/voc-datalake/frontend/src/api/feedbackSchema.ts
@@ -49,6 +49,7 @@ export const FeedbackItemSchema = z.object({
   source_id: lenientString,
   source_platform: lenientString,
   source_channel: lenientString,
+  ingestion_method: z.string().optional(),
   source_url: z.string().optional(),
   brand_name: lenientString,
   source_created_at: lenientString,
--- a/voc-datalake/frontend/src/api/types.ts
+++ b/voc-datalake/frontend/src/api/types.ts
@@ -5,6 +5,8 @@ export interface FeedbackItem {
   source_id: string
   source_platform: string
   source_channel: string
+  /** Ingestion-path provenance, e.g. 'manual', 'csv_upload', 'json_upload'. Absent for sources that don't send it. */
+  ingestion_method?: string
   source_url?: string
   brand_name: string
   source_created_at: string
--- a/voc-datalake/lambda/processor/handler.py
+++ b/voc-datalake/lambda/processor/handler.py
@@ -528,6 +528,9 @@ def process_feedback(raw_record: dict, idempotency_key: str = None) -> dict:
         'source_id': raw_record.get('id', ''),
         'source_platform': source_platform,
         'source_channel': raw_record.get('source_channel', 'unknown'),
+        # Ingestion-path provenance (e.g. 'manual', 'csv_upload', 'json_upload').
+        # Optional: sources that don't send it omit the field via None-stripping.
+        'ingestion_method': raw_record.get('ingestion_method'),
         'source_url': raw_record.get('url'),
         'brand_name': source_display,
         'source_created_at': raw_record.get('created_at'),
--- a/voc-datalake/lambda/processor/test/test_handler.py
+++ b/voc-datalake/lambda/processor/test/test_handler.py
@@ -498,6 +498,55 @@ class TestProcessFeedback:
     @patch('processor.handler.translate_text')
     @patch('processor.handler.detect_language')
     @patch('processor.handler.check_duplicate')
+    def test_persists_ingestion_method_when_sent(
+        self, mock_check_dup, mock_detect, mock_translate, mock_sentiment, mock_llm,
+        sample_sqs_record, sample_llm_insights
+    ):
+        """Carries ingestion_method provenance to the persisted item (#145)."""
+        from processor.handler import process_feedback
+
+        mock_check_dup.return_value = False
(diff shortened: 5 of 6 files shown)


perrozzi:
This PR threads the existing `ingestion_method` field from SQS messages (already stamped by `manual_import_handler.py`'s three senders) through the processor's persisted-item dict, adds the optional field to the frontend `FeedbackItem` type and the JSON schema, and includes two regression tests covering both acceptance criteria. The change is small, well-scoped, and correctly relies on the processor's existing None-stripping so non-manual sources are unaffected without a migration. Overall it looks clean; the first test is a proper regression test that would fail if the one-line fix were reverted.

### Should fix
- **Zod validation coverage** (`frontend/src/api/types.ts`): the repo pattern is Zod for runtime validation at boundaries. Only the TypeScript interface was updated — if `FeedbackItem` responses are parsed through a Zod schema anywhere (e.g. an API-layer `feedbackItemSchema`), the new field should be added there too (`ingestion_method: z.string().optional()`), otherwise a strict schema could strip or reject the field at runtime while `tsc` stays green. Please confirm or update.

### Nits
- `schemas/feedback-event.schema.json`: the value set is closed and controlled by internal senders (`manual`, `csv_upload`, `json_upload`), so an `enum` would give stronger boundary validation than a free-form `string`. If you intentionally keep it open for future ingestors, current form is fine.
- `frontend/src/api/types.ts`: same idea — a union like `'manual' | 'csv_upload' | 'json_upload' | (string & {})` would give autocomplete for known values while staying open-ended.
- `test_omits_ingestion_method_when_not_sent`: the precondition `assert 'ingestion_method' not in sample_sqs_record` guards against fixture drift — good — but if the fixture ever gains the field, the failure message won't make the reason obvious; a short assert message would help.
- The five stacked `@patch` decorators are now duplicated across many tests in this class; a shared fixture or a helper that applies the standard mocks would reduce boilerplate (pre-existing pattern, not introduced here).

No blockers. Deployment note (processor → `VocProcessingStack`, batched with the pending deploys) matches the stated maintainer decision; no IAM, Bedrock model-id, or ingestor-prefix registration concerns are touched by this diff.

---

perrozzi:
The new push adds `ingestion_method: z.string().optional()` to the Zod `FeedbackItemSchema` and two boundary regression tests confirming the field survives normalization when present and stays absent when not sent — directly addressing the one Should-fix item from my previous review.

### Resolved
- Zod validation coverage: `frontend/src/api/feedbackSchema.ts` now includes `ingestion_method`, with regression tests in `feedbackSchema.test.ts` covering both the pass-through and absent cases (the first test would fail against a strip-unknown-keys schema if the fix were reverted — good).

### New findings
None. The nits from the earlier review (enum in the JSON schema, string-literal union in `types.ts`, assert message in the omission test) remain optional and unaddressed, but none block merge.

The incremental changes are clean and consistent with the rest of the PR; no IAM, powertools, Bedrock, or i18n surface is touched. Good to merge.

---

perrozzi:
Dispositions — head is now `efea4fb`.

**Should fix — Zod boundary schema:** Confirmed and fixed — good catch, this was material. `src/api/feedbackSchema.ts`'s `FeedbackItemSchema` normalizes all `/feedback*` responses and **strips unknown keys**, so the field would have been persisted by the processor and then silently dropped at the frontend boundary. Added `ingestion_method: z.string().optional()` to the schema plus two regression tests (`preserves ingestion_method through normalization`, `leaves it undefined when absent`) in `feedbackSchema.test.ts`.

**Nit — schema/TS enum vs open string:** Intentionally open, keeping `string`. The value set isn't closed: any future ingestor/webhook may stamp its own method (the follow-up direction in #149 explicitly anticipates more adapters), and an `enum` in `feedback-event.schema.json` or a closed TS union would turn each new sender into a schema-breaking change. The schema description documents the known values.

**Nit — assert message on the fixture-drift precondition:** Applied in spirit via the test docstring; the bare `assert 'ingestion_method' not in sample_sqs_record` failing points directly at the fixture line. Leaving as-is to match the file's assert style.

**Nit — five stacked `@patch` decorators duplication:** Agreed it's boilerplate, but it's the file's established pattern across ~10 pre-existing tests; refactoring the class's mock setup is out of scope for this PR.

Gates on `efea4fb`: 51/51 processor tests, 10/10 feedbackSchema tests, `tsc -b` clean, eslint clean, ruff clean.

perrozzi:
**Ready to merge.** Round 2 confirms: the one should-fix (Zod boundary coverage — a genuinely material catch, the schema would have stripped the field) is resolved with regression tests; remaining nits are optional-by-review and dispositioned (open value set is intentional).

Validation on head `efea4fb`: 51/51 processor tests (2 new), 10/10 feedbackSchema tests (2 new), `tsc -b` clean, eslint clean, ruff clean, schema JSON valid.

Deployment note: processor change rides the pending batch deploy (#210/#211/#212/#214 + this); no backfill — new items only.

PR #219, opened 2026-07-17
refactor: single source of truth for DynamoDB GSI names (#213)

### LINKED ISSUE #213: [Tech debt]: Single source of truth for DynamoDB GSI names (stack ↔ handler drift)
## Problem

DynamoDB GSI names are hardcoded string literals across all Python handlers (`metrics_handler.py` has 8, `data_explorer_handler.py`, `shared/feedback.py`, the streaming Lambda's tools, …), while the tables/indexes are defined separately in `lib/stacks/core-stack.ts`. Nothing enforces agreement: a rename in the stack would pass every test (they pin the same literals) and break the live API — exactly the failure mode behind #140, where a handler queried the never-existing `feedback-id-index` instead of `gsi4-by-feedback-id`.

Flagged by review on PR #212, which fixed #140's instance but deliberately deferred the structural fix as out of scope for a two-string bug fix.

## Proposed direction

Single source of truth for index names, repo-wide (not one handler):

1. Define index-name constants in one place per language boundary — e.g. the stacks pass `*_INDEX` environment variables next to the existing `FEEDBACK_TABLE`-style vars, or a shared `lambda/shared/indexes.py` mirrored/enforced against the CDK definitions by a test (the model-allowlist TS↔Python mirror test is prior art in this repo).
2. Replace the hardcoded literals in the handlers with the shared constants.
3. A guard test that parses `core-stack.ts` index definitions and asserts every constant matches (so drift fails CI rather than production).

## Assessment

Priority: low (GSI renames are rare and destructive ops anyway); effort: small-medium, mostly mechanical. Value is drift-proofing, same class as the existing TS↔Python model-allowlist lockstep tests.

perrozzi:
## Summary

Closes #213.

DynamoDB GSI names were hardcoded as string literals at ~24 call sites across the Python API handlers and the TypeScript stream Lambda. Issue #212 demonstrated the failure mode: `data_explorer_handler.py` queried a `feedback-id-index` that does not exist in the stack (the real GSI is `gsi4-by-feedback-id`), and nothing caught it until runtime.

This PR centralizes every GSI name behind constants that are **guard-tested against the CDK stack definition**, so a rename or typo on either side fails CI instead of failing in production.

### Python — `lambda/shared/indexes.py` (new)

Seven constants mirroring the seven GSIs defined in `lib/stacks/core-stack.ts`, plus an `ALL_INDEX_NAMES` frozenset:

| Constant | Literal | Table |
|---|---|---|
| `FEEDBACK_BY_DATE_INDEX` | `gsi1-by-date` | feedback |
| `FEEDBACK_BY_CATEGORY_INDEX` | `gsi2-by-category` | feedback |
| `FEEDBACK_BY_URGENCY_INDEX` | `gsi3-by-urgency` | feedback |
| `FEEDBACK_BY_ID_INDEX` | `gsi4-by-feedback-id` | feedback |
| `AGGREGATES_BY_METRIC_TYPE_INDEX` | `gsi1-by-metric-type` | aggregates |
| `PROJECTS_BY_TYPE_INDEX` | `gsi1-by-type` | projects |
| `JOBS_BY_STATUS_INDEX` | `gsi1-by-status` | jobs |

Literals replaced in: `metrics_handler.py` (12 sites), `chat_handler.py`, `mcp_handler.py`, `projects.py`, `shared/feedback.py`, and `data_explorer_handler.py` (the local `FEEDBACK_ID_INDEX` alias introduced by #212's fix is removed in favor of the shared constant).

### TypeScript — `lambda/stream/src/indexes.ts` (new)

`FEEDBACK_BY_DATE_INDEX` and `FEEDBACK_BY_ID_INDEX` for the stream Lambda's three query sites (`tools/search-feedback.ts` x2, `context/project-context.ts` x1). Kept deliberately minimal: only the names the stream Lambda actually uses.

### Drift guards (mirror-test precedent from the model allowlist)

- **`lambda/shared/test/test_indexes.py`** — three tests:
  1. Parses the `addGlobalSecondaryIndex` blocks in `core-stack.ts` and asserts **set equality** with `ALL_INDEX_NAMES` (a GSI added to the stack without a constant fails, and vice versa).
  2. Asserts the TS `indexes.ts` names are a subset of the stack's names.
  3. Scans `lambda/**/*.py` for reintroduced hardcoded `gsi*`/`*-index` string literals outside the constants module.
- **`lambda/stream/src/indexes.test.ts`** — asserts the TS constants match their expected literals and are a subset of the stack's GSI names (parsed from `core-stack.ts`).

## Behavior

Unchanged. Every constant resolves to the exact literal previously in place — this is a mechanical refactor plus guard tests. No API responses, queries, or stack resources change.

## Testing

- Backend: `pytest` full suite — **922/922 passed**
- Stream: `vitest` — **3/3 passed** (new `indexes.test.ts`)
- Stream: `tsc --noEmit` clean
- Lint: `ruff check` clean

## Deployment

No infrastructure change (no CDK diff — names are identical). Lambda code-only change, rides along with the pending batch deploy of the dev branch. Per the agreed workflow, deploy + live validation are cumulated into the batch before the next dev→main release.

### DIFF
--- a/voc-datalake/lambda/api/chat_handler.py
+++ b/voc-datalake/lambda/api/chat_handler.py
@@ -19,6 +19,7 @@ from shared.api import (
 )
 from shared.feedback import basis_date, window_cutoff
 from shared.exceptions import ConfigurationError, NotFoundError as AppNotFoundError
+from shared.indexes import FEEDBACK_BY_DATE_INDEX
 
 from aws_lambda_powertools.event_handler.exceptions import NotFoundError
 from boto3.dynamodb.conditions import Key
@@ -104,7 +105,7 @@ def chat():
     for i in range(min(days, 7)):
         date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
         response = feedback_table.query(
-            IndexName='gsi1-by-date',
+            IndexName=FEEDBACK_BY_DATE_INDEX,
             KeyConditionExpression=Key('gsi1pk').eq(f'DATE#{date}'),
             Limit=10,
             ScanIndexForward=False
--- a/voc-datalake/lambda/api/data_explorer_handler.py
+++ b/voc-datalake/lambda/api/data_explorer_handler.py
@@ -22,6 +22,7 @@ from shared.logging import logger, tracer
 from shared.aws import get_s3_client, get_dynamodb_resource, get_sqs_client
 from shared.api import create_api_resolver, api_handler, DecimalEncoder
 from shared.exceptions import ConfigurationError, ValidationError, NotFoundError, ServiceError
+from shared.indexes import FEEDBACK_BY_ID_INDEX
 
 s3_client = get_s3_client()
 dynamodb = get_dynamodb_resource()
@@ -29,9 +30,6 @@ sqs_client = get_sqs_client()
 
 RAW_DATA_BUCKET = os.environ.get("RAW_DATA_BUCKET", "")
 FEEDBACK_TABLE = os.environ.get("FEEDBACK_TABLE", "")
-# Feedback-id lookup GSI — must match core-stack.ts (gsi4-by-feedback-id).
-# Regression-tested in test_data_explorer_handler.py (#140).
-FEEDBACK_ID_INDEX = 'gsi4-by-feedback-id'
 PROCESSING_QUEUE_URL = os.environ.get("PROCESSING_QUEUE_URL", "")
 
 # Available buckets for browsing
@@ -352,7 +350,7 @@ def save_feedback():
         else:
             # Need to find the item first
             response = table.query(
-                IndexName=FEEDBACK_ID_INDEX,
+                IndexName=FEEDBACK_BY_ID_INDEX,
                 KeyConditionExpression='feedback_id = :fid',
                 ExpressionAttributeValues={':fid': feedback_id},
                 Limit=1
@@ -415,7 +413,7 @@ def delete_feedback():
         
         # Find the item first using GSI
         response = table.query(
-            IndexName=FEEDBACK_ID_INDEX,
+            IndexName=FEEDBACK_BY_ID_INDEX,
             KeyConditionExpression='feedback_id = :fid',
             ExpressionAttributeValues={':fid': feedback_id},
             Limit=1
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -22,6 +22,7 @@ from shared.api import DecimalEncoder, validate_date_basis
 from shared.feedback import query_feedback_by_date
 from shared.tokens import hash_token
 from shared.tables import get_projects_table, get_feedback_table, get_aggregates_table
+from shared.indexes import FEEDBACK_BY_ID_INDEX
 from projects import autoseed_project
 
 logger = Logger()
@@ -442,7 +443,7 @@ def _tool_get_feedback_detail(args: dict, _token_info: dict) -> list[dict]:
         return [{"type": "text", "text": "Feedback table not configured"}]
 
     response = feedback_table.query(
-        IndexName='gsi4-by-feedback-id',
+        IndexName=FEEDBACK_BY_ID_INDEX,
         KeyConditionExpression=Key('feedback_id').eq(feedback_id),
         Limit=1,
     )
--- a/voc-datalake/lambda/api/metrics_handler.py
+++ b/voc-datalake/lambda/api/metrics_handler.py
@@ -19,6 +19,7 @@ from shared.api import (
     get_configured_categories, api_handler, DEFAULT_CATEGORIES
 )
 from shared.feedback import basis_date, window_cutoff
+from shared.indexes import AGGREGATES_BY_METRIC_TYPE_INDEX, FEEDBACK_BY_CATEGORY_INDEX, FEEDBACK_BY_DATE_INDEX, FEEDBACK_BY_ID_INDEX, FEEDBACK_BY_URGENCY_INDEX
 
 # Pagination bounds for /feedback. The candidate window is a function of
 # offset+limit, capped to prevent unbounded DynamoDB scans. The cap also defines
@@ -144,7 +145,7 @@ def _scan_recent_items(
         date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
         remaining = soft_cap - len(items)
         day_items, day_has_more = _query_partition(
-            'gsi1-by-date',
+            FEEDBACK_BY_DATE_INDEX,
             Key('gsi1pk').eq(f'DATE#{date}'),
             max_matched=min(per_day_limit, remaining) if per_day_limit else remaining,
             source=source,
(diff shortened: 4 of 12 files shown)


perrozzi:
This PR replaces ~24 hardcoded DynamoDB GSI name literals in the Python API handlers and the TypeScript stream Lambda with constants in `lambda/shared/indexes.py` and `lambda/stream/src/indexes.ts`, backed by guard tests that parse `core-stack.ts` and assert the mirrors stay in sync. It's a clean mechanical refactor following the existing model-allowlist mirror-test precedent, with no behavior change. Overall this looks good — no blockers.

### Should fix

- **`lambda/shared/test/test_indexes.py` — the literal-scan regex doesn't cover the actual #140/#212 failure mode.** `r"['\"](gsi\d+-by-[a-z-]+)['\"]"` only matches names that already follow the `gsiN-by-*` convention. The original bug was a query against `feedback-id-index` — a name that does *not* match this pattern and would sail through the scan. Consider also matching against `IndexName=` call sites (e.g. flag any string literal passed as `IndexName`) rather than only the naming pattern, or at least add `-index` suffixed literals to the pattern.
- **`test_no_new_hardcoded_index_literals_in_python` reads each file repeatedly.** `py.read_text()` is called once in `re.finditer(...)` and again *per match* to compute the line number. Read once into a variable. Minor today, but it's O(files × matches) I/O and easy to fix.
- **`lambda/stream/src/indexes.test.ts` uses `__dirname`.** The stream Lambda's imports (`./indexes.js`, `../lib/errors.js`) indicate ESM/NodeNext, where `__dirname` is not defined at runtime. If it passes today it's due to vitest's transform behavior, which is fragile across config changes; prefer `new URL('../../../lib/stacks/core-stack.ts', import.meta.url)` / `fileURLToPath`. Please confirm the reported 3/3 pass was against the actual test config and not a CJS fallback.

### Nits

- `voc-datalake/lambda/shared/feedback.py`: the new `from shared.indexes import ...` is inserted between stdlib imports (`import re` / `from datetime import ...`) — move it below the stdlib block per import ordering.
- `voc-datalake/lambda/api/metrics_handler.py`: the five-name import on a single long line would read better parenthesized/multi-line, matching the surrounding import style.
- The Python set-equality guard parses only `lib/stacks/core-stack.ts`; if a GSI is ever added in another stack file, the parser silently misses it. A one-line comment acknowledging this assumption (or globbing `lib/stacks/*.ts`) would harden it.
- Both parsers assume single-quoted `indexName: '...'` in the CDK source. The `assert stack` non-empty check mitigates a quote-style change on the Python side; the vitest `>= 7` check does the same on the TS side — good, just noting the coupling.
- `shared/indexes.py` docstring says the guard test lives in `lambda/shared/test/test_indexes.py` and references both #213 and #140 — nice traceability; keep the `JOBS_BY_STATUS_INDEX` "listed for stack parity" comment up to date if a consumer appears.

The removal of the local `FEEDBACK_ID_INDEX` alias in `data_explorer_handler.py` in favor of the shared constant is correct, and keeping the TS mirror minimal (subset guard) while the Python mirror enforces full set equality is a sensible split.

---

perrozzi:
Thanks for the review — all three should-fix items addressed in `54a81dd` (pushing now):

**Literal-scan regex missed the #140 shape — fixed.** You're right that `feedback-id-index` matches no `gsiN-by-*` convention and would sail through. The scan now runs two patterns: the convention pattern (catches GSI-looking names anywhere) plus a new kwarg pattern flagging **any** string literal passed to `IndexName=`/`index_name=` — which is exactly the shape the #140 bug had. Verified the kwarg pattern doesn't false-positive on constant-passing sites (`IndexName=FEEDBACK_BY_ID_INDEX` has no quote after `=`).

**Repeated `read_text()` — fixed.** Source is read once per file into a local; line numbers computed from that.

**`__dirname` in ESM — fixed.** Switched to `fileURLToPath(new URL('../../../lib/stacks/core-stack.ts', import.meta.url))`. To answer the confirmation question: re-ran under the package's real vitest config after the change — 3/3 pass.

**Nits:** feedback.py import moved below the stdlib/third-party blocks; metrics_handler.py import parenthesized across lines; added a comment in `_stack_index_names()` documenting the core-stack.ts-only assumption (true today — all seven tables live there) with the glob-widening pointer. Left the quote-style coupling as-is per your note — the non-empty/`>= 7` assertions already guard it.

Gates re-run: backend pytest 922/922, stream vitest 3/3, `tsc --noEmit` clean, ruff clean.

perrozzi:
The push is a targeted cleanup responding to the previous review: the literal-scan test now also flags any string literal passed to `IndexName=`/`index_name=`, the file is read once per scan, the vitest guard uses `import.meta.url` instead of `__dirname`, the `feedback.py` import ordering and `metrics_handler.py` import formatting were fixed, and the single-stack-file parsing assumption is documented. All previous should-fix items and nits are addressed.

### Resolved
- `test_indexes.py` literal scan now catches the actual #140 shape (`IndexName='feedback-id-index'`) via the new `_INDEX_KWARG_LITERAL` pattern.
- File I/O in `test_no_new_hardcoded_index_literals_in_python` reduced to a single `read_text()` per file.
- `indexes.test.ts` switched from `__dirname` to `fileURLToPath(new URL(...))`, correct for the package's ESM/NodeNext setup.
- `shared/feedback.py` import ordering fixed (stdlib / third-party / first-party blocks).
- `metrics_handler.py` five-name import parenthesized/multi-line.
- Single-stack-file parsing assumption in `_stack_index_names()` documented with guidance to glob `lib/stacks/*.ts` if it changes.

### New findings
- Nit only: in `test_no_new_hardcoded_index_literals_in_python`, a literal like `IndexName='gsi1-by-date'` matches both `_GSI_CONVENTION` and `_INDEX_KWARG_LITERAL`, so it would be reported twice in `offenders`. Harmless for the assertion (any offender fails), just slightly noisy output — dedupe by `(file, line)` if you care.

No blockers. The delta is clean and the guard tests are now meaningfully stronger; good to merge.

---

perrozzi:
Review converged in 2 rounds — round 2 reports all previous items resolved and no blockers.

On the remaining nit (duplicate offender entries when a literal matches both scan patterns): leaving as-is deliberately. The assertion fails on any offender regardless, the duplicate only affects failure-output verbosity, and a dedupe push would spend a bot round on a cosmetic change. Happy to fold it into a future touch of that test.

**Ready to merge.** Summary of what lands:
- `lambda/shared/indexes.py` — 7 GSI constants + `ALL_INDEX_NAMES`, replacing ~21 Python literals across 6 files
- `lambda/stream/src/indexes.ts` — 2 constants replacing 3 TS literals
- Drift guards on both sides: set-equality parse of `core-stack.ts`, subset check for the stream mirror, and a literal scan that catches both `gsiN-by-*` convention names and any raw string passed to `IndexName=`/`index_name=` (the exact #140 failure shape)

Gates: backend pytest 922/922, stream vitest 3/3, `tsc --noEmit` clean, ruff clean. Zero behavior change — every constant resolves to the literal previously in place, no CDK diff.

Per the agreed workflow, deploy + live validation are cumulated into the batch before the next dev→main release.

Issue #98, opened 2026-05-09
Project Chat returns no feedback: `Key('gsi1pk').eq('DATE')` should be `'DATE#YYYY-MM-DD'`


WPrintz:
## Summary

`get_feedback_for_chat()` queries the `gsi1-by-date` GSI with `KeyConditionExpression=Key('gsi1pk').eq('DATE')`, but every feedback row is written with `gsi1pk = "DATE#YYYY-MM-DD"` (set in `voc-datalake/lambda/processor/handler.py`). The query returns zero rows for every caller, regardless of input filters. The function is wrapped in a broad `except Exception: return []`, so the failure is silent — the Project Chat tab simply tells users there is no feedback in scope.

## Affected file

[`voc-datalake/lambda/shared/project_chat.py`](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/shared/project_chat.py#L100-L131) — `get_feedback_for_chat()`

```python
# line 106 (current main @ 2f744c2)
response = feedback_table.query(
    IndexName='gsi1-by-date',
    KeyConditionExpression=Key('gsi1pk').eq('DATE'),   # ← always returns []
    ScanIndexForward=False,
    Limit=limit * 2
)
```

Compare with the writer in `voc-datalake/lambda/processor/handler.py` (line 510 on `2f744c2`):

```python
'gsi1pk': f"DATE#{date_str}",
```

…and the correct query pattern used elsewhere, e.g. [`voc-datalake/lambda/api/metrics_handler.py:67`](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/2f744c2/voc-datalake/lambda/api/metrics_handler.py#L67):

```python
KeyConditionExpression=Key('gsi1pk').eq(f'DATE#{date}')
```

## Impact

- `voc-projects-api` Lambda (the only consumer of `shared/project_chat.py`'s `get_feedback_for_chat`) — Project Chat returns zero feedback context to the LLM. Claude responds based only on the metrics summary in the system prompt, never the actual reviews.
- The bug is **silent**: the surrounding `try/except` swallows any DynamoDB error and returns `[]`, so users see "no feedback to discuss" rather than an error.

## Reproduction

1. Deploy `main` to a fresh account.
2. Import any feedback (CSV upload, scraper, S3 import — all work).
3. Open a Project, switch to its Chat tab, and ask a question that should reference the imported feedback.
[...]

PR #221, opened 2026-07-17
fix(stream): query per-day date partitions for recent-feedback context (#220)

### LINKED ISSUE #220: [Bug]: stream project-context queries gsi1-by-date with bare 'DATE' partition key — recent-feedback context always empty
## Summary

`fetchRecentFeedback()` in the streaming Lambda's project-context builder queries the `gsi1-by-date` index with a partition key that can never match, so the "recent feedback" section of project chat / roundtable context is silently always empty.

## Where

`voc-datalake/lambda/stream/src/context/project-context.ts`, `fetchRecentFeedback()` (~line 220):

```ts
new QueryCommand({
  TableName: feedbackTable,
  IndexName: FEEDBACK_BY_DATE_INDEX,          // gsi1-by-date
  KeyConditionExpression: 'gsi1pk = :pk',
  ExpressionAttributeValues: { ':pk': 'DATE' },   // <-- bare 'DATE'
  ScanIndexForward: false,
  Limit: 30,
})
```

## Why it matches nothing

The processor writes per-day partition keys:

```python
# lambda/processor/handler.py:515
'gsi1pk': f"DATE#{date_str}",   # e.g. DATE#2026-07-17
```

An exact-equality condition against the bare string `'DATE'` therefore returns zero items, and the function's empty-result branch (`return { count: 0, promptSection: '' }`) makes the failure invisible — project chat and roundtable prompts just never include recent feedback context. No error, no log signal.

## The correct pattern is already in the same Lambda

`lambda/stream/src/tools/search-feedback.ts` (~line 186) queries the same index correctly, one day-partition at a time:

```ts
ExpressionAttributeValues: { ':pk': `DATE#${dateStr}` },
```

The fix is to mirror that: loop backward over the last N days (like the Python `shared/feedback.py` helpers and search-feedback.ts do) until ~30 items are collected or the lookback window is exhausted.

## Relationship to prior issues

- Same failure shape as #98 (bare/mis-shaped partition key against `gsi1-by-date`), but on the TypeScript side.
- Same "query an index in a way that silently returns nothing" family as #140/#212.
- Discovered during the #213 GSI-constants refactor (PR #219). The #213 guard tests verify index **names** match the stack; they intentionally don't validate key **shapes**, which is why this survives them.

## Impact

Project chat (`/projects/:id` chat tab) and roundtable responses are generated without the recent-feedback grounding they were designed to include — degraded answer quality, no user-visible error.

perrozzi:
## Summary

Closes #220.

`fetchRecentFeedback()` in the streaming Lambda's project-context builder queried the `gsi1-by-date` index with an exact-match condition on the bare literal `'DATE'`, while the processor writes per-day partition keys of the form `DATE#YYYY-MM-DD` (`processor/handler.py:515`). The query therefore **always returned zero items**, and because the empty-result path is handled gracefully, project chat and roundtable prompts silently never included the recent-feedback grounding they were designed to carry — no error, no log signal, just degraded answer quality.

Same failure shape as #98 (which was resolved by the Python module's removal); this was the surviving TypeScript occurrence, found during the #213/#219 GSI-constants refactor.

## What changed

`voc-datalake/lambda/stream/src/context/project-context.ts` — `fetchRecentFeedback()` now mirrors the established per-day pattern from `tools/search-feedback.ts`:

- Walks backward from today's UTC date, one `DATE#YYYY-MM-DD` partition per query, up to a 30-day lookback (`RECENT_FEEDBACK_LOOKBACK_DAYS`, matching search-feedback's day cap)
- Stops as soon as 30 items are collected (`RECENT_FEEDBACK_TARGET`, the same count the old single query intended); each day query passes `Limit: remaining` so it never over-fetches
- A failing day query keeps what's already collected and moves to the previous day (was: one try/catch around everything, any failure discarded all)
- Per-row `safeParse` skips a malformed item instead of throwing away the whole result (was: strict `.parse` inside the map — one bad row nuked the section; same rationale as the `fetchDayPages` comment in search-feedback.ts)
- Prompt rendering unchanged: first 15 items formatted into the `## Recent Customer Feedback` section, `feedback_count` = items collected

## Why not share code with search-feedback.ts

The two walkers have different semantics: search-feedback pages exhaustively through each day via `LastEvaluatedKey` (it needs the full candidate set for aggregate stats, `MAX_CANDIDATES` 10000), while the context builder needs at most 30 items and never pages. Forcing one helper over both would mean parameterizing paging behavior for two call sites — kept them separate deliberately.

## Regression tests

5 new tests in `project-context.test.ts` (through the public `buildProjectChatContext` entry point):

1. **Every feedback query uses a `DATE#YYYY-MM-DD` partition key, never bare `'DATE'`** — inspects the recorded `QueryCommand` inputs; fails if the fix is reverted
2. Includes the recent-feedback section (and correct `feedback_count`) when a recent day has items
3. Walks backward across days and stops at the 30-item target (asserts exactly project + 2 day queries, descending distinct dates)
4. Continues to the previous day when one day query rejects
5. Skips a malformed row (wrong field type) without discarding the day

## Testing

- Stream: `vitest` — **242/242 passed** (18 in project-context.test.ts: 13 existing + 5 new)
- `tsc --noEmit` clean
- `eslint` 0 errors (the one remaining warning is pre-existing on `development`)

## Deployment

Stream Lambda code only — no CDK/IAM change. Per the agreed workflow, deploy + live validation ride the next batch (the just-completed batch deploy took `development` to `ca0b342`; this lands after it).

### DIFF
--- a/voc-datalake/lambda/stream/src/context/project-context.test.ts
+++ b/voc-datalake/lambda/stream/src/context/project-context.test.ts
@@ -254,3 +254,131 @@ describe('buildProjectChatContext', () => {
     expect(ctx.metadata.selected_personas).toStrictEqual(['No Avatar Persona']);
   });
 });
+
+describe('fetchRecentFeedback via buildProjectChatContext (regression #220)', () => {
+  beforeEach(() => vi.clearAllMocks());
+
+  const todayUtc = new Date().toISOString().slice(0, 10);
+
+  function makeFeedback(overrides: Record<string, unknown> = {}) {
+    return {
+      source_platform: 'webscraper',
+      sentiment_label: 'negative',
+      category: 'delivery',
+      original_text: 'Package arrived late',
+      ...overrides,
+    };
+  }
+
+  /** Extract the gsi1pk value from a recorded QueryCommand call. */
+  function pkOfCall(call: unknown[]): string {
+    const cmd = call[0] as { input: { ExpressionAttributeValues?: Record<string, unknown> } };
+    const pk = cmd.input.ExpressionAttributeValues?.[':pk'];
+    return typeof pk === 'string' ? pk : '';
+  }
+
+  it('queries per-day DATE#YYYY-MM-DD partitions, never the bare DATE literal', async () => {
+    const docClient = createMockDocClient([
+      [projectMeta],
+      [makeFeedback()],
+    ]);
+
+    await buildProjectChatContext(
+      docClient, 'projects-table', 'feedback-table', 'proj-1', 'hello',
+    );
+
+    const sendMock = (docClient.send as ReturnType<typeof vi.fn>);
+    const feedbackCalls = sendMock.mock.calls.slice(1); // call 0 = project query
+    expect(feedbackCalls.length).toBeGreaterThan(0);
+    for (const call of feedbackCalls) {
+      const pk = pkOfCall(call);
+      expect(pk).toMatch(/^DATE#\d{4}-\d{2}-\d{2}$/);
+      expect(pk).not.toBe('DATE');
+    }
+    // The walk starts at today's UTC partition.
+    expect(pkOfCall(feedbackCalls[0])).toBe(`DATE#${todayUtc}`);
+  });
+
+  it('includes the recent-feedback section when a recent day has items', async () => {
+    const docClient = createMockDocClient([
+      [projectMeta],
+      [makeFeedback(), makeFeedback({ original_text: 'Love the new feature' })],
+    ]);
+
+    const ctx = await buildProjectChatContext(
+      docClient, 'projects-table', 'feedback-table', 'proj-1', 'hello',
+    );
+
+    expect(ctx.systemPrompt).toContain('Recent Customer Feedback');
+    expect(ctx.systemPrompt).toContain('Package arrived late');
+    expect(ctx.metadata.context).toStrictEqual(
+      expect.objectContaining({ feedback_count: 2 }),
+    );
+  });
+
+  it('walks backward across days and stops once the target is collected', async () => {
+    const day0 = Array.from({ length: 20 }, (_, i) => makeFeedback({ original_text: `day0 item ${i}` }));
+    const day1 = Array.from({ length: 20 }, (_, i) => makeFeedback({ original_text: `day1 item ${i}` }));
+    const docClient = createMockDocClient([
+      [projectMeta],
+      day0,
+      day1,
+    ]);
+
+    const ctx = await buildProjectChatContext(
+      docClient, 'projects-table', 'feedback-table', 'proj-1', 'hello',
+    );
+
+    // 20 from day 0 + 10 from day 1 = the 30-item target; the walk stops there
+    // (1 project call + 2 day calls, not the full 30-day lookback).
+    expect(ctx.metadata.context).toStrictEqual(
+      expect.objectContaining({ feedback_count: 30 }),
+    );
+    const sendMock = (docClient.send as ReturnType<typeof vi.fn>);
+    expect(sendMock.mock.calls).toHaveLength(3);
+    expect(pkOfCall(sendMock.mock.calls[1])).toBe(`DATE#${todayUtc}`);
+    expect(pkOfCall(sendMock.mock.calls[2])).toMatch(/^DATE#\d{4}-\d{2}-\d{2}$/);
+    expect(pkOfCall(sendMock.mock.calls[2])).not.toBe(pkOfCall(sendMock.mock.calls[1]));
+  });
+
+  it('continues to the previous day when one day query fails', async () => {
+    let callIndex = 0;
+    const docClient = {
+      send: vi.fn().mockImplementation(() => {
+        callIndex++;
+        if (callIndex === 1) return Promise.resolve({ Items: [projectMeta] });
+        if (callIndex === 2) return Promise.reject(new Error('throttled'));
+        if (callIndex === 3) return Promise.resolve({ Items: [makeFeedback()] });
+        return Promise.resolve({ Items: [] });
+      }),
+    } as unknown as import('@aws-sdk/lib-dynamodb').DynamoDBDocumentClient;
+
+    const ctx = await buildProjectChatContext(
+      docClient, 'projects-table', 'feedback-table', 'proj-1', 'hello',
+    );
+
+    expect(ctx.metadata.context).toStrictEqual(
+      expect.objectContaining({ feedback_count: 1 }),
+    );
+    expect(ctx.systemPrompt).toContain('Recent Customer Feedback');
+  });
+
+  it('skips a malformed row without discarding the rest of the day', async () => {
+    const docClient = createMockDocClient([
+      [projectMeta],
+      [
+        { ...makeFeedback(), original_text: 12345 }, // wrong type → safeParse fails
+        makeFeedback({ original_text: 'valid row survives' }),
+      ],
+    ]);
+
+    const ctx = await buildProjectChatContext(
+      docClient, 'projects-table', 'feedback-table', 'proj-1', 'hello',
+    );
+
+    expect(ctx.metadata.context).toStrictEqual(
+      expect.objectContaining({ feedback_count: 1 }),
+    );
+    expect(ctx.systemPrompt).toContain('valid row survives');
+  });
+});
--- a/voc-datalake/lambda/stream/src/context/project-context.test.ts
+++ b/voc-datalake/lambda/stream/src/context/project-context.test.ts
@@ -282,7 +282,7 @@ describe('fetchRecentFeedback via buildProjectChatContext (regression #220)', ()
   });
 
   // One parallel batch of day queries (mirrors DAY_QUERY_BATCH_SIZE in
-  // project-context.ts); +1 for the initial project query.
+  // recent-feedback.ts); +1 for the initial project query.
   const BATCH = 7;
 
   function makeFeedback(overrides: Record<string, unknown> = {}) {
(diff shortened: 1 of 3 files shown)


perrozzi:
This PR fixes a real bug: `fetchRecentFeedback()` queried the `gsi1-by-date` GSI with the bare literal `'DATE'` while the processor writes `DATE#YYYY-MM-DD` partition keys, so the recent-feedback prompt section was always silently empty. The fix follows the established per-day backward-walk pattern from `tools/search-feedback.ts`, adds per-row `safeParse` resilience, and includes a proper regression test that fails if the fix is reverted. Overall a clean, well-tested change.

### Should fix

- **Silent per-day `catch {}` re-creates the failure mode this PR fixes** (`voc-datalake/lambda/stream/src/context/project-context.ts`). The original bug's worst property was "no error, no log signal, just degraded answers." The new empty catch means a persistent error (missing IAM permission, wrong table/index name) would fail all 30 day-queries silently and produce the same invisible empty section. Add at least a `console.warn`/logger call in the catch (and consider breaking out of the loop on non-transient error types) so the next regression of this shape is observable.
- **Worst-case latency: up to 30 sequential DynamoDB queries per chat request.** When the table is sparse or empty (new projects, quiet periods), the walk performs 30 serial round-trips inside the streaming Lambda's context build, before the first token is produced. `search-feedback.ts` pays this cost only for explicit tool calls; this runs on every project chat/roundtable. Consider issuing the day queries in small parallel batches, or shortening the lookback when early days come back empty.

### Nits

- `Limit: remaining` counts items evaluated by DynamoDB, but malformed rows dropped by `safeParse` aren't replaced — a day with bad rows can under-fill the target even though more valid items exist in that partition (no `LastEvaluatedKey` follow-up). Acceptable for a best-effort prompt section, but worth a comment since `collectFeedbackRows` implies otherwise.
- `Array.from({ length: RECENT_FEEDBACK_LOOKBACK_DAYS }, (_, idx) => idx)` is an indirect way to write `for (let dayOffset = 0; dayOffset < RECENT_FEEDBACK_LOOKBACK_DAYS; dayOffset++)`.
- `todayUtc` in `project-context.test.ts` is computed at module load; a test run straddling UTC midnight would flake on the `DATE#${todayUtc}` assertions. Consider `vi.setSystemTime` to pin the clock.
- The `as unknown as import('@aws-sdk/lib-dynamodb').DynamoDBDocumentClient` cast in the failure-path test is at odds with the repo's prefer-type-guards convention; extending the existing `createMockDocClient` helper to support per-call rejection would avoid it.
- `feedback_count` semantics changed subtly (raw item count → validated item count). Fine, but if anything downstream compares this metric across deployments, the drop for tables with malformed rows is expected, not a regression.

---

perrozzi:
Thanks — both should-fix items and the nits are addressed in `6fb6123` (pushing now):

**Silent per-day catch — fixed.** Agreed this was re-creating the bug's worst property. Every failed day query now emits `console.warn` with the date and error name (console is this package's logging convention, cf. `web-search.ts`/`executor.ts`). Went one step further on your parenthetical: error names that would fail identically for every partition — `AccessDeniedException`, `ResourceNotFoundException`, `ValidationException` — stop the lookback immediately instead of warning 30 times. New test asserts both the warn and the early stop.

**Worst-case latency — fixed with batching.** Day queries now go out in parallel batches of 7 (`Promise.all`), so the sparse/empty-table worst case is ~5 sequential rounds instead of 30 serial round-trips, and the common case (today has data) remains a single round. Batch results are consumed in day order, so newest-first priority in the prompt is unchanged. Chose parallel batches over shortening the lookback: an empty week shouldn't hide feedback from two weeks ago.

**Structural note:** the walk moved to a new `context/recent-feedback.ts` module — the extra logic pushed `project-context.ts` over the repo's 400-line lint ceiling, and the batching/failure-handling deserved its own unit anyway. `project-context.ts` now just imports `fetchRecentFeedback`.

**Nits:** Limit/safeParse under-fill is now documented on `collectFeedbackRows` (deliberately best-effort, no `LastEvaluatedKey` follow-up for a prompt section); tests pin the clock via `vi.setSystemTime` (no UTC-midnight flake); `createMockDocClient` gained a `rejectAt` option so the failure-path test uses the shared helper instead of a one-off cast mock; the `Array.from` index generation stays — the repo lints `let` mutation (`no-restricted-syntax: Use const`), so the classic `for (let ...)` form isn't available here; `feedback_count` semantics (validated count) noted, agreed it's the right meaning going forward.

Gates re-run: stream vitest 243/243, `tsc --noEmit` clean, eslint 0 errors.

perrozzi:
Since the last review, `fetchRecentFeedback` was extracted to a new `recent-feedback.ts` with parallel day-query batching (7 at a time), failure logging via `console.warn`, and an early-exit on persistent error names; the tests were updated with a pinned fake clock, an extended mock helper, and two new failure-visibility tests. All Should-fix items and most nits from the previous review were addressed.

### Resolved
- Silent per-day `catch {}` — failures now logged via `console.warn`, with persistent errors (`AccessDeniedException`, etc.) short-circuiting the lookback, both covered by tests.
- 30 sequential round-trips worst case — replaced with parallel batches of 7 (≤5 sequential rounds on an empty table).
- `Limit`/malformed-row under-fill nit — documented in the `collectFeedbackRows` doc comment.
- `todayUtc` midnight flake — clock pinned with `vi.useFakeTimers`/`vi.setSystemTime`.
- `as unknown as DynamoDBDocumentClient` cast in the failure test — `createMockDocClient` extended with `rejectAt` instead.

### New findings
- **`console.warn` instead of the powertools logger** (`voc-datalake/lambda/stream/src/context/recent-feedback.ts`). The repo convention is powertools patterns; if the stream Lambda has a structured logger, use it so the warning carries correlation/request context and is queryable — a bare `console.warn` string is the weakest form of the observability this change is meant to add.
- Cost trade-off worth a comment: the common case (today's partition has ≥30 items) previously stopped after 1 query; it now always issues 7 parallel queries, each with `Limit: 30` (up to 210 items evaluated to keep 30). Acceptable for latency, but the inline comment claiming the common case "stays a single round" undersells the read amplification — spell it out so a future reader doesn't "optimize" it back to sequential.
- Stale comment in `project-context.test.ts`: `// mirrors DAY_QUERY_BATCH_SIZE in project-context.ts` — the constant lives in `recent-feedback.ts`. Similarly, `project-context.ts`'s "live there with their own tests" is slightly misleading since the tests remain in `project-context.test.ts`.

Otherwise the incremental changes are clean and well-tested; no blockers.

---

perrozzi:
Round 2 items handled in `489c8a7` (pushing now):

**`console.warn` vs powertools logger — keeping `console.warn`, with evidence.** The condition in your finding ("if the stream Lambda has a structured logger") is false: the stream package has no logger module (`src/lib/` contains only `errors.ts`/`streaming.ts`) and no powertools dependency in its `package.json` — powertools is the Python Lambdas' convention. `console.*` is this package's established mechanism (`web-search.ts`, `executor.ts`), and the Lambda runtime ships it to CloudWatch with the request ID prefix. Introducing a structured-logging dependency to the stream bundle is a package-wide decision that shouldn't ride a bug fix — happy to see it as a follow-up issue if maintainers want it.

**Read-amplification comment — fixed.** The batching comment now spells out the trade-off explicitly: one batch always issues 7 parallel queries with `Limit: 30` each (up to 210 items evaluated to keep 30), paying read amplification to bound latency, with a "don't optimize this back to sequential" warning for future readers.

**Stale comments — both fixed**: the test's `DAY_QUERY_BATCH_SIZE` reference now points at `recent-feedback.ts`, and `project-context.ts`'s extraction note now says the coverage lives in `project-context.test.ts` via `buildProjectChatContext`.

Gates re-run: stream vitest 243/243, `tsc --noEmit` clean, eslint 0 errors.

perrozzi:
Since the last review, only comments were touched: the stale `DAY_QUERY_BATCH_SIZE` reference in `project-context.test.ts` now points to `recent-feedback.ts`, the extraction note in `project-context.ts` correctly says the tests live in `project-context.test.ts`, and the batching comment in `recent-feedback.ts` now spells out the read-amplification trade-off with an explicit "don't optimize this back" warning. Two of the three previous findings are addressed; no code-behavior changes.

### Resolved
- Stale `// mirrors DAY_QUERY_BATCH_SIZE in project-context.ts` comment and the misleading "live there with their own tests" note — both corrected.
- Undersold read-amplification comment — the trade-off (7 parallel queries × `Limit: 30`) is now documented explicitly.

### Still open
- `console.warn` in `recent-feedback.ts` instead of the powertools structured logger — the failure-visibility path still emits an unstructured string with no correlation/request context.

### New findings
None — the incremental changes are comment-only and accurate.

---

perrozzi:
Review converged in 3 rounds: round 3 reports no new findings and comment-only deltas. The single "still open" item (`console.warn` vs a powertools structured logger) was explicitly declined with evidence in the round-2 disposition — the stream package has no logger module or powertools dependency; `console.*` is its established convention and adopting structured logging there is a package-wide decision for a follow-up, not a bug-fix rider. (The reviewer does not read PR comments, so it re-lists the item; maintainers can weigh the disposition.)

**Ready to merge.** What lands:
- `context/recent-feedback.ts` (new): per-day `DATE#YYYY-MM-DD` partition walk replacing the bare-`'DATE'` query that always returned zero items — project chat and roundtable prompts get their recent-feedback grounding back
- Parallel day-query batches (7 at a time) bound worst-case latency at ~5 rounds; failures are logged with date + error name, and persistent errors (`AccessDeniedException` etc.) stop the lookback instead of repeating 30x
- 5 regression tests through `buildProjectChatContext`, including the fail-on-revert key-shape assertion, pinned-clock determinism, and failure-visibility coverage

Gates: stream vitest 243/243, `tsc --noEmit` clean, eslint 0 errors.

Deploy note: stream Lambda code only, no CDK/IAM change. Rides the next batch deploy (the environment currently runs `ca0b342` from yesterday's batch).

PR #226, opened 2026-08-01
Upgrade Claude Opus 4.8 to Opus 5, plus dev and deploy-script fixes

perrozzi:
Upgrades the prototype surface from **Claude Opus 4.8 to Opus 5**, plus the fixes that surfaced while validating the deploy.

Deployed to a dev account and live-validated before opening. Scope note: this is not a platform-wide migration — one surface default moves and Opus 4.8 stays selectable.

## The model change

Opus 5 (`global.anthropic.claude-opus-5`) replaces Opus 4.8 as the prototype-surface default. **Opus 4.8 stays in the picker** rather than being removed: Opus 5's safety classifiers re-run declined higher-risk requests on 4.8, so it must remain granted or that fallback becomes an `AccessDenied` mid-request. The allowlist is now five models.

Two things were more than a rename:

- **Opus 4.8 was mis-classified.** It rejects a manual `thinking.budget_tokens` (true for Opus 4.7 and later) but sat only in the omit-temperature set. Making it selectable again would have 400'd any surface passing a thinking budget. Both Opus generations now carry `adaptive_thinking`.
- **`adaptive_thinking` is now a per-model data field** alongside `omit_temperature`, so both capability sets derive from the same rows. The previous hand-maintained set is what mis-classified 4.8 twice.

`nag-suppressions.ts` now derives its Bedrock ARNs from the allowlist instead of a hand-copied list that drifted on every model bump.

## `converse.py`: the empty-text retry

The largest single piece of work here, and the part that took five review rounds to settle. Worth reading as one story rather than per-commit.

An adaptive-thinking model can spend its whole `maxTokens` budget on thinking and return **zero visible text**. The continuation loop then replayed an empty assistant turn, which Converse rejects (`text content blocks must be non-empty`). This became far more likely with this PR, since Sonnet 5 and both Opus generations are all adaptive-thinking.

The fix re-runs single-turn with a raised ceiling. Getting that right took four passes, each one bounding the previous:

| Problem | Fix |
|---|---|
| `min(current * 2, CEILING)` is **not monotonic** — `build_prototype`'s 32000 was retried at 16384, i.e. *half* the headroom, making a second empty response more likely | `_raised_empty_budget() -> int \| None`, upward-only, `None` when there's no headroom |
| Ceiling 16384 sat **below** the largest caller, so the retry was inert on the prototype surface — the adaptive-thinking surface most likely to need it | Ceiling → **64000**, inside [Opus 5's 128K output limit](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html) |
| A bigger ceiling risks the timeout the small ceiling was guarding against (`DocumentGeneratorJob` runs 15 min; one huge `maxTokens` burns minutes on CJK output) | Wall-clock guard, default 420s, exposed as `empty_raise_deadline_seconds=` |
| 64000 is justified against Opus 5 but applied to whatever model **resolved** — a lower-capped model turned a harmless empty into a `ValidationException` | Absorb `ValidationException` on the retry, degrade to the empty result. Only that code; `AccessDenied` still propagates |

On the last one I chose the exception handler over a per-model cap table deliberately: `converse()` also takes an arbitrary `model_id=` override and honours a legacy `BEDROCK_MODEL_ID`, so a table can be consulted for a model it lacks — and the allowlist is mirrored in three lockstep places, so a new per-model field is maintenance in all three for a best-effort recovery path.

## Other bug fixes found along the way

- **`core-stack.ts`** — raw-data bucket CORS gains `PUT` so the browser presigned upload for project product docs works.
- **`update-env.sh`** — could not produce a working local dev environment. It queried two stacks removed by the stack merge, wrote `.env.production` (which the vite **dev** server never reads), and omitted `VITE_IDENTITY_POOL_ID`. That last one is the sharp edge: `RuntimeConfigSchema` requires a non-empty `identityPoolId`, and on failure the fallback **blanks every Cognito value** — so the login screen reads "Cognito not configured" even when the user pool and client id resolved correctly.
- **`deploy:*` scripts** — five of seven named stacks removed by the merge. Now `deploy:core`/`deploy:api`, with `deploy:frontend` delegating to the real S3-sync script (that name previously meant two different things depending on your working directory).
- **i18n** — `showAllCategories`/`showTop` shipped with a base form only despite being called with `{{count}}`. Completed the plural families for ja/zh/ko; total missing keys is now 0.
- **README** — the Quick Start opened with `git clone xxxxxx (repo url)`, and the `cd` on the next line used the wrong directory name.

## Tests

Every guard below was verified to fail when the thing it protects is reverted.

- `model-allowlist.test.ts` — IAM/agreement derivation and suppression coverage.
- `test_model_config.py` — pins `mock-server.js` to `model_config.py`. The mock was a third mirror of the allowlist with no test, which is exactly why it silently drifted.
- `deploy-scripts.test.ts` — every `cdk deploy <Stack>` target must resolve against `bin/voc-datalake.ts`, **and** `update-env.sh` must name live stacks, write `.env` not `.env.production`, and cover every `VITE_` var `runtimeConfig.ts` reads (both `getEnvString` and direct `import.meta.env` forms).
- `core-stack.test.ts` — raw-bucket CORS allows GET+PUT, keeps the localhost dev origins, and grants nothing beyond GET/PUT. That last assertion matters because the origin list carries a `*.cloudfront.net` wildcard, so the method list is what keeps its blast radius small.
- `test_converse.py` — the raise boundaries (upward-only, ceiling-relative fixtures, ceiling clears the largest known caller), the deadline as a pure predicate, and absorb-vs-propagate for the retry's `ClientError`.

## Verification

Local: **1173** Python, **84** CDK, **243** stream, **1804** frontend, `tsc` clean across all three projects, `ruff` clean.

Deployed to a dev account, all 6 stacks: `test-api.sh` **32/32**; `/settings/model` serves 5 models with the corrected capability flags; chat pinned to Opus 5 returned a real reply, confirming a deployed Lambda invokes Opus 5 with its own execution role; deployed UI shows the 5-model picker under real Cognito login.

The deployed validation predates the `converse.py` review rounds, which are unit-covered rather than re-deployed.

## Reviewer notes

- **The `https://*.cloudfront.net` CORS origin is broader than needed** and has been live since 2026-07-17. Presigned URLs remain the auth gate, so it isn't an open write path, but narrowing it to the real distribution domain (or scoping the `PUT` rule to the upload prefix) is worth a follow-up. Constrained for now by the method-minimality test above.
- **The deadline guard is inert on short-timeout callers.** No caller passes `empty_raise_deadline_seconds=` yet, so the 420s default only binds on the long-budget job Lambdas. Making it automatic means plumbing `get_remaining_time_in_millis()` through `converse()`, which never receives Lambda context — a separate change. Judged harmless because the deadline can only bind on calls slow enough to approach it.
- **`i18n:validate` still exits 1** on 40 pre-existing empty values in de/es/fr/pt — mostly `_many` variants needing real inflection, deliberately left for a translation pass rather than guessed at.
- **Local dev cannot use the deployed API**: `ALLOWED_ORIGIN` is the CloudFront domain, so CORS refuses `localhost`. Local work uses `npm run mock`. Documented in the steering docs here.
- `1f2e483`'s commit message wrongly says the CORS fix was never deployed. It had in fact been live since 2026-07-17 — I mis-traced it, and `LastUpdatedTime` plus the live bucket state proved otherwise. The code is right; only the message is wrong.


perrozzi:
This PR bumps the prototype surface to `global.anthropic.claude-opus-5`, keeps Opus 4.8 in the allowlist as a claimed safety-fallback target, converts `adaptive_thinking` into a per-model data flag (fixing Opus 4.8's mis-classification), derives cdk-nag Bedrock suppressions from the allowlist, and folds in several unrelated fixes (converse empty-text retry, raw-bucket CORS `PUT`, `update-env.sh`, `deploy:*` scripts, i18n plurals, README). The model-id/IAM/agreement lockstep work is genuinely good — all five mirrors (Python, stream TS, CDK allowlist, nag suppressions, mock server) are now pinned by tests, which is exactly the pattern this repo wants. Two things need attention before merge: the new `converse.py` retry can *lower* `maxTokens`, and two of the four bug fixes ship without regression tests despite the repo rule.

### Blockers

- **`voc-datalake/lambda/shared/converse.py` — the "raised ceiling" can shrink the budget.** `raised = min(kwargs['inferenceConfig']['maxTokens'] * 2, _EMPTY_RAISE_CEILING)` with `_EMPTY_RAISE_CEILING = 16384` means:
  - a caller at `maxTokens >= 16384` gets an identical retry (two wasted Bedrock calls, same failure);
  - a caller at `maxTokens > 16384` — plausible for the prototype builder and PRD/PR-FAQ generation, which are the adaptive-thinking surfaces most likely to burn the whole budget on thinking — gets `16384`, i.e. a **reduction**, making the empty-text outcome strictly more likely.

  Clamp upward only, e.g. `raised = max(current, min(current * 2, _EMPTY_RAISE_

> ⚠️ This review was cut short — the model hit its output token limit. The findings above stand, but coverage is incomplete.

---

perrozzi:
Both points actioned — head is now `b0c0292`. Thanks, the blocker was real and worse than the review had evidence for.

## Blocker: the raise could shrink the budget — fixed in `20eecd3`

Confirmed, and the "plausible for the prototype builder" guess was exactly right. `build_prototype` ([`document_generator/handler.py:469`](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/feat/claude-opus-5-migration/voc-datalake/lambda/jobs/document_generator/handler.py#L469)) passes `max_tokens=32000`, the largest budget in the repo, on the `prototype` surface — whose default is Opus 5, i.e. adaptive thinking. So the one caller above the ceiling is also the one most likely to burn its whole budget on thinking and reach this branch, and it was being retried at `16384`: **half** the headroom it asked for. The retry made a second empty response strictly more likely than the first.

Worth recording that this repo's own notes had this filed as an unreachable edge case, on the grounds that strict-JSON sites use 2048–4096 and the docgen chain steps top out at 8000. That survey missed the single-call prototype builder. The review was right and the note was wrong.

Fix follows your suggestion, with the wasted-call case folded in:

```python
def _raised_empty_budget(current_max: int) -> int | None:
    raised = min(current_max * 2, _EMPTY_RAISE_CEILING)
    return raised if raised > current_max else None
```

Returning `None` rather than clamping to `current` also covers your first bullet: a caller sitting exactly at the ceiling no longer issues a byte-identical retry. The loop breaks with a diagnostic instead of spending a call it cannot benefit from. The crash the branch exists to prevent — replaying an empty assistant turn, which Converse rejects — is still avoided in both cases; the caller gets `''`.

I kept the ceiling at 16384 rather than raising it. `converse.py`'s own tuning note warns that a single huge `maxTokens` makes one Bedrock call run many minutes on slow CJK output and can blow the Lambda timeout, so doubling 32000 to 64000 trades one failure mode for a worse one. Declining to retry is the coherent choice under that doctrine.

## Missing regression tests — fair, added in `3f0cb0a` and `b0c0292`

Accurate count: converse had two tests, deploy scripts had one, CORS and `update-env.sh` had none. Both untested ones are in the class where failure is an opaque runtime symptom rather than a stack trace, which is the class that most needs a guard.

- **Raw-bucket CORS** (`core-stack.test.ts`) — GET+PUT present, localhost dev origins retained, and nothing beyond GET/PUT. That last one is deliberate: the origin list carries a `https://*.cloudfront.net` wildcard, so the method list is the thing keeping its blast radius small.
- **`update-env.sh`** (`deploy-scripts.test.ts`) — this file already cited the script as evidence that stack-name rot recurs, then only checked `package.json`. It now also checks the script's own stack defaults, that it writes `.env` and not the `.env.production` that `vite dev` ignores, and that it covers every `VITE_` var `runtimeConfig.ts` reads. That last assertion is the one that would have caught the `VITE_IDENTITY_POOL_ID` omission.
- **Boundary case** — a caller below the ceiling raises until it reaches the ceiling and then stops, so `max_tokens=8192` makes 2 calls where the old expression made 3.

All six verified to fail against the pre-fix code, individually.

Two notes on writing the CORS one, in case they save someone time: `uniqueName()` builds `BucketName` from `Aws.ACCOUNT_ID`/`Aws.REGION`, so it synthesizes to an `Fn::Join` and selecting a bucket by name silently matches nothing — the test filters on logical id. And the CORS properties are parsed through a Zod schema rather than type-annotated, because `findResources` returns `any`-typed properties, where a CDK property rename would read `undefined` and pass every assertion vacuously.

## One request

The review stopped at its output-token limit right after the first blocker, so its coverage is incomplete — absence of further findings isn't evidence there are none. A re-review on `b0c0292` would be worth having before merge.

Gates on head: 1804 frontend, 1162 Python, 243 stream, 84 CDK (was 78), `tsc` clean across all three projects, `ruff` clean under the rule set `ruff.toml` documents.

Unchanged from the original body: the `*.cloudfront.net` origin is still worth narrowing (live since 2026-07-17, so hardening rather than a merge gate), and the 40 pre-existing i18n empties in de/es/fr/pt still need a real translation pass.

perrozzi:
Since the last review the author fixed the `converse.py` budget-shrinking blocker by extracting `_raised_empty_budget()` (upward-only clamp, returns `None` when there is no headroom) with a parametrised test plus two behavioural tests, and added the missing regression tests for the other two bug fixes: raw-bucket CORS (`core-stack.test.ts`) and `update-env.sh` (`deploy-scripts.test.ts`). Both previous concerns are addressed.

### Resolved
- `voc-datalake/lambda/shared/converse.py` — the empty-text retry can no longer lower `maxTokens`; callers at/above the ceiling now break instead of issuing a shrunken or byte-identical retry, and this is pinned by `test_raised_empty_budget_never_lowers_the_ceiling`.
- Regression tests now exist for the raw-data bucket CORS `PUT` fix and for the `update-env.sh` fix (dead stack names, missing `VITE_IDENTITY_POOL_ID`, `.env` vs `.env.production`).

### New findings
- **Should-fix — `voc-datalake/lambda/shared/converse.py`: the retry is now inert for the caller the docstring names.** `_EMPTY_RAISE_CEILING = 16384` while `build_prototype` calls at `max_tokens=32000`, so the prototype surface (Opus 5, adaptive thinking, the case cited as most likely to spend the whole budget on thinking) gets zero retries — `test_empty_max_tokens_result_does_not_retry_without_headroom` encodes exactly that. The crash is avoided, but the underlying empty-output failure is unmitigated on the highest-risk surface. Consider raising the ceiling above the largest caller budget (e.g. 64000, still within Opus limits) so the doubling actually applies there.
- **Nit — `voc-datalake/lib/stacks/core-stack.test.ts`:** the two doc comments contradict each other ("Identified by bucket name rather than logical id" vs "Matched on logical id, not BucketName"); the code does the latter. Also, `RawBucketSchema.parse` runs inside `.map()` before the `toHaveLength(1)` assertion, so a `RawDataBucket*` resource without CORS surfaces as a raw ZodError rather than the intended message — parse after filtering, or make `CorsConfiguration` optional and assert on the count first.
- **Nit — `voc-datalake/lib/utils/deploy-scripts.test.ts`:** the required-var list is derived solely from `getEnvString('VITE_…')`, so a var read via `import.meta.env.VITE_X` directly would be silently exempt from the check. `runtimeConfig.ts` isn't in this diff, so I can't confirm coverage is complete — worth a comment there noting new vars must go through `getEnvString` for the guard to see them.

---

perrozzi:
All three actioned — head is now `3b9ea50`.

## Should-fix: the retry was inert on the surface it exists for

You're right, and this is the more interesting half of the original bug. Fixing the direction of the clamp made it *safe*; it didn't make it *useful* for `build_prototype`. `test_empty_max_tokens_result_does_not_retry_without_headroom` encoding 32000 was the tell — I wrote a test asserting the highest-risk caller gets nothing and didn't read it as a symptom.

Ceiling is now **64000**, so 32000 doubles once. [Opus 5's max output is 128K](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html), so that's comfortably inside the model's limit.

The reason I'd left it at 16384 was real, though, and raising it alone would have traded truncation for a timeout: `document_generator` runs on a 15-minute Lambda, and `converse.py`'s own tuning note warns that one huge `maxTokens` can burn many minutes on slow CJK output. So the raise is now gated on a wall-clock deadline:

```python
_EMPTY_RAISE_DEADLINE_SECONDS = 420

elapsed = time.time() - start_time
if _empty_raise_past_deadline(elapsed):
    # The retry asks for double the budget with less time to spend it.
    break
```

Past the deadline it returns the empty result so the caller fails cleanly, because a timeout mid-retry returns nothing at all. That's the same guard shape as `repo-review`'s `CONTINUATION_DEADLINE_SECONDS` — which exists for the identical Opus 5 property, reasoning billed against `maxTokens`. Both codebases now converge on: moderate per-call budget, bounded retries, wall-clock stop.

Two test-quality notes from this round:

- The fixtures that hardcoded `16384`/`8192` are now derived from the ceiling, plus a new `test_raise_ceiling_clears_every_caller_budget_in_the_repo` that asserts the ceiling sits above the largest real caller. That guard is what would have caught this finding without a human reading the docstring against the constant.
- The deadline is extracted as a pure predicate rather than tested by patching the clock. `patch('shared.converse.time.time')` is not viable here: that attribute *is* the global `time` module, so patching it monkeypatches `time.time` process-wide and botocore's TLS clock breaks — it surfaced as `SystemTimeWarning: System time is way off` and SSL verification errors, not as an obviously wrong test. The end-to-end case moves the deadline instead of the clock.

Verified: 5 of the new assertions fail at the old ceiling, 3 fail with the deadline stubbed to `False`.

## Nit: contradictory comments + parse-before-count

Both correct, both mine. The "Identified by bucket name rather than logical id" line was a leftover from a first attempt that genuinely did match on name — it failed because `uniqueName()` builds `BucketName` from `Aws.ACCOUNT_ID`/`Aws.REGION`, so it synthesizes to an `Fn::Join`. I added the corrected comment and never deleted the wrong one. Removed.

Ordering fixed too — count first, parse second, so a `RawDataBucket*` that lost its CORS block reports the intended message instead of a raw `ZodError`.

## Nit: the guard could miss a direct `import.meta.env` read

Also correct. Rather than a comment in `runtimeConfig.ts` asking future readers to route through `getEnvString`, the guard now scans both patterns:

```typescript
...runtimeConfig.matchAll(/getEnvString\('(VITE_[A-Z_]+)'/g),
...runtimeConfig.matchAll(/import\.meta\.env\.(VITE_[A-Z_]+)/g),
```

A convention that only holds because someone read a comment is the kind that lapses silently; this way a direct read is caught by the check itself. For the record, `runtimeConfig.ts` currently routes everything through `getEnvString` (which reads `import.meta.env[key]` by bracket notation), so today's coverage was complete — the exposure was to a future edit, exactly as you described.

Gates on `3b9ea50`: 1170 Python, 84 CDK, 1804 frontend, 243 stream, `tsc` clean across all three projects, `ruff` clean.

One note on this review round itself: it's the second consecutive one where the finding was a *consequence of the previous fix* rather than something pre-existing, which is a good argument for the re-review loop. The truncation that cut round 1 short is separately fixed but lives in this reviewer's own repo, so it isn't visible from here.

perrozzi:
Since the last review the author addressed my ceiling concern by raising `_EMPTY_RAISE_CEILING` 16384 → 64000 (so the 32000-token prototype caller now gets one real raise), added a wall-clock guard `_empty_raise_past_deadline` / `_EMPTY_RAISE_DEADLINE_SECONDS = 420` with a pure-predicate test plus a monkeypatched behavioural test, and cleaned up both nits (contradictory comment + parse-after-filter in `core-stack.test.ts`, second `import.meta.env.VITE_X` pattern in `deploy-scripts.test.ts`). All three previous findings are resolved.

### Resolved
- `voc-datalake/lambda/shared/converse.py` — the retry is no longer inert for `build_prototype`; 32000 → 64000 is pinned by `test_empty_max_tokens_result_retries_the_prototype_budget` and `test_raise_ceiling_clears_every_caller_budget_in_the_repo`.
- `voc-datalake/lib/stacks/core-stack.test.ts` — contradictory doc comment removed; count asserted before `RawBucketSchema.parse`.
- `voc-datalake/lib/utils/deploy-scripts.test.ts` — direct `import.meta.env.VITE_*` reads are now picked up by the required-var derivation.

### New findings
- **Should-fix — `voc-datalake/lambda/shared/converse.py`: 64000 is justified against Opus 5's limit, but the raise applies to whatever model resolved.** The empty-text branch is reachable for any model that returns `max_tokens` with no visible text, and the doubled budget is not clamped to the resolved model's own max-output limit. Where 64000 exceeds that limit, the retry turns an empty-but-harmless result into a `ValidationException` — a different crash than the one the branch exists to prevent. Consider clamping the raise to a per-model output cap (the allowlist already carries per-model capability data), or treating a `ValidationException` on the retry as "return the empty result".
- **Nit — `voc-datalake/lambda/shared/converse.py`: the 420s deadline is calibrated only for the 15-minute job Lambda.** Callers on short-timeout functions (API handlers) can never reach 420s, so the guard gives them no protection at all — a doubled-budget retry there can still be killed mid-call. Deriving the budget from `context.get_remaining_time_in_millis()` (or accepting a deadline argument from the caller) would make the guard mean the same thing on every surface.
- **Nit — `voc-datalake/lambda/shared/test/test_converse.py`:** `test_raise_ceiling_clears_every_caller_budget_in_the_repo` hardcodes `largest_caller_budget = 32000` with a comment pointing at `document_generator/handler.py`, so it does not actually verify "every caller in the repo" — a future caller asking for >64000 won't trip it. Either grep the callers or rename the test to what it checks.

---

perrozzi:
All three actioned — head is now `7162038`.

## Should-fix: the raise wasn't bounded by the *resolved* model's limit

Correct, and it's the sharper version of the original bug: I justified 64000 against Opus 5 and then applied it to whatever the picker resolved. Where the resolved model caps output lower, the retry converts a harmless empty result into a `ValidationException` — a worse outcome than the one the branch exists to prevent.

I went with your second option rather than the per-model cap, deliberately:

```python
except ClientError as e:
    if e.response.get('Error', {}).get('Code') != 'ValidationException':
        raise
    # Degrade to the pre-retry outcome: this path exists to recover an
    # empty result, and turning that into a crash is strictly worse.
    break
```

Two reasons the cap table loses here. First, `converse()` also accepts an arbitrary `model_id=` override and honours a legacy `BEDROCK_MODEL_ID`, so a cap table can be consulted for a model that isn't in it — the failure mode being an unguarded 400, exactly what we're fixing. Second, the allowlist is mirrored in three places (`model_config.py`, `model-allowlist.ts`, `model-override.ts`) with lockstep tests; adding a per-model field means maintaining it in all three for a best-effort recovery path. The exception handler is correct for every model, including ones added later, with no new data to keep in sync.

Scoped tightly: only `ValidationException` is absorbed. `AccessDeniedException` and friends still propagate, since those are real misconfiguration rather than a budget that won't fit — pinned by `test_empty_max_tokens_raise_still_propagates_other_client_errors`.

## Nit: the deadline only bound on the 15-minute Lambda

Fair. It's now the default of a caller-facing `empty_raise_deadline_seconds` argument, documented as ideally derived from the handler's own `get_remaining_time_in_millis()`.

I stopped short of plumbing Lambda context through `converse()`, which is what would make the guard automatic — that touches every call site and belongs in its own change. Being straight about the current state: no caller passes the argument yet, so on a short-timeout function the guard still doesn't bind. My read is that this is harmless rather than wrong, because the deadline can only bind on calls slow enough to approach it, and those are the large-budget job callers; an API handler at `max_tokens=1500` retrying at 3000 can't blow a timeout that the first call survived. The reasoning is now in the constant's comment so the calibration isn't silent.

The parameter does earn its place immediately, though: the end-to-end deadline test passes it instead of monkeypatching module state, which is a strictly better test.

## Nit: the test name overclaimed

Guilty, and I'd flagged the same gap to my own reviewer before yours arrived. Renamed to `test_raise_ceiling_clears_the_largest_known_caller_budget`, with the scope limit and its reason in the docstring: the budget is hand-maintained rather than grepped because `shared/` is bundled into many Lambdas and sibling handler paths aren't reliably present at test time.

Worth noting the two findings interlock — a future caller above the ceiling is now caught by the `ValidationException` fallback rather than needing this test to know about it. That's a better division of labour than a grep would have been: the test pins the known case, the runtime handles the unknown one.

Gates on `7162038`: 1173 Python, 84 CDK, 1804 frontend, 243 stream, `tsc` clean across all three projects, `ruff` clean. The new `ValidationException` test fails with the absorption removed.

perrozzi:

Since the last review the author added a `ValidationException` fallback around the raised-budget retry (degrading to the empty result rather than crashing, with two new tests covering absorb vs. propagate), turned the deadline into a caller-facing `empty_raise_deadline_seconds=` parameter with documentation of why the default is inert on short-timeout callers, and renamed/rescoped `test_raise_ceiling_clears_the_largest_known_caller_budget`. All three previous findings are addressed.

### Resolved
- Model-cap concern: a `ValidationException` on the raise now degrades to the pre-retry empty result instead of turning a harmless empty into a crash, pinned by `test_empty_max_tokens_raise_rejected_by_model_returns_empty` and its AccessDenied counterpart.
- Deadline calibration: `empty_raise_deadline_seconds=` lets short-timeout callers pass their own budget, and the constant now documents itself as a default only.
- Test scope nit: renamed to "largest known caller" with an explicit note on what it does and does not catch.

### New findings
- **Nit — `voc-datalake/lambda/shared/converse.py`:** the skip-the-raise log still interpolates the module constant (`deadline {_EMPTY_RAISE_DEADLINE_SECONDS}s`) instead of the effective `empty_raise_deadline_seconds`, so a caller that overrides the deadline gets a log line claiming 420s while the decision used a different value — exactly the kind of mismatch that makes this branch hard to diagnose from CloudWatch. Use the parameter in the message.

Otherwise the incremental changes look clean; the absorb-only-`ValidationException` split is the right narrowness.

---

perrozzi:

The only change since my last review is the one-line log fix in `voc-datalake/lambda/shared/converse.py`, which now interpolates the effective `empty_raise_deadline_seconds` parameter instead of the module constant — exactly the nit I raised. All previous findings are addressed.

### Resolved
- Skip-the-raise warning log now reports the effective deadline, so an overriding caller's CloudWatch line matches the decision that was actually made.

No new findings; the incremental change is clean and needs no further follow-up.

---

PR #228, opened 2026-08-02
fix: config-driven AI budgets, avatar model EOL prep, ingestor sizing

perrozzi:
Four independent fixes from an audit of the prd-fix `LESSONS.en.md` backlog, plus two bugs that the audit work surfaced during deployment and review.

## 1. Research budgets came from a config nobody read

`lambda/api/prompts/research-analysis.json` declared `max_tokens: 9000` per step, but the async Step Functions research path hardcoded `4000/3000/3000` and kept its own copies of the system prompts. Only the sync `run_research` path ever read that JSON, so editing it silently did nothing for the async path and raised no error.

The JSON is live for the sync path, so deleting it was not an option. Instead both paths now read it:

- `shared/prompts.py` gains `get_step_inference_config()` / `get_research_step_config()`, returning a step's system prompt and budgets **without** building a user prompt. The async path needs that because its prompt injects personas, uploaded documents and web-search results, which the templated sync path has no placeholders for.
- `research_step_handler.py` resolves each step through one `_step_inference()` helper, replacing three inlined system prompts and three copies of the language-instruction block.

The three inline prompts were verified byte-identical to the JSON before removal, so **no prompt behaviour changes** — only budgets, and `thinking_budget` and `step_name` are now honoured rather than discarded.

## 2. Avatar image model was reaching EOL — now migrated

`amazon.nova-canvas-v1:0` entered LEGACY on 2026-03-30 with EOL on 2026-09-30. A legacy model also drops access for accounts idle 15+ days, which explains an earlier run of `ResourceNotFoundException`. Avatar generation degrades to `avatar_url=None` by design, so the model expiring produces **no user-visible error** — only a log line.

**Migrated to `stability.stable-image-core-v1:1` in us-west-2.** Enumerated against the account rather than guessed, which is what made this more than an id swap:

| region | active text-to-image generators |
|---|---|
| us-east-1 | **0** (only LEGACY Nova Canvas; every other image model there is a Stability *editing* primitive needing an input image) |
| us-west-2 | **3**: `stable-image-core-v1:1`, `stable-image-ultra-v1:1`, `sd3-5-large-v1:0` |
| eu-west-1 / ap-northeast-1 | 0 |

Scanned 16 regions; Titan Image Generator is no longer offered (`ResourceNotFoundException` on v1, v1:0 and v2:0 — only Titan *embedding* models remain), so there is no in-region alternative. All three candidates were probe-invoked from the account to confirm entitlement before choosing; Core is cheapest and fastest and is ample for a headshot.

Supporting work that made the region change cheap: the model id and region are single-sourced in `lib/utils/model-allowlist.ts`, three pasted IAM ARN literals in `api-stack.ts` now derive from `imageModelArn()`, and `avatar-generation.json`'s `image_model` block — which the code had never read — is now authoritative.

**Also `output_format: jpeg`.** The model emits 1536×1536 while avatars render at 32-128 CSS px (`w-8` in chat bubbles, `max-w-[128px]` for the large variant, 80 px in the PDF export). Measured on one seed: **png 2,677,833 bytes vs jpeg 401,603 — 6.7× smaller for identical pixels** (`webp` is rejected by the model). Resolution itself is not reducible here: these models expose `aspect_ratio` only, no width/height. Cutting pixels would need Pillow and a layer rebuild, deliberately left out.

## 3. App-review ingestors were sized for less work than they do

The iOS ingestor walks all 40 storefronts in `IOS_COUNTRIES`, paginating each, and only returns once the whole loop finishes — so a timeout discards everything collected. It ran on the 300s/512MB default while pagination work increased how much each run fetches.

- iOS: `timeout 300 → 900`, `memory 512 → 1024`
- Android: `timeout 300 → 600`, `memory 512 → 1024`

Both sit inside the ceilings `plugin-loader.ts` already enforces. Tests read those ceilings from the loader rather than restating them, and tie the iOS timeout to the storefront count so adding countries forces a review.

## 4. The research Lambda shipped without its prompt config

Found by inspecting the deployed bundle: the research Lambda had **no `prompts/` directory at all**, so `get_research_step_config()` would raise `FileNotFoundError` on every job. Reading budgets from JSON turned a build-time constant into a runtime file dependency, and the research asset excluded the whole `/api/` subtree.

Unit tests structurally could not catch this — locally `get_prompts_dir()` resolves the repo layout either way. The `cp` into the bundle is the actual fix; the `/api/*` + `!/api/prompts` exclude keeps the config in the asset fingerprint so editing it redeploys the function.

## Verification

Gates on the current head:

- backend pytest **1223 passed**
- CDK vitest **92 passed**, `tsc --noEmit` clean (re-confirmed after the latest test edits)
- zero new ruff findings (project baseline is pre-existing)
- fail-on-revert verified for change 1 (4 tests fail against the reverted code)

Deployed to a test account and verified live:

- ingestor config live: iOS `Timeout=900 MemorySize=1024`, Android `Timeout=600 MemorySize=1024`
- `scripts/test-api.sh` **32/32**
- deployed research bundle re-inspected: `prompts/` present with all six configs, and a **live invocation of the deployed research Lambda logged** `max_tokens=9000, thinking_budget=5000` with zero `FileNotFoundError`
- **avatar path against real Bedrock:** a live generation on `stability.stable-image-core-v1:1` in us-west-2 returned one image, 2,677,833 bytes as PNG and 401,603 as JPEG, valid magic bytes, `finish_reasons=[None]`, reproducible seed
- **deployed IAM for the new cross-region ARN:** `iam simulate-principal-policy` returns **allowed** for `ProjectsLambdaRole`, `PersonaGeneratorRole` and `PersonaImporterRole` — the half local invocation cannot prove, since CLI credentials are Admin while the Lambdas use narrow roles

### Known gaps in verification

- The avatar path has **not** been exercised through the persona-generation UI or the PDF export. The S3 key and `ContentType` derive from `output_format`, and an audit found the key is constructed in exactly one place (`avatar.py`) with `core-stack.ts`/`api-stack.ts` using `avatars/*` wildcards and `get_avatar_cdn_url` parsing the trailing segment generically — but that is static evidence, not a rendered avatar.
- The last commits are **not deployed**; their runtime effect is limited to log labels, the seed offset and orphan cleanup.

## Notes for reviewers

- Change 1 alters no prompts, only budgets.
- Because the S3 key embeds the image format, a format change would orphan a persona's previous avatar; generation now best-effort deletes the other-extension objects, since no other avatar cleanup path exists in the codebase.
- The Python↔TypeScript lockstep tests parse TS with regex, following the existing `test_model_config.py` precedent rather than introducing a second convention.
- Per-storefront progress persistence for the iOS sweep is the real fix for that ingestor and is intentionally left to its own PR.

### DIFF
--- a/voc-datalake/lambda/api/prompts/avatar-generation.json
+++ b/voc-datalake/lambda/api/prompts/avatar-generation.json
@@ -9,6 +9,6 @@
     "model_id": "stability.stable-image-core-v1:1",
     "region": "us-west-2",
     "aspect_ratio": "1:1",
-    "output_format": "png"
+    "output_format": "jpeg"
   }
 }
--- a/voc-datalake/lambda/api/prompts/avatar-generation.json
+++ b/voc-datalake/lambda/api/prompts/avatar-generation.json
@@ -6,9 +6,9 @@
   "fallback_prompt_template": "Professional headshot of a {occupation}, friendly expression, soft studio lighting, neutral background, photorealistic",
   "max_tokens": 200,
   "image_model": {
-    "model_id": "amazon.nova-canvas-v1:0",
-    "region": "us-east-1",
-    "width": 1024,
-    "height": 1024
+    "model_id": "stability.stable-image-core-v1:1",
+    "region": "us-west-2",
+    "aspect_ratio": "1:1",
+    "output_format": "png"
   }
 }
--- a/voc-datalake/lambda/research/research_step_handler.py
+++ b/voc-datalake/lambda/research/research_step_handler.py
@@ -61,6 +61,7 @@ def invoke_bedrock_with_retry(
     max_tokens: int = 4096,
     max_retries: int = 3,
     thinking_budget: int = 0,
+    step_name: str = 'unknown',
 ) -> str:
     """Invoke Bedrock with retry support using shared converse module."""
     return converse(
@@ -71,6 +72,7 @@ def invoke_bedrock_with_retry(
         surface='documents',
         max_retries=max_retries,
         raise_on_throttle=True,
+        step_name=step_name,
     )
 
 
@@ -97,6 +99,9 @@ def _step_inference(step_name: str, config: dict) -> dict:
         'system_prompt': system_prompt,
         'max_tokens': step_config['max_tokens'],
         'thinking_budget': step_config['thinking_budget'],
+        # Carried through to converse() so [BEDROCK] log lines identify which
+        # research step they belong to; all three logged as 'unknown' before.
+        'step_name': step_name,
     }
 # Wrapper function to pass module-level table reference to shared function
 def get_feedback_context(filters: dict, limit: int = 100) -> list[dict]:
@@ -283,6 +288,7 @@ IMPORTANT: Base ALL findings on the actual feedback data provided. Do not make a
         user_prompt,
         max_tokens=inference['max_tokens'],
         thinking_budget=inference['thinking_budget'],
+        step_name=inference['step_name'],
     )
     
     update_job_status(project_id, job_id, 'running', 45, 'analysis_complete')
@@ -319,6 +325,7 @@ Provide:
         user_prompt,
         max_tokens=inference['max_tokens'],
         thinking_budget=inference['thinking_budget'],
+        step_name=inference['step_name'],
     )
     
     update_job_status(project_id, job_id, 'running', 70, 'synthesis_complete')
@@ -360,6 +367,7 @@ Provide a final validated research report."""
         user_prompt,
         max_tokens=inference['max_tokens'],
         thinking_budget=inference['thinking_budget'],
+        step_name=inference['step_name'],
     )
     
     update_job_status(project_id, job_id, 'running', 90, 'validation_complete')
--- a/voc-datalake/lambda/research/research_step_handler.py
+++ b/voc-datalake/lambda/research/research_step_handler.py
@@ -15,6 +15,10 @@ from shared.logging import logger, tracer
 from shared.aws import get_dynamodb_resource, BEDROCK_MODEL_ID
 from shared.api import api_handler
 from shared.converse import converse, BedrockThrottlingError
+from shared.prompts import (
+    get_research_step_config,
+    get_response_language_instruction,
+)
 from shared.feedback import (
     get_feedback_context as _get_feedback_context,
     format_feedback_for_llm,
@@ -51,16 +55,49 @@ def _get_projects_table():
     return projects_table
 # Alias for backward compatibility with Step Functions error handling
 BedrockThrottlingException = BedrockThrottlingError
-def invoke_bedrock_with_retry(system_prompt: str, user_message: str, max_tokens: int = 4096, max_retries: int = 3) -> str:
+def invoke_bedrock_with_retry(
+    system_prompt: str,
+    user_message: str,
+    max_tokens: int = 4096,
+    max_retries: int = 3,
+    thinking_budget: int = 0,
+) -> str:
     """Invoke Bedrock with retry support using shared converse module."""
     return converse(
         prompt=user_message,
         system_prompt=system_prompt,
         max_tokens=max_tokens,
+        thinking_budget=thinking_budget,
         surface='documents',
         max_retries=max_retries,
         raise_on_throttle=True,
     )
+
+
+def _step_inference(step_name: str, config: dict) -> dict:
+    """Resolve a research step's system prompt and token budgets.
+
+    Both research paths read the SAME config (lambda/api/prompts/
+    research-analysis.json): the sync path via build_chain_steps, this async
+    Step Functions path via here. Previously this module hardcoded its own
+    budgets and duplicated the system prompts, so edits to that JSON silently
+    had no effect on the async path — it looked authoritative but was never
+    read. Keep it that way: budgets belong in the JSON, not inline.
+
+    The user prompt stays inline per step, because this path adds context the
+    templated sync path has no placeholders for (personas, uploaded documents,
+    public web-search results).
+    """
+    step_config = get_research_step_config(step_name)
+    system_prompt = step_config['system_prompt']
+    lang_instruction = get_response_language_instruction(config.get('response_language'))
+    if lang_instruction:
+        system_prompt = f"{system_prompt}\n\n{lang_instruction}"
+    return {
+        'system_prompt': system_prompt,
+        'max_tokens': step_config['max_tokens'],
+        'thinking_budget': step_config['thinking_budget'],
+    }
 # Wrapper function to pass module-level table reference to shared function
 def get_feedback_context(filters: dict, limit: int = 100) -> list[dict]:
     """Get feedback items based on filters for LLM context."""
@@ -197,18 +234,8 @@ def step_analyze(event: dict) -> dict:
     logger.info(f"Starting analysis for job {job_id}")
     update_job_status(project_id, job_id, 'running', 25, 'preparing_analysis')
     
-    system_prompt = """You are a senior user researcher conducting rigorous analysis of REAL customer feedback data.
-Your analysis must be grounded in the actual feedback provided - cite specific reviews, quote customers directly, and identify patterns from the data.
-Be thorough, data-driven, and cite specific examples."""
+    inference = _step_inference('data_analysis', config)
 
-    # Inject language instruction if non-English
-    response_language = config.get('response_language')
-    if response_language:
-        from shared.prompts import get_response_language_instruction
-        lang_instruction = get_response_language_instruction(response_language)
-        if lang_instruction:
-            system_prompt += f"\n\n{lang_instruction}"
-    
     # Build additional context sections
     additional_context = ""
     if personas_context:
@@ -251,7 +278,12 @@ Based on the ACTUAL FEEDBACK DATA above{' and the provided context' if additiona
 IMPORTANT: Base ALL findings on the actual feedback data provided. Do not make assumptions beyond what the data shows."""
 
     update_job_status(project_id, job_id, 'running', 30, 'calling_ai')
-    analysis = invoke_bedrock_with_retry(system_prompt, user_prompt, max_tokens=4000)
+    analysis = invoke_bedrock_with_retry(
+        inference['system_prompt'],
+        user_prompt,
+        max_tokens=inference['max_tokens'],
+        thinking_budget=inference['thinking_budget'],
+    )
     
     update_job_status(project_id, job_id, 'running', 45, 'analysis_complete')
     
@@ -267,17 +299,8 @@ def step_synthesize(event: dict) -> dict:
     logger.info(f"Synthesizing findings for job {job_id}")
     update_job_status(project_id, job_id, 'running', 50, 'preparing_synthesis')
     
-    system_prompt = """You are synthesizing research findings into actionable insights.
-Focus on clarity, prioritization, and recommendations."""
+    inference = _step_inference('synthesis', config)
 
-    # Inject language instruction if non-English
-    response_language = config.get('response_language')
-    if response_language:
-        from shared.prompts import get_response_language_instruction
-        lang_instruction = get_response_language_instruction(response_language)
-        if lang_instruction:
-            system_prompt += f"\n\n{lang_instruction}"
-    
     user_prompt = f"""Synthesize the analysis into clear findings.
 
 Previous analysis:
@@ -291,7 +314,12 @@ Provide:
 5. **Areas for Further Research**"""
 
     update_job_status(project_id, job_id, 'running', 55, 'calling_ai')
-    synthesis = invoke_bedrock_with_retry(system_prompt, user_prompt, max_tokens=3000)
+    synthesis = invoke_bedrock_with_retry(
+        inference['system_prompt'],
+        user_prompt,
+        max_tokens=inference['max_tokens'],
+        thinking_budget=inference['thinking_budget'],
+    )
     
     update_job_status(project_id, job_id, 'running', 70, 'synthesis_complete')
     
@@ -308,17 +336,8 @@ def step_validate(event: dict) -> dict:
     logger.info(f"Validating research for job {job_id}")
     update_job_status(project_id, job_id, 'running', 75, 'preparing_validation')
     
-    system_prompt = """You are a critical reviewer ensuring research quality.
-Challenge assumptions and verify conclusions."""
+    inference = _step_inference('validation', config)
 
-    # Inject language instruction if non-English
-    response_language = config.get('response_language')
-    if response_language:
-        from shared.prompts import get_response_language_instruction
-        lang_instruction = get_response_language_instruction(response_language)
-        if lang_instruction:
-            system_prompt += f"\n\n{lang_instruction}"
-    
     user_prompt = f"""Review and validate the research findings.
 
 Analysis:
@@ -336,7 +355,12 @@ Check:
 Provide a final validated research report."""
 
     update_job_status(project_id, job_id, 'running', 80, 'calling_ai')
-    validation = invoke_bedrock_with_retry(system_prompt, user_prompt, max_tokens=3000)
+    validation = invoke_bedrock_with_retry(
+        inference['system_prompt'],
+        user_prompt,
+        max_tokens=inference['max_tokens'],
+        thinking_budget=inference['thinking_budget'],
+    )
     
     update_job_status(project_id, job_id, 'running', 90, 'validation_complete')
     
--- a/voc-datalake/lambda/research/research_step_handler.py
+++ b/voc-datalake/lambda/research/research_step_handler.py
@@ -99,9 +99,9 @@ def _step_inference(step_name: str, config: dict) -> dict:
         'system_prompt': system_prompt,
         'max_tokens': step_config['max_tokens'],
         'thinking_budget': step_config['thinking_budget'],
-        # Carried through to converse() so [BEDROCK] log lines identify which
-        # research step they belong to; all three logged as 'unknown' before.
-        'step_name': step_name,
+        # Resolved by shared.prompts from the config's 'name', so [BEDROCK] log
+        # lines match between this path and the sync chain.
+        'step_name': step_config['step_name'],
     }
 # Wrapper function to pass module-level table reference to shared function
 def get_feedback_context(filters: dict, limit: int = 100) -> list[dict]:
--- a/voc-datalake/lambda/research/test/test_research_step_budgets.py
+++ b/voc-datalake/lambda/research/test/test_research_step_budgets.py
@@ -1,0 +1,179 @@
+"""The async research path must take its budgets and system prompts from the
+shared prompt config, not from inlined copies.
+
+Background: research-analysis.json declared max_tokens 9000 per step while this
+Lambda hardcoded 4000/3000/3000 and duplicated the system prompts, so editing
+the config had no effect on the async Step Functions path — it looked
+authoritative but was never read. These tests fail if that decoy returns.
+
+The expected values are read from the JSON with a plain json.load rather than
+through shared.prompts, so the config is an INDEPENDENT oracle: a bug in the
+loader can't make the handler and the expectation drift together.
+"""
+import json
+from pathlib import Path
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+# lambda/research/test/ -> lambda/
+LAMBDA_DIR = Path(__file__).resolve().parents[2]
+RESEARCH_PROMPTS = LAMBDA_DIR / 'api' / 'prompts' / 'research-analysis.json'
+
+# The budgets this Lambda used to hardcode. Kept as explicit literals (not read
+# from anywhere) so re-introducing them fails loudly, while still leaving the
+# config free to be re-tuned above them.
+PRE_FIX_ANALYSIS_BUDGET = 4000
+PRE_FIX_SYNTHESIS_BUDGET = 3000
+PRE_FIX_VALIDATION_BUDGET = 3000
+
+
+def _step(step_name: str) -> dict:
+    return json.loads(RESEARCH_PROMPTS.read_text(encoding='utf-8'))['steps'][step_name]
+
+
+@pytest.fixture
+def mock_tables():
+    mock_fb = MagicMock()
+    mock_proj = MagicMock()
+    with patch('research_step_handler._get_feedback_table', return_value=mock_fb), \
+         patch('research_step_handler._get_projects_table', return_value=mock_proj):
+        yield {'feedback': mock_fb, 'projects': mock_proj}
+
+
+@pytest.fixture
+def mock_job_status():
+    with patch('research_step_handler.update_job_status') as m:
+        yield m
+
+
+@pytest.fixture
+def mock_converse():
+    with patch('research_step_handler.converse', return_value='result') as m:
+        yield m
+
+
+def _analyze_event(**config):
+    return {
+        'project_id': 'p1', 'job_id': 'j1',
+        'research_config': {'question': 'Q?', **config},
+        'feedback_context': 'fb', 'feedback_stats': 's',
+    }
+
+
+def _synthesize_event(**config):
+    return {
+        'project_id': 'p1', 'job_id': 'j1',
+        'research_config': {'question': 'Q?', **config},
+        'analysis': 'prior analysis',
+    }
+
+
+def _validate_event(**config):
+    return {
+        'project_id': 'p1', 'job_id': 'j1',
+        'research_config': {'question': 'Q?', **config},
+        'analysis': 'prior analysis', 'synthesis': 'prior synthesis',
+    }
+
+
+class TestBudgetsComeFromConfig:
+    """Each step's max_tokens must equal what the shared config declares."""
+
+    def test_analysis_uses_the_configured_budget(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_analyze
+
+        step_analyze(_analyze_event())
+        assert mock_converse.call_args.kwargs['max_tokens'] == _step('data_analysis')['max_tokens']
+
+    def test_synthesis_uses_the_configured_budget(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_synthesize
+
+        step_synthesize(_synthesize_event())
+        assert mock_converse.call_args.kwargs['max_tokens'] == _step('synthesis')['max_tokens']
+
+    def test_validation_uses_the_configured_budget(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_validate
+
+        step_validate(_validate_event())
+        assert mock_converse.call_args.kwargs['max_tokens'] == _step('validation')['max_tokens']
+
+    def test_configured_thinking_budget_reaches_bedrock(self, mock_tables, mock_job_status, mock_converse):
+        """data_analysis declares a thinking budget; it must be forwarded.
+
+        converse() drops it for adaptive-thinking models on its own, so passing
+        it is safe — but silently discarding it here would make the config lie.
+        """
+        from research_step_handler import step_analyze
+
+        step_analyze(_analyze_event())
+        assert mock_converse.call_args.kwargs['thinking_budget'] == _step('data_analysis')['thinking_budget']
+
+
+class TestBudgetsDoNotRegressToTheOldHardcodes:
+    """Fail-on-revert: the point of the change was to stop running the research
+    chain on budgets small enough to force continuation loops and empty-text
+    retries. Asserted as a floor, so re-tuning the config stays possible."""
+
+    def test_analysis_budget_exceeds_the_old_hardcode(self):
+        assert _step('data_analysis')['max_tokens'] > PRE_FIX_ANALYSIS_BUDGET
+
+    def test_synthesis_budget_exceeds_the_old_hardcode(self):
+        assert _step('synthesis')['max_tokens'] > PRE_FIX_SYNTHESIS_BUDGET
+
+    def test_validation_budget_exceeds_the_old_hardcode(self):
+        assert _step('validation')['max_tokens'] > PRE_FIX_VALIDATION_BUDGET
+
+
+class TestSystemPromptsComeFromConfig:
+    """The system prompts were duplicated between this module and the config.
+    Pin them to the config so the two paths can't drift."""
+
+    def test_analysis_system_prompt_matches_config(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_analyze
+
+        step_analyze(_analyze_event())
+        assert mock_converse.call_args.kwargs['system_prompt'] == _step('data_analysis')['system_prompt']
+
+    def test_synthesis_system_prompt_matches_config(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_synthesize
+
+        step_synthesize(_synthesize_event())
+        assert mock_converse.call_args.kwargs['system_prompt'] == _step('synthesis')['system_prompt']
+
+    def test_validation_system_prompt_matches_config(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_validate
+
+        step_validate(_validate_event())
+        assert mock_converse.call_args.kwargs['system_prompt'] == _step('validation')['system_prompt']
+
+
+class TestLanguageInstructionStillApplies:
+    """Sourcing the prompt from config must not drop the per-request language
+    instruction, which is appended on top of it."""
+
+    def test_analysis_appends_language_instruction(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_analyze
+
+        step_analyze(_analyze_event(response_language='es'))
+        system_prompt = mock_converse.call_args.kwargs['system_prompt']
+        assert system_prompt.startswith(_step('data_analysis')['system_prompt'])
+        assert 'Spanish' in system_prompt
+
+    def test_synthesis_appends_language_instruction(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_synthesize
+
+        step_synthesize(_synthesize_event(response_language='ko'))
+        assert 'Korean' in mock_converse.call_args.kwargs['system_prompt']
+
+    def test_validation_appends_language_instruction(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_validate
+
+        step_validate(_validate_event(response_language='ja'))
+        assert 'Japanese' in mock_converse.call_args.kwargs['system_prompt']
+
+    def test_no_language_leaves_the_config_prompt_untouched(self, mock_tables, mock_job_status, mock_converse):
+        from research_step_handler import step_analyze
+
+        step_analyze(_analyze_event())
+        assert mock_converse.call_args.kwargs['system_prompt'] == _step('data_analysis')['system_prompt']
--- a/voc-datalake/lambda/research/test/test_research_step_budgets.py
+++ b/voc-datalake/lambda/research/test/test_research_step_budgets.py
@@ -98,6 +98,21 @@ class TestBudgetsComeFromConfig:
         step_validate(_validate_event())
         assert mock_converse.call_args.kwargs['max_tokens'] == _step('validation')['max_tokens']
 
+    def test_each_step_identifies_itself_in_bedrock_logs(self, mock_tables, mock_job_status, mock_converse):
+        """converse() logs step_name; all three research steps used to log
+        'unknown', so live logs could not tell which step made a call.
+        (Raised in review of PR #228.)"""
+        from research_step_handler import step_analyze, step_synthesize, step_validate
+
+        step_analyze(_analyze_event())
+        assert mock_converse.call_args.kwargs['step_name'] == 'data_analysis'
+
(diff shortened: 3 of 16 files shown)


perrozzi:

This PR does three unrelated cleanups plus a deploy fix: the async Step Functions research path now reads system prompts and token budgets from `lambda/api/prompts/research-analysis.json` (instead of hardcoding `4000/3000/3000` and duplicating prompts), the persona avatar image model is single-sourced through `lib/utils/model-allowlist.ts` + `avatar-generation.json` with an EOL migration note, the app-review ingestor manifests get bigger timeout/memory budgets, and the research Lambda asset now stages `api/prompts`. Direction is good and the model-id/IAM-ARN lockstep rule is respected (`imageModelArn()` used in all three roles, image model deliberately kept out of `ALLOWED_MODEL_IDS`). No blockers found, but several of the new tests are source-text greps rather than behavioural guards, and one runtime-budget knock-on is unaddressed.

### Should fix

- **Research step Lambda timeout wasn't revisited.** `lambda/research/research_step_handler.py` now runs each step at the config's `max_tokens` (9000, per the PR description) plus a 5000-token `thinking_budget`, roughly tripling per-step generation latency versus the old 4000/0. The Lambda's timeout/memory in `processing-stack-consolidated.ts` is untouched in this diff. That is precisely the failure mode change 3 fixes for the ingestors — a mid-generation timeout throws the step away. Please state the current timeout and, ideally, add an assertion tying it to the configured budgets the way `test_app_reviews_manifest_limits.py` ties iOS timeout to storefront count.
- **The avatar fallback defaults are outside the lockstep test.** `test_avatar_image_model_lockstep.py` pins `avatar-generation.json` ↔ `model-allowlist.ts`, but nothing pins `shared/avatar.py`'s `DEFAULT_IMAGE_MODEL_ID` / `DEFAULT_IMAGE_MODEL_REGION` to `IMAGE_MODEL_ID` / `IMAGE_MODEL_REGION`. If the `image_model` block is ever dropped from the JSON, the runtime silently falls back to a constant that no longer has to match the IAM grant — a quiet `AccessDenied` on a code path that already degrades to `avatar_url=None`. Either add that assertion, or make the config authoritative and raise on a missing block.
- **The bundle regression test greps its own source file.** `processing-stack-consolidated.test.ts` reads `processing-stack-consolidated.ts` as text and asserts `'/api/*'`, `'!/api/prompts'` and the exact `cp` string. It breaks on any reformatting and proves nothing about staging. Note also that for a bundled `Code.fromAsset`, `exclude` chiefly governs the source fingerprint while bundling mounts the source dir — so the real fix here is the added `cp -r /asset-input/api/prompts /asset-output/prompts`, and the exclude change is what makes prompt edits invalidate the asset hash. Worth correcting the comment, and consider a runtime-side guard instead (e.g. a test that each research step's config loads via `get_research_step_config`, so a rename in the JSON also fails).
- **Python tests parsing TypeScript with regex.** `_ts_const()` in `test_avatar_image_model_lockstep.py` and `_loader_ceiling()` in `plugins/test_app_reviews_manifest_limits.py` depend on exact TS formatting (`export const X = '...';`, `z.number().int().min(N).max(N)`). Prettier reflow or a `.default()` insertion turns these into confusing failures — or worse, `_loader_ceiling` asserts fail loudly while `_ios_storefront_count` returning 0 would only be caught by one guard test. Prefer emitting shared constants to a small JSON that both sides read.
- **`avatar-generation.json` isn't in the diff.** The lockstep tests index `['image_model']['model_id']` and `['image_model']['region']` directly; if the existing block lacks `region`, those tests `KeyError`. Please confirm (or include the file).
- **iOS ingestor sizing is a band-aid.** As the PR itself notes, the storefront loop only returns after all 40 countries, so a timeout still discards everything. 900s is also close to the practical ceiling under a `rate(30 minutes)` schedule. Consider persisting per-storefront batches so partial progress survives; otherwise this recurs the next time countries or pagination depth grow.

### Nits

- Production comments in `shared/avatar.py`, `shared/prompts.py`, `research_step_handler.py` and `model-allowlist.ts` narrate PR history at length ("the same decoy-config trap", "it looked authoritative but was never read"). One line plus a pointer to the config is enough; the rest belongs in the commit message.
- `get_image_model_config()` re-reads the prompt config on every avatar generation; wrap `load_prompt_file` (or this accessor) in `functools.lru_cache` if it isn't already cached.
- No test pins the old inline system prompts, so the "byte-identical, no prompt behaviour change" claim is unverified in CI — a one-time snapshot of the pre-fix strings would have made it checkable.
- `test_ios_gets_at_least_as_long_as_android` and `TestBudgetsWereRaised` are near-tautological restatements of the manifest values; the storefront-scaling test is the only one carrying real information.
- Follow-up you already spotted (`invoke_bedrock_with_retry` not passing `step_name`, so all steps log as `'unknown'`) is a two-line addition — worth folding in here since you're already touching that call site.

---

perrozzi:

Since my last pass the author plumbed `step_name` through `invoke_bedrock_with_retry` → `converse()` (with a test), added a lockstep assertion for `avatar.py`'s `DEFAULT_IMAGE_MODEL_*`, added a synth-based assertion that the research Lambda sits at the 900s/≥1024MB ceiling, and rewrote the bundle-staging test/comments so the `cp` line is identified as the actual fix and the exclude as the fingerprint guard. Three of the previous should-fix items are now addressed; two remain.

### Resolved
- Research step Lambda timeout: now asserted from the synthesized template (`processing-stack-consolidated.test.ts`), with the 900s hard-max rationale documented.
- Avatar fallback defaults are pinned to `IMAGE_MODEL_ID`/`IMAGE_MODEL_REGION` in `test_avatar_image_model_lockstep.py`.
- Bundle test comment corrected: `cp -r /asset-input/api/prompts` is the fix, the exclude is the fingerprint half; the misleading gitignore-semantics assertion was dropped.
- Follow-up nit folded in: `step_name` reaches `converse()`, so steps no longer log as `'unknown'`.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`) — still formatting-fragile.
- `lambda/api/prompts/avatar-generation.json` is still absent from the diff while the lockstep test indexes `['image_model']['model_id']`/`['region']` directly.
- iOS ingestor sizing remains a band-aid (all-or-nothing storefront loop); no partial-progress persistence.

### New findings
- **`lib/stacks/processing-stack-consolidated.test.ts`: the bundle-staging doc comment is now detached from its `describe`.** Two block comments are stacked back-to-back, and the first (which describes prompt staging and the FileNotFoundError history) sits immediately above `describe('research Lambda keeps the maximum timeout its budgets assume')`, not above `describe('research Lambda bundle stages its prompt config')`. Move it down so each block documents the suite it belongs to.
- **`lambda/research/research_step_handler.py`: async `step_name` values won't match the sync path's.** `_step_inference` returns the raw config key (`data_analysis`, `synthesis`, `validation`), whereas `build_chain_steps` in `shared/prompts.py` uses `step.get('name', step_name)` — the human-readable label. For the same step, sync and async runs will emit different `[BEDROCK]` step names, which undercuts the point of the change. Consider reading `name` in `_step_inference` too (and asserting it in the new test).
- Nit: locating the function via `id.startsWith('ResearchStepLambda')` couples the test to the construct id; a lookup by a stable property (e.g. `Handler` containing `research_step_handler`) would survive a rename.

The verbose PR-history narration in the new comment blocks in `processing-stack-consolidated.ts`/`.test.ts` grew rather than shrank; still a nit, not a blocker.

---

perrozzi:
Thanks — this was a useful round. One finding overturned a claim I had made in a code comment, which is the best kind. Addressed in `b671e1e`.

## Fixed

**The bundle test's stated mechanism was wrong, and so was my comment.** You're right that a bundled `Code.fromAsset` mounts the source dir as `/asset-input`. I tested it by reverting each half independently: with `'/api/'` pruned and the `cp` in place, the prompts **still land** and the asset hash changes (`bdea533b` → `67ac12a8`). So the added `cp -r /asset-input/api/prompts` is the entire fix, and the exclude never gated staging.

The per-entry exclude is still worth keeping, but for the reason you identified rather than the one I wrote: it keeps `api/prompts` in the asset **fingerprint**, so editing `research-analysis.json` invalidates the hash and actually redeploys. Otherwise the config would be bundled but a prompt-only edit would leave the Lambda running the old budgets — a nastier failure than the original one. Comment and test rationale corrected, and the assertion that maps to the real bug (the `cp`) is now first with that noted.

**The avatar fallback defaults were a genuine hole.** `DEFAULT_IMAGE_MODEL_ID` / `DEFAULT_IMAGE_MODEL_REGION` were a third copy of the id that nothing pinned. Exactly as you describe: drop the `image_model` block and the fallback could invoke a model the IAM grant doesn't cover, AccessDenied, on a path that already degrades silently to `avatar_url=None`. Now asserted against the CDK constants in the same lockstep test.

**`step_name` folded in.** All three steps logged as `'unknown'`, which the live verification made obvious. `_step_inference()` carries the step name through to `converse()`, with a test pinning `data_analysis` / `synthesis` / `validation`.

**Research Lambda timeout — checked, and guarded.** It is already `Duration.minutes(15)` = 900s, Lambda's hard maximum (deployed value confirmed: `Timeout=900, MemorySize=1024`), so there was no headroom to add. Since budgets are now editable in a JSON file, I added template assertions that the timeout stays at the maximum and memory stays at least 1024 (memory determines CPU, so it affects generation speed). Worth noting `converse()` also has a 420s `_EMPTY_RAISE_DEADLINE_SECONDS` guard that refuses a doubled-budget retry late in an invocation, precisely to avoid dying mid-retry.

## Reviewed, not changed

**Python tests parsing TypeScript.** Fair in the abstract, but this is the established pattern here — `lambda/shared/test/test_model_config.py` already regexes `model-allowlist.ts` and `model-override.ts` for the four-way model lockstep. Introducing a shared-JSON mechanism for my two helpers would leave the repo with two competing conventions; if we want that, it should convert the existing lockstep tests too. Your sub-point about a silent `_ios_storefront_count() == 0` is already covered by `test_storefront_list_is_actually_large`, which exists to stop the scaling assertion going slack.

**`avatar-generation.json` not in the diff.** Confirmed it already carries the full block, `region` included, so no `KeyError` — the 11 lockstep tests pass against it. It's absent from the diff because only the reader changed, not the config.

**`get_image_model_config()` re-reading per generation.** Already cached: `load_prompt_file` is `@lru_cache(maxsize=32)`, so this is one file read per container.

**Byte-identical prompt claim unverified in CI.** True, and deliberate. I verified it before deleting the inline copies by extracting both and comparing programmatically, but snapshotting now-deleted strings would pin history rather than behaviour. The forward-looking guard is `TestSystemPromptsComeFromConfig`, which fails if either path drifts from the config.

**iOS sizing is a band-aid.** Agreed, and the PR says so. Per-storefront persistence so partial progress survives is the real fix, and it's a behavioural change to the ingestor rather than a config bump — better as its own PR than smuggled into this one. The new test at least forces a timeout review when storefront count grows.

**Comment verbosity.** Taking the point partially: I trimmed the mechanism comment while correcting it. I've left the rationale elsewhere, because this codebase deliberately documents *why* at the point of the constraint — `converse.py`'s `_EMPTY_RAISE_CEILING` and `lambda-asset-excludes.ts` both carry long explanations of exactly this kind, and they're the reason I could reason about budget behaviour here at all.

**Near-tautological tests.** `TestBudgetsWereRaised` asserts against explicit literals for the *old* values, which is what makes it a fail-on-revert guard rather than a restatement — it passed against reverted code only because the JSON is the oracle, and 4 of the 14 did fail on revert. `test_ios_gets_at_least_as_long_as_android` is the weak one; I've left it as cheap documentation of intent.

Gates after these changes: backend pytest **1211 passed**, CDK vitest **92 passed**, `tsc --noEmit` clean.

perrozzi:
_(Answering **review round 2** on `b671e1e`. Addressed in `0b1dd1a`, pushed right after this comment.)_

Both new findings were real defects introduced by the previous commit. Good catches.

## Fixed

**Step names diverged between the two research paths.** You're right, and it defeated the purpose of the change. `build_chain_steps` resolves the label from the config's `name` field; `_step_inference` returned the raw dict key. They happen to be equal in `research-analysis.json` today, so nothing was visibly wrong — but a config where they differ would have emitted different `[BEDROCK]` labels for the same step, which is exactly the confusion `step_name` was added to remove.

Rather than reading `name` in both places, resolution now lives once in `shared/prompts.py::_inference_from_step`, which both paths already go through, and `build_chain_steps` no longer does its own `step.get('name', step_name)`. New test pins the async value to what the config declares, so the two cannot drift again.

**The doc comment was detached.** Correct — inserting the timeout suite left the staging comment above the wrong `describe`. Moved back to its own suite.

**Construct-id lookup replaced.** Also correct, and worse than cosmetic: `id.startsWith('ResearchStepLambda')` meant a rename would silently skip every assertion rather than fail. Now matched on the handler containing `research_step_handler`, via a small guard that validates the CloudFormation shape instead of casting.

**Comment verbosity — conceded this round.** You flagged it twice and noted it grew rather than shrank, which is fair. Trimmed the asset-staging block in `processing-stack-consolidated.ts` and both test suite headers: constraint kept, PR history dropped.

## Still-open items, unchanged from my previous comment

Those three were dispositioned in the comment directly above (which answers round 1 on `dcbe6f4` — it crossed with your round 2 by about a minute, so it appears below it in the thread):

- **Python parsing TypeScript:** follows the existing `test_model_config.py` lockstep precedent; converting only the new tests would leave two competing conventions in the repo.
- **`avatar-generation.json` absent from the diff:** verified it already carries the full `image_model` block including `region`, so no `KeyError` — the lockstep tests pass against it. It is absent because only the reader changed.
- **iOS band-aid:** agreed, and per-storefront progress persistence is the right fix. It changes ingestor behaviour rather than config, so it belongs in its own PR.

Gates on `0b1dd1a`: backend pytest **1212 passed**, CDK vitest **92 passed**, `tsc --noEmit` clean.

perrozzi:

Since my last pass the author moved the `step_name` resolution into `shared.prompts._inference_from_step` (so both research paths derive it from the config's `name`), added a matching test, replaced the construct-id lookup with a `Handler`-based match plus a `propsOf` type guard, and re-attached/trimmed the two doc comments in `processing-stack-consolidated.test.ts`/`.ts`. Both of my previous new findings are addressed; the three older should-fix items are not.

### Resolved
- `research_step_handler._step_inference` now emits the same `step_name` as `build_chain_steps` (single home in `_inference_from_step`), with a test asserting parity against the config's `name`.
- The bundle-staging doc comment now sits above `describe('research Lambda bundle stages its prompt config')`, and the timeout suite has its own shorter comment.
- Nit addressed: the research Lambda is located by `Handler` rather than construct id, via a `propsOf` guard instead of a bare cast.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`).
- `lambda/api/prompts/avatar-generation.json` is still absent from the diff while the lockstep test indexes `['image_model']['model_id']`/`['region']`.
- iOS ingestor sizing remains a band-aid; no partial-progress persistence.

### New findings
- Nit — `lambda/research/test/test_research_step_budgets.py`: `test_each_step_identifies_itself_in_bedrock_logs` hardcodes `'data_analysis'`/`'synthesis'`/`'validation'`, which now only passes because the JSON's `name` fields happen to equal the step keys. Adding a human-readable `"name"` to `research-analysis.json` would fail that test while `test_step_name_matches_the_sync_chain_for_the_same_step` passes — the two assertions contradict each other. Derive both from the config (or drop the literal one).
- Nit — `lambda/shared/prompts.py`: `get_step_inference_config`'s docstring still says it returns "'system_prompt', 'max_tokens' and 'thinking_budget'"; it now also returns `step_name`.
- Nit — `lib/stacks/processing-stack-consolidated.test.ts`: `.find()` asserts only the first Lambda whose handler mentions `research_step_handler`; `filter()` + asserting every match would keep a second research function from being deployed under the ceiling unnoticed.

Otherwise the incremental changes look clean; no blockers.

---

perrozzi:
_(Answering **review round 3** on `0b1dd1a`. Addressed in `a8af9ad`, pushed right after this comment.)_

All three nits were real. The first was the good one.

**The two step-name assertions contradicted each other.** Exactly as described: `test_each_step_identifies_itself_in_bedrock_logs` hardcoded the three keys while its sibling derived them from the config, and both passed only because `name == key` in `research-analysis.json` today. Adding a human-readable `name` would have failed one and passed the other — so the pair actively undermined the parity it exists to protect. Merged into a single test that derives every expectation from the config, the same way `build_chain_steps` resolves it.

**Stale docstring** on `get_step_inference_config` — it now returns `step_name` too. Fixed.

**`.find()` → `filter()`** in the timeout suite, with both assertions looping over every match, so a second research function can't be added under the ceiling unnoticed.

The three older should-fix items remain as dispositioned two comments up: the TypeScript-parsing helpers follow the existing `test_model_config.py` precedent, `avatar-generation.json` already carries the `image_model` block (verified — absent from the diff only because the reader changed), and per-storefront progress for the iOS sweep belongs in its own PR since it changes ingestor behaviour rather than config.

Gates on `a8af9ad`: backend pytest **1211 passed** (one fewer than the previous head because the two contradicting tests became one), CDK vitest **92 passed**, `tsc --noEmit` clean.

Unless the next pass finds something substantive, I'd treat this as converged — the last two rounds have been nit-only and each one has been a consequence of the previous round's edit rather than anything pre-existing.

perrozzi:

Since my last pass the author only touched the three nits: the two contradictory `step_name` tests were merged into one config-derived loop, the `get_step_inference_config` docstring now mentions `step_name`, and the CDK timeout suite switched from `.find()` to `.filter()` and asserts every matching research Lambda. All three previous new findings are addressed; the older should-fix items are unchanged.

### Resolved
- `test_each_step_reports_the_name_the_config_declares` derives all three expectations from the config, removing the hardcoded-vs-parity contradiction.
- `get_step_inference_config` docstring now documents the `step_name` key.
- `processing-stack-consolidated.test.ts` asserts the timeout/memory ceiling for *every* Lambda whose handler matches `research_step_handler`.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`).
- `lambda/api/prompts/avatar-generation.json` is still absent from the diff while `test_avatar_image_model_lockstep.py` indexes `['image_model']['model_id']`/`['region']`.
- iOS ingestor sizing remains a band-aid; no partial-progress persistence.

### New findings
- Nit — `lib/stacks/processing-stack-consolidated.test.ts`: the memory assertion dropped its `as number` cast (`expect(fn.MemorySize).toBeGreaterThanOrEqual(1024)`) where `fn` is `Record<string, unknown>`; please re-confirm `tsc --noEmit` is still clean, since the previously reported clean run predates this edit.

Otherwise the delta is clean; no blockers.

---

perrozzi:
_(Answering **review round 4** on `a8af9ad`. No code change this round.)_

**Re-confirmed as asked: `tsc --noEmit` is clean on `a8af9ad`** — exit 0, zero output, run just now against the pushed head that contains `expect(fn.MemorySize).toBeGreaterThanOrEqual(1024)` at line 159. The previously reported clean run also postdated that edit, so nothing regressed in between.

It type-checks because `expect()` accepts `unknown`; `toBeGreaterThanOrEqual` constrains its own argument, not the actual value. The `as number` was doing no work, and removing it is consistent with this repo's preference for avoiding assertions that don't validate anything — the same reason the sibling lookup uses the `propsOf` guard rather than a cast.

Treating this as converged: rounds 3 and 4 were nit-only, and every finding since round 1 has been a consequence of the previous round's edit rather than anything pre-existing. The three long-standing items stay as dispositioned above — the TypeScript-parsing test helpers follow the existing `test_model_config.py` precedent, `avatar-generation.json` already carries the `image_model` block (it's absent from the diff because only the reader changed), and per-storefront progress for the iOS sweep is a behavioural change to the ingestor that belongs in its own PR.

Final state: 7 commits, `tsc` clean, backend pytest **1211**, CDK vitest **92**, deployed and live-verified (`test-api.sh` 32/32; the deployed research Lambda logs `max_tokens=9000, thinking_budget=5000` with no `FileNotFoundError`).

perrozzi:
_(New work, not a review response. Addressed in `a3be64b`, pushed right after this comment.)_

## The avatar model is now off the EOL model — and it wasn't an id swap

Reviewers rightly asked why a PR that flags a model reaching EOL in eight weeks wouldn't just replace it. It now does: **`amazon.nova-canvas-v1:0` → `stability.stable-image-core-v1:1`**. I enumerated against the account rather than guessing, which surfaced why this needed more than a constant change.

**us-east-1 has zero active text-to-image models.** Nova Canvas is the only generator offered there and it's legacy. Every other image model in the region is a Stability *editing* primitive — inpaint, upscale, remove-background, recolor, outpaint, style-transfer, control-sketch — all of which require an input image and cannot generate a portrait from a prompt.

| region | active text-to-image generators |
|---|---|
| us-east-1 | **0** (only LEGACY `amazon.nova-canvas-v1:0`) |
| us-west-2 | **3**: `stable-image-core-v1:1`, `stable-image-ultra-v1:1`, `sd3-5-large-v1:0` |
| eu-west-1 / ap-northeast-1 | 0 (legacy Nova Canvas only) |

So the avatar client now calls **us-west-2 cross-region**. That needed no new plumbing, which is the payoff from the earlier single-sourcing: `avatar.py` already builds its own regional client, and `imageModelArn()` derives the IAM grant from the same constants — so `cdk diff` swapped the ARN in all three roles by itself (us-east-1 Nova removed, us-west-2 Stability added, ×3).

I probe-invoked all three generators from this account before choosing, to confirm entitlement rather than assume it. Core is the cheapest and fastest and is ample for a headshot; ultra and sd3-5-large share the identical request shape, so switching is one constant.

**Response shape is compatible, request shape is not.** Stability returns `{images: [base64], seeds, finish_reasons}`, so the decode-and-upload path is untouched. But it takes `prompt` / `mode` / `aspect_ratio` / `output_format` / `seed`, not `taskType` / `textToImageParams` / `imageGenerationConfig` — a `ValidationException`, not a silent difference. The config block therefore swaps `width`/`height` for `aspect_ratio`/`output_format`, and the S3 key extension plus `ContentType` now derive from `output_format` so they cannot disagree with the bytes.

**A latent bug fell out of the rewrite.** The seed was `hash(persona_id)`, and Python randomises string hashing per process — so the "consistent seed per persona" comment was never true, and regenerating a persona produced a different avatar on every cold start. Now sha256-derived, with a test that recomputes the digest independently rather than just calling the function twice in one process.

### Verification

Deployed `VocApiStack` and checked both halves, since neither alone is sufficient:

- **Request shape + entitlement:** real invocation returned 1 image, 2,677,833 bytes, valid PNG magic header, `finish_reasons=[None]`, seed reproducible.
- **Deployed IAM** (the half local invocation can't prove, since my CLI creds are Admin while the Lambdas use narrow roles): `iam simulate-principal-policy` returns **allowed** for `ProjectsLambdaRole`, `PersonaGeneratorRole` and `PersonaImporterRole` against `arn:aws:bedrock:us-west-2::foundation-model/stability.stable-image-core-v1:1`.
- `scripts/test-api.sh` **32/32**.
- Gates: backend pytest **1215**, CDK vitest **92**, `tsc --noEmit` clean, zero new ruff findings.

One note for reviewers: the `image_model` block in `avatar-generation.json` **is** in this diff now, which also resolves the standing "that file is absent while the lockstep test indexes it" finding.

perrozzi:

Since the last pass the PR grew a new change: the avatar model was actually swapped (`amazon.nova-canvas-v1:0` → `stability.stable-image-core-v1:1`, `us-east-1` → `us-west-2`), the request body rewritten to Stability's `prompt/mode/aspect_ratio/output_format` shape, the S3 key/content type now derived from `output_format`, and `hash()` replaced by a sha256-derived seed. The lockstep test and `model-allowlist.ts` were updated in step. The previously-noted missing `avatar-generation.json` is now present; the other older items are unchanged.

### Resolved
- `lambda/api/prompts/avatar-generation.json` is now in the diff, so `test_avatar_image_model_lockstep.py`'s `['image_model']` indexing is grounded.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`).
- iOS ingestor sizing remains a band-aid; no partial-progress persistence.
- `processing-stack-consolidated.test.ts`: `expect(fn.MemorySize).toBeGreaterThanOrEqual(1024)` on a `Record<string, unknown>` — please re-confirm `tsc --noEmit`.

### New findings
- **Should-fix — verification evidence is now stale and contradicts the diff.** The description still says "The model id is intentionally NOT changed in this PR" and cites `cdk diff` showing *zero IAM statement changes* plus a live research invocation; with the model id, region, request body and response handling all changed, the IAM statement demonstrably changes and none of the new avatar path has been exercised against real Bedrock. Please re-run/re-state verification (model access enabled for Stability in `us-west-2`, one live avatar generated end-to-end) and update the description — this was exactly the risk that justified deferring the swap.
- **Should-fix — `lambda/shared/avatar.py`: the S3 key extension is now variable.** `avatars/{persona_id}.{output_format}` is only `.png` by config today, but any other place that reconstructs the avatar key (persona delete/cleanup, regeneration, frontend URL building) will silently miss a `.jpeg` object. Please confirm no other `avatars/{id}.png` construction exists, and add a regression test asserting key + `ContentType` follow `output_format` — the new coupling has no test.
- **Question — `lib/utils/model-allowlist.ts`: the "no active text-to-image model in us-east-1" claim.** `amazon.titan-image-generator-v2:0` is a text-to-image generator offered in `us-east-1`; if it is still active, an in-region model avoids a cross-region Bedrock call (and the residency question that raises for non-US deployments). Worth stating why it was ruled out, since the comment is the documented rationale future readers will trust.
- Nit — `_MAX_SEED`/`_stable_seed`: `% 4294967294` can yield `0`, which Stability treats as "random seed", defeating determinism for that one hash bucket; `1 + digest % (_MAX_SEED - 1)` sidesteps it.
- Nit — `avatar.py` maps `'jpg' → image/jpeg`, but Stability only accepts `png`/`jpeg`/`webp` as `output_format`, so `jpg` would fail validation before the mapping matters; either drop the special case or validate the configured format.

---

perrozzi:
_(Follow-up on avatar payload size, raised in review of the model swap. Addressed in `7dbe7dd`, pushed right after this comment.)_

Fair challenge on the 2.67 MB — it was oversized by roughly two orders of magnitude. Fixed by changing the encoding, with the resolution question answered separately below.

**Where avatars actually render**, which sets the budget:

| site | rendered size |
|---|---|
| `ChatBubbles` | `w-8 h-8` → **32 px** |
| `PersonaAvatar` sm / md / lg | 40 / 48 / 96 px, `max-w-[128px]` |
| `PersonaPDFContent` (export) | 80 px |

So the largest render is 128 CSS px, while the model emits **1536×1536**. Shipping a lossless encoding of a photograph that is always downscaled is simply the wrong trade.

**Measured on one seed and prompt:**

| format | bytes | dimensions |
|---|---|---|
| png | 2,677,833 | 1536×1536 |
| **jpeg** | **401,603** | 1536×1536 |
| webp | rejected — not a valid `output_format` for this model | — |

**6.7× smaller for identical pixels.** A project with 5-10 personas drops from roughly 13-26 MB of avatars to 2-4 MB. I inspected the JPEG at full size before committing: no artifacts worth caring about, and it's displayed at a twelfth of its linear resolution.

Nothing else needed changing, because the earlier commit already derives the S3 key extension and `ContentType` from `output_format` — objects become `.jpeg` with `image/jpeg` automatically. Existing `.png` avatars keep working: `avatar_url` is stored per persona and `get_avatar_cdn_url` takes the filename generically, so both extensions coexist.

**On reducing resolution specifically — not available at this layer.** `stable-image-core` exposes `aspect_ratio` only; there is no width/height parameter, so 1536×1536 is what it emits. Cutting the pixel count means downscaling after generation, which needs Pillow (a native ARM64 wheel) added to `processing-deps` plus a layer rebuild and a redeploy of every consumer. That would get an avatar to roughly 15-25 KB — a further ~20× — but it is a dependency and deployment change rather than a config one, so I have deliberately left it out. Happy to do it as its own PR if the extra reduction is wanted.

Tests: a guard that the default format stays lossy and that the shipped config agrees, plus one asserting the derived `ContentType` and S3 key cannot disagree with the bytes. I also de-hardcoded the `.png` literal in the existing success test so it derives from config and survives the next format change.

Backend pytest **1218 passed**.

perrozzi:

Since my last pass the only change is the avatar `output_format` flipping `png` → `jpeg` (config + `DEFAULT_OUTPUT_FORMAT` with a size-measurement comment), plus a `TestAvatarsAreNotShippedAsPng` guard and a rework of the `test_avatar.py` URL assertion. This addresses none of the previously-open items, and it makes one of them concrete rather than hypothetical.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`).
- iOS ingestor sizing remains a band-aid; no partial-progress persistence.
- `processing-stack-consolidated.test.ts`: `expect(fn.MemorySize).toBeGreaterThanOrEqual(1024)` on `Record<string, unknown>` — please re-confirm `tsc --noEmit`.
- PR description verification section is stale (still claims the model id is unchanged and cites zero IAM statement changes).
- `avatar.py`'s `'jpg' → image/jpeg` special case for a format the model rejects.

### New findings
- **Blocker — the S3 key extension now actually changes, so the previously-flagged key-reconstruction risk is live.** With `output_format: "jpeg"`, `lambda/shared/avatar.py` writes `avatars/{persona_id}.jpeg` instead of `.png`. Two consequences the diff doesn't address: (a) any other code that builds `avatars/{id}.png` (persona delete/cleanup, presign, frontend/PDF URL construction) now misses the object — this needs an actual grep result in the description, not an assumption; (b) regenerating an existing persona no longer overwrites its old `.png` object, so every pre-existing avatar leaks an orphan in `rawDataBucket` with no cleanup path. Previously this coupling was inert because the configured value was `png`; it no longer is.
- **Should-fix — no evidence the JPEG path was exercised end-to-end.** The 2.67 MB vs 401 KB measurement implies a live generation was run, but the description's verification section predates the model swap entirely. Please state explicitly that one persona was generated with `output_format: jpeg` against Stability in `us-west-2` and rendered in the UI and the PDF export (the PDF path in particular is where a hardcoded PNG assumption would surface).
- Nit — `test_default_format_is_lossy` accepts `'jpg'`, but the new comment in `avatar.py` says the model only takes `png`/`jpeg`; the test therefore green-lights a value that would fail Bedrock validation. Restrict it to `{'jpeg'}` (and consider validating the configured format at load time instead of the `'jpg'` mapping).
- Nit — `test_avatar.py` now derives its expectation from `get_image_model_config()`, i.e. from the same production code it's testing, so it can no longer detect a wrong extension. It's redundant with the lockstep test's literal `avatars/p1.jpeg`; either drop it or keep a literal.

---

perrozzi:
_(Answering **review rounds on `a3be64b` and `7dbe7dd`**. Addressed in `5ea6c9b`, pushed right after this comment. PR description rewritten — see below.)_

The blocker was right, and so were the nits. One factual claim was wrong; evidence below.

## Blocker — orphaned avatars on a format change: half right, and the right half is fixed

**(b) is real and now fixed.** The S3 key embeds the image format, so with `output_format: jpeg` a regenerated persona writes `avatars/{id}.jpeg` and leaves `avatars/{id}.png` behind. There is **no other avatar cleanup path anywhere in the codebase** — personas can be deleted without their avatar being removed at all — so those objects would leak permanently. You're also right that this was inert while the configured value was `png` and stopped being inert the moment it changed. `generate_persona_avatar` now best-effort deletes the persona's avatar under the other known extensions after a successful write. Deleting an absent key is a no-op in S3, so it costs nothing when there's nothing to clean, and a cleanup failure cannot fail generation.

**(a) does not apply, and here is the grep rather than an assumption.** The key is constructed in exactly one place:

| site | how it references avatars |
|---|---|
| `lambda/shared/avatar.py:214` | the only construction: `avatars/{persona_id}.{image_format}` |
| `lib/stacks/core-stack.ts:187` | `/avatars/*` CloudFront behavior — wildcard |
| `lib/stacks/api-stack.ts:437,511,625` | `grantReadWrite(..., 'avatars/*')` — wildcard |
| `avatar.py::get_avatar_cdn_url` | parses `parts[-1]`, extension-agnostic |
| `lambda/stream/src/context/project-context.ts:15` | doc comment only |

No presign path touches avatars, and no delete/cleanup path referenced them before this change. The remaining `.png` mentions were all stale doc comments, corrected in this commit.

## Nits — all three fixed

**Seed 0 meant "random".** Genuinely subtle and a real hole in the determinism I'd just added: `% 4294967294` can return 0, which Stability treats as choose-randomly. Now `1 + digest % (MAX - 1)`, with a test scanning 2000 ids for a zero.

**The `jpg → image/jpeg` mapping was dead code.** Rather than just deleting it, `get_image_model_config()` now validates `output_format` against what the model accepts (`png`, `jpeg`) and falls back to the default with a warning. The format names the S3 object as well as the request, so an unusable value previously produced a misleading key and `ContentType` *before* failing at invoke time. `jpg` can no longer reach the mapping, so the special case is gone.

**`test_default_format_is_lossy` accepted `'jpg'`** — restricted to `'jpeg'`, and it now also pins the supported set.

**`test_avatar.py` had become tautological** — correct, and worth calling out: deriving the expectation from `get_image_model_config()` took it from the same production code under test, so a wrong extension could never have failed it. That's the anti-pattern this repo's testing guidance explicitly warns about, and I walked into it while trying to avoid brittleness. Reverted to a literal.

## The Titan question — checked, and it does not hold

`amazon.titan-image-generator-v2:0` is **not offered**, so it cannot avoid the cross-region call:

```
amazon.titan-image-generator-v1     ResourceNotFoundException
amazon.titan-image-generator-v1:0   ResourceNotFoundException
amazon.titan-image-generator-v2:0   ResourceNotFoundException
amazon.nova-canvas-v1:0             LEGACY  IMAGE
```

`get-foundation-model` in us-east-1, all three variants. The only Titan models left in-region are *embeddings* — including `amazon.titan-embed-image-v1`, which is image-related but produces vectors, not images, and is the likely source of the confusion. My 16-region scan also never listed any Titan generator. I've left the rationale comment in `model-allowlist.ts` as-is because it is accurate, and the residency point you raise is worth noting for non-US deployments: there is currently no in-region option to choose.

## `tsc --noEmit` — re-confirmed twice now

Exit 0, zero bytes of output, run against the current head after the latest test edits.

## PR description rewritten

You were right that it was actively misleading — it still claimed the model id was unchanged and cited zero IAM statement changes, both false since the swap. Rewritten end to end: the migration is described as done, the live avatar generation and the deployed-role IAM simulation are stated as evidence, and I've added an explicit **"Known gaps in verification"** section recording that the avatar path has not been exercised through the persona UI or PDF export, and that the last commits are not deployed.

Gates on `5ea6c9b`: backend pytest **1223 passed**, CDK vitest **92**, `tsc` clean.

perrozzi:

Since the last pass the author addressed the JPEG fallout directly: `output_format` is now validated against `_SUPPORTED_OUTPUT_FORMATS` (unknown values coerced to the default with a warning), the old-extension objects are best-effort deleted via `_delete_superseded_avatars`, the seed is offset off 0, the `jpg`→`image/jpeg` special case is gone, the `test_avatar.py` assertion is back to a literal, and new tests cover cleanup/fallback/non-zero-seed. The PR description also now carries the JPEG/us-west-2 verification plus an explicit "known gaps" section.

### Resolved
- Blocker: orphaned `.png` avatars after the format change — `_delete_superseded_avatars` plus three tests (deletes old ext, never deletes the new one, failure doesn't fail generation).
- Nit: `test_default_format_is_lossy` now pins `'jpeg'` and the supported set, no longer green-lighting `'jpg'`.
- Nit: `'jpg' → image/jpeg` mapping removed, ContentType derived from the validated format.
- Nit: `test_avatar.py` back to a literal `avatars/p123.jpeg` instead of deriving from production code.
- Stale verification section — description now covers the model swap, the live JPEG generation and `simulate-principal-policy`.

### Still open
- Python tests parse TypeScript with regex (`_ts_const`, `_loader_ceiling`).
- iOS ingestor sizing remains a band-aid; no partial-progress persistence.
- JPEG path still not exercised through the persona UI or the PDF export (now disclosed as a known gap rather than closed).

### New findings
- **Should-fix — the cleanup adds a new IAM requirement to the avatar write path.** `_delete_superseded_avatars` needs `s3:DeleteObject` on `avatars/*`. `PersonaGeneratorRole` and `PersonaImporterRole` get `rawDataBucket.grantReadWrite(..., 'avatars/*')` (which includes `DeleteObject*`), but `ProjectsLambdaRole` also received `imageModelArn()` in `lib/stacks/api-stack.ts`, implying it generates avatars too — please confirm it has delete on `avatars/*`, otherwise every generation on that role emits three AccessDenied warnings and cleans nothing. Add it to the deployed-IAM simulation list you already ran.
- Nit — `_HISTORICAL_EXTENSIONS` includes `jpg` and `webp`, which this codebase has never written (only `.png`, now `.jpeg`). That's two guaranteed-miss `DeleteObject` calls per generation; trimming to `('png', 'jpeg')` keeps the cleanup honest about what could actually exist.
- Nit — deleting the old object immediately after `put_object` means that if the caller's DynamoDB update of `avatar_url` fails afterwards, the persona is left pointing at a key that no longer exists (previously it kept a stale-but-renderable `.png`). Acceptable, but worth a line in the docstring so the ordering isn't "fixed" later without thought.
- Nit — `test_seed_derives_from_a_process_independent_digest` now imports `_MAX_SEED` from `shared.avatar` to build its expectation, i.e. the same coupling you deliberately removed from `test_avatar.py`. Inline the literal `4294967294` to keep the oracle independent.

---

PR #230, opened 2026-08-03
fix(security): require signed URLs for /avatars/* and /prototypes/*

### LINKED ISSUE #229: security: /avatars/* and /prototypes/* served without authentication
## Summary

`/avatars/*` and `/prototypes/*` on the frontend CloudFront distribution are served **without any viewer-side authorization**. Anyone with a URL can fetch a persona avatar or a generated product prototype — no Cognito session, no signed URL, no cookie.

## Affected branches — the severe half is unreleased

The two paths differ here, and the difference sets the priority:

| path | `main` | `development` | sensitivity |
|---|---|---|---|
| `/prototypes/*` | **absent** — 0 occurrences in `core-stack.ts`, 0 prototype files in the tree | present (`core-stack.ts:217`) | **high** — PRD/PR-FAQ-derived product HTML |
| `/avatars/*` | **present** (`core-stack.ts:172`), same no-auth config | present (`core-stack.ts:187`) | low — AI-generated fictional portraits |

So **the material risk is `development`-only and has never been released.** It can be fixed before the dev→`main` promotion, which is the cheapest possible moment — no deployed prototype data to rotate or invalidate. The already-released half is the low-sensitivity one.

Two clarifications, since "development-only" invites a wrong reading:

- It is **branch** scope, not **environment** scope. `core-stack.ts` has **no dev/prod branching** — `grep -n "isDev\|isProd\|environment"` returns nothing, and `isDev` exists only in `api-stack.ts:104-110` where its sole use is the CORS `allowedOrigin`. Both behaviors are added unconditionally, so whichever branch you deploy, those paths are public *in that deployment* — including a production deployment cut from `development`.
- Avatars being on `main` means that half is already live wherever `main` is deployed. Low severity, but not "not yet shipped".

## Why it happens

Cognito is enforced at **API Gateway**, not at the CDN. The distribution's primary job is serving the SPA, which must be publicly reachable to render the login page at all. `/avatars/*` and `/prototypes/*` were added as additional cache behaviors on that same distribution (`core-stack.ts` ~185-225), so they inherited the public posture.

The S3 bucket itself is correctly locked down — all four public-access blocks enabled, reads only via Origin Access Control. This is specifically a CloudFront→viewer gap, not a bucket misconfiguration.

## Evidence

Live distribution config — no auth attachments on any behavior:

| behavior | CloudFront Functions | Lambda@Edge | signed URLs / cookies |
|---|---|---|---|
| `*` (default) | 0 | 0 | disabled |
| `/avatars/*` | 0 | 0 | disabled |
| `/prototypes/*` | 0 | 0 | disabled |

Unauthenticated fetches, no token and no cookie:

- `GET /avatars/persona_<id>.jpeg` → `200 image/jpeg`
- `GET /prototypes/proj_<id>/prototype_<id>.html` → `200 text/html`, 21,723 bytes of generated prototype

## Risk

**Prototypes are the material concern, not avatars.** Prototypes are generated from PRDs and PR-FAQs, so they can contain product strategy. Avatars are AI-generated fictional portraits — low sensitivity by comparison.

Two factors compound it:

- **Permanence.** Objects are served `Cache-Control: max-age=31536000, immutable` with no expiry, so any URL that is shared, logged, or leaked stays live indefinitely.
- **Key predictability.** Avatar keys are `avatars/persona_<YYYYMMDDHHMMSS>_<index>.<ext>` — a single second-resolution timestamp, enumerable with effort. Prototype keys need two timestamps (`prototypes/proj_<ts>/prototype_<ts>.html`), so brute force is impractical — but "hard to guess" is not access control.

## Options

1. **CloudFront signed cookies — recommended.** `` and `<iframe src>` cannot carry an `Authorization` header, so header-based JWT validation does not fit these paths. A short-lived cookie minted after Cognito login is sent automatically by the browser and covers both paths with one mechanism.
2. **Signed URLs.** Cheap for avatars specifically, since `lambda/shared/avatar.py::get_avatar_cdn_url` already centralises URL construction and could sign there. Works for prototypes too, as they are self-contained single-file HTML with no internal subresource links.
3. **Lambda@Edge JWT validation.** Heavier, and still cookie-based for subresources; a CloudFront Function cannot reach the Cognito JWKS endpoint.
4. **Triage:** accept for avatars, fix prototypes only. Defensible given the sensitivity difference.

## Trap for whoever picks this up

The `/prototypes/*` comment in `core-stack.ts` says the isolation that matters "comes from the frontend loading this via a cross-document `<iframe src=...>`". That is about **script** isolation — the model's inline JS cannot reach the parent app's DOM, storage or cookies — and says nothing about **access control**. It reads like reassurance that prototypes are protected.

## Notes

- Pre-existing; both behaviors predate #228, which only changed the file extension served on `/avatars/*`.
- Branch facts above were read from `git show origin/main:voc-datalake/lib/stacks/core-stack.ts` and the equivalent on `origin/development` (tips `d01b454` and `d005f02` at time of writing), not from a working tree.
- Found while verifying the avatar model migration in #228 (deployed avatar URLs were shared and turned out to be publicly fetchable).

perrozzi:
Closes #229.

`/avatars/*` and `/prototypes/*` were cache behaviors on the SPA's CloudFront distribution, which must stay publicly reachable to serve the login page. Cognito is enforced at API Gateway, never at the CDN, so anyone with a URL could fetch a persona avatar or — the material concern — a generated product prototype derived from PRDs and PR-FAQs. The bucket was never the problem: it is OAC-only with all four public-access blocks set. This was CloudFront → viewer.

Both behaviors now require a CloudFront trusted key group, and the already-authenticated API mints a short-lived signed URL per request.

## Why signed URLs and not signed cookies

The issue recommended cookies. They cannot be delivered: `cloudfront.net` and `execute-api.<region>.amazonaws.com` are both [Public Suffix List](https://publicsuffix.org/list/public_suffix_list.dat) entries, so no `Domain=` value lets an API Gateway response set a cookie for the distribution's host. URLs also need no cookie lifecycle in the SPA, and because `CACHING_OPTIMIZED` forwards no query strings the signature stays out of the cache key — edge caching is unaffected, and CloudFront strips the signature parameters before the origin fetch so OAC signing is untouched.

Presigned **S3** URLs were the other candidate and were rejected for `/prototypes/*` specifically: `prototypeHeadersPolicy` is a CloudFront response-headers policy, S3 cannot emit `Content-Security-Policy`, and that policy's `default-src 'none'` with no `connect-src` is the egress control that stops Bedrock-generated inline JS — running in a document holding PRD-derived content — from making outbound requests. Serving prototypes from S3 would have traded an authz gap for an egress gap.

## Keypair provisioning

A custom resource generates RSA-2048 at **deploy** time, writes the private half straight to Secrets Manager and returns only the public half, which CDK feeds into `cloudfront.PublicKey` + `KeyGroup` so CloudFormation keeps ownership of those. Two alternatives were closed off:

- generating at synth time would break the deterministic-synth guarantee `core-stack.test.ts` already asserts;
- a KMS asymmetric key cannot substitute — [`kms:Sign`](https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html) offers only SHA-256/384/512 RSA variants and CloudFront requires RSA-SHA1.

The handler is Node so keygen is stdlib, and it reuses an existing keypair on every update, since rotating would invalidate URLs already in browsers.

## Scope was larger than the issue suggested

**Avatar URLs are minted in three places, not one.** Besides the projects API, the **TypeScript chat-stream Lambda** emits `avatar_url` in the `persona_turn` SSE event that the SPA renders directly — signing only Python would have silently broken project-chat avatars. So there are two signers, and the TypeScript one is pinned byte-for-byte to botocore's canned-policy serialization by a fixture, so they cannot drift into signing different bytes. The third site, in `persona_importer`, turned out to be provably dead (it mutated a dict after its last use) and is removed.

**Prototype URLs move to read time.** The generator no longer persists an absolute `prototype_url`; the key is derivable, so the API signs one per request and *overwrites* anything stored. Rows written before this change hold an unsigned URL that now 403s.

Everything **fails closed**: with signing unavailable the helpers return `None`/`undefined` rather than a bare URL, and the SPA degrades to its gradient avatar.

## 403s were being laundered into 200s

`errorResponses` mapped 403 to `/index.html` with status **200**, and custom error responses are distribution-**wide** with no per-behavior scoping. Every denial on every path was therefore reported as success carrying the SPA shell. That is a large part of why this went unnoticed, and it would have been actively harmful here: a rejected prototype request would render the entire app inside the prototype iframe, and no test could tell allow from deny. Verified against the live distribution before changing it.

The 403 rule is dropped. SPA deep links keep working because CloudFront now has `s3:ListBucket` on the **website bucket only** (scoped by `AWS:SourceArn`), so a missing key answers 404 and the existing 404 rule serves the shell. The raw-data bucket deliberately does **not** get that grant — there, 403-for-a-missing-key is the right answer, since whether a given avatar or prototype key exists is not information an unauthenticated viewer is owed.

## Tested

Deployed and verified end to end against a live distribution:

| case | avatar | prototype |
|---|---|---|
| unsigned | 403 (was 200) | 403 (was 200, 21,723 bytes — the figure in #229) |
| signed | 200 `image/png` | 200 `text/html` |
| tampered signature | 403 | 403 |
| expired signature | 403 | 403 |

Plus: SPA deep links (`/`, `/settings`, `/projects`, `/a/b/c`) return 200 `index.html`; static assets unaffected; `GET /projects/{id}` returns signed avatars and prototypes that fetch 200 while the stale unsigned form 403s; legacy inline prototypes correctly get no URL; `POST /chat/stream` `persona_turn` avatars are signed and fetch 200. In a browser, both avatars render at 1536px with **zero console errors or CSP violations**, the prototype iframe loads through its signed `src`, and legacy prototypes fall back to `srcDoc`.

Three problems surfaced only by deploying, each now with a regression test:

1. **CoreStack ↔ ApiStack cycle.** `secret.grantRead()` on a CMK-encrypted secret adds a KMS *key-policy* statement naming the grantee; the key is in CoreStack and the roles in ApiStack. Now the ARN crosses as a string with an explicit `secretsmanager:GetSecretValue` statement — the pattern the existing `secretsArn` already uses.
2. **Node 24 rejected the custom resource at init** with "removed support for callback-based function handlers". The handler was async but took `deps` as a third parameter for test injection, and the runtime infers handler style from *arity*. Entry point is single-parameter now, with the injectable core on `onEvent`; a test asserts `handler.length <= 2`.
3. **Legacy inline prototypes.** The oldest prototypes keep their HTML in DynamoDB `content` with no S3 object, so minting a signed URL produced a 403 from S3 for a missing key *with a valid signature*, and because the frontend prefers `prototype_url` over `content` it would have replaced a working inline render with a broken iframe. Presence of `content` is the discriminator; `prototype_format` is `'html'` in both eras.

`core-stack.test.ts` had **no** CloudFront assertions before this, which is how an unprotected behavior shipped. It now asserts the property — every non-default behavior requires a signature — so a third private path is covered without anyone remembering to extend it, alongside the 403 mapping's absence, the `ListBucket` asymmetry, the prototype CSP, and that no key material reaches the template.

Gates: cdk 119, backend 1258, stream 257, frontend 1804 pass; `typecheck:all` clean. `lint:python` reports 545 ruff findings on this branch and 545 on `development` measured in a clean worktree — pre-existing, unchanged.

## Notes for review

- **`prototypeHeadersPolicy`'s comment is corrected here rather than in a separate PR.** #229 asked for that as its own small change; it is folded in because the same comment also needed the point that the policy is the *egress* control on model-authored JS, which is the part most easily dropped by accident. Happy to split it out if you'd rather.
- **The first commit is a `.gitignore` change** that stops local agent-memory notes (which contain AWS account ids) being committed alongside the tracked `.kiro/` directories. Unrelated to the fix — say the word and I'll drop it.
- **Not covered:** `scripts/test-api.sh` has no probe for these paths, so nothing automated guards the *live* behavior; the CDK tests only guard the template. A signed/unsigned probe pair would close that. Keypair rotation (delete the secret value and redeploy) and the `CDN_SIGNING_*` env contract are also undocumented in `docs/`. Both are follow-ups I have not made.


perrozzi:

This PR closes a real viewer-side authz gap: `/avatars/*` and `/prototypes/*` on the SPA distribution move behind a CloudFront trusted key group, a deploy-time custom resource mints the RSA keypair into Secrets Manager, and both the Python (`shared/cloudfront_signing.py`) and TypeScript (`stream/src/lib/cloudfront-signing.ts`) signers mint short-lived signed URLs at read time. The design reasoning is unusually well documented, fail-closed behavior is consistent, and the new CDK/Python/TS tests genuinely pin the regressions (including the cross-language canned-policy bytes and the legacy-inline-prototype discriminator). Two things need resolving before merge, mostly around packaging/dependency reach rather than the security logic itself.

### Blockers

- **Inline custom-resource code almost certainly exceeds the CloudFormation inline limit.** `voc-datalake/lib/stacks/core-stack.ts` uses `lambda.Code.fromInline(fs.readFileSync('.../cdn_signing_keys.js'))`, and `lambda/custom_resources/cdn_signing_keys.js` is 164 lines (~7–8 KB). `AWS::Lambda::Function` `Code.ZipFile` is capped at 4096 characters. Please confirm what was actually deployed here; if it did deploy, say how. The clean fix is `lambda.Code.fromAsset(path.join(__dirname, '../../lambda/custom_resources'))` with `handler: 'cdn_signing_keys.handler'` — that also removes the reason the handler must be plain JS, and with it the hand-written `.d.ts` plus the `!voc-datalake/lambda/custom_resources/*.d.ts` `.gitignore` negation.
- **`cryptography` reach is wider than the one layer that was updated.** `shared/prototypes.py` imports `sign_url` at module scope, so `jobs/document_generator/handler.py` and `jobs/persona_importer/handler.py` (via `api.projects`) and the projects API all now fail at *import* time if `cryptography` is missing — a broken import in `document_generator` kills prototype generation entirely, and nothing in this diff exercises that at deploy. Only `lambda/layers/processing-deps/requirements.txt` gained the dependency. Please confirm every function that transitively imports `shared.cloudfront_signing` has that layer attached, and consider moving the `cryptography`/`botocore.signers` imports inside `sign_url` (or importing `shared.cloudfront_signing` lazily in `shared/prototypes.py`) so a writer-only Lambda like `document_generator`, which only needs `prototype_s3_key`, does not take a crypto dependency at all.

### Should fix

- **The cross-language pin depends on an uncommitted generator.** `stream/src/lib/__fixtures__/cloudfront-signing.botocore.json` cites `.tmp/gen_signing_fixture.py`, and `.tmp/` is gitignored. The fixture is the only thing preventing the two signers from drifting, so its generator should be committed (e.g. as a pytest that asserts `botocore.signers.CloudFrontSigner`'s policy equals the fixture string) — otherwise nobody can regenerate or audit it.
- **The TypeScript signer swallows every failure silently.** `loadPrivateKeyPem` and `signCloudFrontUrl` in `stream/src/lib/cloudfront-signing.ts` both use bare `catch { return undefined; }`. Failing closed is right, but the Python side logs (`logger.exception`, plus the `is_configured` error log); here an IAM denial or malformed PEM manifests only as avatars quietly disappearing from `persona_turn`. Add a log on the error paths.
- **Confirm `_with_signed_prototype_url` covers every read path.** It is wired only into `get_project` in `lambda/api/projects.py`. If any other endpoint returns prototype documents (a single-document GET, a documents list, an export) it will still emit the persisted unsigned `prototype_url`, which is exactly the stale link this change is trying to stop circulating. Same question for `avatar_url` outside the `PERSONA#` branch of `get_project`.
- **Nothing handles expiry on the client.** URLs are valid for 1h and there are no frontend changes in this diff; a TanStack Query cache entry (or an open tab) older than the TTL yields a 403 image/iframe with no refetch trigger, and the 403→200 laundering that used to mask it is now correctly gone. At minimum, set the relevant query `staleTime` below `CDN_SIGNED_URL_TTL_SECONDS` and document the coupling.
- **`resolveAvatarUrl`'s doc comment contradicts its code** (`stream/src/context/project-context.ts`): "A value that is already an https URL is passed through unsigned only if it cannot be signed" — the non-`s3://` branch returns `signCloudFrontUrl(url)`, which yields `undefined`, never a pass-through.

### Nits

- `shared/cloudfront_signing.py`: `sign_url`'s docstring references `DEFAULT_TTL_SECONDS`; the constant is `FALLBACK_TTL_SECONDS`.
- `lib/stacks/api-stack.ts` adds `import * as secretsmanager` but the props are plain ARN strings — looks unused.
- `lambda/conftest.py` places `import json` / `patch` / `pytest` mid-file (ruff E402), and the session-scoped `cdn_signing_keypair` now makes `cryptography` a hard test-env dependency for the whole suite.
- With `enableLogging: true` on the distribution, `Signature`/`Key-Pair-Id` land in the access-logs bucket (`cs-uri-query`). Bounded by the TTL, but worth a line in the module docstring alongside the other trade-offs.
- The `.gitignore` / agent-memory commit is unrelated to the fix, as you note — splitting it would make this diff easier to bisect.

---

perrozzi:
Thanks — the design points landed, and I've acted on all of them in `f5676f5`. Both **blockers turned out to rest on false premises**, though, so evidence first.

## Blocker 1: the 4096-character inline cap is not enforced

You asked what actually deployed. It deployed as inline, at **6,868 characters**:

```
$ aws cloudformation get-template --stack-name VocCoreStack ...
resource: CdnSigningKeysLambdaDE260F93
has ZipFile: True
ZipFile length (chars): 6868
runtime: nodejs24.x

$ aws lambda get-function-configuration --function-name voc-cdn-signing-keys-...
{"CodeSize": 3057, "Runtime": "nodejs24.x", "State": "Active", "LastUpdateStatus": "Successful"}
```

CloudFormation accepted it and the function ran (it generated the keypair). aws-cdk-lib 2.261.0 has no check either — `grep -c 4096 node_modules/aws-cdk-lib/aws-lambda/lib/code.js` is `0`, and the only validation in `fromInline` is "Lambda inline code cannot be empty". The reason nobody had noticed is that the existing `admin_bootstrap.py` is 4,003 bytes, just under the number everyone quotes.

**I still made the change**, because your underlying point is right: relying on undocumented tolerance in a repo other people deploy into their own accounts is a bad bet, and it left a trap where adding a comment to the handler creeps toward a limit with nothing to catch it. Now `fromAsset` with excludes, and the deployed zip contains exactly one file:

```
zip contents: cdn_signing_keys.js
```

One correction to the reasoning: **`fromAsset` does not retire the `.d.ts` or the `.gitignore` negation.** Those exist so the CDK test can type-import the JS, which is orthogonal to inline-vs-asset. Only making the handler TypeScript — i.e. `NodejsFunction`/esbuild — would, and that seemed like more machinery than a 150-line custom resource warrants. Happy to go further if you disagree.

## Blocker 2: `cryptography` does reach every importer

Verified rather than reasoned. All four Lambdas that transitively import `shared.cloudfront_signing` — `projects-api`, `job-document-generator`, `job-persona-generator`, `job-persona-importer` — share `ApiDepsLayer:7`, and that layer contains cryptography (190 zip entries plus `_rust.abi3.so`). More directly, invoking the document generator with a bogus payload distinguishes an import failure from a handler failure:

```
$ aws lambda invoke --function-name voc-job-document-generator-... --payload '{"probe":"import-check"}'
{"errorMessage": "'project_id'", "errorType": "KeyError", ...
   File "/opt/python/aws_lambda_powertools/logging/logger.py", line 554, in decorate
```

A `KeyError` raised *inside* the handler proves the module and all its imports loaded.

**Your suggestion is implemented anyway**, because the fragility is real even if it isn't currently biting: `document_generator` uses this module only for `prototype_s3_key`, so `shared/prototypes.py` now imports `sign_url` inside `prototype_signed_url`. A writer that never signs no longer takes a crypto dependency at cold start.

## Should-fix items

- **Fixture generator committed** as `lambda/shared/test/test_cloudfront_signing_fixture.py`. It recomputes the canned policy from `CloudFrontSigner.build_policy` and fails if the committed fixture disagrees, so the pin is reproducible and auditable, and a botocore change to the serialization is caught there rather than by a 403 in a browser. It also asserts the fixture carries no key material.
- **TS signer now logs**, three `console.warn` paths (missing config, unusable secret, signing failure), error `name` only, never the payload. You were right that failing closed silently just made avatars vanish with nothing to grep for.
- **`resolveAvatarUrl`'s comment corrected** — it did contradict the code.
- **Nits fixed:** `DEFAULT_TTL_SECONDS` → the real constant in the docstring; unused `secretsmanager` import dropped; and I documented the access-log point you raised (`enableLogging` puts `Signature` into `cs-uri-query`; private bucket and TTL-bounded, but replayable by anyone who can read those logs).

Two I checked and left alone, with reasoning:

- **Read paths are covered.** `prototype_url` is attached only in `get_project`, and that is the only documents-returning endpoint — `/projects/prioritization` returns scores from the aggregates table. The Prioritization page's prototypes come from `prioritizationUtils.ts` consuming `detail.documents`, i.e. the same `getProject` payload. Same for `avatar_url`.
- **Client expiry is self-healing.** `frontend/src/App.tsx` sets a global TanStack `staleTime: 30000` against the 1h TTL, and refetch-on-window-focus is on, so URLs refresh long before expiry. I documented the coupling in the module docstring rather than changing code, since the failure mode only appears if the TTL drops below the staleTime.

## Re-verified after these changes

The one that mattered, since changing the handler's code triggers a real CloudFormation Update of the custom resource: **the keypair was not rotated.** Key pair id `K110AE9ETM1F8F` and public-key sha256 `caa85e178f67a9c3`, identical before and after — the idempotency guarantee exercised for real rather than only in unit tests. Also confirmed: template now uses an S3 asset, the CR still starts, the document generator still imports, unsigned requests still 403 on both paths, SPA deep links still 200, and every persona with a real avatar plus both S3-backed prototypes fetch 200 signed while legacy inline prototypes correctly get no URL.

Gates: cdk 119, backend 1264, stream 257 pass; typecheck clean; ruff 545, unchanged from `development`.

Still open for a maintainer call: whether to split the `/prototypes/*` doc-comment into its own PR as #229 asked (it's folded in because the same comment also needed the egress-control point), and whether to drop the unrelated `.gitignore` commit.

perrozzi:

Since the last review the author switched the custom resource from `Code.fromInline` to `Code.fromAsset` (blocker 1 resolved), made `shared/cloudfront_signing` a lazy import inside `prototype_signed_url` (partially addresses blocker 2), committed a pytest that recomputes the cross-language fixture from botocore, added `console.warn` on every fail-closed path in the TS signer, documented the access-log/`staleTime` couplings, fixed the `resolveAvatarUrl` doc comment and the `DEFAULT_TTL_SECONDS`/unused-import nits. No blockers remain; what's left is stale justification comments created by the `fromInline` → `fromAsset` switch.

### Resolved
- Inline 4096-char limit: `core-stack.ts` now uses `Code.fromAsset(.../custom_resources, { exclude: [...] })` with `handler: 'cdn_signing_keys.handler'`.
- `document_generator`'s crypto dependency: `shared/prototypes.py` imports `sign_url` lazily, so the writer path no longer needs `cryptography`.
- Fixture generator is now committed and auditable as `lambda/shared/test/test_cloudfront_signing_fixture.py` (also catches botocore serialization changes).
- TS signer no longer swallows failures silently (`loadPrivateKeyPem` / `signCloudFrontUrl` warn, without logging payloads).
- `resolveAvatarUrl` doc comment now matches behavior; TTL docstring constant corrected; unused `secretsmanager` import dropped from `api-stack.ts`.
- Client-expiry coupling documented (30s global `staleTime` vs 1h TTL) in `shared/cloudfront_signing.py`.

### Still open
- `shared/avatar.py` still imports `sign_url` at module scope, so `api.projects` and `jobs/persona_importer` fail at import without the `cryptography` layer — the in-code "verified" note is an assertion, not an artifact; please confirm the layer is attached to `persona_importer` in `api-stack.ts`.
- `_with_signed_prototype_url` is still wired only into `get_project`; other endpoints returning prototype documents (if any) would still emit the persisted unsigned `prototype_url`.

### New findings
- Stale justification comments now that the handler is an asset, not inline. `.gitignore` ("the handler cannot be TypeScript: inline Lambda code is never bundled"), `lambda/custom_resources/cdn_signing_keys.d.ts` ("core-stack.ts inlines it with `lambda.Code.fromInline`"), `cdn_signing_keys.js` header ("inlines with zero dependencies"), and `lib/stacks/cdn-signing-keys.test.ts` all still claim `fromInline`. With an asset, the "cannot be TypeScript" rationale no longer holds — either correct the comments or drop the `.d.ts` + `.gitignore` negation and make the handler TS.
- `lib/stacks/core-stack.test.ts`: the "never puts signing key MATERIAL in the template" test comment explains the narrow PEM-header search as avoiding a match on "the inlined handler's OWN SOURCE" — no longer true; the assertion still passes but the reasoning misleads.
- `lib/stacks/core-stack.ts`: `exclude: ['*.py', ...]` under the default GLOB ignore mode matches top-level only; nested Python outside `test/` (if any is ever added to `lambda/custom_resources/`) would be packaged. Consider `'**/*.py'`.
- Minor: verify `import * as fs from 'fs'` in `core-stack.ts` is still referenced after removing the `readFileSync` call, or `noUnusedLocals`/lint will flag it.

---

perrozzi:

Since my last review the author added two tests only: a subprocess-based guard in `lambda/shared/test/test_prototypes.py` that `import shared.prototypes` does not pull in `cryptography`, and a CDK assertion in `lib/stacks/core-stack.test.ts` that the signing-key handler is packaged as an asset (`S3Bucket`, no `ZipFile`). No previously reported findings were addressed by these changes; they add regression coverage for the two fixes made in the prior round.

### Still open
- `shared/avatar.py` still imports `sign_url` at module scope — `jobs/persona_importer` import path unconfirmed against the `cryptography` layer.
- `_with_signed_prototype_url` is wired only into `get_project`.
- Stale `fromInline` rationale in `.gitignore`, `lambda/custom_resources/cdn_signing_keys.d.ts`, `cdn_signing_keys.js` header, and `lib/stacks/cdn-signing-keys.test.ts`; plus the "inlined handler's OWN SOURCE" comment on the key-material test in `core-stack.test.ts`.
- `exclude: ['*.py', ...]` matches top-level only (`'**/*.py'`); `import * as fs` usage in `core-stack.ts` unverified.

### New findings
- Should fix — `lambda/shared/test/test_prototypes.py`: the new subprocess test hardcodes `cwd='lambda'`, so it only passes when pytest is invoked from `voc-datalake/`. Run from the repo root or from inside `lambda/` it fails with `FileNotFoundError` on the cwd, not on the property under test. Derive it instead, e.g. `cwd=pathlib.Path(__file__).resolve().parents[2]` (the `lambda` dir), which is the same anchoring `test_cloudfront_signing_fixture.py` already uses for `FIXTURE_PATH`.
- Nit — `lib/stacks/core-stack.test.ts`: the new test's comment ("the inline form DID deploy at ~6.9KB … aws-cdk-lib has no 4096-character check") now directly contradicts the surviving `fromInline` comments listed above, including the `.gitignore`/`.d.ts` claim that the handler "cannot be TypeScript". Worth resolving those in one pass so the asset switch has a single story.

---

perrozzi:
Both follow-up rounds addressed in `ec1f461` (pushing right after this). The stale-comment finding was the best catch of the three rounds — I changed the mechanism and left four places still justifying themselves by the old one.

## Fixed

**Stale `fromInline` rationale.** All four corrected: `.gitignore`, `cdn_signing_keys.d.ts`, the `cdn_signing_keys.js` header, and `cdn-signing-keys.test.ts`. The "cannot be TypeScript" claim did die with the switch — the handler now stays plain CommonJS for the narrower reason that an asset copies files verbatim, and each note records that moving to TypeScript would need a `NodejsFunction` and would retire both the `.d.ts` and the `.gitignore` negation. I've left it as JS deliberately: that's more build machinery than a ~150-line custom resource warrants. Say the word if you'd rather have the TS conversion and I'll do it.

Also fixed the key-material test's comment in `core-stack.test.ts`, which justified its narrow PEM-header search by the handler being inlined into the template. No longer why — though the precise assertion is still correct, and the comment now says so plus notes it stays honest if anyone reverts to inline.

**Real bug in the test I added in `50beb71`** — good catch, and worse than described: hardcoding `cwd='lambda'` meant the test reported the *wrong failure*, a `FileNotFoundError` on the cwd rather than anything about the import graph. Now anchored on `__file__` exactly as you suggested, matching `test_cloudfront_signing_fixture.py`. Verified by running it from inside `lambda/`, the cwd that broke it.

**`shared/avatar.py` now lazy too.** You were right that I'd left the asymmetry: `prototypes.py` got the lazy import and `avatar.py` didn't. Only `get_avatar_cdn_url` signs; the rest of that module is the avatar *writer* path used by the persona-generator and persona-importer jobs, which never mint a URL. Both are lazy now.

On the layer question underneath it — the note in code was indeed an assertion, so here's the artifact. `voc-job-persona-importer` carries `ApiDepsLayer:7`, the same layer as the other three importers, and that layer contains cryptography:

```
voc-projects-api            layers: ApiDepsLayer07D4211E:7
voc-job-document-generator  layers: ApiDepsLayer07D4211E:7
voc-job-persona-generator   layers: ApiDepsLayer07D4211E:7
voc-job-persona-importer    layers: ApiDepsLayer07D4211E:7

layer zip: 13,461,847 bytes
cryptography entries: 190
cryptography native so (_rust.abi3.so): 1
```

With both modules lazy this is no longer load-bearing either way.

## Checked and left alone, with evidence

**`exclude: ['*.py']` does match at any depth.** The suggestion to use `'**/*.py'` rests on a false premise — the default `IgnoreMode.GLOB` uses .gitignore semantics, where a pattern with no slash matches by basename anywhere in the tree. I staged a nested `.py` into a throwaway asset to check rather than assume:

```
staged files: ["handler.js"]
top-level top.py excluded: true
NESTED nested/deep.py excluded: true
```

I documented that in the comment instead of changing a working pattern.

**`import * as fs` is still used** — `core-stack.ts:696`, by the admin-bootstrap handler, which is still `fromInline` (4,003 bytes, under the ceiling). Typecheck and lint are clean.

**Read paths** — same answer as my previous reply, restated since it was re-raised: `prototype_url` is attached only in `get_project`, which is the only documents-returning endpoint. `/projects/prioritization` returns scores from the aggregates table; the Prioritization page's prototypes come from `prioritizationUtils.ts` consuming `detail.documents`, i.e. that same `getProject` payload. There is no other endpoint to wire.

Gates: cdk 120, backend 1266, stream 257 pass; typecheck and stream lint clean; ruff 545, unchanged from `development`.

Still open for a maintainer call, unchanged: whether to split the `/prototypes/*` doc-comment into its own PR as #229 asked, and whether to drop the unrelated `.gitignore` commit. Two follow-ups I have deliberately not done in this PR: `docs/` has no signed-URL or keypair-rotation section, and `scripts/test-api.sh` has no signed/unsigned probe, so nothing automated guards the live behaviour.

perrozzi:

Since my last review the author pushed comment/doc corrections only, plus two functional touches: `shared/avatar.py` now imports `sign_url` lazily inside `get_avatar_cdn_url`, and the subprocess test's `cwd` is anchored on `__file__` instead of `'lambda'`. Most of my previous findings are addressed.

### Resolved
- `shared/avatar.py` no longer imports `shared.cloudfront_signing` at module scope, so the persona-importer/generator writer paths don't pull in `cryptography` at cold start (and an `ImportError` still lands in the existing `except` → fails closed to `None`).
- The stale `fromInline` rationale is corrected consistently in `.gitignore`, `cdn_signing_keys.d.ts`, the `cdn_signing_keys.js` header, `cdn-signing-keys.test.ts`, and the key-material test comment in `core-stack.test.ts`.
- `test_prototypes.py` now derives `cwd` from `pathlib.Path(__file__).resolve().parents[2]`, matching `test_cloudfront_signing_fixture.py`; the test passes from any invocation directory.
- The `exclude` glob concern is answered in `core-stack.ts` — `IgnoreMode.GLOB` does pass `matchBase: true`, so basename patterns do match at any depth; the added comment documents that correctly.

### Still open
- `_with_signed_prototype_url` is wired only into `get_project`.

### New findings
- Nit — `voc-datalake/lambda/shared/avatar.py`: the fix that motivated the lazy import (writer jobs must not need `cryptography`) has no guard, while its sibling `shared/prototypes.py` got `test_importing_the_module_does_not_pull_in_cryptography`. A one-line parallel subprocess assertion on `import shared.avatar` would stop someone hoisting this import back.
- Nit — same file: the two added lines inside `get_avatar_cdn_url` include a whitespace-only line, which will add a ruff `W293` to the count you baselined at 545/545.

---

perrozzi:
Your first new finding is already done — `15806b1`, pushed right after this. Nice convergence: I'd reached the same conclusion independently, that making `avatar.py` lazy without a guard left the asymmetry where only the prototypes half was protected.

Rather than copy ~15 lines of subprocess plumbing into a second test file, the check is now an `imports_cryptography` fixture in `lambda/conftest.py`, used by both `test_avatar.py` and `test_prototypes.py`. It stays a subprocess for the reason that matters: `cryptography` is already in `sys.modules` for most of this suite — the signing tests and the `cdn_signing_keypair` fixture both use it — so an in-process check would pass regardless of the actual import graph. cwd anchored on the conftest, never the caller's.

## On the W293 nit — checked, and it does not apply

`W293` is not enabled in this repo's ruff configuration. Zero findings across the entire tree:

```
$ ruff check lambda plugins --output-format=concise | grep -c W293
0
```

Which matches the measurement: 545 before, 545 after.

The blank lines are also the local convention rather than my deviation — `avatar.py` has **33** whitespace-only lines out of 372, including the two immediately above my addition (lines 350 and 355, both pre-existing). Reformatting only mine would make it inconsistent with its neighbours for a rule that isn't on, so I've left it.

For the record, that's the fifth factual claim across these rounds that didn't hold up — after the 4096-character inline cap, the `cryptography` reach, `'*.py'` matching top-level only, and the unused `fs` import. The design points underneath have been consistently good and I've acted on all of them; it's the supporting specifics that need checking before acting.

## Still open

`_with_signed_prototype_url` being wired only into `get_project` has now been raised in all four rounds. Restating once more: that is correct and complete, because `get_project` is the **only** endpoint that returns documents. `/projects/prioritization` returns scores from the aggregates table, and the Prioritization page's prototypes come from `prioritizationUtils.ts` consuming `detail.documents` — the same `getProject` payload. There is no second read path to wire.

Gates: backend 1267, cdk 120, stream 257; typecheck and stream lint clean; ruff 545, unchanged from `development`.

Unchanged for a maintainer: whether to split the `/prototypes/*` doc-comment into its own PR as #229 asked, and whether to drop the unrelated `.gitignore` commit. And two things deliberately not in this PR: `docs/` has no signed-URL or keypair-rotation section, and `scripts/test-api.sh` has no signed/unsigned probe, so nothing automated guards the live behaviour.

perrozzi:

Since my last review the only change is test plumbing: a new session-scoped `imports_cryptography` predicate fixture in `voc-datalake/lambda/conftest.py`, a matching guard test for `shared.avatar`, and `test_prototypes.py` refactored onto the shared fixture (dropping its now-unused `pathlib` import). This addresses the first of my two previous nits; the second is untouched.

### Resolved
- `shared.avatar`'s lazy `cryptography` import now has a regression guard (`TestNoCryptoDependencyForWriters` in `shared/test/test_avatar.py`), mirroring the prototypes one, with the subprocess mechanics deduplicated into a single cwd-anchored fixture.

### Still open
- `_with_signed_prototype_url` is wired only into `get_project`.
- Whitespace-only line in `get_avatar_cdn_url` (`voc-datalake/lambda/shared/avatar.py`) — a new ruff `W293` against the 545/545 baseline you cited.

### New findings
- Nit — `voc-datalake/lambda/conftest.py`: `subprocess.run(..., check=True)` means a module that fails to import at all surfaces as an opaque `CalledProcessError` rather than the assertion message; passing `check=False` and asserting on `returncode`/`stderr` would keep the failure legible. Not blocking.

---

perrozzi:
Both nits addressed in `6eded47` (pushing right after this). I think this is done from my side.

**`check=False`** — good catch, and it's the same class of problem as the `cwd` bug you found two rounds ago: the test failing for a reason that looks like something else. The fixture now raises an `AssertionError` carrying the child's stderr, so a broken import reports `ModuleNotFoundError: No module named ...` instead of an exit code.

**Whitespace** — cleaned. Your stated impact was wrong, though, and worth pinning down since it came up twice: `W293` is **not** in this repo's ruff select set, so it never counted against the baseline, which sat at 545 before and after every commit in this PR. It only appears if you ask for it explicitly, and then `avatar.py` has 33 of them:

```
$ ruff check lambda plugins | grep -c W293
0
$ ruff check lambda/shared/avatar.py --select W293
Found 33 errors        # → 32 after removing mine
```

So the count claim was wrong but the observation was fair, and the line cost nothing to clean. The other 32 are pre-existing and left alone.

## `_with_signed_prototype_url` — fifth and final restatement

This has now been listed as "still open" in all five rounds. It is not open; it is complete. `get_project` is the **only** endpoint in the codebase that returns documents:

- `/projects/prioritization` returns prioritization scores from the aggregates table, not documents.
- The Prioritization page's prototypes reach the UI through `prioritizationUtils.ts` consuming `detail.documents` — i.e. the `getProject` payload that is already signed.
- `grep -rn "prototype_url" lambda/api lambda/shared` matches only `projects.py`.

There is no second read path to wire. If you have a specific endpoint in mind, name it and I'll wire it; otherwise I'd ask that this drop off the list.

## State

Deployed and live-verified on a real distribution: unsigned/tampered/expired all 403, signed 200, SPA routing intact, avatars and prototypes rendering in a browser with no CSP errors, and the keypair confirmed not to rotate across a real CloudFormation Update of the custom resource.

Gates: backend 1267, cdk 120, stream 257, frontend 1804; typecheck and stream lint clean; ruff 545, unchanged from `development`.

From my side this is ready. Outstanding items are all maintainer calls rather than work: whether to split the `/prototypes/*` doc-comment into its own PR as #229 asked, and whether to drop the unrelated `.gitignore` commit. Two things I deliberately kept out of scope and would suggest as follow-up issues rather than growing this PR further: `docs/` has no signed-URL or keypair-rotation section, and `scripts/test-api.sh` has no signed/unsigned probe — so the template is guarded by CDK tests but nothing automated guards the live behaviour.

perrozzi:

Since my last review the only changes are the two nits: `conftest.py`'s `imports_cryptography` now uses `check=False` and raises an `AssertionError` carrying the child's stderr, and the whitespace-only line in `get_avatar_cdn_url` is gone. Both previous nits are addressed; one earlier Should-fix item remains.

### Resolved
- `voc-datalake/lambda/conftest.py`: an import failure in the subprocess now surfaces as a readable assertion with stderr instead of an opaque `CalledProcessError`.
- `voc-datalake/lambda/shared/avatar.py`: whitespace-only line removed, so no new ruff `W293` against the 545/545 baseline.

### Still open
- `_with_signed_prototype_url` is wired only into `get_project`.

---

PR #268, opened 2026-08-06
chore(workshop): publish VocWebSearchStack with the workshop artifacts

perrozzi:
## Problem

Web search deploys by default since #206, so a plain `cdk synth --all` emits `VocProcessingStack` and `VocApiStack` templates containing `Fn::ImportValue` on the gateway's exports:

```
VocWebSearchStack:ExportsOutputFnGetAttWebSearchGatewayGatewayArnBA97E0DC
VocWebSearchStack:ExportsOutputFnGetAttWebSearchGatewayGatewayUrlE01706EF
```

But `generate:workshop-artifacts` only converted five stacks, so `VocWebSearchStack` was never published alongside them.

Deploying those templates into a Workshop Studio event fails at CREATE:

```
No export named VocWebSearchStack:ExportsOutputFnGetAttWebSearchGatewayGatewayArnBA97E0DC found
```

`VocProcessingStack` and `VocApiStack` both fail in about 30 seconds with **zero resources created**. Because `VocApiStack` owns the frontend `BucketDeployment`, the event is also left with no API Gateway and an empty frontend bucket, so nothing is reachable. The imports are consumed by `ResearchStepLambda` (processing) and `ChatStreamApi` (api) plus their IAM policies.

This was hit on a real workshop test event: 2 of 5 stacks came up, 3 failed.

## Change

One line: add `VocWebSearchStack` to the `convert-template.mjs` stack list in `generate:workshop-artifacts`, so the exporting stack is published with its consumers. Listed first to mirror the `--all` deploy order, where it is stack `[1/6]`.

## Why no special handling is needed

- `addAssetParameters()` runs unconditionally in `convert-template.mjs`, so the converted template declares the usual `AssetBucket`/`AssetPrefix` parameters even though this stack has no Lambda assets. Its workshop `contentspec.yaml` entry is therefore uniform with the other five.
- Export names are emitted from the **CDK stack id**, so they match the consumers' imports regardless of the deployed CloudFormation stack name (Workshop Studio lowercases stack names).
- Every other transform in the converter is guarded (`continue` / optional chaining / `if (policyKey)`), so an asset-less stack passes through cleanly.

## Verification

Synthesized and converted the stack (`AWS_REGION=us-east-1`):

- Resources: `AWS::IAM::Role`, `AWS::BedrockAgentCore::Gateway`, `AWS::IAM::Policy`, `AWS::BedrockAgentCore::GatewayTarget` — 4 total, no Lambda.
- `Parameters`: `[AssetBucket, AssetPrefix]`.
- Export names match `VocProcessingStack`'s imports byte for byte.
- No `Custom::CrossRegionExportWriter` and no SSM cross-region reader when the app region is us-east-1.

`package.json` remains valid JSON (23 scripts). No application code, IaC or IAM is touched, so runtime behaviour and every non-workshop `cdk deploy` are unchanged.

## Note for whoever publishes workshop artifacts

Pin the synth region to us-east-1. `VocWebSearchStack` is hard-pinned to us-east-1, so when the ambient region differs, CDK switches to cross-region references: the stack grows a `Custom::CrossRegionExportWriter` plus a Lambda, and the consumers read the gateway through SSM instead of `Fn::ImportValue` — a different wiring shape than the published templates expect.

Note that the CDK CLI **overwrites** `CDK_DEFAULT_REGION` for the app subprocess, so setting that variable has no effect. Use `AWS_REGION` / `AWS_DEFAULT_REGION` or `--profile`. This is easy to hit silently, since the region otherwise comes from the `[default]` profile in `~/.aws/config`, and `contentspec.yaml` lists us-west-2 as a deployable region.

Pinning the region inside the npm script was deliberately left out of this PR to keep it to the one-line fix; happy to add it if preferred.

## Out of scope

A workshop test event separately showed that the event account's private marketplace can refuse `CreateFoundationModelAgreement` for the newest models (Sonnet 5 and Opus 5 were denied; Sonnet 4.6 and Haiku 4.5 succeeded). That is an event-account entitlement issue, independent of this change, and is not addressed here.

### DIFF
--- a/voc-datalake/package.json
+++ b/voc-datalake/package.json
@@ -24,7 +24,7 @@
     "generate:menu": "ts-node scripts/generate-menu-config.ts",
     "generate:config": "npm run generate:manifests && npm run generate:menu",
     "generate:integrity": "ts-node scripts/generate-integrity.ts",
-    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1 npx cdk synth --all && node scripts/convert-template.mjs VocWebSearchStack VocCoreStack VocIngestionStack VocProcessingStack VocApiStack BedrockAccessStack",
+    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && npx cdk synth --all && node scripts/convert-template.mjs VocWebSearchStack VocCoreStack VocIngestionStack VocProcessingStack VocApiStack BedrockAccessStack",
     "validate:plugins": "ts-node scripts/validate-plugins.ts",
     "prebuild:frontend": "npm run generate:config"
   },
--- a/voc-datalake/package.json
+++ b/voc-datalake/package.json
@@ -24,7 +24,7 @@
     "generate:menu": "ts-node scripts/generate-menu-config.ts",
     "generate:config": "npm run generate:manifests && npm run generate:menu",
     "generate:integrity": "ts-node scripts/generate-integrity.ts",
-    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && npx cdk synth --all && node scripts/convert-template.mjs VocCoreStack VocIngestionStack  VocProcessingStack VocApiStack BedrockAccessStack",
+    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && npx cdk synth --all && node scripts/convert-template.mjs VocWebSearchStack VocCoreStack VocIngestionStack  VocProcessingStack VocApiStack BedrockAccessStack",
     "validate:plugins": "ts-node scripts/validate-plugins.ts",
     "prebuild:frontend": "npm run generate:config"
   },
--- a/voc-datalake/package.json
+++ b/voc-datalake/package.json
@@ -24,7 +24,7 @@
     "generate:menu": "ts-node scripts/generate-menu-config.ts",
     "generate:config": "npm run generate:manifests && npm run generate:menu",
     "generate:integrity": "ts-node scripts/generate-integrity.ts",
-    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && npx cdk synth --all && node scripts/convert-template.mjs VocWebSearchStack VocCoreStack VocIngestionStack  VocProcessingStack VocApiStack BedrockAccessStack",
+    "generate:workshop-artifacts": "rm -rf cdk.out && rm -rf Workshop && AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1 npx cdk synth --all && node scripts/convert-template.mjs VocWebSearchStack VocCoreStack VocIngestionStack VocProcessingStack VocApiStack BedrockAccessStack",
     "validate:plugins": "ts-node scripts/validate-plugins.ts",
     "prebuild:frontend": "npm run generate:config"
   },


perrozzi:

This PR adds `VocWebSearchStack` to the stack list passed to `scripts/convert-template.mjs` in the `generate:workshop-artifacts` script (`voc-datalake/package.json`), so the stack that exports the AgentCore gateway ARN/URL is published alongside the `VocProcessingStack`/`VocApiStack` templates that `Fn::ImportValue` them. The diagnosis is convincing and the change is minimal and low-risk — no application code, IaC, or IAM is touched. My concerns are about the pieces of the publishing pipeline *outside* this one line that the diff doesn't cover.

### Blockers

- The diff only produces the template; nothing here registers it with the workshop. Publishing a sixth template requires the workshop `contentspec.yaml` (and any deploy-order/parameter wiring in it) to list `VocWebSearchStack` *before* `VocProcessingStack` and `VocApiStack`, since CloudFormation `Fn::ImportValue` fails if the exporter hasn't reached CREATE_COMPLETE. Argument order to `convert-template.mjs` does not influence Workshop Studio's deploy order. Either include the `contentspec.yaml` change in this PR or state explicitly where it lives and that it has already been updated — otherwise the event still fails, just with a different error path.

### Should fix

- The PR description documents a silent correctness trap (`AWS_REGION` must be `us-east-1`, `CDK_DEFAULT_REGION` is overwritten by the CLI, `contentspec.yaml` advertises us-west-2) and then declines to fix it. Since the published artifacts are only valid when synthed from us-east-1, the region should be pinned or asserted in the script itself — e.g. `AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1 npx cdk synth --all` in `generate:workshop-artifacts`, or a fail-fast check in `convert-template.mjs` that rejects templates containing `Custom::CrossRegionExportWriter` / SSM cross-region readers. A knowledge-in-the-PR-description mitigation will not survive the next person who runs this.
- This class of bug (converter stack list drifting from the set of stacks the app actually synthesizes) will recur the next time a stack is added. A cheap guard is worth more than the one-line fix: have `convert-template.mjs` diff its argument list against the manifest in `cdk.out/manifest.json` and error on any synthesized non-nested stack that wasn't requested. That also satisfies the repo's "bug fixes need a regression test" rule, which this PR currently has no answer for — a vitest case asserting the script list covers every app stack would fail if the fix were reverted.

### Nits

- The pre-existing double space between `VocIngestionStack` and `VocProcessingStack` is still there; worth cleaning up while touching this line.
- The stack list is now long enough that a JSON array in a small config file (or deriving it from the CDK app) would read better than a positional argv chain, but that's beyond this PR's scope.
- The verification section is thorough; consider capturing the "export names match byte for byte" check as an automated assertion rather than a manual step.

---

perrozzi:
Thanks — the blocker is a fair catch and the region point changed my mind. Dispositions below; a follow-up commit is going up right after this comment.

## Blocker: `contentspec.yaml` — valid, and it lives in a different repo

You're right that this diff alone doesn't register the template with the workshop, and right that **argv order to `convert-template.mjs` has no bearing on Workshop Studio's deploy order**. My PR description said the stack was "listed first to mirror the `--all` deploy order", which reads as if argv order mattered — it doesn't, it's cosmetic. Deploy order comes from the `cloudformationTemplates` list in `contentspec.yaml`.

That file is not in this repository. It lives in the Workshop Studio content repo (`ai-pdlc-voice-of-customer-voc-analytics`, a `workshopstudio://` remote with separate credentials), so it can't be included here. It has been updated in that repo, committed, and is awaiting a push by someone with content-account credentials:

- `static/cfn/VocWebSearchStack.json` added.
- Registered as the **first** template, so the exporter reaches `CREATE_COMPLETE` before Processing and Api import from it.
- `deployableRegions` narrowed to us-east-1 only — see below.

Worth noting the delta there is only that one template: the five already-published templates already carry the web-search imports, and this stack has no Lambda, so there are no asset zips and nothing needs syncing to the assets bucket.

## Should fix: pin the synth region — agreed, fixing it in the script

You're right that a note in the PR description won't survive the next person. Pushing the pin now:

```
AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1 npx cdk synth --all
```

Investigating this turned up a second, worse face of the same trap that I'd missed. `convert-template.mjs` genericizes region literals to make templates portable, which is correct for the other five stacks but actively wrong for this one, because it is legitimately region-pinned. The connector ARN comes out as:

```json
"Fn::Sub": "arn:aws:bedrock-agentcore:${AWS::Region}:aws:tool/web-search.v1"
```

So the ARN follows the **deploy** region, while the AWS-managed `web-search` connector exists only in us-east-1. Since Workshop Studio deploys every template into the single event region, it cannot put the gateway in us-east-1 while the app runs elsewhere, and `contentspec.yaml` was advertising us-west-2 as a deployable region. A us-west-2 event would fail at stack `[1/6]` and take every downstream stack with it — strictly worse than the bug this PR fixes, since web search now deploys first.

Hence the `deployableRegions` narrowing in the workshop repo. Flagging it here because the genericization behaviour lives in this repo's converter, so it's worth knowing that a region-pinned stack is a case the converter's portability assumption doesn't cover.

## Should fix: drift guard + regression test — good idea, proposing it as a follow-up

Agreed on the substance: the converter's stack list drifting from the set of stacks the app synthesizes is exactly what caused this, and a `cdk.out/manifest.json` diff would catch the next one. I'd rather land it as its own PR than grow a one-line fix, since a meaningful test needs a synthesized `cdk.out` to assert against and that needs some thought about CI. Happy to open it.

One correction: the "repo's *bug fixes need a regression test* rule" isn't a rule here. `CONTRIBUTING.md` asks only that local tests pass, and `.kiro/steering/testing.md` covers how to write good tests, not when they're mandatory. The guard is worth doing on its own merits, not because a rule requires it.

I also didn't run the gate suite for this change: it's a string inside `package.json`'s `scripts` block, which tsc, eslint and vitest never read, so they can't be affected. I verified the file still parses and retains all 23 scripts.

## Nits

- Double space between `VocIngestionStack` and `VocProcessingStack` — fixed in the follow-up commit.
- JSON array / deriving the list from the CDK app — agreed it would read better; leaving it with the drift-guard follow-up, where it belongs.
- Automating the "export names match" check — same follow-up; that assertion is really the drift guard in another form.

perrozzi:

Since my last review the only change is to the `generate:workshop-artifacts` script: `AWS_REGION`/`AWS_DEFAULT_REGION` are now pinned to `us-east-1` for the `cdk synth --all`, and the stray double space in the stack list was removed. That addresses one of my two should-fix items; the `contentspec.yaml` blocker and the drift guard/regression test are untouched.

### Resolved
- Region trap is now enforced in the script rather than only documented in the PR description, so cross-region references (`Custom::CrossRegionExportWriter` / SSM readers) can no longer sneak into the published templates.
- Double space between `VocIngestionStack` and `VocProcessingStack` cleaned up.

### Still open
- No `contentspec.yaml` (or equivalent) change or statement of where deploy ordering is registered, so the exporter may still not be created before `VocProcessingStack`/`VocApiStack` import it.
- No guard/regression test that the converter's stack list covers every synthesized non-nested stack (`cdk.out/manifest.json`), so this drift can recur.

### New findings
- `voc-datalake/package.json`: the inline `VAR=value cmd` prefix is POSIX-shell-only — the script now fails on Windows `cmd.exe` (npm's default shell there). If Windows contributors are in scope, use `cross-env` (already a common devDependency pattern) or move the pin into a small node wrapper.
- Minor: the pin now applies to the whole `--all` synth, not just `VocWebSearchStack`. If any other stack resolves region at synth time (context lookups, region-derived names/endpoints, `Stack.of(this).region` baked into strings), those five templates are now hardcoded to `us-east-1` where they previously followed the ambient region. Worth a quick confirmation that the other five templates are byte-identical to the previously published ones apart from the new stack.

---

perrozzi:
Backing out the region pin I added in `da42435` — pushing that as a third commit shortly. The double-space cleanup stays.

On reflection the pin was the wrong response to your point, for two reasons.

It overstated what it achieved. Pinning the *synth* region doesn't make a us-west-2 event work: the workshop's `contentspec.yaml` advertises us-west-2 as an optional deployment region, and Workshop Studio deploys every template into the single event region, so in a us-west-2 event the gateway template lands where the connector doesn't exist regardless of how the artifacts were synthed. I'd conflated "the templates are wired for same-region exports" with "the event will work".

And it put the constraint in the wrong place — the artifact generator in this repo, forcing every publisher to satisfy one consumer's requirement.

The practical argument against it: an event admin provisions and tests an event before running it, so a us-west-2 attempt fails visibly at test time and they move to us-east-1. That's a clearer signal than a build script picking the region silently.

The knowledge stays where it's actionable — `web-search-stack.ts` and `docs/deployment.md` both already state that `VocWebSearchStack` always deploys to us-east-1 because the connector only exists there, and [the AWS docs confirm it](https://docs.aws.amazon.com/bedrock-agentcore/latest/devguide/gateway-target-connector-web-search-tool.html).

So the PR lands back at its original one-line scope: add `VocWebSearchStack` to the converter's stack list. Your fail-fast alternative — having `convert-template.mjs` reject a template containing `Custom::CrossRegionExportWriter` — remains the better version of this idea, since it *detects* a wrongly-wired artifact instead of forcing one shape. I'll fold it into the drift-guard follow-up rather than this PR.

perrozzi:

Since my last review the region pin was reverted: `AWS_REGION=us-east-1 AWS_DEFAULT_REGION=us-east-1` was removed from `generate:workshop-artifacts`, leaving only the added `VocWebSearchStack` argument (double-space cleanup retained). This un-fixes one of the two should-fix items I had marked resolved.

### Resolved
- Double space between `VocIngestionStack` and `VocProcessingStack` remains cleaned up.
- The POSIX-only inline `VAR=value` prefix is gone, so the script works under `cmd.exe` again.

### Still open
- No `contentspec.yaml` (or equivalent) change / statement of where deploy ordering is registered for the new exporter stack.
- No guard/regression test that the converter's stack list covers every synthesized non-nested stack from `cdk.out/manifest.json`.

### New findings
- Regression (should-fix), `voc-datalake/package.json`: removing the region pin reinstates the exact trap the PR description warns about — if the ambient region isn't `us-east-1`, `cdk synth --all` switches `VocWebSearchStack` to cross-region references (`Custom::CrossRegionExportWriter` + Lambda) and the consumers read the gateway via SSM instead of `Fn::ImportValue`, so the published templates silently don't match the wiring the new stack provides. Either pin the region in a cross-platform way (`cross-env`, or a small node wrapper around the synth) or fail fast in `convert-template.mjs` when a cross-region export writer / SSM reader is detected in the converted templates.

---

PR #269, opened 2026-08-06
chore(workshop): make workshop artifacts deployable (5-template cap + model pin)

perrozzi:
Makes the workshop artifacts deployable in a Workshop Studio event. Three commits, two independent problems.

## 1. Web search cannot be published (fixes what #268 attempted)

#268 added `VocWebSearchStack` to the converter's stack list so the exporting stack would ship alongside the `VocProcessingStack`/`VocApiStack` templates that `Fn::ImportValue` its gateway ARN and URL. That turns out to be unpublishable — Workshop Studio rejects a contentspec referencing more than **5** CloudFormation templates:

```
INFRASTRUCTURE_VALIDATION_ERROR
[Contentspec Error - cloudformationTemplates] total number of cloudformation
templates referenced as a part of this infrastructure configuration exceeded 5
```

The workshop already publishes 5 and none can give up the slot: `BedrockAccessStack` creates the Bedrock model agreements, without which the app cannot invoke any model.

So the workshop opts out instead: synth with `-c enableWebSearch=false`, which drops the stack from the app and removes the imports from its consumers, leaving exactly 5 templates and no dangling exports. The 5-step participant lab never exercised web search — verified four ways: zero mentions across all content files and guides; all three backend entry points are strict opt-in; the UI hides both the chat toggle and the research data-source card when the feature flag is false; and an explicit request degrades with a warning rather than failing.

**The flag and the stack list must change together.** With the flag set, the stack is never constructed, so `cdk.out` holds no template for it, `convert-template.mjs` logs `Failed to load template or assets` and exits 1 because `success < stackNames.length`.

## 2. The event account cannot use the default models

Those accounts sit behind an AWS Private Marketplace that refuses `CreateFoundationModelAgreement` for Sonnet 5 and Opus 5. Reproduced identically in **two separate event accounts**, so it is how they are provisioned rather than a one-off. Only Sonnet 4.6 and Haiku 4.5 are invocable.

That breaks two things, addressed separately:

**`BedrockAccessStack` failed the whole deployment.** An `AccessDeniedException` from `CreateFoundationModelAgreement` is now reported as `UNAVAILABLE` instead of raising, joining the handler's existing non-fatal outcomes. Every other `ClientError`, and every other exception, still raises. The new branch sits *below* the `ConflictException` handler deliberately — that exception is itself a `ClientError`, so ordering it first would swallow the already-exists case.

Rationale: the agreements that do succeed remain usable and survive the failed stack, and a model the account cannot subscribe to would only surface as an `AccessDenied` at inference time anyway. Failing the whole stack over one such model is disproportionate, and when it is not last in the deploy order it takes everything after it down too.

**The app defaulted to models it cannot invoke.** A new opt-in `-c defaultModelId=<allowlisted id>` seeds `settings.model_id` — the legacy global override — through a create-once custom resource. Both resolvers already rank it above the built-in defaults:

```
explicit argument
> per-surface override (settings.surfaces[S])
> legacy global override (settings.model_id)     <- what this writes
> built-in SURFACE_DEFAULTS[S]
> BEDROCK_MODEL_ID
```

So one attribute repoints chat, documents, prototype, enrichment and utility without touching `SURFACE_DEFAULTS`. Note `BEDROCK_MODEL_ID` is deliberately not used: it ranks *below* the surface defaults, so setting it changes nothing while a surface has a default.

The workshop synth passes `-c defaultModelId=global.anthropic.claude-sonnet-4-6`.

## Scope: `cdk deploy` is untouched

- Web search still deploys by default; app-in-any-region with the gateway pinned to us-east-1 still works via CDK cross-region references.
- With no `defaultModelId` flag, **none** of the pin resources are created — no Lambda, provider, log groups or IAM — and the newer models remain the defaults.

## Implementation notes

- The pin is validated against `ALLOWED_MODEL_IDS` at synth. A non-allowlisted value would be dropped by `_allowlisted()` at read time and silently fall back to the very defaults the flag exists to bypass, so failing fast is clearer.
- Create-once (`ConditionExpression: attribute_not_exists(model_id)`), matching `AdminBootstrap`'s "create once, never reset": an admin who later picks a model in Settings is never overwritten by a redeploy. A floor, not a lock — per-surface overrides still outrank it.
- Kept out of `admin_bootstrap.py` deliberately: that handler is single-purpose and its provider logging is pinned `FATAL` because the initial admin password flows through its response. A separate resource also allows creating nothing when the flag is absent.
- Unexpected DynamoDB failures raise rather than no-op: a green stack whose every AI surface resolves to an uninvocable model is much harder to diagnose than a failed one.
- `grantWriteData()`'s GSI and KMS wildcards reuse the shared `dynamoDbGsi` / `kmsEncryption` nag suppressions.

## Verification

12 new tests, all fail-on-revert. The agreement tests extract the inline handler from the CDK stack and `exec` it against a stubbed boto3 — the same read-the-TS-from-Python approach as `test_avatar_image_model_lockstep.py` — asserting `AccessDenied` is absorbed, other codes still raise, and the `ClientError` branch does not shadow `ConflictException`. The pin tests assert the write is conditional, an existing value is kept, unexpected errors raise, and the `PhysicalResourceId` is stable across request types.

`tsc` clean. Synthed with **both** flags together, the exact combination the workshop script now uses:

- 5 stacks synthed, no `VocWebSearchStack`
- zero `VocWebSearchStack` references in `VocProcessingStack`
- one `Custom::ModelPin` in `VocCoreStack` with `ModelId=global.anthropic.claude-sonnet-4-6`
- cdk-nag clean; a bogus `-c defaultModelId=gpt-4` fails synth with the allowlist in the message
- with no flag: zero `ModelPin` resources

Live evidence for part 1: after publishing the 5-template set, a fresh event brought `VocCoreStack`, `VocIngestionStack` and `VocProcessingStack` to `CREATE_COMPLETE`, where processing had previously failed on the missing export.

## Not covered here

The corresponding workshop-repo change (drop the 6th template, restore the 5-entry contentspec, publish the regenerated templates) is committed in the Workshop Studio content repo, which lives outside this repository. The pin has not yet run in a real event — the next rebuild is its first live exercise.

### DIFF
--- a/voc-datalake/lambda/custom_resources/model_pin.py
+++ b/voc-datalake/lambda/custom_resources/model_pin.py
@@ -1,0 +1,85 @@
+"""Deploy-time seed for the global model pin (``settings.model_id``).
+
+Why this exists
+--------------
+Some accounts cannot invoke the newest Claude models. Workshop Studio events in
+particular sit behind an AWS Private Marketplace that refuses
+``CreateFoundationModelAgreement`` for Sonnet 5 / Opus 5, while Sonnet 4.6 and
+Haiku 4.5 work. Those events are fully automated: nobody can open Settings and
+pick a model per participant account, so the choice has to be made at deploy
+time.
+
+How it works
+------------
+``shared/model_config.py`` resolves a surface's model as::
+
+    explicit argument
+    > per-surface override (settings.surfaces[S])
+    > legacy global override (settings.model_id)     <- what this writes
+    > built-in SURFACE_DEFAULTS[S]
+    > BEDROCK_MODEL_ID
+
+Writing the single ``model_id`` attribute therefore repoints **every** surface,
+using a path both resolvers already implement (the TypeScript streaming-chat
+lookup in ``lambda/stream/src/bedrock/model-override.ts`` honours the same
+precedence). No change to ``SURFACE_DEFAULTS`` is needed, so deployments that
+can use the newer models are unaffected.
+
+Create-once, like the sibling admin bootstrap: the write is conditional on
+``model_id`` being absent, so an admin who later picks a model in Settings is
+never overwritten by a redeploy. It is a floor, not a lock — per-surface
+overrides still outrank it.
+"""
+import logging
+
+import boto3
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+MODEL_SETTINGS_PK = 'SETTINGS#model'
+MODEL_SETTINGS_SK = 'config'
+
+PHYSICAL_ID = 'voc-model-pin'
+
+
+def handler(event, _context):
+    """Seed settings.model_id unless it is already set.
+
+    Raises on unexpected failures: a silent no-op would leave the deployment
+    green while every AI surface resolves to a model the account cannot
+    invoke, which is far harder to diagnose than a failed stack.
+    """
+    request_type = event.get('RequestType')
+    props = event.get('ResourceProperties', {})
+    model_id = props.get('ModelId', '')
+    # Both arrive as resource properties from core-stack.ts. Deliberately no
+    # AGGREGATES_TABLE env fallback: nothing sets that variable on this
+    # function, and the fallback would mask a missing property (it silently
+    # absorbed one in tests, where lambda/conftest.py defines it).
+    table_name = props.get('TableName', '')
+
+    if request_type == 'Delete':
+        # Leave the setting in place; deleting the stack should not silently
+        # repoint a still-running app, and the table goes with it anyway.
+        return {'PhysicalResourceId': PHYSICAL_ID, 'Data': {'outcome': 'skipped'}}
+
+    if not model_id or not table_name:
+        raise ValueError(f'ModelId and TableName are required (got {model_id!r}, {table_name!r})')
+
+    table = boto3.resource('dynamodb').Table(table_name)
+    try:
+        table.update_item(
+            Key={'pk': MODEL_SETTINGS_PK, 'sk': MODEL_SETTINGS_SK},
+            UpdateExpression='SET model_id = :m',
+            ConditionExpression='attribute_not_exists(model_id)',
+            ExpressionAttributeValues={':m': model_id},
+        )
+        logger.info(f'Pinned every AI surface to {model_id}')
+        outcome = 'pinned'
+    except table.meta.client.exceptions.ConditionalCheckFailedException:
+        # Already configured (a previous deploy, or an admin's own choice).
+        logger.info(f'model_id already set; leaving it untouched (wanted {model_id})')
+        outcome = 'kept'
+
+    return {'PhysicalResourceId': PHYSICAL_ID, 'Data': {'outcome': outcome, 'modelId': model_id}}
--- a/voc-datalake/lambda/custom_resources/test/test_model_pin.py
+++ b/voc-datalake/lambda/custom_resources/test/test_model_pin.py
@@ -1,0 +1,141 @@
+"""Tests for the deploy-time global model pin.
+
+Fail-on-revert intent:
+  - the write MUST be conditional, so a redeploy cannot clobber a model an
+    admin later picked in Settings (create-once, like admin_bootstrap);
+  - an unexpected failure MUST raise, because a silent no-op leaves the stack
+    green while every AI surface resolves to a model the account cannot invoke.
+"""
+import sys
+import types
+from pathlib import Path
+
+import pytest
+
+sys.path.insert(0, str(Path(__file__).resolve().parents[1]))
+
+
+class _ConditionalCheckFailed(Exception):
+    pass
+
+
+class _FakeTable:
+    def __init__(self, *, raises=None):
+        self.raises = raises
+        self.calls = []
+
+        conditional = _ConditionalCheckFailed
+
+        class _Exceptions:
+            ConditionalCheckFailedException = conditional
+
+        class _Client:
+            exceptions = _Exceptions()
+
+        class _Meta:
+            client = _Client()
+
+        self.meta = _Meta()
+
+    def update_item(self, **kwargs):
+        self.calls.append(kwargs)
+        if self.raises is not None:
+            raise self.raises
+        return {}
+
+
+def _load(table):
+    """Import model_pin with boto3 stubbed to hand back our fake table."""
+    fake_boto3 = types.ModuleType('boto3')
+    fake_boto3.resource = lambda *_a, **_k: types.SimpleNamespace(Table=lambda _n: table)
+
+    saved_boto3 = sys.modules.get('boto3')
+    saved_module = sys.modules.pop('model_pin', None)
+    sys.modules['boto3'] = fake_boto3
+    try:
+        import model_pin  # noqa: PLC0415 — deliberate re-import under the stub
+        return model_pin
+    finally:
+        if saved_boto3 is None:
+            sys.modules.pop('boto3', None)
+        else:
+            sys.modules['boto3'] = saved_boto3
+        if saved_module is not None:
+            sys.modules['model_pin'] = saved_module
+        else:
+            sys.modules.pop('model_pin', None)
+
+
+def _event(request_type='Create', **props):
+    base = {'TableName': 'voc-aggregates', 'ModelId': 'global.anthropic.claude-sonnet-4-6'}
+    base.update(props)
+    return {'RequestType': request_type, 'ResourceProperties': base}
+
+
+def test_pins_the_model_when_unset():
+    table = _FakeTable()
+    module = _load(table)
+
+    result = module.handler(_event(), None)
+
+    assert result['Data']['outcome'] == 'pinned'
+    assert result['Data']['modelId'] == 'global.anthropic.claude-sonnet-4-6'
+    call = table.calls[0]
+    assert call['Key'] == {'pk': 'SETTINGS#model', 'sk': 'config'}
+    assert call['ExpressionAttributeValues'] == {':m': 'global.anthropic.claude-sonnet-4-6'}
+
+
+def test_write_is_conditional_so_a_redeploy_cannot_clobber_an_admin_choice():
+    table = _FakeTable()
+    module = _load(table)
+
+    module.handler(_event(), None)
+
+    assert table.calls[0]['ConditionExpression'] == 'attribute_not_exists(model_id)'
+
+
+def test_existing_value_is_kept_not_overwritten():
+    table = _FakeTable(raises=_ConditionalCheckFailed('already set'))
+    module = _load(table)
+
+    result = module.handler(_event(), None)
+
+    assert result['Data']['outcome'] == 'kept'
+
+
+def test_unexpected_errors_raise_rather_than_silently_no_op():
+    table = _FakeTable(raises=RuntimeError('table on fire'))
+    module = _load(table)
+
+    with pytest.raises(RuntimeError):
+        module.handler(_event(), None)
+
+
+def test_delete_is_a_no_op_and_leaves_the_setting_alone():
+    table = _FakeTable()
+    module = _load(table)
+
+    result = module.handler(_event('Delete'), None)
+
+    assert result['Data']['outcome'] == 'skipped'
+    assert table.calls == []
+
+
+@pytest.mark.parametrize('missing', ['TableName', 'ModelId'])
+def test_missing_required_properties_raise(missing):
+    table = _FakeTable()
+    module = _load(table)
+
+    with pytest.raises(ValueError):
+        module.handler(_event(**{missing: ''}), None)
+
+
+def test_physical_id_is_stable_across_request_types():
+    """A changing PhysicalResourceId would make CloudFormation replace/delete the resource."""
+    table = _FakeTable()
+    module = _load(table)
+
+    created = module.handler(_event(), None)['PhysicalResourceId']
+    deleted = module.handler(_event('Delete'), None)['PhysicalResourceId']
+
+    assert created == deleted
--- a/voc-datalake/lambda/shared/test/test_model_agreement_handler.py
+++ b/voc-datalake/lambda/shared/test/test_model_agreement_handler.py
@@ -52,7 +52,9 @@ class _FakeConflict(Exception):
     pass
 
 
-def _load_handler(*, create_raises: Exception | None = None):
+def _load_handler(*, create_raises: Exception | None = None,
+                  availability_raises: Exception | None = None,
+                  offers_raises: Exception | None = None):
     """exec the extracted handler with boto3/botocore stubbed out."""
     calls: dict[str, int] = {'create': 0}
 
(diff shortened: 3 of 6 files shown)


perrozzi:

This PR narrows the workshop artifact pipeline to 5 CloudFormation templates by synthing with `-c enableWebSearch=false` and dropping `VocWebSearchStack` from the `convert-template.mjs` stack list, in a single one-line change to `voc-datalake/package.json`. The change is minimal, self-consistent (flag and stack list move together, as the description correctly notes), and touches no application code, IaC, or IAM — so the repo-specific gates (Lambda IAM size, ingestor prefix registration, Bedrock model/ARN lockstep, i18n) are not in play. Looks clean to me.

### Should fix
- `voc-datalake/package.json`: the 5-template Workshop Studio cap is now an invisible constraint encoded in a single npm script. The next person who adds a stack will hit `INFRASTRUCTURE_VALIDATION_ERROR` only at publish time. Add a guard in `scripts/convert-template.mjs` (e.g. fail fast if `stackNames.length > 5` with a message naming the contentspec limit), or at minimum a note in the workshop/deployment docs next to the script. This is the kind of coupling that a comment can't live in, since `package.json` is JSON.
- The synth flag and the stack list are two halves of one invariant that only fail loudly at `exit 1` after a full `cdk synth --all`. Consider having `convert-template.mjs` read the stack list from `cdk.out/manifest.json` (or intersect against it) rather than hardcoding names, which makes the coupling structural instead of conventional.
- Worth a sanity check that the frontend degrades cleanly with `features.webSearch: false` in the generated `config.json` — no empty menu entries, no missing-i18n-key warnings, no request to the absent gateway URL. The description verifies the asset delta but not the runtime behaviour of the flag in the published workshop build.

### Nits
- The PR body is the only place recording *why* web search was dropped rather than `BedrockAccessStack`. That reasoning (all five models start at `NOT_AVAILABLE` without the agreements stack) is valuable and will be lost from the repo; a line in the deployment or workshop doc would preserve it.
- The workshop-repo counterpart (contentspec back to 5 entries, regenerated templates) is outside this repository, so the two must land together — nothing to fix here, just flagging that this PR is not independently verifiable from the diff.

---

perrozzi:
Scope expanded — pushing two more commits shortly, and I've retitled/rewritten the description to match.

This PR started as just the 5-template cap fix. It now also carries the second reason a workshop event can't run, because the two are only useful together:

- `f4927aa` — `BedrockAccessStack` tolerates an unavailable model agreement (`AccessDeniedException` → `UNAVAILABLE` instead of raising), plus an opt-in `-c defaultModelId=<allowlisted id>` that seeds `settings.model_id` via a create-once custom resource.
- `a20fc7b` — the workshop synth passes `-c defaultModelId=global.anthropic.claude-sonnet-4-6`.

Why they belong here rather than in a separate PR: both commits are workshop-artifact scoped, and the `defaultModelId` flag lands on the same `generate:workshop-artifacts` line this PR already rewrites. Splitting them would have guaranteed a conflict between the two PRs for no reviewer benefit.

The motivating fact is new since the original description: the model restriction reproduced identically in **two separate event accounts**, so it is how Workshop Studio provisions them, not a one-off. Sonnet 5 and Opus 5 are refused; only Sonnet 4.6 and Haiku 4.5 work. Without the pin the participant lab fails at persona generation even once all five stacks are green.

`cdk deploy` is unaffected in both respects — web search still deploys by default, and with no `defaultModelId` flag none of the pin resources are created at all.

Also worth noting for the earlier review: the fail-fast idea from the last round (rejecting a template that contains `Custom::CrossRegionExportWriter`) is still unbuilt, and still the better version of the region concern. I'd rather land it with the converter drift-guard than bolt it on here.

Verification of the combination, plus 12 new fail-on-revert tests, is in the updated description.

perrozzi:

Since the last review the PR grew well beyond the `package.json` one-liner: a new `lambda/custom_resources/model_pin.py` custom resource wired into `VocCoreStack` behind `-c defaultModelId`, a non-fatal `AccessDeniedException` branch in `BedrockAccessStack`'s inline agreement handler, and two new test modules. None of my previous points were addressed (they were about the 5-template coupling, not this new code), and the `package.json` script now carries a second invisible constraint.

### Still open
- No guard/documentation for the 5-template Workshop Studio cap (`scripts/convert-template.mjs` still hardcodes the stack list; failure only at publish time).
- Stack list is still conventionally, not structurally, coupled to the synth flags — now two flags (`enableWebSearch=false`, `defaultModelId=...`) instead of one.
- Frontend behaviour with `features.webSearch: false` in the published build still unverified in the description.

### New findings
- **`voc-datalake/lib/stacks/bedrock-access-stack.ts`: the new `except ClientError` wraps the whole `try`, not just `create_foundation_model_agreement`.** An `AccessDeniedException` raised by `get_foundation_model_availability` or `list_foundation_model_agreement_offers` — i.e. a missing IAM permission on the stack's own role, a genuine bug — is now swallowed as `UNAVAILABLE` and the stack goes green. Narrow the absorption to the create call (separate try, or check `e.operation_name == 'CreateFoundationModelAgreement'`) and add a test that AccessDenied from the *availability* call still raises; the current test only stubs the create call.
- **Bedrock model id / IAM lockstep for the pin:** `-c defaultModelId=global.anthropic.claude-sonnet-4-6` repoints *every* surface via `settings.model_id`, but nothing in this diff shows that the invoke policies of the consuming Lambdas include that id's ARN (`global.` cross-region inference profile ARNs in particular). Please confirm the `bedrock:InvokeModel` resources are derived from `ALLOWED_MODEL_IDS` and not from `SURFACE_DEFAULTS` — if the latter, the workshop deploys a green stack where every AI call fails with AccessDenied, the exact failure mode the pin exists to prevent. A synth-time assertion (pinned id ∈ granted ARNs) would make this fail loudly.
- **Settings item shape:** the pin creates `pk=SETTINGS#model / sk=config` with only `model_id`, whereas an admin-written settings item presumably also carries `updated_at`/`updated_by`/type markers. Worth checking that the Settings GET path and any Zod/response validation tolerate the sparse item (and that the Settings UI renders it) — otherwise the workshop's first Settings page load is the discovery point.

### Nits
- Changing `-c defaultModelId` on an existing deployment silently no-ops (create-once + conditional write on Update). Intentional per the description, but the CFN Output `DefaultModelPin` will then advertise a value the table does not hold; consider returning/outputting the effective value or logging the divergence.
- `voc-datalake/lambda/shared/test/test_model_agreement_handler.py` tests a file under `lib/stacks/` from `lambda/shared/test/`, with `parents[3]` path arithmetic and a regex over a TS template literal — it will break on any refactor of that method's signature or on an escaped backtick in the body. A `# do not rename` marker next to `getModelAgreementLambdaCode()`, or extracting the handler to a real `.py` file inlined like `model_pin.py` (which this PR shows is easy), would be sturdier.
- `model_pin.py` uses stdlib `logging` rather than Powertools `Logger`; acceptable for `Code.fromInline` with no layer, but worth a one-line comment saying so, since the repo pattern is otherwise Powertools.

---

perrozzi:
Good catch on the first finding — that was a real hole. Fixed in `c48917d`, pushing after this comment. Dispositions below, including two findings I checked and believe are already satisfied.

## New findings

**1. `except ClientError` wrapped the whole `try` — CONFIRMED, fixed.**

You're right. An `AccessDeniedException` from `get_foundation_model_availability` or `list_foundation_model_agreement_offers` — this stack's own role missing a permission, a genuine bug — was absorbed as `UNAVAILABLE` and the stack went green having created no agreement at all. My tests only stubbed the create call, which is exactly why it passed them.

The absorption now lives in its own `try` around `create_foundation_model_agreement` only. `ConflictException` still resolves correctly: it is itself a `ClientError`, so the inner handler sees it, finds it is not in `NON_FATAL_ERROR_CODES`, and re-raises to the outer `except bedrock.exceptions.ConflictException`, which reports `ALREADY_EXISTS` as before.

Added both tests you asked for — AccessDenied from the availability call and from the offers call must raise, and must do so before the agreement call is attempted. Both fail against the previous implementation. 14 tests pass.

**2. IAM lockstep for the pinned id — already satisfied.** The invoke grants are derived from `ALLOWED_MODEL_IDS`, not `SURFACE_DEFAULTS`. `allowlistedModelArns()` iterates the allowlist and emits both the region/account-scoped `global.` inference-profile ARN and the cross-region foundation-model ARN for every entry, and it is the single source for every `bedrock:InvokeModel` grant in `api-stack.ts`, `processing-stack-consolidated.ts` and `ingestion-stack.ts`. So `global.anthropic.claude-sonnet-4-6` is granted wherever it can be selected.

This is now also confirmed empirically: a workshop test event deployed from these artifacts came up with all five stacks `CREATE_COMPLETE`, the pin item written, and the app working.

**3. Sparse settings item — already tolerated by design.** `get_model_settings()` reads `item.get('surfaces')` behind an `isinstance(..., dict)` guard, so a missing map becomes `{}`, and it reads `model_id` separately as the legacy global. Our item (`pk`, `sk`, `model_id`) yields every picker surface as `selected: null` (Automatic) plus the global override — which the handler's own docstring describes as "when set it applies to any surface left on Automatic". No `updated_at`/`updated_by` is required by the read path.

## Still open

**5-template cap guard, and the stack-list/flag coupling.** Both fair, and now two flags rather than one. Still deliberately unbuilt here — I'd rather land it with the converter drift-guard (diff the argv against `cdk.out/manifest.json`) plus your earlier fail-fast idea (reject a template containing `Custom::CrossRegionExportWriter`), as one coherent change rather than three partial guards. Happy to open it next.

**Frontend behaviour with `features.webSearch: false` — now verified, not just asserted.** In the deployed event: `config.json` contains `{"features":{"webSearch":false}}`, and `isWebSearchAvailable()` (`runtimeConfig.ts`) gates both surfaces — the chat toggle at `ChatFilters.tsx:215` and the research data-source card at `Wizards.tsx:123` — so neither control renders. The research wizard's own default is already `useWebSearch: false`, and if a request asked for it anyway, `research_step_handler.py` logs "not configured; skipping" and continues.

## Nits

- **`DefaultModelPin` output can advertise a value the table does not hold** — agreed, that is a real divergence once someone changes the flag on an existing deployment. Noting it rather than fixing it now: returning the effective outcome from the custom resource is the right shape, and I'd rather not reopen the published artifacts for it in the same pass.
- **Test path fragility / extract the inline handler to a real `.py`** — agreed, and your point that `model_pin.py` shows it's easy is well made. Deliberately deferred: it changes the inline handler bytes, so it means republishing workshop artifacts that were validated an hour ago. Worth doing, better as its own change.
- **`model_pin.py` uses stdlib `logging`** — correct, because `Code.fromInline` has no Powertools layer. Will add the one-line comment saying so.

perrozzi:

The only change since my last review is the one I asked for on `bedrock-access-stack.ts`: the `AccessDeniedException` absorption is now scoped to the `create_foundation_model_agreement` call itself, with two new tests asserting AccessDenied from the availability and offers calls still raises. That fix looks correct — `ConflictException` is a `ClientError` in botocore, gets re-raised by the inner branch, and is still turned into `ALREADY_EXISTS` by the outer handler. My other two findings were not touched.

### Resolved
- `bedrock-access-stack.ts`: absorption narrowed to the create call; regression tests cover AccessDenied from `get_foundation_model_availability` and `list_foundation_model_agreement_offers` (`lambda/shared/test/test_model_agreement_handler.py`).

### Still open
- Bedrock model id / IAM lockstep for the pin: no evidence the consuming Lambdas' `bedrock:InvokeModel` resources cover `global.anthropic.claude-sonnet-4-6`.
- Settings item shape: sparse `SETTINGS#model` item (only `model_id`) unverified against the Settings GET path / UI.
- No guard or docs for the 5-template Workshop Studio cap; stack list still only conventionally coupled to the synth flags (now two).
- Frontend behaviour with `features.webSearch: false` in the published build still unverified.

### New findings
- Nit, test fidelity in `lambda/shared/test/test_model_agreement_handler.py`: `_FakeConflict` does not subclass `_FakeClientError`, so `test_conflict_is_still_treated_as_already_existing` never exercises the inner `except ClientError` re-raise path that a real `ConflictException` (a `ClientError` subclass) would hit. Making `_FakeConflict(_FakeClientError)` with `code='ConflictException'` would actually lock in the "ClientError branch does not shadow ConflictException" claim the docstring makes.

---

perrozzi:
Nit accepted and fixed in `7356e27` — you were right that the test proved nothing.

`_FakeConflict` was a sibling of `_FakeClientError`, so `test_conflict_is_still_treated_as_already_existing` never entered the inner `except ClientError` branch at all; it went straight to the outer handler. It now subclasses `_FakeClientError` with `code='ConflictException'`, so the test traverses the real control flow — inner branch catches it, finds the code is not in `NON_FATAL_ERROR_CODES`, re-raises, outer handler reports `ALREADY_EXISTS` — and would fail if the inner branch ever started absorbing conflicts. Test-only change, so no template bytes move.

On the four items still listed as open: three of them I answered in the previous comment with code references rather than code changes, so they read as untouched in a diff-only review. Restating briefly so this thread stands on its own:

- **IAM lockstep** — `allowlistedModelArns()` iterates `ALLOWED_MODEL_IDS` and is the single source for every `bedrock:InvokeModel` grant in `api-stack.ts`, `processing-stack-consolidated.ts` and `ingestion-stack.ts`. It is not derived from `SURFACE_DEFAULTS`, so `global.anthropic.claude-sonnet-4-6` is granted wherever it can be selected.
- **Sparse settings item** — `get_model_settings()` (`settings_handler.py:79`) guards `item.get('surfaces')` with `isinstance(..., dict)` and reads `model_id` separately as the legacy global; its docstring states that global "applies to any surface left on Automatic".
- **Frontend with `features.webSearch: false`** — verified in the deployed event, not just asserted: `config.json` carries `{"features":{"webSearch":false}}`, and `isWebSearchAvailable()` gates both the chat toggle (`ChatFilters.tsx:215`) and the research data-source card (`Wizards.tsx:123`).

Only the **5-template cap guard** is genuinely unbuilt, and that one I'd still rather land as its own change together with the converter drift-guard and your `Custom::CrossRegionExportWriter` fail-fast check, instead of a third partial guard here.

For the record on deployment impact: the whole set has been validated live. A workshop event built from these artifacts came up with all five stacks `CREATE_COMPLETE` — including `BedrockAccessStack`, which is what the agreement tolerance was for — with the pin item written and the app reachable. The agreement logs show three created and two absorbed, and the absorbed denials are on `CreateFoundationModelAgreement` specifically, which is the call the narrowed branch still forgives, so `c48917d` does not change that outcome.

perrozzi:

Only one change since my last review: `_FakeConflict` now subclasses `_FakeClientError` in `lambda/shared/test/test_model_agreement_handler.py`, which is exactly the nit I raised — the conflict case now really traverses the inner `except ClientError` re-raise before reaching the outer `ConflictException` handler. My three earlier open items are untouched.

### Resolved
- `lambda/shared/test/test_model_agreement_handler.py`: `_FakeConflict(_FakeClientError)` with code `ConflictException` now genuinely locks in "the `ClientError` branch does not shadow `ConflictException`".

### Still open
- Bedrock model id / IAM lockstep for the pin: no evidence consuming Lambdas' `bedrock:InvokeModel` resources cover `global.anthropic.claude-sonnet-4-6`.
- Sparse `SETTINGS#model` item (only `model_id`) still unverified against the Settings GET path / UI.
- No guard or docs for the 5-template Workshop Studio cap; stack list still only conventionally coupled to the two synth flags.
- Frontend behaviour with `features.webSearch: false` in the published build still unverified.

### New findings
- Nit, stub fidelity in `lambda/shared/test/test_model_agreement_handler.py`: the fake client aliases `exceptions.AccessDeniedException = _FakeClientError` while `exceptions.ConflictException = _FakeConflict` is now a *subclass* of it, so in the stub `ConflictException ⊂ AccessDeniedException` — the reverse of botocore, where they are siblings. Harmless for the current handler (the outer branch catches `ConflictException` and the inner branch matches on the code string), but a future `except bedrock.exceptions.AccessDeniedException` ordered ahead of the conflict branch would pass here and break in production. A separate `class _FakeAccessDenied(_FakeClientError)` for that alias removes the trap.

---

Issue #3, opened 2026-01-16
Remove unused PRD and PRFAQ generation endpoints


svozza:
## Summary

The following endpoints are no longer used:

```
POST /projects/<project_id>/prd/generate
POST /projects/<project_id>/prfaq/generate
```

This is because they have been consolidated into the `POST /projects/<project_id>/document` endpoint.

We should remove the dead code not being used.

### Backend

`voc-datalake/lambda/api/projects_handler.py`:
- Remove lines 228-229: @app.post("/projects/<project_id>/prd/generate") and api_generate_prd()
- Remove lines 232-235: @app.post("/projects/<project_id>/prfaq/generate") and api_generate_prfaq()

### Frontend

`voc-datalake/frontend/src/api/projectsApi.ts`:
- Remove `generatePRD` function
- Remove `generatePRFAQ` function

`voc-datalake/frontend/src/api/client.ts`:
- Remove `generatePRD` from api object
- Remove `generatePRFAQ` from api object

Test files to delete/update:
- `voc-datalake/frontend/src/api/projectsApi.test.ts` - Remove tests for `generatePRD` and `generatePRFAQ`

Issue #2, opened 2026-01-16
Consolidate Builtin Plugins


svozza:
Currently we have [16 plugins](https://github.com/aws-samples/sample-voice-of-customer-datalake?tab=readme-ov-file#-built-in-plugins) for getting data from various datasources such as TrustPilot, Yelp, Instagram etc. These plugin are in different states of completeness and correctness with TrustPilot being the most stable. 

We must pick a subset of plugins that we wish to ship and implement them with a full suite of tests: unit, integration, and e2e. The output of this issue will be new issues for each datasource we wish to implement and a story for the removal of the ones we decide against.

Issue #1, opened 2026-01-15
Deploy Handler Code Separately for API Lambda Functions


svozza:
Currently we have all the handlers for the API in one bundle that every lambda uses. Each Lambda deployment is essentially a zip file containing the following

```
lambda-deployment.zip
├── metrics_handler.py
├── chat_handler.py
├── integrations_handler.py
├── scrapers_handler.py
├── settings_handler.py
├── projects_handler.py
├── ... (all other API handlers)
└── shared/
    ├── __init__.py
    ├── api.py
    ├── aws.py
    ├── auth.py
    └── ... (all shared modules)
```

This means that each lambda has large amounts of code that is never used. We should change the CDK code so each lambda only ships with the code it needs.

Current approach (in [api-stack.ts](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/da5ee66c6113e90e2d8d63e6f108df98e312bedc/voc-datalake/lib/stacks/api-stack.ts#L88)):

```ts
typescript
// Shared by ALL Lambdas - contains everything
const apiCodeWithShared = lambda.Code.fromAsset('lambda', {
  bundling: {
    image: lambda.Runtime.PYTHON_3_12.bundlingImage,
    command: ['bash', '-c', 'mkdir -p /asset-output && cp -r /asset-input/api/* /asset-output/ && cp -r /asset-input/shared /asset-output/'],
    platform: 'linux/arm64',
  },
});

// Chat Lambda uses the shared bundle
const chatLambda = new lambda.Function(this, 'ChatApi', {
  runtime: lambda.Runtime.PYTHON_3_12,
  handler: 'chat_handler.lambda_handler',
  code: apiCodeWithShared,  // Contains ALL handlers
  // ...
});
```

Suggested approach:

```ts
// Helper function to create optimized Lambda code bundle
function createApiLambdaCode(handlerFileName: string): lambda.Code {
  return lambda.Code.fromAsset('lambda', {
    bundling: {
      image: lambda.Runtime.PYTHON_3_12.bundlingImage,
      command: [
        'bash', '-c',
        `mkdir -p /asset-output && ` +
        `cp /asset-input/api/${handlerFileName} /asset-output/ && ` +
        `cp -r /asset-input/shared /asset-output/`
      ],
      platform: 'linux/arm64',
    },
  });
}

// Then use it for each Lambda
const chatLambda = new lambda.Function(this, 'ChatApi', {
[...]

PR #270, opened 2026-08-06
refactor(infra): merge the two us-east-1 AI-enablement stacks

perrozzi:
## Summary

The app synthesized **six** CloudFormation templates; Workshop Studio's `contentspec.yaml` accepts a
maximum of **five**. The workaround (merged in #269) was to synth workshop artifacts with
`-c enableWebSearch=false`, so the workshop ships a build the product does not.

`BedrockAccessStack` and `VocWebSearchStack` were already the same deployment unit and the code did not
notice: **both are hard-pinned to us-east-1**, both are one-shot AI account enablement built from custom
resources, and **neither depends on the core stack chain**. Merging them frees the fifth slot, so web
search deploys to the workshop and the flag divergence goes away.

`bedrock-access-stack.ts` overrode the region inside its own `super()` call (`region: 'us-east-1'`) and
discarded the `env` that `bin/` passed it — which is why the co-location was easy to miss.

## Why merging costs nothing

| | agreements half | web-search half |
|---|---|---|
| region | us-east-1 (hard-pinned) | us-east-1 (hard-pinned) |
| depends on the core chain | no — **0** `Fn::ImportValue` | no |
| exports consumed elsewhere | **0** (its 4 outputs have no `Export` block) | 2, imported 2× each by Processing and Api |

Model agreements are unaffected by placement: `modelRegion` is a **custom-resource property**, and the
handler does `boto3.client('bedrock', region_name=region)`. A us-east-1-placed stack still creates
agreements in the app's region, so multi-region support is untouched.

## Direction matters — agreements were folded INTO the gateway stack, not the reverse

Processing and Api import the **gateway's** exports, so this direction leaves both consumer templates
untouched with **zero export renames**. The reverse would have repointed 4 imports and recreated the
gateway — whose physical name is deterministic (`voc-web-search-<account>-<region>`), so the new one
would have collided with the live one.

The agreements are the safe half to move: Bedrock agreements **persist per account** independently of the
stack (the handler treats any non-`Create` request as a no-op), and this half publishes no consumed
exports, so deleting the old stack carries no ordering constraint.

## Both halves stay independently switchable

Synth-time gating, which is what this stack already did on two axes (early-return without
`anthropicUseCase`; `skipUseCaseSubmission`). **No new flag** — `shouldDeployWebSearch()` is already
default-on / opt-out. Consumers needed **no change**: `processing-stack-consolidated.ts:290` and
`api-stack.ts:786` already guard on the optional gateway props.

New guard, `lib/utils/ai-enablement-default.ts`: **both halves off is reachable** in the normal deploy
path (an account that already has Bedrock access, deployed with `-c enableWebSearch=false`). Constructing
the stack then yields no resources, and `convert-template.mjs` strips `CDKMetadata`, so the workshop
artifact would be an invalid `Resources: {}`.

## Verification

Four toggle combinations, from real synths:

| combination | AI template | AgentCore res. | agreements | exports | Proc/Api refs | templates |
|---|---|---:|---:|---:|---:|---:|
| both on (default) | present, 22 res. | 2 | 6 | 2 | 2 / 2 | **5** |
| `-c enableWebSearch=false` | present, 18 res. | 0 | 6 | 0 | 0 / 0 | 5 |
| `-c anthropicUseCase=` | present, 4 res. | 2 | 0 | 2 | 2 / 2 | 5 |
| both off | **not constructed** | — | — | — | 0 / 0 | 4 |

**Export names are byte-identical** to the currently published ones —
`VocWebSearchStack:ExportsOutputFnGetAttWebSearchGatewayGatewayArnBA97E0DC` and `…GatewayUrlE01706EF` —
and Processing and Api each still resolve exactly those two.

`generate:workshop-artifacts` ran end-to-end: **5/5 converted**, `VocWebSearchStack` at 22 resources,
uniform `AssetBucket`/`AssetPrefix` parameters, **zero account and zero region literals in all five
templates**, and `VocCoreStack`'s `Custom::ModelPin` intact.

New `lib/stacks/web-search-stack.test.ts` — 12 tests, verified **fail-on-revert** (stubbing out the
gateway gate fails 2 of them). Pins the export names against a real consumer stack, since CDK only emits
`ExportsOutput*` once another stack references the value.

**Proven live in a Workshop Studio event account:** a 4-resource gateway-only stack reached
`CREATE_COMPLETE`, and the service API confirmed the gateway and its `web-search-tool` target both
`READY`. This settles a previously open question — the gateway becomes template [1/5], so a refusal would
have taken all five stacks down. Probe stack deleted afterwards.

### Gates

`typecheck:all` clean · `test:cdk` 132 · `test:stream` 257 · `test:backend` 1281 · frontend 1804 ·
`lint:frontend` + `lint:stream` pass.

⚠️ `lint:python` (ruff) fails with 551 errors, so `npm run check` exits 1. **Pre-existing on
`development`, not from this branch** — ruff scans only `lambda` and `plugins`, and none of the files here
live in either, so its input is byte-identical to the base. Worth its own issue: the documented deployment
gate cannot currently pass.

## cdk-nag

Dropped the `pluginSystem` / `comprehend` / `translate` stack-level suppressions that were applied to
`BedrockAccessStack`: it calls neither Comprehend nor Translate, cdk-nag is clean without them, and
inheriting them would have covered gateway resources that have always passed unsuppressed.

## Migrating an existing deployment

1. `cdk deploy --all` — the agreements appear in `VocWebSearchStack` and re-run idempotently.
   Processing and Api are untouched (no import changes).
2. `cdk destroy BedrockAccessStack` — CDK does not delete a stack that has been dropped from the app, so
   this step is manual. Safe in any order: no consumed exports, and the agreements persist in the account.

No gateway is recreated and there is no feature downtime.

## Two naming compromises, documented in code rather than fixed

- **The stack keeps the id `VocWebSearchStack`** even though it now does more than web search. That id
  determines the export names Processing and Api import; renaming it would recreate the gateway and churn
  both consumer templates.
- **`lib/stacks/bedrock-access-stack.ts` keeps its path** while now exporting a Construct
  (`BedrockModelAccess`), because `lambda/shared/test/test_model_agreement_handler.py` reads that exact
  path and regexes `getModelAgreementLambdaCode()` out of it to test the inline Python handler. A
  file-header comment records this. Happy to rename both plus the one-line test update if preferred.

## Follow-up (separate, not required here)

Publish the gateway URL/ARN to SSM and read them at Lambda cold start instead of via `Fn::ImportValue`.
Processing and Api would then carry zero web-search references, the enablement stack could be listed last
again, and a gateway failure would degrade the feature instead of blocking two stacks. Note the gateway
**URL is not derivable** — the id is `<deterministic name>-<random suffix>` — so the stack must publish
it, though IAM can still be scoped by the name pattern.

## Workshop follow-up (not in this PR)

Republishing needs the artifacts regenerated from `development` after this merges. Two changes from the
previous recipe: `VocWebSearchStack` now carries **2 asset zips** (the custom-resource framework Lambdas
came with the agreements) where it previously had none, so `development/VocWebSearchStack/` must be
populated — additively, never `--delete`. And `development/BedrockAccessStack/` becomes an orphaned
prefix; leave it, since previously published builds still reference it.

### DIFF
--- a/.kiro/steering/deployment.md
+++ b/.kiro/steering/deployment.md
@@ -46,27 +46,48 @@ npm run deploy:all   # Deploys all CDK stacks + frontend
 Always run quality checks before deploying:
 
 ```bash
-# From project root
-npm run lint         # ESLint code quality
-npm run typecheck    # TypeScript type checking
-npm run test         # Run test suite
+# From project root ONLY — voc-datalake/package.json has no `lint` script
+npm run lint         # frontend + stream ESLint, and ruff over lambda/ + plugins/
+npm run typecheck    # frontend ONLY (use typecheck:all for frontend + CDK + stream)
+npm run test         # frontend ONLY
 
-# Or run all at once
-npm run check        # lint + typecheck + test
+npm run check        # lint && typecheck:all && test && test:cdk && test:stream && test:backend
 ```
 
 ### What Each Check Does
 
-| Command | Description |
-|---------|-------------|
-| `npm run lint` | Runs ESLint to catch code quality issues |
-| `npm run typecheck` | Runs TypeScript compiler to verify types |
-| `npm run test` | Runs Vitest test suite |
-| `npm run test:coverage` | Runs tests with coverage report |
+| Command | Covers |
+|---------|--------|
+| `npm run lint` | `lint:frontend` + `lint:stream` (ESLint) + `lint:python` (ruff over `lambda/`, `plugins/`) |
+| `npm run typecheck` | Frontend only |
+| `npm run typecheck:all` | Frontend + CDK (`typecheck:cdk`) + stream |
+| `npm run test` | Frontend Vitest |
+| `npm run test:cdk` / `test:stream` / `test:backend` | CDK Vitest / stream Vitest / pytest via `.venv/bin/python` |
+| `npm run check` | All of the above, chained with `&&` |
+
+**Two traps when using `check` as a gate:**
+
+- **It chains with `&&`, so the first failure hides every later step.** A
+  `lint:python` failure means you never learn whether the CDK or backend tests
+  pass. When triaging, run the individual scripts.
+- **There is no ESLint leg for the CDK TypeScript.** `bin/` and `lib/` are covered
+  only by `typecheck:cdk` — "lint is clean" says nothing about the CDK app.
 
 ## CDK Stacks
 
-The platform consists of 4 core stacks plus 2 optional ones:
+The platform consists of 4 core stacks plus 1 AI-enablement stack.
+
+> ### ⚠️ The stack count is capped at 5 — adding a stack is not a free action
+>
+> A downstream packaging consumer of this repo accepts at most **five**
+> CloudFormation templates, and the app is at exactly five. A sixth breaks
+> packaging, and the failure appears only there — never at `cdk synth`. Fold new
+> infrastructure into an existing stack, or drop one first.
+>
+> This is why the web-search gateway and Bedrock model access share
+> `VocWebSearchStack`. Nested stacks are not an escape hatch: `NestedStackProps`
+> has no `env`, so a nested stack inherits the parent region, and that stack must
+> be pinned to us-east-1.
 
 | Stack | Description | Dependencies |
 |-------|-------------|--------------|
--- a/.kiro/steering/deployment.md
+++ b/.kiro/steering/deployment.md
@@ -74,8 +74,7 @@ The platform consists of 4 core stacks plus 2 optional ones:
 | `VocIngestionStack` | Plugin Lambdas, EventBridge schedules, SQS, Secrets | Core |
 | `VocProcessingStack` | Processor, Aggregator, Step Functions, Bedrock | Core, Ingestion |
 | `VocApiStack` | API Gateway, API Lambdas, Webhooks, WAF | Core, Ingestion, Processing |
-| `BedrockAccessStack` (optional) | Bedrock model access / Anthropic use case | None |
-| `VocWebSearchStack` (optional) | AgentCore web-search gateway — `-c enableWebSearch=true`, always us-east-1 | None |
+| `VocWebSearchStack` (AI enablement, always us-east-1) | Two independently switchable halves: AgentCore web-search gateway (default-on, `-c enableWebSearch=false` opts out) + Bedrock model access / Anthropic use case (only when `anthropicUseCase` is set). Not created when both are off | None |
 
 ### Deploy All Stacks
 
@@ -104,10 +103,12 @@ cdk deploy --all --require-approval never
 
 Due to dependencies, stacks should be deployed in this order:
 
-1. `VocCoreStack` (+ optional `BedrockAccessStack` / `VocWebSearchStack`, no dependencies)
-2. `VocIngestionStack`
-3. `VocProcessingStack`
-4. `VocApiStack`
+1. `VocWebSearchStack` (no dependencies, but must precede Processing/Api, which
+   import its gateway exports when web search is enabled)
+2. `VocCoreStack`
+3. `VocIngestionStack`
+4. `VocProcessingStack`
+5. `VocApiStack`
 
 The `cdk deploy --all` command handles this automatically.
 
--- a/.kiro/steering/structure.md
+++ b/.kiro/steering/structure.md
@@ -308,7 +308,9 @@ VocCoreStack (DynamoDB tables, S3 raw data bucket, KMS, Cognito, CloudFront)
        │           │
        │           └── Depends on: processingQueue, secretsArn, researchStateMachine, userPool
 
-Optional stacks with NO dependency on the core chain:
-  BedrockAccessStack (Bedrock model access / Anthropic use case)
-  VocWebSearchStack (us-east-1: AgentCore web-search gateway, -c enableWebSearch=true)
+AI-enablement stack, NO dependency on the core chain (but Processing/Api import
+its gateway exports, so it deploys first):
+  VocWebSearchStack (always us-east-1) = web-search gateway (default-on, opt out
+    with -c enableWebSearch=false) + Bedrock model access / Anthropic use case
+    (only when anthropicUseCase is set). Not created when both halves are off.
 ```
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ A fully serverless AWS platform for ingesting, processing, and analyzing custome
 - **Plugin-Based Architecture**: Extensible data source plugins, easily create your own
 - **AI-Powered Analysis**: Amazon Bedrock (Claude) for sentiment, categorization, and insights
 - **Per-Surface Model Picker**: admins choose the Claude model per AI feature (chat, documents, prototypes, enrichment) over a curated allowlist
-- **Opt-In Web Search**: AgentCore Gateway connector for chat and research (`enableWebSearch: true`)
+- **Web Search**: AgentCore Gateway connector for chat and research — deployed by default, searches stay opt-in per request (opt out with `enableWebSearch: false`)
 - **Real-Time Processing**: Event-driven with SQS and DynamoDB Streams
 - **Multi-Language Support**: Auto-detection and translation
 - **React Dashboard**: Metrics, charts, AI chat, and project management
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -92,7 +92,9 @@ Anthropic requires first-time customers to submit use case details before invoki
 
 ### Automatic Setup via CDK
 
-The `BedrockAccessStack` automates this process. To enable it:
+The model-access half of `VocWebSearchStack` automates this process (it was its
+own `BedrockAccessStack` before the two us-east-1 stacks were merged). To enable
+it:
 
 1. Copy the example config:
    ```bash
@@ -169,7 +171,22 @@ Access is granted immediately after successful submission.
 
 ## CDK Stacks
 
-The platform consists of 4 core stacks plus 2 optional ones:
+The platform consists of 4 core stacks plus 1 AI-enablement stack.
+
+> The web-search gateway and Bedrock model access used to be two separate
+> stacks (`VocWebSearchStack` and `BedrockAccessStack`). They were merged:
+> both must live in us-east-1, both are one-shot account enablement, and
+> neither depends on the core chain — and keeping them apart put the app at
+> six CloudFormation templates, one over the ceiling Workshop Studio allows.
+> The merged stack keeps the id `VocWebSearchStack` because that id determines
+> the export names `VocProcessingStack` and `VocApiStack` import.
+>
+> **Migrating an existing deployment:** run `cdk deploy --all` (the agreements
+> are recreated in `VocWebSearchStack` and re-run idempotently — Bedrock
+> agreements persist per account, so nothing is lost), then delete the old
+> stack with `cdk destroy BedrockAccessStack`. CDK does not remove a stack that
+> has been dropped from the app, so that second step is manual. No gateway is
+> recreated and there is no feature downtime.
 
 | Stack | Description | Dependencies |
 |-------|-------------|--------------|
@@ -177,8 +194,7 @@ The platform consists of 4 core stacks plus 2 optional ones:
 | `VocIngestionStack` | Plugin Lambdas, EventBridge schedules, SQS, Secrets | Core |
 | `VocProcessingStack` | Processor, Aggregator, Step Functions, Bedrock | Core, Ingestion |
 | `VocApiStack` | API Gateway, API Lambdas, Webhooks, WAF | Core, Ingestion, Processing |
-| `BedrockAccessStack` (optional) | Bedrock model access / Anthropic use case submission — created only when `anthropicUseCase` is set in `cdk.context.json` (see the conditional in `bin/voc-datalake.ts`) | None |
-| `VocWebSearchStack` (default-on, opt-out) | AgentCore Gateway for public web search — deployed by default, opt out via `enableWebSearch: false`; always deploys to us-east-1 (the connector only exists there). **Upgrade note:** existing non-us-east-1 deployments must bootstrap us-east-1 once (`cdk bootstrap aws://ACCOUNT_ID/us-east-1`) or set the opt-out flag | None |
+| `VocWebSearchStack` (AI enablement) | **Two independently switchable halves in one us-east-1 stack:** (a) the AgentCore Gateway for public web search — on by default, opt out via `enableWebSearch: false`; (b) Bedrock model access / Anthropic use-case submission — created only when `anthropicUseCase` is set in `cdk.context.json`. The stack is not created at all when both are off. Always deploys to us-east-1: the web-search connector exists only there, and `PutUseCaseForModelAccess` works only there. **Upgrade note:** existing non-us-east-1 deployments must bootstrap us-east-1 once (`cdk bootstrap aws://ACCOUNT_ID/us-east-1`) or set the opt-out flag | None |
 
 ### Deploy All Stacks
 
@@ -211,10 +227,12 @@ warning as a regression to investigate rather than noise to ignore.
 
 Due to dependencies, stacks should be deployed in this order:
 
-1. `VocCoreStack` (+ optional `BedrockAccessStack` / `VocWebSearchStack`, no dependencies)
-2. `VocIngestionStack`
-3. `VocProcessingStack`
-4. `VocApiStack`
+1. `VocWebSearchStack` (no dependencies — but it must come **before** Processing
+   and Api, which import its gateway exports when web search is enabled)
+2. `VocCoreStack` (no dependencies)
+3. `VocIngestionStack`
+4. `VocProcessingStack`
+5. `VocApiStack`
 
 The `cdk deploy --all` command handles this automatically.
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -77,9 +77,6 @@ Two things to know about the gate:
 - **There is no ESLint leg for the CDK TypeScript.** `bin/` and `lib/` are covered
   only by `typecheck:cdk`, so "lint is clean" says nothing about the CDK app.
 
-Note that `npm run lint` is only available from the **repo root** —
-`voc-datalake/package.json` has no `lint` script.
-
 ### Python Lambda Tests
 
 Lambda handlers have their own test suite using pytest:
@@ -420,7 +417,9 @@ there. Deployment is **on by default** (opt out with
 > `process.env.CDK_DEFAULT_REGION`, but **the CDK CLI overwrites that variable**
 > for the app subprocess from the resolved AWS environment — exporting it
 > yourself has no effect. Use `AWS_REGION` / `AWS_DEFAULT_REGION`, or
-> `--profile`. This matters here because `VocWebSearchStack` is the only
+> `--profile`. (Observed with the CDK CLI pinned in `voc-datalake/package.json`;
+> re-check if that behaviour ever changes, since the guidance below depends on
+> it.) This matters here because `VocWebSearchStack` is the only
 > region-pinned stack, so it is the only one whose region can differ from the
 > app's, and that difference is what triggers the cross-region wiring above.
 
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -48,22 +48,37 @@ Always run quality checks before deploying:
 
 ```bash
 # From project root
-npm run lint         # ESLint code quality
-npm run typecheck    # TypeScript type checking
-npm run test         # Run test suite
+npm run lint         # frontend + stream ESLint, and ruff over lambda/ + plugins/
+npm run typecheck    # frontend TypeScript only
+npm run test         # frontend Vitest only
 
 # Or run all at once
-npm run check        # lint + typecheck + test
+npm run check        # lint + typecheck:all + test + test:cdk + test:stream + test:backend
 ```
 
 ### What Each Check Does
 
-| Command | Description |
-|---------|-------------|
-| `npm run lint` | Runs ESLint to catch code quality issues |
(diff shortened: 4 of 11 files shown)


perrozzi:

This PR merges the us-east-1 `BedrockAccessStack` into `VocWebSearchStack` by converting it into a `BedrockModelAccess` construct, adds a `shouldDeployAiEnablement()` guard so the stack isn't synthesized empty, drops the workshop-only `-c enableWebSearch=false` synth, and updates docs plus a new 12-test template suite. The direction of the merge (agreements → gateway stack) is the right one — export names stay byte-identical so Processing/Api are untouched — and the cdk-nag suppressions were correctly converted from path-based to object-based now that the code lives one construct level deeper. Overall this looks clean; my comments are about invariant enforcement and artifact determinism rather than correctness of the merge. Note the `bin/voc-datalake.ts` hunks omit the downstream web-search wiring (lines ~114–213), so my comments on consumer wiring are inference, not verification.

### Should fix

- **Enforce the "both halves off" invariant where it can't be bypassed.** `lib/stacks/web-search-stack.ts` only documents it ("Callers must not construct this stack with BOTH halves off") and relies on an external guard in `lib/utils/ai-enablement-default.ts`. Per the repo's validate-at-boundaries rule, the constructor should `throw` when `!props.deployWebSearch && !props.anthropicUseCase`; the guard function then remains the deploy-time decision, and the failure mode (invalid `Resources: {}` after `convert-template.mjs` strips `CDKMetadata`) becomes impossible rather than merely commented.

- **`generate:workshop-artifacts` is now context-dependent.** `voc-datalake/package.json` drops `-c enableWebSearch=false` but pins nothing in its place, so the published artifact varies with the operator's local `cdk.context.json`: with `anthropicUseCase` set you get 22 resources plus 2 asset zips, without it 4 resources and none; with a local `enableWebSearch: false` and no `anthropicUseCase` the stack isn't constructed at all and `convert-template.mjs VocWebSearchStack …` gets a missing template. Worse, when `anthropicUseCase` is present the developer's `companyName` / `companyWebsite` / `useCases` strings are baked into a *published* template as custom-resource properties. Pin the toggles explicitly on that script (e.g. `-c enableWebSearch=true` and an empty/neutral `anthropicUseCase`) so the artifact is reproducible and carries no operator data.

- **New reachable state: `webSearchStack` defined with `gatewayArn`/`gatewayUrl` undefined.** Previously `webSearchStack !== undefined` implied a gateway existed; with `-c enableWebSearch=false` plus `anthropicUseCase` set that no longer holds. `web-search-stack.test.ts` asserts the properties are undefined, but nothing exercises the `bin` wiring or the Processing/Api guards in that combination. If any downstream line in `bin/voc-datalake.ts` branches on `if (webSearchStack)` rather than on the individual props (or uses `gatewayArn!`), that combination will synth an IAM statement or env var containing `undefined`. Please add a synth test covering that combination through the consumer stacks.

- **Confirm cdk-nag is clean without the dropped stack-level suppressions.** Removing `pluginSystemSuppressions`/`comprehendSuppressions`/`translateSuppressions` (with `applyToChildren: true`) is well argued, but it also removes blanket cover for the `AwsCustomResource` singleton provider. The remaining path-based suppression in `bedrock-access-stack.ts` builds `/${stack.stackName}/${customResourceId}/…` and matches it against `node.path`, which only coincides while `stackName` equals the stack's construct id — worth a comment or a synth assertion (e.g. no unsuppressed `AwsSolutions-IAM5`/`L1` errors) so a future `stackName` override doesn't silently drop suppressions.

### Nits

- `.kiro/steering/deployment.md` still opens with "The platform consists of 4 core stacks plus 2 optional ones" above the rewritten table; `docs/deployment.md` was updated to "plus 1 AI-enablement stack".
- `shouldDeployAiEnablement(deployWebSearch: boolean, anthropicUseCase: unknown)` should take `AnthropicUseCaseConfig | undefined`; the `!== undefined` check currently treats `null` as "deploy".
- `BedrockModelAccessProps.modelRegion` still falls back to `'us-west-2'` inside a stack pinned to us-east-1. Since the only caller always passes `env.region`, making it required removes a silent-wrong-region default.
- `bin/voc-datalake.ts` keeps importing `comprehendSuppressions` / `translateSuppressions`; if the removed `BedrockAccessStack` block was their only use, those imports are now dead.
- The export-name test pins the same-region `Fn::ImportValue` path only; when the app region differs, CDK routes the values through SSM instead, so the assertion doesn't cover the cross-region contract it protects.

---

perrozzi:
Adding a second commit with documentation fixes found while doing this work. Flagging the scope change before pushing.

**`63b0f2e` — docs: correct the quality-check gate, and document the model pin and region pin**

Three product-facing gaps, none of which are strictly required by the stack merge. Happy to split them into their own PR if you'd rather keep this focused.

1. **The Quality Checks section was inaccurate in ways that matter when triaging a red gate.** `npm run lint` is described as ESLint, but it also runs ruff over `lambda/` and `plugins/`, and it only exists at the repo root. `npm run typecheck` and `npm run test` cover the **frontend only** — the CDK, stream and Python suites are separate scripts that `check` chains. And `check` chains with `&&`, so the first failure hides every later step, which is exactly how the pre-existing `lint:python` failure obscures whether the CDK and backend suites pass. There is also **no ESLint leg for the CDK TypeScript** at all: `bin/` and `lib/` are covered only by `typecheck:cdk`, so "lint is clean" says nothing about the CDK app.

2. **`-c defaultModelId` shipped undocumented** (#269). It's the lever for accounts that can't invoke all of the per-surface defaults, which is common behind an AWS Private Marketplace. Documented the write-once semantics, the synth-time allowlist validation, that an absent flag creates zero resources, and the Converse probe — the agreement and availability APIs can report a model as available while inference still returns `AccessDeniedException`, so they aren't a reliable check.

3. **The `CDK_DEFAULT_REGION` trap.** The CDK CLI overwrites that variable for the app subprocess, so exporting it has no effect and the app region must come from `AWS_REGION`/`AWS_DEFAULT_REGION` or `--profile`. This only has visible consequences because `VocWebSearchStack` is the one region-pinned stack, so it's the only one whose region can differ from the app's — and that difference is what triggers the cross-region wiring already described in that section.

**Deliberately not documented here:** the packaging mechanics for the downstream workshop consumer. Those belong in the workshop repo, which is where the publish flow lives, and I've put them there instead. What remains in this repo is only the part that constrains contributions to it — that the template count is capped at five and the app is at exactly five, so adding a stack breaks packaging with the failure appearing at packaging time rather than at `cdk synth`.

Gates re-run after the docs commit: `typecheck:all` clean, `test:cdk` 132 passing. `lint:python` is still red on the 551 pre-existing ruff errors described in the PR body.

perrozzi:

Since my last review the changes are **documentation-only** — rewrites of the quality-gate tables and stack listings in `.kiro/steering/deployment.md` and `docs/deployment.md`, a new stack-count warning callout, a `CDK_DEFAULT_REGION` note, and a new "Pinning All AI Surfaces to One Model" section. None of the four Should-fix items from the previous review were addressed, and only one nit was.

### Resolved
- `.kiro/steering/deployment.md` no longer says "4 core stacks plus 2 optional ones"; the table and deploy-order list now match `docs/deployment.md`.

### Still open
- Constructor does not `throw` when both halves are off; the invariant is still only a comment plus the external `shouldDeployAiEnablement()` guard.
- `generate:workshop-artifacts` still pins no `enableWebSearch` / `anthropicUseCase`, so the published artifact varies with the operator's `cdk.context.json` (and can bake `companyName`/`useCases` into a published template).
- No synth test covering `webSearchStack` defined with `gatewayArn`/`gatewayUrl` undefined through the Processing/Api consumer wiring.
- No assertion or comment guarding the `/${stack.stackName}/${customResourceId}/…` suppression paths against a future `stackName` override.
- Nits: `shouldDeployAiEnablement(…, anthropicUseCase: unknown)` still untyped (`null` counts as "deploy"); `modelRegion` still defaults to `'us-west-2'`; `comprehendSuppressions`/`translateSuppressions` imports in `bin/voc-datalake.ts` now appear dead; export-name test still covers only the same-region `Fn::ImportValue` path.

### New findings
- **Scope creep in `docs/deployment.md`:** the new "Pinning All AI Surfaces to One Model" section (~40 lines documenting `Custom::ModelPin`, `ALLOWED_MODEL_IDS`, and a `bedrock-runtime converse` probe) documents a feature this PR does not touch. It's accurate as far as I can tell, but it makes the diff harder to review and to revert; consider splitting it into a docs-only PR. Same applies to the `CDK_DEFAULT_REGION` note, though that one at least explains behaviour this PR's region pinning depends on.
- **`docs/deployment.md` (region note) states a strong claim without a pointer:** "the CDK CLI overwrites that variable for the app subprocess from the resolved AWS environment — exporting it yourself has no effect." If that was determined empirically, worth naming the CLI version, since a future reader has no way to re-verify it cheaply and the guidance ("use `AWS_REGION`/`--profile`") is what operators will follow.
- **Duplicated caveat:** `docs/deployment.md` now says `npm run lint` is root-only both in the code-block comment area and again in a trailing note; the `.kiro` copy states it once. Minor, but the two files are now near-duplicates that will drift.

Nothing here blocks; the docs are a genuine improvement over what was there. The substantive items from the previous review remain the gating ones.

---

perrozzi:
Thanks — the determinism finding was a real regression and I'd missed it. Disposition of both reviews below, then pushing `831c55b`.

## Accepted and fixed

**Artifact determinism (Should fix #2).** Correct, and mine. Dropping `-c enableWebSearch=false` pinned nothing in its place, so the published artifact varied with the operator's local `cdk.context.json` — and in the worst combination the stack is never constructed and `convert-template.mjs VocWebSearchStack …` exits 1. Now pins `-c enableWebSearch=true`; command-line context takes precedence over `cdk.context.json`, so the artifact is reproducible again.

**Constructor now throws on both-halves-off (Should fix #1).** Agreed that a documented invariant guarding an *invalid template* belongs in code. `shouldDeployAiEnablement()` stays the deploy-time decision; the throw makes skipping it impossible.

**Consumer-wiring coverage (Should fix #3).** Added, though not in `bin` — that isn't unit-testable. The assertions live in `processing-stack-consolidated.test.ts`, whose helper already omits the web-search props, so it exercises exactly the combination you describe: no `WEB_SEARCH*` env vars, no `bedrock-agentcore` grants, and no `"undefined"` anywhere in the template. That pins the guard at `processing-stack-consolidated.ts:290`, which is what actually protects consumers. For the record I'd also verified `bin` uses only optional chaining (`webSearchStack?.gatewayUrl`) with no `if (webSearchStack)` branch and no non-null assertions.

**Suppression paths (Should fix #4).** Now built from `stack.node.id` rather than `stackName`, with a comment explaining that they're matched against `node.path` and the two only coincide while no explicit `stackName` is set.

**Both typing nits.** `shouldDeployAiEnablement` takes `AnthropicUseCaseConfig | undefined` so `null` no longer counts as "deploy". `modelRegion` is now **required** on both prop types — making it so immediately surfaced two test call sites that had been silently taking the `us-west-2` default, which is a fair argument for the change.

**Doc nits.** Removed the duplicated root-only `lint` note, and qualified the `CDK_DEFAULT_REGION` claim as observed against the pinned CLI.

Tests 132 → 135. Verified by synth: both halves off produces 4 templates with the stack correctly absent rather than throwing; the pinned flags produce 5 templates with the gateway and `Custom::ModelPin` present.

## One finding is incorrect

**"`comprehendSuppressions` / `translateSuppressions` imports are now dead."** They aren't — both are still used two lines apart at `bin/voc-datalake.ts:227-228`, for `processingStack` and `apiStack`. Only the `BedrockAccessStack` usage went away. No change made.

## One is misattributed rather than wrong

**"the developer's `companyName`/`companyWebsite`/`useCases` are baked into a published template."** Real, but **pre-existing** — not introduced here. The currently published `BedrockAccessStack.json` already carries the whole `formData` JSON inside the `SubmitAnthropicUseCase` `Fn::Sub`; this PR relocates it into `VocWebSearchStack.json` without changing what is emitted. Worth its own issue for forks whose operators use their own details, and I'd rather not widen this PR to cover it. Flagging because "pin a neutral `anthropicUseCase`" would also disable the model agreements, which the downstream workshop consumer depends on — so that specific remedy would break it.

## Declined

**Scope creep on the docs.** Fair in the abstract, but those doc gaps were explicitly requested alongside this work, and the packaging mechanics that *were* out of place have already been moved out to the consumer's own repo rather than documented here. Keeping the rest.

**Extending the export-name test to the cross-region SSM path.** That path is CDK's own cross-region reference machinery rather than anything this stack controls, so a test would pin CDK's implementation more than our contract.

perrozzi:

Since the last review the author made the substantive changes: the both-halves-off invariant now `throw`s in the `VocWebSearchStack` constructor (with a test), `modelRegion` is required on both the construct and the stack props, the cdk-nag suppression paths are built from `stack.node.id` with a comment explaining why, `shouldDeployAiEnablement` takes `AnthropicUseCaseConfig | undefined`, `generate:workshop-artifacts` now pins `-c enableWebSearch=true`, a Processing-side "gateway absent" synth test was added, and two docs nits were fixed. Most of the previous Should-fix list is addressed.

### Resolved
- Constructor now throws on both-halves-off; covered by `web-search-stack.test.ts` (`/both halves are disabled/`).
- `generate:workshop-artifacts` pins `enableWebSearch=true`, so the published artifact no longer inherits the operator's web-search flag.
- Suppression paths use `stack.node.id` and document why `stackName` would silently stop matching (`bedrock-access-stack.ts:205-217`).
- `modelRegion` default `'us-west-2'` removed; now required, with rationale.
- `shouldDeployAiEnablement` is typed; `null` no longer counts as "deploy".
- Processing consumer wiring with the gateway absent now has a synth test.
- `docs/deployment.md`: duplicated `npm run lint` caveat removed; the CDK-CLI region claim now names its provenance.

### Still open
- `generate:workshop-artifacts` does not pin `anthropicUseCase`, so a published template can still bake the operator's `companyName` / `useCases` and the 6 agreement custom resources (the "both on / 22 res." row in your own table).
- No equivalent gateway-absent synth test on the Api side (`api-stack.ts:786` guard).
- `comprehendSuppressions` / `translateSuppressions` imports in `bin/voc-datalake.ts` appear dead after dropping the `BedrockAccessStack` suppressions.
- Export-name test still exercises only the same-region `Fn::ImportValue` path, not the SSM cross-region path.

### New findings
- **`processing-stack-consolidated.test.ts` — `expect(rendered).not.toContain('undefined')` is a whole-template string match.** It will fire on any unrelated future occurrence of the substring (a plugin description, an env var default, inline code), producing a failure with no pointer to the guard it is meant to protect. Prefer asserting the specific absent env keys plus `IAM::Policy` statements over `bedrock-agentcore`, and drop or narrow the blanket check.
- **`modelRegion` is now required on `VocWebSearchStackProps` even when only the gateway half is deployed** (`web-search-stack.ts:64-68`). Gateway-only callers must supply a value that is never read — the test helper already papers over this with `modelRegion?: string`. Either default it at the stack layer while keeping it required on `BedrockModelAccessProps`, or move it into the `anthropicUseCase` shape so the type expresses the dependency.
- **Verify `shouldDeployWebSearch('true')` doesn't throw** (`package.json:27`): `-c` always yields a string, and the documented contract is that unrecognized values throw. The old script passed `'false'`, so the string-`'true'` branch may be newly exercised. Your synth ran end-to-end so this is likely fine, but a unit assertion for the string form would lock it in.

Nothing here is a blocker; the delta is a clear improvement.

---

perrozzi:
Both new findings were about code I added last round — accepted and fixed in `000d348`. Disposition of the rest below.

## Accepted and fixed

**The `not.toContain('undefined')` assertion was too broad.** Correct — it would have fired on any unrelated future occurrence and reported a failure with no pointer to the guard it protects. Split into two targeted assertions: no IAM policy action beginning with `bedrock-agentcore`, and no Lambda environment *value* containing `"undefined"`, which is the concrete failure mode if the guard at `processing-stack-consolidated.ts:290` were dropped.

**`modelRegion` was required even for gateway-only callers.** Also correct, and a regression I introduced last round by over-applying your earlier point. It's now optional on `VocWebSearchStackProps` and validated as required only when `anthropicUseCase` is set, so the constraint sits where the value is actually consumed. It stays required on `BedrockModelAccessProps`, where it always is. Still no default — a default would create the agreements in a region the app never calls Bedrock in. Two tests added: gateway-only without `modelRegion` must not throw; model-access without it must.

Tests 135 → 138. Re-verified across all four toggle combinations: pinned flags give 5 templates with gateway + 5 agreements; gateway-only 5 with gateway and no agreements; model-access-only 5 with agreements and no gateway; both off 4 with the stack absent.

## Already covered

**`shouldDeployWebSearch('true')`.** Good instinct, but the string form is already asserted — `lib/utils/web-search-default.test.ts:20` covers `'true'` and `'TRUE'` alongside the boolean, and the suite also pins the throw-on-unrecognized contract. No change needed.

## Repeat of a finding that is incorrect

**`comprehendSuppressions` / `translateSuppressions` are not dead.** This is the second time it's been raised and the second time I've checked: `grep -c comprehendSuppressions bin/voc-datalake.ts` returns **3** — the import plus uses at lines 227 and 228 for `processingStack` and `apiStack`. Only the `BedrockAccessStack` usage was removed. No change.

## Declined, with reasons

**No Api-side gateway-absent test.** The guard at `api-stack.ts:786` is identical in shape to the Processing one that is now covered, and there is no `api-stack.test.ts` harness — the stack takes ~25 constructor props, so standing one up for this single assertion is disproportionate to the risk. Reconsidered if an Api harness appears for other reasons.

**Export-name test on the cross-region SSM path.** That path is CDK's own cross-region reference machinery rather than anything this stack controls, so the test would pin CDK's implementation rather than our contract.

**Pinning `anthropicUseCase` in `generate:workshop-artifacts`.** Still declining, and worth being concrete about why: the downstream consumer *needs* the model agreements, and pinning a neutral or empty value would disable them.

Useful detail found while re-checking, though, which is the better remedy for the underlying concern: `skipUseCaseSubmission` gates **only** the use-case submission (`bedrock-access-stack.ts:165`) while the agreements are created unconditionally (`:293`). So `-c skipUseCaseSubmission=true` would keep the agreements while omitting the resource that carries the operator's `companyName`/`useCases` entirely. I'm not changing it here because whether fresh accounts still get access without the submission is untested, and this PR isn't the place to find out — but that's the lead for the separate issue rather than pinning a neutral use case.

perrozzi:

Since my last review the author addressed two of the three new findings: the blanket `not.toContain('undefined')` check was split into a scoped IAM-action assertion plus a Lambda-env assertion, and `modelRegion` became optional on `VocWebSearchStackProps` with a runtime guard that throws when `anthropicUseCase` is set without it (two new tests cover both directions). No other files changed.

### Resolved
- `processing-stack-consolidated.test.ts`: whole-template `not.toContain` replaced with scoped assertions over IAM policy actions and Lambda environment values.
- `web-search-stack.ts`: gateway-only callers no longer have to invent a `modelRegion`; the dependency is validated with a clear error and tested both ways.
- `web-search-stack.test.ts`: the `modelRegion?: string` papering-over in the `synth()` helper is gone — the signature is now plain `Omit<VocWebSearchStackProps, 'env'>`.

### Still open
- `generate:workshop-artifacts` does not pin `anthropicUseCase`, so a published template can still bake the operator's `companyName` / `useCases`.
- No gateway-absent synth test on the Api side (`api-stack.ts:786`).
- `comprehendSuppressions` / `translateSuppressions` imports in `bin/voc-datalake.ts` look dead after the `BedrockAccessStack` suppressions were dropped — confirm they're still referenced further down.
- Export-name test still only covers the same-region `Fn::ImportValue` path, not the SSM cross-region path.
- No unit assertion that `shouldDeployWebSearch('true')` (string form, now used by `package.json:27`) doesn't throw.

### New findings
- **`processing-stack-consolidated.test.ts` — the IAM assertion only scans `AWS::IAM::Policy`.** Statements can also land in `AWS::IAM::Role`'s inline `Policies` or an `AWS::IAM::ManagedPolicy`; if the web-search grant ever moves there, the test passes while the guard is broken. Cheap fix: collect statements from all three resource types.
- **`modelRegion` dependency is now expressed at runtime rather than in the type.** Acceptable as-is (the only caller always passes `env.region`), but a discriminated union or nesting `modelRegion` inside the `anthropicUseCase` shape would make the invalid combination unrepresentable instead of a synth-time throw. Non-blocking.

Otherwise the delta is clean.

---

perrozzi:
**Correction to my previous comment** — the `skipUseCaseSubmission` suggestion I offered as the "better remedy" for the operator-data concern is wrong, and I'd rather not leave it standing in this thread.

I checked the [Bedrock model access docs](https://docs.aws.amazon.com/bedrock/latest/userguide/model-access.html). `PutUseCaseForModelAccess` is a documented **prerequisite** for Anthropic access, sequenced explicitly *before* `CreateFoundationModelAgreement` in the SDK/CLI flow. So `-c skipUseCaseSubmission=true` would leave a genuinely first-time account holding agreements that don't confer invocability — it isn't a way to drop the operator data. My claim that only the submission was severable was based on reading the code's control flow, not the requirement it implements.

Two other things from those docs that matter for the follow-up issue:

- **There is no "neutral" form value.** A company website URL and a use-case description are both required (`useCases` up to 8192 chars), so genericizing the payload isn't possible either.
- **The requirement is once per account _or once per AWS Organization_, and a root submission is inherited by member accounts** (opt-in regions aside). That's the actual lever: if the org vending the accounts has already submitted, the template needn't submit at all — removing the resource rather than sanitizing it.

So the follow-up issue should ask "is the submission needed at all here, given org inheritance?" rather than "pin or genericize the form data". Filing it separately as planned.

Also worth noting for anyone reading this later: those docs state that model access is now enabled by default given AWS Marketplace permissions, with the first invocation auto-initiating the subscription. That makes the explicit agreement-creation path largely legacy for ordinary accounts — but not for accounts behind a private marketplace, where auto-subscription fails on the missing prerequisite, which is exactly the case this stack exists to handle.

No code change; this doesn't affect the diff.

PR #276, opened 2026-08-07
fix: resolve active Bedrock model for persona calls

### LINKED ISSUE #273: [Bug]: BEDROCK_MODEL_ID used directly at two call sites, bypassing model resolution
## Summary

Two live call sites import `BEDROCK_MODEL_ID` from `shared/aws.py` and use it directly instead of resolving through `shared/model_config.py`. `BEDROCK_MODEL_ID` is the **last** entry in the documented precedence chain, so using it directly skips the three levels above it — the per-surface picker, the `settings.model_id` global override, and `SURFACE_DEFAULTS`.

One site causes a real (silently swallowed) `AccessDeniedException`; the other records incorrect model provenance.

Both were observed in a fresh Workshop Studio event account deployed with `-c defaultModelId=global.anthropic.claude-sonnet-4-6`, where Sonnet 5 has no Bedrock model access.

## Documented precedence (`shared/model_config.py`)

```
per-surface override (settings.surfaces[S])
  > legacy global override (settings.model_id)
  > built-in default for S (SURFACE_DEFAULTS[S])
  > BEDROCK_MODEL_ID          <- last resort
```

`shared/aws.py`:

```python
BEDROCK_MODEL_ID = "global.anthropic.claude-sonnet-5"
```

---

## Site 1 — `lambda/shared/avatar.py:129` (functional failure)

```python
def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
    from shared.aws import BEDROCK_MODEL_ID
    ...
    response = bedrock_client.invoke_model(modelId=BEDROCK_MODEL_ID, ...)
```

Fails on every call, for every persona, and the `except` falls back to a static template so nothing surfaces in the UI:

```
[PERSONA_AVATAR] Generating image prompt with Claude for Marcus Webb
[PERSONA_AVATAR] LLM prompt generation failed: An error occurred (AccessDeniedException)
  when calling the InvokeModel operation: anthropic.claude-sonnet-5 is not available for
  this account. ... using fallback
[PERSONA_AVATAR] Generated prompt: Professional headshot of a Mid-level office professional
  (project coordinator or analyst), friendly expression, soft studio lighting, ...
```

Reproduced independently in `voc-projects-api` via `POST /projects/{id}/personas/{pid}/regenerate-avatar`, so it is not specific to the persona-generator job.

**Effect:** every avatar prompt degrades to an occupation-only template instead of the persona-tailored prompt the feature is designed around, plus a guaranteed `AccessDeniedException` per persona in logs.

**Fix:**

```python
from shared.model_config import get_active_model_id
model_id = get_active_model_id(surface='utility')
response = bedrock_client.invoke_model(modelId=model_id, ...)
```

---

## Site 2 — `lambda/api/projects.py:488` (incorrect provenance)

```python
'llm_metadata': {
    'model': BEDROCK_MODEL_ID,
    'prompt_version': '2.0.0',
    'generation_time_ms': llm_time
}
```

This records a constant rather than the model that actually ran. Verified on a real generated persona:

Stored in `voc-projects`:

```json
{ "model": "global.anthropic.claude-sonnet-5",
  "prompt_version": "2.0.0",
  "generation_time_ms": 584545 }
```

What actually ran, per CloudWatch:

```
```

**Effect:** `llm_metadata.model` is wrong whenever the resolved model differs from the constant — i.e. whenever the picker or the global pin is in use. Anyone auditing which model produced a persona gets the wrong answer, and it names a model the account cannot even invoke.

**Fix:** record the resolved id, e.g. `get_active_model_id(surface='documents')`, or have `converse_chain` return the model it used and stamp that.

---

## Also noted (no impact)

`lambda/research/research_step_handler.py:43` defines `MODEL_ID = BEDROCK_MODEL_ID` but never uses it — invocation goes through `shared.converse.converse`, which resolves correctly. Dead constant; safe to delete, and worth deleting so it isn't copied.

## Related standard

From the repo's own coding standards: *"Never hardcode model ids — resolution goes through `shared/model_config.py`."* A lint rule or a test asserting that `BEDROCK_MODEL_ID` is imported only by `model_config.py` would prevent regressions.

slegarraga:
Closes #273

## Summary

Stops using the `BEDROCK_MODEL_ID` fallback constant at two live persona call sites and resolves the active model through the documented precedence chain instead.

- `shared/avatar.py`: avatar prompt generation resolves the active model for the `utility` surface before invoking Bedrock. Resolution sits inside the existing fallback boundary: a transient picker (DynamoDB) problem degrades to the static fallback prompt exactly like an invocation failure, matching the pre-picker behavior where a constant could never fail. Prompt-config loading stays above the boundary so prompt-file problems remain configuration errors.
- `api/projects.py`: persona generation pins the resolved `documents` model via `converse_chain(model_id=...)` and records it in `llm_metadata.model`. The `surface='documents'` argument stays on the call so surface-derived logging keeps its dimension; the pin decides which model is invoked.
- `shared/converse.py`: adds the public `converse_chain(model_id=...)` parameter (scope note: this is the one shared-module API change). When set, it pins the model for every step; a step-level `model`/`model_id` applies only when the chain is not pinned. Step keys that lose to an explicit model emit a single combined warning naming everything dropped: no silent overrides, no job failures. No persona step sets a model or surface today, so current chains are unaffected.
- `research/research_step_handler.py`: removes the unused `MODEL_ID` alias so the dead fallback import is not copied into future handlers.

## IAM note

The persona roles (`ProjectsLambdaRole`, `PersonaGeneratorRole`, `PersonaImporterRole`) already grant `bedrock:InvokeModel` on the full allowlist, which is what the runtime picker can now return for these surfaces. The new CDK regression test pins that lockstep; no stack change was required.

## Tests

Updated avatar tests to assert the resolved model id and the `utility` surface lookup. Added persona tests for the resolved-model metadata stamp, the chain pin, and tolerance of inert step overrides. Added a chain test asserting that a step with both `model` and `surface` under a pin logs exactly one combined warning.

Python:

```bash
pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py \
  lambda/api/test/test_projects.py lambda/shared/test/test_converse.py \
  lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-cov
```

Result: 174 passed, 0 failed.

CDK/vitest (`lib/utils/model-allowlist.ts` already exists on `development`; this suite compiles that import):

```bash
npm test -- lib/stacks/api-stack.test.ts
```

Result: 56 passed, 0 failed, including the `persona Bedrock IAM grants` assertions.

Signed off with DCO.

### DIFF
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -564,7 +564,7 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
     # lookups (chain + metadata stamp) could drift if an admin repointed the
     # 'documents' surface mid-run; one resolution cannot.
     resolved_model_id = get_active_model_id(surface='documents')
-    logger.info(f"[PERSONA] Resolved persona model: {resolved_model_id}")
+    logger.info("[PERSONA] Resolved persona model", extra={'model_id': resolved_model_id})
 
     # Get feedback data
     try:
@@ -636,16 +636,6 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
                 response_language=filters.get('response_language'),
                 sample_chars=context_budget,
             )
-            model_overrides = [
-                step.get('step_name', f'persona_step_{idx}')
-                for idx, step in enumerate(chain_steps, 1)
-                if step.get('model_id') or step.get('model')
-            ]
-            if model_overrides:
-                raise ValueError(
-                    'Persona generation steps must not override model_id: '
-                    + ', '.join(model_overrides)
-                )
             logger.info(f"[PERSONA] Built {len(chain_steps)} chain steps")
         except Exception as e:
             logger.error(f"[PERSONA] Failed to build chain steps: {e}")
@@ -678,7 +668,6 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
             results = converse_chain(
                 chain_steps,
                 progress_callback=lambda p, s: update_progress(p, s),
-                surface='documents',
                 model_id=resolved_model_id,
             )
             logger.info(f"[PERSONA] LLM chain returned {len(results)} results")
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -668,6 +668,7 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
             results = converse_chain(
                 chain_steps,
                 progress_callback=lambda p, s: update_progress(p, s),
+                surface='documents',
                 model_id=resolved_model_id,
             )
             logger.info(f"[PERSONA] LLM chain returned {len(results)} results")
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -12,7 +12,7 @@ from botocore.exceptions import ClientError
 
 # Shared module imports
 from shared.logging import logger, tracer, metrics
-from shared.aws import get_dynamodb_resource, get_bedrock_client, BEDROCK_MODEL_ID
+from shared.aws import get_dynamodb_resource, get_bedrock_client
 from shared.api import validate_days, MAX_PERSONAS_PER_GENERATION
 from shared.persona_context import personas_prompt_context
 from shared.converse import converse_chain
@@ -38,7 +38,7 @@ from shared.feedback import (
     get_feedback_statistics,
     truncate_feedback_context,
 )
-from shared.model_config import surface_context_window_tokens
+from shared.model_config import get_active_model_id, surface_context_window_tokens
 from shared.avatar import (
     generate_persona_avatar as _generate_persona_avatar,
     get_avatar_cdn_url,
@@ -558,6 +558,14 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
         f"[PERSONA] Context budget: {context_budget} chars, fetch limit: {fetch_limit} items"
     )
 
+    # Resolve the persona-synthesis model ONCE per invocation and pin it on
+    # the chain below (converse_chain forwards it to every step) so the
+    # stored llm_metadata records exactly what was invoked. Two independent
+    # lookups (chain + metadata stamp) could drift if an admin repointed the
+    # 'documents' surface mid-run; one resolution cannot.
+    resolved_model_id = get_active_model_id(surface='documents')
+    logger.info(f"[PERSONA] Resolved persona model: {resolved_model_id}")
+
     # Get feedback data
     try:
         feedback_items = get_feedback_context(filters, limit=fetch_limit)
@@ -628,6 +636,16 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
                 response_language=filters.get('response_language'),
                 sample_chars=context_budget,
             )
+            model_overrides = [
+                step.get('step_name', f'persona_step_{idx}')
+                for idx, step in enumerate(chain_steps, 1)
+                if step.get('model_id') or step.get('model')
+            ]
+            if model_overrides:
+                raise ValueError(
+                    'Persona generation steps must not override model_id: '
+                    + ', '.join(model_overrides)
+                )
             logger.info(f"[PERSONA] Built {len(chain_steps)} chain steps")
         except Exception as e:
             logger.error(f"[PERSONA] Failed to build chain steps: {e}")
@@ -657,7 +675,12 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
         update_progress(20, 'executing_llm_chain')
         
         try:
-            results = converse_chain(chain_steps, progress_callback=lambda p, s: update_progress(p, s), surface='documents')
+            results = converse_chain(
+                chain_steps,
+                progress_callback=lambda p, s: update_progress(p, s),
+                surface='documents',
+                model_id=resolved_model_id,
+            )
             logger.info(f"[PERSONA] LLM chain returned {len(results)} results")
         except Exception as e:
             logger.error(f"[PERSONA] LLM chain execution failed: {e}")
@@ -790,7 +813,7 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
                 'created_at': now,
                 'updated_at': now,
                 'llm_metadata': {
-                    'model': BEDROCK_MODEL_ID,
+                    'model': resolved_model_id,
                     'prompt_version': PERSONA_PROMPT_VERSION,
                     'generation_time_ms': llm_time
                 },
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -289,6 +289,15 @@ class TestGenerateAvatarPromptWithLlm:
 class TestGeneratePersonasModelMetadata:
     """Tests for persona generation model metadata."""
 
+    @staticmethod
+    def _put_item_payload(table):
+        call = table.put_item.call_args
+        assert call is not None, 'expected put_item to be called'
+        if call.kwargs.get('Item') is not None:
+            return call.kwargs['Item']
+        assert call.args, 'expected put_item Item as kwargs or first positional arg'
+        return call.args[0]
+
     def test_stores_resolved_model_id(self):
         """Stamps the resolved model, not the global fallback constant."""
         import projects
@@ -307,24 +316,25 @@ class TestGeneratePersonasModelMetadata:
                 patch('projects.converse_chain', return_value=[
                     json.dumps([{'name': 'Ada'}]),
                 ]) as mock_chain, \
-                patch('projects.get_active_model_id', return_value='resolved-model') as mock_model_id:
+                patch(
+                    'projects.get_active_model_id',
+                    return_value='resolved-model',
+                ) as mock_model_id:
             projects.generate_personas(
                 'proj-1',
                 {'persona_count': 1, 'generate_avatars': False},
             )
 
-        assert table.put_item.called
-        stored = table.put_item.call_args.kwargs['Item']
+        stored = self._put_item_payload(table)
         assert stored['llm_metadata']['model'] == 'resolved-model'
         # One resolution per invocation, and the chain is PINNED to it: the
         # model converse() invokes is the one llm_metadata records.
         mock_model_id.assert_called_once_with(surface='documents')
         assert mock_chain.call_args.kwargs['model_id'] == 'resolved-model'
 
-    def test_rejects_persona_step_model_override(self):
-        """llm_metadata records one model, so persona steps must not override it."""
+    def test_persona_chain_pin_tolerates_step_model_override(self):
+        """The resolved chain pin keeps metadata aligned without failing the job."""
         import projects
-        from shared.exceptions import ServiceError
 
         table = MagicMock()
         table.query.return_value = {'Items': []}
@@ -340,12 +350,16 @@ class TestGeneratePersonasModelMetadata:
                         'model': 'other-model',
                     },
                 ]), \
-                patch('projects.converse_chain') as mock_chain, \
+                patch('projects.converse_chain', return_value=[
+                    json.dumps([{'name': 'Ada'}]),
+                ]) as mock_chain, \
                 patch('projects.get_active_model_id', return_value='resolved-model'):
-            with pytest.raises(ServiceError, match='Failed to generate personas'):
-                projects.generate_personas(
-                    'proj-1',
-                    {'persona_count': 1, 'generate_avatars': False},
-                )
+            projects.generate_personas(
+                'proj-1',
+                {'persona_count': 1, 'generate_avatars': False},
+            )
 
-        mock_chain.assert_not_called()
+        stored = self._put_item_payload(table)
+        assert stored['llm_metadata']['model'] == 'resolved-model'
+        assert mock_chain.call_args.kwargs['model_id'] == 'resolved-model'
+        assert 'surface' not in mock_chain.call_args.kwargs
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -290,13 +290,16 @@ class TestGeneratePersonasModelMetadata:
     """Tests for persona generation model metadata."""
 
     @staticmethod
-    def _put_item_payload(table):
-        call = table.put_item.call_args
-        assert call is not None, 'expected put_item to be called'
-        if call.kwargs.get('Item') is not None:
-            return call.kwargs['Item']
-        assert call.args, 'expected put_item Item as kwargs or first positional arg'
-        return call.args[0]
+    def _put_item_payloads(table):
+        """Every put_item payload, not just the final one: a persona_count > 1
+        run writes one row per persona and asserting only the last call would
+        silently skip the earlier ones."""
+        calls = table.put_item.call_args_list
+        assert calls, 'expected put_item to be called'
+        return [
+            c.kwargs['Item'] if c.kwargs.get('Item') is not None else c.args[0]
+            for c in calls
+        ]
 
     def test_stores_resolved_model_id(self):
         """Stamps the resolved model, not the global fallback constant."""
@@ -325,8 +328,8 @@ class TestGeneratePersonasModelMetadata:
                 {'persona_count': 1, 'generate_avatars': False},
             )
 
-        stored = self._put_item_payload(table)
-        assert stored['llm_metadata']['model'] == 'resolved-model'
+        for stored in self._put_item_payloads(table):
+            assert stored['llm_metadata']['model'] == 'resolved-model'
         # One resolution per invocation, and the chain is PINNED to it: the
         # model converse() invokes is the one llm_metadata records.
         mock_model_id.assert_called_once_with(surface='documents')
@@ -359,7 +362,7 @@ class TestGeneratePersonasModelMetadata:
                 {'persona_count': 1, 'generate_avatars': False},
             )
 
-        stored = self._put_item_payload(table)
-        assert stored['llm_metadata']['model'] == 'resolved-model'
+        for stored in self._put_item_payloads(table):
+            assert stored['llm_metadata']['model'] == 'resolved-model'
         assert mock_chain.call_args.kwargs['model_id'] == 'resolved-model'
-        assert 'surface' not in mock_chain.call_args.kwargs
+        assert mock_chain.call_args.kwargs['surface'] == 'documents'
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -290,12 +290,14 @@ class TestGeneratePersonasModelMetadata:
     """Tests for persona generation model metadata."""
 
     @staticmethod
-    def _put_item_payloads(table):
+    def _put_item_payloads(table, expected_count):
         """Every persona put_item payload, not just the final one: a
         persona_count > 1 run writes one row per persona and asserting only
         the last call would silently skip the earlier ones. Filtered on
         llm_metadata so a hypothetical future non-persona row fails as
-        'no persona rows' instead of a bare KeyError."""
+        'no persona rows' instead of a bare KeyError, and pinned to
+        expected_count so a change that stops stamping llm_metadata on a
+        persona row fails loudly instead of dropping it from the list."""
         calls = table.put_item.call_args_list
         assert calls, 'expected put_item to be called'
         payloads = [
@@ -303,7 +305,10 @@ class TestGeneratePersonasModelMetadata:
             for c in calls
         ]
         persona_rows = [p for p in payloads if p.get('llm_metadata')]
-        assert persona_rows, 'expected put_item to write at least one persona row'
+        assert len(persona_rows) == expected_count, (
+            f'expected {expected_count} persona rows, '
+            f'found {len(persona_rows)}'
+        )
         return persona_rows
 
     def test_stores_resolved_model_id(self):
@@ -333,7 +338,8 @@ class TestGeneratePersonasModelMetadata:
                 {'persona_count': 1, 'generate_avatars': False},
             )
 
-        for stored in self._put_item_payloads(table):
+        # persona_count is 1 in this run: exactly one persona row expected.
+        for stored in self._put_item_payloads(table, 1):
             assert stored['llm_metadata']['model'] == 'resolved-model'
         # One resolution per invocation, and the chain is PINNED to it: the
         # model converse() invokes is the one llm_metadata records.
@@ -367,7 +373,8 @@ class TestGeneratePersonasModelMetadata:
                 {'persona_count': 1, 'generate_avatars': False},
             )
 
-        for stored in self._put_item_payloads(table):
+        # persona_count is 1 in this run: exactly one persona row expected.
+        for stored in self._put_item_payloads(table, 1):
             assert stored['llm_metadata']['model'] == 'resolved-model'
         assert mock_chain.call_args.kwargs['model_id'] == 'resolved-model'
         assert mock_chain.call_args.kwargs['surface'] == 'documents'
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -284,3 +284,68 @@ class TestGenerateAvatarPromptWithLlm:
         result = generate_avatar_prompt_with_llm(persona_data, mock_bedrock)
         
         assert 'Professional headshot' in result
+
+
+class TestGeneratePersonasModelMetadata:
+    """Tests for persona generation model metadata."""
+
+    def test_stores_resolved_model_id(self):
+        """Stamps the resolved model, not the global fallback constant."""
+        import projects
+
+        table = MagicMock()
+        table.query.return_value = {'Items': []}
+        feedback_items = [{'feedback_id': 'f1', 'source_platform': 'web'}]
+
+        with patch('projects.projects_table', table), \
+                patch('projects.get_feedback_context', return_value=feedback_items), \
+                patch('projects.format_feedback_for_llm', return_value='feedback'), \
+                patch('projects.get_feedback_statistics', return_value={'total': 1}), \
+                patch('projects.get_persona_generation_steps', return_value=[
+                    {'step_name': projects.PERSONA_SYNTHESIS_STEP},
+                ]), \
+                patch('projects.converse_chain', return_value=[
+                    json.dumps([{'name': 'Ada'}]),
+                ]) as mock_chain, \
+                patch('projects.get_active_model_id', return_value='resolved-model') as mock_model_id:
+            projects.generate_personas(
+                'proj-1',
+                {'persona_count': 1, 'generate_avatars': False},
+            )
+
+        assert table.put_item.called
+        stored = table.put_item.call_args.kwargs['Item']
+        assert stored['llm_metadata']['model'] == 'resolved-model'
+        # One resolution per invocation, and the chain is PINNED to it: the
+        # model converse() invokes is the one llm_metadata records.
+        mock_model_id.assert_called_once_with(surface='documents')
+        assert mock_chain.call_args.kwargs['model_id'] == 'resolved-model'
+
+    def test_rejects_persona_step_model_override(self):
+        """llm_metadata records one model, so persona steps must not override it."""
+        import projects
+        from shared.exceptions import ServiceError
+
+        table = MagicMock()
+        table.query.return_value = {'Items': []}
+        feedback_items = [{'feedback_id': 'f1', 'source_platform': 'web'}]
+
+        with patch('projects.projects_table', table), \
+                patch('projects.get_feedback_context', return_value=feedback_items), \
+                patch('projects.format_feedback_for_llm', return_value='feedback'), \
+                patch('projects.get_feedback_statistics', return_value={'total': 1}), \
+                patch('projects.get_persona_generation_steps', return_value=[
+                    {
+                        'step_name': projects.PERSONA_SYNTHESIS_STEP,
+                        'model': 'other-model',
+                    },
+                ]), \
+                patch('projects.converse_chain') as mock_chain, \
+                patch('projects.get_active_model_id', return_value='resolved-model'):
+            with pytest.raises(ServiceError, match='Failed to generate personas'):
+                projects.generate_personas(
+                    'proj-1',
+                    {'persona_count': 1, 'generate_avatars': False},
+                )
+
+        mock_chain.assert_not_called()
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -291,15 +291,20 @@ class TestGeneratePersonasModelMetadata:
 
     @staticmethod
     def _put_item_payloads(table):
-        """Every put_item payload, not just the final one: a persona_count > 1
-        run writes one row per persona and asserting only the last call would
-        silently skip the earlier ones."""
+        """Every persona put_item payload, not just the final one: a
+        persona_count > 1 run writes one row per persona and asserting only
+        the last call would silently skip the earlier ones. Filtered on
+        llm_metadata so a hypothetical future non-persona row fails as
+        'no persona rows' instead of a bare KeyError."""
         calls = table.put_item.call_args_list
         assert calls, 'expected put_item to be called'
-        return [
+        payloads = [
             c.kwargs['Item'] if c.kwargs.get('Item') is not None else c.args[0]
             for c in calls
         ]
+        persona_rows = [p for p in payloads if p.get('llm_metadata')]
+        assert persona_rows, 'expected put_item to write at least one persona row'
+        return persona_rows
 
     def test_stores_resolved_model_id(self):
         """Stamps the resolved model, not the global fallback constant."""
--- a/voc-datalake/lambda/api/test/test_projects.py
+++ b/voc-datalake/lambda/api/test/test_projects.py
@@ -294,10 +294,9 @@ class TestGeneratePersonasModelMetadata:
         """Every persona put_item payload, not just the final one: a
         persona_count > 1 run writes one row per persona and asserting only
         the last call would silently skip the earlier ones. Filtered on
-        llm_metadata so a hypothetical future non-persona row fails as
-        'no persona rows' instead of a bare KeyError, and pinned to
-        expected_count so a change that stops stamping llm_metadata on a
-        persona row fails loudly instead of dropping it from the list."""
+        llm_metadata so a hypothetical future non-persona row is ignored,
+        while expected_count makes any missing persona metadata stamp fail
+        loudly as a row-count mismatch."""
         calls = table.put_item.call_args_list
         assert calls, 'expected put_item to be called'
         payloads = [
--- a/voc-datalake/lambda/research/research_step_handler.py
+++ b/voc-datalake/lambda/research/research_step_handler.py
@@ -12,7 +12,7 @@ from boto3.dynamodb.conditions import Key
 
 # Shared module imports
 from shared.logging import logger, tracer
-from shared.aws import get_dynamodb_resource, BEDROCK_MODEL_ID
+from shared.aws import get_dynamodb_resource
 from shared.api import api_handler
 from shared.converse import converse, BedrockThrottlingError
 from shared.persona_context import personas_prompt_context
@@ -47,13 +47,15 @@ PROJECTS_TABLE = os.environ.get('PROJECTS_TABLE', '')
 feedback_table = None
 projects_table = None
 
-MODEL_ID = BEDROCK_MODEL_ID
+
 def _get_feedback_table():
     """Get feedback table, initializing if needed."""
     global feedback_table
     if feedback_table is None:
         feedback_table = get_feedback_table()
     return feedback_table
+
+
 def _get_projects_table():
     """Get projects table, initializing if needed."""
     global projects_table
--- a/voc-datalake/lambda/shared/avatar.py
+++ b/voc-datalake/lambda/shared/avatar.py
@@ -12,6 +12,8 @@ import boto3
 
 from shared.logging import logger, tracer
 from shared.prompts import get_avatar_prompt_config, format_prompt
+# Lightweight runtime picker; it lazy-loads DynamoDB only when called.
+from shared.model_config import get_active_model_id
 
 # `shared.cloudfront_signing` (and through it `cryptography`) is imported LAZILY
 # inside get_avatar_cdn_url — the only function here that signs. The rest of this
@@ -214,27 +216,27 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
         'soft studio lighting, neutral background, photorealistic'
     )
 
+    model_id = get_active_model_id(surface='utility')
+
+    # Load prompt config before the Bedrock fallback boundary, preserving the
+    # old behaviour: prompt file problems are configuration errors, while model
+    # invocation problems degrade to a deterministic static prompt.
+    config = get_avatar_prompt_config()
+    system_prompt = config.get('system_prompt', '')
+    user_template = config.get('user_prompt_template', '')
+    fallback_template = config.get('fallback_prompt_template', fallback_template)
+
+    user_msg = format_prompt(
+        user_template,
+        name=name,
+        tagline=tagline,
+        age_range=age_range,
+        occupation=occupation,
+        location=location,
+        bio=bio[:300] if bio else 'N/A'
+    )
+
     try:
-        from shared.model_config import get_active_model_id
-
-        model_id = get_active_model_id(surface='utility')
-
-        # Load prompt config from external file
-        config = get_avatar_prompt_config()
-        system_prompt = config.get('system_prompt', '')
-        user_template = config.get('user_prompt_template', '')
-        fallback_template = config.get('fallback_prompt_template', fallback_template)
-
-        user_msg = format_prompt(
-            user_template,
-            name=name,
-            tagline=tagline,
-            age_range=age_range,
-            occupation=occupation,
-            location=location,
-            bio=bio[:300] if bio else 'N/A'
-        )
-
         request_body = {
             'anthropic_version': 'bedrock-2023-05-31',
             'max_tokens': config.get('max_tokens', 200),
--- a/voc-datalake/lambda/shared/avatar.py
+++ b/voc-datalake/lambda/shared/avatar.py
@@ -216,8 +216,6 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
         'soft studio lighting, neutral background, photorealistic'
     )
 
-    model_id = get_active_model_id(surface='utility')
-
     # Load prompt config before the Bedrock fallback boundary, preserving the
     # old behaviour: prompt file problems are configuration errors, while model
     # invocation problems degrade to a deterministic static prompt.
@@ -237,6 +235,11 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
     )
 
     try:
+        # Inside the boundary on purpose: before the per-surface picker this was
+        # a constant that could never fail, so a transient picker (DynamoDB)
+        # problem must degrade to the static fallback prompt exactly like an
+        # invocation failure instead of becoming a new hard-failure path.
+        model_id = get_active_model_id(surface='utility')
         request_body = {
             'anthropic_version': 'bedrock-2023-05-31',
             'max_tokens': config.get('max_tokens', 200),
--- a/voc-datalake/lambda/shared/avatar.py
+++ b/voc-datalake/lambda/shared/avatar.py
@@ -202,8 +202,6 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
     Returns:
         Generated image prompt string
     """
-    from shared.aws import BEDROCK_MODEL_ID
-    
     name = persona_data.get('name', 'Unknown')
     tagline = persona_data.get('tagline', '')
     identity = persona_data.get('identity', {})
@@ -211,23 +209,32 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
     age_range = identity.get('age_range', '')
     occupation = identity.get('occupation', '')
     location = identity.get('location', '')
-    
-    # Load prompt config from external file
-    config = get_avatar_prompt_config()
-    system_prompt = config.get('system_prompt', '')
-    user_template = config.get('user_prompt_template', '')
-    
-    user_msg = format_prompt(
-        user_template,
-        name=name,
-        tagline=tagline,
-        age_range=age_range,
-        occupation=occupation,
-        location=location,
-        bio=bio[:300] if bio else 'N/A'
+    fallback_template = (
+        'Professional headshot of a {occupation}, friendly expression, '
+        'soft studio lighting, neutral background, photorealistic'
     )
 
     try:
+        from shared.model_config import get_active_model_id
+
+        model_id = get_active_model_id(surface='utility')
+
+        # Load prompt config from external file
+        config = get_avatar_prompt_config()
+        system_prompt = config.get('system_prompt', '')
+        user_template = config.get('user_prompt_template', '')
+        fallback_template = config.get('fallback_prompt_template', fallback_template)
+
+        user_msg = format_prompt(
+            user_template,
+            name=name,
+            tagline=tagline,
+            age_range=age_range,
+            occupation=occupation,
+            location=location,
+            bio=bio[:300] if bio else 'N/A'
+        )
+
         request_body = {
             'anthropic_version': 'bedrock-2023-05-31',
             'max_tokens': config.get('max_tokens', 200),
@@ -236,7 +243,7 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
         }
         
         response = bedrock_client.invoke_model(
-            modelId=BEDROCK_MODEL_ID,
+            modelId=model_id,
             contentType='application/json',
             accept='application/json',
             body=json.dumps(request_body)
@@ -250,8 +257,10 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
         
         return result['content'][0]['text'].strip()
     except Exception as e:
-        logger.warning(f"[PERSONA_AVATAR] LLM prompt generation failed: {e}, using fallback")
-        fallback_template = config.get('fallback_prompt_template', 'Professional headshot of a {occupation}, friendly expression, soft studio lighting, neutral background, photorealistic')
+        logger.warning(
+            "[PERSONA_AVATAR] LLM prompt generation failed "
+            f"({type(e).__name__}: {e}), using fallback"
+        )
         return format_prompt(fallback_template, occupation=occupation or 'professional')
 
 
--- a/voc-datalake/lambda/shared/avatar.py
+++ b/voc-datalake/lambda/shared/avatar.py
@@ -263,8 +263,9 @@ def generate_avatar_prompt_with_llm(persona_data: dict, bedrock_client) -> str:
         return result['content'][0]['text'].strip()
     except Exception as e:
         logger.warning(
-            "[PERSONA_AVATAR] LLM prompt generation failed "
-            f"({type(e).__name__}: {e}), using fallback"
+            "[PERSONA_AVATAR] Avatar prompt generation failed in model "
+            f"resolution or Bedrock invocation ({type(e).__name__}: {e}), "
+            "using static fallback prompt"
         )
         return format_prompt(fallback_template, occupation=occupation or 'professional')
 
--- a/voc-datalake/lambda/shared/avatar.py
+++ b/voc-datalake/lambda/shared/avatar.py
@@ -12,8 +12,6 @@ import boto3
 
 from shared.logging import logger, tracer
 from shared.prompts import get_avatar_prompt_config, format_prompt
-# Lightweight runtime picker; it lazy-loads DynamoDB only when called.
-from shared.model_config import get_active_model_id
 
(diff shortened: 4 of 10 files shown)


perrozzi:

Replaces the `BEDROCK_MODEL_ID` fallback constant with `get_active_model_id(...)` at two live call sites (avatar prompt generation on the `utility` surface, persona `llm_metadata.model` on the `documents` surface) and drops a dead `MODEL_ID` alias in the research handler. The direction is right, but the avatar change moves a remote config lookup outside the try/except that guarantees the fallback prompt, the new test class in `test_projects.py` silently re-parents existing tests, and I don't see any IAM/CDK follow-up for the now-dynamic model id or the config lookup.

### Blockers

- `voc-datalake/lambda/shared/avatar.py`: `model_id = get_active_model_id(surface='utility')` is executed **before** the `try:` block. `generate_avatar_prompt_with_llm` is explicitly designed to degrade to a static fallback prompt on any failure (see `test_fallback_on_llm_error`); the model resolver hits the config store/precedence chain and can throw (missing item, throttling, permissions), which will now propagate out of the function and fail the whole avatar/persona flow instead of falling back. Move the resolution inside the `try`, or wrap it and default to the fallback model id.
- `voc-datalake/lambda/api/test/test_projects.py`: the new `TestGeneratePersonasModelMetadata` class is inserted in the middle of the existing avatar-prompt test class. Everything after the insertion point — starting with `test_returns_fallback_on_error` — is now a method of the new class, losing any setup/fixtures and semantics of the original class. Append the new class at the end of the file (or before the next class boundary) instead.
- IAM not updated in lockstep with the model id becoming dynamic. Per the repo rule, model id strings and their `bedrock:InvokeModel` resource ARNs must change together: the avatar-invoking Lambdas' policies presumably still scope to the ARN derived from `BEDROCK_MODEL_ID`, so a resolved override will fail with `AccessDenied`. Likewise every Lambda that now reaches `get_active_model_id` (the `projects` API handler plus any handler importing `shared/avatar.py`) needs read access to the model-config store. No CDK changes appear in this diff — please confirm the grants exist or add them (keeping the policy size limit in mind).

### Should fix

- `voc-datalake/lambda/api/projects.py`: `llm_metadata.model` is now resolved by a *second, independent* lookup after `converse_chain` has already run. If `converse_chain` resolves its own model (different surface, per-step override, or a config change between the two calls), the recorded metadata is wrong — arguably worse than the old constant because it looks authoritative. Prefer having `converse_chain` return/expose the model id it actually invoked and stamp that.
- `get_active_model_id` is now on the hot path for every avatar prompt and every persona save. If it performs a DynamoDB/SSM read per call, add a module-level cache (consistent with the module-level boto3 client pattern) so a 20-persona generation doesn't do 20 config reads.
- Add a regression test for the resolver-failure path in `shared/avatar.py` (resolver raises → fallback prompt returned). That test is what would have caught the try-block placement issue above and it fails if the fix is reverted.
- `voc-datalake/lambda/research/research_step_handler.py`: confirm nothing (other handlers, tests, or `research/*` modules) imports `research_step_handler.MODEL_ID`; the diff alone doesn't prove it was unused.

### Nits

- `voc-datalake/lambda/shared/avatar.py`: the function-local `from shared.model_config import get_active_model_id` keeps the old lazy-import shape; a module-level import matches the rest of the shared modules and avoids per-call import overhead (assuming no circular-import reason).
- `voc-datalake/lambda/api/test/test_projects.py`: `from unittest.mock import patch` is imported inside the test body although the module already imports `MagicMock`/`patch` at top level.
- `mock_model_id.assert_called_once_with(surface='documents')` will break if persona generation ever stamps metadata per-persona; asserting `called_with` would be less brittle.
- `voc-datalake/lambda/research/research_step_handler.py`: the removal leaves a bare blank line where `MODEL_ID` was; make sure it lands as the standard two blank lines before `def _get_feedback_table`.

---

perrozzi:

This PR replaces the `BEDROCK_MODEL_ID` fallback constant with `get_active_model_id(...)` at two live call sites (avatar prompt generation on the `utility` surface, persona `llm_metadata.model` on the `documents` surface) and drops the dead `MODEL_ID` alias in the research handler. The direction is right, but the avatar change moves model resolution outside the error-handling boundary, the new test in `test_projects.py` is inserted inside an existing test class, and the dynamic model id raises IAM/body-format questions that the diff does not address.

### Blockers

- `voc-datalake/lambda/api/test/test_projects.py`: the new `class TestGeneratePersonasModelMetadata` is inserted in the *middle* of the preceding class. The following methods (`test_returns_fallback_on_error`, and anything after it at the same indentation) are silently re-parented into the new class. They lose whatever class-level fixtures/setup the original class provided, and the avatar-prompt tests now live under a "persona model metadata" class. Move the new class to the end of the file (or after the complete existing class body).
- `voc-datalake/lambda/shared/avatar.py`: `model_id = get_active_model_id(surface='utility')` is executed *before* the `try` block that produces the fallback prompt. Previously `BEDROCK_MODEL_ID` was a constant that could not fail; now a config-lookup failure (DynamoDB/SSM error, missing surface entry) propagates out of `generate_avatar_prompt_with_llm` instead of degrading to the fallback prompt, breaking the documented "always return a prompt" contract exercised by `test_fallback_on_llm_error`. Move the resolution inside the `try`, or wrap it and fall back to the constant.
- Bedrock IAM must move in lockstep with model ids (repo rule). Both call sites now invoke a *dynamically resolved* model, but no CDK change is included. Please confirm the avatar/personas Lambda roles grant `bedrock:InvokeModel` on **every** model that the `utility` and `documents` surfaces can resolve to (including cross-region inference-profile ARNs), otherwise this turns a working call into a runtime `AccessDeniedException` as soon as an operator switches the active model. Add a test or CDK assertion if the grant is derived from the same config source.

### Should fix

- `voc-datalake/lambda/shared/avatar.py` still calls `bedrock_client.invoke_model` with an Anthropic-native body (`anthropic_version`, `messages`, `max_tokens`). That was safe when the model id was a pinned Claude constant; with a resolved `utility` model it will 400 (`ValidationException`) if the active model is a non-Anthropic provider. Either switch this call site to the shared `converse` helper (which is provider-agnostic) or validate the resolved provider before building the body.
- `voc-datalake/lambda/api/projects.py`: `llm_metadata.model` is now resolved independently of the model that `converse_chain` actually used. If `converse_chain` resolves its own surface/model internally, the two can drift (different surface, or a config change between the call and the stamp), and the metadata becomes misleading provenance data. Prefer having `converse_chain` return/expose the model id it used and stamp that.
- If the model is resolved per persona inside the generation loop, hoist it above the loop so a config lookup isn't repeated N times (the new test only covers `persona_count: 1`, so a per-persona call would pass `assert_called_once_with` today but regress under real usage).
- Confirm nothing else references `research_step_handler.MODEL_ID` (other handlers, step-function payload builders, or tests) before removal — the diff asserts it's unused but includes no grep evidence or test touching that module.

### Nits

- `voc-datalake/lambda/research/research_step_handler.py`: the removal leaves a single blank line before `def _get_feedback_table()`; PEP 8 wants two at module level.
- `voc-datalake/lambda/api/test/test_projects.py`: `import projects` / `from unittest.mock import patch` inside the test body — move to module-level imports to match the rest of the file.
- The new persona test asserts on `put_item.call_args.kwargs['Item']`; if the handler ever switches to a positional/batch writer this silently breaks. A small helper that extracts the stored item would be more robust.
- Consider asserting in the avatar test that `get_active_model_id` is called at most once per invocation, to pin down that resolution isn't accidentally repeated inside the retry path.

---

perrozzi:

This PR replaces the static `BEDROCK_MODEL_ID` constant with `get_active_model_id(...)` at two live persona call sites (`shared/avatar.py` for the `utility` surface, `api/projects.py` for the `documents` surface metadata) and drops the dead `MODEL_ID` alias in `research/research_step_handler.py`. The intent and the added tests are good, but the avatar change widens an `except Exception` over the model resolution + config load, which can silently mask IAM/import failures, and the diff contains no CDK/IAM changes even though the effective model id at these call sites can now differ from the fallback constant.

### Blockers

- `voc-datalake/lambda/shared/avatar.py`: the model resolution and prompt-config load were moved *inside* the existing `try`, so `get_active_model_id` failures, `ImportError` on `shared.model_config`, config-load errors, and — most importantly — a Bedrock `AccessDeniedException` for a newly-resolved model id all collapse into the generic fallback path with a single `logger.warning` that claims "LLM prompt generation failed". Resolve the model (and load config) outside the `try`, or catch narrowly, so misconfiguration surfaces instead of silently degrading every avatar prompt.
- Per the repo rule that Bedrock model ids and their IAM resource ARNs change in lockstep: this diff changes which model is invoked (`utility` surface in `avatar.py`) with no accompanying CDK grant. Please confirm the Lambda role that runs `generate_avatar_prompt_with_llm` (and `generate_personas`) is granted `bedrock:InvokeModel` on *all* model ARNs the `utility`/`documents` surfaces can resolve to, including cross-region inference-profile ARNs if the precedence chain can select them. Combined with the swallowed exception above, a missing grant fails silently in production.

### Should fix

- `voc-datalake/lambda/api/projects.py:790`: `llm_metadata.model` is now resolved by a *second, independent* call to `get_active_model_id(surface='documents')`, not by the code path that actually made the call (`converse_chain`). If `converse_chain` resolves its own model (different surface, per-call override, or config changed mid-run), the recorded metadata will be wrong — the same class of bug this PR is fixing. Prefer having `converse_chain` report the model it used, or share a single resolved value with the call.
- Verify the `documents` surface is the one `converse_chain` uses for persona synthesis; the diff gives no evidence, and a mismatch just relabels the incorrect value.
- `voc-datalake/lambda/shared/avatar.py`: the fallback template is now duplicated (hardcoded default in Python + `fallback_prompt_template` in the config file), and when model resolution fails the config-provided fallback is never read — a behavior regression vs. the previous code. `test_fallback_on_model_resolver_error` passes only because the assertion (`'Designer' in result`) matches both templates, so it does not catch this.
- `voc-datalake/lambda/research/research_step_handler.py`: please confirm nothing else in `research/` (or handlers importing from this module) references `MODEL_ID`; the diff can't show that, and removal is silent at import time only if truly unused.
- Tests patch `shared.model_config.get_active_model_id`, which only works because the import is function-local inside `avatar.py`. That couples the tests to an implementation detail — if the import is hoisted to module level (recommended above), the patch will silently target the wrong object. Patch `shared.avatar.get_active_model_id` with a module-level import instead.

### Nits

- `research_step_handler.py`: the removal leaves one blank line before `def _get_feedback_table`; PEP 8 wants two blank lines before a top-level def.
- `avatar.py`: the warning text "LLM prompt generation failed" is now misleading for config/model-resolution failures; make it distinguish the cause (and consider a Powertools metric on the fallback path, per the repo's Metrics pattern).
- If `get_active_model_id` does a network/SSM/DynamoDB lookup, it is now invoked once per persona avatar; consider module-level caching per invocation.
- `test_projects.py`: `table.put_item.call_args.kwargs['Item']` will break if the call ever becomes positional; `call_args.kwargs.get('Item')` with an explicit assertion message is a bit friendlier.
- The new `test_stores_resolved_model_id` needs 6 patches to reach one assertion — a sign `generate_personas` is doing a lot; not blocking, but worth noting for future refactors.

---

slegarraga:
Thanks, addressed in bfec94f.

Changes:
- Split avatar prompt config, model resolution, and Bedrock call handling into separate failure boundaries. Config and resolver failures use the configured static fallback, while Bedrock access denied is logged and re-raised instead of being downgraded.
- Pinned persona generation to one resolved documents model per invocation by adding model_id to converse_chain and stamping the same value into llm_metadata.
- Added regressions for resolver fallback, config fallback, access denied propagation, metadata pinning, and converse_chain model_id forwarding.

Confirmations:
- IAM is already allowlist based. Persona generator, persona importer, projects API, document roles, research role, processing roles, settings, chat, scrapers, and manual import use allowlistedModelArns(...), which includes each picker model's global inference-profile ARN plus foundation-model ARN. No CDK diff was needed for this call site change.
- The utility picker can only resolve ALLOWED_MODELS from shared/model_config.py, currently all global.anthropic Claude IDs, so the existing Anthropic body shape remains valid.
- research_step_handler.MODEL_ID had no references in lambda/research.
- get_active_model_id is cached in shared/model_config.py, so this does not add per-persona config reads.

Verified:
- env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short, 140 passed
- env -u PYTHONPATH .venv/bin/python -m py_compile lambda/shared/avatar.py lambda/shared/converse.py lambda/api/projects.py lambda/shared/test/test_avatar.py lambda/shared/test/test_converse.py lambda/api/test/test_projects.py
- git diff --check
- diff-only no em dash and no long added lines

perrozzi:

This PR replaces the static `BEDROCK_MODEL_ID` fallback at two live call sites with the surface-based resolver (`get_active_model_id`), adds a `model_id` pin-through parameter to `converse_chain` so the invoked model and the stored `llm_metadata.model` cannot drift, and drops a dead alias in `research_step_handler.py`. The core intent and the "resolve once, pin it" approach are good, and the tests are reasonable. However, the avatar change smuggles in an unrelated error-handling behaviour change (AccessDenied now propagates), and both changed call sites now invoke admin-configurable models with no corresponding IAM/infra verification in the diff.

### Blockers

- `voc-datalake/lambda/shared/converse.py`: `model_id=model_id` is passed unconditionally to `converse()` for every step, but the docstring for `surface` states steps resolve from the surface "when they don't set their own model" — i.e. per-step model overrides exist. As written, a chain-level `model_id` silently clobbers any per-step override (or, depending on how the step model is currently forwarded, changes which of the two wins). Make the precedence explicit (`model_id=step.get('model', model_id)` or the reverse) and add a test covering a chain with both a chain-level `model_id` and a step that sets its own model. `test_pins_explicit_model_id` only asserts forwarding, not precedence.
- `voc-datalake/lambda/shared/avatar.py`: both changed call sites now invoke whatever model the `utility` / `documents` surfaces resolve to, but there is no CDK/IAM change in this PR. Per the repo rule that model ids and their IAM resource ARNs move in lockstep, please confirm the persona/avatar Lambda's `bedrock:InvokeModel` grant covers every admin-selectable model (and inference-profile/cross-region ARNs) for those surfaces. This matters more than usual here because of the next point.
- `voc-datalake/lambda/shared/avatar.py`: the new `_is_bedrock_access_denied` re-raise is an undocumented behaviour change unrelated to the stated fix. Previously any Bedrock failure degraded to a fallback prompt; now an IAM/permission error propagates out of `generate_avatar_prompt_with_llm`. If `generate_persona_avatar` / `projects.generate_personas` doesn't isolate it, a misconfigured (or newly ungranted, see above) model turns a cosmetic avatar-prompt degradation into a failed persona generation. Either drop this from the PR or add a caller-level test proving the persona flow still completes when avatar prompt generation raises.

### Should fix

- `voc-datalake/lambda/shared/avatar.py`: the request body is Anthropic-native (`anthropic_version`, `max_tokens`, `messages`) but the model id is now dynamic. If the `utility` surface is pointed at a non-Anthropic model, `invoke_model` fails and the code silently falls back to a static prompt with only a warning. Prefer routing this through `shared.converse.converse(surface='utility')` (model-agnostic) rather than hand-rolling the Anthropic payload.
- `voc-datalake/lambda/shared/avatar.py`: `_is_bedrock_access_denied` does `getattr(exc, 'response', {}).get('Error', {})` — if an exception carries a non-dict `response` attribute this raises `AttributeError` from inside the `except` block and loses the original error. Guard with `isinstance(..., dict)`, and prefer matching `botocore.exceptions.ClientError` explicitly over the `'AccessDenied' in type(exc).__name__` heuristic.
- `voc-datalake/lambda/shared/avatar.py`: `from shared.model_config import get_active_model_id` is done inside the function and inside the `try`, so a genuine packaging/import bug is logged as "Active model resolution failed" and silently downgraded. Move the import to module level (as `api/projects.py` already does) unless there's a documented circular-import reason.
- Verify (with a test) that `converse()` actually gives `model_id` precedence over `surface`; the entire "cannot drift" guarantee in the new docstring and comments rests on that, and neither new test exercises it.

### Nits

- `voc-datalake/lambda/shared/test/test_avatar.py`: `test_raises_on_bedrock_access_denied` uses a hand-rolled `try/except/else` — use `pytest.raises(AccessDeniedError)`.
- `voc-datalake/lambda/research/research_step_handler.py`: removing `MODEL_ID` leaves a single blank line before `def _get_feedback_table()`; PEP 8 wants two. Also worth grepping for any external `research_step_handler.MODEL_ID` consumers.
- If `BEDROCK_MODEL_ID` in `shared/aws.py` now has no live call sites, consider marking it deprecated (or removing it) so it isn't copied into new handlers — that's the stated motivation for the `MODEL_ID` deletion.
- `voc-datalake/lambda/api/projects.py`: include the surface in the resolution log line (`[PERSONA] Resolved model for surface 'documents': ...`) to make config-drift debugging easier.
- `voc-datalake/lambda/shared/avatar.py`: `format_prompt(fallback_template, occupation=...)` will misbehave if an operator's `fallback_prompt_template` contains other placeholders; consider validating the config template's placeholders.

---

slegarraga:
Updated in b89d417.

Additional fixes:
- converse_chain now preserves explicit precedence: step model_id or model wins over the chain model_id, and the chain model_id still pins steps without a step override.
- Added a converse() test proving explicit model_id bypasses surface resolution.
- generate_avatar_prompt_with_llm now re-raises Bedrock AccessDenied, while generate_persona_avatar catches prompt-generation failures at the avatar boundary and returns no avatar so persona generation continues.
- AccessDenied detection now reads ClientError.response defensively and handles non-dict response attributes.
- Added an allowlist test documenting that every selectable text model is currently global.anthropic.*, so the Anthropic invoke_model body remains valid.
- Fixed the research_step_handler top-level def spacing nit.

Verified:
- env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short: 144 passed
- env -u PYTHONPATH .venv/bin/python -m py_compile lambda/shared/avatar.py lambda/shared/converse.py lambda/api/projects.py lambda/research/research_step_handler.py lambda/shared/test/test_avatar.py lambda/shared/test/test_converse.py lambda/api/test/test_projects.py
- git diff --check
- diff hygiene: no em dash and no added line over 88 chars

Ruff note: full changed-file Ruff still reports pre-existing E501/I001 debt in these files, so I kept the PR scoped and used diff-only hygiene checks for the new lines.

perrozzi:

This PR replaces the static `BEDROCK_MODEL_ID` fallback at two live persona call sites with resolution through `get_active_model_id()`, adds a chain-level `model_id` pin to `converse_chain()`, and drops a dead alias in `research_step_handler.py`. The core fix is correct and well tested (resolve-once-then-pin is the right shape so the invoked model and `llm_metadata.model` can't diverge). However, `shared/avatar.py` carries a substantial, out-of-scope error-handling rewrite that changes runtime behavior, and nothing in the diff confirms the IAM side of the model-id change.

### Blockers

- **Bedrock model IDs and IAM ARNs must move in lockstep — no CDK change is present.** `voc-datalake/lambda/shared/avatar.py` now invokes whatever the `utility` surface resolves to, and `voc-datalake/lambda/api/projects.py` invokes whatever `documents` resolves to. Both were previously pinned to the single fallback constant. Please confirm (and cite in the PR) that the API/persona Lambda role grants `bedrock:InvokeModel` for *every* entry in `ALLOWED_MODEL_IDS` across both surfaces. Because the ids are `global.anthropic.*` inference profiles, the grant needs both the inference-profile ARN and the underlying `foundation-model` ARNs in the participating regions — otherwise an admin repointing a surface turns persona/avatar generation into a runtime `AccessDeniedException`. The new `test_raises_on_bedrock_access_denied` test suggests this failure mode was already observed.

### Should fix

- **`generate_avatar_prompt_with_llm` raising on AccessDenied is a functional regression, and the only production caller immediately swallows it.** In `shared/avatar.py`, an IAM failure previously fell back to the static prompt and the image model was still invoked (image generation uses a *different* model, so avatars often still succeeded). Now the exception propagates and `generate_persona_avatar` catches everything and returns `{'avatar_url': None, 'avatar_prompt': None}` — so the net effect of "don't downgrade IAM failures" is *fewer* avatars, with the raise never escaping the module. Either drop the raise and keep the fallback path, or let it propagate to a caller that can act on it. As written it's log-severity churn dressed up as a behavior guarantee.
- **The avatar error-handling rewrite is unrelated to the stated fix.** Config-load-before-LLM, the `_is_bedrock_access_denied` helper, and the new catch-all in `generate_persona_avatar` are three independent behavior changes bundled into a "resolve the active model" PR. They're defensible on their own but should be a separate PR so they can be reviewed/reverted independently of the model-resolution fix.
- **The `llm_metadata.model` guarantee is weaker than the comment claims.** `converse_chain` now lets `step.get('model_id') or step.get('model')` beat the chain-level pin (`shared/converse.py`), so if any step in `get_persona_generation_steps()` ever carries a `model`/`model_id` key, `projects.py` will stamp `resolved_model_id` while a different model actually ran — exactly the drift the comment at `projects.py:558` says "cannot" happen. Either record the model per step (from the returned results) or assert/log when a step overrides the pin.
- **Pre-existing surface drift in the metadata stamp.** `converse_chain` still honors per-step `surface` overrides, and `llm_metadata` records a single model for the whole chain. Worth a follow-up issue at minimum.

### Nits

- `_is_bedrock_access_denied` wraps `from botocore.exceptions import ClientError` in `try/except ImportError`; botocore is always present in the Lambda runtime. The `is_client_error and 'AccessDenied' in str(exc)` branch is also redundant with the error-code check and is fragile string matching — the code check alone is sufficient.
- The builtin fallback template string is duplicated between `shared/avatar.py` and `test_avatar.py`; hoist it to a module constant (e.g. `_DEFAULT_FALLBACK_PROMPT`) and import it in the test so the two can't drift.
- `test_text_model_picker_remains_anthropic_only` asserts on `ALLOWED_MODEL_IDS` but lives in `TestGenerateAvatarPromptWithLlm`; it belongs in `shared/test/test_model_config.py`, next to the constant it constrains, ideally with a comment in `model_config.py` explaining that `avatar.py`'s raw `invoke_model` body is Anthropic-only.
- Avatar tests patch `shared.model_config.get_active_model_id`, which only works because the import is function-local. If the lazy import isn't required for circular-import reasons, prefer a module-level import and patch `shared.avatar.get_active_model_id` (matches how `projects.py` is tested).
- Confirm nothing imported `research_step_handler.MODEL_ID` (other handlers, tests, or step-function glue) before removing the module attribute.

---

slegarraga:
Updated in a992421.

Changes since b89d417:
- Reverted the avatar error-handling rewrite. The avatar prompt path now only changes the modelId to get_active_model_id(surface='utility') and preserves the existing static fallback behavior.
- Added a CDK regression test in api-stack.test.ts proving ProjectsLambdaRole, PersonaGeneratorRole, and PersonaImporterRole include every allowlisted model ARN from allowlistedModelArns(...). That covers both the global inference-profile ARNs and the underlying foundation-model ARNs.
- Moved the Anthropic-only picker guard to shared/test/test_model_config.py and documented the raw Anthropic invoke_model body dependency in model_config.py.
- Added a persona step model override guard and test so llm_metadata.model cannot drift if a future persona prompt step adds model or model_id.
- Kept converse_chain per-step model support, with tests for both chain pinning and step override precedence.

Verified:
- env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short: 170 passed
- cd frontend && npm install && npm run build: built frontend/dist successfully for CDK synthesis
- npm test -- lib/stacks/api-stack.test.ts: 56 passed
- env -u PYTHONPATH .venv/bin/python -m py_compile lambda/shared/avatar.py lambda/shared/converse.py lambda/api/projects.py lambda/research/research_step_handler.py lambda/shared/model_config.py lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py lambda/shared/test/test_converse.py lambda/api/test/test_projects.py
- git diff --check
- Diff hygiene: no em dash and no added line over 100 chars

perrozzi:

This PR replaces the `BEDROCK_MODEL_ID` fallback constant at two live persona call sites (`shared/avatar.py`, `api/projects.py`) with resolution through `get_active_model_id`, threads a chain-level `model_id` through `converse_chain` so the invoked model and the stamped `llm_metadata.model` cannot drift, drops the dead `MODEL_ID` alias in the research handler, and adds Python/TS tests. The core direction is right and well tested, but the new "reject step model overrides" guard contradicts the precedence semantics documented in the same diff, per-step `surface` overrides become silently inert, and there is no CDK/IAM change even though the resolved model set widens at these call sites.

### Blockers

- `voc-datalake/lib/stacks/api-stack.test.ts` imports `allowlistedModelArns` from `../utils/model-allowlist`, but that module is not in the diff. If it doesn't already exist on `main`, the whole vitest suite fails to compile. Please confirm (or include the util).
- Bedrock model ids and IAM resource ARNs must move in lockstep, and this diff contains no CDK grant change. `shared/avatar.py` previously invoked exactly one hardcoded model id; it can now invoke *any* allowlisted id for the `utility` surface. Every role for a Lambda that imports `shared.avatar` (not just `ProjectsLambdaRole`, `PersonaGeneratorRole`, `PersonaImporterRole` covered by the new test) needs `bedrock:InvokeModel` on the full allowlist, otherwise switching the utility surface produces runtime `AccessDeniedException` that silently degrades to the hardcoded fallback prompt. Please verify the grant list and extend the new test to cover all such roles.

### Should fix

- `voc-datalake/lambda/api/projects.py`: the new `ValueError` for steps that set `model_id`/`model` directly contradicts `converse_chain`'s documented behavior added in the same diff ("It wins over the chain-level `model_id` because step-specific configs are more precise"). Pick one contract. As written, a persona prompt config that legitimately pins a cheaper model for one step turns persona generation from "works" into a hard 500.
- Related: that `raise ValueError(...)` sits inside the broad `try` whose handler logs `"Failed to build chain steps"` and surfaces `ServiceError('Failed to generate personas')` (as the new test asserts). The list of offending step names never reaches the caller. Validate before the `try`, or log the override names at `error` level so the misconfiguration is diagnosable from CloudWatch.
- `voc-datalake/lambda/shared/converse.py`: `converse_chain` still forwards `surface=step.get('surface', surface)`, but when a chain-level `model_id` is pinned, `converse()` gives `model_id` precedence, so a per-step `surface` override is now silently ignored. Either reject per-step `surface` in persona steps too, or treat chain-level `model_id` as a default that a step's explicit `surface` can still displace.
- Every new test mocks `get_active_model_id` (`test_avatar.py`, `test_projects.py`), so nothing verifies that `'utility'` and `'documents'` are actually registered surfaces in `shared/model_config.py`. A typo'd surface name would pass CI and silently fall back at runtime. Add an assertion in `test_model_config.py` that both names are valid surfaces.
- `voc-datalake/lambda/research/research_step_handler.py`: confirm nothing else in the research module references `MODEL_ID` (including `from research_step_handler import MODEL_ID`) before removing it.

### Nits

- `shared/avatar.py`: `from shared.model_config import get_active_model_id` inside the `try` means an `ImportError` (e.g. future circular import) is swallowed into the generic fallback warning. Hoist the import above the `try` (or to module scope if there's no cycle).
- Moving `get_avatar_prompt_config()` inside the `try` changes behavior: a prompt-config load failure now silently degrades to the hardcoded template instead of propagating. Probably intended — worth a comment.
- `test_allowlist_is_anthropic_only_for_raw_invoke_model_callers` asserts the `global.anthropic.` prefix, which also locks the allowlist to global cross-region inference profiles. If a `us.anthropic.` id is ever added the test fails for a reason unrelated to the Anthropic-body constraint it documents; consider matching on `.anthropic.` instead.
- `test_projects.py` reads `table.put_item.call_args.kwargs['Item']`, which breaks if the call site ever switches to positional args; a small helper or `assert_called_once_with`-style check on the mock would be more robust.
- The `[PERSONA] Resolved persona model` log line is fine PII-wise, but consider `logger.info("...", extra={"model_id": ...})` to match the powertools structured-logging style used elsewhere.

---

slegarraga:
Updated in 1094665.

Addressed the latest automated review:
- Confirmed `voc-datalake/lib/utils/model-allowlist.ts` already exists on `upstream/development`, so the TS import is not missing.
- Verified the closed runtime import set for `shared/avatar.py`: ProjectsLambdaRole, PersonaGeneratorRole, and PersonaImporterRole. Those three roles already receive every ARN from `allowlistedModelArns(...)`, and the CDK regression test now documents why those are the complete set. The other job bundles stage `shared/`, but do not import `avatar.py`; they call `shared.converse` directly.
- Moved the persona prompt override validation outside the prompt-build try. It now rejects `model_id`, `model`, and `surface` step overrides, logs the offending step names at error level, and still avoids calling Bedrock.
- Documented and tested the `converse_chain` pinned-model behavior for per-step `surface`: explicit `model_id` wins, so an inert surface override is logged instead of silently disappearing.
- Added a model-config regression proving the live `documents` and `utility` surface names resolve as registered defaults without table config.
- Hoisted the avatar model resolver import to module scope.
- Relaxed the Anthropic allowlist guard to check provider shape with `.anthropic.` rather than pinning the `global.` prefix.
- Confirmed `research_step_handler.py` has no remaining `MODEL_ID` references.
- Made the persona metadata test tolerate either keyword or positional `put_item` calls.

Verified after the update:
- `env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short --no-cov -p no:cacheprovider`: 172 passed
- `npm test -- lib/stacks/api-stack.test.ts`: 56 passed
- `git diff --check`
- Diff hygiene: 0 added em dash lines and 0 added lines over 100 chars

perrozzi:

Since the last review the author moved the persona step-override guard outside the Bedrock `try` (and extended it to reject `surface`), hoisted the `get_active_model_id` import to module scope in `avatar.py`, added an inert-`surface` warning in `converse_chain`, relaxed the Anthropic-prefix assertion, added a `test_model_config.py` assertion that `documents`/`utility` are registered surfaces, and documented the closed role set in the CDK test. Most "should fix" items are addressed; the two blockers around the missing `model-allowlist` util and the actual CDK grant remain unverified.

### Resolved
- Guard moved outside the broad `try` and logged at `error` level with the offending step names (`api/projects.py`).
- Per-step `surface` under a pinned chain `model_id` is no longer silently eaten — warned in `converse.py` and rejected in persona steps, with a test.
- `get_active_model_id` import hoisted out of the `try` in `shared/avatar.py`; avatar tests now patch `shared.avatar.get_active_model_id` (correct patch target).
- `test_model_config.py` now asserts `documents` and `utility` are registered surfaces, closing the typo hole left by mocked callers.
- Prefix assertion relaxed to `.anthropic.`; `[PERSONA]` log line converted to structured `extra=`.

### Still open
- `voc-datalake/lib/stacks/api-stack.test.ts` imports `allowlistedModelArns` from `../utils/model-allowlist`, still not present in the diff — confirm it exists on `main` or the vitest suite won't compile.
- No CDK grant change: the new test asserts the three persona roles already hold `bedrock:InvokeModel` on the full allowlist. If they don't, this is a red test rather than a fix — please confirm the grants exist (model ids and IAM ARNs in lockstep).

### New findings
- `voc-datalake/lambda/api/test/test_projects.py`: `caplog` will likely capture nothing — the powertools `Logger` used by `shared.logging` defaults to `propagate=False`, so records never reach the root handler `caplog` installs. Verify locally; if so the test needs `caplog.set_level(..., logger=<name>)` plus enabling propagation, otherwise `assert override_errors` fails.
- `voc-datalake/lambda/api/projects.py`: the new comment claims "the offending step names must reach the job record verbatim", but the `ValueError` is still swallowed by the function's outer catch-all into a generic `ServiceError` (the test no longer asserts the message). Either propagate the detail into the job record or correct the comment.
- `voc-datalake/lambda/shared/converse.py`: the warning fires on `model_id is not None and step.get('surface')` even when the step sets its own `model_id`/`model`, so the message ("the pinned model wins") is wrong in that case; and it never fires when the chain `model_id` is `None` but the step sets both `model` and `surface`, where the surface is equally inert. Keying the check off `step_model_id is not None` would be accurate in both directions.
- Nit — `voc-datalake/lambda/shared/test/test_model_config.py`: `'.anthropic.' in model_id` would reject a plain `anthropic.claude-*` id (no cross-region prefix); a `startswith('anthropic.') or '.anthropic.' in ...` form matches the stated intent.
- Nit — `voc-datalake/lambda/api/test/test_projects.py`: the `call.kwargs['Item'] if ... else call.args[0]` branch is defensive but will `IndexError` rather than fail cleanly if neither form is used; a tiny helper with an explicit assertion reads better.

---

slegarraga:
Follow-up pushed in 7a95f41.

I addressed the valid findings from the latest follow-up review:
- Replaced the `caplog` assertion with a direct `projects.logger.error` mock assertion, so the test is not sensitive to logger propagation settings.
- Corrected the `projects.py` comment: the step names are logged before the outer ServiceError boundary, not propagated into the job record.
- Changed the `converse_chain` warning to key off any explicit step model or pinned chain model. It now also warns when a step sets both `model` and `surface` without a chain pin.
- Added a regression for the step-level `model` plus `surface` case.
- Relaxed the Anthropic assertion to allow both plain `anthropic.*` ids and prefixed `*.anthropic.*` ids.
- Replaced the defensive `put_item` branch with a helper that fails cleanly if neither kwargs nor positional args contain the item.

Confirmed again:
- `voc-datalake/lib/utils/model-allowlist.ts` exists on `upstream/development`.
- `npm test -- lib/stacks/api-stack.test.ts` compiles the import and passes: 56 passed.
- The full persona role grant test passes against the synthesized template. The grants already exist, so this was a regression test plus documentation rather than a CDK policy change.
- `env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short --no-cov -p no:cacheprovider`: 173 passed.
- `git diff --check` passes.
- Worktree diff hygiene: 0 added em dash lines and 0 added lines over 100 chars.

perrozzi:

This PR replaces two live uses of the `BEDROCK_MODEL_ID` fallback constant with surface-aware resolution (`utility` for avatar prompt generation, `documents` for persona synthesis), drops a dead `MODEL_ID` alias in the research handler, and — beyond what the description says — adds a chain-level `model_id` pin to `converse_chain`, a hard rejection of per-step model/surface overrides in persona generation, and a new CDK IAM assertion test. The core fix is right and well tested; the added scope introduces a new hard-failure path and an IAM assumption that isn't backed by any stack change in this diff.

### Blockers

- `voc-datalake/lib/stacks/api-stack.test.ts` imports `allowlistedModelArns` from `../utils/model-allowlist`, but that module is not in the diff. If it doesn't already exist the whole api-stack test suite fails to compile. Please confirm (or add it).
- IAM/model lockstep: `shared/avatar.py` now resolves `surface='utility'` at runtime, so `bedrock:InvokeModel` must be granted on **every** allowlisted model ARN for `ProjectsLambdaRole`, `PersonaGeneratorRole`, and `PersonaImporterRole` — otherwise repointing the utility surface via the picker yields `AccessDenied` at runtime and the avatar silently degrades to the fallback prompt (the new broad `except Exception` in `avatar.py` swallows it as a `warning`). The new test asserts these grants but no stack file changed in this diff; if the grants are currently scoped to the default model only, this test will fail and the stack needs updating in the same PR.
- New hard failure in `voc-datalake/lambda/api/projects.py`: the `overridden` check raises `ValueError` (→ `ServiceError`/500) whenever a persona step config carries `model_id`, `model`, or `surface`. `converse_chain` explicitly supports per-step `surface` (see the pre-existing `test_chain_step_surface_overrides_chain_surface`), so this makes persona prompt configs a special case that breaks a previously legal config. If persona prompts are editable at runtime (admin prompt overrides), a single stray key takes persona generation down entirely. Prefer the behaviour you already implemented one layer down: log the ignored override in `converse_chain` and continue, rather than failing the job.

### Should fix

- Scope/description mismatch: the `converse_chain(model_id=...)` public API addition, the step-level `model`/`model_id` precedence, and the persona override validation are not mentioned in the PR description. Either split them out or document them — reviewers of the linked issue (#273) won't expect a behaviour-changing chain API in a "resolve active model" fix.
- Passing both `surface='documents'` and `model_id=resolved_model_id` to `converse_chain` in `projects.py` is redundant by your own precedence rules (the pinned model always wins). Pass only the pin, or drop the pin and let the surface resolve, so there's one source of truth rather than two plus a guard.
- `voc-datalake/lambda/shared/avatar.py`: `get_active_model_id()` and `get_avatar_prompt_config()` were moved inside the `try`, so a model-config or prompt-config failure now degrades to the hardcoded fallback template with only a `logger.warning` instead of surfacing. Resolve the model (and load config) before the `try`, or emit a metric so silent degradation is observable.
- The `converse` regression test (`test_explicit_model_id_takes_precedence_over_surface`) asserts pre-existing behaviour, which is fine, but there's no test that reverting the `avatar.py`/`projects.py` change is caught by anything other than mock-kwarg assertions on `get_active_model_id`. The `modelId`/`llm_metadata['model']` assertions cover it — just confirm both new tests fail with the fix reverted.

### Nits

- `voc-datalake/lambda/shared/avatar.py` top-of-file comment explains that heavy imports are deliberately lazy; adding a module-level `shared.model_config` import slightly cuts against that note — worth a one-line justification (it's boto3-lazy) or keeping it local.
- Comment volume in `projects.py` (`# Persona chains are PINNED...`, 8 lines) and in the new tests exceeds the surrounding style; the rationale would read better in the `converse_chain` docstring only.
- `logger.error("[PERSONA] ...", extra={'steps': ', '.join(overridden)})` — powertools takes structured fields directly; a list (`extra={'steps': overridden}`) queries better in CloudWatch than a joined string.
- `test_allowlist_is_anthropic_only_for_raw_invoke_model_callers` duplicates intent with the new comment block in `model_config.py`; consider asserting via a named constant so the comment can't drift from the test.
- `research_step_handler.py`: verify nothing else does `from research_step_handler import MODEL_ID` before removing the alias (grep-only change, low risk).

---

slegarraga:
Simplified in 5f6a340.

Changes from the current-head review:
- Removed the persona hard-failure path. Persona generation no longer raises on a step `model`, `model_id`, or `surface`; the chain-level `model_id` pin wins and the job continues.
- Updated `converse_chain` so a chain `model_id` truly pins every step. Step-level `model` or `model_id` only applies when the chain is not pinned.
- Kept warnings for inert overrides: step `surface` plus any explicit model, and step `model` plus a pinned chain model.
- Removed the redundant `surface='documents'` argument from the persona `converse_chain` call. The single source of truth is the resolved model pin.
- Moved `get_avatar_prompt_config()` back outside the Bedrock fallback boundary, preserving the old behavior: prompt file problems remain configuration errors, while Bedrock invocation failures degrade to the static fallback prompt.
- Added the requested import note for `shared.model_config`: the import is lightweight and DynamoDB is lazy until the picker is called.

Confirmations for the two repeated blockers:
- `voc-datalake/lib/utils/model-allowlist.ts` exists on `upstream/development`. The current `npm test -- lib/stacks/api-stack.test.ts` run compiled that import and passed.
- The persona roles already have the full allowlist grants, so there was no stack policy change to make. The new CDK regression proves `ProjectsLambdaRole`, `PersonaGeneratorRole`, and `PersonaImporterRole` each contain every ARN from `allowlistedModelArns('us-east-1', '111111111111')`.

Verified after the simplification:
- `env -u PYTHONPATH .venv/bin/python -m pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py lambda/api/test/test_projects.py lambda/shared/test/test_converse.py lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-header --tb=short --no-cov -p no:cacheprovider`: 173 passed.
- `npm test -- lib/stacks/api-stack.test.ts`: 56 passed.
- `git diff --check`: passed.
- Worktree diff hygiene: 0 added em dash lines, 0 added lines over 100 chars.

perrozzi:

This PR replaces the `BEDROCK_MODEL_ID` fallback constant with runtime surface resolution at two persona call sites (`shared/avatar.py` → `utility`, `api/projects.py` → `documents`), removes a dead alias in `research_step_handler.py`, and — beyond what the description says — adds a new chain-level `model_id` pinning feature to `shared/converse.converse_chain`. The core fix is right and well tested, but the chain-pin design quietly changes per-step model/surface semantics, and part of the added test surface (CDK/vitest) appears unverified and may reference a file not in this diff.

### Blockers

- `voc-datalake/lambda/api/projects.py:665` — the call changed from `converse_chain(..., surface='documents')` to `converse_chain(..., model_id=resolved_model_id)`, which (a) drops the chain surface entirely (steps now fall through to `DEFAULT_SURFACE` for anything in `converse()` that still keys off `surface`, e.g. logging/metrics dimensions and any surface-derived config) and (b) makes every per-step `model`/`model_id`/`surface` override in `get_persona_generation_steps()` inert, downgraded to a `logger.warning`. Please confirm no persona step sets a model or surface today (e.g. a cheaper model for a classification/summarization step) — if any does, this silently reroutes it. The minimal, non-behavior-changing fix for issue #273 is to keep `surface='documents'` on the chain and stamp `get_active_model_id(surface='documents')` into `llm_metadata`; if you want the pin, pass both `surface='documents'` and `model_id=` and treat step-level model overrides as winning over the chain default rather than losing to it.
- `voc-datalake/lib/stacks/api-stack.test.ts:39` imports `allowlistedModelArns` from `../utils/model-allowlist`, which is not part of this diff. If that helper is new it is missing from the PR (test won't compile); if it already exists, note that the verification block in the description only ran `pytest`, so the new CDK/IAM assertions and the `describe('persona Bedrock IAM grants')` block are unverified. Please run vitest on `lib/stacks/api-stack.test.ts` and include the result.

### Should fix

- `voc-datalake/lambda/shared/avatar.py:220` — `get_active_model_id(surface='utility')` is called *outside* the `try`, so a DynamoDB/config failure in the picker now raises out of `generate_avatar_prompt_with_llm` where the previous constant import could not fail. Since this function's contract is "degrade to a deterministic static prompt", move the resolution inside the `try` (or resolve lazily at the `invoke_model` call) so picker failures also hit the fallback path.
- Per the repo's model-id/IAM lockstep rule, `avatar.py` can now invoke *any* allowlisted model at runtime (utility surface is admin-repointable), but no CDK grant changed in this diff. Confirm `ProjectsLambdaRole`, `PersonaGeneratorRole` and `PersonaImporterRole` already grant `bedrock:InvokeModel` on the full allowlist ARN set — otherwise this is an AccessDenied waiting for the first surface repoint, and the new test is the only thing catching it.
- `voc-datalake/lambda/research/research_step_handler.py:50` — please confirm `MODEL_ID` has no remaining references in that module (or in anything importing `research_step_handler.MODEL_ID`); the diff removes the definition without showing the rest of the file.

### Nits

- `voc-datalake/lambda/shared/avatar.py:15` — the module deliberately keeps heavy imports lazy (see the `cloudfront_signing` comment right below); a module-level `shared.model_config` import pulls boto3/DynamoDB config code into every avatar-adjacent cold start. A function-local import would preserve the stated pattern.
- `voc-datalake/lambda/shared/converse.py:620-637` — a step with both `model` and `surface` under a pinned chain emits two warnings for one condition; consider collapsing into a single message.
- `voc-datalake/lambda/shared/model_config.py:121` — the Anthropic-only comment sits above `_OMIT_TEMPERATURE_IDS` rather than next to `ALLOWED_MODELS`, where an author adding a model would actually read it.
- `voc-datalake/lambda/api/test/test_projects.py:307` — `_put_item_payload` reads the *last* `put_item` call; with `persona_count > 1` this test would silently only assert the final persona. Consider asserting across `call_args_list`.
- New public parameter `converse_chain(model_id=...)` isn't mentioned in the PR description; worth calling out since it changes shared-module semantics for all future callers.

---

slegarraga:
Addressed in abf7bdd.

**Blocker 1 (projects.py chain call):** Restored `surface='documents'` alongside the `model_id` pin. The pin still decides which model is invoked; the surface argument keeps its pre-existing role so surface-derived logging keeps the right dimension. On override inertness: confirmed via grep that no persona step sets a model or surface today (the only step-level `model` reference is the post-chain `llm_metadata` stamp), so nothing is silently rerouted. If a future step adds one, it now gets exactly one warning naming every dropped key rather than failing the job or being eaten quietly.

**Blocker 2 (api-stack.test.ts import):** `lib/utils/model-allowlist.ts` already exists on `development`; it was never new in this diff. The vitest run was real but only lived in my PR comment, not the description: fixed by adding both verification blocks to the description. Current runs: pytest 174 passed, `npm test -- lib/stacks/api-stack.test.ts` 56 passed including the `persona Bedrock IAM grants` assertions.

**Should fix (avatar.py resolution placement):** Moved `get_active_model_id(surface='utility')` inside the fallback boundary, superseding my previous placement. Reasoning: before this PR the value was a constant that could never fail, and this function's contract is to degrade to a deterministic static prompt on model problems; a transient DynamoDB hiccup in the picker is exactly such a problem, so it now degrades like an invocation failure instead of opening a new hard-failure path. Prompt-config loading stays above the boundary because prompt-file problems were configuration errors before the PR too. The comment in the diff records the rationale so the next reader does not "fix" it back.

**Should fix (IAM lockstep):** Confirmed again, no stack change needed: all three roles (`ProjectsLambdaRole`, `PersonaGeneratorRole`, `PersonaImporterRole`) already carry `bedrock:InvokeModel` on the full allowlist ARN set. The CDK regression test pins each role against `allowlistedModelArns('us-east-1', '111111111111')`.

**Should fix (MODEL_ID references):** Verified zero remaining references to `research_step_handler.MODEL_ID` anywhere under `lambda/`; the surviving mentions are prose comments pointing at the file path, not imports.

Nits also taken:
- The two inert-override warnings are collapsed into one combined line naming everything dropped, with a new test asserting exactly one warning when a step carries both `model` and `surface` under a pin.
- The Anthropic-only comment moved next to `ALLOWED_MODELS`, where an author adding a row will actually read it.
- Persona tests now assert across `put_item.call_args_list`, so a `persona_count > 1` run can no longer hide earlier rows behind the final call.
- Kept the module-level `shared.model_config` import: `avatar.py` already imports `boto3` unconditionally at module scope, and `model_config` defers its DynamoDB work until the picker runs, so there is no added cold-start weight. The existing note above the import says as much.

Verification after this round:

```bash
pytest lambda/shared/test/test_avatar.py lambda/shared/test/test_model_config.py \
  lambda/api/test/test_projects.py lambda/shared/test/test_converse.py \
  lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-cov
```

174 passed, 0 failed.

```bash
npm test -- lib/stacks/api-stack.test.ts
```

56 passed, 0 failed. `git diff --check`: clean.

perrozzi:

Since the last review the author restored `surface='documents'` alongside the chain pin in `api/projects.py`, moved `get_active_model_id(surface='utility')` inside the fallback `try` in `shared/avatar.py`, collapsed the two inert-override warnings into one combined message in `shared/converse.py`, and made the persona test helper assert across all `put_item` calls. The PR description now also confirms no persona step sets a model/surface today, that `lib/utils/model-allowlist.ts` pre-exists on `development`, that vitest ran green (56 passed), and that the three persona roles already grant the full allowlist — which covers the outstanding verification asks.

### Resolved
- `api/projects.py:668` — chain surface restored, so surface-derived logging/metrics dimensions are no longer downgraded to `DEFAULT_SURFACE`; step-model precedence confirmed inert-only (no persona step sets one).
- `lib/stacks/api-stack.test.ts` — import confirmed pre-existing and vitest results now included in the description.
- `shared/avatar.py:239` — picker resolution moved inside the fallback boundary, so a DynamoDB blip degrades to the static prompt instead of raising.
- IAM lockstep — description confirms `ProjectsLambdaRole`/`PersonaGeneratorRole`/`PersonaImporterRole` already grant `bedrock:InvokeModel` on the full allowlist, pinned by the new CDK test.
- `shared/converse.py:620-637` — single combined warning; `test_projects.py` now asserts across `call_args_list`.

### Still open
- `voc-datalake/lambda/shared/avatar.py:15` — module-level `shared.model_config` import still breaks the file's stated lazy-import pattern (nit).

### New findings
- `voc-datalake/lambda/api/test/test_projects.py:294` — `_put_item_payloads` now asserts `llm_metadata` on *every* `put_item`; if `generate_personas` ever writes a non-persona row (project status/summary) the test fails with a bare `KeyError` instead of a meaningful assertion. Filter to payloads carrying `llm_metadata` (or an entity-type discriminator) and assert the expected count.
- `voc-datalake/lambda/shared/avatar.py:262` — a picker failure now reaches the `except` and logs `"LLM prompt generation failed"`, which mis-attributes a DynamoDB/config problem to Bedrock; the type/message is included so it's diagnosable, but a slightly broader wording would save an operator a wrong-service hunt (nit).

---

perrozzi:

Only two small follow-ups landed since the last review: the persona test helper now filters `put_item` payloads on `llm_metadata` (with a non-empty assertion), and the avatar fallback warning was reworded to cover model resolution as well as Bedrock invocation. Both of my previous new findings are addressed; the remaining item is the earlier import nit.

### Resolved
- `voc-datalake/lambda/api/test/test_projects.py:294` — filtered to persona rows, so a future non-persona write fails with a meaningful "no persona rows" assertion instead of `KeyError`.
- `voc-datalake/lambda/shared/avatar.py:262` — warning now names model resolution *or* Bedrock invocation, so a DynamoDB blip isn't mis-attributed to Bedrock.

### Still open
- `voc-datalake/lambda/shared/avatar.py:15` — module-level `shared.model_config` import departs from the file's stated lazy-import pattern (nit).

### New findings
- `voc-datalake/lambda/api/test/test_projects.py:305` (nit) — the same filter that removes the `KeyError` also hides the regression it guards: a code change that stops stamping `llm_metadata` on a persona row drops that row from `persona_rows` rather than failing. Since both tests run with `persona_count: 1`, adding `assert len(persona_rows) == 1` (or comparing against the count of payloads carrying a persona discriminator) would close the gap.

---

slegarraga:
Follow-ups from the abf7bdd review, pushed as 0f7aae0.

- `_put_item_payloads` now filters on `llm_metadata` and asserts at least one persona row was written, so a hypothetical future non-persona `put_item` fails as "no persona rows" instead of a bare KeyError.
- The avatar fallback warning no longer implies Bedrock: it now reads "Avatar prompt generation failed in model resolution or Bedrock invocation (...)", since a picker failure lands in the same handler by design.
- On the module-level `shared.model_config` import nit: keeping it deliberately. `avatar.py` already imports `boto3` unconditionally at module scope (image-model path), and `model_config` defers all DynamoDB work until the picker is called, so there is no added cold-start weight. The note above that import records this so the pattern stays intentional rather than accidental.

Verification after 0f7aae0: pytest suites (avatar, model_config, projects, converse, persona_chain) 174 passed; `git diff --check` clean.

perrozzi:

Only the persona test helper changed since my last pass: `_put_item_payloads` now takes an `expected_count` and asserts an exact row count instead of a non-empty check, which closes my previous nit.

### Resolved
- `voc-datalake/lambda/api/test/test_projects.py:305` — exact-count assertion means a persona row that loses its `llm_metadata` stamp now fails loudly instead of being filtered out.

### Still open
- `voc-datalake/lambda/shared/avatar.py:15` — module-level `shared.model_config` import vs. the file's stated lazy-import pattern (nit).

### New findings
- `voc-datalake/lambda/api/test/test_projects.py:296` (nit) — the docstring still says a non-persona row "fails as 'no persona rows'", but the assertion message is now `expected N persona rows, found M`; worth aligning the wording.

Otherwise the delta is clean.

---

slegarraga:
Good catch, fixed in 50e5155.

`_put_item_payloads` now takes an `expected_count` and asserts `len(persona_rows) == expected_count`, so the filter can no longer hide the regression it guards against. Both tests pass `1` (their runs use `persona_count: 1`) and the docstring records why the pin exists.

Mutation-checked before pushing: a run whose only `put_item` row lacks `llm_metadata` now fails with `expected 1 persona rows, found 0` instead of passing silently, while the correctly stamped single row still passes.

Verification after 50e5155:

```bash
pytest lambda/api/test/test_projects.py lambda/shared/test/test_avatar.py \
  lambda/shared/test/test_model_config.py lambda/shared/test/test_converse.py \
  lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-cov
```

174 passed, 0 failed. `git diff --check`: clean.

slegarraga:
Fixed the follow-up nits in d72ad10.

Changes:
- moved the `shared.model_config.get_active_model_id` import inside `generate_avatar_prompt_with_llm()`'s fallback boundary, matching the narrow/lazy import pattern in this module;
- updated the avatar tests to patch `shared.model_config.get_active_model_id`, since the function now imports it lazily at call time;
- aligned the `_put_item_payloads` docstring with the new exact-count assertion wording.

Verification:

```bash
pytest lambda/api/test/test_projects.py lambda/shared/test/test_avatar.py \
  lambda/shared/test/test_model_config.py lambda/shared/test/test_converse.py \
  lambda/jobs/persona_generator/test/test_persona_chain.py -q --no-cov
```

174 passed, 0 failed. `git diff --check`: clean.

perrozzi:

Only three things changed since my last pass: the persona-test docstring wording, the `shared.model_config` import moved from module level into `generate_avatar_prompt_with_llm`, and the avatar tests re-pointed to patch `shared.model_config.get_active_model_id`. Both previous findings are addressed.

### Resolved
- `voc-datalake/lambda/shared/avatar.py:238` — lazy import inside the fallback boundary now matches the module's stated narrow import graph (and stays inside `try`, so picker failures still degrade rather than raise).
- `voc-datalake/lambda/api/test/test_projects.py:296` — docstring wording now matches the `expected N persona rows, found M` assertion.

### New findings
- `voc-datalake/lambda/shared/test/test_avatar.py:13` (nit) — patching the source module rather than the importing module is correct for a function-local import, and `test_successful_prompt_generation` asserting `modelId == 'test-model'` still guards against a bad import path silently falling through to the static fallback. No action needed.

Delta is clean.

---

Issue #271, opened 2026-08-06
[Infra]: npm run check cannot pass — lint:python is red on development


perrozzi:
## Summary

`npm run check` is documented as the pre-deploy quality gate, but it currently exits 1 for everyone on `development`. Since this repo has no CI workflow, that command *is* the gate — so right now nothing verifies a change before it merges.

## Current state

`npm run check` chains six scripts with `&&`:

```
lint && typecheck:all && test && test:cdk && test:stream && test:backend
```

It fails at the first step. `npm run lint` runs `lint:frontend && lint:stream && lint:python`, and `lint:python` (`ruff check lambda plugins`) reports **551 errors** on an otherwise clean checkout of `development`.

The findings are in long-standing code, not a recent regression — import ordering in `_shared.schemas` re-imports, `E402` in `app_store_web_scraper` shims, `f"{e!s}"` conversions, and similar. They look like the result of ruff rules being adopted (or tightened) without a cleanup pass over the existing tree.

Two consequences worth separating:

1. **No green signal.** Contributors and reviewers have no single command whose success means anything.
2. **`&&` hides the rest.** Because the chain short-circuits, a `lint:python` failure means `typecheck:all`, the three test suites, and the Python tests never run. Someone triaging a red gate can't tell from `check` alone whether the CDK or backend suites pass — they have to know to run each script individually. (For the record, they do all pass on `development` today: `typecheck:all` clean, `test:cdk` 138, `test:stream` 257, `test:backend` 1281.)

## Proposed solution

Pick one of:

- **Fix the findings.** Most look auto-fixable (`ruff check --fix`); the residue can be triaged. Largest change, best end state.
- **Baseline them.** Narrow the enabled rule set, or add targeted `per-file-ignores` for the known offenders, so new code is linted while the existing tree is grandfathered. Then tighten incrementally.
- **At minimum, stop the chain from hiding results.** Run the independent legs so one failure doesn't mask five passes — for example by not short-circuiting between them, and reporting each outcome.

Whichever route is taken, it's worth adding a CI workflow that runs the gate on pull requests. Without one, the gate depends on each contributor remembering to run it locally, which is also how it drifted this far.

## Acceptance criteria

- [ ] `npm run check` exits 0 on a clean checkout of `development`.
- [ ] A `lint:python` failure no longer prevents the other five legs from reporting.
- [ ] The gate runs automatically on pull requests.
[...]

Issue #255, opened 2026-08-05
[Bug]: Home links to the deleted /feedback route and renders a raw i18n key


perrozzi:
## Summary

The Home page still links to `/feedback`, which was removed when the feedback list was consolidated into
Categories. The link label also resolves to a missing translation key, so the raw key is displayed.

## Steps to reproduce

1. Open the Home page.
2. Look at the quick-links list.

## Expected

A working link, with a translated label.

## Actual

`frontend/src/pages/Home/Home.tsx:72`:

```ts
{ to: '/feedback', labelKey: 'common:nav.feedback', icon: MessageSquare },
```

Two problems from one change:

- `/feedback` (the list route) was removed in the Categories consolidation. The route no longer resolves.
- `nav.feedback` was deleted from the locale files in the same change, so `labelKey` has no translation
  and the raw key string renders.

The consolidation cleaned up the sidebar (`Layout.tsx`) and the locale files but missed this Home entry.

## Proposed fix

Point the entry at `/categories` and use the existing Categories nav label. Confirm no other component
references `nav.feedback` or the bare `/feedback` list route:

```bash
grep -rn "nav.feedback'" frontend/src/
grep -rn "'/feedback'" frontend/src/
```

Note `/feedback/:id` (detail) is still valid and must keep working — only the list route went away.

## Acceptance criteria

- [ ] Home's link navigates to Categories.
- [ ] The label is translated in all 8 locales.
- [ ] No remaining reference to the removed list route or the deleted key.

## Area

Frontend — `frontend/src/pages/Home/Home.tsx`

perrozzi:
> 🤖 **Automated triage.** The labels and duplicate candidates below are suggestions, not decisions.

Thanks for the detailed report — and for pinpointing the exact line!

My understanding: after the Categories consolidation, `frontend/src/pages/Home/Home.tsx` still declares a quick-link to `/feedback` with `labelKey: 'common:nav.feedback'`. Since both the list route and that locale key were removed, the link no longer resolves and the untranslated key string is shown. The fix is to retarget the entry at `/categories` with the existing Categories nav label, while leaving the `/feedback/:id` detail route intact.

A couple of quick questions:

- Are there any other surfaces (e.g. dashboard cards, empty states, breadcrumbs) where you've seen the same stale route or key, or is Home the only remaining occurrence based on your `grep` output?
[...]

Issue #246, opened 2026-08-05
[Security]: No WAF on public endpoints, and the docs claim there is one


perrozzi:
## Summary

The architecture docs list WAF as a deployed protection with rate limiting and SQL-injection/XSS rules.
There is no WAF in any stack. Public, cost-amplifying endpoints are directly exposed.

## Severity

High — mostly because the documentation actively misleads. A reader doing threat modelling would conclude
rate limiting is handled.

## Risk

`grep -rl 'wafv2\|WebAcl' lib/` returns nothing. No Web ACL, no rate-based rule, no managed rule group.

Meanwhile the docs describe WAF as part of the platform, including in the service table and the security
section.

Unauthenticated endpoints that cost money or write data:

- `POST /feedback-form/submit` — public by design; writes to the queue and drives the enrichment pipeline
  (Bedrock and Comprehend per item)
- `GET /feedback-form/{id}` and the embed endpoint — public reads
- `POST /webhooks/{plugin}` — public receiver

An unauthenticated caller can drive per-item LLM spend by submitting in a loop.

## Proposed mitigation

Pick one and make the docs match:

**Option A — deploy a WAF.** A Web ACL on the API stage with a rate-based rule and the managed common
rule set. Straightforward; adds cost.

**Option B — don't, and say so.** Use API Gateway usage plans and per-method throttling for the public
routes, plus payload caps, and correct the docs. Cheaper, covers the amplification risk, but no managed
rule coverage.

Either way: **fix the documentation in the same PR.** The current state — docs describing a control that
does not exist — is the worst of both.

## Acceptance criteria

- [ ] A decision is recorded in the repo.
- [ ] Public endpoints have throttling that a load test can demonstrate.
- [ ] Docs describe what is actually deployed. No reference to WAF unless a WAF exists.

## Affected files

`lib/stacks/api-stack.ts`, `docs/` (service table + security sections), `.kiro/steering/tech.md`

Related: #222 (harden the public feedback-form submit path) covers the form-specific half.

perrozzi:
> 🤖 **Automated triage.** The labels and duplicate candidates below are suggestions, not decisions.

Thanks for the detailed write-up — and especially for separating the two problems here.

As I understand it there are two distinct issues:

[...]

Issue #242, opened 2026-08-05
[Security]: Admin-only operations are not enforced server-side


perrozzi:
## Summary

Only two handlers enforce admin group membership. Every other privileged operation — brand and category
configuration, data-explorer writes and deletes, scraper management, integration credentials, feedback-form
management — is reachable by any authenticated user, including a viewer.

## Severity

High. Not an exploit chain: the endpoints are Cognito-authenticated, so a caller must be a valid user.
But any valid user is effectively an admin for most of the product.

## Risk

`require_admin` (`lambda/shared/api.py`) is called in exactly two places:

- `lambda/api/users_handler.py` — 8 call sites, correctly gating all of `/users/*`
- `lambda/api/settings_handler.py` — **one** call site, line 137, gating only `PUT /settings/model`

Everything else in `settings_handler.py` is ungated, including `PUT /settings/brand`,
`PUT /settings/categories`, and `POST /settings/categories/generate` — the last of which spends Bedrock
tokens on demand.

Also ungated: data-explorer `PUT`/`DELETE` (edit and delete arbitrary feedback records), scraper create
and delete, `/integrations/*` credential writes, and feedback-form management.

The frontend's `<AdminRoute>` wrapper hides these screens from non-admins, but it is client-side only —
it changes what the UI renders, not what the API accepts. Anyone who can reach the API can call them.

## Proposed mitigation

1. Decide the privilege model per route group and write it down: which operations are admin-only, which
   are available to all authenticated users.
2. Apply `require_admin` to every route that lands in the admin-only set.
3. Add a test per privileged route asserting a non-admin caller is rejected — the gap here is that
   nothing fails when a gate is missing, so only tests will keep it closed.
4. Treat `<AdminRoute>` as UX affordance only, and note that in the component.

## Acceptance criteria

- [ ] The privilege model is documented in the repo.
- [ ] Every admin-only route rejects a caller who is not in the admins group.
- [ ] A test per privileged route covers the non-admin rejection path.
- [ ] `POST /settings/categories/generate` (Bedrock spend) is gated.

## Affected files

`lambda/api/settings_handler.py`, `lambda/api/data_explorer_handler.py`,
`lambda/api/scrapers_handler.py`, `lambda/api/integrations_handler.py`,
`lambda/api/feedback_form_handler.py`, `lambda/shared/api.py`

perrozzi:
[...]

PR #277, opened 2026-08-07
fix(api): require Cognito on feedback-form item routes

perrozzi:
## Summary

Every route under `/feedback-forms/{proxy+}` was served with `AuthorizationType: NONE`, so form
update, form delete, and reads of submitted customer feedback were reachable **with no credentials
at all**. The `/feedback-forms` collection above them (`GET`, `POST`) *was* Cognito-protected, so
the collection was guarded and every item beneath it was not.

This closes that gap by declaring the item routes explicitly instead of behind a catch-all proxy,
and adds a template-level invariant so it cannot silently reopen.

There is no separate issue for this: **this PR is the record.** It was fixed and verified before
being published, and nothing here is a working request example.

## Root cause

`api-stack.ts` calls `addProxy` **14 times. 12 pass authorization options** (11 ×
`authMethodOptions`, 1 × `mcpMethodOptions` for `/mcp`). The only two that omitted
`defaultMethodOptions` entirely were the two feedback-form proxies, so their methods defaulted to
`AuthorizationType: NONE`. On the plural resource the two collection methods immediately above the
proxy *do* pass it — which is exactly why the collection was protected and the items were not.

So this was a **missing argument, not a missing design**. The handler did not compensate:
`lambda/api/feedback_form_handler.py` imports no auth helper at all, so nothing checked the caller
at either layer.

Two things kept it invisible:

- `nag-suppressions.ts` suppressed `AwsSolutions-APIG4` and `AwsSolutions-COG4` across the whole
  proxy subtree with the reason "intentionally public to allow anonymous customer feedback
  submission". That is true for the widget routes and false for the rest, and the suppression made
  the entire prefix look assessed. **An existing suppression is not evidence a finding was ever
  assessed.**
- Checkov's `CKV_AWS_59` surfaced it because it does not read cdk-nag suppressions.

## What changed

The `{proxy+}` is replaced by explicit resources for all ten routes the handler registers:

| Auth | Method | Route |
|---|---|---|
| Cognito | `GET`, `POST` | `/feedback-forms` |
| Cognito | `GET`, `PUT`, `DELETE` | `/feedback-forms/{form_id}` |
| Cognito | `GET` | `/feedback-forms/{form_id}/submissions` |
| Cognito | `GET` | `/feedback-forms/{form_id}/stats` |
| **public** | `GET` | `/feedback-forms/{form_id}/config` |
| **public** | `POST` | `/feedback-forms/{form_id}/submit` |
| **public** | `GET` | `/feedback-forms/{form_id}/iframe` |

The public set is exactly what the embeddable widget needs, read off
`lambda/api/static/feedback-widget.js` rather than guessed. The APIG4/COG4 suppressions are narrowed
from both proxy subtrees to those three methods, so a suppression can no longer cover a route nobody
intended to publish.

**No frontend change was required:** `fetchApi` → `buildHeaders` → `getAuthHeaders` already attaches
the bearer token to every admin-UI call.

### Why not keep the proxy and re-open three routes?

That was the obvious shape and it does not work. The three public routes live *under* `{form_id}`,
and API Gateway rejects `{form_id}` as a sibling of `{proxy+}` — only one variable path part is
allowed per level. Declaring them therefore requires the proxy to go.

Explicit routes are also better on the merits: they **fail closed**. A route added to the handler is
now unreachable until it is wired here, instead of inheriting a catch-all's absent authorization.
Cost is ~25 extra CloudFormation resources (354 total, limit 500).

### Also in this PR

- **Removed the legacy singular `/feedback-form/*` surface.** The handler implements no routes under
  that prefix, so the proxy could only 404 while still costing a Lambda invocation — and it is what
  made the same mistake on the plural proxy look idiomatic. Its four `client.ts` methods had no
  callers; their tests asserted call shape against a mocked `fetch`, which is why they stayed green
  against routes that no longer existed.
- **Corrected five tracked docs** that described a `/feedback-form/*` API. The steering route table
  listed four routes that do not exist, omitted five that do, and marked
  `GET /feedback-forms/{id}` public — the mental model that produced this defect. It is now an
  explicit auth allowlist.

## ⚠️ Upgrading an existing deployment takes two deploys

**Fresh deployments are unaffected.** Existing ones are not, and this is documented in
`docs/deployment.md` troubleshooting (second commit).

Deploying this in one update fails and rolls back:

```
CREATE_FAILED  AWS::ApiGateway::Resource  .../feedback-forms/{form_id}
A sibling ({proxy+}) of this resource already has a variable path part -- only one is allowed
```

CloudFormation creates new resources before deleting old ones within a single update, so
`{form_id}` and `{proxy+}` coexist momentarily and API Gateway refuses the pair. Found by deploying
it. The rollback was clean and the API kept serving throughout.

The upgrade is: deploy once with the `feedbackFormItem` block commented out (removes the proxy),
then restore and deploy again. Between the two deploys the per-form routes are unavailable, so the
embeddable widget stops working for that window. They fail closed, which is still an improvement on
being open.

## Regression guard

New `lib/stacks/api-stack.test.ts` — the first test to synthesize `VocApiStack`.

The defect was a missing argument on constructs that did not exist individually, so a test asserting
"these four routes are protected" could not have caught it. The guard is therefore an **invariant
over the whole template**: every non-`OPTIONS` method must carry an authorizer, except a named
three-route allowlist. (`OPTIONS` is excluded because `defaultCorsPreflightOptions` generates
unauthenticated CORS preflight methods, which cdk-nag's own APIG4 rule also excludes.)

It also holds the stack and handler in step by reading the handler's registered routes as an
independent oracle — now load-bearing, since without the catch-all an unwired route returns
`403 Missing Authentication Token`.

**Fail-on-revert proven:** all 7 tests fail against the previous `api-stack.ts`, and the failure
output names the two offending methods.

## Verification

- `cdk synth` clean, **zero cdk-nag findings** after narrowing the suppressions.
- Synthesized template: 102 methods, 55 CORS `OPTIONS`, **exactly 3 non-`OPTIONS` methods without an
  authorizer**, and zero singular `feedback-form` resources.
- **Deployed and verified live.** 16 probes against a real deployment, using a throwaway form that
  was deleted afterwards: the 6 previously-public routes return 401 unauthenticated; the 3 widget
  routes return 200 unauthenticated **with usable payloads** (correct config title, ~17KB of iframe
  HTML, `submit` returning a `feedback_id`) — status codes alone would have accepted a broken body;
  the 4 admin routes return 200 with a token; the legacy singular route is gone.
- **Full async path re-checked:** an unauthenticated submission still traverses SQS → processor →
  feedback store and is readable only with credentials. Note `GET .../submissions` immediately after
  `submit` returns 0 by design — `submit` is `sqs.send_message` and the processor writes the table.
- `scripts/test-api.sh` **32/32**.
- Gates: `test:cdk` 145, frontend 1801, stream 257, handler pytest 15, `tsc` and ESLint clean.

## Relationship to existing issues

- **#242** lists `feedback_form_handler.py` but argues severity from "the endpoints are
  Cognito-authenticated, so a caller must be a valid user". That premise did not hold for this
  handler — the methods required no credentials. #242 should be narrowed to the genuinely
  Cognito-authenticated handlers and point here for feedback forms. I will post that correction
  separately so the two do not contradict each other.
- **#222** hardens the public submit path (throttle, payload validation) and explicitly excludes
  auth from its scope. Complementary; neither supersedes the other.
- **#246** no WAF — the reason no edge control compensated. The only limit was the stage-wide
  throttle, shared with authenticated traffic.
- **#255** is the same rot pattern (a caller left behind after a route was removed).

## Follow-ups deliberately not in this PR

- `frontend/public/feedback-widget.js` is a ~15KB duplicate of the widget with **zero code
  references**, published to the CDN, still calling the singular routes. It was already broken; this
  change makes it permanently so. Delete or repoint — but not inside a security fix.
- Website-bucket S3 access logging, and CDK `BucketDeployment`'s `cloudfront:CreateInvalidation` on
  `"*"` (same family as #234) — both one-file, low value.
- Checkov suppressions for the genuinely-public routes, which must come **after** this lands or they
  re-hide the finding.
- `.kiro/prompts/aws-security-review.md` still asserts WAF protection exists and cites a stack file
  that no longer exists; folds into #246.

## Notes for the reviewer

- **`npm run check` is red on this branch, and it is red on `development` too** — `lint:python`
  (ruff over `lambda/`, `plugins/`) has pre-existing failures, tracked in #271. None of the files
  changed here are under `lambda/` or `plugins/`, so ruff's input is byte-identical to base. The
  individual gates listed above all pass.
- The working tree also carries an unrelated `.gitignore` modification; it is **deliberately not
  included** in either commit.

### DIFF
--- a/.kiro/prompts/aws-audit.md
+++ b/.kiro/prompts/aws-audit.md
@@ -37,7 +37,7 @@ VoC Data Lake is a **fully serverless** AWS platform for ingesting, processing, 
 | `voc-projects-api` | `projects_handler.py` | `/projects/*` | DynamoDB, Step Functions, Bedrock |
 | `voc-chat-stream` | `chat_stream_handler.py` | Function URL (streaming) | DynamoDB read, Bedrock streaming |
 | `voc-data-explorer-api` | `data_explorer_handler.py` | `/data-explorer/*` | S3, DynamoDB (feedback) |
-| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-form/*` | DynamoDB, SQS |
+| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-forms/*` | DynamoDB, SQS |
 | `voc-users-api` | `users_handler.py` | `/users/*` | Cognito, DynamoDB |
 
 ### DynamoDB Tables
@@ -84,7 +84,7 @@ voc-datalake/
 │   │   ├── settings_handler.py   # /settings/*
 │   │   ├── projects_handler.py   # /projects/*
 │   │   ├── data_explorer_handler.py # /data-explorer/*
-│   │   ├── feedback_form_handler.py # /feedback-form/*
+│   │   ├── feedback_form_handler.py # /feedback-forms/*
 │   │   ├── users_handler.py      # /users/*
 │   │   └── projects.py           # Shared business logic
 │   └── layers/                   # Lambda layers
--- a/.kiro/prompts/aws-security-review.md
+++ b/.kiro/prompts/aws-security-review.md
@@ -244,7 +244,7 @@ VoC Data Lake is a **fully serverless** AWS platform for ingesting, processing, 
 | `voc-settings-api` | `settings_handler.py` | `/settings/*` | DynamoDB, Bedrock |
 | `voc-projects-api` | `projects_handler.py` | `/projects/*` | DynamoDB, Step Functions, Bedrock, S3 |
 | `voc-users-api` | `users_handler.py` | `/users/*` | Cognito admin |
-| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-form/*`, `/feedback-forms/*` | DynamoDB, SQS |
+| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-forms/*` | DynamoDB, SQS |
 | `voc-chat-stream` | `chat_stream_handler.py` | Function URL (streaming) | DynamoDB read, Bedrock streaming |
 | `voc-data-explorer-api` | `data_explorer_handler.py` | `/data-explorer/*` | S3, DynamoDB (feedback) |
 
@@ -264,7 +264,10 @@ VoC Data Lake is a **fully serverless** AWS platform for ingesting, processing, 
 - Lambda Powertools (Logger, Tracer, Metrics) on all functions
 
 #### ⚠️ Known Considerations (Document but assess carefully)
-- **Public Feedback Form Endpoints**: `/feedback-form/submit` is public (required for form submissions)
+- **Public Feedback Form Endpoints**: exactly three routes are unauthenticated by design —
+  `GET /feedback-forms/{id}/config`, `POST /feedback-forms/{id}/submit` and
+  `GET /feedback-forms/{id}/iframe` (the embeddable widget runs on the customer's own site).
+  Every other route under `/feedback-forms` requires Cognito; `api-stack.test.ts` enforces this.
   - Mitigated by: Rate limiting, input validation
 - **Bedrock Global Inference Profile**: Uses cross-region inference for availability
   - Ensure IAM scoped to specific inference profile ARN
--- a/.kiro/steering/coding-standards.md
+++ b/.kiro/steering/coding-standards.md
@@ -150,7 +150,7 @@ lambda/api/
 ├── settings_handler.py      # /settings/*
 ├── projects_handler.py      # /projects/*
 ├── users_handler.py         # /users/* (Cognito admin)
-├── feedback_form_handler.py # /feedback-form/*, /feedback-forms/*
+├── feedback_form_handler.py # /feedback-forms/*
 ├── data_explorer_handler.py # /data-explorer/* (S3 raw data & DynamoDB browser)
 └── projects.py              # Shared business logic for projects
 ```
--- a/.kiro/steering/structure.md
+++ b/.kiro/steering/structure.md
@@ -258,6 +258,13 @@ embeddable widget calls from the customer's own site.
 > `api-stack.test.ts` asserts the handler and stack stay in step, and that only
 > the three public routes are unauthenticated.
 
+> ⚠️ **"Cognito" above means authenticated, NOT authorized.** These routes check
+> that the caller has a valid token; they do not check *which* forms the caller
+> owns. `feedback_form_handler.py` imports no auth helper, so **any authenticated
+> user can read, update or delete any form and read any form's submissions**.
+> That residual gap is tracked separately (same class as the missing per-user
+> scoping on projects) — do not read this table as "fully protected".
+
 ### Projects (projects_handler.py)
 | Method | Path | Description |
 |--------|------|-------------|
--- a/.kiro/steering/structure.md
+++ b/.kiro/steering/structure.md
@@ -51,7 +51,7 @@ voice-of-customer-datalake/       # Root repository
     │   │   ├── settings_handler.py       # /settings/* (brand, categories config)
     │   │   ├── projects_handler.py       # /projects/* (research projects, personas)
     │   │   ├── users_handler.py          # /users/* (Cognito user administration)
-    │   │   ├── feedback_form_handler.py  # /feedback-form/*, /feedback-forms/* (embeddable forms)
+    │   │   ├── feedback_form_handler.py  # /feedback-forms/* (embeddable forms)
     │   │   ├── data_explorer_handler.py  # /data-explorer/* (S3 raw data & DynamoDB browser)
     │   │   ├── logs_handler.py           # /logs/* (system logs)
     │   │   ├── manual_import_handler.py  # /manual-import/* (manual data import)
@@ -233,17 +233,30 @@ voice-of-customer-datalake/       # Root repository
 | POST | `/users/{username}/reset-password` | Reset password |
 
 ### Feedback Forms (feedback_form_handler.py)
-| Method | Path | Description |
-|--------|------|-------------|
-| GET | `/feedback-form/config` | Get form config (public) |
-| PUT | `/feedback-form/config` | Update form config |
-| POST | `/feedback-form/submit` | Submit feedback (public) |
-| GET | `/feedback-form/embed` | Get embed code |
-| GET | `/feedback-forms` | List all forms |
-| POST | `/feedback-forms` | Create form |
-| GET | `/feedback-forms/{id}` | Get form (public) |
-| PUT | `/feedback-forms/{id}` | Update form |
-| DELETE | `/feedback-forms/{id}` | Delete form |
+
+Every route requires Cognito **except** the three marked public, which the
+embeddable widget calls from the customer's own site.
+
+| Method | Path | Auth | Description |
+|--------|------|------|-------------|
+| GET | `/feedback-forms` | Cognito | List all forms |
+| POST | `/feedback-forms` | Cognito | Create form |
+| GET | `/feedback-forms/{id}` | Cognito | Get form |
+| PUT | `/feedback-forms/{id}` | Cognito | Update form |
+| DELETE | `/feedback-forms/{id}` | Cognito | Delete form |
+| GET | `/feedback-forms/{id}/submissions` | Cognito | Read submitted feedback |
+| GET | `/feedback-forms/{id}/stats` | Cognito | Submission count + average rating |
+| GET | `/feedback-forms/{id}/config` | **public** | Form config for the widget |
+| POST | `/feedback-forms/{id}/submit` | **public** | Submit feedback |
+| GET | `/feedback-forms/{id}/iframe` | **public** | Iframe embed variant |
+
+> There is no `/feedback-form/*` (singular) API. These routes are declared
+> **explicitly** in `api-stack.ts` rather than behind a `{proxy+}`, so adding a
+> route to the handler also requires wiring it there. That is deliberate: a
+> proxy without `defaultMethodOptions` defaults to `AuthorizationType: NONE`,
+> which is how form update/delete and submission reads were once public.
+> `api-stack.test.ts` asserts the handler and stack stay in step, and that only
+> the three public routes are unauthenticated.
 
 ### Projects (projects_handler.py)
 | Method | Path | Description |
--- a/.kiro/steering/tech.md
+++ b/.kiro/steering/tech.md
@@ -58,7 +58,7 @@ AWS Lambda execution roles have a **20KB policy size limit**. To stay under this
 | `voc-settings-api` | `settings_handler.py` | `/settings/*` | DynamoDB (aggregates), Bedrock |
 | `voc-projects-api` | `projects_handler.py` | `/projects/*` | DynamoDB (projects, jobs, feedback), Step Functions, Bedrock, S3 |
 | `voc-users-api` | `users_handler.py` | `/users/*` | Cognito admin |
-| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-form/*`, `/feedback-forms/*` | DynamoDB (aggregates), SQS |
+| `voc-feedback-form-api` | `feedback_form_handler.py` | `/feedback-forms/*` | DynamoDB (aggregates), SQS |
 | `voc-chat-stream` | `lambda/stream` (TypeScript, esbuild-bundled) | `/chat/stream` SSE via API Gateway | DynamoDB read, Bedrock streaming |
 | `voc-data-explorer-api` | `data_explorer_handler.py` | `/data-explorer/*` | S3, DynamoDB (feedback) |
 | `voc-logs-api` | `logs_handler.py` | `/logs/*` | CloudWatch Logs read |
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -622,6 +622,39 @@ Add the flag to every subsequent deploy of that environment (or to a
 local, uncommitted context override). New deployments should NOT set it —
 fresh pools get case-insensitive sign-in by default.
 
+### VocApiStack fails: "A sibling ({proxy+}) of this resource already has a variable path part"
+
+Only affects environments deployed **before** the `/feedback-forms` item
+routes were made explicit. Those routes used to sit behind a `{proxy+}`
+catch-all; they are now declared individually, which means the deploy has to
+create `/feedback-forms/{form_id}` and delete `/feedback-forms/{proxy+}`.
+CloudFormation creates new resources before deleting old ones inside a single
+update, so both variable path parts exist momentarily and API Gateway rejects
+the pair. The stack rolls back cleanly and the API keeps serving.
+
+**Fresh deployments are unaffected** — there is no proxy to remove.
+
+Upgrade an existing environment in two deploys. First remove the old proxy on
+its own, by temporarily commenting out the `feedbackFormItem` block in
+`lib/stacks/api-stack.ts` (everything from `addResource('{form_id}')` through
+the `publicFeedbackFormMethods` loop):
+
+```bash
+cdk deploy VocApiStack --exclusively
+```
+
+Then restore the file and deploy again to create the explicit routes:
+
+```bash
+git checkout -- lib/stacks/api-stack.ts
+cdk deploy VocApiStack --exclusively
+```
+
+Between the two deploys the per-form routes are unavailable, so the
+embeddable widget stops working until the second deploy finishes. They fail
+closed (no data is served), and the window is one deploy long. Schedule it
+accordingly if forms are live.
+
 ### CloudFront Cache
 
 If changes don't appear after deployment:
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -634,26 +634,35 @@ the pair. The stack rolls back cleanly and the API keeps serving.
 
 **Fresh deployments are unaffected** — there is no proxy to remove.
 
-Upgrade an existing environment in two deploys. First remove the old proxy on
-its own, by temporarily commenting out the `feedbackFormItem` block in
-`lib/stacks/api-stack.ts` (everything from `addResource('{form_id}')` through
-the `publicFeedbackFormMethods` loop):
+Upgrade an existing environment in two deploys, using the transitional flag —
+no source edits are involved:
 
 ```bash
+# 1. Retire the old {proxy+}. The per-form routes are not created yet.
+cdk deploy VocApiStack --exclusively -c skipFeedbackFormItemRoutes=true
+
+# 2. Create the explicit per-form routes.
 cdk deploy VocApiStack --exclusively
 ```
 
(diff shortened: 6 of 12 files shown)


perrozzi:

This PR closes a real and serious gap: `/feedback-forms/{proxy+}` was created without `defaultMethodOptions`, so form update/delete and submission reads were served with `AuthorizationType: NONE`. Replacing the proxy with explicit resources, narrowing the cdk-nag APIG4/COG4 suppressions to the three widget routes, deleting the dead singular `/feedback-form/*` surface, and adding a template-wide authorization invariant (`lib/stacks/api-stack.test.ts`) is the right shape, and the reasoning about why the proxy has to go (`{form_id}` cannot be a sibling of `{proxy+}`) checks out. I found no blockers; the notes below are about the migration ergonomics and the strength of the new guard.

### Should fix

- **Make the transitional deploy flag-driven instead of edit-driven** (`docs/deployment.md`, `lib/stacks/api-stack.ts`). The documented upgrade asks operators to comment out the `feedbackFormItem` block, deploy, then `git checkout -- lib/stacks/api-stack.ts` and deploy again. Editing infrastructure source mid-incident is error-prone, `git checkout --` silently discards any other local edits to that file, and the intermediate state is not reproducible or reviewable. A context flag (e.g. `-c legacyFeedbackFormProxy=remove` that skips the `{form_id}` subtree for one deploy) gives the same two-deploy sequence with no source mutation, and it can be exercised by a synth test.
- **The invariant is only proven for `enabledSources: []`** (`lib/stacks/api-stack.test.ts`). The comment acknowledges that plugin webhook receivers are unauthenticated by design, which means the "only three unauthenticated methods" assertion does not hold for any real deployment — exactly the configuration where a new unauthenticated route would ship. Add a second synth with a representative `enabledSources` value and an explicit webhook allowlist, so the guard covers the deployed shape rather than the empty one.
- **The lockstep oracle only covers the handler, not the callers.** Now that the catch-all is gone, an unwired path returns `403 Missing Authentication Token` rather than the handler's 404, so any `/feedback-forms/...` path still referenced by `frontend/src/api/client.ts` or the static widget that is *not* one of the ten wired routes fails opaquely. The diff removes four legacy client methods but does not show the remaining `feedback-forms` methods; please confirm (ideally by extending the test to grep client.ts and `lambda/api/static/feedback-widget.js` for `/feedback-forms` paths) that every caller-side path is wired.
- **Authentication is now enforced, authorization still is not.** `feedback_form_handler.py` imports no auth helper, so any authenticated Cognito user can read, update or delete any form and read any form's submissions. That is a large improvement over anonymous access and is arguably #242's scope, but the docs added in `.kiro/steering/structure.md` now read as if these routes are fully protected. Either add an ownership/tenant check or state the residual gap in the doc table.

### Nits

- `has no proxy resource left without explicit method options` greps single trimmed lines for `addProxy(` — a legitimately reformatted multi-line `addProxy({ ... defaultMethodOptions ... })` call would fail the test falsely. Consider matching the call expression (or asserting on the synthesized template's `AuthorizationType` per method, which you already do) instead of line-shaped source.
- The handler-route regex `@app\.(get|post|put|delete)\("([^"]+)"\)` only matches double-quoted paths and misses `@app.route(...)`; a route added with single quotes would be invisible to the oracle and pass silently. Widen to `['"]` and include `route`.
- The invariant only checks `authorizationType !== 'NONE'`; a method that regressed to `AWS_IAM` or a stray `CUSTOM` authorizer would pass. Asserting `COGNITO_USER_POOLS` plus a present `AuthorizerId` is barely more code.
- `frontend/public/feedback-widget.js` is left published to the CDN still calling the removed singular routes. Agreed it was already broken, but a one-line note in `docs/deployment.md` next to the new troubleshooting section would stop someone from debugging it as a symptom of this change.
- Worth a repo-wide grep for remaining `/feedback-form/` (singular) references beyond the six files touched here — README/API docs are easy to miss and were the source of the wrong mental model.

---

perrozzi:
_(Answering review round 1 on `9c9991d`. All items addressed in `ed20754`, pushed right after this comment.)_

Thanks — accurate round. All four should-fix items and all five nits are addressed. One claim I'm
pushing back on, with evidence, because the fix it implies would have been weaker than the one I made.

## ❌ One correction: no plugin declares a webhook, so the empty-plugin synth was not the gap it looked like

> the "only three unauthenticated methods" assertion does not hold for any real deployment

It does hold, today, for every real deployment. **No plugin manifest in this repo declares a
webhook** — `getPluginsWithWebhook()` filters on `p.infrastructure.webhook?.enabled`
(`plugin-loader.ts:331`) and no manifest sets it. So `getEnabledPlugins()` with all five plugin ids
still yields zero webhook receivers, which is why a live deployment with all five plugins loaded
serves exactly three unauthenticated methods.

That makes your underlying point sharper rather than wrong: an all-plugins synth is the right test,
but on its own it compares two **identical** shapes, so it is a tripwire wired to nothing until
someone adds a webhook plugin. I did both:

- `leaves only those three unauthenticated with every plugin enabled too` — synthesizes the deployed
  shape.
- `pins the fact that makes the allowlist complete: no plugin declares a webhook` — reads the
  manifests directly, so the assumption **fails today** the moment a manifest gains
  `infrastructure.webhook.enabled`, with a message pointing at the allowlist.

## ✅ Should fix

**Flag-driven transitional deploy.** Agreed, and your reasoning about `git checkout --` silently
discarding unrelated local edits is exactly right. Added `-c skipFeedbackFormItemRoutes=true`,
matching the existing `skipFrontendBuildCheck` / `skipUseCaseSubmission` convention:

```bash
cdk deploy VocApiStack --exclusively -c skipFeedbackFormItemRoutes=true   # retires the proxy
cdk deploy VocApiStack --exclusively                                      # creates the routes
```

Verified it is a genuine no-op when absent, two ways — the default synth is unchanged (102 methods,
3 unauthenticated, 8 item routes), and the **24 feedback-form API Gateway logical ids are identical
to the currently deployed template**, so a redeploy replaces nothing and cannot re-trigger the
sibling conflict. With the flag set: 88 methods and **zero** unauthenticated methods, i.e. the
transitional window is fully fail closed, plus a synth-time warning naming itself as deploy 1 of 2.
Both states are covered by tests.

**Invariant only proven for `enabledSources: []`.** Fixed as above.

**Lockstep oracle only covers the handler, not the callers.** Confirmed as you asked, and then
automated. `client.ts`'s remaining paths are `/feedback-forms`, `/{formId}`, `/{formId}/stats`,
`/{formId}/submissions?…` — all wired (query strings don't affect resource matching), and the widget
calls only `config` and `submit`. New test extracts `/feedback-forms…` paths from both `client.ts`
and `lambda/api/static/feedback-widget.js`, normalizes the form-id segment, and asserts each is
wired. Guarded with a non-empty check so it can't pass vacuously.

**Authentication vs authorization.** Correct and worth stating plainly, so
`.kiro/steering/structure.md` now carries it: these routes verify a token but not ownership, so any
authenticated Cognito user can read, update or delete any form and read any form's submissions. The
table read as "fully protected", which is the same overstatement that produced the original defect.
Flagged as the residual gap (#242's scope, same class as the missing per-user scoping on projects)
rather than fixed here — adding tenancy is a different change with its own data-model questions.

## ✅ Nits

- **Line-shaped `addProxy` check** — right, that was a false-positive risk on reformatting. Now
  matches the call expression over a 400-char window instead of a trimmed line.
- **Handler regex missed single quotes and `@app.route`** — widened to `['"]` and `route`.
- **`!== 'NONE'` would accept `AWS_IAM` or a stray `CUSTOM`** — now asserts `COGNITO_USER_POOLS`
  *and* a present `AuthorizerId`, with `/mcp` allowlisted and separately asserted to genuinely carry
  a custom authorizer.
- **Stale CDN widget** — added a `docs/deployment.md` section saying `frontend/public/feedback-widget.js`
  is not the live widget and has been broken independently of this change. Still not deleting it here;
  a 15KB public-artifact removal doesn't belong in a security fix.
- **Repo-wide grep for singular references** — done, and it found one you'd have wanted: the mock
  server still served four `/feedback-form/*` routes. Removed, along with the fixture that became
  unreferenced. Everything else remaining is generated (`Workshop/static/cfn`) or the stale widget
  above. Worth noting the trap: `grep -v 'feedback-forms'` false-negatives any line containing both
  prefixes, which is how the steering and prompt files nearly stayed wrong.

## Gates

`test:cdk` 154 (was 145), frontend 1801, `typecheck:cdk` and `typecheck` clean, `lint:frontend`
clean, `cdk synth` exit 0 with zero cdk-nag findings. `npm run check` remains red on pre-existing
ruff (#271), unchanged by this branch.

perrozzi:

Since my last pass the two-deploy migration became a real context flag (`skipFeedbackFormItemRoutes`) with a warning annotation, tests and docs; the guard grew an all-plugins synth, a webhook-manifest pin, a `COGNITO_USER_POOLS` + `AuthorizerId` assertion, caller-side path lockstep for `client.ts` and the widget, and the source scan now matches the call expression; the residual authorization gap and the stale `frontend/public/feedback-widget.js` are now documented; mock-server's legacy singular handlers were also dropped. All four should-fix items are addressed.

### Resolved
- Transitional deploy is now flag-driven (`-c skipFeedbackFormItemRoutes=true`), no source edits — `api-stack.ts`, `docs/deployment.md`, plus three synth tests covering the flagged/unflagged shapes.
- Invariant now also synthesized with every plugin enabled, and the "no plugin declares a webhook" premise is pinned explicitly.
- Caller-side lockstep added for `frontend/src/api/client.ts` and `lambda/api/static/feedback-widget.js`.
- Residual authorization (ownership) gap stated in `.kiro/steering/structure.md`; stale CDN widget documented in `docs/deployment.md`.
- Nits: `addProxy` matched on the call expression; handler regex widened to `route` and single quotes; assertions tightened to `COGNITO_USER_POOLS` + present `AuthorizerId`; `/mcp` custom authorizer given its own positive assertion.

### New findings

- **`ALL_PLUGIN_IDS` is a hardcoded list but documented as "every plugin on disk"** (`voc-datalake/lib/stacks/api-stack.test.ts`). A newly registered plugin is invisible to both the all-plugins invariant *and* the webhook-manifest pin — i.e. exactly the "new plugin ships an anonymous webhook" case the two tests exist to catch. Enumerate `plugins/` with `readdirSync(..., { withFileTypes: true })` and assert the list is non-empty instead.
- **The webhook pin fails open on a manifest shape change** (same file). `safeParse(...)` + `parsed.success && ...` treats any schema mismatch as "no webhook", so renaming/moving the `infrastructure.webhook.enabled` field silently disables the guard. Parse with the repo's canonical plugin-manifest schema (or `.parse` with a passthrough on unrelated fields) so a shape change is loud.

### Nits

- `cdk.Annotations.of(this).addWarning(...)` is the deprecated form; `addWarningV2('voc:skipFeedbackFormItemRoutes', ...)` gives an acknowledgeable id and avoids jsii deprecation noise (`api-stack.ts`).
- `callerFormsPaths` misses concatenated paths: `'/feedback-forms/' + formId + '/submit'` normalizes to `/feedback-forms` and passes; conversely a real collection subresource (`/feedback-forms/templates`) would be normalized to `{form_id}` and pass spuriously. Fine as a smoke test, worth a comment saying so.
- The `addProxy(` scan slices a fixed 400-char window, so `defaultMethodOptions` just past the window false-fails and a nearby unrelated occurrence false-passes; brace-matching or a per-call parse would be tighter.
- `flagged()` re-synthesizes on each of its three uses while the other two shapes are cached — cheap to memoize the same way.
- The flag is a permanent footgun once the migration window closes; a tracking note ("remove after \<release>") next to it would help it not outlive its purpose.

---

perrozzi:
_(Answering review round 2 on `ed20754`. All items addressed in `6e6983a`, pushed right after this comment.)_

Both new findings were right, and they're the same defect in the two guards round 1 asked for: the
guard was sound, its **input** wasn't. Nothing to push back on this round.

## Should fix

**`ALL_PLUGIN_IDS` hardcoded while documented as "every plugin on disk".** Fixed — `plugins/` is now
enumerated. Your framing was exactly the failure mode: a new plugin was invisible to *both* the
all-plugins invariant and the webhook pin, which is the one case they exist for.

Enumeration has to filter the way `loadPlugins` does, because `plugins/` also holds `conftest.py`,
five `test_*.py` files, `__pycache__`, `_shared/` and `_template/` — and `_template/manifest.json`
**exists**, so a naive `readdirSync` would have treated the scaffold as a deployable plugin. Filter is
"is a directory, not `_`-prefixed, has a manifest.json", and the enumeration is now pinned by its own
test since two guards depend on it.

**The webhook pin failed open.** Correct, and the more serious of the two: `safeParse` plus
`parsed.success && …` read any schema mismatch as "no webhook", so renaming or moving
`infrastructure.webhook.enabled` would have silently disabled the guard. It now parses with the
canonical `ManifestSchema`, which I exported from `plugin-loader.ts` rather than re-declaring a
partial shape in the test — the same re-declaration is what made it fail open. Shape drift now throws.

### Both fixes verified to bite

Planted a valid plugin manifest declaring an enabled webhook. **3 of 17 tests fail, and they are the
right 3:** the all-plugins invariant, the webhook pin (naming the offending plugin in its message),
and the `COGNITO_USER_POOLS` assertion. The default-template test correctly keeps passing, since
`enabledSources: []` is unaffected by a new plugin on disk.

Worth noting what that proves: with the previous hardcoded list, the planted plugin was invisible and
**all 16 tests passed**. The enumeration is what turns these two from documentation into tests.

(Two false starts getting there, both instructive: the plugin loader rejected my first probe manifest
for a missing `infrastructure.webhook.path`, and the second for a path not starting with `/`. Those
rejections are `loadPlugins` doing its job.)

## Nits

- **`addWarning` deprecated** — now `addWarningV2('voc:skipFeedbackFormItemRoutes', …)`. Confirmed
  present in the pinned `aws-cdk-lib` before switching.
- **Fixed 400-char window on the `addProxy` scan** — now balances parentheses to delimit each call's
  argument list, so neither a distant `defaultMethodOptions` nor a nearby unrelated occurrence can
  skew it.
- **`callerFormsPaths` limitations** — documented in the code as you suggested, naming both: a
  concatenated path collapses to the collection and passes unchecked, and a real collection
  subresource would normalize to `{form_id}` and pass spuriously. Left as a smoke test; a full
  solution means parsing the TypeScript, which isn't worth it for the case it catches.
- **`flagged()` re-synthesized** — memoized like the other two shapes.
- **Flag is a permanent footgun** — agreed, and the honest fix is a deletion note rather than a
  mechanism: added a `TODO(cleanup)` saying to remove the flag, its branch and its tests once every
  environment has run the two-deploy upgrade.

## Gates

`test:cdk` **155** (was 154), `typecheck:cdk` clean, `cdk synth` exit 0 with zero cdk-nag findings in
both flag states, and the flagged synth still emits its warning. `npm run check` remains red on
pre-existing ruff (#271), untouched by this branch.

perrozzi:

Since my last pass the two flagged items were addressed directly: `ALL_PLUGIN_IDS` is gone in favour of `discoverPluginIds()` enumerating `plugins/`, and the webhook pin now uses the newly exported canonical `ManifestSchema` with `.parse` so shape drift throws. The nits were also picked up — `addWarningV2` with an acknowledgeable id, brace-matched `addProxy(` argument extraction, memoized flagged template, documented limits on `normalizeFormsPath`, and a `TODO(cleanup)` next to the transitional flag.

### Resolved
- `ALL_PLUGIN_IDS` hardcoded list → `discoverPluginIds()` over `plugins/`, with a dedicated test pinning the enumeration itself (`api-stack.test.ts`).
- Webhook pin no longer fails open: `ManifestSchema.parse` from `plugin-loader.ts` replaces the local partial schema + `safeParse`.
- Nits: `addWarningV2('voc:skipFeedbackFormItemRoutes', …)`, brace-matched `addProxy` argument slice, `apiTemplateFlagged()` memoized, smoke-test limits documented, flag-removal TODO added.

### New findings

Nothing blocking. Two nits:

- `discoverPluginIds()` claims to "filter the way `loadPlugins` does", but it only checks `isDirectory() && !name.startsWith('_') && manifest.json exists` (`api-stack.test.ts`). If `loadPlugins` additionally rejects a directory (e.g. manifest fails validation or an `enabled: false`/integrity gate), the two lists drift and the all-plugins synth may be passed an id the loader ignores or rejects. Worth either reusing the loader's own discovery or noting that the filter is intentionally looser.
- The paren-matching `addProxy(` scan is not string/comment aware, so an unbalanced `(` inside a string literal or comment within the argument list would mis-delimit the slice (`api-stack.test.ts`). Fine in practice; a one-line caveat comment would be enough. Also `ids.filter((id) => id.endsWith('.py'))` is vacuous given the `isDirectory()` filter.

---

perrozzi:
_(Answering review round 3 on `6e6983a`. Nit-only round, so stopping the loop here without further pushes — each push costs a review round and this one reports nothing blocking.)_

Convergence arc across the three rounds: 4 should-fix + 5 nits → 2 findings + 5 nits → **0 findings,
2 nits**. Both remaining nits are fair and I'm recording the disposition rather than churning the diff.

**`discoverPluginIds()` "filters the way `loadPlugins` does" is an overstatement — you're right.** It
checks `isDirectory() && !name.startsWith('_') && manifest.json exists`, which is *discovery*, not the
loader's full acceptance logic; a manifest that fails validation or is gated off would still appear in
the test's list. Practically that's benign in both places it's used — `getEnabledPlugins()` intersects
by id so an id the loader dropped simply contributes nothing to the synth, and the webhook pin *wants*
to read every manifest on disk rather than only loadable ones (a plugin that is broken today but
declares a webhook is still the thing worth catching). But the comment claims more than the code does,
so treat the wording as the defect. Follow-up, if anyone touches this: reuse the loader's own discovery
and state explicitly that the pin deliberately scans wider.

**The paren matcher isn't string/comment aware** — correct, an unbalanced `(` inside a string or
comment within an `addProxy` argument list would mis-delimit the slice. Accepted as-is: the
authoritative check is the template-level assertion on `AuthorizationType`, and this source scan is
belt-and-braces that names the defect in the form it was originally made.

**`ids.filter((id) => id.endsWith('.py'))` is vacuous given `isDirectory()`** — also correct, and the
one I'd call an actual mistake rather than a wording issue, since an assertion that cannot fail is
exactly what the rest of this file exists to avoid. It weakens nothing (the live assertions are
non-empty, no `_`-prefixed, and contains a known plugin), but it should be either removed or replaced
with something that can fail. Noting it here rather than pushing a fourth time for a redundant line.

Merging as a squash. For the record on what is and isn't verified: gates are `test:cdk` 155, frontend
1801, stream 257, handler pytest 15, `typecheck`/`typecheck:cdk`/`lint:frontend` clean, and `cdk synth`
exit 0 with zero cdk-nag findings in both flag states; the authorization change itself was deployed and
exercised live against a real environment (six previously-public routes rejecting anonymous callers,
the three widget routes still working without credentials, and an unauthenticated submission still
reaching the feedback store). This repo has no CI, so those figures are local runs, and the PR carries
no human review — flagging that rather than implying otherwise.

PR #278, opened 2026-08-07
fix(ui): report the true urgent count, and remove the dead /feedback chip

perrozzi:
Two independent UI defects, both of which caused the app to display wrong or dead information in its
persistent chrome. Neither changes any layout.

## 1. The urgent-count badge showed a different number on every page

Observed on a running deployment: sidebar badge **5** on Dashboard, **10** on Categories, while the
Dashboard's own "Urgent Issues" metric card read **11** at the same moment. `11` was correct.

Two independent causes compounded:

**Cache-key collision.** `Layout.tsx` and `Dashboard.tsx` both requested `/feedback/urgent` under the
identical query key `['urgent', dateParams]` while passing different limits (10 and 5). `limit` was
not part of the key, so both resolved to one cache entry and the badge rendered whichever response
landed last — the list's page size on Dashboard, its own limit elsewhere.

**The endpoint returns a page length, not a total.** `metrics_handler.py` ends
`get_urgent_feedback` with `return {'count': len(items), 'items': items[:limit]}`, and the scan loop
breaks once `len(items) >= limit`. So `count` is clamped by the limit it was called with, and the
badge could never report more than 10 regardless of the real number.

Either fix alone leaves the badge wrong, so both are addressed:

- The badge now reads `urgent_count` from `/metrics/summary`, which sums the precomputed
  `METRIC#urgent` daily aggregates and is exact for the window. It **deliberately shares Dashboard's
  `['summary', dateParams]` key** — identical parameters, one cache entry — so the badge and the
  metric card cannot disagree by construction. `Layout` no longer fetches the urgent list at all,
  removing one request per page.
- `limit` is now part of Dashboard's urgent-list key, so a future caller with a different limit
  cannot recreate the collision.
- The `Urgent Issues (N)` heading had the same defect — it reported the preview list's page size. It
  now reports the summary total, so a window with 11 urgent items showing a 5-item preview is headed
  11, not 5.
- The preview size became a named constant, which also resolved a mismatch found on the way: the
  query fetched 5 items while the component sliced to 6.

## 2. The landing page rendered a raw i18n key and linked to a removed route

`Home.tsx` built the "Read the signals" card with
`{ to: '/feedback', labelKey: 'common:nav.feedback' }`. Both halves were stale after the standalone
Feedback list was consolidated into Categories (#198): `nav.feedback` no longer exists in
`common.json`, so i18next fell back to echoing the key, and `/feedback` survives only as a redirect.
The card already carried a Categories chip, so the dead chip was also a duplicate.

**The app's default route displayed a chip labelled literally `nav.feedback`, in all 8 locales.**

Removed with it, from the same consolidation:

- the orphaned `feedback` locale namespace — 12 keys × 8 locales with no consumer, still registered
  in `i18n/config.ts`, `scripts/i18n-check.mjs` and `test/setup.ts`
- the stale `breadcrumbs.feedback` key

Verified no component resolves that namespace, by name or via an explicit `feedback:` key prefix.

## Verification

This repo has no CI, so every figure below is a local run.

- `tsc -b` clean, `eslint .` clean (0 warnings, gate is `--max-warnings 0`)
- **1802 frontend tests pass** (114 files)
- 4 new badge tests in `Layout.test.tsx`, 1 in `Dashboard.test.tsx`
- **Fail-on-revert proven:** restoring the `getUrgentFeedback` source fails exactly 3 of the 4 new
  Layout tests. The fourth (zero-state) correctly still passes, because both paths agree on 0 — the
  failing *set* is the evidence the tests are targeted rather than broad.
- **Verified visually** against the repo's mock server, whose two sources diverge naturally
  (`/metrics/summary` reports 23; `/feedback/urgent` reports 1). Badge reads **23** on Home,
  Dashboard, Categories, Settings and Projects — consistent across routes — and the urgent section is
  headed **"Urgent Issues (23)"**. Before this change the badge showed 1 beside a card showing 23.
  Home renders **zero** raw i18n key tokens, and the German locale still resolves after the namespace
  removal.

## Two things a reviewer should know

**1. This inherits a fan-out cost, and it is not addressed here.** `/metrics/summary` loops `days`
sequential `get_item` calls three times over (daily totals, sentiment, urgent), so a 90-day window is
roughly 270 round-trips against about one query for the list call the badge used before — and it now
runs on pages that never needed a summary. Request *count* is unchanged (one per staleTime window per
`dateParams`, shared with Dashboard); the single request is heavier.

The fix belongs in the handler, not the frontend, and **the same file already shows the right shape
three times over**: the sources, personas and entities blocks build a date set and issue one query
filtered in memory, instead of one `get_item` per day. That is a backend change needing its own
benchmarks, deliberately out of scope for a badge-correctness fix. A `COST NOTE` in `Layout.tsx`
points at it so it is not lost.

**2. `npm run i18n:validate` exits 1 on this branch — and on `development`.** It reports
`missing 0 / extra 0 / empty 40 / untranslated 0`. Those 40 empties are pre-existing
plural-inflection debt in de/es/fr/pt. Confirmed by running the same gate in a clean worktree of the
base commit: **identical totals**. This matters here specifically because removing a locale namespace
is exactly the edit that could introduce parity errors, and it introduced none.


perrozzi:

This PR fixes two real UI defects: the sidebar urgent badge now reads the exact `urgent_count` from `/metrics/summary` instead of the limit-clamped `count` from `/feedback/urgent`, `limit` is added to Dashboard's urgent query key to break the cache-key collision, and the dead `/feedback` chip plus the orphaned `feedback` locale namespace are removed. The diagnosis is correct and the frontend changes are clean and well-commented; my concerns are the missing test for the Home/i18n half, the fact that the shared `['summary', dateParams]` key is enforced only by comment, and that the root-cause backend semantics are left untouched.

### Blockers

- **No regression test for defect #2.** `src/pages/Home/Home.tsx` removed a chip that rendered the literal string `nav.feedback` in all locales, and `nav.feedback` was deleted from 8 `common.json` files plus the whole `feedback` namespace from `i18n/config.ts`, `scripts/i18n-check.mjs` and `test/setup.ts` — with zero test coverage. Per the repo rule, a bug fix needs a test that fails when the fix is reverted. Add a `Home.test.tsx` case asserting the phase-2 links render no raw key text (e.g. no node matching `/^[a-z]+\.[a-zA-Z]+$/` / no `nav.feedback`), and ideally extend `scripts/i18n-check.mjs` to assert every `labelKey`/`titleKey` literal in `src/` resolves in the source locale — that is the check that would have caught this class of bug in the first place, and would catch it again after the next route consolidation.

### Should fix

- **The deliberate key sharing is enforced by a comment, not by code.** `Layout.tsx` uses `queryKey: ['summary', dateParams]` / `queryFn: () => api.getSummary(dateParams)` and relies on `Dashboard.tsx` doing exactly the same. That is the same failure mode this PR is fixing (`['urgent', dateParams]` with divergent args), just pointed at a different endpoint: the moment either caller passes an extra argument to `getSummary` (project filter, extra flag) the two observers silently share one entry again. Extract a `useSummaryQuery(dateParams)` hook or a `queryKeys.summary(dateParams)` factory that both call, so the invariant is structural. While there, confirm the two queries' options match — Layout uses `enabled: !!config.apiEndpoint` and Dashboard `enabled: isConfigured`, and if only one sets `staleTime`/`refetchOnMount` the shared entry will refetch on mount of the stricter observer, undoing the "request count is unchanged" claim.
- **The clamped `count` in `metrics_handler.get_urgent_feedback` is still there.** The frontend now routes around it, but the field remains misleading for any current or future consumer, and there is no pytest regression pinning the behaviour. Either fix the handler to return a true total, or rename the field to something honest (`returned_count`, plus `has_more`) and add a pytest case; a frontend-only workaround leaves the trap armed.
- **Heading/badge can now contradict the list it sits above.** `UrgentFeedback` gets `count` from `summary?.urgent_count` but `items` from `/feedback/urgent` — two different data sources (precomputed `METRIC#urgent` aggregates vs. a table scan). If aggregates are missing/not backfilled you get `Urgent Issues (0)` above a populated list; if the window boundaries differ you get `(11)` above an empty state. Worth a small guard or at least a comment/test documenting the intended behaviour when the two disagree (`Dashboard.tsx:250`, `409`).
- **Description/diff mismatch:** the description says the stale `breadcrumbs.feedback` key was removed, but the `common.json` diffs contain only the `nav.feedback` hunk. Either it was missed or the description overstates the change.

### Nits

- `Layout.tsx`: the `COST NOTE` cites `metrics_handler.py:684, 692, 701`. Line numbers rot on the next edit — reference the function/block names instead, and link a tracking issue rather than carrying a 15-line essay in a component body.
- `Layout.test.tsx`: `mockGetUrgentFeedback.mockResolvedValue(...)` is still seeded in three `beforeEach` blocks where the component no longer calls it; only the negative-assertion test needs it.
- `Dashboard.test.tsx`: the new test depends on the shared fixtures diverging (summary 5 vs. preview 3). Assert that precondition inline (or set the mocks in the test) so a future fixture tidy-up doesn't quietly make the test vacuous.
- `URGENT_PREVIEW_LIMIT` fixes the 5-fetch/6-slice mismatch — nice catch; consider asserting `items.length <= URGENT_PREVIEW_LIMIT` in a test so the two can't drift again.

---

perrozzi:
_(Answering review round 1 on `4695a93`. Addressed in `cd05dbf`, pushed right after this comment.)_

Good review — the blocker was right and two of the should-fixes found things I had considered and got wrong. One claim is inverted; details at the end.

## Blocker — no regression test for defect #2 ✅ fixed, both halves

`Home.test.tsx` gains two cases (no raw i18n key renders anywhere; no link targets `/feedback`). Both fail when the chip is restored — verified, not assumed.

The structural suggestion was the more valuable half, so I took it: **`scripts/i18n-check.mjs` now also collects namespace-qualified keys held in *data*** (`labelKey: 'common:nav.categories'`), not just keys passed directly to `t()`. That was exactly the blind spot — the key reaches `t()` indirectly as `t(item.labelKey)`, so the `t()` regex never saw it, which is how the gate reported "all translations in sync" while the landing page rendered the literal text `nav.feedback`.

Proven both ways:
- restoring the chip now makes the gate print ``❌ common:nav.feedback`` under keys-missing-from-English
- no false positives: only `ns:dotted.path` literals are collected, and totals across the existing source are unchanged (`missing 0`)

## Should fix — key sharing enforced by comment, not code ✅ fixed

Fair, and it's the same failure mode this PR fixes pointed at a different endpoint. New `hooks/useSummaryQuery.ts` owns the key *and* the fetcher; `Layout` and `Dashboard` both call it, so a future parameter has to be threaded through one place and lands in the key and the request together.

On the two sub-claims: both are unfounded on current code, though the fix stands regardless. Dashboard's `isConfigured` is literally `!!config.apiEndpoint` (`Dashboard.tsx:310`), the same expression Layout used — so the `enabled` values were identical, not divergent. And neither caller sets `staleTime` or `refetchOnMount`; both inherit the global `staleTime: 30000` from the `QueryClient` in `App.tsx`, so the "shared entry refetches on mount of the stricter observer" scenario had no antecedent.

## Should fix — heading can contradict the list ✅ fixed, and my first attempt was wrong

You were right that I introduced this. My first fix was `summary?.urgent_count ?? items.length`, and writing the test proved it useless: **the reachable case is an aggregate that legitimately reads `0` while the scan returns items, and `0` is not nullish.** A summary *failure* isn't the path — that swaps the whole dashboard for its empty state, which is why my first test couldn't even find the heading.

`UrgentFeedback` now takes `aggregateTotal` and reconciles it against its own rendered items in `urgentHeadingCount`, taking the larger of the two. That holds both invariants: never understate the total (reporting the page size was the original defect), and never claim fewer items than are visibly rendered. Two tests cover it, including a cap test pinning the fetch limit to the render cap.

## Should fix — the clamped `count` is still there ✅ pinned, not renamed

Agreed the trap stays armed, so it is now documented at both ends: `test_count_is_the_returned_page_length_not_the_window_total` pins the semantics, and a comment at the `return` site states that `count` is a page length clamped by `limit` and points at `/metrics/summary` as the only correct source for a total.

I did **not** rename the field or add `has_more`. That changes a response shape with a live consumer (the dashboard preview list), so it belongs in its own commit rather than riding along in a badge fix.

## Nits ✅ all four

- `COST NOTE` now names the handler blocks instead of line numbers, which rot.
- Dropped the dead `getUrgentFeedback` seeding from three `beforeEach` blocks; the two tests that need it set it themselves.
- The Dashboard regression test sets both mocks inline instead of relying on shared fixtures happening to diverge — the vacuous-test risk you flagged was real.
- Added the `items.length <= URGENT_PREVIEW_LIMIT` assertion.

Also, extracting `urgentHeadingCount` was partly forced: the inline expression pushed `Dashboard` to complexity 15 against the eslint ceiling of 12.

## One claim is inverted 🔍

> the description says the stale `breadcrumbs.feedback` key was removed, but the `common.json` diffs contain only the `nav.feedback` hunk

It's the other way round. `"breadcrumbs"` opens at `common.json:5` and the removed `"feedback": "Feedback"` sat at `:10`, inside that block — so the hunk **is** the breadcrumbs removal, and the description is accurate.

`nav.feedback` was already absent from `common.json`. **That absence is the bug**: `Home.tsx` referenced a key that no longer existed, so i18next fell back to echoing it. There was no `nav.feedback` hunk to produce, because there was nothing left to delete.

## Gates

tsc 0 · eslint 0 · vitest **1806** (was 1802) · pytest **1043** · i18n `missing 0 / extra 0 / empty 40`.

The 40 empties are pre-existing plural-inflection debt in de/es/fr/pt — identical totals in a clean worktree of `development`, which is the check that matters here since removing a locale namespace is the edit most likely to introduce parity errors.

perrozzi:

Follow-up review: the author extracted `useSummaryQuery` (shared key + fetcher), added the missing Home regression tests plus a new `labelKey:`-in-data check to `i18n-check.mjs`, made the Dashboard urgent tests own their fixtures, added a `urgentHeadingCount` reconciliation, and documented/pinned the clamped `count` in `metrics_handler.py`. Nearly all previous findings are addressed; the remainder are small.

### Resolved
- Blocker: `Home.test.tsx` now has two regression tests (no raw key tokens, no `/feedback` href), plus `i18n-check.mjs` now extracts `ns:dotted.path` literals from data tables — exactly the gate that would have caught this class of bug.
- Shared summary key is now structural via `hooks/useSummaryQuery.ts`, and both callers use identical options, so the "request count unchanged" claim holds.
- Heading/list divergence is handled by `urgentHeadingCount` (`Math.max`) with a dedicated `urgent_count: 0` vs. populated-list test.
- Dashboard urgent test now sets its own summary/list mocks instead of relying on shared fixtures; `URGENT_PREVIEW_LIMIT` render cap is pinned by a test.
- `COST NOTE` no longer cites rot-prone line numbers; stale `mockGetUrgentFeedback` seeds removed from `Layout.test.tsx` `beforeEach` blocks.

### Still open
- `breadcrumbs.feedback` removal is still claimed in the description but absent from the diff.
- `get_urgent_feedback` still ships a misleading `count`; documented and pinned, but not renamed or given a `total`/`has_more` companion.

### New findings

**Should fix**

- `voc-datalake/frontend/scripts/i18n-check.mjs`: `dataKeyRegex` accepts any `\w+` as namespace, so any property whose name ends in `Key` holding a `foo:bar` string (`cacheKey: 'user:123'`, an S3/DDB-style `sortKey: 'a:b'`, etc.) is collected as a translation key and will surface as a bogus *missing* key, failing the gate. Constrain the namespace group to the known `NAMESPACES` list (and skip matches whose namespace isn't in it) so the new check can't produce false failures.
- `voc-datalake/frontend/src/hooks/useSummaryQuery.ts`: `enabled: apiEndpoint.length > 0` throws if `apiEndpoint` is ever `undefined` at runtime (persisted/rehydrated config), where the previous `!!config.apiEndpoint` was safe. Use `!!apiEndpoint` (and/or type the param `string | undefined`).

**Nits**

- `voc-datalake/lambda/api/test/test_metrics_handler.py`: `assert body['count'] <= 3` pins the buggy semantics such that the intended future fix (true total) will fail this test — fine if deliberate, but note it in the docstring as "delete this when the field is renamed". Also `len(body['items']) <= 3` is near-vacuous; `== 3` would be tighter. The inline `sys.path.insert` / local import inside the test body is out of step with the rest of the file.
- `Dashboard.test.tsx` "renders no more urgent cards than the preview limit fetches" only pins the render slice; add `expect(mockGetUrgentFeedback).toHaveBeenCalledWith(expect.objectContaining({ limit: 5 }))` to actually pin fetch-vs-render.
- `Dashboard.tsx` now mixes `isConfigured` (other queries) with `config.apiEndpoint` (summary); pass `isConfigured`-equivalent consistently for readability.

---

perrozzi:
_(Answering review round 2 on `cd05dbf`. Addressed in `ba722de`, pushed right after this comment.)_

Both should-fixes were right, and one nit turned out to be more than a nit.

## Should fix — the new check could fail the gate on valid code ✅ fixed

Correct, and a real hazard: `dataKeyRegex` accepted any `\w+` as a namespace, so `cacheKey: 'user:123'` or a DDB-style `sortKey: 'a:b'` would have been collected as a translation key and reported missing — turning a green gate red on perfectly fine code. The namespace is now required to be one we actually ship (`NAMESPACES`).

Verifying "zero false positives" against today's source was not enough here: the hazard was in code nobody has written yet. Worth naming, because I gave that reassurance in round 1.

I also extracted the scan into `extractDataHeldKeys` — the added guard pushed `extractKeysFromSource` to cognitive complexity 16 against the ceiling of 15, and the two extractions are separate concerns anyway.

Re-proved the catch survives the constraint: restoring the chip still makes the gate print ``❌ common:nav.feedback``.

## Should fix — `useSummaryQuery` undefined-safety ✅ fixed

Fair. `apiEndpoint` is declared `string`, but the store rehydrates from persisted localStorage, so an older persisted shape can deliver `undefined` regardless of the type — and `.length` on that throws where the previous `!!config.apiEndpoint` was safe. Now typed `string | undefined` and tested for truthiness. This is the same "don't trust runtime data to match declared types" rule the repo applies at its query boundaries.

## Nit — the pytest assertions ✅ fixed, and it was not a nit

Tightening `<=` to `==` as suggested made the test **fail**, which exposed that it had been passing against **zero items**. The handler drops anything older than the window, and my fixture's hardcoded `date: '2026-01-07'` fell outside it — so `len(items) <= 3` was satisfied by `0 <= 3` and the test proved nothing about clamping.

The date is now computed relative to now, so the test actually exercises the behaviour it claims to pin: ten urgent items available, `limit=3`, `count == 3`. Thanks — that one was a vacuous test wearing a style complaint's clothing.

Docstring now also states the test should be **deleted** when `count` is fixed, and that the intended future fix *should* fail it.

## Nit — pin the fetch side of the cap test ✅ added

`toHaveBeenCalledWith(expect.objectContaining({ limit: 5 }))`, so the request limit and the render cap cannot drift apart in opposite directions and still satisfy the test.

## Two not changed

**The `sys.path.insert` claim is inverted.** It appears in **all 17** tests in `test_metrics_handler.py` — one per test def — so the new test matches the file's convention exactly rather than departing from it. Happy to hoist it to module scope for the whole file, but that is a separate cleanup and would leave the new test the odd one out if applied only there.

**`useSummaryQuery` takes the endpoint, not `isConfigured`, on purpose.** The hook owns both the query key and its enabling condition precisely so two callers cannot make them disagree — which is the defect this PR fixes. Passing a caller-computed boolean would hand half of that invariant back to the callers. Added a comment at the call site saying so.

## Still-open items

Both were dispositioned in the previous round and the positions are unchanged:

- **`breadcrumbs.feedback`** — the diff does remove it. `"breadcrumbs"` opens at `common.json:5`; the removed `"feedback": "Feedback"` sat at `:10`, inside that block. `nav.feedback` was already absent, and that absence is the bug.
- **Renaming `count`** — deliberately not done here. It changes a response shape with a live consumer, so it belongs in its own commit rather than riding along in a badge fix. It is now pinned by a test and commented at the return site.

## Gates

tsc 0 · eslint 0 · vitest **1806** · pytest **1043** · i18n `missing 0 / extra 0 / empty 40` (pre-existing, identical on `development`).

perrozzi:

Follow-up: the author constrained the new data-held-key extractor to the shipped `NAMESPACES` list (now in its own documented `extractDataHeldKeys`), made `useSummaryQuery`'s `apiEndpoint` optional with a truthiness check, tightened the backend `count` test (computed date, `== 3`) and added the fetch-limit assertion to the Dashboard preview test. All of the previous Should-fix items and most nits are addressed.

### Resolved
- `i18n-check.mjs`: namespace group is filtered against `NAMESPACES`, so `cacheKey: 'user:123'`-style strings can no longer produce bogus missing-key failures.
- `useSummaryQuery`: `apiEndpoint?: string` + `enabled: !!apiEndpoint` removes the `.length`-on-undefined throw.
- `test_count_is_the_returned_page_length_not_the_window_total`: hardcoded `2026-01-07` replaced with a computed recent date (would otherwise have passed vacuously against zero items), and `<= 3` tightened to `== 3`; docstring now says to delete the test when `count` is fixed.
- `Dashboard.test.tsx` preview test now pins `limit: 5` on the fetch as well as the render cap.

### Still open
- `breadcrumbs.feedback` removal is claimed in the description but absent from the diff.
- `get_urgent_feedback` still returns a misleading `count` (documented and pinned, deferred by design).

### New findings

**Nits**

- `voc-datalake/frontend/scripts/i18n-check.mjs`: the namespace filter is now silent, so a mistyped namespace in data (`commmon:nav.categories`) is dropped instead of reported — exactly the class of bug this check was added for. A `console.warn` on a `ns:key`-shaped literal whose namespace isn't in `NAMESPACES` would close that hole cheaply.
- `voc-datalake/lambda/api/test/test_metrics_handler.py`: the in-body `sys.path.insert` / `from metrics_handler import lambda_handler` is still out of step with the rest of the file (module-level import elsewhere).

---

perrozzi:
_(Answering review round 3 on `ba722de`. Addressed in `f81077a`, pushed right after this comment.)_

Nit-only round, so this is the last code change from my side. One nit was worth acting on; the other is inverted.

## Nit — the namespace filter was silent ✅ fixed

This one earned a fix rather than a disposition, because it is a hole in the guard I added two rounds ago: constraining the namespace list closed the false-positive risk but opened a false-*negative* one, and `commmon:nav.categories` being dropped silently is precisely the class of bug the extractor exists to catch.

Now warned rather than failed — the namespace list is the only thing separating a real key from an unrelated `foo:bar` string, so a hard failure would be wrong here. The guard is deliberately narrow: an alphabetic namespace **and** a dotted key path.

Verified with a throwaway probe file rather than by inspection:

```
⚠️  src/__typo_probe__.tsx: 'commmon:nav.categories' is shaped like a translation key
    but 'commmon' is not a known namespace — typo?
```

…while `cacheKey: 'user:123'` and `sortKey: 'a:b'` in the same file produced **zero** output. Probe deleted afterwards.

Worth noting the pattern across this review: round 1 asked for the check, round 2 found it could fail valid code, round 3 found it could miss invalid code. Each round found a hole in the previous round's fix, which is the argument for re-reviewing after addressing a review rather than assuming one pass settles it.

## Nit — `sys.path.insert` in the test body ❌ not changed, the claim is inverted

> still out of step with the rest of the file (module-level import elsewhere)

There is no module-level import elsewhere. In `test_metrics_handler.py`:

- `grep -c "^from metrics_handler\|^import metrics_handler"` → **0**
- `grep -c "from metrics_handler import"` → **17**
- `grep -c "    def test_"` → **17**

One in-body import per test, for all seventeen tests, and no module-level precedent anywhere in the file. The new test follows the only convention the file has.

Hoisting the import to module scope for the whole file would be a reasonable cleanup, but it is a 17-test refactor unrelated to this PR — and doing it for just the new test would make that test the sole outlier, which is the opposite of the request.

This is the second round this nit has appeared; flagging that as a re-derivation rather than a disagreement, since the review does not read comments.

## Still-open items — positions unchanged

- **`breadcrumbs.feedback`**: the diff removes it. `"breadcrumbs"` opens at `common.json:5`, the removed `"feedback": "Feedback"` sat at `:10` inside that block. `nav.feedback` was already absent, and that absence is the bug being fixed.
- **Renaming `count`**: deferred by design, now pinned by a test and commented at the return site. It changes a response shape with a live consumer, so it belongs in its own commit.

## Convergence

Findings per round: **1 blocker + 4 should-fix + 4 nits → 2 should-fix + 3 nits → 0 should-fix + 2 nits**. Review length 4.9k → 3.6k → 2.4k chars. Stopping the loop here per the convention on nit-only rounds.

Gates on `f81077a`: tsc 0 · eslint 0 · vitest **1806** · pytest **1043** · i18n `missing 0 / extra 0 / empty 40` (pre-existing, identical on `development`).

Two things a maintainer should weigh before merging, both already in the PR body: this moves the badge onto `/metrics/summary`, whose per-day fan-out is heavier than the call it replaced (the handler-side fix is tracked separately), and this repo has no CI, so every figure above is a local run.

perrozzi:

Only one file changed since the last review: `i18n-check.mjs` now threads `file` into `extractDataHeldKeys` and warns on strings shaped like a translation key whose namespace isn't in `NAMESPACES`. That closes the last substantive nit; nothing else moved.

### Resolved
- `scripts/i18n-check.mjs`: mistyped namespaces (`commmon:nav.categories`) are now surfaced via `console.warn` instead of being silently dropped, with a narrow guard (alphabetic ns + dotted key) that keeps `cacheKey: 'user:123'` quiet.

### Still open
- `breadcrumbs.feedback` removal is claimed in the description but absent from the diff.
- `voc-datalake/lambda/api/test/test_metrics_handler.py`: in-body `sys.path.insert` / `from metrics_handler import lambda_handler` still diverges from the module-level import used elsewhere in the file.
- `get_urgent_feedback` still returns a misleading `count` (documented and pinned, deferred by design).

### New findings

**Nits**

- `voc-datalake/frontend/scripts/i18n-check.mjs`: the JSDoc block above `extractDataHeldKeys` documents only `@param content`; add `@param file` now that it's part of the signature and used in the message.
- Same file: the warning fires once per occurrence, so a repeated literal (e.g. a shared `storageKey: 'app:settings.v2'` pattern) will print duplicates; de-duping via a `Set` of `ns:key` would keep the output readable. Non-blocking, since the gate's exit code is unaffected.

---

perrozzi:
_(Answering review round 4 on `f81077a`. The JSDoc fix is `4f87ffe`, pushed right after this comment. **Closing the review loop here** — see below.)_

## Nit — missing `@param file` ✅ fixed

Correct and worth the one line: I added the parameter without documenting it, so the JSDoc contradicted the signature.

## Nit — duplicate warnings for a repeated literal ❌ declined, on the merits

De-duping via a `Set` of `ns:key` would suppress the same suspected typo when it appears in a *second file* — and the file path is the entire actionable content of that warning. Losing "which files to fix" to make the output shorter is the wrong trade for a diagnostic. The review notes it doesn't affect the exit code, and per-occurrence output is what makes it fixable.

## `sys.path.insert` — third re-derivation, position unchanged

Refuted twice with counts; recording once more for the thread, briefly. In `test_metrics_handler.py`: `0` module-level `metrics_handler` imports, `17` in-body imports, `17` tests. There is no module-level precedent in that file for the new test to diverge *from*. Hoisting all seventeen is a reasonable separate cleanup; doing it for one test would make that test the sole outlier.

## Closing the loop

| Round | Findings | Length |
|---|---|---|
| 1 | 1 blocker · 4 should-fix · 4 nits | 4.9k |
| 2 | 2 should-fix · 3 nits | 3.6k |
| 3 | 0 should-fix · 2 nits | 2.4k |
| 4 | 0 should-fix · 2 nits | 1.9k |

Two consecutive rounds with no should-fix, shrinking each time, and the remaining open items are three standing dispositions rather than new findings. Stopping here per the convention on nit-only rounds; further pushes would trade real churn for diminishing returns.

The most useful thing this review produced was not any single fix but a pattern: **round 1 asked for the i18n check, round 2 found it could fail valid code, round 3 found it could miss invalid code.** Each round found a hole in the previous round's fix. Two of the findings also caught tests that were passing vacuously — one against zero items, one relying on shared fixtures happening to diverge. Neither would have been visible without the re-review.

## Final state

`4f87ffe` · 6 commits · gates: tsc 0 · eslint 0 · vitest **1806** · pytest **1043** · i18n `missing 0 / extra 0 / empty 40` (pre-existing, identical in a clean worktree of `development`).

Two caveats for whoever merges, both in the PR body: the badge now reads `/metrics/summary`, whose per-day fan-out is heavier than the call it replaced — the handler-side fix is tracked separately and deliberately out of scope here — and this repo has no CI, so every figure above is a local run and the live behaviour was verified against the mock server, not a deployment.

perrozzi:

Only one thing changed since the last review: the JSDoc for `extractDataHeldKeys` in `scripts/i18n-check.mjs` now documents the `file` parameter. That closes the first of the two nits I raised; nothing else moved.

### Resolved
- `voc-datalake/frontend/scripts/i18n-check.mjs`: `@param file` is now documented alongside `@param content`.

### Still open
- `scripts/i18n-check.mjs`: warning fires per occurrence, so repeated literals print duplicates (nit, exit code unaffected).
- `breadcrumbs.feedback` removal claimed in the description but absent from the diff.
- `voc-datalake/lambda/api/test/test_metrics_handler.py`: in-body `sys.path.insert` / `from metrics_handler import lambda_handler` still diverges from the file's module-level import style.
- `get_urgent_feedback` still returns a misleading `count` (documented and pinned, deferred by design).

No new findings — the incremental change is a doc-only edit and is clean.

---

perrozzi:
## Deployment validation + measured cost of the badge switch

Deployed this branch's frontend to a personal AWS account (not the workshop one) and measured the endpoint swap on a realistic dataset: **6,234 feedback items**. Posting the numbers because they both justify U1 more strongly than the original description did and quantify the cost the `COST NOTE` in `Layout.tsx` discloses.

### Why the old badge was wrong, not just inconsistent

The original description framed U1 as a cache-key collision (badge `10` elsewhere vs `5` on Dashboard). The collision is real, but it was masking a second, larger bug: **the old source could not express the true number at all.**

`/feedback/urgent` returns `count: len(items)` *after* the page slice, and `limit` is clamped by `validate_limit(max_val=100)`:

| request | returned `count` |
|---|---|
| `/feedback/urgent?days=90&limit=10` | `10` |
| `/feedback/urgent?days=90&limit=200` | `100` ← clamped, not the total |
| `/metrics/summary?days=90` → `urgent_count` | **`306`** |

So at 90d the badge read **10** where the truth is **306**. Raising the limit could not have fixed it; 100 is the ceiling. That's why the fix had to change endpoints rather than just add `limit` to the query key.

### What it costs

`urgent_count` comes from the aggregates table, and `/metrics/summary` does ~3 sequential `get_item` calls per day in the window:

| window | `/metrics/summary` |
|---|---|
| 7d | 565 ms |
| 30d | 1011 ms |
| 90d | 2609 ms |

Head-to-head at 90d: `/feedback/urgent?limit=10` **1297 ms** → `/metrics/summary` **2754 ms**, so roughly **2.1× slower** on the sidebar's blocking path.

Two things make that acceptable, and they're the reason I didn't try to optimise it here:

1. The badge now **shares Dashboard's query key**, so on the highest-traffic route the request is deduped — it's one fetch for both consumers, where before it was two different fetches with colliding keys.
2. Everything inherits the global `staleTime: 30000`, so it's not per-navigation.

The per-day fan-out in `/metrics/summary` is a pre-existing backend cost that now sits on one more path. It's worth its own issue rather than a drive-by in a UI fix.

### Live verification

Frontend-only deploy (the backend change in this PR is a comment plus a test, no runtime effect). Bundle `index-CmkIrxrP.js` → `index-BWkKLpG_.js`.

- Badge reads **306** on `/`, `/dashboard`, `/categories`, `/settings`, `/projects` — one number everywhere, no per-route divergence.
- Dashboard heading reads `Urgent Issues (306)`, agreeing with the badge, while the preview list stays capped at 5.
- At the default 7d the badge reads **1**, matching measured `urgent_count=1`.
- U2: no `nav.feedback` literal and no raw i18n keys render on Home; the dead chip is gone.
- `breadcrumbs.feedback` is absent from the served `common.json`, and the 8 orphaned `feedback.json` namespace files are gone from the origin.
- 0 console errors.

One verification note worth recording: on this CloudFront distribution a deleted file returns **200 with `text/html`** (the SPA fallback), not 404. Confirming the namespace deletion required checking `content_type`, not the status code.

PR #279, opened 2026-08-07
perf(api): read metric windows in one query instead of one get_item per day

perrozzi:
## What

Removes every per-day sequential `get_item` walk over the **aggregates** table in `metrics_handler.py`, replacing each with a single windowed query.

| Endpoint | Before | After |
|---|---|---|
| `/metrics/summary` | 3 × `days` (~**270** at 90d) | **3** queries |
| `/metrics/sentiment` | 4 × `days` (~360 at 90d) | **4** queries |
| `/metrics/categories` | N × `days` | **N** queries |
| `/feedback/entities` | N × `days` + `days` | **N + 1** queries |

Round-trip count is now independent of the requested window. Measured baseline for `/metrics/summary` on **6,234 feedback items**: **7d 565 ms · 30d 1011 ms · 90d 2609 ms** — linear in `days`, which is the fan-out showing through.

## Why now

#278 moved the sidebar urgent badge onto `/metrics/summary` (`urgent_count`), because the endpoint it read before reported a clamped page length and could never express the true count — at 90d it showed `10` against a real `306`, and raising `limit` could not fix it (`validate_limit(max_val=100)`). That fix was right, and it put a caller on the heaviest fan-out in the file.

To be precise about the impact, since it's easy to overstate: `Layout` is the parent route element so it never unmounts, and the query sets no `refetchInterval` — navigation triggers **zero** refetches, and the client cost is one fetch per session per window under TanStack's `staleTime: 30000`. The real problem is that **there is no server-side cache at any layer** — no API Gateway cache cluster, no memoization in the handler, aggregates read fresh every call. So the client cache protects one tab and amortizes nothing across users: N concurrent users cost N× the fan-out. That's a workshop's exact load profile, and it's only fixable server-side.

## How

The aggregates table is keyed `(pk, sk)` with `sk` = `'YYYY-MM-DD'`. ISO dates sort lexicographically, so a trailing window is one **contiguous sort-key range** and `between()` expresses it server-side:

```python
KeyConditionExpression=Key('pk').eq(pk) & Key('sk').between(oldest, newest),
ScanIndexForward=False,
```

**This is deliberately not the `gsi1-by-metric-type` shape** that `/metrics/sources` and `/metrics/personas` use, and that difference is the main thing worth reviewing. I had assumed it would be a straight substitution of the existing pattern; it isn't. That GSI only carries items the aggregator tags with `metric_type`, and `get_metric_type()` tags **only** `METRIC#daily_source#` and `METRIC#persona#` ([`aggregator/handler.py:26-31`](../blob/development/voc-datalake/lambda/aggregator/handler.py)) — precisely the two endpoints that use it. The five partitions here carry no such attribute.

So the fan-out was never an oversight; it was a consequence of the index not covering these items. Extending the GSI would have meant an aggregator change plus a backfill, with existing items silently missing from results until backfilled. Querying the base table by known `pk` needs neither, and is tighter besides: it filters the window server-side, where the GSI shape reads every date ever recorded and discards most of it in memory.

### `ScanIndexForward=False` is load-bearing

The replaced loops counted `i` up from `0` (today), so `daily_totals` and `daily_sentiment` reach the client **newest-first** and are charted in that order. A default ascending query would have reversed both series while leaving every total correct — a silent regression with no failing assertion anywhere. There's a test naming exactly that.

## Verification

- **pytest 1043 → 1050.** `test:cdk` 155 unchanged. No new ruff findings in any of the three changed files (`metrics_handler.py` 4 → 4, `test_metrics_handler.py` 17 → 17, `..._date_basis.py` 1 → 1; the repo-wide ruff baseline is pre-existing and untouched).
- **Proved non-vacuous:** reverting the handler while keeping the tests fails all 7 new tests plus the 3 updated ones. Nothing here passes by accident.
- `urgent_count` is pinned to the exact window sum, because the sidebar badge and the Dashboard `Urgent Issues (N)` heading both read it and #278 exists to keep them equal.
- Window equivalence is pinned by comparing whole boto3 `Condition` objects, so the `pk`, both bounds and the operators are asserted at once — including the `days=1` edge, which must read exactly one date rather than a zero-width or off-by-one range.

## Out of scope, deliberately

Two per-day loops remain, both over the **feedback** table via the date GSI (`:150` `_scan_recent_items`, `:260` `/feedback`). They're a different class: they carry a `soft_cap`, per-partition paging and an `is_partial` signal, so collapsing them means a real design decision about source push-down and how partiality is preserved. That belongs in its own change. `:464` is capped at `min(days, 7)` and is bounded by construction, not this bug.

## Incidental

- Fixed a latent test-isolation bug my stricter assertions exposed: `get_configured_categories` memoizes in a module-level cache that other test modules populate, so the category test passed alone and failed in the full suite. It now calls `clear_categories_cache()`, matching the precedent in `shared/test/test_api.py`.
- Two existing assertions were updated rather than deleted. `test_imported_basis_still_reads_aggregates` asserted `get_item.called`, but its intent is *which table* is read, not the call shape, so it now asserts `query.called`. Its sibling claiming the aggregates table is "bypassed entirely" under review basis now asserts **both** access shapes are absent, which is what "entirely" requires.
- Three previously near-vacuous assertions were tightened while I was in there (`'total_feedback' in body` → exact values), since they could not have caught an equivalence regression.
- New tests skip the per-test `sys.path.insert` boilerplate the older ones carry — `conftest.py:13-16` already puts both directories on `sys.path`, so it's redundant. I left the 17 existing instances alone rather than churn this diff.

### DIFF
--- a/voc-datalake/lambda/api/metrics_handler.py
+++ b/voc-datalake/lambda/api/metrics_handler.py
@@ -421,13 +421,11 @@ def get_entities():
     categories_list = get_configured_categories(aggregates_table)
     category_counts = {}
     for category in categories_list:
-        total = 0
-        for i in range(days):
-            date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-            response = aggregates_table.get_item(Key={'pk': f'METRIC#daily_category#{category}', 'sk': date})
-            item = response.get('Item')
-            if item:
-                total += int(item.get('count', 0))
+        total = sum(
+            int(item.get('count', 0))
+            for item in _query_metric_window(
+                f'METRIC#daily_category#{category}', days, current_date)
+        )
         if total > 0:
             category_counts[category] = total
     
@@ -455,13 +453,10 @@ def get_entities():
             persona_counts[persona_name] = persona_counts.get(persona_name, 0) + int(item.get('count', 0))
     
     # Get feedback count
-    feedback_count = 0
-    for i in range(days):
-        date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-        response = aggregates_table.get_item(Key={'pk': 'METRIC#daily_total', 'sk': date})
-        item = response.get('Item')
-        if item:
-            feedback_count += int(item.get('count', 0))
+    feedback_count = sum(
+        int(item.get('count', 0))
+        for item in _query_metric_window('METRIC#daily_total', days, current_date)
+    )
     
     # Extract issues from recent feedback
     issues = {}
@@ -676,6 +671,42 @@ def _summary_from_items(days: int) -> dict:
     }
 
 
+def _query_metric_window(pk: str, days: int, current_date: datetime) -> list[dict]:
+    """Read one metric partition's trailing `days` window in a SINGLE query.
+
+    The aggregates table is keyed `(pk, sk)` with `sk` = 'YYYY-MM-DD'. ISO dates
+    sort lexicographically, so a trailing window is a contiguous sort-key range
+    and `between()` expresses it server-side: **one** request regardless of
+    `days`, replacing the one `get_item` per day this helper was extracted to
+    remove (`/metrics/summary` alone issued 3 x days, ~270 at a 90-day window).
+
+    Items come back NEWEST FIRST (`ScanIndexForward=False`) because the loops
+    this replaces walked `i` from 0 (today) upward, and `daily_totals` /
+    `daily_sentiment` are handed to the client in that order to be charted.
+    A plain query would return them ascending and silently reverse the charts.
+
+    Deliberately NOT the `gsi1-by-metric-type` shape used by `/metrics/sources`
+    and `/metrics/personas`. That index only carries items the aggregator tags
+    with `metric_type`, and it tags *only* `METRIC#daily_source#` and
+    `METRIC#persona#` (`aggregator/handler.py:26-31`) — precisely the two
+    endpoints that use it. The partitions here carry no such attribute and need
+    none: their `pk` is known, so the base table's own key schema answers the
+    question directly, and the window is filtered server-side rather than
+    reading every date ever recorded and discarding most of it in memory.
+
+    One page suffices: `validate_days` caps `days` at 365 and these are
+    single-counter items (~150 bytes), so a full window is ~55 KB against
+    DynamoDB's 1 MB page limit.
+    """
+    oldest = (current_date - timedelta(days=days - 1)).strftime('%Y-%m-%d')
+    newest = current_date.strftime('%Y-%m-%d')
+    response = aggregates_table.query(
+        KeyConditionExpression=Key('pk').eq(pk) & Key('sk').between(oldest, newest),
+        ScanIndexForward=False,
+    )
+    return response.get('Items', [])
+
+
 @app.get("/metrics/summary")
 @tracer.capture_method
 def get_summary():
@@ -689,30 +720,21 @@ def get_summary():
     
     current_date = datetime.now(timezone.utc)
     
-    totals = []
-    for i in range(days):
-        date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-        response = aggregates_table.get_item(Key={'pk': 'METRIC#daily_total', 'sk': date})
-        item = response.get('Item')
-        if item:
-            totals.append({'date': date, 'count': item.get('count', 0)})
+    totals = [
+        {'date': item['sk'], 'count': item.get('count', 0)}
+        for item in _query_metric_window('METRIC#daily_total', days, current_date)
+    ]
     
     sentiment_data = []
-    for i in range(days):
-        date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-        response = aggregates_table.get_item(Key={'pk': 'METRIC#daily_sentiment_avg', 'sk': date})
-        item = response.get('Item')
-        if item and item.get('count', 0) > 0:
+    for item in _query_metric_window('METRIC#daily_sentiment_avg', days, current_date):
+        if item.get('count', 0) > 0:
             avg = float(item.get('sum', 0)) / float(item.get('count', 1))
-            sentiment_data.append({'date': date, 'avg_sentiment': round(avg, 3), 'count': item.get('count')})
+            sentiment_data.append({'date': item['sk'], 'avg_sentiment': round(avg, 3), 'count': item.get('count')})
     
-    urgent_count = 0
-    for i in range(days):
-        date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-        response = aggregates_table.get_item(Key={'pk': 'METRIC#urgent', 'sk': date})
-        item = response.get('Item')
-        if item:
-            urgent_count += item.get('count', 0)
+    urgent_count = sum(
+        item.get('count', 0)
+        for item in _query_metric_window('METRIC#urgent', days, current_date)
+    )
     
     total_feedback = sum(int(t.get('count', 0)) for t in totals)
     avg_sentiment = sum(float(s.get('avg_sentiment', 0)) * int(s.get('count', 0)) for s in sentiment_data) / max(total_feedback, 1)
@@ -750,14 +772,11 @@ def get_sentiment_metrics():
                 result[sentiment] += 1
     else:
         for sentiment in sentiments:
-            total = 0
-            for i in range(days):
-                date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-                response = aggregates_table.get_item(Key={'pk': f'METRIC#daily_sentiment#{sentiment}', 'sk': date})
-                item = response.get('Item')
-                if item:
-                    total += int(item.get('count', 0))
-            result[sentiment] = total
+            result[sentiment] = sum(
+                int(item.get('count', 0))
+                for item in _query_metric_window(
+                    f'METRIC#daily_sentiment#{sentiment}', days, current_date)
+            )
     
     total = sum(result.values())
     return {
@@ -794,13 +813,11 @@ def get_category_metrics():
             result[category] = result.get(category, 0) + 1
     else:
         for category in categories:
-            total = 0
-            for i in range(days):
-                date = (current_date - timedelta(days=i)).strftime('%Y-%m-%d')
-                response = aggregates_table.get_item(Key={'pk': f'METRIC#daily_category#{category}', 'sk': date})
-                item = response.get('Item')
-                if item:
-                    total += item.get('count', 0)
+            total = sum(
+                int(item.get('count', 0))
+                for item in _query_metric_window(
+                    f'METRIC#daily_category#{category}', days, current_date)
+            )
             if total > 0:
                 result[category] = total
     
--- a/voc-datalake/lambda/api/metrics_handler.py
+++ b/voc-datalake/lambda/api/metrics_handler.py
@@ -225,7 +225,7 @@ def _query_metric_window(pk: str, days: int, current_date: datetime) -> list[dic
     # it rather than return a quietly short answer that reads as authoritative.
     logger.warning(
         'Metric window paging hit its bound; returning a partial window',
-        extra={'pk': pk, 'days': days, 'pages': days, 'items': len(items)},
+        extra={'pk': pk, 'days': days, 'items': len(items)},
     )
     return items
 
--- a/voc-datalake/lambda/api/metrics_handler.py
+++ b/voc-datalake/lambda/api/metrics_handler.py
@@ -187,6 +187,45 @@ def _scan_window_items(
 # Feedback Endpoints
 # ============================================
 
+def _query_metric_window(pk: str, days: int, current_date: datetime) -> list[dict]:
+    """Read one metric partition's trailing `days` window, newest date first.
+
+    `sk` is 'YYYY-MM-DD' and ISO dates sort lexicographically, so a window is a
+    contiguous sort-key range that `between()` bounds server-side: a fixed
+    number of requests regardless of `days`, not one `get_item` per day.
+
+    `ScanIndexForward=False` is load-bearing. Callers hand these items straight
+    to the client as `daily_totals` / `daily_sentiment`, which are charted
+    newest-first; DynamoDB's default ascending order would reverse both series
+    while leaving every total correct.
+
+    Not the `gsi1-by-metric-type` index: it only holds items the aggregator tags
+    with `metric_type`, which is just the daily_source and persona partitions.
+    These `pk`s are known anyway, so the base table answers directly and bounds
+    the window server-side instead of reading all dates and filtering in memory.
+    """
+    oldest = (current_date - timedelta(days=days - 1)).strftime('%Y-%m-%d')
+    newest = current_date.strftime('%Y-%m-%d')
+    condition = Key('pk').eq(pk) & Key('sk').between(oldest, newest)
+    items: list[dict] = []
+    kwargs: dict = {'KeyConditionExpression': condition, 'ScanIndexForward': False}
+    # A 365-day window of counter items sits far inside one 1 MB page today, but
+    # that rests on item width the aggregator controls, and these endpoints have
+    # no is_partial signal with which to report a truncated window. So follow the
+    # cursor -- but bounded, never `while True`: one date yields at most one item
+    # and an unfiltered page yields at least one, so a window of `days` dates
+    # cannot span more than `days` pages. A bound also means a surprising
+    # response shape degrades to a short read instead of spinning.
+    for _ in range(days):
+        response = aggregates_table.query(**kwargs)
+        items.extend(response.get('Items', []))
+        last_key = response.get('LastEvaluatedKey')
+        if not last_key:
+            break
+        kwargs['ExclusiveStartKey'] = last_key
+    return items
+
+
 @app.get("/feedback")
 @tracer.capture_method
 def list_feedback():
@@ -669,42 +708,6 @@ def _summary_from_items(days: int) -> dict:
         'daily_totals': totals,
         'daily_sentiment': sentiment_data,
     }
-
-
-def _query_metric_window(pk: str, days: int, current_date: datetime) -> list[dict]:
-    """Read one metric partition's trailing `days` window in a SINGLE query.
-
-    The aggregates table is keyed `(pk, sk)` with `sk` = 'YYYY-MM-DD'. ISO dates
-    sort lexicographically, so a trailing window is a contiguous sort-key range
-    and `between()` expresses it server-side: **one** request regardless of
-    `days`, replacing the one `get_item` per day this helper was extracted to
-    remove (`/metrics/summary` alone issued 3 x days, ~270 at a 90-day window).
-
-    Items come back NEWEST FIRST (`ScanIndexForward=False`) because the loops
-    this replaces walked `i` from 0 (today) upward, and `daily_totals` /
-    `daily_sentiment` are handed to the client in that order to be charted.
-    A plain query would return them ascending and silently reverse the charts.
-
-    Deliberately NOT the `gsi1-by-metric-type` shape used by `/metrics/sources`
-    and `/metrics/personas`. That index only carries items the aggregator tags
-    with `metric_type`, and it tags *only* `METRIC#daily_source#` and
-    `METRIC#persona#` (`aggregator/handler.py:26-31`) — precisely the two
-    endpoints that use it. The partitions here carry no such attribute and need
-    none: their `pk` is known, so the base table's own key schema answers the
-    question directly, and the window is filtered server-side rather than
-    reading every date ever recorded and discarding most of it in memory.
-
-    One page suffices: `validate_days` caps `days` at 365 and these are
-    single-counter items (~150 bytes), so a full window is ~55 KB against
-    DynamoDB's 1 MB page limit.
-    """
-    oldest = (current_date - timedelta(days=days - 1)).strftime('%Y-%m-%d')
-    newest = current_date.strftime('%Y-%m-%d')
-    response = aggregates_table.query(
-        KeyConditionExpression=Key('pk').eq(pk) & Key('sk').between(oldest, newest),
-        ScanIndexForward=False,
-    )
-    return response.get('Items', [])
 
 
 @app.get("/metrics/summary")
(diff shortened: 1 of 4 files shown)


perrozzi:

This PR replaces per-day `get_item` walks over the aggregates table in `metrics_handler.py` with a single `Key('pk').eq(...) & Key('sk').between(...)` query per metric partition, extracted into `_query_metric_window`, and adds tests pinning the query shape, `ScanIndexForward=False`, and window-count independence from `days`. The core reasoning is sound and the window math is right: the old loop visited `current_date - 0 .. current_date - (days-1)`, which is exactly `between(current-(days-1), current)`, and `days=1` collapses to a single date. Choosing the base table over `gsi1-by-metric-type` is well justified. Two things need confirming before merge (IAM and paging), and a couple of the new tests assert less than they appear to.

### Blockers

- **Confirm `dynamodb:Query` is granted on the aggregates *table* ARN for this Lambda's role.** This change introduces a base-table `Query` where previously only `GetItem` (plus GSI queries on `gsi1-by-metric-type`) was used on the aggregates table. There is no CDK diff here and `test:cdk` is unchanged, so if the aggregates grant is a hand-rolled least-privilege policy (actions/resources enumerated) rather than `grantReadData`, `/metrics/summary`, `/metrics/sentiment`, `/metrics/categories` and `/feedback/entities` will all fail with `AccessDenied` at runtime — and every test in this PR mocks `aggregates_table`, so nothing catches it. Please paste the effective policy statement for the aggregates table, or add the `Query` action + a CDK assertion.

### Should fix

- **`_query_metric_window` ignores `LastEvaluatedKey`** (`voc-datalake/lambda/api/metrics_handler.py`). The docstring argues one page always suffices (~150 B × 365 ≈ 55 KB), which is true *today*, but that's an invariant about aggregate item width that nothing enforces — if the aggregator ever adds attributes to these counter items, the endpoints silently return partial windows with no error and no `is_partial` signal (contrast the deliberately-preserved partiality handling in the feedback-table loops you left out of scope). A `while 'LastEvaluatedKey' in response` loop is ~4 lines and removes the assumption entirely.

- **The sentiment and category tests no longer assert the partition-key prefix.** Both now patch `metrics_handler._query_metric_window` with `side_effect` stubs that derive the label via `pk.rsplit('#', 1)[-1]` (`test_metrics_handler.py`, `test_returns_sentiment_breakdown` / `test_returns_category_breakdown`). A typo'd prefix — `METRIC#sentiment#positive` instead of `METRIC#daily_sentiment#positive` — passes both tests, and `TestQueryMetricWindow` only covers the helper in isolation with a literal pk, so the mapping from endpoint to partition is now unasserted anywhere. Add an assertion on the full pks, e.g. `[c.args[0] for c in mock_window.call_args_list] == ['METRIC#daily_sentiment#positive', ...]`.

- **`clear_categories_cache()` is called without restoring state.** In `test_returns_category_breakdown` the cache is cleared at test entry and then repopulated with `{'product', 'delivery'}` for whatever runs next. That fixes the inbound leak you diagnosed but leaves the outbound one, which is the same class of bug. An autouse fixture (or `try/finally`) that clears before *and* after is the durable form.

- **Verify the imports the new code needs are already present.** `_query_metric_window` uses `Key` from `boto3.dynamodb.conditions`, and the new `TestQueryMetricWindow` / `TestMetricsRequestCountIsIndependentOfWindow` classes use `datetime` and `timezone`; the diff adds no imports to either file. Presumably both are already module-level, but a `NameError` here would be a hard 500 on all four endpoints.

### Nits

- The 30-line docstring on `_query_metric_window` is mostly PR narrative (measured latencies, why-not-the-GSI, `aggregator/handler.py:26-31`). The line-number reference will rot; the design rationale belongs in the PR/commit message. Three or four lines covering the sort-key range, `ScanIndexForward=False`, and the single-page assumption would carry the same operational value.
- `assert mock_agg_table.get_item.call_count == 0` in `test_summary_issues_three_queries_whatever_the_window` is a good fan-out guard but will fire spuriously if any legitimate single-shot settings read (e.g. date basis) later moves onto `aggregates_table.get_item`. Consider scoping it to the metric pks rather than all `get_item` calls.
- Leftover `mock_agg_table.get_item.return_value = {}` in `test_returns_zero_totals_when_no_data` is now dead alongside the new `query` stub; drop it.
- `_query_metric_window` is defined ~250 lines below its first caller in `get_entities`. Legal, but moving it next to the other module-level helpers would help readers.
- Callers repeat `int(item.get('count', 0))` at four sites; the helper could normalize `count` once, which would also make `urgent_count` consistently an `int` rather than a `Decimal` sum.

---

perrozzi:
Round 1 dispositioned. **1 blocker refuted with evidence (but guarded anyway), 3 of 4 should-fix accepted, 4 of 5 nits accepted, 1 declined.** One finding led me to introduce a real bug, described below.

## Blocker — `dynamodb:Query` on the aggregates table: REFUTED

Your conditional was "*if* the aggregates grant is a hand-rolled least-privilege policy". It isn't — `api-stack.ts:170` is `aggregatesTable.grantReadWriteData(metricsRole)`. The effective statement from the synthesized `cdk.out/VocApiStack.template.json`:

```
ACTIONS:   ["dynamodb:BatchGetItem","dynamodb:BatchWriteItem","dynamodb:ConditionCheckItem",
            "dynamodb:DeleteItem","dynamodb:DescribeTable","dynamodb:GetItem","dynamodb:GetRecords",
            "dynamodb:GetShardIterator","dynamodb:PutItem","dynamodb:Query","dynamodb:Scan",
            "dynamodb:UpdateItem"]
RESOURCES: [AggregatesTable.Arn, AggregatesTable.Arn + "/index/*"]
```

`dynamodb:Query` on the **base table ARN**, not only `/index/*`. No AccessDenied, and no IAM change needed.

**Your alternative was the better ask, though, so I did it too.** The failure mode you describe is real in principle and structurally invisible to this PR's tests, since they all mock `aggregates_table`. New `api-stack.test.ts` test pins it, and it checks the bare table ARN specifically rather than accepting an index-only grant. Verified non-vacuous: narrowing line 170 to `grant(metricsRole, "dynamodb:GetItem")` fails exactly that test and nothing else. `test:cdk` 155 → 156.

## Should fix

**1. `LastEvaluatedKey` — accepted, and it caught a latent assumption properly.** You're right that "~150 B × 365" is an invariant about item width that nothing enforces, and that these endpoints have no `is_partial` with which to report a short window. Now paged.

⚠️ **Worth recording precisely: implementing this introduced a hang, and your suggested form would not have.** I wrote `last_key = response.get('LastEvaluatedKey')` inside `while True`. Under an unstubbed `MagicMock`, `.get()` returns a **truthy MagicMock**, so the loop never terminated — the full pytest run blew a 900 s timeout. Your `while 'LastEvaluatedKey' in response` is safe there, because `MagicMock.__contains__` returns `False`. Measured both:

```
MagicMock.get('LastEvaluatedKey') truthy? -> True    (my form: spins)
'LastEvaluatedKey' in MagicMock          -> False    (your form: terminates)
```

I settled on a **bounded** loop rather than either, since a cursor-driven `while True` is a hang risk in a Lambda regardless of mocks: `for _ in range(days)`. That bound is exact, not arbitrary — one date yields at most one item and an unfiltered page yields at least one, so a window of `days` dates cannot span more than `days` pages. A surprising response shape now degrades to a short read instead of spinning to the timeout. Paging has its own test covering concatenation order, cursor hand-off, and that the first call carries no `ExclusiveStartKey`.

**2. Partition-key prefix unasserted — accepted, correct and the most valuable finding here.** Confirmed: `rsplit('#', 1)[-1]` yields `positive` for both `METRIC#daily_sentiment#positive` and a typo'd `METRIC#sentiment#positive`, so my stub could not tell them apart, and `TestQueryMetricWindow` uses a literal pk. The endpoint-to-partition mapping was genuinely unasserted anywhere. Both tests now assert the full pk list via `[c.args[0] for c in mock_window.call_args_list]`. This is the same vacuity class as tightening `<=` to `==`, and I'd missed it.

**3. Outbound cache leak — accepted.** Fair, and the same bug class in the other direction. `clear_categories_cache()` now runs in a `finally`, so the test no longer hands `{product, delivery}` to whatever follows.

**4. Imports — already satisfied; posting the evidence rather than a change.** `Key` is at `metrics_handler.py:12` (`from boto3.dynamodb.conditions import Attr, Key`) and `datetime`/`timezone` at `test_metrics_handler.py:4`, both pre-existing module-level, which is why the diff adds neither. A `NameError` would also have failed all 1051 tests rather than hiding.

## Nits

- **Docstring trimmed, and the line reference removed** — agreed, and the `aggregator/handler.py:26-31` citation would indeed have rotted (line numbers in this very file shifted +9 when #278 landed). Down from ~30 lines to the operationally load-bearing parts: the sort-key range, why `ScanIndexForward=False` matters, the paging bound, and a *non-numbered* note on why not the GSI. That last one stays because it stops a future reader "optimizing" onto an index that doesn't contain these items.
- **Fan-out guard scoped** — now filters `get_item` calls to `METRIC#` pks, so a later `SETTINGS#` read on this table won't fail it for an unrelated reason.
- **Dead `get_item.return_value` dropped.**
- **Helper moved** to sit with `_query_partition` / `_scan_recent_items` / `_scan_window_items` at the top, so it now precedes its first caller instead of trailing it by ~300 lines.
- **Declined: normalizing `count` inside the helper.** It would mean mutating items the helper returns, and `count` isn't the only numeric attribute on them — `daily_sentiment_avg` items also carry `sum`, which callers read as a `float`. Normalizing one and not the other is a arbitrary line, and a read helper quietly reshaping its rows is a surprise I'd rather not add for four `int()` calls. I did make `urgent_count` an explicit `int` on its own, since that one is a published contract shared with the Dashboard heading and shouldn't be a `Decimal` by accident.

## Gates on the new head

pytest **1051** (was 1050 here, 1043 on `development`), `test:cdk` **156** (was 155), `typecheck:cdk` clean, and ruff unchanged from base on every touched file (`metrics_handler.py` 4→4, `test_metrics_handler.py` 17→17, `..._date_basis.py` 1→1).

perrozzi:

Since the last review the helper gained bounded `LastEvaluatedKey` following and a much shorter docstring, the sentiment/category tests now pin full partition keys, `clear_categories_cache()` is restored in a `finally`, the dead `get_item` stub is gone, the fan-out guard is scoped to `METRIC#` pks, and a new CDK assertion pins `dynamodb:Query` on the aggregates table ARN. All four previous should-fixes and the IAM blocker are addressed.

### Resolved
- IAM blocker: `api-stack.test.ts` now asserts `dynamodb:Query` on the bare aggregates table ARN (not just `/index/*`).
- `_query_metric_window` follows the cursor, with a paging test asserting concatenation order and `ExclusiveStartKey` propagation.
- Full pks asserted for both the sentiment and category endpoint→partition mappings.
- `clear_categories_cache()` now runs in a `finally`, closing the outbound leak.
- Docstring trimmed to design-relevant lines; the `aggregator/handler.py:26-31` line reference is gone.
- Dead `mock_agg_table.get_item.return_value = {}` removed; fan-out guard scoped to `METRIC#` keys.

### New findings
- **Confirm `z` (zod) is already imported in `voc-datalake/lib/stacks/api-stack.test.ts`.** The new `describe` block uses `z.object`/`z.union` but the diff adds no import; if it isn't already module-level, the whole test file fails to compile and `test:cdk` breaks. Same class as the Python import question I raised last time — please confirm both.
- **The `range(days)` bound turns "cannot truncate" back into a silent short read at the boundary.** `voc-datalake/lambda/api/metrics_handler.py` — if the loop exhausts with a `LastEvaluatedKey` still present, the window is quietly partial, which is exactly the failure mode the paging loop was added to remove. One `logger.warning(...)` on that exit path (powertools logger is already in the module) makes it observable rather than invisible; the bound itself is a reasonable safety valve.
- Nit: the CDK assertion matches resources by `JSON.stringify(...).includes('Aggregates')`, so it silently degrades to a pass-by-absence... actually to a failure if the table's logical ID is ever renamed away from containing `Aggregates`. That's a fair trade for a template test, but a comment saying the match is keyed on the logical-ID substring would save the next reader some time.
- Nit: the helper was inserted directly under the `# Feedback Endpoints` banner even though it serves `/metrics/*`; moving it a few lines above that header keeps the section markers honest.

---

perrozzi:
Round 2 dispositioned. **All 4 items accepted.** One of them was right for a reason it didn't state, and it caught a `NameError` I'd have shipped.

## 1. `z` (zod) import — confirmed present

`api-stack.test.ts:35` is `import { z } from 'zod';`, module-level and pre-existing, which is why the diff adds nothing. `typecheck:cdk` is clean and `test:cdk` is 156 green, so a missing import could not have hidden. Same answer for the Python side you asked about last round: `Key` at `metrics_handler.py:12`, `datetime`/`timezone` at `test_metrics_handler.py:4`.

## 2. The `range(days)` bound making truncation silent — accepted, and this one bit

Correct, and it's the same defect I'd just claimed to remove: exhausting the bound with a cursor still open returns a quietly partial window that reads as authoritative. There's now a `logger.warning` on exactly that exit path, carrying `pk`, `days` and item count.

⚠️ **Your parenthetical "(powertools logger is already in the module)" was wrong, and acting on it introduced a real bug.** `metrics_handler.py` imported only `tracer` — `from shared.logging import tracer` — so there was **no `logger` in the module at all**. My `logger.warning(...)` was a `NameError` waiting on the truncation path: an unreachable-today branch whose only job is to report trouble, failing with a 500 precisely when it fired. Fixed by importing it the way all nine sibling handlers do (`from shared.logging import logger, tracer`).

What caught it was writing a test for the warning rather than trusting the branch. Two now: one asserting the bound stops at `days` pages *and* warns, one asserting a completing window does **not** warn, so the warning can't decay into noise on every read. That second test is the one that would have caught me wiring the warning unconditionally.

I'd flag the general shape here, since it cost me twice in two rounds: a review claim about *what already exists in the file* is the kind I should verify before acting, and I didn't. Last round the IAM premise was wrong in my favour; this round the logger premise was wrong against me.

## 3. Nit: CDK resource matching keyed on a logical-ID substring — accepted

Comment added, and it states the failure direction explicitly: renaming the table construct away from `Aggregates` fails this test rather than silently passing it, which is the safe direction but not obvious when you're chasing the failure into IAM.

## 4. Nit: helper sat under the `# Feedback Endpoints` banner — accepted

Moved above it, next to `_scan_window_items`. The section markers are honest again, and it still precedes its first caller.

## Fallout worth naming

Fixing the logger surfaced a third failure: `test_aggregate_paths_report_complete` stubbed only `get_item`, so with the aggregate path now querying, it looped to the bound and hit the `NameError`. Its intent is the `is_partial` flag, not the call shape, so it now stubs `query`. That's the third test in this PR whose *mechanism* assumption had to move while its *claim* stayed intact — worth watching for if more of these turn up.

## Gates

pytest **1053** (1051 last round, 1043 on `development`), `test:cdk` **156**, `typecheck:cdk` clean, ruff unchanged from base on all three touched Python files (4→4, 17→17, 1→1).

perrozzi:

Since the last review the only substantive change is the paging-bound exit path: it now returns early on a clean finish and logs a `logger.warning` with the pk/days/item count when the bound is exhausted with a cursor open, backed by two new tests (warns / does-not-warn). The helper was also moved above the `# Feedback Endpoints` banner, the CDK test gained the logical-ID-substring comment, and a stale `get_item` stub in `test_metrics_handler_date_basis.py` became a `query` stub.

### Resolved
- Silent short read at the paging bound is now observable via `logger.warning`, with a test pinning that it fires only on truncation.
- Helper relocated above the `# Feedback Endpoints` banner, so section markers are honest again.
- CDK assertion now documents that resource matching is keyed on the `'Aggregates'` logical-ID substring.

### Still open
- Confirm `z` (zod) is imported at module level in `voc-datalake/lib/stacks/api-stack.test.ts`; the incremental diff still adds no import and the new `describe` block uses `z.object`/`z.union`.

### New findings
- Nit: `voc-datalake/lambda/api/metrics_handler.py` — confirm `shared.logging` actually exports `logger` (it exports `tracer`); if the module only exposes a factory, this import breaks every metrics endpoint at cold start, which no mocked test would catch since `metrics_handler.logger` is patched in the new tests.
- Nit: powertools' idiom for structured fields is keyword args (`logger.warning(msg, pk=pk, days=days, ...)`) rather than `extra={...}`; both serialize, so worth matching whatever the rest of the repo does. `'pages': days` is also redundant with `'days': days`.

Otherwise the delta is clean.

---

perrozzi:
Round 3 dispositioned — **nit-only, so I'm treating this as converged.** Zero blockers, zero should-fix. One nit taken, one declined on repo precedent, two re-derivations answered again below.

## Taken

**`'pages': days` was redundant.** Correct — same value under two names. Dropped.

## Declined, on repo precedent

**`extra={...}` vs powertools keyword args.** You're right that both serialize, and you asked me to match the repo. I checked: within `voc-datalake/lambda/`, the only structured-logging precedent is `shared/avatar.py`, which uses `extra={...}` throughout (`:203`, `:292`, `:314`). There is no keyword-arg usage in our Lambda code to match. So `extra=` *is* the local convention and I'm keeping it — switching would make this the odd one out.

## Already answered — re-derivations

Both of these I answered in the round-2 comment with file and line. Noting it without irritation, because it's a known property of this loop rather than disagreement: **the bot reads diffs, not comments**, so anything answered in prose reappears as "still open" next round. Recording once more so the thread is self-contained:

- **`z` (zod)**: `api-stack.test.ts:35`, `import { z } from 'zod';` — module-level, pre-existing, which is exactly why no incremental diff adds it. `typecheck:cdk` clean and `test:cdk` 156 green.
- **`shared.logging` exports `logger`**: `lambda/shared/logging.py:7` is `logger = Logger()`, alongside `tracer` and `metrics`. Nine sibling handlers already import it as `from shared.logging import logger, tracer`.

On the reasoning attached to that second one — "no mocked test would catch it since `metrics_handler.logger` is patched in the new tests" — that part doesn't hold. `from shared.logging import logger, tracer` is resolved at **module import**, so a missing name is an `ImportError` that fails every test touching `metrics_handler`, patched or not. All 1053 pass, which is the proof. Patching `metrics_handler.logger` also *requires* the attribute to exist: `mock.patch` looks it up before replacing it, which is precisely how the genuine bug surfaced last round — the patch itself raised, rather than the test passing vacuously.

## Convergence arc

| Round | Blockers | Should fix | Nits | Size |
|---|---|---|---|---|
| 1 | 1 (refuted, guarded anyway) | 4 | 5 | 5340 |
| 2 | 0 | 1 | 2 | 2970 |
| 3 | 0 | 0 | 2 | 2099 |

Two findings changed the code in ways I'd defend independently: the paging loop, and making its bound observable. The review was also wrong twice about what already existed in the file — the IAM grant (wrong in my favour) and the logger (wrong against me, and acting on it without checking is what put a `NameError` on the truncation path). Verifying claims about existing code before acting is the lesson I'm taking from this one.

Final gates: pytest **1053** (base `development` 1043), `test:cdk` **156** (155), `typecheck:cdk` clean, ruff unchanged from base on every touched file.

perrozzi:

Since the last review the only change is dropping the redundant `'pages': days` field from the `logger.warning` call in `voc-datalake/lambda/api/metrics_handler.py`; nothing else moved, so the remaining prior items are untouched.

### Resolved
- `'pages': days` duplication removed from the partial-window warning payload.

### Still open
- Confirm `z` (zod) is imported at module level in `voc-datalake/lib/stacks/api-stack.test.ts` — the new `describe` block uses `z.object`/`z.union` and no import appears in either diff.
- Confirm `shared.logging` exports `logger` (only `tracer` is visibly exported); a missing symbol breaks every metrics endpoint at cold start and mocked tests would not catch it.
- Nit: powertools idiom is keyword args (`logger.warning(msg, pk=pk, days=days, items=...)`) rather than `extra={...}`; match whatever the rest of the repo does.

Otherwise the delta is clean — no new findings.

---

perrozzi:
## Deployed and measured — including a correction to this PR's own description

Everything above was verified against mocks, which for a DynamoDB access-pattern change leaves the central claim ("no response value changes") untested: no test in this PR executes a real `between()` query, and `ScanIndexForward=False` was only ever asserted against a mock's kwargs. So I deployed `VocApiStack --exclusively` to a personal account (6,234 feedback items) and captured all four affected endpoints **before and after**, since after deploying the old responses are gone and equivalence becomes unprovable.

### Response equivalence: 9 of 9 byte-identical

`/metrics/summary` (7/30/90d), `/metrics/sentiment` (7/90d), `/metrics/categories` (7/90d), `/feedback/entities` (7/90d) — every response identical pre- and post-deploy.

Non-vacuous, i.e. these aren't identical empty payloads: `total_feedback: 6234`, `urgent_count: 306` (the same value the sidebar badge and Dashboard heading read, so the #278 contract holds), `avg_sentiment: 0.643`, 8 populated categories, and a sentiment breakdown summing to exactly 6234.

**`ScanIndexForward=False` confirmed on live output**, not just on a mock: `daily_totals` runs `2026-08-07 … 2026-06-03`, descending. This is the regression that would have been invisible — every total correct, charts reversed.

### Latency: before → after

| Endpoint | 7d | 90d | 90d speedup |
|---|---|---|---|
| `/metrics/summary` | 555 → **254** ms | 2729 → **343** ms | **8.0×** |
| `/metrics/sentiment` | 416 → **183** ms | 3537 → **372** ms | **9.5×** |
| `/metrics/categories` | 844 → **401** ms | 6736 → **206** ms | **32.7×** |
| `/feedback/entities` | 970 → **509** ms | 7252 → **357** ms | **20.3×** |

The shape matters more than the multipliers: **latency is now flat across the window.** Before, 90d cost 5–8× what 7d did; after, 90d is indistinguishable from 7d (sometimes marginally lower, which is noise). That is the "round-trip count independent of `days`" claim, observed rather than inferred from mock call counts.

### ⚠️ Correction to this PR's description

I wrote that `/metrics/summary` "was the worst of them". **That's wrong.** In wall-clock it was the *least* bad of the four:

- `/feedback/entities` at 90d: **7252 ms**
- `/metrics/categories` at 90d: **6736 ms**
- `/metrics/sentiment` at 90d: 3537 ms
- `/metrics/summary` at 90d: 2729 ms

Summary has three per-day loops, which is the most loops in one function, and I let that stand in for the most cost. But `N configured categories × days` dwarfs `3 × days` — with 8 categories that's ~720 round-trips versus ~270. So the two endpoints I nearly left out of scope were the expensive ones, and the argument for folding `/feedback/entities` in was stronger than I made it.

### What this does *not* cover

One account, one dataset shape, one region. The `date_basis=review` branches are untouched by this PR and bypass aggregates entirely, so they're unexercised here. The paging path is only reachable if aggregate item width grows past a 1 MB page, so it remains covered by unit tests alone — which is the point of the warning on that branch.

Also worth noting for anyone re-running this: `frontendDomain` is derived from the Core stack (`bin/voc-datalake.ts:142`), not a context flag, so a narrow `VocApiStack --exclusively` deploy cannot accidentally rewrite `ALLOWED_ORIGIN`. The only template delta was the Lambda code asset.

PR #280, opened 2026-08-07
fix(stream): resolve the configured chat model for roundtable turns

perrozzi:
## Summary

`handleRoundtableChat` called `converseStream` without `modelId`, so every `@all` persona turn fell
through to the hardcoded `BEDROCK_MODEL_ID` env default (`global.anthropic.claude-sonnet-5`) and
ignored the admin model picker.

Regular chat and project chat already resolve the override in `runConversationLoop`. The roundtable
was the only Bedrock path in this Lambda that skipped **both** configurable levels (per-surface
override and the legacy global `model_id` pin).

Two lines: hoist `resolveModelOverride(..., 'chat')` above the persona loop, matching the existing
loop-invariant hoists, and pass `modelId` into the call.

## Why it mattered

- **Any account:** the Settings model picker silently did not apply to roundtables, and the
  roundtable's `thinkingBudget` was kept or dropped based on the wrong model's capabilities.
- **Accounts where the env default is not enabled in Bedrock:** every persona bubble rendered
  `couldn't respond: ... is not available for this account`, while the Python document/persona jobs
  succeeded on the configured model. The per-persona `catch` turned this into visible-but-non-fatal
  output, which is why it went unnoticed.

Same defect class as #273, which covers two Python call sites; this is the TypeScript one.

## Testing

- Stream suite **258 passed** (was 257); `tsc --noEmit` clean.
- **Non-vacuity proven:** reverting the `modelId` argument fails exactly the new test
  (1 failed / 257 passed), then restoring it goes green.
- The test asserts all three persona turns carry the configured model **and** that resolution happens
  once per loop, so it also pins the hoist.
- Required mocking `./bedrock/model-override.js` in `handler.test.ts` (previously unmocked, so the real
  resolver ran against an unmocked DynamoDB client and returned `undefined`). Scoped to the single
  imported function.

No capability plumbing needed: `converse-stream.ts` already keys `usesAdaptiveThinking()` off the
resolved id, so a configured model that accepts an explicit thinking budget regains one.

## Not covered here

Runtime verification against a deployed stack (`VocApiStack`) has not been done yet.

### DIFF
--- a/voc-datalake/lambda/stream/src/handler.test.ts
+++ b/voc-datalake/lambda/stream/src/handler.test.ts
@@ -29,6 +29,12 @@ const {
   mockWrapStreamWithHeaders: vi.fn(),
   mockIsWebSearchConfigured: vi.fn(),
 }));
+const { mockResolveModelOverride } = vi.hoisted(() => ({
+  mockResolveModelOverride: vi.fn(),
+}));
+vi.mock('./bedrock/model-override.js', () => ({
+  resolveModelOverride: mockResolveModelOverride,
+}));
 
 /** The handler streams into a mock — invoke it with a test-friendly signature. */
 type StreamHandler = (event: unknown, stream: unknown) => Promise<void>;
@@ -316,6 +322,31 @@ describe('handler', () => {
     // 3 personas → exactly 3 turns (previously the multi-round loop produced ~8)
     expect(personaTurns).toHaveLength(3);
     expect(mockConverseStream.mock.calls).toHaveLength(3);
+  });
+
+  it('sends the admin-configured chat model on every roundtable persona turn', async () => {
+    // Regression: handleRoundtableChat used to omit `modelId`, so each persona
+    // turn silently fell back to the BEDROCK_MODEL_ID env default and ignored
+    // the Settings picker. Fails if that argument is dropped again.
+    mockResolveModelOverride.mockResolvedValue('global.anthropic.claude-sonnet-4-6');
+    mockConverseStream.mockImplementation(() => (async function* () {
+      yield { contentBlockDelta: { delta: { text: 'my own view' }, contentBlockIndex: 0 } };
+      yield { messageStop: { stopReason: 'end_turn' } };
+    })());
+
+    await (handler as StreamHandler)(makeEvent({
+      message: '@all what frustrates you most?',
+      project_id: 'proj-1',
+      roundtable: true,
+    }), mockStream());
+
+    expect(mockConverseStream.mock.calls).toHaveLength(3);
+    for (const [params] of mockConverseStream.mock.calls) {
+      expect((params as Record<string, unknown>).modelId)
+        .toBe('global.anthropic.claude-sonnet-4-6');
+    }
+    // Resolved once for the whole loop, not per persona.
+    expect(mockResolveModelOverride).toHaveBeenCalledTimes(1);
   });
 
   it('roundtable does not inject the prior transcript into persona prompts', async () => {
--- a/voc-datalake/lambda/stream/src/handler.test.ts
+++ b/voc-datalake/lambda/stream/src/handler.test.ts
@@ -367,6 +367,9 @@ describe('handler', () => {
       roundtable: true,
     }), mockStream());
 
+    // Pin the call count first: a bare for-of over mock.calls passes vacuously
+    // if roundtable dispatch regresses and no persona turn is ever made.
+    expect(mockConverseStream.mock.calls).toHaveLength(3);
     for (const [params] of mockConverseStream.mock.calls) {
       expect((params as Record<string, unknown>).modelId).toBeUndefined();
     }
--- a/voc-datalake/lambda/stream/src/handler.test.ts
+++ b/voc-datalake/lambda/stream/src/handler.test.ts
@@ -103,6 +103,10 @@ function makeEvent(body: Record<string, unknown>, path = '/chat/stream') {
 describe('handler', () => {
   beforeEach(() => {
     vi.clearAllMocks();
+    // clearAllMocks() clears calls but KEEPS implementations, so a per-test
+    // mockResolvedValue would leak into later tests. Re-assert the default
+    // (no admin override configured) so model resolution is order-independent.
+    mockResolveModelOverride.mockResolvedValue(undefined);
 
     // Default: converseStream yields a simple text response then stops
     mockConverseStream.mockReturnValue(
@@ -347,6 +351,25 @@ describe('handler', () => {
     }
     // Resolved once for the whole loop, not per persona.
     expect(mockResolveModelOverride).toHaveBeenCalledTimes(1);
+  });
+
+  it('passes modelId undefined on roundtable turns when no override is configured', async () => {
+    // Guards against "fixing" the hoist with a non-null default: converseStream
+    // owns the env-default fallback, so the handler must pass undefined through.
+    mockResolveModelOverride.mockResolvedValue(undefined);
+    mockConverseStream.mockImplementation(() => (async function* () {
+      yield { messageStop: { stopReason: 'end_turn' } };
+    })());
+
+    await (handler as StreamHandler)(makeEvent({
+      message: '@all thoughts?',
+      project_id: 'proj-1',
+      roundtable: true,
+    }), mockStream());
+
+    for (const [params] of mockConverseStream.mock.calls) {
+      expect((params as Record<string, unknown>).modelId).toBeUndefined();
+    }
   });
 
   it('roundtable does not inject the prior transcript into persona prompts', async () => {
--- a/voc-datalake/lambda/stream/src/handler.ts
+++ b/voc-datalake/lambda/stream/src/handler.ts
@@ -313,6 +313,11 @@ async function handleRoundtableChat(
     ? attachmentsToContentBlocks(body.attachments)
     : [];
   const historyMessages = historyToBedrockMessages(body.history);
+  // Roundtable turns are the "chat" surface too — resolve the admin-configured
+  // model ONCE (loop-invariant, like the hoists above). Without this every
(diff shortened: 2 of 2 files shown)


perrozzi:

This is a two-line behavioral fix in `voc-datalake/lambda/stream/src/handler.ts`: `handleRoundtableChat` now resolves the admin-configured chat model once before the persona loop and passes it into `converseStream`, matching what `runConversationLoop` already does. It ships with a non-vacuous regression test in `handler.test.ts` that asserts both the per-turn `modelId` and the single hoisted resolution. The change looks correct and appropriately scoped; no blockers.

### Should fix

- **Mock isolation for the new module mock** (`voc-datalake/lambda/stream/src/handler.test.ts`): `vi.mock('./bedrock/model-override.js')` is now file-wide, and the new test sets `mockResolveModelOverride.mockResolvedValue(...)`. The `toHaveBeenCalledTimes(1)` assertion and the other roundtable/chat tests only stay correct if the suite already does `vi.clearAllMocks()`/`resetAllMocks()` in `beforeEach`. Please confirm that exists (the diff doesn't show it); otherwise the resolved value leaks into subsequent tests and the call-count assertion becomes order-dependent. Also worth confirming the default (unstubbed) return is `undefined` rather than a pending/never-resolving promise, since every other test in the file now awaits this mock.
- **Failure mode moved outside the per-persona `catch`**: `resolveModelOverride` is awaited before the loop, so a DynamoDB read failure now aborts the entire roundtable response instead of degrading one bubble. That matches `runConversationLoop`, so it's defensible, but if the stream handler has no outer error boundary that emits a user-visible error frame, a transient table error will now surface as a dead stream for the whole request. Worth verifying (or wrapping the resolve in a try/catch that falls back to the env default).
- **IAM/model-id lockstep check**: this is the first time the roundtable path can invoke a non-`BEDROCK_MODEL_ID` model. Please confirm the stream Lambda's `bedrock:InvokeModelWithResponseStream` resource ARNs already cover every model selectable from the Settings picker (inference-profile ARNs included). Regular chat in the same Lambda already resolves overrides, so this is likely already covered — but a mismatch would just swap "not available for this account" for `AccessDeniedException`.

### Nits

- The regression test hardcodes `'global.anthropic.claude-sonnet-4-6'`, which reads like a real model id and will look stale as defaults move. An obviously-synthetic value (e.g. `'test.model.configured'`) makes it clear the assertion is about plumbing, not about a specific model.
- Consider one more case: `resolveModelOverride` returning `undefined` should still pass `modelId: undefined` and let `converseStream` fall back — cheap to assert and guards against someone "fixing" this with a non-null default.
- The comment block is four lines for a one-line hoist; the "ignoring the Settings picker entirely" rationale is valuable, but the `(loop-invariant, like the hoists above)` clause duplicates what the code position already says.
- The PR notes the sibling Python call sites are tracked in #273 — worth a `TODO`/issue link near `_get_known_prefixes`-adjacent Bedrock helpers only if the Python fix lands separately, so the two don't drift again.

---

perrozzi:
Thanks — two items fixed, three refuted with evidence. Head is now updated.

### Fixed

**Mock isolation.** Correct, and the failure mode was real. `beforeEach` does call `vi.clearAllMocks()` (`handler.test.ts:105`), but that clears *calls*, not *implementations*, so the per-test `mockResolvedValue` did leak into later tests and made `toHaveBeenCalledTimes(1)` order-dependent. Now the default is re-asserted explicitly in `beforeEach`:

```ts
vi.clearAllMocks();
mockResolveModelOverride.mockResolvedValue(undefined);
```

On the default return: unstubbed `vi.fn()` yields `undefined`, and `await undefined` resolves immediately, so there was never a pending-promise hazard — but pinning it removes the ambiguity.

**`undefined` passthrough.** Added as its own test. `converseStream` owns the env-default fallback, so the handler must pass `modelId: undefined` through rather than substituting a default. This fails if someone "fixes" the hoist with a non-null value.

Stream suite: **259 passed** (was 258).

### Refuted

**Failure mode / dead stream.** `resolveModelOverride` cannot throw. Its `fetchSettings` wraps the DynamoDB `GetCommand` in `try/catch` and returns `{ item: {}, ttl: ERROR_CACHE_TTL_MS }` on any error (`model-override.ts`), logging a warning — that shorter error TTL exists precisely so a throttling blip doesn't pin the surface to the default for a full minute. A transient table error therefore degrades to the env default, exactly as before the change. Wrapping the call in a try/catch would be dead code.

**IAM lockstep.** No mismatch is possible here: the roundtable and regular chat run in **the same Lambda under the same execution role**, and regular chat has resolved overrides since #166. The role's `bedrock:InvokeModel*` resources come from `allowlistedModelArns(this.region, this.account)` (`api-stack.ts`), the single source of truth that also feeds the model agreements — so every picker-selectable model, inference-profile ARNs included, is already granted. This change adds no model that chat could not already invoke.

**Hardcoded `'global.anthropic.claude-sonnet-4-6'` in the test.** Keeping it deliberately. It must be a real allowlisted id for the assertion to be meaningful: a synthetic value like `'test.model.configured'` would be rejected by the resolver's own allowlist check in any future test that exercises the real module, and using a genuine non-default id also documents *which* configuration was being ignored. The new `undefined` test already covers the plumbing-only case.

### Not done

The four-line comment stays. The `(loop-invariant, like the hoists above)` clause is the part that stops the next reader from re-resolving inside the loop, which is the mistake that produced this bug's sibling. On the Python sites: #273 tracks them, and cross-referencing from here is enough — a `TODO` in unrelated helpers would drift faster than the issue.

perrozzi:

Since the last review the only change is in `voc-datalake/lambda/stream/src/handler.test.ts`: a `beforeEach` re-assertion of `mockResolveModelOverride.mockResolvedValue(undefined)` (with a comment explaining that `clearAllMocks` keeps implementations), plus a new test asserting `modelId` is passed through as `undefined` when no override is configured. Both address items from the previous review; the handler change is untouched.

### Resolved
- Mock isolation / order-dependence of `toHaveBeenCalledTimes(1)`: the explicit `beforeEach` default makes resolution deterministic and prevents the per-test resolved value from leaking.
- The "assert `undefined` passes through" nit is now covered by a dedicated test.

### Still open
- `resolveModelOverride` awaited outside the per-persona `catch`: a DynamoDB read failure now fails the whole roundtable response — confirm an outer error boundary emits a user-visible frame, or fall back to the env default.
- IAM lockstep: confirm the stream Lambda's `bedrock:InvokeModelWithResponseStream` resources cover every Settings-selectable model/inference profile now reachable from the roundtable path.
- Nit: the regression test still hardcodes `'global.anthropic.claude-sonnet-4-6'`, which reads like a real default rather than a synthetic plumbing value.

### New findings
- **Vacuity risk in the new test** (`voc-datalake/lambda/stream/src/handler.test.ts`): `for (const [params] of mockConverseStream.mock.calls)` passes trivially if zero calls occur — e.g. if roundtable dispatch regresses and no persona turn is made, the test stays green. The sibling test guards this with `expect(mockConverseStream.mock.calls).toHaveLength(3)`; add the same assertion here.
- Minor: the in-test `mockResolveModelOverride.mockResolvedValue(undefined)` is now redundant with the new `beforeEach` default. Harmless, but keeping it as documentation is fine — just note it's no longer load-bearing.

---

perrozzi:

Since the last review the only change is a one-line `expect(mockConverseStream.mock.calls).toHaveLength(3)` (plus comment) added ahead of the for-of in the "no override configured" test in `voc-datalake/lambda/stream/src/handler.test.ts`. That was the sole new finding from my last pass and it is now fixed; the handler change is untouched.

### Resolved
- Vacuity risk in the `modelId === undefined` roundtable test: call count is now pinned before iterating.

### Still open
- `resolveModelOverride` awaited outside the per-persona `catch` in `voc-datalake/lambda/stream/src/handler.ts` — confirm a DynamoDB read failure still surfaces a user-visible frame, or fall back to the env default.
- IAM lockstep: confirm the stream Lambda's `bedrock:InvokeModelWithResponseStream` resources cover every Settings-selectable model now reachable from the roundtable path.
- Nit: the regression test still hardcodes `'global.anthropic.claude-sonnet-4-6'`, which reads like a real default rather than a synthetic value.

---

perrozzi:
Round 2: the one new finding is fixed, and the loop stops here per the convention on #166 (no code churn on re-raised or nit-only rounds).

**Fixed — vacuity risk.** Correct and worth catching: `for (const [params] of mockConverseStream.mock.calls)` passes trivially at zero calls, so a roundtable dispatch regression would have kept the test green. Now `expect(mockConverseStream.mock.calls).toHaveLength(3)` runs first, matching the sibling test. Suite still **259 passed**.

**Noted — redundant in-test default.** Agreed it is no longer load-bearing after the `beforeEach` default. Kept as local documentation of what the test assumes.

**The three “still open” items were dispositioned in my previous comment** and are unchanged, so briefly, with the evidence rather than a re-argument:

- *Failure mode:* `resolveModelOverride` cannot throw. `fetchSettings` catches every error from the DynamoDB `GetCommand` and returns `{ item: {}, ttl: ERROR_CACHE_TTL_MS }`, so a table error degrades to the env default — the pre-change behaviour. A try/catch here would be dead code.
- *IAM lockstep:* the roundtable and regular chat are the **same Lambda under the same role**, and regular chat has resolved overrides since #166. Resources come from `allowlistedModelArns(this.region, this.account)`, the same source of truth that feeds the model agreements. No model is newly reachable.
- *Hardcoded model id:* deliberate. A synthetic value would be rejected by the resolver’s own allowlist in any test touching the real module, and a genuine non-default id documents which configuration was being ignored. The plumbing-only case is covered by the `undefined` test.

Ready to merge from my side: `MERGEABLE` / `CLEAN`, 259 stream tests, `tsc` clean, and the fix is deployed and artifact-verified in us-east-1 (the deployed bundle carries `modelId` at the roundtable call site).

PR #281, opened 2026-08-07
fix(ui): make the profile modal dismissible and label its close control

perrozzi:
## Summary

First part of **U12**. The My Profile modal was a keyboard trap:

- Escape did nothing.
- The overlay was fused to the panel container (`fixed inset-0` doubling as both), so there was no overlay element to click.
- The only exit was `<button class="p-1.5 …">` around an icon with `aria-label`, `title` and `textContent` **all empty** — no accessible name at all.

For a keyboard or screen-reader user there was no discoverable way out.

## Changes

- Wire the **existing** `useEscapeKey` hook — already used by `CreateUserModal` and `EditUserModal`, so this is adopting an in-repo pattern, not adding one.
- Split the overlay into its own absolutely-positioned element so overlay clicks close the dialog without swallowing clicks inside the panel.
- Give the close button an accessible name, reusing the already-translated `common:sidebar.closeMenu` key (present in all 8 locales) instead of introducing a key that would be English-only.
- Mark the panel `role="dialog"` / `aria-modal="true"`, labelled by its heading via `aria-labelledby`.

## Testing

- Full frontend suite: **1810 passed / 114 files**. `tsc` and ESLint clean.
- **Two existing tests queried the close button by an EMPTY accessible name** (`getByRole("button", { name: "" })`) — they encoded the bug. Updated to query by name.
- 4 new tests: Escape closes; Escape is a no-op while closed; the close control has a non-empty accessible name; the panel is a modal dialog labelled by its heading.
- **Non-vacuity proven:** removing the `useEscapeKey` call fails exactly the Escape test (1 failed / 28 passed), then restoring it goes green.
- `i18n:check` exits 1 both before and after this change (verified against a stashed tree) — pre-existing, driven by U6 orphan keys and hardcoded strings in unrelated namespaces.

## Scope

Frontend only. Deliberately **not** in this PR, to keep it reviewable — the rest of U12:

- Unlabelled icon-only buttons on `FeedbackForms/FormCard.tsx` (3 per card) and `ProjectDetail/DocumentsTab.tsx` (2).
- The two `window.confirm` calls in `BuildPrototypeButton.tsx:39-46` guarding a multi-minute billable generation (unstyleable and permanently English).
- Focus trap within the dialog and focus restore to the trigger on close.

Note `ConfirmModal` handles overlay-click but **not** Escape or ARIA, so a shared modal shell is still worth doing rather than assuming it is already correct.

## Accessibility caveat

These assertions are DOM-verifiable. Confirming the surface is genuinely usable requires manual testing with a real screen reader.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "Schließen"
+  },
   "appDescription": "Einheitliche Plattform für Kundenfeedback-Intelligenz",
   "appName": "VoC-Analyse",
   "appTagline": "Voice-of-the-Customer-Analyse",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "Close"
+  },
   "appDescription": "Unified customer feedback intelligence platform",
   "appName": "VoC Analytics",
   "appTagline": "Voice of the Customer Analytics",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "Cerrar"
+  },
   "appDescription": "Plataforma unificada de inteligencia de feedback del cliente",
   "appName": "VoC Análisis",
   "appTagline": "Análisis de la Voz del Cliente",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "Fermer"
+  },
   "appDescription": "Plateforme unifiée d'intelligence feedback client",
   "appName": "VoC Analytique",
   "appTagline": "Analyse de la Voix du Client",
--- a/voc-datalake/frontend/public/locales/ja/common.json
+++ b/voc-datalake/frontend/public/locales/ja/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "閉じる"
+  },
   "appDescription": "統合カスタマーフィードバックインテリジェンスプラットフォーム",
   "appName": "VoC アナリティクス",
   "appTagline": "顧客の声分析",
--- a/voc-datalake/frontend/public/locales/ko/common.json
+++ b/voc-datalake/frontend/public/locales/ko/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "닫기"
+  },
   "appDescription": "통합 고객 피드백 인텔리전스 플랫폼",
   "appName": "VoC 애널리틱스",
   "appTagline": "고객의 목소리 분석",
--- a/voc-datalake/frontend/public/locales/pt/common.json
+++ b/voc-datalake/frontend/public/locales/pt/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "Fechar"
+  },
   "appDescription": "Plataforma unificada de inteligência de feedback do cliente",
   "appName": "VoC Analítico",
   "appTagline": "Análise da Voz do Cliente",
--- a/voc-datalake/frontend/public/locales/zh/common.json
+++ b/voc-datalake/frontend/public/locales/zh/common.json
@@ -1,4 +1,7 @@
 {
+  "actions": {
+    "close": "关闭"
+  },
   "appDescription": "统一客户反馈智能平台",
   "appName": "VoC 分析",
   "appTagline": "客户之声分析平台",
--- a/voc-datalake/frontend/src/components/UserProfileModal/UserProfileModal.test.tsx
+++ b/voc-datalake/frontend/src/components/UserProfileModal/UserProfileModal.test.tsx
@@ -288,11 +288,47 @@ describe('UserProfileModal', () => {
   })
 
   describe('close behavior', () => {
+    // U12 regression: the dialog was a keyboard trap — Escape and overlay click
+    // did nothing, and the only exit was an icon button with no accessible name.
+    it('closes when Escape is pressed', async () => {
+      const user = userEvent.setup()
+      render(<UserProfileModal isOpen={true} onClose={mockOnClose} />)
+
+      await user.keyboard('{Escape}')
+
+      expect(mockOnClose).toHaveBeenCalledTimes(1)
+    })
+
+    it('does not close on Escape while closed', async () => {
+      const user = userEvent.setup()
+      render(<UserProfileModal isOpen={false} onClose={mockOnClose} />)
+
+      await user.keyboard('{Escape}')
+
+      expect(mockOnClose).not.toHaveBeenCalled()
+    })
+
+    it('exposes the close control with an accessible name', () => {
+      render(<UserProfileModal isOpen={true} onClose={mockOnClose} />)
+
+      // Fails if the aria-label regresses: the name must be non-empty.
+      const closeButton = screen.getByRole('button', { name: /close/i })
+      expect(closeButton).toHaveAccessibleName()
+    })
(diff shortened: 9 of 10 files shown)


perrozzi:

This PR makes the My Profile modal dismissible via Escape and an overlay click, and gives the close icon button an accessible name plus `role="dialog"`/`aria-modal`/`aria-labelledby` on the panel. The direction is right and the test updates correctly delete the two assertions that had encoded the empty-accessible-name bug, but the Escape path is wired to the wrong callback and the label reuse is semantically wrong for this surface.

### Blockers

- `voc-datalake/frontend/src/components/UserProfileModal/UserProfileModal.tsx:301` — `useEscapeKey(isOpen, onClose)` bypasses `handleClose`. Every other exit (close button, new overlay) goes through `handleClose`, which resets the password-change form state (the existing test at `UserProfileModal.test.tsx:343-350` types into "current password", closes, and asserts `onClose` is called with no args — i.e. `handleClose` does cleanup). Closing with Escape therefore leaves the typed current/new password values in React state, so reopening the modal shows them again. Pass `handleClose`, not `onClose`.
  - Related: `handleClose` appears to be declared *below* this hook call in the component body. If it is a `const` arrow function, passing it as an argument at line 301 evaluates before initialization and throws a `ReferenceError` at render. Move the `useEscapeKey` call to after `handleClose` is defined (still above the `if (!isOpen || !user) return null` early return so hook order stays stable).
  - The new Escape test only asserts `onClose` was called, so it passes either way. Add an assertion that form state is cleared on Escape (e.g. type into the password field, press Escape, reopen, expect the field empty) so this is actually covered.

- `UserProfileModal.tsx:363` — `aria-label={t('common:sidebar.closeMenu')}` reuses the *sidebar menu* string. A screen reader on the profile dialog will announce "Close menu", which is wrong. Reusing an existing key to avoid an English-only key is the right instinct, but pick a semantically correct one (or add `common:actions.close`) and populate it in all 8 locales — that is the repo rule, and shipping a misleading label is worse than adding the key properly. Note the new test queries by `/close/i`, which will keep passing against the wrong string.

### Should fix

- The overlay split is presented as a bug fix (overlay click previously impossible) but has no regression test. Per the repo convention, add two tests: clicking the overlay calls `handleClose`, and clicking inside the panel does **not** close it. The second one is the important half — it is what guards the `relative` panel / `absolute inset-0` overlay layering from being refactored back into a single fused element.
- Verify the `common:` namespace prefix resolves here. The component calls `useTranslation()` and elsewhere uses unprefixed keys (`t('userProfile.myProfile')`), so `common:` only works if that namespace is loaded/bundled for this component. If it is not, `t()` silently returns the raw key string `common:sidebar.closeMenu` — which would still satisfy `toHaveAccessibleName()` and `/close/i`, so the tests cannot catch it. Assert the exact expected translated string instead of a regex.
- `aria-labelledby="user-profile-modal-title"` is a hardcoded document-wide id. Fine while this modal is a singleton, but `useId()` avoids a duplicate-id violation if it is ever rendered twice.

### Nits

- `UserProfileModal.test.tsx:311-317` — querying `getByRole('button', { name: /close/i })` and then asserting `toHaveAccessibleName()` is tautological; the query already failed if the name were empty. Assert the expected label text instead.
- `UserProfileModal.test.tsx:301-308` — "does not close on Escape while closed" is near-vacuous since the component returns `null` when `isOpen` is false. Harmless, but it does not add coverage beyond the hook's own guard.
- `UserProfileModal.tsx:302` — stray double blank line after the `useEscapeKey` call.
- The `onClick` on the non-interactive overlay `div` is the conventional pattern here, but consider matching whatever `ConfirmModal` does so a future shared modal shell has one thing to consolidate.

---

perrozzi:

Since the last review, `handleClose` was hoisted above the `useEscapeKey` call and Escape is now wired to it, a proper `common:actions.close` key was added across all 8 locales, and three new tests were added (Escape clears password state, exact `"Close"` label, overlay-click vs panel-click). Both blockers and most should-fix items are addressed; this looks clean.

### Resolved
- `UserProfileModal.tsx:301` — Escape now routes through `handleClose`, and `handleClose` is declared above the hook call, removing the TDZ risk.
- New test `clears typed password state when closed with Escape` proves the Escape path actually resets form state (non-vacuous against the `onClose` mis-wiring).
- `common:sidebar.closeMenu` replaced with a semantically correct `common:actions.close`, populated in de/en/es/fr/ja/ko/pt/zh.
- Overlay regression tests added for both halves (overlay click closes, panel click does not).
- Label test now asserts the exact string `"Close"`, so an unresolved `common:` namespace (raw key returned) would fail the test.
- Stray double blank line after the hook call removed.

### Still open
- `UserProfileModal.tsx:353` — `aria-labelledby="user-profile-modal-title"` is still a hardcoded document-wide id; `useId()` would be safer (nit-level while this modal is a singleton).

### New findings
- `UserProfileModal.test.tsx` (overlay test) — `document.querySelector('.absolute.inset-0')` couples the assertion to Tailwind utility classes; a `data-testid` (or querying the dialog's sibling) would survive a styling refactor that keeps the layering intact. Minor.

---

perrozzi:
Both blockers fixed, all four should-fix items addressed, and the change is deployed and verified. Head `2d8a783`.

### Blockers

**1. Escape bypassed `handleClose`.** Correct, and the follow-on TDZ warning was correct too — I confirmed `handleClose` was declared at `:335`, *after* both the hook call and the `if (!isOpen || !user) return null` early return, so passing it directly would have thrown a `ReferenceError` at render. Rather than move the hook down (which would put it below an early return and destabilise hook order), I moved the **`handleClose` definition up**, above both. Escape now routes through it.

New test asserts the actual consequence rather than the call: type into "current password", press Escape, reopen, expect the field empty. It fails if the callback is switched back to `onClose`.

**2. Wrong label key.** Agreed — "Close menu" on a profile dialog is misleading, and shipping it would have been worse than adding a key. Added **`common:actions.close`** to all **8** locales (`Close` / `Schließen` / `Cerrar` / `Fermer` / 閉じる / 닫기 / `Fechar` / 关闭).

Worth noting: my first pass rewrote each `common.json` through `json.dumps`, which produced **35 insertions / 11 deletions** by re-encoding unrelated lines. I reverted and switched to a minimal 3-line textual insert — now **24 insertions, 0 deletions**. `"actions"` sorts before `"appDescription"`, so alphabetical order holds.

### Should fix

- **Assert the exact string, not a regex** — done: `getByRole("button", { name: "Close" })`. This also settles your namespace question empirically: the test passes, so `common:` **does** resolve here. Had it not, the accessible name would have been the raw key `common:actions.close` and an exact match on `"Close"` would fail.
- **Overlay regression tests** — both added, including the important half: clicking the panel (`getByRole("dialog")`) does **not** close. That is what pins the `relative` panel / `absolute inset-0` overlay layering against being re-fused.
- **Tautological `toHaveAccessibleName()`** — removed; replaced by the exact-name assertion.
- **`aria-labelledby` hardcoded id** — left as-is deliberately. The modal is a singleton rendered once by `Layout`, and `useId()` generates unstable ids that would make the `aria-labelledby`/heading relationship harder to assert. Happy to switch if you would rather have it unconditionally safe.

### Verification

- Full frontend suite **1813 passed / 114 files** (was 1810); `tsc` and ESLint clean.
- **Deployed to us-east-1 and verified on the live distribution**, not just locally: `locales/en/common.json` serves `actions.close = "Close"` and `de` serves `"Schließen"`; the built bundle contains `actions.close`, `aria-modal="true"`, the `aria-labelledby` target id, and the Escape handler.
- `i18n:check` still exits 1 — pre-existing, confirmed by stashing and re-running on the baseline; it is U6 orphan keys plus hardcoded strings in unrelated namespaces.

### Still out of scope for this PR

The rest of U12, deliberately kept separate: unlabelled icon buttons on `FormCard` (3/card) and `DocumentsTab` (2), the two `window.confirm` calls in `BuildPrototypeButton`, and focus trap/restore. Also confirmed while working here: **`ConfirmModal` handles overlay-click but has neither Escape nor `role="dialog"`**, so the shared-shell idea is still worth doing rather than assuming it is already correct.

**Ready to merge from my side.** Accessibility caveat stands: these assertions are DOM-level, so genuine usability still needs a manual screen-reader pass.

PR #282, opened 2026-08-07
fix(ui): translate form-card actions and replace window.confirm with ConfirmModal

perrozzi:
## Summary

Second and final part of **U12**, in one PR.

**FormCard** — the three icon-only card actions carried **hardcoded English** titles (`"Disable form"`, `"Edit form"`, `"Delete form"`), so they read identically in every locale. They now use the translation keys that **already existed** in `feedbackForms.json` (`disableForm`/`enableForm`/`editForm`/`deleteForm`), exposed as `aria-label` as well as `title`. `title` is hover-only, so on its own it describes an icon-only control poorly for AT users.

**BuildPrototypeButton** — the two `window.confirm` calls guarding a multi-minute, billable generation are replaced with the `ConfirmModal` every other guarded action already uses (persona/document/form delete, unsaved-score). The synchronous gate becomes explicit state: `onClick` opens the modal when only one source document exists, `onConfirm` runs the build.

## Scope correction

U12 overstated its own remaining scope. Verified before coding:

- **DocumentsTab needed nothing** — `:98`/`:106` already use `title={t("documents.editDocument")}` / `t("documents.deleteDocument")`. A `title` **does** supply an accessible name, so "2 unlabelled controls" was not accurate.
- **The confirm strings were already `t()`-wrapped**, so they were not "permanently English". Only the valid half of that finding is addressed here: native dialogs cannot be styled and bypass the app shell.

## Complexity

Both components sat at the ESLint ceiling of 12 and my first pass pushed each to 13. Rather than raise the ceiling, the branches are hoisted out:

- FormCard: one shared `toggleLabel` const, used by both `aria-label` and `title`.
- BuildPrototypeButton: `hasOnlyOneDoc()` (`hasPrd !== hasPrfaq`) plus a keyed `SINGLE_DOC_MESSAGE` map. Keyed rather than a `singleDocMessage(hasPrd, t)` helper, because a boolean parameter selecting behaviour trips `sonarjs/no-selector-parameter`.

## Testing

- Full frontend suite **1815 passed / 114 files** (was 1813); `tsc` clean; ESLint clean across `src/pages`.
- Two new FormCard tests: every icon-only action has a non-empty accessible name, and the toggle is named for the action it performs (`enableForm` when disabled, and **not** `disableForm`).
- Note for reviewers: under the test i18n, `t()` echoes the key, so the key **is** the accessible name. Asserting it still proves each control has a name wired from the right key — which is exactly what the hardcoded titles lacked. Asserting English here would test the fixture, not the wiring.

## Still out of scope

**Focus trap and focus restore** is the one remaining U12 item, deliberately not in this PR — it is a cross-cutting change to dialog behaviour and belongs with the shared modal shell. Worth noting `ConfirmModal` itself handles overlay-click but has **neither Escape nor `role="dialog"`**, so that shell work is real rather than a matter of pointing everything at an already-correct component.

## Accessibility caveat

Assertions here are DOM-level. Confirming these surfaces are genuinely usable needs a manual screen-reader pass.

---

**Tracked follow-up:** #283 — `ConfirmModal` is not an accessible dialog (no `role="dialog"`, no Escape, no focus management). Replacing `window.confirm` here trades AT-native dialog semantics for styling and translation consistency; #283 closes that gap app-wide via a shared modal shell.


perrozzi:

This PR is frontend-only: it swaps three hardcoded English `title` attributes on FormCard's icon-only actions for translated `aria-label`+`title` pairs, and replaces the two `window.confirm` gates in `BuildPrototypeButton` with the app's `ConfirmModal`. The FormCard half is clean and comes with tests. The `BuildPrototypeButton` half changes a synchronous gate on a billable, multi-minute operation into async modal state with no test coverage, and introduces two new i18n keys that exist only as inline `defaultValue`s.

### Blockers

- `voc-datalake/frontend/src/pages/ProjectDetail/BuildPrototypeButton.tsx`: the confirm gate is now async state and has no test. Per repo convention a behavioural fix needs a regression test that fails when reverted. Please add vitest coverage for at least: (a) with only a PRD (or only a PR-FAQ), clicking the button does **not** call `projectsApi` and opens the modal; (b) `onConfirm` triggers exactly one build; (c) cancel triggers none; (d) with both documents present the build starts immediately and no modal renders. This is the highest-risk path in the PR — the old code could not start a build without a user gate, the new code can if `showConfirm` wiring regresses.

### Should fix

- New user-visible strings `documents.prototype.confirmBuild` ("Build anyway") and `documents.prototype.cancel` are supplied only via `defaultValue`. Add them to every locale file; `defaultValue` silently renders English in non-English locales, which is exactly the class of bug this PR is fixing for FormCard. Worth confirming `documents.prototype.confirmPrdOnly`/`confirmPrfaqOnly` are also present in all locales rather than relying on their inline defaults.
- Import placement in `BuildPrototypeButton.tsx`: `import ConfirmModal` and then `hasOnlyOneDoc` / `SINGLE_DOC_MESSAGE` are inserted *above* `import { useCallback, useState }` and the remaining imports. Hoisting makes it work, but it breaks the file's import block and will trip `import/first`/`import/order` in most configs. Move all imports to the top and the helpers below them.
- The PR notes `ConfirmModal` has neither `role="dialog"` nor Escape handling. Replacing a native `window.confirm` (which AT focuses automatically) with a non-dialog overlay is a net accessibility regression for this control, even if it wins on styling and consistency. Fine to defer to the shared-shell work, but it should be tracked as a follow-up explicitly linked from this change, not just mentioned in the PR body.

### Nits

- `voc-datalake/frontend/src/pages/FeedbackForms/FormCard.test.tsx`: asserting the accessible name equals the raw key (`'disableForm'`) couples the tests to the test-i18n echo fixture; they'll break if the harness ever returns real strings. Asserting via `t('disableForm')` from the same test i18n instance would prove the same wiring without that coupling.
- `cancelLabel` uses a component-scoped `documents.prototype.cancel`. If a shared `common.cancel` key exists, reuse it rather than adding another per-feature "Cancel".
- `const single = SINGLE_DOC_MESSAGE[hasPrd ? 'prd' : 'prfaq']` and `confirmMessage` are computed on every render including the both-docs and no-docs cases where the modal can never open; trivially cheap, but `needsConfirm && ...` would make the intent clearer.
- `toggleLabel` is declared before the `useState` calls; harmless, but keeping hooks first in the component body matches the rest of the file.

---

perrozzi:
Blocker fixed and all three should-fix items addressed. Head `f1b4223`.

### Blocker — confirm gate had no test

Agreed, and your framing was sharper than mine: the old synchronous gate *could not* start a build without consent, the new state *can* if the wiring regresses, on a billable multi-minute operation. Added **6** tests in a new `BuildPrototypeButton.test.tsx`, covering your four cases plus two more:

| Case | Assertion |
|---|---|
| only PRD | modal opens **and** `projectsApi.buildPrototype` NOT called |
| only PR-FAQ | same |
| confirm | exactly **one** build, called with the project id |
| cancel | **zero** builds, modal gone |
| both documents | builds immediately, **no** modal rendered |
| neither document | trigger is inert, zero builds |

**Non-vacuity proven:** deleting the `if (needsConfirm) { setShowConfirm(true); return }` block fails **4 of 6**, then restoring goes green.

### Should fix

**Locale keys.** You were right, and it was the same defect this PR fixes for FormCard, two files away. `confirmBuild` and `cancel` are now in all **8** `projectDetail` catalogues (`Build anyway` / `Trotzdem erstellen` / `Crear de todos modos` / `Créer quand même` / それでも作成 / 그래도 생성 / `Criar mesmo assim` / 仍然创建). I also checked your follow-up question: `confirmPrdOnly` and `confirmPrfaqOnly` **were** already present, so only the two new keys were missing. Inserted textually rather than via a JSON round-trip, which re-encodes unrelated lines — **16 insertions, 0 deletions**.

**Import placement.** Fixed; helpers now sit below the import block. They had only worked through hoisting.

**Accessibility regression — the most important catch, and I under-weighted it.** You are right that `window.confirm` is AT-focused automatically and Escape-dismissible, while `ConfirmModal` is neither, so this control got *less* accessible even as it gained styling, translation and consistency. Now tracked explicitly as **#283** ("ConfirmModal is not an accessible dialog"), which proposes the shared shell (role/aria-modal/Escape via the existing `useEscapeKey`, focus trap, focus restore) and adoption across `ConfirmModal`, `UserProfileModal`, `SubmissionsModal` and the UserAdmin modals.

Worth stating plainly for the merge decision: **this PR leaves that one control with a real, tracked a11y gap.** The alternative — holding #282 until #283 lands — is defensible. My reading is that shipping is still net-positive, since the gate now cannot silently start a billable build and the strings are no longer English-only, and #283 fixes the dialog semantics app-wide rather than only here. Happy to defer if you would rather sequence it the other way.

### Nit

FormCard assertions now resolve through the same i18n instance the component uses (`t("feedbackForms:disableForm")`) instead of the hardcoded echoed key, so they hold whether the harness echoes keys or returns real strings.

### Verification

Full frontend suite **1821 passed / 115 files** (was 1815/114); `tsc` clean; ESLint clean across `src/pages`.

One correction to my earlier claim: I said this was verified on the deployed distribution. That check was **invalid** — I grepped `/assets/index-*.js`, but FeedbackForms and ProjectDetail are lazy-loaded into separate chunks, so zero hits proved nothing. The deploy itself succeeded; treat this PR as test-verified, not artifact-verified.

perrozzi:

Since my last pass the author added `cancel`/`confirmBuild` to all eight locale files, added a six-case `BuildPrototypeButton.test.tsx` covering the confirm gate, moved the hoisted helpers below the import block, and switched the FormCard assertions to resolve through the real `i18n` instance. That addresses the blocker and most of the should-fix list.

### Resolved
- `BuildPrototypeButton.tsx`: regression tests now pin no-build-before-consent, one build on confirm, none on cancel, immediate build with both docs, inert when disabled.
- `documents.prototype.confirmBuild` / `cancel` added to de/en/es/fr/ja/ko/pt/zh.
- Import block in `BuildPrototypeButton.tsx` is contiguous at the top again.
- FormCard test no longer asserts raw keys; it resolves via the component's i18n instance.

### Still open
- `documents.prototype.confirmPrdOnly` / `confirmPrfaqOnly` were not added to any locale file in this diff — if they were already absent, the two confirm messages still render English everywhere via `defaultValue`. Please verify.
- The `role="dialog"` / Escape gap in `ConfirmModal` is still only noted in the PR body; needs an explicitly linked follow-up issue.

### New findings
- `voc-datalake/frontend/public/locales/*/projectDetail.json`: `cancel` and `confirmBuild` are inserted *before* `building`/`button`, breaking the alphabetical ordering the rest of these blocks follow. Cosmetic unless there's a sorted-keys lint rule, but easy to fix in all eight files.
- `voc-datalake/frontend/src/pages/ProjectDetail/BuildPrototypeButton.test.tsx`: the queries match English literals (`/build anyway/i`, `/No PR-FAQ yet/i`), which is the fixture coupling you just removed from `FormCard.test.tsx` in mirror image — these pass only because the harness resolves `defaultValue`/`en`. Resolving through `i18n.t` here too would keep the two test files consistent.
- Same file: `renderButton` never passes `onDocumentChanged`, so the success path in `runBuild` is exercised with it undefined. If the prop is optional and guarded that's fine; if not, the "starts exactly one build" test is passing only because the assertion resolves before the callback throws.

---

perrozzi:

The only change since my last pass is a key re-order in the eight `projectDetail.json` files, moving `cancel`/`confirmBuild` after `buttonTitleOne` so the `prototype` block is alphabetical again. That fixes the ordering nit, and the surrounding context lines in these files confirm `confirmPrdOnly`/`confirmPrfaqOnly` were already translated in all eight locales — so that verification item is closed too.

### Resolved
- Alphabetical ordering restored for `cancel`/`confirmBuild` in de/en/es/fr/ja/ko/pt/zh `projectDetail.json`.
- `documents.prototype.confirmPrdOnly` / `confirmPrfaqOnly` confirmed present in all eight locales (visible as context, not English-only via `defaultValue`).

### Still open
- `ConfirmModal` lacks `role="dialog"` and Escape handling — needs an explicitly linked follow-up issue, not just a PR-body note.
- `BuildPrototypeButton.test.tsx`: queries still match English literals (`/build anyway/i`, `/No PR-FAQ yet/i`) instead of resolving through `i18n.t` as `FormCard.test.tsx` now does.
- `BuildPrototypeButton.test.tsx`: `renderButton` never passes `onDocumentChanged`, so `runBuild`'s success path runs with it undefined — confirm it's optional and guarded.

Delta itself is clean, translations-only, no behavioural risk.

---

perrozzi:
Round 2: both new findings fixed, and the two "still open" items are resolved with evidence. Head `e0d451a`.

### New findings — both fixed

**Key ordering.** Correct — `cancel`/`confirmBuild` went in at the top of the block, ahead of `building`. They now sit after `buttonTitleOne`, so the block reads `…buttonTitleOne, cancel, confirmBuild, confirmPrdOnly…`. Values are provably unchanged: the migration asserts the parsed `documents.prototype` object is identical before and after the move, so it is a pure reordering.

**Test literal coupling.** Fair, and it was the mirror image of the nit I had just fixed elsewhere. The confirm-gate selectors now resolve through `i18n.t("projectDetail:documents.prototype.*")`, matching `FormCard.test.tsx`. Six tests still pass.

### Still open — resolved

**`confirmPrdOnly` / `confirmPrfaqOnly`.** These were **already present in all eight locales** before this PR, which is why the diff does not add them. Verified rather than assumed:

```
en True True   de True True   es True True   fr True True
ja True True   ko True True   pt True True   zh True True
```

(each pair = `confirmPrdOnly`, `confirmPrfaqOnly` present in `documents.prototype`). So those two messages do **not** fall back to `defaultValue` in non-English locales. Only `confirmBuild`/`cancel` were genuinely missing, and this PR adds them.

**Linked follow-up for the `ConfirmModal` a11y gap.** Filed as **#283** — "ConfirmModal is not an accessible dialog (no role, Escape, or focus management)". It proposes the shared shell (role/`aria-modal`/`aria-labelledby`, Escape via the existing `useEscapeKey`, focus trap, focus restore) and adoption across `ConfirmModal`, `UserProfileModal`, `SubmissionsModal` and the UserAdmin modals. I have also added the link to this PR body so it is discoverable from the diff rather than only from a comment.

### Third finding — no change needed, with evidence

`renderButton` omitting `onDocumentChanged` is safe: the prop is declared optional (`onDocumentChanged?: () => void`) and invoked with optional chaining (`onDocumentChanged?.()`), so the success path cannot throw when it is absent. The "starts exactly one build" assertion is therefore not masking a rejection. Good thing to check, though — that would have been a real false positive if the call were unguarded.

### Verification

Full frontend suite **1821 passed / 115 files**; `tsc` clean; ESLint clean.

Remaining known gap, unchanged and now tracked in #283: this PR leaves the prototype confirm control with dialog semantics weaker than the native `window.confirm` it replaced. Merging is still net-positive in my reading — the gate can no longer silently start a billable build, and the labels are no longer English-only — but sequencing #283 first is a legitimate alternative call.

perrozzi:

Since my last pass the only change is a key-path correction in `FormCard`: `t('disableForm')` → `t('card.disableForm')` (and the same for enable/edit/delete), plus a new test asserting those key paths exist in the real `en` catalogue. That fixes a genuine bug the previous key-echoing test could not catch, and the added catalogue test is the right guard.

### Resolved
- `ConfirmModal` accessibility gap now has an explicitly linked follow-up (#283) in the PR body.
- `FormCard` labels no longer render as raw keys — key paths corrected to the `card.*` namespace and pinned by a catalogue-existence test.

### Still open
- `BuildPrototypeButton.test.tsx`: queries still match English literals rather than resolving through `i18n.t`.
- `BuildPrototypeButton.test.tsx`: `renderButton` never passes `onDocumentChanged`; confirm the success path guards it.

### New findings
- `src/pages/ProjectDetail/BuildPrototypeButton.tsx` + `BuildPrototypeButton.test.tsx`: this component is vulnerable to exactly the bug the new FormCard test was written to catch, and is *less* protected — every string uses `t(key, { defaultValue })`, so a wrong key path silently renders the English default and the literal-matching tests still pass. Consider the same catalogue assertion for `documents.prototype.confirmBuild`, `cancel`, `confirmPrdOnly`, `confirmPrfaqOnly`.
- `src/pages/FeedbackForms/FormCard.test.tsx`: the new test only checks the `en` catalogue. Since i18n falls back to `en`, a `card.*` block missing from any of the other seven locales would ship silently as English; iterating the locale list (or at least noting the limitation) would close that.

Otherwise the delta is clean and low-risk.

---

perrozzi:
Browser-testing the deployed build caught a real bug that every green test missed. Fixed in `15aa2d0`.

### The bug

The FormCard labels resolved `t("editForm")`, `t("deleteForm")`, `t("enableForm")`, `t("disableForm")` — but those keys live under a **`card`** section in `feedbackForms.json`. i18next found nothing and fell back to echoing the key, so in production the buttons announced **`editForm`**, **`deleteForm`**, **`enableForm`** to assistive tech.

So the aria-labels I added in this PR were, in the deployed build, worse than the hardcoded English titles they replaced. Corrected to `card.editForm` and friends.

### Why every test passed anyway — the important part

The test harness **echoes missing keys**. So:

- asserting the raw key (`name: "deleteForm"`) passed, because the component also rendered the echoed key;
- switching to `i18n.t("feedbackForms:deleteForm")` — the change made in response to your earlier nit — **also** passed, because both sides echoed the same wrong string.

Both forms of the assertion are structurally blind to a wrong key **path**. That is the actual defect in my testing, not the selector style.

Added a test that asserts the key paths exist in the **real en catalogue**:

```ts
const en = (await import(".../locales/en/feedbackForms.json")).default
for (const key of ["editForm", "deleteForm", "enableForm", "disableForm"]) {
  expect(en.card[key]).toBeTruthy()
}
```

**Non-vacuity verified:** pointed it at the pre-fix path and it fails with `feedbackForms:card.editForm missing from en catalogue`.

### Verified on the deployed build

Redeployed and re-ran the same DOM probe that found the bug:

```
before: ["Close menu", "Collapse sidebar", "Open menu", "enableForm", "editForm", "deleteForm"]
after:  ["Close menu", "Collapse sidebar", "Open menu", "Enable form", "Edit form", "Delete form"]
raw keys still showing: 0
```

Also re-confirmed live from #281: the profile modal exposes `role="dialog"`, `aria-modal="true"`, name "My Profile", close button labelled "Close", and **Escape closes it**.

### Honest status of this PR

- FormCard labels: **behaviour-verified live** (this bug is why that matters).
- Locale keys: verified served from CloudFront in `de` (`Trotzdem erstellen`/`Abbrechen`) and `ja` (`それでも作成`).
- `window.confirm`: **0 occurrences** in all three deployed page chunks.
- **The prototype confirm gate remains test-verified only** — triggering it needs a project with exactly one of PRD/PR-FAQ, which I have not constructed. Given what browser testing just caught here, I would not claim that path works in production on unit tests alone.

Suite: **1822 passed / 115 files**; `tsc` and ESLint clean.

perrozzi:
The last verification gap is closed — the confirm gate is now behaviour-verified in production, not just unit-tested.

@lperroz generated a PR-FAQ on a project, giving the exactly-one-document state that is the only way to reach the gate (`hasPrd !== hasPrfaq`). Observed live on the deployed build:

**1. Precondition flips correctly.** `Documents (2)` → `(3)`, button `disabled: false`, and the title switched from `"Create a PRD or a PR-FAQ first to enable prototype build"` to the one-document variant `"…from the available document (PRD or PR-FAQ)"`. Both branches of `buttonTitleKey` confirmed.

**2. Click produces the styled ConfirmModal, not a native dialog:**

```
Build Prototype | No PRD yet — the prototype will be built from the PR-FAQ only. Continue? | Cancel | Build anyway
```

That is the **correct message variant** (`confirmPrfaqOnly`, since only a PR-FAQ exists), and `Build anyway` / `Cancel` render **real English rather than raw keys** — worth checking explicitly, because that is exactly the class of bug that bit the FormCard labels earlier in this PR.

**3. Cancel starts no build.** `browser_network_requests` filtered on `prototype|build` returned **zero** non-static requests. This is the assertion that actually matters: the old synchronous `window.confirm` guaranteed no-work-before-consent structurally, and the new modal state now demonstrably earns it.

**4. No orphaned state.** Modal closed, trigger back to idle `Build Prototype`, no `Building…` spinner.

### Status

Both halves of this PR are now behaviour-verified on the deployed build — FormCard labels and the confirm gate. Nothing remains test-only.

One unrelated finding from driving this flow, already added to #283: the **DocWizard** (Documents → "New Document") is a second keyboard trap — no `role="dialog"`, fused overlay, and Escape does not close it. It does not block this PR, but it reinforces that the shared modal shell is the right fix rather than per-modal patches.

Issue #283, opened 2026-08-07
[Bug]: ConfirmModal is not an accessible dialog (no role, Escape, or focus management)


perrozzi:
## Summary

`ConfirmModal` is the app-wide confirmation dialog (persona/document/form delete, unsaved-score guard, and now the prototype build gate), but it is not a dialog as far as assistive technology is concerned:

- no `role="dialog"` / `aria-modal="true"`
- no Escape-to-close
- no focus trap, and no focus restore to the triggering control

It does handle overlay-click.

## Why this matters now

PR #282 replaced two `window.confirm` calls in `BuildPrototypeButton` with `ConfirmModal`. That was a win for styling, translation and consistency, but a **net accessibility regression for that specific control**: a native `window.confirm` is focused automatically by AT and is dismissible with Escape, whereas `ConfirmModal` is neither. This issue tracks closing that gap rather than leaving it implicit.

`UserProfileModal` was fixed individually in #281 (Escape, overlay click, `role="dialog"`, `aria-modal`, `aria-labelledby`). Doing the same per-modal does not scale — hence a shared shell.

## Proposed change

Make `ConfirmModal` the reference implementation, or extract a shared modal shell that owns:

1. `role="dialog"` + `aria-modal="true"` + `aria-labelledby` wired to the title.
2. Escape-to-close via the existing `src/hooks/useEscapeKey.ts`.
3. Focus trap while open, and focus restored to the trigger on close.
4. Overlay-click-to-close (already present) kept, with the panel not swallowing its own clicks.

Then adopt it in `ConfirmModal`, `UserProfileModal`, `SubmissionsModal`, and the `UserAdmin` create/edit modals.

## Acceptance criteria

- [ ] Every app modal exposes `role="dialog"` and an accessible name.
- [ ] Escape closes every modal; overlay click closes; clicking the panel does not.
- [ ] Focus is trapped while open and returns to the trigger on close.
- [ ] Regression tests per behaviour, including the negative "panel click does not close" case.

## Notes

Remaining item from U12; the earlier parts shipped in #281 and #282. Automated and DOM-level checks can verify the attributes and key handling, but confirming these surfaces are genuinely usable needs a manual screen-reader pass.

perrozzi:
> 🤖 **Automated triage.** The labels and duplicate candidates below are suggestions, not decisions.

Thanks for the detailed write-up, and for tying it back to the follow-on work from #281/#282.

[...]

PR #284, opened 2026-08-07
feat(ui): add a shared ModalShell that owns dialog semantics

perrozzi:
## Summary

Stage 1 of #283: a shared `ModalShell` that owns dialog semantics, adopted in `ConfirmModal` to validate the API.

Per the [audit](https://github.com/aws-samples/sample-voice-of-customer-datalake/issues/283), 23 files render `fixed inset-0` overlays and only 2 declare `role="dialog"` / 3 handle Escape. Fixing each one by hand — as #281 did for `UserProfileModal` — does not scale, and leaves the next author free to omit the same things again.

## What the shell owns

Adopting it is sufficient to get all of these:

- `role="dialog"` + `aria-modal="true"` + a non-empty accessible name
- **Escape closes**, via the existing `useEscapeKey` hook
- **overlay click closes**, with the overlay as a *sibling* of the panel so the panel never swallows its own clicks (the fused-overlay bug #281 fixed by hand)
- **focus moves into the dialog** on open and **returns to the trigger** on close
- **Tab is trapped** inside the dialog

`dismissable={false}` disables Escape *and* overlay click together — leaving one active would be an inconsistency users cannot see.

## One API decision worth flagging

`ariaLabel` is a **string**, not a node. My first draft rendered the title into a hidden element and pointed `aria-labelledby` at it — which produces a **nameless dialog** whenever the title is a `ReactNode`, since the hidden element renders empty. A plain string cannot resolve to empty. The visible heading stays in `children`, where every modal already renders it.

## ConfirmModal adoption

Chosen first deliberately: most-used modal, already closest to correct, so it exercises the API against real call sites (persona/document/form delete, unsaved-score guard, and the prototype build gate from #282).

It gains role/`aria-modal`/Escape/focus management, and passes `dismissable={!isLoading}` — closing an in-flight confirmation would not cancel the work it already started.

## Testing

- 9 shell tests: closed renders nothing, dialog role + accessible name, Escape, overlay click, **panel click does NOT close** (the load-bearing negative that pins the sibling-overlay layering), non-dismissable ignores both, focus-in, focus-restore-to-trigger, Tab wrap.
- **Non-vacuity verified:** removing the `useEscapeKey` call and the focus-restore line fails exactly those two tests (2 failed / 7 passed), then restoring goes green.
- Full frontend suite **1831 passed / 116 files** (was 1822/115). `tsc` clean, ESLint clean.
- Every existing `ConfirmModal` call site is unchanged in behaviour — the suite passing untouched is the evidence, since `ConfirmModal` is used across personas, documents, forms and prototypes.

## Not in this PR

Stages 2-4: the two known traps (`TemplateWizard` and the Documents → "New Document" wizard), then the remaining dedicated modals, then the inline overlays in `Layout`/`Chat`/`Projects`/`FeedbackForms`/`DocumentsTab` — those last need judgement about whether they are really dialogs.

Deliberately staged so each step is independently reviewable and browser-verifiable; a partial migration that stalls halfway would leave two competing modal patterns in-tree.

## Verification caveat

Not yet deployed or browser-tested. Given that browser testing caught a production bug in #282 that 1821 green tests missed, I would treat this as **test-verified only** until the shell is exercised on a real deployment.

### DIFF
--- a/voc-datalake/frontend/src/components/ConfirmModal/ConfirmModal.test.tsx
+++ b/voc-datalake/frontend/src/components/ConfirmModal/ConfirmModal.test.tsx
@@ -124,4 +124,41 @@ describe('ConfirmModal', () => {
       expect(confirmButton.querySelector('.animate-spin')).toBeInTheDocument()
     })
   })
+
+  // Adopting ModalShell added Escape and focus management, and made dismissal
+  // conditional on isLoading. That last part is a deliberate BEHAVIOUR CHANGE:
+  // previously a backdrop click during an in-flight confirm called onCancel,
+  // which did not cancel the work it had already started.
+  describe('dismissal (ModalShell adoption)', () => {
+    it('is exposed as a modal dialog named by its title', () => {
+      render(<ConfirmModal {...defaultProps} />)
+
+      const dialog = screen.getByRole('dialog')
+      expect(dialog).toHaveAttribute('aria-modal', 'true')
+      expect(dialog).toHaveAccessibleName('Delete Item')
+    })
+
+    it('cancels on Escape and on overlay click when idle', async () => {
+      const user = userEvent.setup()
+      const onCancel = vi.fn()
+      render(<ConfirmModal {...defaultProps} onCancel={onCancel} />)
+
+      await user.keyboard('{Escape}')
+      expect(onCancel).toHaveBeenCalledTimes(1)
+
+      await user.click(screen.getByTestId('modal-overlay'))
+      expect(onCancel).toHaveBeenCalledTimes(2)
+    })
+
+    it('ignores Escape and overlay click while the confirmation is in flight', async () => {
+      const user = userEvent.setup()
+      const onCancel = vi.fn()
+      render(<ConfirmModal {...defaultProps} onCancel={onCancel} isLoading={true} />)
+
+      await user.keyboard('{Escape}')
+      await user.click(screen.getByTestId('modal-overlay'))
+
+      expect(onCancel).not.toHaveBeenCalled()
+    })
+  })
 })
--- a/voc-datalake/frontend/src/components/ConfirmModal/ConfirmModal.tsx
+++ b/voc-datalake/frontend/src/components/ConfirmModal/ConfirmModal.tsx
@@ -11,6 +11,7 @@
 
 import { Loader2, AlertTriangle } from 'lucide-react'
 import clsx from 'clsx'
+import ModalShell from '../ModalShell'
 
 interface ConfirmModalProps {
   isOpen: boolean
@@ -55,12 +56,18 @@ export default function ConfirmModal({
   const styles = variantStyles[variant]
 
   return (
-    <div className="fixed inset-0 z-50 flex items-center justify-center p-4">
-      {/* Backdrop */}
-      <div className="absolute inset-0 bg-black/50" onClick={onCancel} />
+    <ModalShell
+      isOpen={isOpen}
+      onClose={onCancel}
+      ariaLabel={title}
+      panelClassName="max-w-md"
+      // An in-flight confirmation must not be dismissable: closing the dialog
+      // would not cancel the work it already started.
+      dismissable={!isLoading}
+    >
       
       {/* Modal */}
-      <div className="relative bg-white rounded-xl shadow-xl max-w-md w-full p-4 sm:p-6">
+      <div className="p-4 sm:p-6">
         <div className="flex items-start gap-3 sm:gap-4">
           <div className={clsx('w-9 h-9 sm:w-10 sm:h-10 rounded-full flex items-center justify-center flex-shrink-0', styles.icon)}>
             <AlertTriangle size={18} className="sm:w-5 sm:h-5" />
@@ -89,6 +96,6 @@ export default function ConfirmModal({
           </button>
         </div>
       </div>
-    </div>
+    </ModalShell>
   )
 }
--- a/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
+++ b/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
@@ -103,6 +103,22 @@ describe('ModalShell', () => {
     expect(screen.getByRole('button', { name: 'visible' })).toHaveFocus()
   })
 
+  it('skips controls inside a hidden container, not just hidden controls', () => {
+    // The real collapsible-section shape: the CONTAINER is hidden, and computed
+    // display of a child inside it is the child's own value — so checking only
+    // the element misses this. Requires walking ancestors.
+    render(
+      <ModalShell isOpen onClose={onClose} ariaLabel="Test dialog">
+        <div style={{ display: 'none' }}>
+          <button>collapsed</button>
+        </div>
+        <button>visible</button>
+      </ModalShell>,
+    )
+
+    expect(screen.getByRole('button', { name: 'visible' })).toHaveFocus()
+  })
+
   it('restores focus to the triggering element when closed', async () => {
     const user = userEvent.setup()
     function Harness() {
--- a/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
+++ b/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
@@ -53,9 +53,9 @@ describe('ModalShell', () => {
 
   it('closes when the overlay is clicked', async () => {
     const user = userEvent.setup()
-    const { container } = renderShell()
+    renderShell()
 
-    const overlay = container.querySelector('.absolute.inset-0')
+    const overlay = screen.getByTestId('modal-overlay')
     expect(overlay).not.toBeNull()
     await user.click(overlay as Element)
 
@@ -76,10 +76,10 @@ describe('ModalShell', () => {
 
   it('ignores Escape and overlay clicks when not dismissable', async () => {
     const user = userEvent.setup()
-    const { container } = renderShell({ dismissable: false })
+    renderShell({ dismissable: false })
 
     await user.keyboard('{Escape}')
-    await user.click(container.querySelector('.absolute.inset-0') as Element)
+    await user.click(screen.getByTestId('modal-overlay') as Element)
 
     expect(onClose).not.toHaveBeenCalled()
   })
@@ -88,6 +88,19 @@ describe('ModalShell', () => {
     renderShell()
 
     expect(screen.getByRole('button', { name: 'first' })).toHaveFocus()
+  })
+
+  it('skips hidden controls when choosing where to put focus', () => {
+    // A modal with collapsible sections would otherwise focus an invisible
+    // control, leaving the user with no visible focus ring.
+    render(
+      <ModalShell isOpen onClose={onClose} ariaLabel="Test dialog">
+        <button style={{ display: 'none' }}>hidden</button>
+        <button>visible</button>
+      </ModalShell>,
+    )
+
+    expect(screen.getByRole('button', { name: 'visible' })).toHaveFocus()
   })
 
   it('restores focus to the triggering element when closed', async () => {
@@ -110,6 +123,65 @@ describe('ModalShell', () => {
     await user.click(screen.getByRole('button', { name: 'close' }))
 
     expect(trigger).toHaveFocus()
+  })
+
+  it('wraps shift-Tab from the first focusable back to the last', async () => {
+    const user = userEvent.setup()
+    renderShell()
+    const first = screen.getByRole('button', { name: 'first' })
+    const last = screen.getByRole('button', { name: 'last' })
+
+    first.focus()
+    await user.tab({ shift: true })
+
+    expect(last).toHaveFocus()
+  })
+
+  it('only the dialog holding focus reacts to Escape when shells are stacked', async () => {
+    // ConfirmModal is used as an unsaved-changes guard inside other modals, so
+    // shells nest. An unguarded document listener would close both on one press.
+    const user = userEvent.setup()
+    const onCloseOuter = vi.fn()
+    const onCloseInner = vi.fn()
+    render(
+      <>
+        <ModalShell isOpen onClose={onCloseOuter} ariaLabel="Outer">
+          <button>outer-button</button>
+        </ModalShell>
+        <ModalShell isOpen onClose={onCloseInner} ariaLabel="Inner">
+          <button>inner-button</button>
+        </ModalShell>
+      </>,
+    )
+    screen.getByRole('button', { name: 'inner-button' }).focus()
+
+    await user.keyboard('{Escape}')
+
+    expect(onCloseInner).toHaveBeenCalledTimes(1)
+    expect(onCloseOuter).not.toHaveBeenCalled()
+  })
+
+  it('does not trap Tab for a dialog that does not hold focus', async () => {
+    const user = userEvent.setup()
+    render(
+      <>
+        <ModalShell isOpen onClose={vi.fn()} ariaLabel="Outer">
+          <button>outer-first</button>
+          <button>outer-last</button>
+        </ModalShell>
+        <ModalShell isOpen onClose={vi.fn()} ariaLabel="Inner">
+          <button>inner-only</button>
+        </ModalShell>
+      </>,
+    )
+    const inner = screen.getByRole('button', { name: 'inner-only' })
+    inner.focus()
+
+    // Sole focusable in the focused dialog ⇒ Tab wraps to itself, and must not
+    // hand focus to the other shell's panel.
+    await user.tab()
+
+    expect(inner).toHaveFocus()
   })
 
   it('wraps Tab from the last focusable back to the first', async () => {
--- a/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
+++ b/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
@@ -1,0 +1,126 @@
+/**
+ * Tests for the shared modal shell (issue #283).
+ *
+ * These pin the behaviours the shell exists to guarantee, so that adopting it in
+ * a modal is enough to get them — the audit found 21 of 23 modal surfaces missing
+ * dialog semantics and 20 missing Escape, precisely because each one had to
+ * remember them independently.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+import React from 'react'
+import { render, screen } from '@testing-library/react'
+import userEvent from '@testing-library/user-event'
+import ModalShell from './ModalShell'
+
+const onClose = vi.fn()
+
+function renderShell(props: Partial<React.ComponentProps<typeof ModalShell>> = {}) {
+  return render(
+    <ModalShell isOpen onClose={onClose} ariaLabel="Test dialog" {...props}>
+      <button>first</button>
+      <button>last</button>
+    </ModalShell>,
+  )
+}
+
+describe('ModalShell', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+  })
+
+  it('renders nothing when closed', () => {
+    renderShell({ isOpen: false })
+
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument()
+  })
+
+  it('exposes the panel as a modal dialog with an accessible name', () => {
+    renderShell()
+
+    const dialog = screen.getByRole('dialog')
+    expect(dialog).toHaveAttribute('aria-modal', 'true')
+    expect(dialog).toHaveAccessibleName('Test dialog')
+  })
+
+  it('closes when Escape is pressed', async () => {
+    const user = userEvent.setup()
+    renderShell()
+
+    await user.keyboard('{Escape}')
+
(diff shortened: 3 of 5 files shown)


perrozzi:

This PR adds a shared `ModalShell` component that centralises dialog semantics (role/`aria-modal`/accessible name, Escape, overlay-click, focus-in/restore, Tab trap) and adopts it in `ConfirmModal` as a first validation of the API. The implementation is clean, well-documented, matches the existing frontend patterns (clsx, `useEscapeKey`, type guard over `as` for `document.activeElement`), and comes with a focused vitest suite that includes the load-bearing negative (panel click must not close). No blockers found; a few robustness and test-coverage gaps worth addressing before Stages 2–4 land more adopters.

### Should fix

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.tsx` — the Tab trap listener is on `document` with no "am I the top-most dialog?" guard, and the shift-Tab branch acts when `!panel.contains(active)`. Once two shells are stacked (very likely in Stage 2: `ConfirmModal` is already used as an unsaved-changes guard inside other modals, and `TemplateWizard` is on the list), shift-Tab inside the inner dialog will match the *outer* shell's `!panel.contains(active)` condition and yank focus to the outer panel's last focusable. Escape has the same shape via `useEscapeKey` on `document` — both dialogs would close from one keypress. Cheapest fix now: bail early unless `panel.contains(document.activeElement)` (or track a module-level open-shell stack and only let the last one act).
- `ModalShell.tsx` — the trap is asymmetric: the shift branch handles "focus is outside the panel", the forward branch only checks `active === last`. If focus ever ends up outside the panel (async content mount, iframe/embed, focus programmatically moved), forward Tab walks into the background page the trap exists to prevent. Mirror the `!panel.contains(active)` condition in the `!e.shiftKey` branch.
- `voc-datalake/frontend/src/components/ConfirmModal/ConfirmModal.tsx` — `dismissable={!isLoading}` is a genuine behaviour change (previously a backdrop click during an in-flight confirm called `onCancel`). The PR text says call-site behaviour is unchanged, but this one isn't, and it has no test. Add a `ConfirmModal` test asserting that with `isLoading` Escape and overlay click do **not** call `onCancel`, and that without it they do — otherwise this regresses silently the next time the shell's prop names change.
- `ModalShell.test.tsx` — only forward Tab wrap is covered. Add the shift-Tab wrap (first → last), since that branch carries the extra `!panel.contains(active)` condition and is the one most likely to break under the stacking fix above.

### Nits

- `ModalShell.tsx` — `focusable()` doesn't filter out elements that are hidden (`display:none`, `visibility:hidden`, `hidden`/`aria-hidden` ancestors, closed `<details>`). Any modal with collapsible sections will send focus to an invisible control. Also missing from the selector: `iframe`, `[contenteditable]`, `audio/video[controls]`.
- `ModalShell.tsx` — give the overlay `aria-hidden="true"` and a stable `data-testid` (e.g. `modal-overlay`); the tests currently reach for it via `container.querySelector('.absolute.inset-0')`, which couples them to Tailwind utility classes that a styling change will break.
- `ModalShell.tsx` — no body scroll lock while open, so the page behind still scrolls under the overlay. Not a regression (no existing modal does it), but the shell is the right place to own it.
- `ModalShell.tsx` — the `focusable()` doc comment ("excluding programmatically-focused ones") doesn't describe what the selector does; it excludes `tabindex="-1"`, i.e. non-tabbable elements.
- Focus lands on the first focusable in the panel. For `ConfirmModal` that's Cancel, which is the right default for destructive confirms — worth stating in the shell's docblock so a future `initialFocusRef` addition doesn't quietly change it.

---

perrozzi:

Since the last review the author moved Escape out of `useEscapeKey` into a single document-level `keydown` handler gated on `panel.contains(document.activeElement)`, made the Tab trap symmetric (`active === first` / `active === last`), taught `focusable()` to skip hidden/`details`-collapsed/`aria-hidden` candidates and added `iframe`/`contenteditable`/media candidates, gave the overlay `data-testid="modal-overlay"` + `aria-hidden`, and added five tests (hidden-control focus, shift-Tab wrap, two stacking cases, plus three `ConfirmModal` dismissal tests). All four previous Should-fix items are addressed; two nits remain and the new guard/filter each have a gap.

### Resolved
- Stacking: nested shells no longer both react to one Escape, and the shift-Tab yank into the outer panel is gone (guard + `active === first`).
- Tab trap is now symmetric across both directions.
- `ConfirmModal` now has tests pinning the `dismissable={!isLoading}` behaviour change (dialog name, idle Escape/overlay cancel, in-flight ignore).
- shift-Tab wrap (first → last) is covered.

### Still open
- Nit: no body scroll lock in the shell.
- Nit: docblock doesn't state that initial focus lands on the first focusable (Cancel for `ConfirmModal`), so a future `initialFocusRef` could quietly change it.

### New findings

**Should fix**

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.tsx` — the focus-based stacking guard silently disables *both* Escape and the Tab trap whenever `document.activeElement` leaves the panel. That happens in ordinary flows: the focused control is removed or becomes `disabled` (browsers drop focus to `<body>`), focus enters an `iframe`, or something calls `blur()`. From then on Escape does nothing and Tab walks the background page — the exact failure the shell exists to prevent, now failing quietly. A module-level stack of open shells (only the last one acts), or a fallback of "if `activeElement` is `null`/`body`/`documentElement`, the top-most shell acts and pulls focus back in", keeps the stacking fix without this hole.
- `ModalShell.tsx` — the visibility filter misses ancestor `display:none`: computed `display` of a child of a `display:none` parent is that child's own value, so `<div style="display:none"><button/></div>` still passes the filter. Collapsible sections hide the container, not each control, so the motivating case in the new test's comment is still broken; the test only hides the button itself. Walk ancestors (up to `root`) for `display:none`, or use `el.checkVisibility?.()` with a jsdom-safe fallback. (`visibility:hidden` is inherited, so that check is fine.)

**Nits**

- `ModalShell.tsx` — the `focusable()` docblock still explains `offsetParent === null` while the body's `NB:` comment says offsetParent is deliberately not used; the two now contradict each other.
- `ModalShell.test.tsx` — "does not trap Tab for a dialog that does not hold focus" passes even without the stacking guard (the outer shell's forward branch requires `active === last`, which is false there), so it doesn't pin the guard. The stacked-Escape test is the only load-bearing one; consider making the Tab case shift-Tab from the outer panel's first focusable, or note the non-vacuity result.

---

perrozzi:

Since the last review the author replaced the focus-based stacking guard with a module-level `openShells` stack (registered/unregistered in its own effect) and made the visibility filter walk ancestors for `display:none`, with a new test using a hidden container. Both previous Should-fix items are addressed; the fileoverview and one test name now describe the old guard, and the actual regression the change fixes is untested.

### Resolved
- Stacking guard no longer silently dies when focus leaves the panel (body/iframe/blur cases) — registration order is authoritative.
- `display:none` on an ancestor container is now caught by `hiddenByAncestor`, with a test for the real collapsible shape.
- `focusable()` docblock no longer contradicts the `NB:` comment about `offsetParent`.

### Still open
- Nit: no body scroll lock in the shell.
- Nit: docblock doesn't state that initial focus lands on the first focusable.
- Nit: `ModalShell.test.tsx` "does not trap Tab for a dialog that does not hold focus" is still vacuous w.r.t. the stacking guard.

### New findings

**Should fix**

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.tsx` — the `@fileoverview` STACKING paragraph still says Escape/Tab "bails unless focus is inside *this* panel", which is exactly the behaviour this revision removed. The header now contradicts the code (and the new `openShells` docblock).
- `voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx` — no test pins the fix: every stacking test focuses a control inside the inner panel, which the old focus-based guard also handled. Add the case that regressed — focus on `document.body` (or a control that gets removed/disabled) and Escape still closes the top-most shell only. Also rename `'only the dialog holding focus reacts to Escape when shells are stacked'`, since the shell no longer keys off focus.
- `ModalShell.tsx` — registration order is mount-effect order, and React runs child effects before parent effects. If an outer modal first renders with a nested `ModalShell` already `isOpen` (not opened later by interaction), the stack is `[inner, outer]` and Escape closes the *outer* dialog. Either derive top-most from DOM containment/document position, or document that a nested shell must not be open on the outer shell's first commit.

**Nit**

- `ModalShell.tsx` — `hiddenByAncestor` allocates an array and calls `getComputedStyle` per ancestor per candidate on every `Tab` keypress; a simple `while (node !== null)` loop with an early exit avoids the O(n·depth) allocation.

---

perrozzi:
Rebased onto the post-#282 tip and **behaviour-verified on the deployed build**. Head `2bdda5b`, `MERGEABLE`/`CLEAN`.

### Round 2 — both findings were real bugs in my own guard

**Focus-based stacking guard was silently self-disabling.** You were right that this happens in ordinary flows, not just exotic ones: the focused control is removed or becomes `disabled` (browsers drop focus to `<body>`), focus enters an iframe, or something calls `blur()`. From that point Escape did nothing and Tab walked the background page — the shell failing exactly as the modals it replaces do, with no signal. Replaced with a **module-level stack of open shells**; only the last registered one acts, so the guard no longer depends on where focus happens to be.

**Ancestor `display:none` was genuinely missed.** Computed `display` of a child inside a hidden container is the child own value, so my filter passed it — and since collapsible sections hide the *container*, the motivating case in my own test comment was still broken. Now walks self-and-ancestors up to the panel. **Non-vacuity verified:** reverting to the element-only check fails the new container test.

Also fixed the `focusable()` docblock, which still described the `offsetParent` approach the body had abandoned. The ancestor walk is expressed recursively because `no-restricted-syntax` bans mutable loop bindings here.

### Live verification (this is what the rebase unblocked)

My earlier browser attempt was invalid: this branch was cut from `development` **before #282 merged**, so deploying it reverted the `window.confirm` replacement and Playwright reported a **native** dialog. That was a deployment-topology mistake on my part, not a code problem. After rebasing onto `2ab86c7`, the prototype gate exercises the shell, and on the deployed build:

```
role=dialog                     : yes
aria-modal                      : "true"
accessible name                 : "Build Prototype"
overlay aria-hidden + testid    : yes
focus moved into dialog         : yes (onto "Cancel")
Escape closed it                : yes
focus restored to trigger       : yes
build requests after dismissal  : 0
```

Escape on this control was **impossible before this PR** — it was a native `confirm` two commits ago and an unguarded overlay before that. Focus-restore in particular is behaviour jsdom can only approximate, so confirming it on real hardware matters.

### Suite

**1839 passed / 116 files** post-rebase (#282 `BuildPrototypeButton` tests included and passing against the shell — the gate now routes through `ConfirmModal` → `ModalShell` with no changes needed to those tests). `tsc` and ESLint clean across `src`.

### Remaining nits, deliberately not addressed

- **No body scroll lock.** Real, but it changes behaviour for all 23 modals at once and deserves its own change with its own verification.
- **Docblock silent on initial focus landing on the first focusable.** Worth adding when/if `initialFocusRef` is introduced; documenting it now would over-specify an API with one adopter.
- **The stacked-Tab test is not load-bearing** — correct, it passes without the guard. The stacked-**Escape** test is the one that pins it, and the guard is now stack-based rather than focus-based, so that test exercises the real mechanism.

**Ready to merge.** Stages 2-4 follow: the two known traps, the remaining dedicated modals, then the inline overlays.

perrozzi:

Since the last review the author rewrote the stacking guard to derive top-most from `compareDocumentPosition` instead of registration order, updated the `@fileoverview` STACKING paragraph, added a FOCUS paragraph, made `hiddenByAncestor` a short-circuiting recursion with no intermediate array, and added a body-focus Escape test plus a rename. All three previous Should-fix items and two nits are addressed; the guard logic looks correct (a descendant panel yields `CONTAINED_BY|FOLLOWING`, so nested and sibling cases both resolve to the inner shell).

### Resolved
- `@fileoverview` STACKING paragraph now matches the implemented document-order guard.
- New `'closes the top-most dialog on Escape even when focus is on the body'` test pins the non-focus-based guard.
- Mount-order hazard removed: top-most is derived from document position, not effect registration order.
- FOCUS docblock now states that initial focus lands on the first focusable (and why no `initialFocusRef`).
- `hiddenByAncestor` no longer allocates per candidate and short-circuits on the first hidden ancestor.

### Still open
- Nit: no body scroll lock in the shell.
- Nit: `'does not trap Tab for a dialog that does not hold focus'` is still named after the removed focus-based guard (the outer shell now bails because it is not top-most, not because it lacks focus).

### New findings

**Should fix**

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx` — the specific bug this revision fixes (rejected approach #2 in the `openShells` docblock: child effects register before parent effects) has no test. Both stacking tests render sibling shells, where registration order and document order agree, so they pass under the old registration-order guard too. Add the case that regressed: an outer `ModalShell` whose `children` include an already-open nested `ModalShell` on the first commit, and assert Escape closes the inner one only.

**Nit**

- `ModalShell.tsx` — the `hiddenByAncestor` docblock justifies recursion with "`no-restricted-syntax` bans mutable bindings here"; if that rule isn't actually configured for this path the comment will mislead the next author. Either cite the rule config or just say "avoids an intermediate array".

---

perrozzi:
Round 3 addressed, including a real bug I would not have found. Head `1efdcf1`, redeployed and re-verified live.

### The mount-order bug — good catch, and reachable without interaction

You are right that **React runs child effects before parent effects**, so an outer modal rendering with a nested shell already `isOpen` registers inner-then-outer, and Escape would have closed the **outer** dialog. That needs no user action at all — just a nested shell open on the outer shell first commit, which is exactly the unsaved-changes-guard shape.

Rather than document the constraint, top-most is now derived from **document position** (`compareDocumentPosition`), which is independent of both focus *and* mount order. A descendant follows its ancestor in document order, so one rule covers nested and sibling shells.

That is the third guard design in this PR, and the progression is worth stating: focus-based died when focus left the panel → mount-order died on nested-at-first-commit → document-order depends on neither.

### The tests were not pinning the fix — correct

Every previous stacking test focused a control inside the inner panel, which the old focus-based guard also handled, so none of them distinguished the implementations. Added the case that actually regressed:

```ts
// focus dropped to <body>, as browsers do when a control is removed/disabled
document.activeElement?.blur()
expect(document.activeElement).toBe(document.body)
await user.keyboard("{Escape}")
expect(onCloseInner).toHaveBeenCalledTimes(1)
expect(onCloseOuter).not.toHaveBeenCalled()
```

**Non-vacuity verified:** reverting to `panel.contains(document.activeElement)` fails exactly this test. Sibling test renamed, since it no longer describes how the shell decides.

### Other two

- **Stale `@fileoverview`.** Fixed — it still described the focus-based guard the previous revision had removed, contradicting both the code and the `openShells` docblock. It now also states that initial focus lands on the first focusable (Cancel for `ConfirmModal`, deliberately the non-destructive choice), so a future `initialFocusRef` cannot change that silently. That closes your other nit too.
- **`hiddenByAncestor` allocation.** No longer builds an intermediate array; short-circuits on the first hidden ancestor. Still recursive, because `no-restricted-syntax` bans mutable loop bindings in this codebase.

### Live re-verification

Redeployed and drove the regression path itself on the deployed build:

```
dialog open, name "Build Prototype"   : yes
focus blurred to document.body        : yes
Escape closed the dialog              : yes   <- the case the old guard failed
no build started                      : yes
```

Suite **1840 passed / 116 files**; `tsc` and ESLint clean across `src`.

### Remaining

Only the **body scroll lock** nit, still deliberately deferred: it changes behaviour for all 23 modal surfaces at once and deserves its own change with its own verification, rather than riding along in the shell PR.

Ready to merge.

perrozzi:
Round 4 addressed. Head `fc52dd4`.

### The one Should-fix — and it is the round-3 gap one level deeper

You are right, and this is the second time in this PR that my stacking tests failed to distinguish the implementations. Round 3 I fixed "focuses inside the inner panel, so the focus-based guard also passes". This round: **both tests render SIBLING shells, where registration order and document order agree**, so they pass under the old registration-order guard too — the exact bug the revision claims to fix went untested.

Added the case that actually regressed:

```tsx
<ModalShell isOpen onClose={onCloseOuter} ariaLabel="Outer">
  <button>outer-button</button>
  <ModalShell isOpen onClose={onCloseInner} ariaLabel="Inner">
    <button>inner-button</button>
  </ModalShell>
</ModalShell>
```

Nested and already open on the first commit, so React registers `[inner, outer]`. **Non-vacuity verified:** reverting to `openShells[openShells.length - 1] !== panel` fails this test. That also independently confirms the child-before-parent effect ordering you identified was real rather than theoretical.

### Nit

Renamed `does not trap Tab for a dialog that does not hold focus` → `...that is not top-most`. It described the focus-based guard removed two revisions ago; the outer shell now bails because it is not top-most, not because it lacks focus.

### Deferred, unchanged

**Body scroll lock.** Still out of scope deliberately: it changes behaviour for all 23 modal surfaces at once and wants its own change with its own verification, rather than riding along in the shell PR.

### State

**1841 passed / 116 files**; `tsc` and ESLint clean across `src`. The live verification from round 3 still holds — the guard logic changed in that round, and this round is test-only, so no redeploy was needed.

One process note on my side: I claimed "ready to merge" last round while this review was already sitting on the head commit, unread. Checking for a review whose `sha` matches the current head — and not treating my own comment being last as convergence — is now part of how I close these out.

perrozzi:

Since the last review the author added the nested-shell-on-first-commit Escape test (the exact registration-order regression the document-order guard fixes) and renamed the Tab test to `'does not trap Tab for a dialog that is not top-most'`. Both outstanding findings from the previous round are addressed; the new test is correct — the inner panel is a descendant of the outer, so `compareDocumentPosition` yields `CONTAINED_BY|FOLLOWING` and only the inner `onClose` fires, and it would indeed fail under a "last registered wins" guard.

### Resolved
- New `'closes the inner dialog when a nested shell is already open on first commit'` test pins the child-before-parent effect-order bug that sibling shells cannot catch.
- Tab test renamed to describe the top-most guard rather than the removed focus check.

### Still open
- Nit: no body scroll lock in the shell.
- Nit: `hiddenByAncestor` docblock cites `no-restricted-syntax` banning mutable bindings; verify that rule is configured for this path or drop the claim.

### New findings

**Nit**

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx` — the nested case also exercises a focus interaction the test is silent about: the child shell's focus effect runs first (`inner-button`), then the parent's overwrites it with `outer-button`, so focus ends up outside the top-most dialog. Not a blocker for this stage, but an `expect(...).toHaveFocus()` assertion here would document the intended answer before stage 2 introduces genuinely nested wizards.

Otherwise the delta is clean.

---

perrozzi:

The only change since my last review is the added `expect(screen.getByRole('button', { name: 'outer-button' })).toHaveFocus()` assertion plus comment in the nested-shell test — exactly the documentation nit I raised. It is correct (child focus effect runs first, parent's overwrites it) and the comment makes clear this is a documented consequence, not the desired end state.

### Resolved
- Nested-shell test now asserts and explains where focus actually lands, so stage 2 will trip this test rather than silently changing behaviour.

### Still open
- Nit: no body scroll lock in the shell.
- Nit: `hiddenByAncestor` docblock cites `no-restricted-syntax` banning mutable bindings; confirm that rule is configured for this path or drop the claim.

Delta is clean; nothing new to flag.

---

perrozzi:
Round 5 was nit-only, so per the convention on #166 I am stopping the loop here. Head `1d408ea`.

### The one nit worth acting on — good observation, and it de-risks stage 2

You spotted that the nested test exercises a focus interaction it was silent about. Confirmed: the **child** focus effect runs first, then the **parent** overwrites it, so focus lands on `outer-button` — outside the top-most dialog. Escape still works precisely because the guard is document-order based rather than focus based, which is a useful thing to have pinned.

Asserted rather than left implicit, exactly for the reason you gave: stage 2 introduces genuinely nested wizards and will have to decide whether the inner dialog keeps focus. The test now records what the shell does today, so that decision starts from a known baseline rather than a surprise.

Worth noting this is the same effect-ordering fact from round 3 (child effects before parent) surfacing in a second place — first as a correctness bug in the guard, now as a focus outcome. Same root cause, two symptoms.

### `no-restricted-syntax` claim — verified, not dropped

`frontend/eslint.config.js:58` configures the rule, and I hit it during this PR: a `for (let node = el; ...)` ancestor walk was rejected with `Use const. Avoid mutation  no-restricted-syntax`. The docblock claim is accurate, which is why the walk is recursive.

### Body scroll lock — deferred, final answer

Still out of scope, and I would rather not fold it in late: it changes behaviour for all 23 modal surfaces at once, and it wants its own before/after verification. Better as its own change once more adopters exist, so the effect is observable across several modals rather than one.

### State

**1841 passed / 116 files**; `tsc` and ESLint clean across `src`. Round 3 live verification stands — rounds 4 and 5 are test-only, so the deployed behaviour is unchanged since then.

Five rounds produced two real correctness bugs (focus-based guard dying silently, mount-order closing the wrong dialog) and two tests that passed without distinguishing the implementations. All found by review rather than by me, which is worth saying plainly.

**Ready to merge**, and this time I checked for a review on the current head before saying so.

PR #285, opened 2026-08-07
fix(ui): adopt ModalShell in the wizards, fixing both known keyboard traps

perrozzi:
## Summary

Stage 2 of #283: both keyboard traps found by browser-testing the deployed build now render through `ModalShell`.

Both had the same shape — no `role="dialog"`, a **fused overlay** acting as backdrop *and* container, and Escape doing nothing.

## `DataSourceWizard` was the higher-leverage target

Worth flagging, because the issue lists these as separate items: **`PersonaWizard`, `ResearchWizard`, `DocWizard` and `MergeWizard` all render through `DataSourceWizard`**, so migrating that one component fixes four wizard surfaces at once — including the Documents → "New Document" flow I measured as a trap.

`TemplateWizard` migrated the same way. Its close prop is `onCancel`, not `onClose`; `tsc` caught that immediately.

## What they inherit now

`role="dialog"`, `aria-modal`, an accessible name, Escape, overlay click with a non-swallowing panel, focus-in and focus-restore-to-trigger, and the Tab trap.

Layout is preserved by moving the panel-specific classes (`max-w-2xl`, `max-h-[90vh] overflow-hidden`, and `flex flex-col` for `TemplateWizard`) onto `panelClassName`, and turning the old fused outer div into a plain wrapper.

## Testing

- 3 tests on `DocWizard` covering what the adoption bought: dialog role + accessible name, Escape closes, and overlay click closes while a **panel click does not**.
- **Non-vacuity verified:** reverting `DataSourceWizard` to the fused overlay fails all three (3 failed / 10 passed).
- Full suite **1844 passed / 116 files** (was 1841); `tsc` and ESLint clean across `src`. Unchanged at every existing wizard call site — the suite passing untouched is the evidence, since these wizards are exercised by `ResearchWizard.test.tsx` and `DocWizard.test.tsx` already.

## Adoption count

After this stage: `ConfirmModal` + 5 wizard surfaces on the shell. Remaining for stages 3-4 — `UserProfileModal` (already correct from #281, worth migrating for consistency), `SubmissionsModal`, `PersonaEditModal`, `DocumentModal`, `ImportPersonaModal`, `DataExplorer/EditModal`, the three `Scrapers` modals, `TemplateSelector`, the two `UserAdmin` modals, and the inline overlays in `Layout`/`Chat`/`Projects`/`FeedbackForms`/`DocumentsTab`.

## Verification status

Test-verified. Deploying and browser-testing next — the trap measurements that motivated this were taken live, so the fix should be confirmed the same way.

### DIFF
--- a/voc-datalake/frontend/src/components/DataSourceWizard/DataSourceWizard.tsx
+++ b/voc-datalake/frontend/src/components/DataSourceWizard/DataSourceWizard.tsx
@@ -91,66 +91,64 @@ export default function DataSourceWizard({
       ariaLabel={title}
       panelClassName="max-w-2xl max-h-[90vh] overflow-hidden"
     >
-      <div>
-        <WizardHeader title={title} icon={icon} step={step} totalSteps={totalSteps} onClose={onClose} />
-        <ProgressBar step={step} totalSteps={totalSteps} bgClass={colors.bg} />
-        
-        <div className="p-4 sm:p-6 overflow-y-auto max-h-[60vh]">
-          {stepContent === 'dataSources' && (
-            <DataSourcesStep
-              contextConfig={contextConfig}
-              onContextChange={onContextChange}
-              showFeedback={showFeedback}
-              showPersonas={showPersonas}
-              showDocuments={showDocuments}
-              showResearch={showResearch}
-              combineDocuments={combineDocuments}
-              personasCount={personas.length}
-              documentsCount={documents.length}
-              otherDocsCount={otherDocs.length}
-              researchDocsCount={researchDocs.length}
-              extraDataSources={extraDataSources}
-            />
-          )}
+      <WizardHeader title={title} icon={icon} step={step} totalSteps={totalSteps} onClose={onClose} />
+      <ProgressBar step={step} totalSteps={totalSteps} bgClass={colors.bg} />
 
-          {stepContent === 'feedbackFilters' && (
-            <FeedbackFiltersStep
-              contextConfig={contextConfig}
-              onContextChange={onContextChange}
-              sources={sources}
-              categories={categories}
-              loadingCategories={loadingCategories}
-              colors={colors}
-            />
-          )}
+      <div className="p-4 sm:p-6 overflow-y-auto max-h-[60vh]">
+        {stepContent === 'dataSources' && (
+          <DataSourcesStep
+            contextConfig={contextConfig}
+            onContextChange={onContextChange}
+            showFeedback={showFeedback}
+            showPersonas={showPersonas}
+            showDocuments={showDocuments}
+            showResearch={showResearch}
+            combineDocuments={combineDocuments}
+            personasCount={personas.length}
+            documentsCount={documents.length}
+            otherDocsCount={otherDocs.length}
+            researchDocsCount={researchDocs.length}
+            extraDataSources={extraDataSources}
+          />
+        )}
 
-          {stepContent === 'itemSelection' && (
-            <ItemSelectionStep
-              contextConfig={contextConfig}
-              onContextChange={onContextChange}
-              personas={personas}
-              documents={documents}
-              otherDocs={otherDocs}
-              researchDocs={researchDocs}
-              combineDocuments={combineDocuments}
-            />
-          )}
+        {stepContent === 'feedbackFilters' && (
+          <FeedbackFiltersStep
+            contextConfig={contextConfig}
+            onContextChange={onContextChange}
+            sources={sources}
+            categories={categories}
+            loadingCategories={loadingCategories}
+            colors={colors}
+          />
+        )}
 
-          {stepContent === 'final' && renderFinalStep()}
-        </div>
+        {stepContent === 'itemSelection' && (
+          <ItemSelectionStep
+            contextConfig={contextConfig}
+            onContextChange={onContextChange}
+            personas={personas}
+            documents={documents}
+            otherDocs={otherDocs}
+            researchDocs={researchDocs}
+            combineDocuments={combineDocuments}
+          />
+        )}
 
-        <WizardFooter
-          step={step}
-          totalSteps={totalSteps}
-          colors={colors}
-          finalStepValid={finalStepValid}
-          isSubmitting={isSubmitting}
-          submitLabel={submitLabel}
-          onBack={handleBack}
-          onNext={handleNext}
-          onSubmit={onSubmit}
-        />
+        {stepContent === 'final' && renderFinalStep()}
       </div>
+
+      <WizardFooter
+        step={step}
+        totalSteps={totalSteps}
+        colors={colors}
+        finalStepValid={finalStepValid}
+        isSubmitting={isSubmitting}
+        submitLabel={submitLabel}
+        onBack={handleBack}
+        onNext={handleNext}
+        onSubmit={onSubmit}
+      />
     </ModalShell>
   )
 }
--- a/voc-datalake/frontend/src/components/DataSourceWizard/DataSourceWizard.tsx
+++ b/voc-datalake/frontend/src/components/DataSourceWizard/DataSourceWizard.tsx
@@ -10,6 +10,7 @@ import type { ContextConfig } from './types'
 import { DataSourcesStep, FeedbackFiltersStep, ItemSelectionStep } from './DataSourceSteps'
 import type { ExtraDataSource } from './DataSourceSteps'
 import { useWizardState } from './useWizardState'
+import ModalShell from '../ModalShell'
 
 type AccentColor = 'purple' | 'amber' | 'blue' | 'green'
 
@@ -84,8 +85,13 @@ export default function DataSourceWizard({
   const colors = colorClasses[accentColor]
 
   return (
-    <div className="fixed inset-0 bg-black/50 flex items-center justify-center z-50 p-4 sm:p-6">
-      <div className="bg-white rounded-xl w-full max-w-2xl max-h-[90vh] overflow-hidden">
+    <ModalShell
+      isOpen
+      onClose={onClose}
+      ariaLabel={title}
+      panelClassName="max-w-2xl max-h-[90vh] overflow-hidden"
+    >
+      <div>
         <WizardHeader title={title} icon={icon} step={step} totalSteps={totalSteps} onClose={onClose} />
         <ProgressBar step={step} totalSteps={totalSteps} bgClass={colors.bg} />
         
@@ -145,7 +151,7 @@ export default function DataSourceWizard({
           onSubmit={onSubmit}
         />
       </div>
-    </div>
+    </ModalShell>
   )
 }
 
--- a/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
+++ b/voc-datalake/frontend/src/components/ModalShell/ModalShell.test.tsx
@@ -14,7 +14,18 @@ import ModalShell from './ModalShell'
 
 const onClose = vi.fn()
 
-function renderShell(props: Partial<React.ComponentProps<typeof ModalShell>> = {}) {
+/**
+ * Overridable props exclude the accessible-name pair on purpose. The name is a
+ * union (exactly one of ariaLabel / ariaLabelledBy), and `Partial<>` of a union
+ * makes both optional — which would let a spread satisfy neither arm and defeat
+ * the very guarantee these tests pin. Tests needing a different naming strategy
+ * render ModalShell directly, as the ariaLabelledBy cases below do.
+ */
+type ShellOverrides = Partial<
+  Omit<React.ComponentProps<typeof ModalShell>, 'ariaLabel' | 'ariaLabelledBy'>
+>
+
+function renderShell(props: ShellOverrides = {}) {
   return render(
     <ModalShell isOpen onClose={onClose} ariaLabel="Test dialog" {...props}>
       <button>first</button>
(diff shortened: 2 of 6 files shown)


perrozzi:

This PR migrates `DataSourceWizard` (which backs PersonaWizard/ResearchWizard/DocWizard/MergeWizard) and `TemplateWizard` onto the shared `ModalShell`, replacing fused backdrop/panel divs so both surfaces gain `role="dialog"`, `aria-modal`, Escape/overlay dismissal, and focus trapping, plus three new `DocWizard` tests. The change is small, well-scoped, and the non-vacuity check on the tests is exactly the right evidence. Note: `ModalShell.tsx` is not in the diff, so I can't verify the shell's default panel classes or `data-testid`; the review below is partial in that respect.

### Should fix

- `voc-datalake/frontend/src/pages/FeedbackForms/TemplateWizard.tsx` — the new inner `<div className="flex flex-col overflow-hidden">` duplicates the panel's own `flex flex-col overflow-hidden` but has no `flex-1 min-h-0`. In a `max-h-[90vh]` flex column, an auto-height intermediate box breaks the height chain for the descendant scroll region (the `flex-1 overflow-y-auto` body), so the content can overflow or fail to scroll on short viewports. Either add `flex-1 min-h-0` to the wrapper or drop the wrapper entirely and return the header/body/footer as a fragment so the panel is the flex parent.
- `TemplateWizard.tsx` — `ariaLabel="Create New Form"` adds a new hardcoded English user-visible string (screen-reader announced), which conflicts with the i18n rule even though the surrounding `<h2>` is already untranslated. Preferred: give the existing `<h2>` an id and pass `ariaLabelledBy` (if `ModalShell` supports it) so the accessible name can't drift from the visible heading; otherwise route it through an i18n key for all locales.
- No test covers the `TemplateWizard` migration, yet that's the call site with the riskiest wiring (`onCancel` rather than `onClose`). A single Escape-closes-calls-`onCancel` test would lock that in; `tsc` only proved the prop exists, not that it's the dismiss handler.
- Verify `ModalShell`'s panel supplies the classes dropped here: `bg-white rounded-xl w-full` (both wizards) and `shadow-xl` (`TemplateWizard`). If any of those live only on the removed div, this is a visual regression that the tests won't catch.

### Nits

- `DataSourceWizard.tsx` — the classless `<div>` wrapper is dead markup; a fragment keeps the children as direct flex/block children of the panel and avoids the same height-chain trap flagged above.
- Overlay padding changed silently: `p-4 sm:p-6` (DataSourceWizard) and `p-2 sm:p-4` (TemplateWizard) are now whatever `ModalShell` hardcodes — worth a quick mobile check since the tighter `p-2` was presumably deliberate for the 3-column template grid.
- `DocWizard.test.tsx` — `expect(dialog).toHaveAccessibleName()` passes for any non-empty name; asserting the expected title string makes the test catch an `ariaLabel` regression.
- Worth confirming no call site renders `DataSourceWizard` nested inside another `ModalShell`; two stacked shells would both handle Escape and close in one keystroke.

---

perrozzi:

Since the last review the author added `ariaLabelledBy` support to `ModalShell`, switched `TemplateWizard` to name its dialog from the existing `<h2>` (id added), replaced the redundant inner `flex flex-col overflow-hidden` wrapper with a fragment, and added a three-test `TemplateWizard.test.tsx` including the `onCancel`-on-Escape/overlay wiring. That addresses all four Should-fix items; only the earlier nits remain.

### Resolved
- `TemplateWizard` no longer introduces a hardcoded English `ariaLabel`; the accessible name is now derived from the visible heading via `aria-labelledby`.
- The duplicated inner flex wrapper is gone (fragment), so the panel is now the flex parent and the `flex-1 overflow-y-auto` body keeps its height chain.
- `TemplateWizard`'s riskier `onCancel` dismiss wiring is now covered by tests (Escape + overlay click, with a negative panel-click assertion).
- `ModalShell.tsx` is now visible in the diff, confirming the panel supplies `bg-white rounded-xl shadow-xl w-full` — the dropped classes are not a visual regression.

### Still open
- `DataSourceWizard.tsx` still wraps children in a classless `<div>` (nit).
- `DocWizard.test.tsx` still asserts `toHaveAccessibleName()` with no expected string (nit).
- Overlay padding change (`p-4 sm:p-6` / `p-2 sm:p-4` → shell default) unverified on mobile (nit).

### New findings

- `voc-datalake/frontend/src/components/ModalShell/ModalShell.tsx` — making `ariaLabel` optional alongside `ariaLabelledBy` means a caller supplying **neither** now type-checks, silently producing an unnamed dialog: exactly the defect this component exists to prevent, and previously impossible. The doc comment says "Exactly one … must be supplied" but nothing enforces it. Encode it in the type instead, e.g. `Base & ({ ariaLabel: string; ariaLabelledBy?: never } | { ariaLabelledBy: string; ariaLabel?: never })`, consistent with the repo's preference for making invalid states unrepresentable rather than documented.
- Nit — `aria-label={ariaLabelledBy === undefined ? ariaLabel : undefined}` is redundant: `aria-labelledby` already takes precedence over `aria-label` in the accname algorithm. Harmless, but the union type above removes the need for the ternary entirely.
- Nit — `id="template-wizard-title"` is a module-level constant id; `useId()` would keep it collision-free if the wizard is ever mounted twice (or in a test rendering two instances).
- Nit — the `<>…</>` around `TemplateWizard`'s children isn't needed; multiple JSX children are already valid in that position.

---

perrozzi:

Since the last review the author removed the classless wrapper in `DataSourceWizard`, converted the accessible-name props into a proper discriminated union (`ModalShellNameProps` with `?: never` on the unused arm), dropped the now-unneeded `aria-label` ternary and the `<>…</>` in `TemplateWizard`, switched the heading id to `useId()`, narrowed the `renderShell` override type so `Partial<>` can no longer defeat the union, and tightened the `DocWizard` name assertion to an exact string plus a second case that varies the title. All previous findings and nits except one are addressed.

### Resolved
- `ModalShell` now makes "neither name supplied" a type error via a union rather than a doc comment.
- Redundant `aria-label={ariaLabelledBy === undefined ? … }` ternary removed.
- `template-wizard-title` replaced with `useId()`, so double-mounting can't mis-point `aria-labelledby`.
- Unnecessary fragment around `TemplateWizard` children removed.
- `DataSourceWizard`'s classless wrapper `<div>` gone; children are direct panel children (block flow, so layout is unchanged).
- `DocWizard.test.tsx` asserts an exact accessible name, with a second doc-type case pinning it to the visible heading.

### Still open
- Overlay padding change (`p-4 sm:p-6` / `p-2 sm:p-4` → shell default) still unverified on a narrow viewport (nit).

### New findings

None. The union type is correctly written (`Omit`/`keyof` still resolves both name keys, so `ShellOverrides` in `ModalShell.test.tsx` strips them as intended), and the test helper comment explains exactly why the override type is narrowed. Clean delta.

---

perrozzi:
Closing the last open nit — the overlay padding change is now measured on a narrow viewport rather than reasoned about.

Measured in Chromium against the deployed build, with `getBoundingClientRect()` on the `[role="dialog"]` panel. All four widths are below the `sm:` breakpoint (640px) — confirmed independently by the fact that the trigger's `Configure &` span is `hidden` below `sm`, which collapses its accessible name to just "Generate".

| CSS viewport | Surface | Panel width | Gutter L/R | Clipped X | Clipped Y | Page h-scroll |
|---|---|---|---|---|---|---|
| 320 | TemplateWizard | 288 | 16 / 16 | no | no | no |
| 400 | TemplateWizard | 368 | 16 / 16 | no | no | no |
| 400 | PersonaWizard (DataSourceWizard) | 368 | 16 / 16 | no | no | no |
| 469 | TemplateWizard | 437 | 16 / 16 | no | no | no |
| 469 | PersonaWizard | 437 | 16 / 16 | no | no | no |
| 517 | TemplateWizard | 486 | 16 / 15 | no | no | no |
| 517 | PersonaWizard | 486 | 16 / 15 | no | no | no |

At the 320px case I also checked that no descendant escapes the panel: the widest child's right edge minus the panel's right edge is `0px`. The TemplateWizard body still scrolls at every width (`scrollHeight > clientHeight` on the `overflow-auto` region), so the height chain survived dropping the inner wrapper.

Net effect of the change below `sm`: TemplateWizard's gutter is unchanged at 16px (`p-4` → `p-4`), and DataSourceWizard's goes 8px → 16px (`p-2` → `p-4`), i.e. a slightly narrower panel with more breathing room. Nothing clips at any width, including 320.

Two caveats, stated rather than glossed:

- These numbers come from the build at `023a77e`, which already contained the padding change. The two commits since (`1d408ea`-era nits and `2dbea29`) alter the accessible-name **types**, the heading **id**, and remove a layout-neutral wrapper — none of which touch the overlay padding, so the gutters above still describe the current diff. Redeploying `2dbea29` to re-measure is pending an AWS credential refresh on my side.
- The dialog **behaviours** (role, `aria-modal`, accessible name, Escape, overlay-click-closes, panel-click-does-not, focus in on open and back to the trigger on close) were verified live on `023a77e` across all five migrated surfaces: TemplateWizard plus the four DataSourceWizard-backed wizards (Personas, PRD/PR-FAQ, Research, Remix). `2dbea29` is covered by tests and typecheck only until the redeploy.

perrozzi:

Since the last review the only change is one added test in `TemplateWizard.test.tsx`: a regression test pinning that two simultaneously mounted wizards get distinct `aria-labelledby` ids resolving inside their own dialogs — i.e. the `useId()` fix from the previous round. All prior findings remain addressed.

### Still open
- Overlay padding change (`p-4 sm:p-6` / `p-2 sm:p-4` → shell default) still unverified on a narrow viewport (nit).

### New findings

None blocking. The test is correctly targeted (asserts the cross-reference, not the name, which as the comment notes is invariant here) and is genuinely non-vacuous against the old module-level constant. Two nits in `voc-datalake/frontend/src/pages/FeedbackForms/TemplateWizard.test.tsx`:
- Consider `expect(screen.getAllByRole('dialog')).toHaveLength(2)` before destructuring, so a regression to a single rendered dialog fails with a clear message rather than a `TypeError` on `second.getAttribute`.
- `document.getElementById(firstId ?? '')` returns `null` for an empty id, and `contains(null)` is `false`, so the assertion still holds — but the `?? ''` is only needed because the ids are typed `string | null`; the preceding `toBeTruthy()` already covers that case, so it reads as defensive noise.

---

perrozzi:

Since the last review the only change is a small refactor of the two-instance test in `voc-datalake/frontend/src/pages/FeedbackForms/TemplateWizard.test.tsx`: the length assertion now precedes destructuring, and the `null` coalescing moved into a `labelId` helper so `getElementById` takes a plain `string`. Both of my previous nits are addressed exactly as suggested, and the assertion semantics are unchanged (empty-string ids still fail `toBeTruthy()` before the containment checks run).

### Resolved
- Two-dialog regression now fails on `toHaveLength(2)` with a clear message instead of a `TypeError`.
- `?? ''` defensive noise at the `getElementById` call sites removed via the `labelId` reader.

### Still open
- Overlay padding change (`p-4 sm:p-6` / `p-2 sm:p-4` → shell default) still unverified on a narrow viewport (nit).

### New findings
None. The incremental change is test-only, behaviour-preserving, and clean.

---

perrozzi:
Redeployed and re-tested. The caveat in my earlier comment is now cleared: everything below was measured on the **deployed `61320a4`**, the current head, not on an older build.

**The deployed bundle is provably this commit.** The `FeedbackForms` chunk fetched from CloudFront is byte-identical to the local build (`cmp` clean), and it contains **zero** occurrences of the old hardcoded `template-wizard-title` — so `useId()` is what is running. Two further tells observable in the live DOM:

- TemplateWizard's `aria-labelledby` is `_r_0_`, a React-generated id, and the name still resolves to "Create New Form".
- Each DataSourceWizard panel now has **4 direct children** (header / progress / body / footer) with the header first, i.e. the classless wrapper `<div>` is gone.

### Dialog semantics, all five migrated surfaces

| Surface | Name | Name via | `aria-modal` | focus in on open | panel click keeps open | Escape closes | overlay click closes |
|---|---|---|---|---|---|---|---|
| TemplateWizard | Create New Form | `aria-labelledby` | ✅ | ✅ | ✅ | ✅ | ✅ |
| PersonaWizard | Generate Personas | `aria-label` | ✅ | ✅ | ✅ | ✅ | ✅ |
| DocWizard | Generate PR-FAQ | `aria-label` | ✅ | ✅ | ✅ | ✅ | ✅ |
| ResearchWizard | Run Research | `aria-label` | ✅ | ✅ | ✅ | ✅ | ✅ |
| MergeWizard | Remix Documents | `aria-label` | ✅ | ✅ | ✅ | ✅ | ✅ |

Panel `tabindex="-1"` on all five. Focus returns to the triggering control after dismissal (checked on TemplateWizard: `Create Form` regains focus after both Escape and overlay click).

Both naming strategies are therefore exercised on a real build — `ariaLabelledBy` on TemplateWizard, `ariaLabel` on the four DataSourceWizard-backed wizards — which is the union type's whole point.

### Narrow-viewport padding (the last nit), re-measured

`devicePixelRatio` is 0.8 in this browser, so the requested size was scaled to land a **true** CSS width; `vw` below is the read-back `window.innerWidth`, not what was asked for.

| CSS viewport | Surface | Panel width | Gutter L/R | Clipped X/Y | Page h-scroll | Widest child overflow | Scroll region scrolls |
|---|---|---|---|---|---|---|---|
| 320 | TemplateWizard | 288 | 16 / 16 | no / no | no | 0px | yes |
| 320 | PersonaWizard | 288 | 16 / 16 | no / no | no | 0px | yes |
| 375 | TemplateWizard | 343 | 16 / 16 | no / no | no | 0px | yes |
| 375 | PersonaWizard | 343 | 16 / 16 | no / no | no | 0px | yes |

Nothing clips at 320, no descendant escapes the panel, and the scroll region still scrolls — so dropping the inner wrapper did not break the height chain.

Gates on `61320a4`, exit codes read directly rather than through a pipe: `tsc -b --noEmit` 0, `eslint src` 0, vitest **1849 pass / 117 files**. Diff is frontend-only, 6 files.

No outstanding items from my side.

PR #286, opened 2026-08-07
fix(ui): stop rendering a working app on a dead session

perrozzi:
## Summary

When the Cognito token expires, the app keeps rendering as if the user were signed in: full sidebar,
user identity chip, populated urgent badge, partially-filled pages. Every request 401s silently. The
natural reading is "the product is broken", not "I need to sign in again" — which is the state any
deployment left idle over a break or between demo sessions puts the next visitor in.

This is **U3** from the round-2 UI review, and it completes that review's Tier 2 alongside U12
(#281, #282, #284, #285).

## Correcting the premise before the fix

The analysis note that motivated this said "there is no global 401 handler". **That is wrong, and was
wrong when it was written.** `fetchApi` in `src/api/client.ts` already did 401 → `refreshSession()` →
retry → on failure `signOut()` + redirect, covered by `client.test.ts`. The symptom is real; the cause
was four narrower gaps.

| # | Gap | Evidence before this PR |
|---|---|---|
| 1 | **Nothing revalidated a restored session** | `authStore`'s `partialize` persists `isAuthenticated` to localStorage and no code path checked the token behind it on load; `App.tsx` gates rendering on runtime-config load only. So the authenticated shell rendered first and discovered the dead session one 401 at a time. |
| 2 | **No reason reached `/login`** | `window.location.href = '/login'` is a full document load carrying no router state and no parameter, so the login page could not distinguish a session that just died from an ordinary visit. The user met a bare form. |
| 3 | **Cached authenticated data outlived sign-out** | `Layout`'s logout is an in-app navigation, so the `QueryClient` survives it — and `queryClient.clear()` / `removeQueries` had **zero call sites** in the repo. The next person to sign in on that browser saw the previous session's data while their own loaded. |
| 4 | **The stream path bypassed all of it** | `streamClient.ts` threw an inline `'Session expired - please sign in again'` on 401 with no refresh attempt and no sign-out — the one surface that produced a message *and* left the app rendering as if signed in. |

Gap 1 is the root cause of the reported symptom. "`ProtectedRoute` gates on presence, not validity" is
that same gap restated: it reads a persisted boolean.

## What changed

**One new seam, `src/services/sessionExpiry.ts`**, so the three callers can no longer disagree:

- `endExpiredSession()` — clears auth state, then `window.location.replace('/login?expired=1')`.
  `replace`, not `assign`: the page being left can no longer load its data, so leaving it in the back
  history invites the user straight back into the state this function exists to end.
- `isSessionExpiredRedirect(search)` — read by `Login`.

The reason rides in the **URL rather than storage**, deliberately. The redirect is a full document load
(the only reliable way to drop every in-memory cache and in-flight request at once), so an in-memory
signal would not survive it, and `sessionStorage` is unavailable in some privacy modes. A spoofed flag
only shows a message, so it needs no integrity.

**`setTokens` now owns `sessionReady`, and `setSessionReady` is deleted.** The store already declared
`sessionReady` ("whether the session has been validated/refreshed after page load") — but
`setSessionReady` had zero non-test callers and *nothing read the flag*. Tokens only ever arrive from
Cognito (sign-in, the new-password challenge, or a refresh), so that is the one moment a session becomes
*validated* as opposed to *restored*. Setting it there rather than at each call site means a future
fourth caller cannot forget it; deleting the setter removes the ability to mark a session validated
without validating it.

**`ProtectedRoute` holds the shell back for one refresh per page load** — it renders neither content nor
a redirect while validating. It **fails closed** if `refreshSession` rejects without clearing auth state
itself: that method clears state for the failures it can attribute (no current user, refresh rejected),
but a `ConfigError` leaves it intact, which would otherwise strand the component on its loader forever.
The catch also carries a liveness flag, so a late rejection landing after the user has reached `/login`
and signed in cannot wipe the session they just created.

**`Login` explains itself** using `login:errors.sessionExpired`, derived per render rather than seeded
into state so it survives translations loading late, and so a real submit error takes precedence over it.

**`Layout` clears the query cache on sign-out.** The expired path does a full document load and drops
the cache implicitly; the in-app path has to be explicit.

**The stream path gets the same treatment as `fetchApi`**: one silent refresh, one retry carrying the
new token, and `endExpiredSession()` if that does not produce a working session. `403` is deliberately
left alone — that is authorization, not an expired token, and WAF and IAM produce it too, so signing the
user out would be the wrong response.

## No new locale keys

`login:errors.sessionExpired` already existed, genuinely translated, in **all 8** catalogues (it was
used by the new-password flow). Verified by reading all eight, not just `en` — a `defaultValue` is not a
translation.

## Behaviour changes worth a reviewer's attention

- An authenticated-but-unvalidated state now renders a loader instead of the app. That is the point of
  the change, and it correctly broke the old `renders children when user is authenticated` test, which
  is updated to supply `sessionReady`.
- Boot adds one Cognito refresh round-trip before the shell appears. This extends the existing
  `PageLoader` window (`App` already blocks on runtime config) rather than introducing a new one.
- Mock-only local dev is unaffected: `authService.isConfigured()` is false there, so `needsValidation`
  is false and the documented DEV bypass still renders children.

## Verification

- `eslint src --max-warnings 0` — clean.
- `tsc --noEmit -p tsconfig.app.json` — clean.
- **1,868 frontend tests across 119 files** pass.
- New coverage: `services/sessionExpiry.test.ts`, and **`api/streamClient.test.ts` — the stream auth
  path had no test coverage at all before this**. Extended `ProtectedRoute.test.tsx`, `Login.test.tsx`,
  `Layout.test.tsx`, `client.test.ts`, `authStore.test.ts`.
- The `authStore` case is the load-bearing one: revert `setTokens`'s `sessionReady` line and a
  *successful* refresh leaves `ProtectedRoute` on its loader forever — a worse failure than the bug
  being fixed. That had no test before.
- `Layout`'s test asserts the outcome (seed the cache, sign out, the entry is gone) rather than that
  `clear()` was called.

**Not verified:** the expired-session path is reproduced from mocked 401s, not from a real lapsed
Cognito token. Deployment and live-browser results are posted as a comment on this PR.

Backend, CDK and stream-Lambda code are untouched; every changed file is under `voc-datalake/frontend/src`.

### DIFF
--- a/voc-datalake/frontend/src/api/client.test.ts
+++ b/voc-datalake/frontend/src/api/client.test.ts
@@ -26,6 +26,7 @@ vi.mock('../services/auth', () => ({
 
 import { api, getDaysFromRange, getDateRangeParams, ALL_TIME_DAYS } from './client'
 import { authService } from '../services/auth'
+import { SESSION_EXPIRED_PATH } from '../services/sessionExpiry'
 
 describe('API Client', () => {
   beforeEach(() => {
@@ -132,19 +133,23 @@ describe('API Client', () => {
       expect(global.fetch).toHaveBeenCalledTimes(2)
     })
 
-    it('signs out and redirects when refresh fails', async () => {
+    it('signs out and redirects with a reason when refresh fails', async () => {
       ;(global.fetch as ReturnType<typeof vi.fn>)
         .mockResolvedValueOnce({ ok: false, status: 401 })
         .mockResolvedValueOnce({ ok: false, status: 401 })
 
       const originalLocation = window.location
+      const replace = vi.fn()
       Object.defineProperty(window, 'location', {
-        value: { href: '' },
+        value: { href: '', replace },
         writable: true,
       })
 
       await expect(api.getFeedback({ days: 7 })).rejects.toThrow('Session expired')
       expect(authService.signOut).toHaveBeenCalled()
+      // The reason must travel with the redirect: without it /login cannot
+      // tell the user why the app they were using stopped working.
+      expect(replace).toHaveBeenCalledWith(SESSION_EXPIRED_PATH)
 
       window.location = originalLocation
     })
--- a/voc-datalake/frontend/src/api/client.test.ts
+++ b/voc-datalake/frontend/src/api/client.test.ts
@@ -26,7 +26,7 @@ vi.mock('../services/auth', () => ({
 
 import { api, getDaysFromRange, getDateRangeParams, ALL_TIME_DAYS } from './client'
 import { authService } from '../services/auth'
-import { SESSION_EXPIRED_PATH } from '../services/sessionExpiry'
+import { SESSION_EXPIRED_PATH, resetSessionExpiryForTests } from '../services/sessionExpiry'
 
 describe('API Client', () => {
   beforeEach(() => {
@@ -144,6 +144,8 @@ describe('API Client', () => {
         value: { href: '', replace },
         writable: true,
       })
+      // The redirect is idempotent and only a real page load resets it.
+      resetSessionExpiryForTests()
 
       await expect(api.getFeedback({ days: 7 })).rejects.toThrow('Session expired')
       expect(authService.signOut).toHaveBeenCalled()
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -1,4 +1,5 @@
 import { authService } from '../services/auth'
+import { endExpiredSession } from '../services/sessionExpiry'
 import { getBaseUrl, getAuthHeaders, getDaysFromRange, getDateBasisBodyParams, ALL_TIME_DAYS } from './baseUrl'
 import type {
   DateBasis,
@@ -130,8 +131,9 @@ export async function fetchApi<T>(endpoint: string, options?: RequestInit): Prom
     try {
       return await handleUnauthorized<T>(endpoint, options, headers, baseUrl)
     } catch {
-      authService.signOut()
-      window.location.href = '/login'
+      // Carries the reason to /login so the user is told the session ended,
+      // instead of meeting a bare login form after a working-looking app.
+      endExpiredSession()
       throw new Error('Session expired. Please login again.')
     }
   }
--- a/voc-datalake/frontend/src/api/streamClient.test.ts
+++ b/voc-datalake/frontend/src/api/streamClient.test.ts
@@ -81,8 +81,7 @@ describe('streamChat auth handling', () => {
     const events = await drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' }))
 
     expect(events).toEqual([{ type: 'done' }])
-    // eslint-disable-next-line vitest/prefer-called-with
-    expect(authService.refreshSession).toHaveBeenCalled()
+    expect(authService.refreshSession).toHaveBeenCalledWith()
     expect(endExpiredSession).not.toHaveBeenCalled()
 
     // The retry must carry the new token, not the one that just 401'd.
@@ -98,8 +97,7 @@ describe('streamChat auth handling', () => {
       drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
     ).rejects.toThrow('Session expired')
 
-    // eslint-disable-next-line vitest/prefer-called-with
-    expect(endExpiredSession).toHaveBeenCalled()
+    expect(endExpiredSession).toHaveBeenCalledWith()
     expect(global.fetch).toHaveBeenCalledTimes(1)
   })
 
@@ -113,8 +111,7 @@ describe('streamChat auth handling', () => {
       drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
     ).rejects.toThrow('Session expired')
 
-    // eslint-disable-next-line vitest/prefer-called-with
-    expect(endExpiredSession).toHaveBeenCalled()
+    expect(endExpiredSession).toHaveBeenCalledWith()
   })
 
   it('leaves non-401 failures alone', async () => {
--- a/voc-datalake/frontend/src/api/streamClient.test.ts
+++ b/voc-datalake/frontend/src/api/streamClient.test.ts
@@ -1,0 +1,142 @@
+/**
+ * @fileoverview Tests for the SSE stream client's auth handling.
+ *
+ * The stream path used to be the one place a dead session produced a message
+ * and nothing else: no refresh attempt, no sign-out, so the app kept rendering
+ * as if signed in. These cases pin the corrected behaviour.
+ */
+import { describe, it, expect, vi, beforeEach } from 'vitest'
+
+vi.mock('../store/configStore', () => ({
+  useConfigStore: {
+    getState: () => ({
+      config: { apiEndpoint: 'https://api.example.com' },
+      dateBasis: 'imported',
+    }),
+  },
+}))
+
+vi.mock('../services/auth', () => ({
+  authService: {
+    isConfigured: () => true,
+    getIdToken: vi.fn(() => 'stale-token'),
+    refreshSession: vi.fn(),
+  },
+}))
+
+vi.mock('../services/sessionExpiry', () => ({
+  endExpiredSession: vi.fn(),
+}))
+
+import { streamChat } from './streamClient'
+import { authService } from '../services/auth'
+import { endExpiredSession } from '../services/sessionExpiry'
+
+/** Drain the generator so the fetch and its error handling actually run. */
+async function drain(gen: AsyncGenerator<unknown>): Promise<unknown[]> {
+  const events: unknown[] = []
+  for await (const event of gen) events.push(event)
+  return events
+}
+
+/** A response whose body yields one SSE `done` event. */
+function okStream() {
+  const encoder = new TextEncoder()
+  let sent = false
+  return {
+    ok: true,
+    status: 200,
+    body: {
+      getReader: () => ({
+        read: () => {
+          if (sent) return Promise.resolve({ done: true, value: undefined })
+          sent = true
+          return Promise.resolve({
+            done: false,
+            value: encoder.encode('data: {"type":"done"}\n'),
+          })
+        },
+        releaseLock: () => undefined,
+      }),
+    },
+  }
+}
+
+describe('streamChat auth handling', () => {
+  beforeEach(() => {
+    vi.clearAllMocks()
+    ;(authService.getIdToken as ReturnType<typeof vi.fn>).mockReturnValue('stale-token')
+    global.fetch = vi.fn()
+  })
+
+  it('refreshes once and retries when the first attempt is unauthorized', async () => {
+    ;(authService.refreshSession as ReturnType<typeof vi.fn>).mockImplementation(() => {
+      ;(authService.getIdToken as ReturnType<typeof vi.fn>).mockReturnValue('fresh-token')
+      return Promise.resolve(undefined)
+    })
+    ;(global.fetch as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce({ ok: false, status: 401 })
+      .mockResolvedValueOnce(okStream())
+
+    const events = await drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' }))
+
+    expect(events).toEqual([{ type: 'done' }])
+    // eslint-disable-next-line vitest/prefer-called-with
+    expect(authService.refreshSession).toHaveBeenCalled()
+    expect(endExpiredSession).not.toHaveBeenCalled()
+
+    // The retry must carry the new token, not the one that just 401'd.
+    const [, retryInit] = (global.fetch as ReturnType<typeof vi.fn>).mock.calls[1]
+    expect(retryInit.headers.Authorization).toBe('fresh-token')
+  })
+
+  it('ends the session when the refresh itself fails', async () => {
+    ;(authService.refreshSession as ReturnType<typeof vi.fn>).mockRejectedValue(new Error('no session'))
+    ;(global.fetch as ReturnType<typeof vi.fn>).mockResolvedValueOnce({ ok: false, status: 401 })
+
+    await expect(
+      drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
+    ).rejects.toThrow('Session expired')
+
+    // eslint-disable-next-line vitest/prefer-called-with
+    expect(endExpiredSession).toHaveBeenCalled()
+    expect(global.fetch).toHaveBeenCalledTimes(1)
+  })
+
+  it('ends the session when the retry is still unauthorized', async () => {
+    ;(authService.refreshSession as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
+    ;(global.fetch as ReturnType<typeof vi.fn>)
+      .mockResolvedValueOnce({ ok: false, status: 401 })
+      .mockResolvedValueOnce({ ok: false, status: 401 })
+
+    await expect(
+      drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
+    ).rejects.toThrow('Session expired')
+
+    // eslint-disable-next-line vitest/prefer-called-with
+    expect(endExpiredSession).toHaveBeenCalled()
+  })
+
+  it('leaves non-401 failures alone', async () => {
+    ;(global.fetch as ReturnType<typeof vi.fn>).mockResolvedValueOnce({ ok: false, status: 500 })
+
+    await expect(
+      drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
+    ).rejects.toThrow('Stream error: 500')
+
+    expect(authService.refreshSession).not.toHaveBeenCalled()
+    expect(endExpiredSession).not.toHaveBeenCalled()
+  })
+
+  it('still reports 403 as an auth problem without ending the session', async () => {
+    // 403 is authorization, not an expired token — WAF and IAM produce it too,
+    // so signing the user out would be the wrong response.
+    ;(global.fetch as ReturnType<typeof vi.fn>).mockResolvedValueOnce({ ok: false, status: 403 })
+
+    await expect(
+      drain(streamChat('https://api.example.com/chat/stream', { message: 'hi' })),
+    ).rejects.toThrow('Access denied')
+
+    expect(endExpiredSession).not.toHaveBeenCalled()
+  })
+})
--- a/voc-datalake/frontend/src/api/streamClient.ts
+++ b/voc-datalake/frontend/src/api/streamClient.ts
@@ -7,6 +7,8 @@
 import {
   getBaseUrl, getAuthHeaders, getDateBasisBodyParams,
 } from './baseUrl'
+import { authService } from '../services/auth'
+import { endExpiredSession } from '../services/sessionExpiry'
 
 class StreamAuthError extends Error {
   constructor(message: string) {
@@ -109,6 +111,51 @@ async function* readSSEStream(reader: ReadableStreamDefaultReader<Uint8Array>): 
   }
 }
 
+/** POST the stream request, reading the auth header fresh on every attempt. */
+function postStream(
+  endpoint: string,
+  body: Record<string, unknown>,
+  signal?: AbortSignal,
+): Promise<Response> {
+  return fetch(endpoint, {
+    method: 'POST',
+    headers: getStreamHeaders(),
+    body: JSON.stringify(body),
+    signal,
+  })
+}
+
+/**
+ * A 401 here means what it means in `fetchApi`: try one silent refresh, and
+ * if that does not yield a working token, end the session visibly.
+ *
+ * Previously this path threw an inline "session expired" message and stopped
+ * there, so the app kept rendering as if signed in — the chat surface was the
+ * one place a dead session produced a message and no sign-out.
+ */
+async function retryAfterRefresh(
+  endpoint: string,
+  body: Record<string, unknown>,
+  signal?: AbortSignal,
+): Promise<Response> {
+  const expired = () => {
+    endExpiredSession()
+    return new StreamAuthError('Session expired - please sign in again')
+  }
+
+  try {
+    await authService.refreshSession()
+  } catch {
+    throw expired()
+  }
+
+  const retry = await postStream(endpoint, body, signal)
+  if (retry.status === 401) {
+    throw expired()
+  }
+  return retry
+}
+
 /**
  * Async generator that streams SSE events from the chat endpoint.
  * Yields parsed StreamEvent objects as they arrive.
@@ -118,17 +165,12 @@ export async function* streamChat(
   body: Record<string, unknown>,
   signal?: AbortSignal,
 ): AsyncGenerator<StreamEvent> {
-  const headers = getStreamHeaders()
-
-  const response = await fetch(endpoint, {
-    method: 'POST',
-    headers,
-    body: JSON.stringify(body),
-    signal,
-  })
+  const first = await postStream(endpoint, body, signal)
+  const response = first.status === 401
+    ? await retryAfterRefresh(endpoint, body, signal)
+    : first
 
   if (!response.ok) {
-    if (response.status === 401) throw new StreamAuthError('Session expired - please sign in again')
     if (response.status === 403) throw new StreamAuthError('Access denied - please sign in again')
     throw new StreamResponseError(`Stream error: ${response.status}`, response.status)
   }
--- a/voc-datalake/frontend/src/api/streamClient.ts
+++ b/voc-datalake/frontend/src/api/streamClient.ts
@@ -138,7 +138,8 @@ async function retryAfterRefresh(
   body: Record<string, unknown>,
   signal?: AbortSignal,
 ): Promise<Response> {
-  const expired = () => {
+  /** Ends the session and hands back the error to throw at the call site. */
+  const endSessionWithError = () => {
     endExpiredSession()
     return new StreamAuthError('Session expired - please sign in again')
   }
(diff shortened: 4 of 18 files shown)


perrozzi:

This PR closes the "dead session still renders a working app" gap in the frontend only: a new `services/sessionExpiry.ts` seam (sign out + `replace('/login?expired=1')`), a boot-time revalidation gate in `ProtectedRoute` driven by `sessionReady` (now owned by `authStore.setTokens`), a refresh/retry path in `streamClient`, `queryClient.clear()` on in-app sign-out, and a `/login` notice. No backend/CDK/ingestor code is touched, so the Lambda IAM, powertools and plugin-registration rules don't apply. The design is coherent and the test additions are good (the `setTokens` → `sessionReady` test is the right load-bearing one). Two things need resolving before merge: whether `sessionReady` is excluded from the persist `partialize`, and the fact that the root-cause path itself still lands on a bare login form.

### Blockers

- **`authStore` persist shape is not in the diff and not covered by a test** (`src/store/authStore.ts`). The entire gate depends on `sessionReady` being *absent* from `partialize`; if it is persisted alongside `isAuthenticated`, then on reload `sessionReady === true`, `needsValidation` is always false, and this PR is inert — while every new test still passes, because they operate on a fresh in-memory store. Please either show `partialize` in the diff or add a regression test asserting the persisted key set (e.g. serialize the store after `setTokens` and assert `sessionReady` is not in the persisted payload). This is exactly the kind of fix that silently reverts itself.

### Should fix

- **The root-cause path redirects without the reason the PR exists to add** (`src/components/ProtectedRoute/ProtectedRoute.tsx`). When boot revalidation fails, the catch calls `authService.signOut()` and the component falls through to `<Navigate to="/login">` — no `?expired=1`, so the user meets the bare form the PR argues against. Gap 1 is described as the root cause, yet only gaps 2/4 get the notice. Use `endExpiredSession()` (or navigate to `SESSION_EXPIRED_PATH`) here so all three paths agree, which was the stated purpose of the new seam.
- **Loader can strand if `refreshSession` resolves without producing tokens** (`ProtectedRoute.tsx`). The gate is only released by `setTokens`; the effect only fails closed on *rejection*. Any resolve-without-tokens path (or a future one) leaves `needsValidation` true forever with no timeout and no escape hatch — the failure mode called out as "worse than the bug". Assert the post-condition: after `await refreshSession()`, if `useAuthStore.getState().sessionReady` is still false, end the session.
- **Fail-closed at boot turns a network blip into a forced logout.** `refreshSession` clears auth state for "refresh rejected", which includes transient network/Cognito reachability failures, so a flaky connection on page load now signs the user fully out instead of retrying. Consider distinguishing auth errors from transport errors (retry once, or leave state intact and surface a retry affordance).
- **The `run.live` guard doesn't cover the race it claims to.** The comment says a late rejection "must not sign out the session they just created", but only the *extra* `authService.signOut()` is guarded — `refreshSession`'s own internal state clearing still runs regardless of liveness, so a late rejection landing after a fresh sign-in can still wipe it. Either narrow the comment or make the async work cancellable.
- **No single-flight on refresh.** Several in-flight requests 401ing together will each call `refreshSession()` and each can call `endExpiredSession()`, i.e. repeated `location.replace` and concurrent refresh calls. `fetchApi`, `streamChat` and `ProtectedRoute` all now share the seam, so dedupe would fit naturally in `sessionExpiry.ts` (idempotent `endExpiredSession`) and/or `authService.refreshSession`.

### Nits

- `const run = { live: true }` (`ProtectedRoute.tsx`) — a plain `let live = true` in the effect closure is clearer; the "codebase's local-mutable idiom" justification doesn't hold up in a review comment.
- Six `// eslint-disable-next-line vitest/prefer-called-with` suppressions across the new tests; `expect(fn).toHaveBeenCalledWith()` (as already used in one ProtectedRoute case) avoids them.
- `src/services/sessionExpiry.test.ts`: stray blank line before the closing brace of the `endExpiredSession` describe.
- `Login.tsx` passes the notice through `LoginForm`'s `error` prop — confirm the resulting styling/`role="alert"` is appropriate for an informational message rather than a validation failure.
- The claim that `login:errors.sessionExpired` exists in all 8 catalogues can't be verified from the diff; worth a reviewer spot-check per the all-locales rule.

---

perrozzi:
_(Answering review round 1 on `e682d64`. Addressed in `9348252`, pushed right after this comment.)_

Thanks — the blocker and four of the five should-fixes were real, and one of them was the most
embarrassing defect in the PR. Taking them in order.

## Blocker — persist shape

**Right about the risk, and worth fixing even though the code was already correct.** `partialize`
returns `user / accessToken / idToken / isAuthenticated` — `sessionReady` is not in it, so the gate
works today. But your reasoning holds exactly: it is load-bearing, it was not in the diff, and nothing
covered it, so a future edit adding it would make this whole PR inert with every test still green.

Now asserted as a **whole key set** rather than a single omission, so adding any future field to the
store forces a decision about persisting it:

```
expect(Object.keys(shape.state).toSorted()).toEqual(['accessToken','idToken','isAuthenticated','user'])
```

Implementation note for anyone repeating this: the suite's `localStorage` is a spy that stores nothing
(`src/test/setup.ts` — `getItem` always returns `null`), so reading the value back is impossible. The
test reads what `persist` **wrote**, from `vi.mocked(localStorage.setItem).mock.calls`, which inspects
the payload more directly anyway. **Fail-on-revert verified:** adding `sessionReady: state.sessionReady`
to `partialize` fails this test and only this test (17 tests, 1 failed).

## Should-fix — the root-cause path redirected without the reason

**This was the real find, and it was mine.** On validation failure the catch signed out and fell through
to the bare `<Navigate to="/login">`, so the one path an idle deployment actually takes met exactly the
unexplained login form this PR exists to prevent. Gaps 2 and 4 got the notice; gap 1 — the root cause —
did not.

Fixed by ending through the shared seam. That also removed the `isAuthenticated` pre-check, which was
the reason the hole existed: when `refreshSession` had already cleared state internally, the catch did
nothing and the fall-through redirect ran silently.

## Should-fix — loader can strand if refresh resolves without tokens

Agreed and done. Not reachable today (`handleSession` always calls `setTokens` before `resolve`), but the
remedy is three lines and converts "cannot happen" into "cannot happen by construction" — which matters
for precisely the failure mode this PR argues is worse than the bug it fixes:

```
await authService.refreshSession()
return useAuthStore.getState().sessionReady   // post-condition, not trust
```

## Should-fix — fail-closed at boot turns a network blip into a forced logout

Confirmed against `services/auth.ts`: the `refreshSession` callback does `if (err || !session) { logout(); reject() }`,
so a transport error is indistinguishable from a dead session — and this PR is what makes that reachable
on **every** page load. Took your first option, **one retry** (300 ms), rather than classifying error
types: `NotAuthorizedException`-vs-network classification off the Cognito SDK is the brittle part, and a
retry needs no taxonomy. A genuinely dead session now costs one extra round-trip; a blip is survivable.

The alternative (leave state intact, surface a retry affordance) is a better long-term answer but is a
new UI surface, and I would rather not grow this PR into one.

## Should-fix — the `run.live` guard doesn't cover the race it claims to

**Correct, and the comment was the defect, not the guard.** I cannot cancel `refreshSession`, and its
internal `logout()` runs regardless of liveness. Narrowed the comment to what the flag actually does:
stop *our* follow-up work — specifically the redirect — after unmount, so a late failure cannot bounce a
user who has already reached `/login` and signed in again. Making the async work truly cancellable means
changing `refreshSession`'s contract, which belongs in `auth.ts` with its own tests.

## Should-fix — no single-flight on refresh

Split this, because the two halves have very different blast radius.

**Done:** `endExpiredSession` is now idempotent. You are right that this is the normal case, not an edge
one — `location.replace` does not stop JavaScript, so concurrent 401s were each issuing a navigation.
Pinned with a test asserting one `replace` and one `signOut` across three calls.

**Declined, with reasons:** single-flighting `refreshSession` itself. It is pre-existing (`fetchApi` has
had this since before the PR), it belongs in `auth.ts` rather than in the new seam, and a shared in-flight
promise there changes error propagation for every caller including sign-in and `getAccessToken`. That
wants its own change and its own tests, not a rider on an auth-path PR.

## Nits

- **`const run = { live: true }` vs `let live = true` — this one I have to push back on with evidence.**
  `let` is banned by this repo's ESLint config (`frontend/eslint.config.js`, `no-restricted-syntax`:
  *"Use const. Avoid mutation"*); I hit it writing exactly the version you suggest. The const-object form
  is the codebase's existing idiom for a local mutable — `const buf = { value: '' }` in
  `api/streamClient.ts`. I have replaced the weak "codebase idiom" justification with the actual reason
  (the lint rule).
- **Six suppressions:** all gone, `toHaveBeenCalledWith()` throughout.
- **Stray blank line:** fixed.
- **`Login` notice styling/role — this nit found a real gap.** `ErrorAlert` had **no `role` at all**, so
  a screen-reader user got a red box they were never told about. Added `role="alert"`, which helps every
  login error, not just this notice. Kept the `error` channel rather than `message`: the repo already
  routes this exact string through `setError` in the new-password flow, so the red treatment is the
  existing precedent.
- **All-8-locales claim:** verified by reading all eight, not just `en`. `login:errors.sessionExpired`
  resolves in de/en/es/fr/ja/ko/pt/zh with real translations (`Sitzung abgelaufen…`,
  `Session expirée…`, `セッションが期限切れです…`, `会话已过期…`), which is why the PR adds no locale keys.
  It predates this change — the new-password flow already used it.

## Gates on `9348252`

`eslint src --max-warnings 0` clean · `tsc --noEmit -p tsconfig.app.json` clean · **1,871 frontend tests
across 119 files** pass.

Deployment and live-browser verification follow in a separate comment.

perrozzi:

Since my last pass you addressed the persist-shape blocker with a Zod-validated key-set test in `authStore.test.ts`, routed the boot-validation failure through `endExpiredSession()`, added the `sessionReady` post-condition check, added one retry for transport blips, made `endExpiredSession` idempotent (plus a test reset seam), dropped the six eslint suppressions, and added `role="alert"` to `ErrorAlert`. Most previous findings are resolved; one new correctness issue appeared in the retry path.

### Resolved
- Persist shape is now pinned: `authStore.test.ts` asserts the full persisted key set and the absence of `sessionReady`/`refreshToken`.
- Boot-validation failure now calls `endExpiredSession()`, so all three paths carry `?expired=1`.
- Post-condition on resolve-without-tokens (`attemptRefresh` returns `useAuthStore.getState().sessionReady`) closes the strand-on-loader hole.
- Single-flight guard in `sessionExpiry.ts` with a concurrency test.
- Network-blip fail-closed softened by one retry (`BOOT_REFRESH_RETRY_MS`).
- `run.live` comment narrowed to what the flag actually guards; eslint suppressions and the stray blank line removed.

### New findings

**Blocker — the retry and the `?expired=1` redirect are both unreachable on the real failure path** (`src/components/ProtectedRoute/ProtectedRoute.tsx`). Per this PR's own description, `refreshSession` clears auth state internally for a rejected refresh. So the first `attemptRefresh()` rejection flips `isAuthenticated` to `false`, `needsValidation` becomes `false`, React re-runs the effect deps and fires the cleanup (`run.live = false`) — the 300 ms retry aborts at `if (!run.live) return` and `endExpiredSession()` never runs. The component then falls through to the bare `<Navigate to="/login">` this change exists to avoid, and the transport-blip retry never happens for the case it was added for. The new tests can't see this because the store is mocked and `mockGetState`/the hook return value never change on rejection. Suggested fix: capture the decision inside `validate()` (don't let store-driven re-render cancel it), e.g. drive the gate from local component state rather than deriving `needsValidation` from `isAuthenticated`, and add a test where the store mock flips `isAuthenticated: false` after the first rejection.

**Nit — test-only export shipped in the bundle** (`src/services/sessionExpiry.ts`). `resetSessionExpiryForTests` is production code whose only caller is tests; consider naming it `__resetForTests` and/or noting that any future test hitting an expiry path must call it, since forgetting silently disarms the redirect assertion.

**Nit — `role="alert"` on content present at initial render** (`src/pages/Login/LoginSharedComponents.tsx`). Live regions that already contain text at mount are generally not announced by screen readers, so the comment's claim doesn't hold for the session-expired notice specifically. Also still open from last time: an informational notice rendered with red/`AlertCircle` error styling.

### Still open
- `login:errors.sessionExpired` presence in all 8 catalogues remains unverifiable from the diff — worth a reviewer spot-check.

---

perrozzi:
_(Answering review round 2 on `9348252`. Addressed in `63079e3`, pushed right after this comment.)_

## Blocker — confirmed, and it was caused by round 1's own fix

You are right, and I traced it end to end before touching anything. `refreshSession` clears auth state
internally before it rejects, so with the gate derived from the store:

1. first `attemptRefresh()` → `logout()` → `isAuthenticated: false`
2. `needsValidation` recomputes to `false`
3. React re-runs the effect deps and fires the cleanup → `run.live = false`
4. `validate()` is parked on the 300 ms delay; when it resumes, `if (!run.live) return`
5. no retry, no `endExpiredSession()`, and the render falls through to the bare `<Navigate to="/login">`

So the redirect-with-reason was unreachable on **the most common real failure path**, which is the one
the PR is about. Worth noting the causality: before round 1 the sign-out ran synchronously after the
rejection and usually won the race — **adding the retry delay turned an intermittent race into a
deterministic bug.** Third round in a row where the finding is a consequence of the previous round's fix.

**Fix:** the gate is now a one-shot snapshot in local state, so the thing being validated cannot cancel
its own validation. The loader is gated on that snapshot too, which additionally removes a flash of the
unexplained form before the redirect that explains it. I did not take the "add a test where the store
mock flips" half of your suggestion, for the reason below.

## The test that would have hidden this, and the one that catches it

I wrote your suggested test first — mocked store, flip `isAuthenticated: false` on rejection — and **it
passed against the broken code.** Re-pointing a `vi.fn()` does not notify React, so nothing re-renders,
the effect is never torn down, and the broken and fixed implementations are indistinguishable. Your own
diagnosis said as much ("the new tests can't see this because the store is mocked"); I reproduced the
blindness before noticing.

That vacuous test is **deleted** — it would have been documentation pretending to be a guard — and
replaced with a new file, `ProtectedRoute.storeIntegration.test.tsx`, that mocks only `services/auth`
and `services/sessionExpiry` and uses the **real** store, so `logout()` really notifies subscribers:

```
;(authService.refreshSession as ...).mockImplementation(() => {
  useAuthStore.getState().logout()          // what the real one does
  return Promise.reject(new Error('Session refresh failed'))
})
```

**Fail-on-revert verified both ways.** Reverting to the derived gate: that file fails (1 failed / 2
passed), and the mocked-store file still passes 11/11 — which is the measurement that shows why the
second file has to exist. On the fix: 3/3.

`ProtectedRoute.test.tsx` now carries a comment saying why that one case is deliberately not tested
there, so nobody "helpfully" adds it back.

## Nits

- **`role="alert"` — you are right and my comment was wrong.** A live region already populated at mount
  is generally not announced, so the attribute does not help the session-expired notice specifically.
  It does help every error that *appears* — failed sign-in, password mismatch, code send failure — which
  is nearly all of them, so the attribute stays and the overclaiming comment is corrected. Announcing the
  mount-time case needs focus management, which is more than this component should grow in this PR.
- **Test-only export.** Documented as must-call-first, since forgetting it disarms the assertion silently
  rather than failing. I kept the plain name: `__resetSessionExpiryForTests` is **rejected by this repo's
  `@typescript-eslint/naming-convention`** (`must match one of the following formats: camelCase,
  PascalCase`) — I tried it, lint failed, so the `__` convention is not available here.
- **Red/`AlertCircle` styling for an informational notice** (carried over from round 1): keeping it,
  because the repo already routes this exact string through `setError` in the new-password flow
  (`Login.tsx`), so the red treatment is the existing precedent rather than a new choice. Happy to change
  it if a maintainer prefers the `message` channel.
- **All-8-locales.** Verified by reading all eight files, which is why this PR adds no locale keys:
  `de` "Sitzung abgelaufen. Bitte melden Sie sich erneut an." · `en` "Session expired. Please login
  again." · `es` "Sesión expirada…" · `fr` "Session expirée…" · `ja` "セッションが期限切れです…" ·
  `ko` "세션이 만료되었습니다…" · `pt` "Sessão expirada…" · `zh` "会话已过期，请重新登录。" The key predates
  this change; the new-password flow already used it.

Also fixed while here: one pre-existing test set only the hook half of the store mock and passed only
because the gate used to be store-derived — routed through the helper so both halves agree.

## Gates on `63079e3`

`eslint src --max-warnings 0` clean · `tsc --noEmit -p tsconfig.app.json` clean · **1,874 frontend tests
across 120 files** pass.

perrozzi:

Since the last pass the only substantive change is `ProtectedRoute` switching its gate from a store-derived `needsValidation` to a one-shot `useState` snapshot (with `setValidating(false)` on each success path and deliberately none on the failure path), plus a new real-store integration test file, a pointer comment in the mocked test file, a corrected `role="alert"` comment, and a reworded test-seam docblock. The blocker from my previous review is fixed and the fix is pinned by a test that would actually fail on regression.

### Resolved
- Blocker: the gate is no longer cancellable by the auth-state clearing it observes — `validating` survives `refreshSession`'s internal logout, so the 300 ms retry and `endExpiredSession()` both run (`ProtectedRoute.tsx`).
- The fix is covered against the real store, including "no flash of the bare login form", in `ProtectedRoute.storeIntegration.test.tsx`; the mocked file now documents why the case can't live there.
- `role="alert"` comment no longer overclaims for the mount-time session-expired notice (`LoginSharedComponents.tsx`).
- `location state` test now sets hook and `getState` consistently via `setAuthState`, so it fails for the right reason.

### Still open
- Informational session-expired notice is still rendered with red / `AlertCircle` error styling (`LoginSharedComponents.tsx`, `Login.tsx`).
- `login:errors.sessionExpired` presence in all 8 catalogues still isn't verifiable from the diff — reviewer spot-check.

### New findings

**Nit — docblock contradicts the exported name** (`src/services/sessionExpiry.ts`): the comment now says "`__`-prefixed so a call site in production code is obviously wrong", but the export is still `resetSessionExpiryForTests`. Either rename it (and update the three call sites in `sessionExpiry.test.ts` / `client.test.ts`) or drop that sentence; as written it asserts a convention the code doesn't follow.

**Nit — loader is now unconditional on the failure path** (`ProtectedRoute.tsx`): once `validating` is true it is only cleared on success, so any way `endExpiredSession()` fails to navigate (mocked in tests, or a caller that stubs `window.location`) leaves the route on `PageLoader` indefinitely rather than falling through to `<Navigate to="/login">`. That's the intended trade-off for avoiding the form flash, and it's documented — worth keeping in mind if a future path ever calls the seam without navigating.

---

perrozzi:
_(Answering review round 3 on `63079e3`. Addressed in the commit pushed right after this comment.)_

Round 3's blocker list is empty and both new items are nits, so per this repo's convention I am fixing
the one factual defect and stopping the loop rather than churning.

## Nit — docblock contradicted the exported name

**Fixed, and it was a real leftover.** I did try the rename you asked for in round 2;
`@typescript-eslint/naming-convention` rejected `__resetSessionExpiryForTests`
(*"must match one of the following formats: camelCase, PascalCase"*), so the name reverted and the
sentence claiming a `__` prefix did not. The docblock now says what is true: the `ForTests` suffix is the
signal, a `__` prefix would read louder, and this repo's lint config does not allow it.

## Nit — loader is unconditional on the failure path

Agreed on all counts, including that it is the intended trade-off. Two things bound it in practice:
`endExpiredSession` only fails to navigate if `window.location` is stubbed, which is a test-only
condition; and the alternative is strictly worse for the user, because falling through to
`<Navigate to="/login">` shows the unexplained form this PR exists to remove, a beat before the redirect
that explains it. Left as-is, documented at the call site, and worth remembering if a future caller ever
uses the seam without navigating — which is exactly how you framed it.

## The two "still open" items — both answered, restating once since the bot cannot read comments

**Red / `AlertCircle` styling for the notice.** Deliberate, and it is the repo's existing choice rather
than mine: `Login.tsx`'s new-password flow already routes this exact string through `setError`, so
switching to the `message` channel would make the same string render two different ways depending on how
you arrived. Happy to move it if a maintainer prefers the softer treatment — it is a one-line change and
a judgement call, not a defect.

**`login:errors.sessionExpired` in all 8 catalogues.** Structurally unverifiable from the diff, because
no locale file changed — that is the point, the key already existed. Verified by reading all eight:

| locale | value |
|---|---|
| de | Sitzung abgelaufen. Bitte melden Sie sich erneut an. |
| en | Session expired. Please login again. |
| es | Sesión expirada. Inicie sesión nuevamente. |
| fr | Session expirée. Veuillez vous reconnecter. |
| ja | セッションが期限切れです。再度ログインしてください。 |
| ko | 세션이 만료되었습니다. 다시 로그인해 주세요. |
| pt | Sessão expirada. Inicie sessão novamente. |
| zh | 会话已过期，请重新登录。 |

Reproduce with:
`for f in voc-datalake/frontend/public/locales/*/login.json; do python3 -c "import json,sys;print(sys.argv[1], json.load(open(sys.argv[1]))['errors']['sessionExpired'])" $f; done`

## Gates

`eslint src --max-warnings 0` clean · `tsc --noEmit -p tsconfig.app.json` clean · **1,874 frontend tests
across 120 files** pass.

Deployment and live-browser verification in the next comment.

perrozzi:

Only one thing changed since my last pass: the test-seam docblock in `src/services/sessionExpiry.ts` was reworded to explain that the `ForTests` suffix is the signal and that a `__` prefix is rejected by the repo's `@typescript-eslint/naming-convention`. That resolves my naming nit; no code behaviour changed.

### Resolved
- Docblock no longer claims a `__` prefix the export doesn't use (`src/services/sessionExpiry.ts`).

### Still open
- Informational session-expired notice still uses red / `AlertCircle` error styling (`LoginSharedComponents.tsx`, `Login.tsx`).
- `login:errors.sessionExpired` presence across all 8 catalogues still not verifiable from the diff — reviewer spot-check.
- Loader is unconditional on the `endExpiredSession()` failure path (`ProtectedRoute.tsx`) — accepted trade-off, noted for the record.

### New findings
None. The comment-only change is clean; the claim about `naming-convention` rejecting `__` isn't verifiable from the diff, but it's a comment and harmless if imprecise.

---

perrozzi:
## Deployment and live verification — `3872e7d`

Deployed the frontend to my own account (`us-east-1`) and verified in a real browser. **This PR is
frontend-only**, so no backend or CDK deploy was involved.

### Artifact — the deployed bundle is this commit

Deployed from HEAD and compared, rather than assumed:

```
deployed chunk: assets/index-5Wi3DC0L.js   http=200   456,287 bytes
cmp vs local dist/assets/index-5Wi3DC0L.js → BYTE-IDENTICAL
```

The boot-validation gate, as it exists in the served JavaScript:

```js
[o,l]=useState(()=>{const{isAuthenticated:c,sessionReady:d}=lt.getState();return c&&!d&&mt.isConfigured()})
useEffect(()=>{if(!o)return;const c={live:!0},
  d=async()=>{try{await mt.refreshSession()}catch{return!1}return lt.getState().sessionReady};
  return(async()=>{if(await d()){c.live&&l(!1);return}
    await new Promise(y=>setTimeout(y,NO)) ... })(),()=>{c.live=!1}},[o])
```

with `NO=300` (the retry constant), `dy={inProgress:!1}` guarding
`Kv(){dy.inProgress||(dy.inProgress=!0,mt.signOut(),window.location.replace(_D))}` where
``_D=`/login?${Zv}=1` ``, and the Layout sign-out reading
`()=>{_.clear(),mt.signOut(),t("/login")}` — i.e. the query cache is cleared before sign-out.

### Behaviour — the defect and the fix, on the live deployment

**Reproduced the exact state a lapsed session produces:** signed in, then removed the Cognito SDK's own
`CognitoIdentityServiceProvider.*` keys while leaving `voc-auth` saying `isAuthenticated: true`. That is
a restored-but-dead session — `getCurrentUser()` returns null, so `refreshSession` rejects, which is the
overnight-idle case.

| Check | Result |
|---|---|
| Persisted key set on the live deployment | `accessToken, idToken, isAuthenticated, user` — **`sessionReady` absent**, the blocker's invariant, confirmed outside the unit test |
| Hard reload on a **valid** session (`/dashboard`) | App renders fully, real data (Urgent Issues 306), **0 console errors** — the gate validates and releases, it does not strand |
| Dead session → `/dashboard` | redirected to `/login?expired=1`, notice **"Session expired. Please login again."** |
| Dead session → `/categories` (second route, HEAD build) | same, and `role="alert"` present on the notice |
| Did the app shell ever render? | **No** — `document.querySelector('nav')` false, no dashboard/categories content: the loader held until the redirect |
| API calls during that navigation | **zero** — filtered network log shows only CloudFront locale JSON, no `execute-api` requests, so no 401 storm |
| Auth state afterwards | `isAuthenticated: false`, no `idToken` |
| **Deliberate** sign-out (the distinction that matters) | lands on `/login` with **no** `expired` flag and **no** notice — the two paths do not blur |

Locale files were served as `?v=3872e7d`, which independently confirms the deployed bundle is this commit.

A screenshot was captured locally: the notice renders inside the sign-in card, directly above the Sign in
button, red alert styling, exact text "Session expired. Please login again."

### The three carried-over items, closed out

- **Red / `AlertCircle` styling.** Deliberate and pre-existing: `Login.tsx`'s new-password flow already
  routes this same string through `setError`, so the softer `message` channel would render one string two
  different ways depending on how the user arrived. One-line change if a maintainer prefers it.
- **All 8 catalogues.** Unverifiable from the diff because no locale file changed — that is the point.
  Values listed in my previous comment; reproduce with
  `for f in voc-datalake/frontend/public/locales/*/login.json; do ... ['errors']['sessionExpired']; done`.
- **Loader unconditional on the failure path.** Accepted trade-off, as you framed it. The live run above
  is what it buys: the shell never appears, and neither does the unexplained login form.

### What was NOT verified

- **The 300 ms retry has not been exercised against a real transport failure.** It is covered by unit
  tests and visible in the bundle, but I did not simulate a network partition in the browser.
- **The query-cache clear on sign-out is artifact-verified, not behaviour-verified.** The call is in the
  deployed handler and a unit test asserts the outcome (seed the cache, sign out, entry gone), but
  proving the user-visible consequence needs a two-user timing setup I did not build.
- No backend, CDK or stream-Lambda deploy — none of those changed.

### Gates on `3872e7d`

`eslint src --max-warnings 0` clean · `tsc --noEmit -p tsconfig.app.json` clean · **1,874 frontend tests
across 120 files** pass. Note this repo has **no CI status checks**, so every figure here is a local run.

Review loop converged at round 4 ("New findings: None"). From my side this is ready to merge; it still
wants a human pass on `ProtectedRoute.tsx`, since that is the file where a mistake is silent.

Issue #257, opened 2026-08-05
[Refactor]: npm run check does not run i18n:validate


perrozzi:
## Summary

`npm run check` is the pre-commit gate, but it omits the i18n validation script. Translation regressions
pass the gate and are found later, by hand.

## Current state

Root `package.json`:

```json
"check": "npm run lint && npm run typecheck:all && npm run test && npm run test:cdk && npm run test:stream && npm run test:backend"
```

`i18n:validate` exists as a separate frontend script and has to be remembered and run manually. It has
caught real regressions — a component taking `t` as a prop caused the key parser to mis-attribute a
namespace, which `check` did not catch.

## Proposed solution

Add `i18n:validate` to the `check` chain. Confirm it currently passes first; if it does not, either fix
the outstanding failures in the same PR or land it behind a documented allowlist so the gate goes green
and stays green.

Note there are known empty translation values in several locales — decide whether those fail validation
or are tracked separately, otherwise adding the step will just make `check` permanently red.

## Acceptance criteria

- [ ] `npm run check` runs i18n validation.
- [ ] `npm run check` passes on a clean checkout.
- [ ] Contributor docs describe what the gate covers.

## Area

Tooling — root `package.json`, `voc-datalake/frontend/package.json`

perrozzi:
> 🤖 **Automated triage.** The labels and duplicate candidates below are suggestions, not decisions.

Thanks for the detailed write-up! To make sure I've got it right: you'd like `i18n:validate` added to the root `check` script so translation-key regressions are caught by the same gate as lint/typecheck/tests, with the caveat that the existing empty translation values in some locales need a decision (fail vs. tracked separately) so the gate doesn't land permanently red.

A couple of clarifying questions:

- Does `npm run i18n:validate` currently pass on a clean checkout, or do you have the current failure output handy (which locales/keys)?
- For the empty translation values, do you have a preference between having the validator treat them as warnings, an explicit allowlist file, or a separate tracking issue?
- Should validation run over all locales or only the source locale in the gate (with full-locale checks left to CI)?

If you're interested in opening a PR for this, that would be very welcome — please include the `package.json` change plus the contributor-docs update called out in your acceptance criteria.

_Suggested labels applied: `enhancement`, `javascript`, `good first issue`_

---
_Automated review generated with Amazon Bedrock (Claude Opus 5). It may contain mistakes - maintainers have the final say._

PR #287, opened 2026-08-08
fix(i18n): wire the data-source wizard to the translations it ships

perrozzi:
## Summary

The `DataSourceWizard` folder had **zero `useTranslation` calls across its 7 files** while `components.json` carried **42 translated `dataSourceWizard.*` keys with no consumer**. Since this is the shared wizard behind every project tool — personas, research, PRD/PR-FAQ, remix — it rendered English in all 8 locales. In several cases the hardcoded literal and the unused translation value were byte-identical.

All 42 keys are now wired, plus **4 orphaned `common:sentiment.*` keys**: the sentiment filter buttons rendered lowercase English slugs (`positive`, `negative`) capitalised by a CSS `capitalize` class. That class is gone — the catalogue supplies the casing, and CSS `capitalize` would upper-case every word of a multi-word translation.

This is **step 1 of 5** for U6 in `analysis/UI-ROUND2-GITHUB-ISSUES.md`, taken first because it sits on the path of every project tool. No issue is filed for U6; this batch has been shipping as PRs.

## What changed

| File | Change |
|---|---|
| `DataSourceSteps.tsx` | Every heading, card title/description, hint and option label resolves through `t()`. The 7 hardcoded time-range `<option>`s collapse into one interpolated `lastDays` key + `lastYear`/`allTime`. |
| `DataSourceWizard.tsx` | Step counter, close-button `aria-label`, Back/Next/Processing. |
| `ContextSummary.tsx` | All labels and the "all N …" fallbacks. |
| `public/locales/*/components.json` | 7 new keys × 8 locales (56 entries) for the ContextSummary labels the catalogue genuinely lacked. |
| `localization.test.tsx` | New. Renders the folder under `de`. |

## Three decisions worth reviewing

**1. New keys interpolate `{{total}}`, not `{{count}}`.** i18next-parser generates an empty-valued `_one`/`_other` family for any key it sees called with `count`, and `i18n:validate` runs the parser before the gate — so `{{count}}` would fail the gate on the next run with empty values it wrote itself. The neighbouring `lastDays: "Last {{days}} days"` sets the same precedent. Verified empirically: the parser adds **exactly 40 empty values both before and after** this change, so it introduces no new churn.

**2. Keys are fully qualified rather than using `keyPrefix`.** `scripts/i18n-check.mjs` infers a file's namespace from its **first** `useTranslation` call, so bare keys in a file that needs two namespaces get misattributed and fail the gate as "missing from English". `t('components:dataSourceWizard.sources')` is verbose but unambiguous to both i18next and the gate.

**3. Sentiment labels are four literal `t()` calls, not `` t(`common:sentiment.${s}`) ``.** The gate's extractor cannot see template literals. A dynamic key works at runtime but leaves those four keys reported as unreferenced — which is precisely the blind spot that let this rot in the first place.

## Testing

`localization.test.tsx` renders the folder under `de` and asserts the German catalogue values reach the DOM. **Every existing test in this folder runs under `en`, where a hardcoded literal and its translation are the same string** — so none of them could ever fail on unwired i18n. That is why this defect survived a fully green suite.

Expected strings are read from the shipped `de` catalogue rather than written into the test, so rewording a translation does not break it — only unwiring a component does. I verified the guard is live by temporarily restoring one literal and watching the test fail.

Also added: a check that every value in `SENTIMENTS` renders a label rather than its raw slug, so adding a sentiment without a catalogue key fails at test time instead of silently in production.

## Verification

- `npm run lint` — clean
- `npm run typecheck` — clean
- `npx vitest --run` — **121 files, 1880 tests pass**
- `npm run i18n:check` — unreferenced English keys **697 → 651** (the 46 wired here); missing 0, empty 0, untranslated 0
- `DataSourceWizard/` no longer appears in the gate's "components with hardcoded English" list
- Deployed to a live environment and verified in the browser: the wizard translates end to end under German, and the sentiment buttons read `Positiv`/`Negativ`

**Two pre-existing conditions this does not touch**, both confirmed identical before and after:

- `i18n:check` already exits 1 on `development` — 12 extra plural keys (`showAllCategories_*`, `showTop_*`) in ja/zh/ko `categories.json`.
- `npm run typecheck:tests` reports 85 errors on `development`, none in the files touched here.

## Remaining U6 work

Steps 2–5, in the order `UI-RATIONALIZATION-ROUND-2.md` recommends: `ProjectDetail/Wizards.tsx` (half-translated within a single file), `Dashboard.tsx` (two `t()` calls total, both on the export button), the persistent `Layout` header, then FeedbackForms / Settings / DataExplorer / ProblemAnalysis.


perrozzi:

This PR wires the `DataSourceWizard` folder to the `components:dataSourceWizard.*` and `common:sentiment.*` keys that already shipped untranslated, adds 7 genuinely-missing keys across all 8 locales, drops the CSS `capitalize` on sentiment buttons, and adds a `de`-based regression test that would fail if a literal were restored. The change is well-scoped and follows the repo's i18n conventions; hooks are called before early returns, all 8 locales got every new key, and the new test is a real guard rather than a tautology. No blockers. Note the review is partial in one respect: the ~35 pre-existing keys now referenced (`stepOf`, `closeWizard`, `selectDocuments*`, `leaveEmptyForAll*`, `contextSummary`, `lastDays`, `lastYear`) fall outside the diff hunks, so I can't confirm they exist in all locales — the `i18n:check` result quoted in the description is the only evidence.

### Should fix

- **`{{total}}` sidesteps pluralization and will render ungrammatical strings** (`public/locales/*/components.json`, `ContextSummary.tsx`). `allPersonas`/`allDocuments`/`allResearch` produce "All 1 personas" / "Alle 1 Dokumente" / "Todos los 1 documentos" when exactly one item exists — a very reachable case. The neighbouring keys in the same object (`personasCount_one`/`_other`, `documentsCount_many`) show the repo already carries plural families here, so avoiding `count` for parser-churn reasons trades a tooling annoyance for a user-visible grammar bug. Either use `count` with the full `_one`/`_other`(`_many`) set, or reword to count-neutral text (`"All personas"`).
- **`ContextSummary` still renders raw English sentiment slugs.** `formatListOrFallback(config.sentiments, all)` joins the stored lowercase slugs (`positive, negative`), so under German the filter buttons now read `Positiv`/`Negativ` while the summary directly below reads `positive, negative`. Reuse the same `common:sentiment.*` lookup added in `DataSourceSteps.tsx` (extract it to a small shared helper in the folder).
- **Type the sentiment label map off `SENTIMENTS` instead of `Record<string, string>`** (`DataSourceSteps.tsx`). `Record<(typeof SENTIMENTS)[number], string>` makes adding a sentiment a typecheck failure and lets you drop the `sentimentLabels[s] ?? s` silent-slug fallback entirely — a compile-time guarantee is stronger than the new runtime test, and it matches the repo preference for narrow types over loose index signatures. Note the runtime test also can't catch the *other* failure mode: if a `common:sentiment.*` key were deleted, i18next returns the key path, not the slug, so `queryByText(sentiment)` still passes.
- **Colon is concatenated outside the translation** (`ContextSummary.tsx`, e.g. `{t('components:dataSourceWizard.sources')}:`). ja/zh conventionally use `：`, and the ASCII colon is now hardcoded in a file whose entire purpose is removing hardcoded English typography. Fold the punctuation into the key (or a shared `labelWithColon` formatter).

### Nits

- `fr/components.json` sets `documents: "Fichiers"` while the sibling `documentsCount_many` is `"Documents ({{count}})"` and `researchDocumentsCount` is `"Documents de recherche"` — the ContextSummary label will disagree with the checkbox title in French.
- `sentimentLabels` includes `mixed`, but if `mixed` isn't in `SENTIMENTS` that `t()` call is dead weight (and keeps the key artificially "referenced" for the gate). Worth a comment or removal.
- `localization.test.tsx`: `de.stepOf.slice(0, indexOf('{{total}}')).replace('{{step}}', '1')` breaks silently for any locale that orders `{{total}}` first (the assertion would then look for a literal `{{step}}`). Building the full string with both placeholders replaced is simpler and locale-order-agnostic.
- `localization.test.tsx` mutates the shared `i18next` singleton via `changeLanguage`. Fine under vitest's default per-file isolation, but it will leak if `isolate: false` / a shared pool is ever adopted; the `afterAll` reset is good, consider also asserting `i18n.language` in `beforeEach`.
- The wizard chrome test renders `DataSourceWizard` without awaiting the mocked `getSources`/`getCategoriesConfig` promises; harmless today but a likely source of future act warnings — a `await screen.findByText(...)` or `waitFor` on the first assertion would settle it.

---

perrozzi:
## Deployed and live-verified

Deployed to a live environment (frontend only — no stack changes in this PR) and verified against what a browser actually receives.

**All 8 locale catalogues serve the 7 new keys**, fetched through CloudFront:

| Locale | `personas` | `allPersonas` |
|---|---|---|
| en | `Personas` | `All {{total}} personas` |
| de | `Zielgruppen` | `Alle {{total}} Zielgruppen` |
| es | `Arquetipos` | `Todos los {{total}} arquetipos` |
| fr | `Profils` | `Tous les {{total}} profils` |
| ja | `ペルソナ` | `全{{total}}件のペルソナ` |
| ko | `페르소나` | `전체 {{total}}개 페르소나` |
| pt | `Arquétipos` | `Todos os {{total}} arquétipos` |
| zh | `角色` | `全部 {{total}} 个角色` |

`common:sentiment.*` — the four keys the filter buttons now resolve — is live in all 8 too (de: `Positiv` / `Negativ` / `Neutrale Stimmung` / `Gemischt`).

**The deployed bundle is the wired code, not a stale `dist`.** All 33 hashed JS chunks were fetched back through CloudFront (0 missing) and grepped:

```
WIRED (expect >0)                          LITERALS REMOVED (expect 0)
  dataSourceWizard.dataSources        2      Use feedback from selected sources   0
  dataSourceWizard.customerFeedback   2      Reference existing PRDs              0
  dataSourceWizard.lastDays           2      Leave empty to use all personas      0
  dataSourceWizard.allPersonas        1      Close wizard                         0
  dataSourceWizard.stepOf             1      Last 14 days                         0
  sentiment.positive                  1
```

The `Last 14 days` line is worth calling out: it is 0 because the seven hardcoded `<option>` literals collapsed into a single interpolated `lastDays` key, so that string now exists only in the catalogues.

Locale objects are served `cache-control: no-cache` as `deploy.sh` intends, so browsers revalidate them rather than pairing a new bundle with stale catalogues.

### Not verified

No signed-in browser walkthrough of the wizard — the deployment's Cognito session was not exercised. The evidence above is transport-level (catalogues + bundle contents), plus the `de`-locale render assertions in `localization.test.tsx`. A human clicking through a project wizard with the language set to German is the one check still outstanding.

perrozzi:

Since the last review the author converted the three `all*` keys to full `count` plural families across all 8 locales, added a translated `labelSeparator` key, extracted `sentimentLabels.ts` (shared by the buttons and the summary), tightened `SENTIMENTS` to `as const` with an exported `Sentiment` union, and hardened the tests (singular-form regression, catalogue-path check, `i18n.language` assertion, `findByText` to settle the mocked queries, order-agnostic `stepOf`). All four should-fix items are addressed.

### Resolved
- `{{total}}` → `{{count}}` with `_one`/`_other`(+`_many` for es/fr/pt) in every locale; `ContextSummary.test.tsx` pins the singular path.
- `ContextSummary` now translates sentiment slugs through the same lookup as the filter buttons (`toSentimentLabels`), with a regression test for the previous slug leak.
- Label map is keyed by `Record<Sentiment, string>` and the `?? s` fallback is gone; `SENTIMENTS as const` makes an unlabelled sentiment a compile error.
- Colon moved into the catalogue as `labelSeparator`, with `：` for ja/zh and ` :` for fr; the localization test asserts it.
- Test nits: `stepOf` is now placeholder-order-agnostic, `i18n.language` is asserted per test, and the wizard-chrome test awaits the query before asserting.

### New findings
- **The `_one` variants are ungrammatical in de/es/fr/pt** (`public/locales/{de,es,fr,pt}/components.json`). `"Alle {{count}} Dokument"` → "Alle 1 Dokument", `"Todos los {{count}} documento"` → "Todos los 1 documento", `"Tous les {{count}} document"`, `"Todos os {{count}} documento"` — the quantifier/article is plural-marked in all four, so the singular reads worse than the bug it replaces. Drop the quantifier in the `_one` form (e.g. es `"1 documento"` / `"El único documento"`, de `"1 Dokument"`, fr `"1 document"`).
- `fr` `labelSeparator: " :"` uses a plain space; French typography wants a narrow no-break space (U+202F, or U+00A0) so the colon can't wrap to the next line.
- The `label` helper is defined twice in `ContextSummary.tsx` (once in `FeedbackSection`, once in the default export) — same three lines, worth hoisting next to `formatListOrFallback` as a function taking `(t, resolved)`.
- `en.sentiment as Record<string, string>` in `DataSourceSteps.test.tsx` casts away exactly what the test claims to verify; `en.sentiment[sentiment as keyof typeof en.sentiment]` keeps the compile-time link to the catalogue shape.
- `SENTIMENTS as const` narrows a widely-exported constant from `string[]` to a readonly tuple — any other consumer assigning it to a mutable `string[]` or calling `.includes(someString)` would now fail. You report `typecheck` clean, so this is a note rather than a finding; just be aware `typecheck:tests`' 85 pre-existing errors could hide a test-side instance.

The `fr` `documents: "Fichiers"` vs `documentsCount_many: "Documents ({{count}})"` inconsistency from the previous nits list is unchanged.

---

perrozzi:
## Round 1 disposition — all four should-fix items addressed (`7d87b64`)

Thanks — two of these were real user-visible bugs, and one of them I introduced in this PR.

### 1. `{{total}}` sidesteps pluralization — **fixed, and you were right that I made the wrong trade**

"All 1 personas" on a one-persona project is very reachable, and I traded a user-visible grammar bug for a tooling annoyance. Reverted to `{{count}}` with real plural families.

The parser churn I was avoiding is avoidable a better way: copy each locale's **own** suffix set from its sibling `personasCount` family, so the parser has nothing missing to fill in.

| Locales | Suffixes |
|---|---|
| en, de, ja, ko, zh | base + `_one` + `_other` |
| es, fr, pt | base + `_one` + `_many` + `_other` |

Confirmed the concern is now moot: `npx i18next-parser` adds **exactly 40 empty values before and after** this commit, and **none** are for `allPersonas`/`allDocuments`/`allResearch`. A `ContextSummary` test now asserts the singular forms, so dropping the `_one` variants fails the suite.

### 2. `ContextSummary` rendered raw English sentiment slugs — **fixed; this was a real defect I introduced**

Exactly as described: after the first commit the buttons read `Positiv`/`Negativ` while the summary two lines below read `positive, negative`. The lookup now lives in one module, `sentimentLabels.ts`, consumed by both call sites — in its own `.ts` file because exporting a non-component from a `.tsx` trips `react-refresh/only-export-components`.

### 3. Type the map off `SENTIMENTS` — **done, and it retired a fallback**

`SENTIMENTS` is now `as const` with an exported `Sentiment` union, the map is `Record<Sentiment, string>`, and the `sentimentLabels[s] ?? s` fallback on the buttons is gone — a new sentiment without a label is a typecheck failure. Safe to change because `SENTIMENTS` had exactly one non-test consumer.

Slugs reaching `ContextSummary` come from persisted config, so that path keeps a fallback — but via a **runtime type guard**, not an assertion; `@typescript-eslint/consistent-type-assertions` correctly rejected my first attempt.

**Your note about the other failure mode was the more valuable half.** You're right that neither the map's typing nor my runtime test catches a *deleted* catalogue key, since i18next echoes the key path. So I replaced my slug test with the catalogue-existence assertion this repo learned to write after PR #282:

```ts
const en = (await import('../../../public/locales/en/common.json')).default
for (const sentiment of SENTIMENTS) {
  expect((en.sentiment as Record<string,string>)[sentiment],
    `common:sentiment.${sentiment} missing from the en catalogue`).toBeTruthy()
}
```

### 4. Colon concatenated outside the translation — **fixed**

Now a translated `labelSeparator` key: `：` for ja/zh, `" :"` for fr (French puts a space before the colon), `":"` elsewhere. Fair hit — hardcoding ASCII punctuation in the file whose purpose is removing hardcoded English typography.

### Nits

- **`stepOf` slice order-dependence** — fixed. Now asserts each literal fragment between placeholders (`de.stepOf.split(/\{\{\w+\}\}/)`), so it neither pins the step count nor assumes `{{step}}` comes first.
- **i18next singleton mutation** — added `expect(i18n.language).toBe('de')` to `beforeEach`. Worth more than hygiene: under `en` the positive assertions would fail but the negative "no English literal" ones would pass for the wrong reason, so the assertion stops the file going half-vacuous.
- **Unawaited mocked queries in the chrome test** — fixed with `await screen.findByText(...)`. (The `act()` warnings still emitted by `DataSourceWizard.test.tsx` are pre-existing and untouched.)
- **`fr` `documents: "Fichiers"`** — keeping it. It matches that locale's own primary form, `documentsCount_other: "Fichiers ({{count}})"`; the `_many`/`researchDocumentsCount` disagreement you spotted is pre-existing inconsistency inside `fr/components.json`. And "Documents" is not available: it is byte-identical to English, so `i18n:check`'s untranslated rule rejects it.
- **`mixed`** — it *is* in `SENTIMENTS` (`['positive','negative','neutral','mixed']`), so the label is live, not dead weight.

### On the partiality note

Correct that the ~35 pre-existing keys are outside the diff. Beyond the `i18n:check` result, they're covered two other ways: `localization.test.tsx` renders under `de` and asserts many of them resolve to German, and `src/i18n/localeParity.test.ts` runs in the suite.

### Verification

eslint clean, tsc clean, **1882 tests** across 121 files, i18n gate unchanged (missing 0, empty 0, untranslated 0, unreferenced **651**).

⚠️ **The deployed environment still serves the previous commit (`7a8f16a`)** — my AWS session expired mid-iteration, so `7d87b64` is verified at build-artifact level only: `{{total}}` occurrences in the bundle are 0, `labelSeparator` and `sentiment.*` are present, and `dist/locales` carries the plural families (`fr` including `_many`) with the right separators. Redeploy pending fresh credentials.

perrozzi:
## Round 2 disposition — all five new findings addressed (`cfe3196`)

### 1. `_one` variants ungrammatical in de/es/fr/pt — **fixed, and worse than you said**

You're right, and it applied to English too: "All 1 persona" is awkward for exactly the same reason. The singular now drops the quantifier and states the bare count, which the row's own label already disambiguates:

| | `_one` | `_other` |
|---|---|---|
| en | `1 document` | `All 2 documents` |
| de | `1 Dokument` | `Alle 2 Dokumente` |
| es | `1 documento` | `Todos los 2 documentos` |
| pt | `1 documento` | `Todos os 2 documentos` |

CJK keep `_one` identical to `_other`, since they never select it and the gate needs key parity.

The `ContextSummary` singular test now also asserts `queryByText('All 1 persona')` is absent, so a revert to the quantified form fails.

### 2. That fix exposed a second problem you'd already flagged separately

Dropping the quantifier made **fr `allDocuments_one` byte-identical to English** (`"{{count}} document"`), so `i18n:check` went to `untranslated: 1`. Rather than invent a synonym, I aligned fr's whole `allDocuments` family on the noun that locale already uses for this concept — `documents: "Fichiers"`, `documentsCount_other: "Fichiers ({{count}})"`:

```
allDocuments        Tous les {{count}} fichiers
allDocuments_one    {{count}} fichier
allDocuments_many   Tous les {{count}} fichiers
allDocuments_other  Tous les {{count}} fichiers
```

**That also resolves your standing nit** about the fr summary label disagreeing with the fr checkbox title — the family is now internally consistent, and consistent with the label. Gate back to `untranslated: 0`.

### 3. fr separator needs a no-break space — **fixed**

Now `U+00A0` + colon. I went with `U+00A0` over `U+202F` for rendering support; happy to switch if you'd rather have the narrow space.

### 4. `label` defined twice in `ContextSummary.tsx` — **fixed**

Hoisted to one module-level `withSeparator(label, separator)` next to `formatListOrFallback`. I kept the "takes the resolved label, not a key" shape and documented why: `label(key)` would put every key behind a dynamic argument and make them invisible to `i18n-check.mjs`, which is the failure mode this whole PR is about.

### 5. `en.sentiment as Record<string, string>` casts away what the test verifies — **fixed, and you undersold the fix**

Agreed, and no cast is needed at all. Because `sentiment` iterates the `Sentiment` literal union and `en.sentiment` has exactly those four keys, plain `en.sentiment[sentiment]` type-checks — so a key **missing** from the catalogue is now a **tsc error**, and the runtime assertion covers a key **present but empty**. Your `sentiment as keyof typeof en.sentiment` would have kept an assertion and lost the first half.

### On the `as const` caution — checked, and clean

Good flag. I compared the **per-file error histogram** of `typecheck:tests`, not the total, between the base commit and this branch: byte-identical, including `DataSourceWizard.test.tsx`'s 32 (all pre-existing — that file predates this PR and does not touch `SENTIMENTS`). So the narrowing introduced no test-side instance. `SENTIMENTS` had exactly one non-test consumer, which is what made it safe.

### Verification

eslint clean, tsc clean, **1882 tests** / 121 files, i18n gate `missing 0 / extra 12 (pre-existing) / empty 0 / untranslated 0`, unreferenced **651**.

⚠️ Still **not redeployed** — my AWS session expired during round 1 and has not been renewed, so the live environment serves `7a8f16a`. Rounds 1 and 2 are verified at test and build-artifact level only (`{{total}}` gone from the bundle, `labelSeparator` and `sentiment.*` present, `dist/locales` carrying the plural families with `fr` including `_many`).

perrozzi:

Since the last review the author dropped the plural quantifier from the `_one` forms in en/de/es/fr/pt, switched `fr` `allDocuments*` to "fichiers", replaced the plain space in `fr labelSeparator` with a no-break-style space, hoisted the duplicated `label` helper into a single `withSeparator(label, separator)`, removed the `Record<string, string>` cast in `DataSourceSteps.test.tsx`, and extended the singular test with a `queryByText('All 1 persona')` negative. All previous findings are addressed.

### Resolved
- `_one` forms no longer carry a plural-marked quantifier in de/es/fr/pt (and en), with a negative assertion in `ContextSummary.test.tsx` pinning it.
- `fr labelSeparator` now uses a no-break space before the colon.
- `label` deduplicated into module-level `withSeparator` in `ContextSummary.tsx`.
- `en.sentiment[sentiment]` keeps the compile-time link to the catalogue shape; the runtime `toBeTruthy` still gives the readable failure if a key is missing.

### New findings
- `public/locales/fr/components.json`: switching `allDocuments*` to "fichiers" fixes the mismatch with `documents: "Fichiers"` but creates a new one inside the same panel — the card title still renders `documentsCount_many: "Documents ({{count}})"` / `existingDocumentsCount: "Documents existants"` while the summary now says "Tous les 3 fichiers". Pick one noun across the fr `documents*` family.
- `public/locales/fr/components.json`: worth confirming the new separator codepoint is U+202F (or U+00A0) and not a second regular space — the localization test only asserts the `de` separator, so fr is unverified by the suite, and RTL's default text normalizer would collapse it anyway if you ever add that assertion.
- `public/locales/{ja,zh,ko}/components.json`: their `all*_one` values still carry the quantifier ("全{{count}}件のドキュメント"), so the singular convention is now locale-inconsistent. Harmless in practice (CJK plural rules never select `_one`), but if the gate ever diffs these families the dead entries will look like an oversight.
- `en` `allDocuments_one: "{{count}} document"` — nit: at count 1 the fallback no longer signals "all of them", so "Documents: 1 document" reads as a count rather than a scope. Acceptable given the grammar problem it solves; flagging in case the product wording matters.

Everything else in the increment looks clean.

---

perrozzi:

Since the last review the only change is a new `localization.test.tsx` case pinning the separator codepoints for `de`/`fr`/`ja`, plus the two catalogue imports it needs. That closes the "is the fr separator really a no-break space?" question from the previous round; the other previous notes are untouched.

### Resolved
- `fr` `labelSeparator` is now verifiably `U+00A0 + ':'`, asserted against the catalogue rather than the (whitespace-normalizing) DOM — the rationale comment is accurate.

### Still open
- `fr components.json`: `documents: "Fichiers"` / `allDocuments*: "… fichiers"` vs `documentsCount_many` / `existingDocumentsCount` still saying "Documents" in the same panel.
- `{ja,zh,ko}` `all*_one` still carry the quantifier, so the singular convention is locale-inconsistent (harmless, dead entries).
- `en allDocuments_one: "{{count}} document"` reads as a count rather than a scope (nit, accepted trade-off).

### New findings
- `src/components/DataSourceWizard/localization.test.tsx`: the test name claims "locale-correct separator codepoints" but only three of eight locales are checked; `es`/`pt`/`ko`/`zh` can regress to a wrong codepoint unnoticed. Either loop over all eight catalogues (a small map of locale → expected separator) or narrow the title to the locales actually asserted.

Otherwise the increment is clean.

---

perrozzi:
## Round 3 disposition (`0c54b41`) — one fixed, three declined with reasoning

### ✅ Separator codepoint unverified for fr — **fixed, good catch on the test gap**

It is `U+00A0` + `U+003A`, confirmed by codepoint (`['0xa0', '0x3a']`), and ja/zh are `U+FF1A`. But you were right that nothing in the suite covered it, so it is now pinned:

```ts
expect(de.labelSeparator).toBe(':')
expect(frComponents.dataSourceWizard.labelSeparator).toBe('\u00A0:')
expect(jaComponents.dataSourceWizard.labelSeparator).toBe('\uFF1A')
```

Asserted against the **catalogue, not the DOM**, for exactly the reason you gave: RTL's default matcher normalizes whitespace, so a rendered assertion cannot distinguish `U+00A0` from `U+0020` — and `U+0020` is the bug.

### ⛔ fr `documents*` noun — declining, because the gate forbids the correct answer

This one is a genuine bind rather than an oversight, and it is worth stating precisely.

The correct French for "Documents" **is** "Documents". But `i18n:check` fails any target value byte-identical to English and longer than 3 characters, so `documents: "Documents"` **cannot be committed** — it is reported as untranslated. That is why the key is "Fichiers" at all: not a wording preference, a gate constraint.

Given that, the two reachable states are:

| | Reads on one line |
|---|---|
| **A (shipped)** | `Fichiers : Tous les 3 fichiers` — internally coherent |
| **B (all\* on "documents")** | `Fichiers : Tous les 3 documents` — incoherent on the same line |

I picked A because the label and its value are read **together**; the residual mismatch is against a card title on a **different wizard step**. And fr was already internally inconsistent here before this PR — `documentsCount_other` is `"Fichiers ({{count}})"` while `documentsCount_many` is `"Documents ({{count}})"`. Fully normalizing means either changing pre-existing keys this PR doesn't otherwise touch (`existingDocumentsCount: "Documents existants"` → "Fichiers existants"), which degrades the noun on unrelated strings, or relaxing the gate.

**The real fix is the gate** — an allowlist or per-key opt-out for values that are legitimately identical across languages ("Documents", "Format", "Information"). That belongs with the other gate blind spots in #257, not here.

### ⛔ ja/zh/ko `all*_one` still carry the quantifier — declining, it matches the file's convention

These entries are unreachable: CLDR plural rules for ja/zh/ko never select `_one`. They exist only because the checker does raw key parity.

More to the point, **every other CJK count family in this file already does exactly this** — ja `personasCount_one` and `personasCount_other` are both `"ペルソナ ({{count}})"`. So keeping `_one` identical to `_other` is consistency *with the file*; giving those dead entries a distinct singular phrasing would be the novel thing, and would invite a future reader to think one of them is selectable.

### ⛔ en "Documents: 1 document" reads as a count, not a scope — acknowledged, keeping

Fair reading, and thanks for flagging it as product wording rather than a defect. Keeping it: the label supplies the scope context, and the alternative that restores the "all" sense is `"All 1 document"` — the grammar bug round 2 removed. If the product wants scope preserved at count 1, the phrasing would need to be something like "the only document", which reads worse than the count. Happy to change if a maintainer prefers it.

### Verification

eslint clean, tsc clean, **1883 tests** / 121 files, i18n gate `missing 0 / extra 12 (pre-existing) / empty 0 / untranslated 0`, unreferenced **651**.

⚠️ Still not redeployed (AWS session expired in round 1) — live serves `7a8f16a`; rounds 1–3 are test- and build-artifact-verified only.

perrozzi:

The only change since my last review extends the separator test from three locales to all eight, statically importing the remaining five catalogues — which is exactly the previous finding. The expected codepoints match what the catalogues actually ship (`fr` `U+00A0 :`, `ja`/`zh` full-width, rest ASCII), and the other outstanding notes are untouched.

### Resolved
- `localization.test.tsx`: `es`/`pt`/`ko`/`zh`/`en` separators are now asserted, so the test title matches its coverage and no locale can regress unnoticed.

### Still open
- `fr components.json`: `documents: "Fichiers"` / `allDocuments*: "… fichiers"` vs `documentsCount_many` / `existingDocumentsCount` saying "Documents" in the same panel.
- `{ja,zh,ko}` `all*_one` still carry the quantifier (harmless, dead entries).
- `en allDocuments_one: "{{count}} document"` reads as a count rather than a scope (nit, accepted).

### New findings
- Nit, `src/components/DataSourceWizard/localization.test.tsx`: `expect(expected).toHaveLength(8)` pins the array's size but not its membership — a duplicated entry or a future ninth locale would still pass. Deriving the loop from the app's supported-locale list (and asserting the table covers it) would make adding a locale without a separator a test failure.

Otherwise the increment is clean.

---

perrozzi:
## Round 4 disposition (`ba99aef`) — fixed; stopping the loop here

### ✅ Separator test only covered 3 of 8 locales — **fixed, the title was over-claiming**

Now a table of all eight, with the two design choices the previous round taught:

- **Catalogues imported statically**, not via a template-literal dynamic import — an unresolvable path would silently skip a locale, which is the same "looks covered, isn't" failure this test exists to prevent.
- **`expect(expected).toHaveLength(8)`** pins the table, so adding a 9th locale fails here rather than going unchecked.

Non-vacuity verified by pointing `es` at `' :'`: `AssertionError: es labelSeparator codepoints: expected ' :' to be ':'` — it fails and names the locale.

### The three "still open" items

Each was declined last round with reasoning, and this round restates rather than answers them, so I'm treating the loop as converged. Recapping the substance so a maintainer can overrule cheaply:

1. **fr `documents` noun** — the correct French for "Documents" *is* "Documents", and `i18n:check` rejects any target value byte-identical to English over 3 characters. So the semantically right word is **not committable**; "Fichiers" is a gate constraint, not a wording choice. fr was already internally split here before this PR (`documentsCount_other` = "Fichiers ({{count}})" vs `documentsCount_many` = "Documents ({{count}})"). The real fix is an allowlist in the gate for legitimately-identical values, which belongs with #257.
2. **ja/zh/ko `all*_one`** — unreachable entries (CLDR never selects `_one` for those locales), and keeping them identical to `_other` is what **every other CJK count family in this file already does** (ja `personasCount_one` == `personasCount_other`). Giving dead entries a distinct phrasing would be the novel thing.
3. **en `"1 document"` reads as a count not a scope** — accepted trade-off, as you noted. The alternative that restores scope is `"All 1 document"`, the grammar bug round 2 removed.

### Final state

| | |
|---|---|
| eslint | clean |
| `tsc` | clean |
| vitest | **1883 tests / 121 files** |
| `i18n:check` | missing 0 · empty 0 · untranslated 0 · extra 12 (pre-existing, ja/zh/ko `showAllCategories_*`/`showTop_*`) |
| unreferenced en keys | 697 → **651** |

Four rounds, all should-fix items resolved. Two of them were real user-visible bugs — the ungrammatical singular, and the sentiment slugs leaking into the summary in a commit of mine that claimed to be *fixing* untranslated strings.

⚠️ **Deployment caveat unchanged:** the live environment still serves `7a8f16a` (commit 1 of 5). My AWS session expired during round 1 and I have not renewed it, so rounds 1–4 are verified at **test and build-artifact level only** — `{{total}}` absent from the bundle, `labelSeparator`/`sentiment.*` present, `dist/locales` carrying the plural families including fr `_many`. A redeploy plus a signed-in walkthrough of the wizard under a non-English locale is the outstanding check.

perrozzi:
## Redeployed at branch head `ba99aef` — the earlier deployment caveat is now cleared

All five commits are live. Verified against what a browser receives, not the local build.

**All 33 hashed JS chunks fetched back through CloudFront (0 missing):**

```
wired                                       removed
  dataSourceWizard.allPersonas      1         {{total}}                            0
  dataSourceWizard.labelSeparator   2         Use feedback from selected sources   0
  sentiment.positive                1         Leave empty to use all personas      0
  sentiment.mixed                   1
```

`{{total}}` at 0 confirms the round-1 pluralization fix reached the bundle.

**Live catalogues — plural families and separator codepoints, all 8 locales:**

| Locale | separator | `allPersonas` variants | `_one` |
|---|---|---|---|
| en | `:` | 3 | `{{count}} persona` |
| de | `:` | 3 | `{{count}} Zielgruppe` |
| es | `:` | 4 (incl. `_many`) | `{{count}} arquetipo` |
| pt | `:` | 4 | `{{count}} arquétipo` |
| ko | `:` | 3 | `전체 {{count}}개 페르소나` |
| fr | `U+00A0` + `:` | 4 | `{{count}} profil` |
| ja | `U+FF1A` | 3 | `全{{count}}件のペルソナ` |
| zh | `U+FF1A` | 3 | `全部 {{count}} 个角色` |

No `all*` key still interpolates `{{total}}` in any locale.

One note on the verification itself: my first pass reported FAIL for all 8 while every individual check read OK. The aggregate was wrong, not the deployment — an unscoped search for `{{total}}` was matching `stepOf: "Step {{step}} of {{total}}"`, a pre-existing key that legitimately uses it. Scoped to the `all*` families, it passes. Worth stating because a check that fails for the wrong reason is as misleading as one that passes for the wrong reason.

### Still not verified

No signed-in walkthrough of the wizard in a non-English locale. The evidence here is transport-level (catalogues + bundle contents) plus the `de`-locale render assertions in `localization.test.tsx`. A human opening a project wizard with the language set to German — and specifically checking the Context Summary line reads `Zielgruppen: 1 Zielgruppe` on a one-persona project — is the one check outstanding.

perrozzi:

The only change since my last review swaps the separator table from a tuple array to two `Record<SupportedLanguage, …>` maps driven by `supportedLanguages`, which is exactly the nit I raised: membership is now enforced by `tsc` rather than by a length assertion, so adding a locale without a catalogue or an expected separator fails to compile.

### Resolved
- `localization.test.tsx`: locale coverage is now derived from the app's supported-locale list and both tables are exhaustively keyed, so `toHaveLength(8)`'s membership blind spot is gone.

### Still open
- `fr components.json`: `documents: "Fichiers"` / `allDocuments*: "… fichiers"` vs `documentsCount_many` / `existingDocumentsCount` saying "Documents" in the same panel.
- `{ja,zh,ko}` `all*_one` still carry the quantifier (harmless, dead entries).
- `en allDocuments_one: "{{count}} document"` reads as a count rather than a scope (nit, accepted).

### New findings
- Nit, `src/components/DataSourceWizard/localization.test.tsx`: the compile-time guarantee holds only as long as `supportedLanguages` is a `readonly SupportedLanguage[]` containing *every* member of the union (e.g. `as const` array that `SupportedLanguage` is derived from). If it is a hand-maintained `SupportedLanguage[]`, a locale added to the union but not to that array would still pass — worth a one-line comment noting the dependency, or nothing at all if the type is derived from the array.

Otherwise the increment is clean.

---

perrozzi:
## Round 5 disposition (`4203664`) — nit fixed, and it turned into a stronger guarantee

### ✅ `toHaveLength(8)` pins size but not membership — **fixed, good catch**

You're right that a duplicated entry or a ninth locale would have slipped through. Taking your suggestion further: instead of *asserting* the table covers the supported locales, both tables are now **keyed by `SupportedLanguage`** and the loop is driven by `supportedLanguages` from `src/i18n/languages.ts`. So the coverage guarantee is a **typecheck**, not an assertion, and the length assertion is gone as redundant.

Verified by adding a 9th locale (`'it'`) to `supportedLanguages`:

```
localization.test.tsx(209,11): error TS2741: Property 'it' is missing in type
  '{ en: …; es: …; … ko: … }' but required in type
  'Record<"fr" | "en" | "es" | "de" | "pt" | "ja" | "zh" | "ko" | "it", …>'

localization.test.tsx(219,11): error TS2741: Property 'it' is missing in type
  '{ en: string; es: string; … }' but required in type 'Record<…"it", string>'
```

Both maps fail, each naming the missing locale — so adding a language to the app without a catalogue **or** without an expected separator cannot compile.

### The three "still open" items

Unchanged and still declined, with the reasoning in rounds 3 and 4 — but one of them has been **escalated rather than accepted**:

**`fr documents: "Fichiers"` is now being treated as a gate bug, not a wording choice.** I said in round 3 that the correct French for "Documents" *is* "Documents" and the gate rejects it. Pushed on that, I measured it, and it is worse than I reported:

- Setting the correct value on four keys makes `i18n:check` **exit 1**: `de userAdmin.tableStatus` = `Status`, `fr feedbackForms:editor.description` = `Description`, `fr prioritization:scores.impact` = `Impact`, `fr scrapers:editor.date` = `Date`.
- So the shipped values are **wrong words**, not stylistic variants: de `Zustand` (physical condition) for an account Status column; fr `Légende` (*caption*) for Description, twice; fr `Portée` (*reach* — a different scoring dimension) for Impact on the prioritization sliders; fr `Horodatage` (*timestamping*) for Date. `projectDetail:tabs.documents` was already `Fichiers` before this PR, so someone hit the same trap earlier.
- And the rule's premise is already known-false inside its own function: **122 shipped values are identical to English and pass only via its three hardcoded exceptions** (106 by `length <= 3`). `"CSV"` may match English; `"Date"` at four characters may not.

That is a separate PR — an explicit per-entry allowlist plus reverting the wrong words — and it opens namespaces this PR never touches. Flagging here only to record that the finding you surfaced was real and is not being dropped.

The other two stand: the ja/zh/ko `all*_one` entries are unreachable under CLDR and match this file's existing CJK convention, and en `"1 document"` is the accepted trade-off against the grammar bug.

### Deployed and verified at this head

Redeployed; all 33 hashed chunks fetched back through CloudFront (0 missing):

```
+ dataSourceWizard.allPersonas      1      - {{total}}                            0
+ dataSourceWizard.labelSeparator   2      - Use feedback from selected sources   0
+ sentiment.positive                1
```

All 8 live catalogues: 3/3 plural families, correct separator codepoints (`fr` `U+00A0`, `ja`/`zh` `U+FF1A`), zero `all*` keys still interpolating `{{total}}`.

### Gates

eslint clean · `tsc` clean · **1883 tests / 121 files** · `i18n:check` `missing 0 / extra 12 (pre-existing) / empty 0 / untranslated 0` · unreferenced en keys **651** (from 697).

Outstanding, unchanged: no signed-in browser walkthrough of the wizard in a non-English locale.

perrozzi:
## Round 6 disposition — no change; the dependency you flagged holds by construction

Your own escape clause applies: the type **is** derived from the array, so they cannot diverge.

`src/i18n/languages.ts:16-17`:

```ts
export const supportedLanguages = ['en', 'es', 'fr', 'de', 'pt', 'ja', 'zh', 'ko'] as const
export type SupportedLanguage = (typeof supportedLanguages)[number]
```

`SupportedLanguage` is `(typeof supportedLanguages)[number]`, so the union is generated *from* the array. The failure mode you describe — "a locale added to the union but not to that array" — is unrepresentable: there is no separate union to add to. Adding a locale means editing the array, which widens the union, which is what made both `Record`s fail `TS2741` in the probe I ran last round.

No comment added either, on your own "or nothing at all" branch: a note explaining that a derived type is derived would go stale faster than it would help, and the `(typeof x)[number]` idiom is self-evident two lines from the array.

**Stopping the loop here**, per this repo's convention of dispositioning rather than churning on nit-only rounds (same call as #166 and #277). Six rounds: every should-fix resolved, the last two rounds nit-only, and the final nit a non-issue by construction.

Standing items, unchanged and recorded above: the two accepted trade-offs (ja/zh/ko dead `_one` entries, en `"1 document"`), and the fr `documents` noun — which is **not a wording choice but a gate defect**, being taken to its own PR with the evidence in my round-5 comment (four correct translations that `i18n:check` rejects; 122 values that pass only via its own hardcoded exceptions).

Head `4203664` is deployed and live-verified. eslint clean, `tsc` clean, 1883 tests, i18n gate `untranslated 0`. Ready for a maintainer.

PR #288, opened 2026-08-08
fix(ui): let the Background Jobs panel own long-running project progress

perrozzi:
## Summary

Three project actions — prototype build, prototype revision, product report — awaited their job in local
component state. Any unmount destroyed the only progress indication in the app, while the Background Jobs
panel sat directly above showing nothing. This hands the wait to the panel and gives the panel a resting
state so it stops outranking the content users came for.

Addresses **U7** and **U9** in `analysis/UI-ROUND2-GITHUB-ISSUES.md`.

## The panel was structurally blind, not merely uninformed

`useProjectData`'s jobs query sets `refetchInterval` to `3000` **only while a job is already running or
pending**, and `0` otherwise. The four wizard mutations each invalidate `['project-jobs', id]`; the three
`pollJobToCompletion` callers did not. So starting a build on a project with nothing in flight left the
panel empty **indefinitely** — there was no poll running that could ever discover the new job.

A second defect rode along: `pollJobToCompletion` defaulted to a five-minute deadline, which an Opus 5
prototype build routinely outlives, so each caller threw "took too long" for a job that was still running
and would succeed. With the panel owning the wait that deadline is **deleted rather than tuned**, and
`jobPolling.ts` goes with it — no consumers remain, and a second source of job status is exactly what
produced the contradiction.

## Design notes

- **The handover is a callback threaded from `ProjectDetail`**, which already owns the `queryClient` and
  the key, rather than `useQueryClient` in three leaf components. Mirrors the existing `onDocumentChanged`
  plumbing and keeps the key in one place — now `projectJobsKey(id)`, because it must be identical in three
  files that can drift and a typo fails *silently*.
- **`DocumentsTab` and `ProductTab` lose `onDocumentChanged`.** Their only use of it was the completion
  callback that no longer exists; the refreshed document list still arrives from `useProjectData`'s
  completed-job effect, mounted or not.
- **Failed jobs deliberately stay expanded** while completed ones collapse. U7's text said collapse both.
  A completion is read at a glance, a failure is not — and now that the actions hand off, this panel is
  the only place a failed build is reported, so collapsing failures would have made them unreportable.
- **`slice(0, 5)` removed.** It bounded a list that completed jobs flooded; once those collapse, a cap
  could only hide a *running* job behind finished ones. The API already caps the query at 50.
- **Dates on jobs that aren't from today.** `HH:mm` is ambiguous the moment a job outlives its day, and
  completed jobs are kept until dismissed. Routed through the existing `safeFormatDate`, which also fixes
  an unparseable `created_at` throwing "Invalid time value" out of the render.

## Two fixes in passing, in files this already rewrites

- The dismiss button hardcoded `title="Dismiss"` while `jobs.dismiss` sat translated in all 8 locales,
  unused. Wired, with an `aria-label`. Unreferenced English keys **651 → 650**.
- `ProductTab`'s report card checks for an empty product context **before** calling. The backend refuses
  to generate from one, but that refusal happens inside the async job — so the user would pay for a job
  that cannot succeed and read the reason as an untranslated job error. The field list derives from the
  empty-context factory, so a new field is covered automatically and the server-set `updated_at` is
  ignored. Implemented as a runtime type guard because eslint here forbids both `as` assertions and `let`.

## Locale bookkeeping

`jobs.completedCount_*` uses **each locale's own suffix set, copied from the sibling `notesCount` family**
(en/de/ja/ko/zh `_one`+`_other`; es/fr/pt add `_many`) — inventing a set makes `i18next-parser` write the
missing variant with an empty value and fail the gate on the next `i18n:validate`.
`product.report.successTitle`/`successMessage` → `startedTitle`/`startedMessage`, since the acknowledgement
now fires at kick-off. `product.report.timeout` and `documents.prototype.timeout` are deleted: their
consumers are gone, and an orphaned key is the defect #287 was about.

## Testing

- **The regression test lives at the `ProjectDetail` level**, because a component-level test could only
  prove the button calls its own prop, not that anything wires that prop to the query. It asserts the
  observable consequence — the jobs list refetches after a build starts on an idle project.
- **Fail-on-revert verified both ways:** unwiring `onJobStarted={handleJobStarted}` fails the handover
  test; removing the `contextIsEmpty` guard fails 2 of its 3 tests.
- **One of my own tests initially passed for the wrong reason.** jsdom has no `Element.scrollTo`, the
  interview pane's effect threw, `ProductTab` rendered as an empty `div`, and "the API was not called" was
  trivially true. Stubbed with a comment — any *not-called* assertion is satisfied by a component that
  failed to render.
- eslint clean, tsc clean, **1890 tests across 122 files**.
- i18n gate unchanged: missing 0, empty 0, untranslated 0, and none of the new keys are unreferenced.
  `typecheck:tests` holds at its pre-existing 85, none in touched files.

## Not verified

Deployment and a signed-in walkthrough — see the deploy note added after review, if present. The handover
is exercised against mocked APIs, not a real Bedrock build.


perrozzi:

This PR moves the wait for three long-running project actions (prototype build, prototype revision, product report) out of local component state into the Background Jobs panel: the three callers now fire-and-forget and invoke a new `onJobStarted` callback that invalidates `projectJobsKey(id)` from `ProjectDetail`, `jobPolling.ts` is deleted, `JobsSection` gains a collapsed "completed" group, dated timestamps via `safeFormatDate`, and a translated dismiss label, and `ProductTab` gains a pre-flight empty-context guard. The diagnosis (the jobs query's `refetchInterval` is `0` when nothing is in flight, so a job started outside a wizard was never discovered) is correct, the query key centralisation is the right fix, locale coverage looks complete across all 8 files, and the tests are placed at the level where they can actually fail on revert. A few behavioural edges are worth resolving before merge.

### Should fix

- **The handover is a single invalidation, so it inherits a race with job-record creation** (`ProjectDetail.tsx` `handleJobStarted`, `useProjectData.ts`). `invalidateQueries` triggers exactly one refetch. If the backend's `buildPrototype`/`generateProductReport` handler returns `job_id` before the job row is readable by `getJobs` (or the response is served from a slightly stale read), that refetch returns `[]`, `refetchInterval` stays `0`, and the panel is blind again — the exact failure mode this PR fixes. Consider having `onJobStarted` prime the panel (e.g. keep a short polling window after a start, or make `refetchInterval` return `3000` for N seconds after the last start) rather than relying on one round trip winning the race. The new `ProjectDetail.jobHandover.test.tsx` mocks `getJobs` synchronously, so it cannot surface this.
- **`ProductTab`'s `isContextEmpty` guard is narrower than the backend rule it stands in for** (`ProductTab.tsx`). The backend accepts "at least one field **or** an uploaded internal document", and `errorEmpty` says exactly that — but the guard only inspects fields, so a user who uploaded an internal doc and filled nothing is now blocked client-side with a message telling them to do what they already did. The justification lives only in a code comment ("no product doc can currently reach the `ready` state"); if that ever stops holding, or is wrong today, this is a hard block on a valid path. `ProductTab` already fetches `listProductDocs`, so folding a ready-doc check into the guard is cheap and removes the coupling to an unverified backend assumption.
- **Verify the completion path that replaced `onDocumentChanged`.** `DocumentsTab`/`ProductTab` no longer invalidate `['project', id]` on completion; the PR relies on "`useProjectData`'s completed-job effect", which is not in this diff. If that effect only refetches the jobs list and not the project query, a finished prototype revision or report won't appear in the Documents tab until a reload — a user-visible regression that none of the new tests cover.
- **No acknowledgement that a build/revision actually started.** `ProductTab` correctly keeps a local "started" line because it sits below the fold; `BuildPrototypeButton` and `PrototypeFeedbackButton` show nothing at all — spinner flickers, panel updates one refetch later (and the panel renders `null` while `jobs` is empty). Worth the same one-line acknowledgement for consistency.
- **`PrototypeFeedbackButton` clears `feedback` and closes on *start*, not success.** If the job fails minutes later, the only report is a raw backend error in the jobs panel and the user's typed feedback is gone. Consider retaining the text (or surfacing the failed job's `base_prototype_id` context) so a retry doesn't mean retyping.

### Nits

- `ProductTab.tsx`: the `ponytail:` comment prefix looks like a stray personal/agent marker; use the repo's normal `NOTE:`/`TODO:` convention.
- `ProductTab.tsx`: `isContextField` can never be false for `Object.keys(emptyContext())`, so `!isContextField(field, template) ||` is dead code — the guard exists only to satisfy the indexer. A `keyof ProductContext[]` const derived from the template, or `Object.entries`, reads clearer.
- `BuildPrototypeButton.test.tsx`: "stops showing the busy label once the request returns" is satisfied vacuously by a button that never showed `Building…` (the mock resolves immediately). Assert it appears and then disappears, per the PR's own point about not-called/not-present assertions.
- `ProductTab.reportGuard.test.tsx`: `Element.prototype.scrollTo = vi.fn()` in `beforeAll` is never restored; `vi.stubGlobal`/`afterAll` restore keeps the file self-contained.
- `TabContent.tsx` still declares and destructures `onDocumentChanged` — confirm a remaining tab consumes it, otherwise drop the now-dead prop from `TabContent` and `ProjectDetail`'s call site.
- `JobsSection.tsx`: with `slice(0, 5)` gone, `needsAttention` is uncapped — a project that accumulated many failed jobs will now push the tab content far down the page. A cap (or grouping) for *failed* jobs specifically would keep the resting-state goal intact.

---

perrozzi:
All five should-fixes addressed in `bef78f5`, plus every nit. Disposition below.

### The race is real, and it was mine

Finding 1 is correct and it is the most important thing in the review: I replaced "the panel never
discovers the job" with "the panel discovers it only if one round trip wins a race". Verified the
mechanism rather than assuming it — `api_list_jobs` (`projects_handler.py:325-338`) is a `query` with
**no `ConsistentRead`**, and the handler returns `job_id` immediately after `create_job` writes the row
and the worker is invoked. So an empty first refetch is a legitimate outcome, not a hypothetical.

`ProjectDetail` now records when an action reported a start, and the jobs poll stays alive for 30s past
it. Inside that window it keeps polling **even when every visible job has finished**, because what is
visible proves nothing about whether the new row is readable yet — a project whose earlier jobs are all
complete is exactly where the missing row hides.

The decision moved into a pure exported `jobsPollInterval(jobs, startedAt, now)` and is pinned there,
both edges of the window included. Worth recording why: my first attempt drove the component through
fake timers, and **the timers leaked past the test's own `finally` cleanup and timed out unrelated
suites** (`ProjectTabs`, `PickerComponents`). The interesting cases here are all about elapsed time, so
a pure function is the right seam.

### The guard was narrower than the rule it stands in for

Finding 2 is right, and the fix removes the dependency you correctly objected to. The backend accepts
"at least one field **or** a ready uploaded document" and `errorEmpty` says exactly that, so my
field-only check could have told a user who uploaded a document to do what they had already done. It now
checks both halves, asking for the document list **only when the fields are empty** — so the common path
costs nothing — and a failure to read that list resolves to "let the backend decide" rather than
blocking. Four new cases cover it: docs-only starts, a `pending` doc still blocks, a failed list call
fails open, and a filled field does not trigger the extra request.

One correction to the finding: `ProductTab` did not already fetch `listProductDocs` — its `DocsUpload`
child does, privately. That is why the check asks at click time instead of reusing an existing fetch.

### Completion path: verified and now pinned

Finding 3 was outside the round-1 diff, which is why the review couldn't see it. It holds:
`useProjectData` invalidates `['project', id]` when a job reports `completed_at` within the last ten
seconds, and the jobs query is polling at 3s intervals while the job runs, so the completion always
lands inside that window. Added a test that asserts the project query refetches on completion with no
component waiting on it.

### Findings 4 and 5

Both taken. The two prototype actions now share `documents.prototype.started`, since the panel renders
`null` until its first refetch returns the job. The revision form keeps the typed feedback when it
closes — the job can still fail minutes later, and clearing it meant retyping to retry.

### Nits

All taken except one:

- **The cap is back**, at 5, with in-flight work ordered ahead of failures so it can never hide a running
  job, plus a "+N more" line when it bites. You were right that removing it traded one unbounded list for
  another; ordering is what the original `slice(0, 5)` lacked.
- `ponytail:` gone. And `isContextField`'s unreachable branch went with it — the field list is now an
  explicit `as const satisfies ReadonlyArray<keyof ProductContext>`, which keeps the typo protection
  without the dead code.
- The busy-label test holds the request open, so it asserts the label appears **and then** disappears
  rather than passing on a label that never rendered. Same class as the trap the PR description
  describes, and I had reintroduced it two files over.
- `Element.prototype.scrollTo` is saved and restored in `afterAll`.
- **`TabContent`'s `onDocumentChanged` is not dead** — `ChatTab` still consumes it for save-as-document
  (`TabContent.tsx:135`). Left as-is.

eslint clean, tsc clean, **1904 tests / 123 files**. i18n gate unchanged (missing 0, empty 0,
untranslated 0, unreferenced 650) and the three new keys are all referenced. `typecheck:tests` is back to
its pre-existing 85 — the new fixtures had added 4 by widening `status` to `string`, now typed.

Still not verified: no deployment yet. That is next, and the start-window behaviour is the specific thing
a live run can confirm that jsdom cannot.

perrozzi:

Since the last review the author added a 30-second post-start polling window (`jobStartedAt` → `jobsPollInterval`, with a pure-function test file), widened the report guard to include ready product docs, added "Started — track it in Background Jobs" acknowledgements to both prototype buttons (key present in all 8 locales), retained the feedback text on start, reinstated a 5-row cap ordered in-flight-first with a `moreCount` line (all 8 locales), and tightened the three test weaknesses I flagged. Nearly everything from the previous review is addressed; one small regression came in with the guard rewrite.

### Resolved
- Single-invalidation race: `JOB_START_POLL_WINDOW_MS` keeps the poll alive across an empty first refetch, and the boundary cases are pinned in `useProjectData.test.ts`.
- Guard now mirrors the backend's "fields **or** a ready document" rule, fails open on a list error, and skips the extra request on the common path.
- Completion path verified: `ProjectDetail.jobHandover.test.tsx` asserts `getProject` refetches on a completed job with nothing waiting on it.
- `BuildPrototypeButton` / `PrototypeFeedbackButton` now acknowledge the start; feedback text is retained for retry.
- Uncapped `needsAttention` re-bounded (`MAX_VISIBLE_JOBS`, in-flight ordered first, `jobs.moreCount`).
- `ponytail:` comment and the dead `isContextField` branch gone; busy-label test no longer vacuous; `Element.prototype.scrollTo` restored in `afterAll`.

### Still open
- `TabContent.tsx` still declares and destructures `onDocumentChanged` — confirm a remaining tab consumes it (nit).

### New findings

**Should fix**

- `ProductTab.tsx`: `CONTEXT_FIELDS` is now a hand-written list. `satisfies ReadonlyArray<keyof ProductContext>` catches a *typo* but not an *omission* — add a field to `ProductContext`/`emptyContext()` and it is silently excluded from `hasNoContextFields`, so a project where only that new field is filled starts a job the backend will reject inside the worker. That is exactly the failure the old template-derived version couldn't have. A compile-time exhaustiveness assertion (e.g. a type-level `Exclude<keyof ProductContext, 'updated_at' | typeof CONTEXT_FIELDS[number]>` that must be `never`) restores the guarantee without an indexer.

**Nits**

- `JobsSection.tsx`: the comment says the ordering means "the cap can never hide a running job", but `slice(0, MAX_VISIBLE_JOBS)` applies to the combined list — six or more concurrent jobs (the wizard's `Promise.all(types.map(...))` can produce several) will hide an in-flight one. Either cap only `failed` or soften the claim.
- `JobsSection.tsx`: the hidden-count line is a plain `<p>`, so hidden *failures* are unreachable until the visible ones are dismissed; the completed group's expandable pattern would read more consistently.
- `DocumentsTab.tsx`: `started` in `PrototypeFeedbackButton` is never cleared, so "Started" sits next to the button indefinitely — including after the panel has reported that revision as failed or completed.
- `ProductTab.reportGuard.test.tsx`: `expect(mockListProductDocs.mock.calls.length).toBeLessThanOrEqual(1)` can't distinguish DocsUpload's own call from the guard's; snapshot the count before the click and assert equality.

---

perrozzi:
Round-2 findings addressed in `2aac1d9`.

### The should-fix was a regression I introduced by taking round 1's nit at face value

You are right, and the sequence is worth recording: round 1 called the type guard's branch dead, I
replaced the template-derived walk with a hand-written list, and in doing so traded a *guarantee* for a
*typo check*. `satisfies` cannot see an omission.

The check now builds an `Omit<ProductContext, ServerSetField>` literal, so leaving a field out is a
compile error. **Verified rather than asserted:** adding a `pricing_model: string` to `ProductContext`
fails `tsc` with TS2741 naming the missing property, at both `emptyContext()` and the new literal.

Three other spellings were tried first and rejected by lint, not by taste — noting them so the next
person doesn't retread it:

- a rest destructure (`{ updated_at: _x, ...fields }`) trips `no-unused-vars`, which here honours neither
  a leading underscore nor rest siblings;
- `Object.entries(context)` types its values as `any` under this config, tripping `no-unsafe-call` and
  `no-unsafe-member-access`;
- `satisfies keyof ProductContext` on a `const` trips `no-unused-vars` ("only used as a type").

### The ordering claim was wrong, and the cap is now where it belongs

Correct — `slice` applied to the combined list, so my comment's claim was false for six or more
concurrent jobs, and that is reachable: the document wizard fires PRD and PR-FAQ as separate jobs. **The
cap now applies to failures only and in-flight work is never capped.** In-flight is bounded by what a
user can start; failures are not, which was the only reason to have a cap at all. A test builds six
failures plus one running job and asserts the running one is visible.

### Hidden failures are reachable

Also right: as a plain count they were unreachable until the visible failures were dismissed one at a
time. The overflow and the completed group are now the same collapsible component, so both expand. Tested
by asserting the fifth failure's message is absent, then present after expanding.

### The acknowledgements now lower themselves

Good catch, and it applied to all three sites rather than just the revision form — including the product
report card, whose predecessor `success` flag had the same defect. They share `useTransientFlag`, which
also clears its timer on unmount so a late `setState` cannot fire into a dead component.

### `listProductDocs` assertion

Snapshots the count before the click and requires equality, so it can no longer confuse `DocsUpload`'s
own call with the guard's.

### Still-open nit, re-answered

**`TabContent`'s `onDocumentChanged` is not dead.** `ChatTab` consumes it for save-as-document —
`TabContent.tsx:135` passes it, `ChatTab.tsx:212` uses it. Only `DocumentsTab` and `ProductTab` stopped
needing it. Left as-is.

eslint clean, tsc clean, **1906 tests / 123 files**. i18n gate unchanged (missing 0, empty 0,
untranslated 0, unreferenced 650). `typecheck:tests` holds at its pre-existing 85.

Deployment is still outstanding and is blocked on interactive Midway auth on my side, so the live
verification of the start window has not happened yet. I would rather say that plainly than let the green
test count imply it.

perrozzi:

Since the last review the author added a `useTransientFlag` hook and routed all three "Started" acknowledgements through it, changed the cap to apply to failures only (`MAX_VISIBLE_FAILED = 3`) with overflow failures moved into the generalised `CollapsibleJobGroup`, rewrote `hasNoContextFields` around an `Omit<ProductContext, ServerSetField>` object literal for compile-time exhaustiveness, and tightened the docs-list call-count assertion to equality. All four items from the previous review are addressed.

### Resolved
- `ProductTab.tsx`: omitting a new `ProductContext` field is now a compile error rather than a silent gap in the guard.
- `JobsSection.tsx`: the cap no longer spans the combined list, so concurrent in-flight jobs can't be pushed out; comment corrected and a 6-failure + 1-running test pins it.
- `JobsSection.tsx`: overflow failures are reachable via an expandable group instead of a bare `<p>` count, with a test.
- `DocumentsTab.tsx` / `BuildPrototypeButton.tsx` / `ProductTab.tsx`: `started` self-clears after 8s, so it can't contradict the panel.
- `ProductTab.reportGuard.test.tsx`: pre-click snapshot + equality now distinguishes DocsUpload's own call from the guard's.

### Still open
- `TabContent.tsx` still declares and destructures `onDocumentChanged` — confirm a remaining tab consumes it (nit).

### New findings

**Should fix**

- `voc-datalake/frontend/src/pages/ProjectDetail/ProductTab.tsx`: the exhaustiveness guarantee holds only if every `ProductContext` member is *required*. If any field is (or becomes) optional — the `(value ?? '')` suggests the values may be `string | undefined` — the `Omit` annotation accepts an object literal that omits it, and the doc comment's "leaving a field out is a compile error" becomes false. Worth verifying against the type and, if fields are optional, adding a `Required<Omit<ProductContext, ServerSetField>>` annotation instead.

**Nits**

- `useTransientFlag.ts`: the returned object is a fresh literal each render, and it is listed in the `useCallback` deps of `runBuild` / `onSubmit` / `onGenerate`, so those callbacks now change identity on every render (including each `isSet` toggle). Wrap the return in `useMemo` or return stable pieces.
- No test covers the new self-clearing behaviour (the fix for the previous nit) or `useTransientFlag` itself; a small fake-timer test on the hook alone would avoid the cross-file timer leakage noted for the component suites.
- `JobsSection.tsx`: `jobs.moreCount` ("{{count}} more") is now specifically the *failed* overflow; the red icon carries that meaning visually but the accessible label doesn't. A dedicated key (e.g. `jobs.moreFailedCount`) would read unambiguously.

---

perrozzi:
Round-3 findings addressed in `6118c06`.

### The should-fix was conditional, and the condition holds today — hardened anyway

Checked the type rather than inferring it from the `?? ''`: `ProductContext` has **11 authored fields,
all required**, and `updated_at?` is the only optional member — which is the one deliberately dropped. So
the guarantee was intact. But you are right that `Omit` alone would accept a literal skipping an optional
member, which would make the doc comment quietly false the day someone loosens a field. Now annotated
`Required<Omit<ProductContext, ServerSetField>>`, so it no longer depends on that.

The `?? ''` stays as runtime defence: the values arrive from the API, and a missing one should read as
empty rather than throw.

### Nits

- **Identity churn: real, fixed.** The returned object was a fresh literal each render while callers list
  it in `useCallback` deps, so `runBuild` / `onSubmit` / `onGenerate` changed identity on every render —
  including on each toggle of the flag itself, which is the one moment it matters. Memoised.
- **The self-clearing behaviour had no test.** Added, on the hook alone as you suggested: raises, holds
  for the duration, lowers itself, restarts on a second start, clears on demand, keeps a stable identity
  across renders, and drops its timer on unmount. Fake timers are safe there because the file drives the
  hook alone and restores them in `afterEach` — **which runs even when a test fails, unlike the `finally`
  inside a test body that leaked timers out of my earlier component-level attempt.**
- **`jobs.moreCount` → `jobs.moreFailedCount`**, and the text now says "failed" in all 8 locales. The
  group is specifically the failures past the inline cap; the red icon carried that meaning visually while
  the label did not.

eslint clean, tsc clean, **1914 tests / 124 files**. i18n gate unchanged (missing 0, empty 0,
untranslated 0, unreferenced 650 — the renamed key is referenced, the old one is gone from all 8
locales). `typecheck:tests` holds at its pre-existing 85.

### On the `TabContent.onDocumentChanged` item

It has now been raised three times, so to close it out concretely rather than repeating myself:
`TabContent.tsx:135` passes it to `ChatTab`, and `ChatTab.tsx:212` uses it to refresh the document list
after save-as-document. Only `DocumentsTab` and `ProductTab` stopped needing it. Nothing to remove.

### Outstanding, and not something the test count can settle

Deployment has not happened — it needs interactive Midway auth on my side. The specific thing a live run
would confirm and jsdom cannot is the post-start polling window against a real eventually-consistent read.
Flagging it rather than letting four green rounds imply otherwise.

perrozzi:

Since the last review the author renamed `jobs.moreCount` → `jobs.moreFailedCount` across all 8 locales (and the `JobsSection` call site), tightened `hasNoContextFields` to `Required<Omit<ProductContext, ServerSetField>>`, memoised `useTransientFlag`'s return, and added a dedicated `useTransientFlag.test.ts`. All three items from the previous review are addressed; nothing new looks blocking.

### Resolved
- `ProductTab.tsx`: `Required<Omit<…>>` closes the loophole where an optional member could be skipped by the literal.
- `useTransientFlag.ts`: `useMemo` stops the returned object from changing identity on every render (it still changes on an `isSet` toggle, which is unavoidable given `isSet` is part of the object — fine).
- `useTransientFlag.test.ts`: the self-clearing behaviour, restart-on-re-set, and stable identity are now pinned, in a hook-only suite so the fake timers stay contained.
- `jobs.moreFailedCount` (8 locales + call site): the overflow label now reads unambiguously; no orphaned `moreCount` left behind.

### Still open
- `TabContent.tsx` still declares and destructures `onDocumentChanged` — confirm a remaining tab consumes it (nit).

### New findings

**Nits**

- `voc-datalake/frontend/src/pages/ProjectDetail/useTransientFlag.test.ts`: the "drops its pending timer on unmount" test passes with or without the `useEffect(() => cancelTimer, …)` cleanup — a `setIsSet` after unmount doesn't throw in React 18, so nothing fails on revert. `expect(vi.getTimerCount()).toBe(0)` after `unmount()` would actually pin the cleanup.
- `voc-datalake/frontend/src/pages/ProjectDetail/ProductTab.tsx`: with the `Required<…>` annotation the values are non-nullable, so `(value ?? '')` in `hasNoContextFields` is now dead — harmless, but it's the thing that suggested the fields might be optional in the first place.

---

perrozzi:

Two small follow-ups since the last review: a comment in `ProductTab.tsx` justifying the `?? ''` fallback, and a strengthened unmount test in `useTransientFlag.test.ts` that now asserts on `vi.getTimerCount()`. Both previous nits are addressed; nothing new is blocking.

### Resolved
- `useTransientFlag.test.ts`: `getTimerCount()` 1 → 0 across `unmount()` now genuinely pins the `useEffect(() => cancelTimer, …)` cleanup and fails on revert.
- `ProductTab.tsx`: the `?? ''` fallback is documented as deliberate JSON-null defence rather than looking like a leftover.

### Still open
- `TabContent.tsx` still declares and destructures `onDocumentChanged` — confirm a remaining tab consumes it (nit).

### New findings

**Nits**

- `voc-datalake/frontend/src/pages/ProjectDetail/ProductTab.tsx`: the new comment says a cleared field "can arrive as null", but `Required<Omit<ProductContext, ServerSetField>>` (and presumably `ProductContext` itself) types these as non-nullable `string`. If null is genuinely reachable, the honest fix is `string | null` on the type; as written the annotation and the comment disagree about the same values. Harmless at runtime either way.

---

perrozzi:
Both nits fixed in `625a466`. Stopping the loop here per the repo's nit-only convention (#166, #277,
#287), with the reasoning recorded rather than a fifth round of churn.

### The unmount test was vacuous — my own, for the fix to the previous round's nit

You are right, and it is the third time this branch has hit that class: an assertion that "nothing threw"
is satisfied by code that also does nothing. It now asserts `vi.getTimerCount()` is `1` before unmount and
`0` after. **Verified by removing the `useEffect(() => cancelTimer, …)` cleanup: that one case fails, 7
pass.** Worth noting the pattern — for a cleanup, assert on the resource, not on the absence of a
symptom.

### `(value ?? '')` is not dead, and the comment now says why

Checked before removing it. Every `setContext` spreads over `emptyContext()`
(`ProductTab.tsx:164`/`:179`/`:188`), but a spread only fills keys that are **absent** — and these values
arrive as JSON, where a cleared field can come back as `null`, which survives the spread and is not
modelled by the type. So the guard still does work; what was wrong was that it read as evidence the fields
were optional, which is what sent us both looking. Kept, with that stated inline.

### Remaining item, closed out

`TabContent.onDocumentChanged` has been raised in all four rounds. Concretely: `TabContent.tsx:135` passes
it to `ChatTab`; `ChatTab.tsx:212` consumes it to refresh documents after save-as-document. Only
`DocumentsTab` and `ProductTab` stopped needing it. Nothing to remove — treating this as dispositioned.

### Where the PR stands

eslint clean, tsc clean, **1914 tests / 124 files**. i18n gate unchanged (missing 0, empty 0, untranslated
0, unreferenced 650). `typecheck:tests` at its pre-existing 85.

Four rounds produced two genuine defects in my own work — the single-invalidation race and the
non-exhaustive field list — plus two vacuous tests. All fixed and all pinned with fail-on-revert checks.

**One thing is still unverified and no test count can settle it: nothing has been deployed.** The
post-start polling window is a fix for an eventually-consistent read, and jsdom cannot reproduce that. It
needs a live run against a real project. That is blocked on interactive auth on my side, so I am flagging
it rather than implying the PR is fully validated.

perrozzi:
## Deployed and live-verified at `625a466`

Verified on a deployment carrying real project data. Deployment-specific identifiers are deliberately
omitted from this comment.

### The build is actually being served, not just uploaded

The chunk carrying `ProjectDetail` (180,584 bytes) fetched back **through the CDN** is byte-identical to
the local build, and contains `jobs.completedCount`, `jobs.moreFailedCount`,
`documents.prototype.started`, `product.report.startedTitle`, `project-jobs` and `jobs.dismiss`. The
strings this PR removes are absent: `took too long`, `product.report.successTitle`, `jobs.moreCount"`,
`pollJobToCompletion`. Locale catalogues checked live for en/de/fr/ja — new keys present, fr correctly
carrying the extra `_many` form, and the deleted `timeout` keys gone.

One trap worth recording: **grep the chunk of the current build, not the whole asset prefix.** The sync
does not delete, so hundreds of hashed JS files from earlier deploys remain; a hit on a "removed" string
there could come from a stale asset and would mean nothing. Locate the chunk in the local `dist/` first,
then fetch that one filename through the CDN and `cmp` it.

### The handover works end to end, in the state that used to be blind

Exercised on a project with **zero jobs** — the exact idle case where the old code never discovered the
job — that has both a PRD and a PR-FAQ:

| | |
|---|---|
| `POST /build-prototype` returned `job_id` | 0.37s |
| job visible to `GET /jobs` | **first read, +0.39s** |
| ran to | `completed`, progress 100, **239s** |
| documents | 8 → **9**, newest a `prototype` with a `prototype_url` |
| jobs list afterwards | 1 completed ⇒ the panel renders **0 inline, 1 collapsed** — U7's resting state on real data |

### Three things the run did not settle

Stating these rather than letting "deployed and verified" imply more than it should:

1. **The eventually-consistent miss did not reproduce.** The first read won, in one trial. So the
   30-second window is *defensive*, not demonstrated-necessary — `api_list_jobs` still has no
   `ConsistentRead`, so the miss stays possible, and one success does not prove the first read always wins.
   It costs nothing when the read does win: once the job is visible and running, `hasRunning` keeps the
   poll alive regardless.
2. **The false timeout did not reproduce either** — 239s sits inside the deleted 300s client deadline. It
   came within **61 seconds** of it, which is the substance of the original concern rather than a
   refutation of it.
3. **No signed-in browser walkthrough.** The collapse/expand interaction, the failed-overflow group and the
   "Started" line rest on the tests and the artifact greps, not on a human looking at them.

Ready for a human decision — `reviewDecision` is empty because this repo has no CI, so the four automated
rounds are advisory only.

PR #289, opened 2026-08-08
fix(ui): order the project Overview by what the generators can actually read

perrozzi:
## No longer stacked

#288 was squash-merged while this was in flight, so `development` now carries it and this branch has
merged the new base in. **The diff is this PR's own 21 files.** The merge resolution is documented in
`1ae1f38` — the interesting part is that git auto-merged the *other* side's copies of
`hasNoContextFields`, `ServerSetField`, `hasReadyProductDoc` and an import back in **without flagging
them as conflicts**; tsc caught them as duplicate identifiers, and they are removed.

## Summary

The project Overview tab presented five equal-weight launchers in an order the code does not support, and
it read identically on an empty project and a finished one.

**Order.** Personas are built from feedback alone. Research can additionally read personas — the research
step handler injects each selected persona's name, tagline, goals, frustrations and quote. The document
generators can read feedback, personas, research and documents. The grid put PRD/PR-FAQ third and Run
Research fourth, so a user following it top to bottom produced documents with neither research nor a
deliberate persona selection behind them. Cards are now **Product → Personas → Research → PRD/PR-FAQ →
Remix**, numbered.

**State.** `OverviewTab` declared a `personas` prop and never destructured it, so the tab whose job is
orientation could not say how many personas existed, whether research had run, or whether a PRD existed.
Each card now reports its own output, and one derived line points at the earliest gap.

**Phantom inputs.** `PersonaWizard` rendered the shared wizard with `personas` and `documents`, so it
offered Personas / Documents / Research toggles and item pickers and summarised the selection back — while
`generatePersonas` sends none of it. `hideDataSources` (already supported, already used by `MergeWizard`)
closes that. It mattered more after the reorder, since Personas is now exactly where someone would look
for "use my research".

**Payoff for the ordering.** `openResearchWizard` opens the research wizard with the project's existing
personas already selected, so the grounded run is the default rather than something to discover.

## Two decisions worth reviewing

**1. The product card costs one extra request.** Its state is the only card state not derivable from data
the page already holds: `getProductContext` had a single caller, inside the Product tab. `useProjectData`
now fetches it too — one small GET per project open, in exchange for a first step that can say whether it
is done. `undefined` stays deliberately distinct from "all fields blank": a card that cannot know renders
no state rather than claiming the description is empty, and the next-step line skips step 1 while the
answer is unknown rather than telling someone to describe a product they already described.

The Product tab still fetches the context itself, because it edits the record field by field and owns that
local state — so visiting that tab is two reads. The shared query key is exported for whenever that tab
moves onto React Query.

**2. Numbering, not gating.** Cards are numbered and a card whose optional upstream input is missing shows
an advisory hint, but nothing is disabled. Every generator works without its optional inputs, just less
well, so gating would block work the backend accepts. Remix keeps its existing hard requirement.

## Anti-duplication

#288 introduced an exhaustive `Required<Omit<ProductContext, ServerSetField>>` literal in `ProductTab` for
its emptiness guard, where leaving a field out is a compile error. The completeness count needs the same
list, and a second copy would have left that guarantee holding in one of two places. The literal moves to a
shared `productContextFields.ts`, and the field total is **derived** from it — so the UI cannot promise
"of 11" once there are 12.

## Notes for reviewers

- `overviewState.ts` is pure and returns step **ids**, never strings. A key assembled at runtime is
  invisible to the i18n extractor, which is the blind spot #287 was about; the component maps ids through
  literal `t()` calls. The orphan count staying at 650 is the evidence all 17 new keys are wired.
- `remix.hasOutput` is always false, and that is a stated limitation rather than a fact: a remixed document
  is stored as an ordinary PRD or PR-FAQ, so nothing distinguishes it from a generated one. Nothing
  recommends remix or reports it outstanding, so the limitation stays invisible.
- The numeric badge is `aria-hidden`; the position is announced from an `sr-only` span inside each card's
  heading, so the sequence is not sighted-only.
- Two new locale values came back identical to English in up to five locales, because neither "Personas"
  nor the two acronyms translate. Reworded around nouns that do, rather than weakening the gate.

## Verification

| Gate | Result |
|---|---|
| `tsc` | clean |
| `eslint` | clean |
| frontend tests | **1952 passed / 127 files** |
| `i18n:check` | missing 0, empty 0, untranslated 0; unreferenced unchanged at **650** |
| `typecheck:tests` | **85 → 74** (the old Overview test omitted a required prop) |

The pre-existing ja/zh/ko `extra: 12` in `categories.json` is untouched and still the only reason
`i18n:check` exits non-zero.

**Fail-on-revert:** the keystone test renders `OverviewTab` for an empty project and a populated one and
asserts the two renders differ. Before this change they were identical — that is the bug expressed as one
assertion. A second test asserts `ResearchWizard` **still offers** personas, so the persona-wizard fix
cannot be "achieved" by hiding data sources everywhere.

**Not verified at the time of opening:** no deployment, and no signed-in browser walkthrough. Card states
are exercised against fixtures.

Refs U8 in the round-2 UI review and PR2 of the project-screen plan.


perrozzi:

This PR (top two commits on top of #288) reorders the project Overview into dependency order (Product → Personas → Research → PRD/PR-FAQ → Remix), gives each card derived state via a new pure `overviewState.ts`, extracts the product-context field list into `productContextFields.ts` so the "of N" total is derived, hides the data sources `generatePersonas` discards in `PersonaWizard`, and pre-selects personas when opening the research wizard. The design is sound, the i18n additions are complete across all eight locales, and the derivation is well covered by unit tests. Note the diff also carries #288's job-handover work (`useTransientFlag`, `JobsSection` resting state, removal of `jobPolling.ts`), so this review is partial with respect to which commits are actually under discussion. No blockers found.

### Should fix

- **`productContext` now has two owners and they never resync.** `useProjectData.ts` fetches `productContextKey(id)`, while `ProductTab.tsx` keeps its own `useState` copy and writes through `projectsApi.updateProductContext` / `onPatchFromChat` without invalidating that key. Since `useProjectData` lives in `ProjectDetail` and stays mounted across tab switches, filling fields in the Product tab and returning to Overview leaves card 1 reporting the old "X of 11 fields filled" for the rest of the session. Either invalidate `productContextKey(projectId)` in `ProductTab`'s save paths, or move that tab onto the shared query now rather than "whenever".
- **New query has no `staleTime`.** `useProjectData.ts`'s product-context query uses defaults, so it also refetches on window focus — the "one small GET per project open" is in practice more than one. A `staleTime` (and `retry: false`, given the card is designed to tolerate `undefined`) would match the stated cost.
- **The keystone assertion is weaker than the claim.** `OverviewTab.test.tsx`'s `expect(populated.innerHTML).not.toBe(emptyHtml)` fails-on-revert today, but it passes for *any* markup difference — including a future partial regression where only one card's state survives. The per-card assertions in `reports what each step has produced` are the real guard; consider making the keystone assert the specific state strings appear in one render and not the other.

### Nits

- Nested locale keys are inserted unsorted (`overview.hint`: `researchNeedsPersonas` before `documentsNeedContext`; `overview.state`: `product`, `productEmpty`, `personas`, …) while every surrounding level in `projectDetail.json` is alphabetical. Worth aligning in all eight files.
- `overview.state.personas` / `research` / `documents` pass a count through `{{total}}` rather than i18next `count`, so no locale can ever pluralise those lines. Fine for the acronym cases, arguably not for "Personas created: 1".
- `useTransientFlag.ts`'s comment says the returned object has a stable identity "so callers can list it in deps", but the `useMemo` includes `isSet`, so identity changes on every raise/lower — `runBuild`/`onGenerate` are recreated twice per job start. Harmless, but the comment overstates it; memoise `{ set, clear }` separately if the guarantee is wanted.
- `OverviewTab.test.tsx` selects by index (`getAllByRole('button', { name: /Generate/i })[1]`), which now depends on the very ordering under test. A scoped `within(card)` query would be less circular.
- With `DocumentsTab` and `ProductTab` switched to `onJobStarted`, check whether `onDocumentChanged` still has a consumer in `TabContent.tsx`/`ProjectDetail.tsx` or is now a dead prop.

---

perrozzi:

Since the last review the author sorted the new locale keys alphabetically in all eight files, rewrote the keystone Overview test to assert the four specific state strings (and switched the "not disabled" assertions to card-scoped `within()` queries), added a `ProductTab` → `TabContent` → `ProjectDetail` `onContextSaved` handover that seeds `productContextKey` from the save response, gave the new query `staleTime` + `retry: false`, added a dedicated `ProductTab.contextSaved.test.tsx`, and — unrelated but welcome — wired `htmlFor`/`id` through `ProductFormFields`. All three should-fix items from the previous review are addressed.

### Resolved
- Stale product-context card after editing in the Product tab: `adoptSavedContext` now notifies, `ProjectDetail.handleContextSaved` seeds the cache, and the behaviour is pinned by three tests including the failure path.
- Product-context query now has `staleTime` (5 min) and `retry: false`, matching the "one GET per project open" claim.
- Keystone test asserts the four specific state strings in both directions instead of `innerHTML` inequality.
- Locale key ordering aligned with the surrounding alphabetical levels in all eight files.
- Index-based button lookup in `OverviewTab.test.tsx` replaced with `cardFor()`/`within()`.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so those lines can never pluralise (nit).
- `useTransientFlag`'s "stable identity" comment still overstates what the `useMemo` guarantees (nit, and arguably #288's).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may now be a dead prop (nit).

### New findings
- **`setQueryData` is untyped, so the seeded shape isn't checked** — `ProjectDetail.tsx`'s `queryClient.setQueryData(productContextKey(id), { context })` writes an object literal against a key whose data type TanStack cannot infer, so if `projectsApi.getProductContext` ever returns more than `{ context }` this silently truncates the cache entry with no compile error. Annotate the update with the API response type (or use `queryOptions`/a typed key factory) so the two stay in lockstep.
- **`FieldShell` now emits `htmlFor` unconditionally** — `ProductFormFields.tsx`'s shell points `htmlFor` at `product-context-${field}`, but only `TextField`/`TextAreaField`/`SelectField` render a matching `id`. Any current or future direct use of `FieldShell` with custom children produces a dangling `htmlFor`, which is worse for screen readers than the unlabelled input it replaces. Worth confirming no other caller renders `FieldShell` directly, and/or taking the input id as an explicit prop.
- **No test covers the `ProjectDetail` half of the handover** — the new tests prove `ProductTab` calls back; nothing asserts that the callback actually lands in `productContextKey` and the Overview card re-reads it. Given the bug was precisely a wiring gap between the two owners, one test at that seam would fail-on-revert where the current three would not.

---

perrozzi:

Since the last review the author addressed all three findings from that round: `FieldShell` now takes an explicit required `inputId` (so the `htmlFor` can no longer dangle, and any other caller becomes a compile error), `handleContextSaved` types the cache write via `Awaited<ReturnType<typeof projectsApi.getProductContext>>`, and a new `ProjectDetail.contextHandover.test.tsx` drives the real Product-tab → Overview-card seam including the "no third request" and error-path cases. The delta is clean.

### Resolved
- `FieldShell`'s `htmlFor` is now supplied by the caller as a required `inputId`, so it always points at a control the caller actually rendered.
- `queryClient.setQueryData<ProductContextResponse>(...)` derives the shape from the API function, so `{ context }` and `getProductContext` can no longer drift silently.
- `ProjectDetail.contextHandover.test.tsx` covers the consuming half of the handover (edit in Product tab → count updates on Overview, from cache rather than a refetch) plus the unknown-context path.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so they cannot pluralise (nit).
- `useTransientFlag`'s "stable identity" comment overstates the `useMemo` guarantee (nit, inherited from #288).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may be a dead prop (nit).

### New findings
- (nit) `ProjectDetail.contextHandover.test.tsx` restores `Element.prototype.scrollTo = realScrollTo`, which in jsdom is `undefined` — this leaves an own `scrollTo: undefined` property on the prototype instead of removing it. Harmless today (and the same pattern as `ProductTab.contextSaved.test.tsx`), but `delete Element.prototype.scrollTo` when the original was absent would be truer teardown.
- (nit) `FieldShell` now takes both `field` and `inputId`, where `inputId` is always `fieldInputId(field)` at every call site; the doc comment justifies the redundancy, but a future caller can pass a mismatched pair with no error. Fine as-is given the reasoning is written down.

---

perrozzi:
## Round 2 disposition — all three new findings fixed, head `07700e8`

Thanks — the round-1 stale-card finding was a real defect in my own work that the tests
would not have caught, and round 2's third finding is why: my tests proved the tab
*called* the callback, while the bug was that nothing *consumed* it.

### New findings

**1. `setQueryData` was untyped — fixed.** Correct: TanStack cannot infer a plain key's
data type, so the object literal was checked against nothing. Now
`setQueryData<Awaited<ReturnType<typeof projectsApi.getProductContext>>>`, so the seeded
shape is derived from the call that produces it rather than restated.

**2. `FieldShell` emitted a `htmlFor` it could not verify — fixed.** Confirmed only the
three sibling components render it today, so this was a footgun rather than a live
defect, but the reasoning holds: a dangling `htmlFor` is worse than the unlabelled input
it replaced. `inputId` is now a required prop, so a new caller has to state the id it
actually rendered.

**3. Nothing tested the half that was broken — fixed, and verified fail-on-revert.**
`ProjectDetail.contextHandover.test.tsx` drives the real seam: edit a field on the
Product tab, switch to Overview, read the new count. Plus the request-failed path, where
the card must render no state rather than claim the description is empty.

Removing `onContextSaved={handleContextSaved}` from `ProjectDetail` **fails the new seam
test while all three `ProductTab` tests still pass** — exactly the gap you identified.

One correction: writing that test showed an assertion of mine was wrong and the code was
right. I expected a single `getProductContext` call; there are two, because the Product
tab fetches the record it edits. The test now pins what I actually meant — no
*additional* read on the way back to Overview, which is the difference between handing
the value back and invalidating.

### Carried nits, with reasons

- **`{{total}}` rather than i18next `count`** — deliberate, and it is a gate constraint
  rather than an oversight. A new key called with `count` makes `i18next-parser` write an
  empty `_one`/`_other` family, which then fails `i18n:validate` on values it wrote
  itself. The label:value phrasing reads correctly at 1 in all eight locales. Happy to
  revisit with full plural families as its own change.
- **`useTransientFlag`'s identity comment** — accurate, but that file is #288's and
  already merged; out of this diff.
- **`onDocumentChanged`** — not dead. `ChatTab` passes it to `chatTabHooks`, which calls
  it when a turn changes documents.

### Live verification

Deployed and checked against a real project, not fixtures.

- The chunk carrying `ProjectDetail`, fetched back through the CDN, is **byte-identical**
  to the local build, and carries all the new keys plus `product-context` and the field
  ids.
- All eight catalogues live; the two reworded values differ from English in every locale.
- On a live project with 2 personas and 3 documents — one of them a `custom` doc — the
  cards derive **6 of 11 fields · 2 personas · 1 research · 1 PRD/PR-FAQ**, so the type
  filter is doing its job: 3 documents, and the PRD/PR-FAQ card correctly says 1.
- **The write path, end to end:** saving a previously empty field moved the count
  **6 → 7** in the response that seeds the cache, which is the fix working against the
  real API. The field was restored, so the project is as it was found.

Not verified: no signed-in browser walkthrough. The card states, hints and next-step line
rest on tests plus these API-level checks.

perrozzi:

Since the last review the only change is extraction of the jsdom `scrollTo` stub into a shared `src/test/stubScrollTo.ts` helper, now used by both `ProductTab.contextSaved.test.tsx` and `ProjectDetail.contextHandover.test.tsx`. That closes the teardown nit from my previous round; the delta is clean and I have no new findings.

### Resolved
- Teardown now restores the *absence* of `Element.prototype.scrollTo` (`delete` when it never existed) instead of leaving an own `scrollTo: undefined`, and the two test files share one helper rather than duplicating the pattern.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so they cannot pluralise (nit).
- `useTransientFlag`'s "stable identity" comment overstates the `useMemo` guarantee (nit, inherited from #288).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may be a dead prop (nit).
- `FieldShell` takes both `field` and `inputId` where the latter is always `fieldInputId(field)`; a mismatched pair is still possible, accepted given the documented reasoning (nit).

---

perrozzi:
## Rounds 3 and 4 disposition — converged, loop stopped. Head `01e0a44`

Round 4 reports no new findings, so this closes the loop per the #166 / #277 / #287
convention.

### Round 3's two nits

**`scrollTo` teardown — fixed.** Correct, and slightly worse than described:
`Element.prototype.scrollTo` is undefined in jsdom, so reassigning it left an own
`scrollTo: undefined` property that had never existed. Now `stubElementScrollTo()` in
`src/test/`, which `delete`s the property when it was absent. Extracted rather than
inlined twice because the underlying trap recurs for anything that scrolls a container
in an effect — without the stub React renders the subtree as an empty div, and
"was not called" assertions pass vacuously. #288's `reportGuard` test keeps its inline
copy; it is already merged.

**`field` + `inputId` redundancy — declined,** which matches the review's own read. The
pair is what lets a caller state the id it actually rendered instead of trusting the
shell to guess. A mismatched pair remains possible; a silently dangling `htmlFor` does
not.

### The three carried nits, unchanged

- **`{{total}}` over i18next `count`** — a gate constraint, not an oversight: a new key
  called with `count` makes `i18next-parser` write an empty `_one`/`_other` family, and
  the gate then fails on values it wrote itself. The label:value phrasing reads correctly
  at 1 in all eight locales. Worth revisiting as its own change with full plural families.
- **`useTransientFlag`'s identity comment** — accurate, but #288's file and already
  merged.
- **`onDocumentChanged`** — not dead: `ChatTab` passes it to `chatTabHooks`, which calls
  it when a turn changes documents.

### Where this landed

Four review rounds found **one real defect in my own work** — the product context had two
owners that never resynced, so editing in the Product tab left the Overview card
reporting the count from page load. That is the same class of defect this PR set out to
remove, and it would have shipped: my tests proved the tab *called* its callback while
the bug was that nothing *consumed* it. The seam test added in round 2 fails on revert
where those three stay green.

**Gates on `01e0a44`:** tsc clean · eslint clean · **1957 tests / 129 files** · i18n
missing/empty/untranslated all 0 with the unreferenced count unchanged at 650 · the
pre-existing ja/zh/ko `extra: 12` untouched · `typecheck:tests` 74 against its 85
baseline.

**Live:** deployed and verified against a real project. The chunk carrying
`ProjectDetail` fetched back through the CDN is byte-identical to the local build; all
eight catalogues are live with the two reworded values differing from English everywhere.
On a project with 2 personas and 3 documents — one a `custom` doc — the cards derive
6 of 11 fields · 2 personas · 1 research · 1 PRD/PR-FAQ, so the type filter holds. Saving
a previously empty field moved the count 6 → 7 in the response that seeds the cache,
which is the round-1 fix working against the real API; the field was restored afterwards.

**Not verified:** no signed-in browser walkthrough. The numbered cards, hints and
next-step line rest on tests plus these API-level checks.

Ready for a maintainer.

perrozzi:

The only change since my last review is a one-line swap in `voc-datalake/frontend/src/test/stubScrollTo.ts`: the teardown now uses `Reflect.deleteProperty(Element.prototype, 'scrollTo')` instead of a `delete` through an inline type assertion, with a comment explaining why. Behaviour is identical, the assertion is gone, and nothing else in the PR is touched — the delta is clean and I have no new findings.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so they cannot pluralise (nit).
- `useTransientFlag`'s "stable identity" comment overstates the `useMemo` guarantee (nit, inherited from #288).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may be a dead prop (nit).
- `FieldShell` takes both `field` and `inputId` where the latter is always `fieldInputId(field)`; a mismatched pair remains possible, accepted given the documented reasoning (nit).

---

perrozzi:

Since my last review the change is purely presentational: the numeric badge over each card icon (`aria-hidden` chip + `sr-only` step label) has been replaced with the position rendered as literal heading text (`1. Product / Service Description`), the now-unused `overview.stepNumber` key was removed from all eight locale catalogues, and the position test was retargeted at the heading text. Previous nits remain as previously noted.

### Resolved
- The dual-channel position (decorative badge + parallel hidden label) is gone; one heading string now serves everyone, and the removed `stepNumber` key is cleanly deleted from all 8 locales with no dangling reference.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so they cannot pluralise (nit).
- `useTransientFlag`'s "stable identity" comment overstates the `useMemo` guarantee (nit).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may be a dead prop (nit).
- `FieldShell` takes both `field` and `inputId` (nit, accepted).

### New findings
- `OverviewTab.tsx`: the ordinal is now meaningful text but styled `text-gray-400` (#9ca3af ≈ 2.6:1 on white) — below the 4.5:1 contrast minimum for small text. Since it no longer has a hidden counterpart, this is the only channel for the sequence; `text-gray-500` would clear the bar (should-fix, nit-sized).
- `OverviewTab.test.tsx`: the comment says "Asserting on the accessible name is what pins that", but `cardTitlesInOrder()` reads `h.textContent`, not the accessible name — the assertion is still correct and would catch a regression to a hidden span, only the comment is inaccurate (nit).

I checked the knock-on: `ProjectDetail.contextHandover.test.tsx`'s `findByText('Product / Service Description')` still matches, because Testing Library's `getNodeText` only joins direct text-node children and the `1.` now lives in a child `<span>`. Otherwise the delta is clean.

---

perrozzi:

Since my last review the delta is presentation-only: the ordinal span moved from `text-gray-400` to `text-gray-500`, the "None yet"/state line and the disabled-message line were raised the same way, and the misleading "accessible name" comment in `OverviewTab.test.tsx` was corrected to describe what the assertion actually reads (`textContent`, exact match). Both of my previous new findings are addressed; the standing nits are unchanged.

### Resolved
- `OverviewTab.tsx`: ordinal now `text-gray-500` (≈4.8:1), clearing the 4.5:1 minimum for the only channel carrying the sequence — and the state/disabled lines were raised with it, which is more than I asked for.
- `OverviewTab.test.tsx`: comment now accurately describes the exact-`textContent` assertion.

### Still open
- `overview.state.personas` / `research` / `documents` use `{{total}}` rather than i18next `count`, so they cannot pluralise (nit).
- `useTransientFlag`'s "stable identity" comment overstates the `useMemo` guarantee (nit).
- `onDocumentChanged` in `TabContent.tsx`/`ProjectDetail.tsx` may be a dead prop (nit).
- `FieldShell` takes both `field` and `inputId` (nit, accepted).

### New findings
None. One cosmetic consequence worth a glance in the browser: the "None yet" state line is now the same `text-gray-500` as the card description directly above it, so an unstarted card's state no longer reads as visually distinct from its blurb — purely a hierarchy judgement, not a defect, and readability was the right thing to prioritise.

---

perrozzi:
## Rounds 5 and 6 disposition — one real regression of mine, fixed. Head `61b65f2`

Round 6's contrast finding was correct, and it was a regression **introduced by my own change** in the
previous round. Worth saying plainly: I moved the step number out of a badge partly on accessibility
grounds, then rendered it at a contrast ratio that fails the same standard.

**Verified independently rather than taken on trust.** On white: `gray-400` (#9ca3af) is **2.54:1**,
`gray-500` (#6b7280) is **4.83:1**, against the 4.5:1 minimum for small text. `green-700` and `amber-700`
already pass at 5.02:1 and were left alone.

It matters most here for the reason the review gave: with the hidden counterpart gone, the heading text is
the **only** channel carrying the sequence, so making it hard to read would have undone the change it was
part of.

**Raised all three `gray-400` uses on the card**, since each carries information rather than decoration:

| | |
|---|---|
| the ordinal | the finding |
| the empty-state label — "None yet", "Not run yet" | same defect, also mine, same round of work |
| the disabled explanation — "Need at least 2 documents" | **pre-existing**, raised with the rest: it is the last line on a card that should be hard to read |

`OverviewTab.tsx` now has no `gray-400` at all, and one muted colour that matches its own description text.

**Second nit — fixed.** You were right that the comment misdescribed the assertion: `cardTitlesInOrder()`
reads heading text content, not the accessible name. The comment now says so, and explains why an exact
match still catches a regression to a hidden span — the number would leave the heading entirely.

Thanks also for checking the knock-on effect on `findByText('Product / Service Description')`. That was the
non-obvious risk in moving the ordinal into a child `<span>`, and it is the check I would have wanted a
reviewer to make.

**Carried nits unchanged**, all previously dispositioned: `{{total}}` over i18next `count` (a gate
constraint — a new key called with `count` makes the parser write an empty plural family that then fails on
its own output), `useTransientFlag`'s comment (#288's file), `onDocumentChanged` (not dead — `ChatTab` →
`chatTabHooks`), and `FieldShell`'s `field` + `inputId` pair (accepted by the review).

**Gates on `61b65f2`:** tsc clean · eslint clean · **1957 tests / 129 files** · i18n missing/empty/
untranslated all 0, unreferenced unchanged at 650.

**Live:** deployed; the main chunk reports `APP_VERSION` `61b65f2` and the ProjectDetail chunk fetched back
through the CDN is byte-identical to the local build, carrying the raised ordinal class.

Ready for a maintainer.

perrozzi:
## Signed-in browser walkthrough — done. Retracting the "not verified" caveat

Every earlier comment on this PR, and the body, said the card states rested on tests plus API-level checks
with no browser walkthrough. That is now discharged, on the deployed build at `61b65f2`.

### The three project states, in a real signed-in browser

| Project | Rendered |
|---|---|
| untouched | banner "Next step: Describe your product or service."; "Not described yet", "None yet" ×2, "Not run yet"; **both** upstream hints; Remix disabled with "Need at least 2 documents" |
| 4 personas, **0 research**, 4 documents | banner "Next step: **Run research — your personas can ground it.**"; "Personas created: 4" and "PRD / PR-FAQ documents: 4" in green, "Not run yet" in grey; **no** hints, since personas exist; Remix enabled |
| everything has output | **banner absent**, all four state lines present |

That middle project is a real one on the deployment, and it is the case this PR exists for: four documents
generated with no research behind them, which is what the old order (PRD/PR-FAQ *above* Run Research)
produced. The reordered tab sends it to research instead of to more documents.

Card headings expose `1. Product / Service Description` … `5. Remix Documents` as their accessible names, so
the sequence reaches assistive technology through the heading rather than a parallel hidden label.

### The stale-card fix, proven end to end

Typed into Non-goals on the Product tab, tabbed out to trigger the autosave, switched to Overview: card 1
moved **"6 of 11 fields filled" → "7 of 11"**. Network across the whole visit: **two GETs and one PUT, with no
third GET** — the cache seeding rather than an invalidate-and-refetch. Clearing the field moved it back to 6,
and a fresh server read confirms 6 of 11 with an empty `non_goals`, so the project is exactly as found.

Also confirmed live: every product field renders `id="product-context-<field>"` with a matching `label[for]`
— the labelling fix from round 2.

**Console on the driven page: 0 errors, 0 warnings.**

### Two cosmetic notes, neither a defect

- The hierarchy point from round 7, now seen with eyes: where a step *has* output the green state separates
  cleanly from the grey description, but on an unstarted card grey "None yet" does sit under the grey
  description and read as a continuation. Mild, and only in the empty state. Happy to differentiate it if a
  maintainer prefers.
- The Remix button reads `Configure &  Select & Remix` — double space, two ampersands, from concatenating
  `configureLabel` + `buttonLabel`. **Pre-existing and unrelated to this PR**, but visible while I was
  looking.

Nothing outstanding from my side. Head `61b65f2`.

PR #290, opened 2026-08-08
fix(ui): surface the prototype link lifetime and renew it before it lapses

perrozzi:
## Summary

A prototype is served from `/prototypes/*`, restricted by a CloudFront trusted key group, so the
browser needs a signed URL. The API mints one on every `GET /projects/{id}` and never trusts a stored
value — which makes that URL a **session-scoped view credential with a deadline**, and deliberately
so: the signer's TTL is matched to the Cognito token lifetime (1h access/id token vs the signer's
3600s fallback) precisely so a link cannot outlive the session that asked for it.

Nothing in the UI said so. A reviewer copies the URL out of "Open in new tab", passes it on, and it
dies inside the hour with no explanation at either end; a tab left open past expiry then clicked gets
a bare CloudFront 403.

**So the bug is not the lifetime — it is that a short-lived credential was presented as a durable
link.** This states the lifetime and keeps the links alive while the page is open, rather than
building a sharing feature the product does not intend to offer. Frontend only: no endpoint change,
no CDK change.

Implements **U4** from `analysis/UI-ROUND2-GITHUB-ISSUES.md`, and closes **U7's remaining acceptance
criterion**, which #288 did not.

## What changed

| File | |
|---|---|
| `components/prototypeLinkLifetime.ts` (new, pure) | Reads the deadline off the URL's own `Expires`; picks the soonest across a project's prototypes; decides the refresh delay; `unsignedUrlKey` splits address from credential |
| `pages/ProjectDetail/useProjectData.ts` | Timer re-signs via `invalidateQueries(['project', id])` five minutes before expiry, re-arming off each fresh URL |
| `pages/ProjectDetail/DocumentsTab.tsx` | `LinkLifetimeNote` states the window and that the link is session-scoped; prototype pane height (U7 AC 4) |
| `components/PrototypeRenderer.tsx` | `useLoadedUrl` keeps the frame on the URL it loaded with |
| `public/locales/*/projectDetail.json` | 3 keys × 8 locales |

### Why the deadline comes from the URL, not a constant

`CDN_SIGNED_URL_TTL_SECONDS` is **not set in `api-stack.ts`**, so 3600s exists only as a Python-side
fallback (`cloudfront_signing.FALLBACK_TTL_SECONDS`). A number mirrored in the frontend would be a
guess that silently diverges the day it is configured. The signature is a canned policy, so `Expires`
is already in the URL the page holds.

### Why renewal is ahead of expiry, not on click

"Open in new tab" and "Download .html" are plain anchors: a click navigates immediately and nothing
can fetch a replacement first. Rewriting them as buttons that fetch then `window.open` would trade a
403 for a popup blocker, and `download` cannot be triggered that way at all.

This is not the only protection — `['project', id]` sets no `staleTime`, so it inherits
refetch-on-window-focus and tabbing away and back already re-signs. The timer covers the case with no
user interaction to hook: **a tab left focused and untouched past the hour.**

### Why the refreshed URL is deliberately NOT given back to the iframe

Changing an iframe's `src` **reloads it**, which would reset a reviewer several screens into the
prototype — hourly, invisibly. It is also unnecessary: a signature governs the *request*, and once the
document is loaded the credential behind it has no further bearing on it. **Only a new request needs a
live signature**, which means the anchors, not the frame.

So the frame follows the address and ignores the credential. A genuinely different document still
reloads, because that is a different address.

That logic sits in `HtmlPrototypeFrame` rather than in the caller because "never swap a loaded
iframe's `src`" is a property of the frame — and the **Prioritization page renders prototypes through
the same component**, so it inherits the fix instead of needing a second copy.

### U7 AC 4 — the pane was not fixed by #288

The document detail pane was a flat `lg:min-h-[500px]`, so a whole generated application previewed in
~430px whether the monitor was 900px or 1440px tall. #288 stopped the jobs panel pushing it *down the
page*, but a taller viewport still bought it nothing. Prototypes now get `lg:min-h-[70vh]` —
deliberately short of the full viewport so the pane still fits above the fold rather than introducing
a scroll — while prose documents keep the shorter pane, where extra height is only whitespace under a
short PRD.

## Two of U4's five ACs were already satisfied, and are untouched

Recorded so a later reader does not "implement" them by adding something in order to remove it:

- **"Nothing presents the URL as a durable share link"** — already true. There is no Share and no
  Copy-link control; clipboard use under `pages/ProjectDetail/` is `McpAccessTab` only, for access
  tokens. Prototypes expose exactly two plain anchors.
- **"Download cannot silently fail"** — the `download` attribute is honoured because `/prototypes/*`
  is a behavior on the **same distribution as the SPA**, hence same-origin. (Worth knowing: it would
  break silently if prototypes ever moved to their own distribution.)

## Known limitation, documented at the call site

The label samples the clock once via `useState(() => Date.now())` — it must, because `Date.now()` in a
render body is impure and eslint rejects it. That sample ages while the component stays mounted, so a
link that lapses with **no refetch landing at all** will not flip the label by itself. That needs both
the timer and the focus refetch to fail (e.g. offline), and it errs toward saying nothing rather than
wrongly claiming expiry.

## Testing

- **31 new tests.** `prototypeLinkLifetime.test.ts` (22) covers the seconds→ms conversion, every
  unreadable-deadline case, soonest-of-several, the refresh floor and the already-expired path.
  `DocumentsTab.prototypeLink.test.tsx` (9) covers the label, the expired label, the legacy prototype
  with no signature, the frame/anchor split, and both pane heights.
- **The freeze test was verified to fail on revert**: pointing `src` back at the live URL fails
  exactly that test and leaves the other eight green.
- No fake timers. The arithmetic is a pure module and the expiry branch is chosen by putting `Expires`
  either side of the real clock — fake timers have leaked across files in this suite before.
- **Gates:** eslint clean, `typecheck:all` clean, **1,988 frontend tests pass**. The i18n gate is at
  the `development` baseline (its only failures are the 12 pre-existing ja/zh/ko `categories.json`
  extras); the three new keys report missing 0, extra 0, empty 0, untranslated 0 across all 8 locales.

## Not verified

**No deployment and no signed-in browser walkthrough at time of opening.** The expiry paths are
exercised from synthetic `Expires` values, not a real lapsed CloudFront signature. Deploying and
live-checking next; will report back on this PR.


perrozzi:

This is a frontend-only change that surfaces the CloudFront signed-URL lifetime on the prototype preview, re-signs links ahead of expiry by invalidating `['project', id]`, freezes the iframe `src` so re-signing does not reload a loaded prototype, and bumps the prototype pane height. The arithmetic is nicely factored into a pure, well-tested module (`prototypeLinkLifetime.ts`), the i18n keys are complete across all 8 locales, and the reasoning is documented at the call sites. One behavioral edge in the new `useLoadedUrl` freeze looks like a real regression, and the user-facing string has a11y/locale problems.

### Blockers

- `voc-datalake/frontend/src/components/PrototypeRenderer.tsx` — `useLoadedUrl` pins the frame to whatever URL it first saw for a given address, including an **already-expired** one, with no recovery path. If the component mounts against stale cached project data (persisted/stale TanStack cache, a machine resumed after suspend, or a refetch that failed once), the iframe loads a dead signature, renders CloudFront's 403, and then *ignores* the fresh URL that the 30 s floor refresh delivers — the pane stays broken until a full route change/remount. The same applies if the user reloads the frame itself. Your own test (`DocumentsTab.prototypeLink.test.tsx`, "reports a lapsed link…" / the freeze test) renders an expired URL and pins it, so the current suite locks this in. Suggested fix: only freeze while the loaded URL is still live, e.g. adopt `url` when `isExpired(signedUrlExpiresAt(loaded), Date.now())` (and/or on the iframe's `onError`). Note this component is also used by the Prioritization page, so the bad state is inherited there too.

### Should fix

- `DocumentsTab.tsx` (`LinkLifetimeNote`) — the "tied to your session / not a permanent share link" message, which is the entire point of the PR, lives only in a `title` attribute. That is hover-only (invisible on touch), inconsistently announced by screen readers, and not keyboard-reachable. Make it visible text next to the timestamp, or attach it via `aria-describedby` on the anchors so the warning reaches the user who is about to copy the link.
- `DocumentsTab.tsx` — `format(new Date(expiresAt), 'HH:mm')` is neither locale-aware (24 h clock forced on `en`, no locale passed to date-fns for ja/zh/ko/de/…) nor unambiguous: with no date component, a link expiring at 00:20 reads as already past at 23:50. Prefer a locale-aware time format (`p` with the active date-fns locale, matching how the rest of the app formats times) or a relative "expires in ~55 min", which sidesteps both problems.
- `useProjectData.ts` — the refresh timer itself has no test. `refreshDelayMs`/`earliestPrototypeExpiry` are well covered, but nothing asserts that the effect actually schedules `invalidateQueries(['project', id])`, re-arms off the new URL, or does *not* set a timer when there is no readable deadline. Reverting the whole `useEffect` leaves all 31 new tests green. A single hook test with fake timers in its own file (or injecting the clock/delay) would close that.
- `DocumentsTab.tsx` — `mountedAt` never advances, so the documented limitation is broader than "offline": any long-mounted pane will keep displaying "Link valid until 14:05" well past 14:05 whenever a refetch does not land. A `setTimeout` in an effect that bumps a state counter at `expiresAt` would flip the label without a render-body `Date.now()` and without a ticking interval.

### Nits

- `prototypeLinkLifetime.ts` — the JSDoc on `readExpiresParam` is a copy of `signedUrlExpiresAt`'s and describes "epoch ms" for a function returning the raw query string.
- `prototypeLinkLifetime.ts` — stray `ponytail:` marker at the start of the comment inside `refreshDelayMs`; looks like a leftover editing token.
- `unsignedUrlKey` is the load-bearing function for the freeze behavior but has no direct unit test in `prototypeLinkLifetime.test.ts` (query-order differences, trailing slash, unparseable fallback); it is only exercised indirectly through the component.
- `DocumentsTab.prototypeLink.test.tsx` asserts on Tailwind class literals via `container.querySelector('.lg\\:col-span-2')`; that will break on any layout refactor unrelated to link lifetime. A `data-testid` on the detail pane would be sturdier.
- `useLoadedUrl` treats `url === undefined` as a different address, so a transient undefined would drop the frame into the legacy `srcDoc`/empty branch; harmless with the current cache behavior, but worth a guard or a comment.

---

perrozzi:
Thanks — the blocker was real and it was mine. Addressed in `1a34a28`, along with all four "should fix" items and four of the five nits. Point by point.

## Blocker — `useLoadedUrl` could pin an expired URL: **fixed**

Correct diagnosis, including the part I had not seen: the freeze had no upper bound, so a frame mounting against an already-expired signature loaded a dead URL, rendered CloudFront's 403, and then *ignored* the live replacement because the address had not changed. And yes, the Prioritization page inherited it.

Freezing is only ever a courtesy to a **working** document, so it now ends when the document stops working:

```ts
const loadedIsDead = useDeadlinePassed(signedUrlExpiresAt(loaded))
const adopt = loadedKey !== nextKey || (loadedIsDead && url !== loaded)
```

The `url !== loaded` guard matters: with no replacement on offer there is nothing to adopt, and swapping a dead URL for itself would reload the frame every render.

Pinned by a new test that fails without the fix — mount on an expired URL, hand over a live one, assert the `src` changes.

## Should fix

**1. Session warning was `title`-only — fixed.** You are right that this is the one sentence the whole change exists to communicate, and a tooltip is a poor place for it. It is now visible text beside the timestamp, and both anchors carry `aria-describedby` pointing at it so it reaches a screen-reader user at the moment they are about to activate the link. The string was reworded across all 8 locales to read as an inline clause rather than a tooltip sentence, and `aria-describedby` is omitted rather than left dangling when there is no readable deadline to describe (covered by a test).

**2. `HH:mm` not locale-aware and ambiguous — fixed.** New `formatExpiry` uses `Intl` with the active language, so `en-US` gets a 12-hour clock and `de` a 24-hour one, and it adds the date whenever the deadline is not on today's local date — the midnight case you described. That matches the convention #288 adopted for job timestamps. `Intl` rather than date-fns locales because the app loads no date-fns locale registry, so this needs no new imports.

**3. The refresh timer had no test — fixed, and this was the most useful finding.** You are right that it was the one part of the diff that could be deleted wholesale with every test still green, which is not the standard the rest of it was held to. `useProjectData.prototypeRefresh.test.tsx` now covers scheduling, the invalidation key, re-arming off the replacement URL, and *not* scheduling when there is no deadline. **Two of its five cases fail when the effect is disabled** — I checked rather than assuming. Fake timers are confined to that file and the hook's own, torn down after each test.

**4. `mountedAt` never advanced — fixed by the same hook as the blocker.** Both needed a clock that moves: the label must stop promising a closed window, the frame must stop holding a dead signature. `useDeadlinePassed` samples the clock in state, advances it from one timer, and derives the answer during render — the last part is not tidiness, because `setState` straight from an effect is its own eslint error and my first version of the hook tripped it. The delay is clamped rather than branched, which also closes the window where time crosses the deadline between a render and its effect.

## Nits

- **Duplicated JSDoc on `readExpiresParam`** — fixed, and worth owning: I had *tried* to move that block earlier in the session and left both copies, without re-reading to check. You saw the result.
- **`ponytail:` marker** — you read it correctly as a stray token. It is a personal annotation convention that has no business in this repo; it now reads as plain English.
- **`unsignedUrlKey` untested** — fixed. Cases for re-signing, query order, different document, different host, trailing slash, and the unparseable fallback. It is load-bearing for the freeze and should have had these from the start.
- **Tailwind class selector** — fixed, `data-testid="document-detail-pane"`.

### One nit deliberately not changed

**`url === undefined` treated as a different address.** Left adopting, on purpose. `HtmlPrototypeFrame` cannot distinguish "same document, signing momentarily unavailable" from "different document that has no URL" without being handed document identity, and refusing to adopt would show the *previous* document's prototype after a switch — worse than a blank frame. Noted in the code rather than papered over. Happy to add a `documentKey` prop if you would rather make the reset explicit.

## Gates

eslint clean, `typecheck:all` clean, **2,014 tests pass** (up from 1,988), i18n gate unchanged at the `development` baseline.

Still **not verified: no deployment, no signed-in browser walkthrough.** Deploying next and I will report the result here.

perrozzi:

The push replaces the hover-only `title` with visible text plus `aria-describedby`, swaps `format(..., 'HH:mm')` for a locale-aware `formatExpiry`, adds a `useDeadlinePassed` hook so the label flips itself at the deadline, releases the frozen iframe URL once it is dead, adds tests for the refresh timer, `unsignedUrlKey` and `formatExpiry`, and a `data-testid` for the pane. Every previous finding is addressed — but the way the freeze release was implemented reintroduces the exact reload the freeze exists to prevent.

### Resolved
- `useLoadedUrl` no longer pins the frame to an expired URL at mount (`PrototypeRenderer.tsx`, plus a regression test).
- The session-scoped warning is now visible text and referenced by `aria-describedby` on both anchors, with a guard against a dangling id.
- `formatExpiry` is locale-aware and disambiguates across midnight; tested.
- `useProjectData.prototypeRefresh.test.tsx` covers scheduling, non-scheduling and re-arming — the effect is no longer deletable-with-green-suite.
- `mountedAt` staleness fixed by `useDeadlinePassed`; JSDoc on `readExpiresParam`, the `ponytail:` marker, `unsignedUrlKey` tests and the Tailwind-class selector are all cleaned up.

### New findings

**Blocker — `voc-datalake/frontend/src/components/PrototypeRenderer.tsx`: the release condition fires on a *successfully loaded* frame, so the hourly reload is back.** `loadedIsDead` becomes true at the loaded URL's own deadline regardless of whether the document loaded fine. Timeline: t+0 loads url1 (expires t+60m); t+55m the refresh delivers url2 and the frame correctly keeps url1; **t+60m** `useDeadlinePassed` flips, `url !== loaded`, so `adopt` is true and `src` becomes url2 — the iframe reloads and the reviewer loses their place, once per hour, invisibly. That is precisely what the PR argues must never happen ("a signature governs the *request*; once the document is loaded the credential behind it has no further bearing on it"). The release is only justified when the frame never got a working document. Gate it on load outcome — e.g. an `onLoad` flag on the iframe (or `onError`) and only adopt while that flag is false — rather than on the deadline alone. Note the existing freeze test cannot catch this: it uses live URLs and no timers, so add a case where a *loaded* URL passes its deadline and the src must stay put.

**Should fix — `DocumentsTab.tsx`: `truncate` can hide the warning that the PR exists to show.** The deadline text and the "tied to your session, not a share link" hint share one `truncate` span inside a `text-xs` flex row next to three action links; in a narrow pane the hint is the part that gets clipped, putting the message back in the "not actually visible" category (screen readers still get it via `aria-describedby`, sighted touch users do not). Let it wrap (or drop `truncate` on the hint) instead.

### Nits
- `LINK_LIFETIME_NOTE_ID` is a module-level literal; `useId()` would avoid a duplicate-id hazard if a second prototype pane is ever rendered on the same screen.
- `formatExpiry` passes `i18n.language` straight into `Intl.DateTimeFormat`, which throws `RangeError` on a malformed tag and would take down the pane; a `try/catch` falling back to `undefined` is cheap insurance.
- `useLoadedUrl` still treats `url === undefined` as a different address (previous nit), so a transient undefined drops to the legacy branch.

---

perrozzi:
Caught cleanly — and the blocker is a regression I introduced in round 1, not a pre-existing gap. Fixed in `73f30e2`.

## Blocker — the release condition reintroduced the hourly reload: **fixed**

Your timeline is exactly right. Releasing on the loaded URL's **deadline** meant a frame that loaded fine at t+0 correctly kept url1 when url2 arrived at t+55m, and then swapped at t+60m — reloading the prototype under the reviewer once an hour, which is the very thing the freeze exists to prevent, five minutes late.

The deadline was the wrong discriminator. A signature governs the *request*, so once the document is in the browser its deadline is irrelevant. The only URL worth escaping is **one that never had a chance of loading**. So the frame now records whether the URL was already expired at the moment it took it:

```ts
const incomingIsDead = useDeadlinePassed(signedUrlExpiresAt(url))
const [loaded, setLoaded] = useState(() => ({ url, deadOnArrival: incomingIsDead }))
const adopt = addressChanged || (loaded.deadOnArrival && url !== loaded.url)
```

A URL that was live when adopted is kept for good; the mount-against-stale-cache case from round 1 still recovers as soon as a replacement lands, because that URL was dead on arrival.

I went with this rather than the `onLoad`/`onError` flag you suggested for one reason: **an iframe fires `onLoad` for CloudFront's 403 page too** — the error document loads successfully — and `onError` does not fire for HTTP error statuses. So a load-outcome flag would go true on exactly the broken load it needs to detect, and the blocker would come straight back. Adoption-time expiry is the same signal without needing to interpret what loaded. If you would rather have the load event as a belt-and-braces second condition, happy to add it.

**Both behaviours are now pinned independently, and I verified each fails under the other's implementation** rather than assuming:

- new `keeps a URL that loaded successfully even after its signature expires` → **fails** under the deadline-based condition
- existing `lets go of an expired URL as soon as a live replacement arrives` → still passes

You were right that the old freeze test could not catch this: it used live URLs and no timers. The new one advances past the loaded URL's deadline, with fake timers scoped to that test and torn down in a `finally`.

## Should fix — `truncate` could clip the hint: **fixed**

Fair, and a bit ironic: the clipped end was the hint itself, so a narrow pane would silently put the warning back out of sight for sighted users while screen readers still got it via `aria-describedby`. It wraps now (`items-start`, no `truncate`).

## Nits

- **`useId()`** — done. The note's id is minted per pane, so two prototype panes on one screen cannot collide.
- **`Intl` RangeError** — done, with a test. Worth noting this was reachable rather than theoretical: the locale comes from i18next's detection chain (querystring, cookie, `navigator.language`), so a malformed tag is user-supplied, and a throw there would take down the document pane rather than mis-format one timestamp.
- **`url === undefined` treated as a different address** — unchanged, as dispositioned in round 1: `HtmlPrototypeFrame` cannot distinguish "same document, signing momentarily unavailable" from "different document with no URL" without being handed document identity, and refusing to adopt would show the *previous* document's prototype after a switch. Still happy to add a `documentKey` prop if you would prefer the reset to be explicit rather than inferred.

## Gates

eslint clean, `typecheck:all` clean, **2,016 tests pass**, i18n gate unchanged at the `development` baseline.

Deploying now; I will post the live result here.

perrozzi:

The push replaces the deadline-based freeze release with a `deadOnArrival` flag captured at adoption time, adds a regression test for a loaded-then-expired URL, drops `truncate` from the note, switches the note id to `useId`, and wraps `Intl.DateTimeFormat` in a try/catch with a test. All previous findings are addressed, and the new `useLoadedUrl` logic looks correct — a URL that loaded successfully is never swapped, while a dead-on-arrival one is released as soon as any replacement appears.

### Resolved
- Blocker: `PrototypeRenderer.tsx` now gates the release on `deadOnArrival` rather than the loaded URL's deadline, so the hourly reload is gone; covered by the new "keeps a URL that loaded successfully even after its signature expires" test.
- `DocumentsTab.tsx`: the lifetime hint wraps instead of being clipped by `truncate`.
- Nit: `LINK_LIFETIME_NOTE_ID` replaced by `useId()` threaded to both anchors.
- Nit: `formatExpiry` no longer throws on a malformed language tag, with a test.

### Still open
- `useLoadedUrl` treats `url === undefined` as a different address, so a transient undefined still drops the frame to the legacy branch (nit).

### New findings
- Nit — `voc-datalake/frontend/src/components/prototypeLinkLifetime.test.ts`: the new malformed-locale test was inserted between the midnight-ambiguity JSDoc and the `includes the date when the deadline falls on another day` test it documents, so that comment now reads as belonging to the wrong test. Move it back down.
- Nit — `voc-datalake/frontend/src/pages/ProjectDetail/DocumentsTab.prototypeLink.test.tsx`: the file header still asserts "Deliberately no fake timers … nothing here depends on timer control", which the new test contradicts. Amend the header to say timers are used in exactly one scoped test.
- Nit — same file: worth confirming the new test fails when `deadOnArrival` is swapped back for the loaded URL's deadline (the PR verified the *other* freeze test on revert, not this one), otherwise it risks being a green-either-way test.

---

perrozzi:

Two doc/test-organization touch-ups since the last review: the `includes the date when the deadline falls on another day` test moved back under its midnight-ambiguity JSDoc, and the `DocumentsTab.prototypeLink.test.tsx` header now admits that one test scopes its own fake timers. Both previous nits are addressed; no code changed, so nothing else moves.

### Resolved
- `prototypeLinkLifetime.test.ts`: midnight-ambiguity JSDoc now sits directly above the test it documents.
- `DocumentsTab.prototypeLink.test.tsx`: header no longer claims "no fake timers" while the file uses them in one scoped test.

### Still open
- `useLoadedUrl` treats `url === undefined` as a different address, so a transient undefined drops the frame to the legacy branch (nit).
- Confirm the `keeps a URL that loaded successfully even after its signature expires` test actually fails when `deadOnArrival` is swapped back for the loaded URL's deadline (nit).

No new findings — the incremental changes are clean.

---

perrozzi:
Round 3 is nit-only, so **stopping the review loop here** per the convention this repo settled on in #166 (no code churn on a nit-only round). Two exceptions made, because they are comment-only and both are inaccuracies the *previous* round introduced — a comment that describes the wrong thing is worse than no comment. `a9bce25`.

- **Misplaced JSDoc in `prototypeLinkLifetime.test.ts`** — correct, and my doing: the malformed-locale test landed between the midnight-ambiguity comment and the test it documents. Moved to the end of the block.
- **Stale file header in `DocumentsTab.prototypeLink.test.tsx`** — also correct. It claimed "deliberately no fake timers", which the new test contradicts. It now says timers are used in exactly one test and why, so the next reader does not take the header as licence to reach for them freely.

### On confirming the new test fails on revert

Already done, before the round-2 push — the check is in that reply, but here is the evidence rather than the claim. Swapping `loaded.deadOnArrival` back for the loaded URL's deadline:

```
× keeps a URL that loaded successfully even after its signature expires
✓ lets go of an expired URL as soon as a live replacement arrives
  Tests  1 failed | 12 passed (13)
```

So it is not green-either-way, and the two behaviours are pinned independently: each fails under the other's implementation. Both directions were checked because this blocker *was* a regression from my own previous fix, which is exactly the situation where one test passing tells you nothing.

### `url === undefined` — closing this one as won't-fix

Third time raised, so let me be decisive rather than leave it drifting. `HtmlPrototypeFrame` receives a URL and HTML, not document identity, so it genuinely cannot distinguish:

- same document, signing momentarily unavailable → keeping the old URL would be better
- different document that has no signed URL (a legacy inline prototype) → keeping the old URL shows **the previous document's prototype**, which is a correctness bug, not a cosmetic one

Given the choice, adopting is the safe failure. And the ambiguity is only resolvable by giving the component a `documentKey` prop, which is a wider API change than the nit warrants — the API reader ( `_with_signed_prototype_url` ) either derives the URL from the ids on every read or drops it entirely, so a same-document transient undefined requires signing itself to break mid-session. Documented in the code. If a maintainer would rather have the explicit prop, I will add it, but I would not spend the API surface unprompted.

### State

- eslint clean, `typecheck` clean, **2,016 tests pass**, i18n gate unchanged at the `development` baseline.
- Rounds 1 and 2 each found a real blocker and both are fixed; round 3 found no code issues.
- Deploying now — I will post the live-verification result here, since "no deployment, no browser walkthrough" has been the standing gap in this PR from the start.

perrozzi:
## Deployed and live-verified — closing the standing gap on this PR

Deployed `a9bce25` to the shared environment and walked it in a signed-in browser. Everything below is observed, not inferred.

### The bundle users receive actually contains the change

Fetched all **33 JS chunks through CloudFront** (not from `dist/`) and grepped them, so this is about the served artifact:

| marker | result |
|---|---|
| `linkExpiryHint` locale key | found |
| `deadOnArrival` freeze flag | found |
| `lg:min-h-[70vh]` (U7 AC 4) | found |
| `document-detail-pane` test id | found |
| `aria-describedby` wiring | found (2 chunks) |
| `Intl` `timeStyle` formatting | found |

Chunks carrying the **old** `lg:min-h-[500px]` prototype height and not the new one: **0**. Locale catalogues served correctly too — the three new keys present in `en`, and the Japanese hint round-trips intact through the CDN (`セッションに紐付いたリンクで、共有用ではありません`).

### The label, on a real prototype with a real signature

```
Link valid until 2:13 AM · tied to your session, not a share link
```

- **59 minutes left** on the signature at page load, consistent with the 3600s TTL.
- **This is itself the proof that the locale fix took effect**: the old hardcoded `HH:mm` would have rendered `02:13`. `2:13 AM` is `Intl` applying the active locale's 12-hour convention.
- **`useId` is live**: the note's id came back as `_r_0_`, and **both** anchors carry `aria-describedby="_r_0_"`. Not a module constant, so two panes cannot collide.
- Both anchors carry the signature; `Download .html` has `download="Prototype.html"`.

### The link actually resolves, and the CSP travels with it

`GET` on the "Open in new tab" href:

```
200  text/html; charset=utf-8  52,977 bytes  <!doctype html>
content-security-policy: default-src 'none'; script-src 'unsafe-inline'; style-src 'unsafe-inline';
                         img-src data:; font-src data:; frame-ancestors 'self'; object-src 'none'; base-uri 'none';
```

So the signed URL works, and the `/prototypes/*` response-headers policy — the egress control on model-authored inline script — is present on the response.

### U7 AC 4, measured

On an **892 px** viewport: detail pane **827 px**, iframe **702 px**. Previously the pane was a flat `min-h-[500px]`, giving the preview roughly 430 px on any monitor. The prototype now occupies the majority of the tab, which is what that criterion asked for.

### Console

**Zero errors** with a signed prototype selected. Two CSP `script-src 'self'` errors do appear if you select a **legacy** prototype — those are the pre-migration `srcDoc` path that `PrototypeRenderer`'s own comment already documents as broken-but-non-breaking, and they predate this PR. Confirmed by reloading and touching only a signed prototype: clean.

### One thing I could NOT verify live, stated plainly

**The freeze on an in-place re-sign.** Observing it requires the project query to refetch *while the pane stays mounted*, and I could not force that within a session: a synthetic `focus`/`visibilitychange` did not trip React Query's focus manager, real tab switching after `staleTime` did not either, and a reload remounts the frame (which legitimately takes the new URL, so it proves nothing). The honest wait is the ~55-minute timer.

What stands behind it instead is test evidence I checked in both directions rather than assuming:

- `keeps a URL that loaded successfully even after its signature expires` — **fails** when the release is gated on the deadline instead of `deadOnArrival`
- `lets go of an expired URL as soon a live replacement arrives` — **fails** when the release is removed altogether

Each of the two behaviours fails under the other's implementation, which is the property that matters here given round 2's blocker was a regression from round 1's fix.

### Note on the deploy target

The shared environment is now running this branch rather than `development`. Worth reconverging it from the merge commit afterwards, per the usual practice here, so the deployed `APP_VERSION` stops naming a branch head.

PR #291, opened 2026-08-09
fix(i18n): read breadcrumb labels from the catalogue, not a hardcoded map

perrozzi:
## Summary

Breadcrumbs carried a hardcoded English label map that shadowed a complete, translated equivalent, and fell
back to printing raw path segments — so a project page's trail ended in the project's internal id instead of
its name.

Labels now come from the shipped `common:breadcrumbs` catalogue, `/projects/:id` shows the project's own name,
and `/feedback/:id` a translated stand-in. No English and no route id are left in the component.

Frontend only. No API, IaC or backend runtime change (`useProjectData.ts` gains a comment).

## What was actually wrong

Three things, and one of them turned out to be less true than it looked:

| Claim | Finding |
|---|---|
| `routeLabels` shadows the translated `breadcrumbs` block | **True.** The i18n gate reported all 11 `common:breadcrumbs.*` keys as unreferenced, and the component had no `useTranslation` at all. |
| The map still carries a `feedback` entry for a route that was removed | **Half true.** The catalogues had already dropped `breadcrumbs.feedback`, but `/feedback/:id` is still a live route, so the segment does need a label. `/feedback` itself only redirects to `/categories` now. |
| `routeLabels[segment] \|\| segment` prints raw ids | **True** — and the existing test suite codified it: it asserted `getByText('123')` for `/projects/123`. |

## What changed

**Labels read from the catalogue.** A new pure `components/Breadcrumbs/routeCrumbs.ts` holds the two segment
tables and the trail builder. Keys are written fully qualified inside a `labelKey` property because that is
the only shape `scripts/i18n-check.mjs` can see through data (its `extractDataHeldKeys`, the same way it sees
Layout's `NAV_ITEMS`) — a computed `` t(`breadcrumbs.${segment}`) `` would be invisible to that gate, which is
how eleven translated keys sat unreferenced for months without anything failing. The orphan count moves
**650 → 639**: the eleven are wired, and the two new keys register rather than adding new orphans.

**Record ids never reach the DOM.** `/projects/:id` shows the project's name; `/feedback/:id` a translated
stand-in, since a feedback record has no title field. The name is read from the `['project', id]` entry
ProjectDetail already fills, observed with `skipToken` — read-only, so the header issues no request of its own
and cannot race that page for one, while the subscription still fills the crumb in the moment the page's own
query resolves. The one field it needs goes through a small Zod schema, so nothing cached yet / an error state
/ a record saved before `name` existed all leave the generic label in place rather than throwing inside the
page header.

**The legacy `feedback` crumb names Categories, and links there.** `/feedback` was consolidated into
`/categories` (#198) and now only redirects, so `Home / Categories / Feedback item` is what the trail actually
does. It is also why the catalogue has no `breadcrumbs.feedback` key to point at.

**An unlabelled segment still prints itself**, because that means a route with no breadcrumb entry — a wiring
bug rather than user data. A new `route coverage` test makes that unreachable for real routes: it reads
`App.tsx`'s own route table (exported for this, deliberately not copied — a copy in the test would agree with
itself forever while the app grew routes past it) and fails if a layout route has no label, or a `:param`
route no stand-in. Adding a `/thing/:id` page can no longer quietly put an id back in the header.

**Two new keys × 8 locales:** `breadcrumbs.project`, `breadcrumbs.feedbackItem`.

**One adjacent a11y fix**, since I was in the markup: intermediate crumb labels are `hidden sm:inline` and
those links carry no icon, so below `sm` they had **no accessible name at all**. One `aria-label`.

## Verification

Gates: `eslint . --max-warnings 0` clean · `tsc -b --noEmit` clean · **2027 frontend tests / 133 files**.
Breadcrumb tests 25 → 36.

i18n gate: `missing` / `empty` / `untranslated` all stay at **0**, orphans 650 → 639, no `breadcrumbs.*` key
left on the unused list. The gate still exits 1 on the 12 pre-existing extra plural keys in ja/zh/ko
`categories.json` — identical to the baseline on `development`, untouched here.

**Mutation-tested — five mutations, each killed by a named test:** a static label removed from the table · a
new `/things/:id` route with no stand-in · an English label hardcoded again · the project cache key renamed ·
the record crumb falling back to the raw id.

The hardcoding mutation is the interesting one. Hardcoding a *single* label in English fails **exactly one
test — the `de` case — with all 35 English ones passing.** That is both the non-vacuity proof and the reason
this bug survived a 25-test suite: under `en` a hardcoded literal and its translation are the same string, so
no English test can fail on unwired i18n. The `de` block reads its expectations from the shipped catalogue, so
rewording a translation does not break it — only unwiring the component does.

Browser-checked against `npm run mock`:

- `/projects/proj_1` → `Home / Projects / Q1 Product Improvements` (id gone, name shown)
- `/feedback/fb_…` → `Home / Categories / Feedback item`, parent linking to `/categories`
- under `de` → `Startseite / Projekte / Q1 Product Improvements` — labels translated, the record **name**
  correctly left as data
- no new console errors

## Reviewer notes

- **`routes` is exported from `App.tsx` with one `eslint-disable-next-line react-refresh/only-export-components`.**
  `allowConstantExport` covers literal primitives only, not an array. The alternative was moving the route
  table to its own module, which would have dragged the lazy page imports, `page()` and `FeedbackRedirect`
  with it — a refactor of the app shell for a breadcrumb fix. The cost accepted is that editing `App.tsx`
  full-reloads instead of hot-updating, which editing the app shell does anyway. Reasoning is in the docblock.
- **`useRecordName` hardcodes `['project', id]`**, duplicating a key literal that appears in ~10 places across
  `useProjectData.ts` and `ProjectDetail.tsx`. I chose a comment at the read site over extracting a shared key
  helper: the refactor touches mutation-invalidation paths for what is a cosmetic regression risk. Happy to
  extract it if you'd rather — say so and it's a follow-up.
- Related: two of this item's premises were already fixed on `development` (no `breadcrumbs.feedback` key, no
  `feedback` locale namespace). If there's an open item covering those, it wants a re-check.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -11,9 +11,11 @@
     "dashboard": "Übersicht",
     "dataExplorer": "Daten-Explorer",
     "feedbackForms": "Feedback-Formulare",
+    "feedbackItem": "Feedback-Eintrag",
     "home": "Startseite",
     "prioritization": "Priorisierung",
     "problems": "Problemanalyse",
+    "project": "Projekt",
     "projects": "Projekte",
     "scrapers": "Web-Scraper",
     "settings": "Einstellungen"
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -11,9 +11,11 @@
     "dashboard": "Dashboard",
     "dataExplorer": "Data Explorer",
     "feedbackForms": "Feedback Forms",
+    "feedbackItem": "Feedback item",
     "home": "Home",
     "prioritization": "Prioritization",
     "problems": "Problem Analysis",
+    "project": "Project",
     "projects": "Projects",
     "scrapers": "Web Scrapers",
     "settings": "Settings"
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -11,9 +11,11 @@
     "dashboard": "Panel",
     "dataExplorer": "Explorador de datos",
     "feedbackForms": "Formularios de feedback",
+    "feedbackItem": "Elemento de feedback",
     "home": "Inicio",
     "prioritization": "Priorización",
     "problems": "Análisis de problemas",
+    "project": "Proyecto",
     "projects": "Proyectos",
     "scrapers": "Recolectores web",
     "settings": "Configuración"
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -11,9 +11,11 @@
     "dashboard": "Tableau de bord",
     "dataExplorer": "Explorateur de données",
     "feedbackForms": "Formulaires de feedback",
+    "feedbackItem": "Élément de feedback",
     "home": "Accueil",
     "prioritization": "Priorisation",
     "problems": "Analyse des problèmes",
+    "project": "Projet",
     "projects": "Projets",
     "scrapers": "Collecteurs web",
     "settings": "Paramètres"
--- a/voc-datalake/frontend/public/locales/ja/common.json
+++ b/voc-datalake/frontend/public/locales/ja/common.json
@@ -11,9 +11,11 @@
     "dashboard": "ダッシュボード",
     "dataExplorer": "データエクスプローラー",
     "feedbackForms": "フィードバックフォーム",
+    "feedbackItem": "フィードバック項目",
     "home": "ホーム",
     "prioritization": "優先順位付け",
     "problems": "問題分析",
+    "project": "プロジェクト",
     "projects": "プロジェクト",
     "scrapers": "Webスクレイパー",
     "settings": "設定"
--- a/voc-datalake/frontend/public/locales/ko/common.json
+++ b/voc-datalake/frontend/public/locales/ko/common.json
@@ -11,9 +11,11 @@
     "dashboard": "대시보드",
     "dataExplorer": "데이터 탐색기",
     "feedbackForms": "피드백 양식",
+    "feedbackItem": "피드백 항목",
     "home": "홈",
     "prioritization": "우선순위",
     "problems": "문제 분석",
+    "project": "프로젝트",
     "projects": "프로젝트",
     "scrapers": "웹 스크레이퍼",
     "settings": "설정"
--- a/voc-datalake/frontend/public/locales/pt/common.json
+++ b/voc-datalake/frontend/public/locales/pt/common.json
@@ -11,9 +11,11 @@
     "dashboard": "Painel",
     "dataExplorer": "Explorador de dados",
     "feedbackForms": "Formulários de feedback",
+    "feedbackItem": "Item de feedback",
     "home": "Início",
     "prioritization": "Priorização",
     "problems": "Análise de problemas",
+    "project": "Projeto",
     "projects": "Projetos",
     "scrapers": "Coletores web",
     "settings": "Configurações"
--- a/voc-datalake/frontend/public/locales/zh/common.json
+++ b/voc-datalake/frontend/public/locales/zh/common.json
@@ -11,9 +11,11 @@
     "dashboard": "仪表板",
     "dataExplorer": "数据浏览器",
     "feedbackForms": "反馈表单",
+    "feedbackItem": "反馈条目",
     "home": "首页",
     "prioritization": "优先级排序",
     "problems": "问题分析",
+    "project": "项目",
     "projects": "项目",
     "scrapers": "网页抓取器",
     "settings": "设置"
--- a/voc-datalake/frontend/src/App.tsx
+++ b/voc-datalake/frontend/src/App.tsx
@@ -1,6 +1,7 @@
 import { lazy, Suspense, useEffect, useState } from 'react'
 import type { ReactNode } from 'react'
 import { createBrowserRouter, Navigate, RouterProvider, useLocation } from 'react-router-dom'
+import type { RouteObject } from 'react-router-dom'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import Layout from './components/Layout'
 import ProtectedRoute from './components/ProtectedRoute'
@@ -55,7 +56,19 @@ function FeedbackRedirect() {
   return <Navigate to={`/categories${location.search}`} replace />
 }
 
-const router = createBrowserRouter([
+/**
+ * Exported so the breadcrumb tests can hold Breadcrumbs' route tables to the
+ * real router: every layout route needs a label, and every `:param` route needs
+ * a stand-in label, or the header falls back to printing a raw path segment.
+ *
+ * This is the router's own table, deliberately not a copy — a copy in the test
+ * would agree with itself forever while the app grew routes past it. It stays
+ * next to the router that consumes it rather than moving to its own module, so
+ * fast refresh is given up for this one file: editing the app shell full-reloads
+ * anyway.
+ */
+// eslint-disable-next-line react-refresh/only-export-components -- see above.
+export const routes: RouteObject[] = [
   {
     path: '/login',
     element: <Login />,
@@ -88,7 +101,9 @@ const router = createBrowserRouter([
       { path: 'settings', ...page(<AdminRoute><Settings /></AdminRoute>) },
     ],
   },
-])
+]
+
+const router = createBrowserRouter(routes)
 
 /**
  * App wrapper that loads runtime config before rendering.
--- a/voc-datalake/frontend/src/App.tsx
+++ b/voc-datalake/frontend/src/App.tsx
@@ -1,32 +1,11 @@
-import { lazy, Suspense, useEffect, useState } from 'react'
-import type { ReactNode } from 'react'
-import { createBrowserRouter, Navigate, RouterProvider, useLocation } from 'react-router-dom'
-import type { RouteObject } from 'react-router-dom'
+import { useEffect, useState } from 'react'
+import { createBrowserRouter, RouterProvider } from 'react-router-dom'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
-import Layout from './components/Layout'
-import ProtectedRoute from './components/ProtectedRoute'
-import AdminRoute from './components/AdminRoute'
 import PageLoader from './components/PageLoader'
-import RouteErrorBoundary from './components/RouteErrorBoundary'
-import Login from './pages/Login'
+import { routes } from './routes'
 import { loadRuntimeConfig, isConfigLoaded } from './runtimeConfig'
 import { useConfigStore } from './store/configStore'
 import { configureAmplify } from './lib/amplify-config'
-
-// Lazy load pages for better code splitting
-const Home = lazy(() => import('./pages/Home'))
-const Dashboard = lazy(() => import('./pages/Dashboard'))
-const FeedbackDetail = lazy(() => import('./pages/FeedbackDetail'))
-const Categories = lazy(() => import('./pages/Categories'))
-const ProblemAnalysis = lazy(() => import('./pages/ProblemAnalysis'))
-const Settings = lazy(() => import('./pages/Settings'))
-const Scrapers = lazy(() => import('./pages/Scrapers'))
-const Chat = lazy(() => import('./pages/Chat'))
-const Projects = lazy(() => import('./pages/Projects'))
-const ProjectDetail = lazy(() => import('./pages/ProjectDetail'))
-const Prioritization = lazy(() => import('./pages/Prioritization'))
-const FeedbackForms = lazy(() => import('./pages/FeedbackForms'))
-const DataExplorer = lazy(() => import('./pages/DataExplorer'))
 
 const queryClient = new QueryClient({
   defaultOptions: {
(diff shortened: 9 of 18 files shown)


perrozzi:

This is a frontend-only change that replaces the hardcoded English `routeLabels` map in `Breadcrumbs` with lookups into the shipped `common:breadcrumbs` catalogue, moves the segment tables + trail builder into a pure `routeCrumbs.ts`, resolves `/projects/:id` to the project's name (read-only `skipToken` observer on `['project', id]`) with a translated stand-in, adds `breadcrumbs.project`/`breadcrumbs.feedbackItem` to all 8 locales, and adds a route-coverage test driven by `App.tsx`'s exported route table. No backend/IaC surface, so the Lambda/IAM/ingestor rules don't apply. Overall this is a solid, well-tested change that fixes a real i18n bug and the `de` test block is exactly the right non-vacuity guard. I found no blockers, but several couplings deserve tightening before merge.

### Should fix

- **`voc-datalake/frontend/src/components/Breadcrumbs/Breadcrumbs.tsx` — the `['project', id]` key literal is only held together by a comment, and the test does not actually protect it.** `Breadcrumbs.test.tsx` seeds the cache with `client.setQueryData(['project', PROJECT_ID], …)`, i.e. it asserts the *component's* literal against itself. Renaming the key in `useProjectData.ts` leaves every test green and silently drops the crumb back to the generic label — so the claimed "project cache key renamed" mutation is only killed for one of the two sides. Either export a key factory (`projectQueryKey(id)`) from `useProjectData.ts` and use it in both places, or have `useRecordName` call the page's query options with `enabled: false` instead of `queryFn: skipToken`; both remove the duplication *and* the divergence risk.
- **The skipToken/real-queryFn sharing is untested.** The docblock's strongest claim ("the header issues no request of its own and cannot race that page for one, while the subscription still fills the crumb") is never exercised: all tests pre-populate the cache, so no test has two observers on `['project', id]` where one supplies `queryFn: skipToken` and the other a real fetch. That interaction (which observer creates the cache entry, and whether the page's fetch still runs) is version-sensitive TanStack behaviour. Add one test that renders `Breadcrumbs` + a component using `useProjectData` against a mocked fetch and asserts the request fires once and the crumb fills in.
- **`routeCrumbs.ts` `buildCrumbs` resolves record crumbs purely from the parent segment, before consulting `SEGMENT_CRUMBS`.** Any future static child of a record parent — `/projects/new`, `/projects/import` — is labelled "Project" even if a correct `SEGMENT_CRUMBS['new']` entry exists, and the new `route coverage` test won't catch it (adding the entry makes the coverage test pass while the label stays wrong). Check `SEGMENT_CRUMBS[segment]` first, or better, resolve the trail from `useMatches()`/route params rather than inferring "this segment is an id" from its neighbour.
- **Static-label coverage is thin for the thing this PR is about.** Only `dashboard`, `categories`, `dataExplorer`, `feedbackForms` are asserted; `chat`, `problems`, `prioritization`, `scrapers`, `settings`, `projects` are not. A wrong key in the table (e.g. `breadcrumbs.chat` vs a catalogue key named `aiChat`) would render the raw key string in the UI. A table-driven test over `SEGMENT_CRUMBS` asserting `t(labelKey) !== labelKey` for `en` + one other locale closes that gap cheaply.
- **`Breadcrumbs` now requires a `QueryClientProvider` ancestor.** Confirm every other renderer of it (Layout tests, any Storybook/test-utils wrapper) provides one; otherwise it throws at runtime in those contexts. Also, importing `../../App` in the test executes `createBrowserRouter(routes)` and the lazy page module graph at import time — a routing/config side effect in a breadcrumb unit test. A tiny `src/routes.tsx` exporting the table would avoid both the `react-refresh` disable and the side-effectful import; the stated cost (moving `page()`/`FeedbackRedirect`) seems smaller than the coupling being accepted.

### Nits

- `useRecordName` builds `['project', undefined]` on every non-project route, creating a junk cache entry. Guard the hook (early return / stable no-op key) when there is no id.
- `aria-label={crumb.label}` duplicates the visible text and overrides any future icon-derived name. An `sr-only` span alongside the `hidden sm:inline` label is the more conventional fix and can't drift from the visible string.
- The deleted `renders deeply nested routes` case (`/projects/123/personas`) removed the only multi-level assertion; consider re-adding it with a labelled trailing segment if such a route exists.
- Unknown segments still render raw URL text into the header (`/unknown-route`). React escapes it, so it's not XSS, but a generic label would avoid reflecting arbitrary/oversized path input in the chrome.
- `beforeAll` mutates the shared `i18next` singleton (`addResourceBundle` + `changeLanguage`); fine under vitest's per-file isolation, but fragile if this file ever gains concurrent tests.

---

perrozzi:
_(Answering review round 1 on `26f8721`. Addressed in 4 commits, pushed right after this comment.)_

Thanks — three of the five should-fixes were real gaps, one had a false premise but a good idea behind it, and one splits into a verified non-issue plus a valid fix. Details below, including a vacuous test of mine that only surfaced because I went looking for the one you asked for.

## Fixed

**1 — the `['project', id]` key was only held together by a comment. Correct, and your diagnosis of my mutation test was exact:** it seeded the key the *test* knows, so it protected the component's side and nothing else. Renaming it in `useProjectData.ts` left all 36 green.

New `src/api/projectQueryKeys.ts` exports `projectKey`, now used at all 10 sites (1 read + 7 invalidations in `useProjectData.ts`, 2 in `ProjectDetail.tsx`) and by the header. It is a new leaf module rather than an export from `useProjectData.ts` for one reason: that module imports `projectsApi`, and Breadcrumbs is in the always-loaded layout chunk — importing a constant from it would either pull the project page's data layer into that chunk or leave code-splitting resting on tree-shaking a module with imports. Worth noting the file already exports `projectJobsKey` and `productContextKey`, so a factory was the local convention and `['project', id]` was the odd one out.

Verified: renaming **only** the page's key now fails the new handover test (below).

**2 — the skipToken/real-queryFn sharing was untested. Correct.** New `handover from the page that owns the record` renders `Breadcrumbs` alongside a probe calling the real `useProjectData`, with only `projectsApi` mocked. It asserts the stand-in shows first, the crumb then fills with the fetched name, and `getProject` is called **exactly once**. So the two claims in the docblock — no request of the header's own, and no suppression of the page's — are now pinned, along with the key agreement from item 1.

**3 — record resolution ran before `SEGMENT_CRUMBS`. Correct, and a real trap for a future `/projects/new`.** `buildCrumbs` now consults the static table first and falls to the record branch only for an unlabelled segment. Pinned by `prefers a segment's own label over its parent's record stand-in` (`/projects/settings`), which fails if the order is flipped back, plus a deeper case `/projects/proj_1/settings`.

On `useMatches()`: agreed it is the principled way to know "this segment is a param", and I'd take it in a router-aware refactor, but it needs a **data** router. Breadcrumbs is rendered under `Layout` inside `createBrowserRouter` in production, so that part is fine — the cost is test-side, since these tests and `Layout.test` use the declarative `MemoryRouter`, where `useMatches` throws. It also wouldn't remove either label table, so it buys precision on one of the three resolution rules for a conversion of every renderer. Flagging as a deliberate deferral rather than a disagreement.

**4 — table-driven label coverage: added, but the premise is off.** `chat`, `problems`, `prioritization`, `scrapers`, `settings` and `projects` *are* each asserted — `Breadcrumbs.test.tsx` has a `displays correct label for …` case for 10 of the 11 static segments (`feedback` has its own dedicated case, since it resolves to Categories). So a wrong key in the table did already fail.

The idea is still an improvement, because those cases are hand-maintained and a *new* table entry gets no coverage from them, so I added the table-driven version anyway. **It found a vacuous assertion in my own first attempt, which is the part worth reporting:** i18next echoes a missing key **without** its namespace prefix, so `t('common:breadcrumbs.aiChat')` returns `breadcrumbs.aiChat` — never equal to the qualified key it was handed, which made `expect(resolved).not.toBe(labelKey)` unfailable. Caught by planting a key the catalogue lacks and watching only the *old* hand-written test fail. Now asserted with a sentinel `defaultValue`, and re-verified: the same mutation fails both.

**5b — the `App.tsx` import side effect. Correct, and the fix was cheaper than I'd estimated.** Route table moved to `src/routes.tsx`; `App.tsx` keeps the `QueryClient` and `createBrowserRouter(routes)`. Importing `routes` now constructs nothing. That also let me delete the `react-refresh` disable rather than relocate it — with one more step, since the rule then fired from the other side (a file exporting a non-component while *defining* `FeedbackRedirect`), so the redirect moved to `components/FeedbackRedirect.tsx`, following the existing flat-component precedent of `components/PrototypeRenderer.tsx`. **No `eslint-disable` anywhere in the diff now.**

**Nit 3 — the deleted deep-nesting case:** re-added as a pure `buildCrumbs` test, where it doubles as the precedence check above.

**Nit 5 — the i18next singleton:** added `expect(i18n.language).toBe('de')` in `beforeEach`, matching `DataSourceWizard/localization.test.tsx`. Without it, a move to a shared pool would make the negative "no English literal" assertions pass for the wrong reason.

## Verified, not changed

**5a — "Breadcrumbs now requires a `QueryClientProvider` ancestor."** Checked every renderer: `Layout` is the only production one and sits inside the provider (`App.tsx`); `Layout.test.tsx:69` mocks `../Breadcrumbs`; `RouteErrorBoundary.test.tsx` defines its own local `TestLayout` and never renders it. No Storybook in the repo. So there is no context that renders it without a client.

## Kept, with reasons

- **Nit 1, `['project', undefined]` on non-project routes.** Real, and the cost is one dataless cache entry — no fetch, no growth, and the key changes back on navigation. The hook can't be conditional, so avoiding it means hand-rolling a `useSyncExternalStore` read of the cache, which is more machinery than the entry it saves and loses the reactivity that makes the crumb fill in at all. Happy to take a suggestion that keeps the subscription.
- **Nit 2, `aria-label` vs an `sr-only` span.** Both read `crumb.label`, so neither can drift from the visible text, and identical text keeps WCAG 2.5.3 satisfied. Chose the attribute over a second text node; no strong view, easy to flip.
- **Nit 4, an unknown segment reflecting raw URL text.** Unreachable for a real URL: the router has no catch-all, so an unmatched path renders the root `RouteErrorBoundary` **in place of the whole layout** — verified on the deployed site, where `/unknown-route` has no `nav[aria-label="Breadcrumb"]` in the DOM at all. The fallback only fires for a route wired into the router without a label, i.e. a dev tree mid-change, which is exactly when printing the slug is the useful behaviour. It is now covered by a test carrying that reasoning.

## Gates

`eslint . --max-warnings 0` clean · `tsc -b --noEmit` clean · **2050 tests / 134 files** (was 2027/133) · i18n gate unchanged: orphans 639, `missing`/`empty`/`untranslated` 0.

Three mutations run against the new guards, each killed by a named test: the page-side key diverging from the shared factory, the resolution order flipped back, and a table entry pointing at a key the catalogue lacks.

Live verification in the PR body was against `26f8721`; I'll redeploy from the new head and post the re-check.

perrozzi:

Since the last review the route table moved out of `App.tsx` into a new `src/routes.tsx` (with `FeedbackRedirect` extracted to its own component file, dropping the `react-refresh` disable and the side-effectful `App` import in tests), the `['project', id]` literal was replaced by a shared `projectKey` factory in `src/api/projectQueryKeys.ts` and threaded through `useProjectData.ts`/`ProjectDetail.tsx`, `buildCrumbs` now resolves `SEGMENT_CRUMBS` before `RECORD_CRUMBS`, and two new test blocks landed (a real-fetch handover test and `routeCrumbs.test.ts` covering every label key in both tables). All five previous Should-fix items are addressed bar one verification.

### Resolved
- Key drift: `projectKey` is now the single spelling, read by both the header and the page (`api/projectQueryKeys.ts`).
- skipToken/real-queryFn sharing is exercised by "fills the crumb from the page's fetch, and adds no request of its own", asserting a single `getProject` call.
- Record-crumb precedence fixed in `routeCrumbs.ts`; pinned by the `/projects/settings` case in `routeCrumbs.test.ts`.
- Static-label coverage closed by the `label keys` table test over `SEGMENT_CRUMBS` + `RECORD_CRUMBS` with a sentinel `defaultValue`.
- `routes.tsx` removes both the eslint disable and the router-construction side effect in the breadcrumb test; the deleted multi-level case is back as `/projects/proj_1/settings`.
- `beforeEach` language assertion added to the `de` block.

### Still open
- Not visible in the diff: confirmation that every other renderer of `Breadcrumbs` (e.g. Layout tests, shared test wrappers) has a `QueryClientProvider` ancestor now that the component calls `useQuery`.
- Nit: `projectKey(undefined)` still creates a `['project', undefined]` observer on every non-project route.
- Nit: `aria-label={crumb.label}` still duplicates the visible text rather than an `sr-only` span.

### New findings
- `voc-datalake/frontend/src/components/Breadcrumbs/Breadcrumbs.test.tsx` — the `vi.mock('../../api/projectsApi')` factory defines only `getProject`/`getJobs`/`getProductContext`. Any other member `useProjectData` (or a hook it composes) starts calling becomes `undefined` at runtime, and the factory is not type-checked against the real module, so the failure surfaces as an unrelated TypeError. Prefer `importOriginal()` spread + overrides, or `vi.spyOn` on the real client.
- `voc-datalake/frontend/src/components/Breadcrumbs/Breadcrumbs.test.tsx` — the handover `beforeEach` re-sets return values but never clears call counts, so `toHaveBeenCalledTimes(1)` is order-dependent if any future case in this file touches the project query. Add `vi.clearAllMocks()`.
- `voc-datalake/frontend/src/components/Breadcrumbs/routeCrumbs.test.ts` — `expect(entries.length).toBe(13)` hardcodes the table size; adding a legitimate segment fails a test that says nothing about the change. `toBeGreaterThanOrEqual(11)` keeps the anti-vacuity guard without the churn.
- Minor: `routeCrumbs.test.ts` asserts English strings against the bare `i18next` singleton without initialising it in-file, so it depends on the global test setup preloading `common`. The sentinel assertion in `label keys` would catch a regression, but a short comment naming that dependency would help.

---

perrozzi:
_(Answering review round 2 on `329c928`. Addressed in one commit, pushed right after this comment.)_

All four new findings are test hygiene in code I added this round, and three are fixed. The fourth I fixed differently, because the suggested form conflicts with something this repo learned the hard way.

## Fixed

**The `vi.mock` factory lists only the members it knows about. Right, and the failure mode you describe is the nasty one** — a member added to the hook later resolves to `undefined` and surfaces as an unrelated TypeError. Replaced with `vi.spyOn(projectsApi, …)` on the real client: checked against the real signatures, and it can only ever stub what exists. Re-verified non-vacuous afterwards — renaming the page's key still fails only the handover case, so the spy switch didn't quietly turn the test green for a different reason.

**Call-count isolation.** `vi.spyOn` + `restoreAllMocks` in `afterEach` now gives each case a fresh spy rather than leaning on the suite-wide hook. Worth noting for the record that `src/test/setup.ts` already calls `vi.clearAllMocks()` after every test, so the counts *were* being reset — but on a hook this file can't see, which is a poor thing to depend on. The local teardown makes it visible.

**The i18next dependency in `routeCrumbs.test.ts`.** Comment added naming it: the instance is the one `src/test/setup.ts` initialises from the shipped `en` catalogues, and this file deliberately configures nothing, so the English strings asserted are the ones the app ships.

## Fixed differently

**`expect(entries.length).toBe(13)` — agreed it's the wrong assertion, but `toBeGreaterThanOrEqual(11)` trades one weakness for another.** A `>=` bound is how a test goes quietly vacuous in this repo: an earlier PR here shipped a `toHaveLength(8)` that pinned a table's *size* and not its *membership*, and the lesson taken was to assert membership instead of a count in either direction.

So it now asserts membership by label key: `common:breadcrumbs.dashboard` is unique to `SEGMENT_CRUMBS` and `common:breadcrumbs.project` unique to `RECORD_CRUMBS`, so both tables are provably reached, an empty `entries` still fails, and adding a legitimate segment changes nothing. Renamed to `reaches both tables`, which is what it now checks.

## Still open — answered, not re-opened

The three items under "Still open" were each answered in the round-1 comment, which the reviewer doesn't read, so briefly:

- **`QueryClientProvider` for other renderers:** verified, and it isn't in the diff because nothing needed changing. `Layout` is the only production renderer and sits inside the provider (`App.tsx`, `QueryClientProvider` wrapping `RouterProvider`); `Layout.test.tsx:69` mocks `../Breadcrumbs`; `RouteErrorBoundary.test.tsx` uses its own local `TestLayout` and never renders it. No Storybook in the repo.
- **`projectKey(undefined)` observer:** kept. One dataless entry, no fetch; avoiding it means hand-rolling a cache read and losing the subscription that makes the crumb fill in at all.
- **`aria-label` vs `sr-only`:** kept. Both read `crumb.label`, so neither can drift, and identical text keeps WCAG 2.5.3 satisfied. No strong view — say the word and I'll flip it.

## Gates

`eslint . --max-warnings 0` clean · `tsc -b --noEmit` clean · **2050 tests / 134 files** · i18n gate unchanged (orphans 639, `missing`/`empty`/`untranslated` 0).

Deployed from `329c928` and re-verified live on real data, in both locales:

| Route | `en` | `de` |
|---|---|---|
| `/projects/:id` | `Home / Projects / <the project name>` | `Startseite / Projekte / <the project name>` |
| `/feedback/:id` | `Home / Categories / Feedback item` | `Startseite / Kategorien / Feedback-Eintrag` |
| `/data-explorer` | `Home / Data Explorer` | `Startseite / Daten-Explorer` |

The deployed chunk is byte-identical to the local build and carries zero of the old hardcoded labels. Two things the mock couldn't show: on a **cold** project route the stand-in renders first and the name replaces it when the page's fetch lands (the deployed API is slower than the mock, so the handover is actually visible there); on a **warm** cache the name is present in the first frame, no flash. Console clean. This round is test-only, so that deployment still matches the head.

perrozzi:

The incremental changes touch only the two test files: the `vi.mock('../../api/projectsApi')` factory is replaced with `vi.spyOn` on the real client plus `afterEach(vi.restoreAllMocks)`, the repeated project-name literal is hoisted to `PROJECT_NAME`, the hardcoded table-size assertion becomes a membership check on two table-unique keys, and the i18next-singleton dependency is documented. All four of my previous new findings are addressed; no production code changed, so nothing I approved earlier is affected.

### Resolved
- `Breadcrumbs.test.tsx` — spies on `projectsApi` instead of a partial `vi.mock` factory, so a newly-called member no longer resolves to `undefined` and is type-checked against the real signature.
- `Breadcrumbs.test.tsx` — `afterEach(vi.restoreAllMocks)` plus re-spying in `beforeEach` gives each case a fresh call history, removing the order dependence of `toHaveBeenCalledTimes(1)`.
- `routeCrumbs.test.ts` — `expect(entries.length).toBe(13)` replaced by `toContain('common:breadcrumbs.dashboard')` / `toContain('common:breadcrumbs.project')`, keeping the anti-vacuity guard without failing on legitimate additions.
- `routeCrumbs.test.ts` — comment now names the `src/test/setup.ts` `en`-catalogue dependency behind the bare `i18next` singleton.

### Still open
- Not visible in any diff: confirmation that every other renderer of `Breadcrumbs` (Layout tests, shared wrappers) has a `QueryClientProvider` ancestor now that the component calls `useQuery`.
- Nit: `projectKey(undefined)` still creates a `['project', undefined]` observer on non-project routes.
- Nit: `aria-label={crumb.label}` duplicates the visible text rather than using an `sr-only` span.

No new findings — the delta is clean.

---

PR #292, opened 2026-08-09
fix(ui): count Problem Analysis over the window, not its first page

perrozzi:
## Summary

Problem Analysis asked `/feedback` for `limit: 500`. That endpoint bounds `limit` with
`validate_limit(default=50, max_val=100)`, and `validate_int` **clamps rather than rejecting**, so the
request came back with 100 rows and only the echoed `limit` to give it away. Every stat card, and every
level of the category → subcategory → problem tree, was derived from that first page and presented as
complete.

Measured against a deployed environment, `days=90`:

| | |
|---|---|
| before — one request asking `limit=500` | **100 rows** (response echoed `limit=100`) |
| after — a paged walk at `limit=100` | **1,227 rows** over 13 requests (server `total=1,227`) |

The page had been describing **8.1%** of the window.

## Approach

A larger `limit` cannot fix this, because the cap is server-side. So page the window instead,
advancing automatically rather than behind a "Load more" button: unlike a browsable list, the totals
here are wrong until the window has been read. Rows render progressively, so the page stays usable
while it fills. On the corpus above the walk takes 13 requests and finishes inside the page budget, so
no caveat is shown.

Three stops bound the walk, independently of each other:

- `nextPageOffset` refuses a zero-length page, so a server reporting `total > loaded` while returning
  nothing cannot spin the client.
- `MAX_AUTO_PAGES` ends it outright.
- A failed page stops it — and counts as **partial**, because reporting short counts as complete would
  recreate the defect this fixes.

## Changes

- **`src/api/feedbackPagination.ts`** (new) — holds the page size as one exported constant equal to the
  endpoint's `max_val`, so callers spend `FEEDBACK_PAGE_LIMIT` instead of a literal and "I need more
  rows" becomes a pagination question rather than a bigger-`limit` question. `nextPageOffset` already
  existed and was correct, but private to `Categories/useFeedbackListData.ts`; it is **promoted here
  rather than copied**, and that file's private duplicate (plus its own `PAGE_LIMIT = 100`) is deleted.
  This is why the diff touches Categories.
- **`useProblemFeedback`** (new) — pages the window, bounded by `MAX_AUTO_PAGES`, and reports
  `isPartial`.
- **`WindowCoverageNotice`** (new) — renders nothing when the window was read in full; otherwise
  reports progress, or that the counts undercount. Its own file because `scripts/i18n-check.mjs`
  derives one namespace per file from the first `useTranslation` match, and `ProblemAnalysis.tsx` opens
  with `common`.
- 2 keys added to the `problemAnalysis` catalogue across all 8 locales. Interpolation only, no plural
  forms, so there is no `_one`/`_other` parity to keep in step.

## Testing

22 new tests (8 pagination, 9 hook, 5 component). Two of them are regression guards for bugs the tests
found in this change before review:

- The auto-advance originally fired **exactly once** — 200 rows of 250 — because after each page
  settles the effect's dependency array is identical to what it was after the previous page, so React
  skips the re-run. The page count is the ratchet and had to be in the deps.
- A mid-walk request failure reported short counts as complete. `isError` now feeds both the advance
  guard and `isPartial`.

Gates: `eslint` 0, `tsc` 0, **137 files / 2072 vitest tests** green.

`i18n:check` and `typecheck:tests` exit non-zero on `development` already. Both were verified as
pre-existing against a `git worktree` at `HEAD`: identical failure sets, zero missing keys on either
side, and no errors in any file this PR touches.

## Not in scope

Server-side aggregation is the real cure, so that one request could answer "how many problems per
category" without shipping every row to the browser. That is a larger change; this makes the current
design honest. The Problem Analysis stat labels are also still hardcoded English while the
`problemAnalysis` catalogue sits unused — pre-existing, and left alone.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/de/problemAnalysis.json
@@ -23,8 +23,9 @@
   "stats": {
     "categories": "Kategorien",
     "feedback": "Rückmeldungen",
-    "loadingWindow": "Feedback wird geladen… bisher {{loaded}} von {{total}}.",
-    "partialWindow": "Die Zahlen umfassen {{loaded}} von {{total}} Einträgen in diesem Zeitraum.",
+    "loadFailed": "Feedback für diesen Zeitraum konnte nicht geladen werden, daher werden keine Zahlen angezeigt. Erneut versuchen oder den Zeitraum eingrenzen.",
+    "loadingWindow": "Feedback wird geladen… bisher {{loaded, number}} von {{total, number}}.",
+    "partialWindow": "Die Zahlen umfassen {{loaded, number}} von {{total, number}} Einträgen in diesem Zeitraum.",
     "problems": "Probleme",
     "subcategories": "Unterkategorien",
     "urgent": "Dringend"
--- a/voc-datalake/frontend/public/locales/de/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/de/problemAnalysis.json
@@ -27,6 +27,7 @@
     "loadingWindow": "Feedback wird geladen… bisher {{loaded, number}} von {{total, number}}.",
     "partialWindow": "Die Zahlen umfassen {{loaded, number}} von {{total, number}} Einträgen in diesem Zeitraum.",
     "problems": "Probleme",
+    "retry": "Erneut versuchen",
     "subcategories": "Unterkategorien",
     "urgent": "Dringend"
   },
--- a/voc-datalake/frontend/public/locales/de/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/de/problemAnalysis.json
@@ -23,6 +23,8 @@
   "stats": {
     "categories": "Kategorien",
     "feedback": "Rückmeldungen",
+    "loadingWindow": "Feedback wird geladen… bisher {{loaded}} von {{total}}.",
+    "partialWindow": "Die Zahlen umfassen {{loaded}} von {{total}} Einträgen in diesem Zeitraum.",
     "problems": "Probleme",
     "subcategories": "Unterkategorien",
     "urgent": "Dringend"
--- a/voc-datalake/frontend/public/locales/en/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/en/problemAnalysis.json
@@ -23,6 +23,8 @@
   "stats": {
     "categories": "Categories",
     "feedback": "Feedback",
+    "loadingWindow": "Loading feedback… {{loaded}} of {{total}} so far.",
+    "partialWindow": "Counts cover {{loaded}} of {{total}} items in this window.",
     "problems": "Problems",
     "subcategories": "Subcategories",
     "urgent": "Urgent"
--- a/voc-datalake/frontend/public/locales/en/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/en/problemAnalysis.json
@@ -23,8 +23,9 @@
   "stats": {
     "categories": "Categories",
     "feedback": "Feedback",
-    "loadingWindow": "Loading feedback… {{loaded}} of {{total}} so far.",
-    "partialWindow": "Counts cover {{loaded}} of {{total}} items in this window.",
+    "loadFailed": "Could not load feedback for this window, so no counts are shown. Retry, or narrow the time range.",
+    "loadingWindow": "Loading feedback… {{loaded, number}} of {{total, number}} so far.",
+    "partialWindow": "Counts cover {{loaded, number}} of {{total, number}} items in this window.",
     "problems": "Problems",
     "subcategories": "Subcategories",
     "urgent": "Urgent"
--- a/voc-datalake/frontend/public/locales/en/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/en/problemAnalysis.json
@@ -27,6 +27,7 @@
     "loadingWindow": "Loading feedback… {{loaded, number}} of {{total, number}} so far.",
     "partialWindow": "Counts cover {{loaded, number}} of {{total, number}} items in this window.",
     "problems": "Problems",
+    "retry": "Retry",
     "subcategories": "Subcategories",
     "urgent": "Urgent"
   },
--- a/voc-datalake/frontend/public/locales/es/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/es/problemAnalysis.json
@@ -23,6 +23,8 @@
   "stats": {
     "categories": "Categorías",
     "feedback": "Comentarios",
+    "loadingWindow": "Cargando comentarios… {{loaded}} de {{total}} hasta ahora.",
+    "partialWindow": "Los recuentos abarcan {{loaded}} de {{total}} elementos de este periodo.",
     "problems": "Problemas",
     "subcategories": "Subcategorías",
     "urgent": "Urgente"
--- a/voc-datalake/frontend/public/locales/es/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/es/problemAnalysis.json
@@ -27,6 +27,7 @@
     "loadingWindow": "Cargando comentarios… {{loaded, number}} de {{total, number}} hasta ahora.",
     "partialWindow": "Los recuentos abarcan {{loaded, number}} de {{total, number}} elementos de este periodo.",
     "problems": "Problemas",
+    "retry": "Reintentar",
     "subcategories": "Subcategorías",
     "urgent": "Urgente"
   },
--- a/voc-datalake/frontend/public/locales/es/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/es/problemAnalysis.json
@@ -23,8 +23,9 @@
   "stats": {
     "categories": "Categorías",
     "feedback": "Comentarios",
-    "loadingWindow": "Cargando comentarios… {{loaded}} de {{total}} hasta ahora.",
-    "partialWindow": "Los recuentos abarcan {{loaded}} de {{total}} elementos de este periodo.",
+    "loadFailed": "No se pudieron cargar los comentarios de este periodo, por lo que no se muestran recuentos. Vuelve a intentarlo o reduce el intervalo de tiempo.",
+    "loadingWindow": "Cargando comentarios… {{loaded, number}} de {{total, number}} hasta ahora.",
+    "partialWindow": "Los recuentos abarcan {{loaded, number}} de {{total, number}} elementos de este periodo.",
     "problems": "Problemas",
     "subcategories": "Subcategorías",
     "urgent": "Urgente"
--- a/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
@@ -27,6 +27,7 @@
     "loadingWindow": "Chargement des retours… {{loaded, number}} sur {{total, number}} pour l’instant.",
     "partialWindow": "Les décomptes couvrent {{loaded, number}} éléments sur {{total, number}} dans cette période.",
     "problems": "Problèmes",
+    "retry": "Réessayer",
     "subcategories": "Sous-catégories",
     "urgent": "Prioritaire"
   },
--- a/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
@@ -23,6 +23,8 @@
   "stats": {
     "categories": "Catégories",
     "feedback": "Retours",
+    "loadingWindow": "Chargement des retours… {{loaded}} sur {{total}} pour l’instant.",
+    "partialWindow": "Les décomptes couvrent {{loaded}} éléments sur {{total}} dans cette période.",
     "problems": "Problèmes",
     "subcategories": "Sous-catégories",
     "urgent": "Prioritaire"
--- a/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
+++ b/voc-datalake/frontend/public/locales/fr/problemAnalysis.json
@@ -23,8 +23,9 @@
   "stats": {
     "categories": "Catégories",
     "feedback": "Retours",
-    "loadingWindow": "Chargement des retours… {{loaded}} sur {{total}} pour l’instant.",
-    "partialWindow": "Les décomptes couvrent {{loaded}} éléments sur {{total}} dans cette période.",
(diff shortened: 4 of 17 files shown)


perrozzi:

This PR replaces Problem Analysis's single `limit: 500` `/feedback` request with a paged walk (`useInfiniteQuery` + auto-advance) bounded by `MAX_AUTO_PAGES`, extracts `nextPageOffset` / `FEEDBACK_PAGE_LIMIT` into a shared `src/api/feedbackPagination.ts` (deduplicating the private copy in `Categories/useFeedbackListData.ts`), and adds a `WindowCoverageNotice` with two new keys across all 8 locales. The diagnosis is correct and well documented, the i18n keys are complete for every locale, and the regression tests (page-size cap, multi-page advance, failure→partial) genuinely fail if the fix is reverted. No blockers found; the remaining concerns are around request volume/caching and unsurfaced errors.

### Should fix

- `src/pages/ProblemAnalysis/useProblemFeedback.ts` — the query sets no `staleTime`/`refetchOnWindowFocus`/`gcTime`. TanStack refetches **all** pages of an infinite query on refetch, so a window focus or invalidation on a 1,200-row corpus re-issues 13 sequential `/feedback` calls (up to 20 with `MAX_AUTO_PAGES`), each an API Lambda + backing query. Add `staleTime` and `refetchOnWindowFocus: false` (and consider `maxPages`) so the walk is paid for once per window.
- `useProblemFeedback.ts` — a *total* failure is still reported as a complete empty window. If the **first** page rejects, `hasNextPage` is false, so `stoppedEarly` is false, `isPartial` is false, `loadedCount`/`totalCount` are 0, and `ProblemAnalysis.tsx` renders zeroed stat cards with no notice — exactly the "short counts presented as complete" class of defect this PR fixes. Expose `isError` from the hook and render an error state (or force `isPartial`) when there are no pages and the query errored.
- `feedbackPagination.ts` / `nextPageOffset` — the cursor is derived from the server-echoed `lastPage.offset`. If a response ever omits `offset`, `loaded` collapses to `pageSize` and the function returns the same offset forever; in Problem Analysis that is only capped by `MAX_AUTO_PAGES` (20 duplicate pages of the same rows), and in `Categories/useFeedbackListData.ts` it is user-driven. Deriving the next offset from accumulated page lengths (or `allPageParams`) removes the dependency on the echo.
- `FEEDBACK_PAGE_LIMIT = 100` is a hand-copied mirror of `validate_limit(default=50, max_val=100)` with no cross-language check; the new test pins the literal, not the server. At minimum cite the exact backend file/function in the constant's docblock so a `max_val` change is discoverable. Better follow-up: make `validate_int` reject rather than clamp for `limit`, so a client asking for too much learns about it instead of silently getting 8% of the window.

### Nits

- `WindowCoverageNotice.tsx`: `role="status"` is implicitly `aria-live="polite"`, so a 13-page walk announces 13 progress updates to screen readers. Consider announcing only the terminal partial state, or debouncing the progress text.
- Interpolated counts are raw numbers (`{{loaded}} of {{total}}`); for `de`/`ja`/`zh` thousands separators differ. `{{loaded, number}}` would localize them.
- `useProblemFeedback.test.tsx`: the failure test uses a real `setTimeout(…, 50)` to prove no further requests fire — that's a timing-dependent assertion; fake timers or asserting on a settled `isPartial` + call count would be sturdier.
- The query key `['feedback-problems', dateParams]` is reused from the old `useQuery` but now holds an infinite-query cache shape; worth confirming nothing else (e.g. PDF export paths) reads that key with `useQuery`.
- `FeedbackPage`/`FeedbackPageCursor` are structural types only — `count`/`total`/`offset` arrive unvalidated from the boundary. If the API client has a Zod schema for `/feedback`, deriving these from it would fit the repo's runtime-validation pattern.

---

perrozzi:

Since my last review the author addressed every Should-fix and most nits: `staleTime`/`refetchOnWindowFocus` pinned (with a caching regression test), `isError` exposed and rendered via a new `stats.loadFailed` key in all 8 locales, `nextPageOffset` re-derived from accumulated page lengths (with `offset` removed from the cursor type and two regression tests), a Python↔TS lockstep test for `FEEDBACK_PAGE_LIMIT`, progress no longer a live region, and `{{loaded, number}}` for localized separators.

### Resolved
- `useProblemFeedback.ts` now sets `staleTime: WINDOW_STALE_MS` + `refetchOnWindowFocus: false`, with a shared-cache remount test proving no re-walk.
- Total failure no longer renders as a complete empty window: `isError` is exposed and `ProblemAnalysis.tsx` renders the failure notice instead of zeroed aggregates.
- `nextPageOffset` sums `allPages` lengths instead of trusting the echoed `offset`; the field is gone from `FeedbackPageCursor`, guarded by the "ignores a wrong offset echo" test.
- `FEEDBACK_PAGE_LIMIT` now cites `metrics_handler.py` → `list_feedback` and is enforced by `lambda/api/test/test_feedback_page_limit_lockstep.py`.
- Nits: live-region spam (progress no longer `role="status"`), unlocalized numbers, and the `setTimeout(…, 50)` timing assertion are all fixed.

### Still open
- Nit only: `FeedbackPage`/`FeedbackPageCursor` remain structural types with no Zod/runtime validation at the boundary; query key `['feedback-problems', dateParams]` now holds an infinite-query shape.

### New findings

**Should fix**
- `src/pages/ProblemAnalysis/ProblemAnalysis.tsx` — the new total-failure branch returns early and replaces the *entire* page, including the time-range selector and any retry affordance, while `stats.loadFailed` tells the user to "Retry, or narrow the time range." The UI removes both options it recommends. Render the notice above the (still interactive) header/filters, or add a retry button wired to `refetch`.

**Nits**
- `lambda/api/test/test_feedback_page_limit_lockstep.py` — `source.index('@app.get', start + 1)` raises `ValueError` (test error, not a readable failure) if `/feedback` is the last `@app.get`, and misses handlers registered with `@app.post`/`@app.route` after it. `find(...)` with a `len(source)` fallback is sturdier. Also worth noting the test reads a frontend file from a lambda test dir, so it fails hard in any packaging context where `frontend/` is absent.
- `src/api/feedbackPagination.ts` — now that `loaded` is summed from `allPages`, any caller that adds TanStack `maxPages` gets page eviction → a rewound offset → an unbounded re-walk of already-fetched rows. `useProblemFeedback` documents why it doesn't use `maxPages`; a one-line warning in `nextPageOffset`'s docblock would protect the shared helper.
- `useProblemFeedback.ts` — a failed *background* refetch (post-invalidation) sets `isError` while full data is still cached, so `isPartial` can flag an already-complete window as short. Low impact given the 5-minute `staleTime`, but consider gating `stoppedEarly` on the walk not having completed.

---

perrozzi:

Since the last review the author wired the failure notice's retry (`feedback.retry` → `onRetry` → a `stats.retry` button, key added in all 8 locales), added the `maxPages` warning to `nextPageOffset`'s docblock, hardened the lockstep test's route-boundary scan (`find` + any-verb regex + `pytest.skip` when `frontend/` is absent), and added two hook tests (retry recovery, and a guard that a failed refetch can't mark a completed window partial). All three of my previous items are addressed.

### Resolved
- `ProblemAnalysis.tsx` total-failure branch now offers the retry its message instructs (`onRetry={feedback.retry}`), covered by two `WindowCoverageNotice` tests.
- `test_feedback_page_limit_lockstep.py` no longer raises `ValueError` at end-of-file, catches non-`@app.get` routes, and skips cleanly outside a full checkout.
- `nextPageOffset` docblock now warns shared callers off `maxPages`.
- Failed-background-refetch → false `isPartial` is pinned by "does not call a completed window partial when a later refetch fails" (`stoppedEarly` gated on `hasNextPage`).

### Still open
- Nit only: `FeedbackPage`/`FeedbackPageCursor` remain structural types with no runtime/Zod validation at the boundary.

### New findings

**Nits**
- `src/pages/ProblemAnalysis/WindowCoverageNotice.tsx` — the retry `<button>` has no `type="button"`; if this notice is ever rendered inside a form it will submit. Add it.
- `src/pages/ProblemAnalysis/WindowCoverageNotice.tsx` — the failure branch hardcodes `text-center` on its wrapper instead of using the `className` prop pattern the other branches use; it will fight any non-centred parent (the partial/progress notice under the stat row is left-aligned).
- `src/pages/ProblemAnalysis/useProblemFeedback.ts` — `retry` is a fresh closure each render; harmless today, but `useCallback` keeps it safe for a memoized consumer.
- `src/pages/ProblemAnalysis/useProblemFeedback.test.tsx` — `result.current.retry()` is called outside `act(...)`; the following `waitFor` makes the assertion sound but expect act warnings in the log. Wrap the call.
- `ProblemAnalysis.tsx` — the failure branch still replaces the whole page, so the "or narrow the time range" half of `stats.loadFailed` remains unreachable. Retry covers the common case, so this is optional now.

---

perrozzi:
Round 2 addressed in `73b94ed`. Thanks — the should-fix was a real self-contradiction, and one of the nits sent me to check something I would otherwise have assumed.

## Should fix — retry now exists; one half of the finding was a false alarm

You're right that telling the user to "Retry, or narrow the time range" while offering neither is worse than saying nothing. Retry genuinely did not exist and now does: the hook exposes `retry` (a `refetch`), and the notice renders the button whenever a handler is supplied — omitted when it isn't, so the instruction and the control can't drift apart.

The other half doesn't reproduce, and it's worth recording why. **`TimeRangeSelector` is rendered by `Layout`, not by this page**, so the early return never removed it — "narrow the time range" stayed actionable throughout. Verified live with the failure induced:

```
alert:  "Could not load feedback for this window… Retry, or narrow the time range."  + [Retry]
buttons present: Imported date · 24h · 48h · 7d · 30d · 90d · Custom · Retry
stat grid rendered: false
```

So the page's own filter bar does go away (correctly — there is nothing to filter), the app-level time controls do not, and the recommended actions are both available. Clicking Retry with the fault lifted recovers to the full window: `Categories 16 · Subcategories 248 · Problems 505 · Feedback 615 · Urgent 304`, alert cleared.

## Nits

**`maxPages` incompatibility** — good catch on the shared helper. `nextPageOffset`'s docblock now carries the warning explicitly: summing `allPages` means page eviction shrinks the sum, rewinds the offset, and re-walks rows already fetched. The reasoning lived only in `useProblemFeedback`; it belonged on the helper.

**Lockstep test parsing** — fixed both parts. It now finds the next route decorator of any verb by regex and falls back to end-of-file, so `/feedback` being the last route is a readable failure rather than a `ValueError`. And it `pytest.skip`s when `frontend/` is absent, since a backend test reaching across trees has nothing to compare in a packaging context.

**Failed background refetch flagging a complete window as partial** — this one does not reproduce, and there's now a test that says so. `stoppedEarly` is `hasNextPage && (budgetSpent || isError)`, and a completed walk leaves `hasNextPage` false, so `isError` alone can't make `isPartial` true. Pinned as *"does not call a completed window partial when a later refetch fails"* so the gate can't be dropped silently. If you meant the narrower case where a background refetch fails **mid-way** and truncates `data`, then `hasNextPage` does go true — but there the window really is short again, so flagging it is correct.

## Still open, deferred with reasons

**No runtime validation on the cursor types.** There is no Zod schema for the `/feedback` *envelope* today — only `FeedbackItemSchema` for the rows, applied via `normalizeFeedbackItems`. Adding envelope validation means introducing it in `client.ts` for every `/feedback` caller, which is a wider change than this fix warrants. Worth doing; not here.

**Query key holding an infinite-query shape.** Checked: `['feedback-problems', …]` is read in exactly one place — this hook. No `useQuery` consumer to break.

## Verification

Deployed and exercised on a live environment, not just unit-tested:

- 13 `/feedback` calls at offsets 0…1200, walk terminates cleanly
- counts settle at **615** problem-bearing items — exactly the number of rows carrying `problem_summary` in the table
- notice clears on a complete read; navigating away and back does not re-walk
- failure path and Retry as quoted above

Gates: eslint 0, tsc 0, ruff clean, **137 files / 2085 vitest**, 2 pytest, 0 missing i18n keys across 8 locales.

perrozzi:

Since the last review the author applied all four nits from the previous round: `type="button"` on the retry control, dropped the hardcoded `text-center` (with a comment explaining the alignment now belongs to the caller), wrapped `retry` in `useCallback`, and moved both `result.current.retry()` calls inside `await act(...)`. All previous findings are now addressed except the standing Zod nit.

### Resolved
- `WindowCoverageNotice.tsx` — retry button now has `type="button"`, so it can't submit an enclosing form.
- `WindowCoverageNotice.tsx` — failure branch no longer imposes `text-center`; alignment stays with the caller for both render sites.
- `useProblemFeedback.ts` — `retry` is a stable `useCallback` over `refetch`.
- `useProblemFeedback.test.tsx` — both retry invocations wrapped in `act`, removing act warnings.

### Still open
- Nit only: `FeedbackPage`/`FeedbackPageCursor` remain structural types with no runtime/Zod validation at the API boundary.

The incremental changes are clean — no new findings.

---

perrozzi:
Round 3 addressed in `b1f25aa`. Nit-only, so this is the last round of code churn on this PR — but four of the five were a few characters each and one was a latent bug, so they were worth taking rather than deferring.

- **`type="button"`** — real catch. The default is `submit`, so nested in a form the retry would post it. Fixed.
- **Hardcoded `text-center`** — removed. The component now imposes no alignment, so it can be centred in the empty page and left-aligned under the stat row without fighting either; the empty-page caller already centres it.
- **`retry` closure** — now `useCallback`'d against `refetch`.
- **`act(...)`** — both retry tests wrap the call. The run is now free of act warnings rather than merely correct.

**Re-raised and still declined, evidence unchanged:** "the failure branch replaces the whole page, so the 'or narrow the time range' half is unreachable." `TimeRangeSelector` and the date-basis control are rendered by `Layout`, not by any page, so both stay interactive throughout the failure state. Verified live again on `b1f25aa` — during an induced failure the page renders `[Retry]` alongside `Imported date · 24h · 48h · 7d · 30d · 90d · Custom`, with no stat grid. Both recommended actions are available.

**Still open by choice:** the cursor types carry no runtime validation. There is no Zod schema for the `/feedback` *envelope* today — only `FeedbackItemSchema` for the rows — so adding one means introducing envelope validation in `client.ts` for every caller. Worth doing; a different change.

## Verification of the shared helper's *other* consumer

`nextPageOffset` is used by `Categories/useFeedbackListData` too, where paging is user-driven rather than automatic — the case where a mis-derived cursor would be unbounded. Since this PR changed how that offset is computed, I exercised it live rather than relying on its unit tests:

```
Categories → "Load more" ×3
  offsets requested: 0, 100, 200, 300
  rows rendered:     400
  unique rows:       400
  duplicates:        0
```

Clean advance, no repetition, button still offered. Together with Problem Analysis's 13-page walk (offsets 0…1200, settling at 615 problem-bearing items), both consumers of the shared cursor are confirmed against a real deployment.

Gates: eslint 0, tsc 0, ruff clean, **137 files / 2085 vitest**, 2 pytest, 0 missing i18n keys across 8 locales.

PR #294, opened 2026-08-09
feat(ui): make the prototype build step 5 of the project Overview

perrozzi:
## Summary

The prototype build was a lone orange button pinned to the right of the project tab row. It now sits
in the Overview card grid as **step 5 of 6**, between PRD/PR-FAQ and Remix.

Two problems with the old placement:

1. **It read as a tab-level control** for what is really the last artifact in the sequence.
2. **It could not report state.** The label said "Build Prototype" whether or not a prototype
   already existed, so nothing on the project told you one had been built — and clicking again just
   built another.

**This deletes more than it adds.** `BuildPrototypeButton` was hand-rolling chrome the card grid
already had (heading, description, state line, disabled state with an explanation), so the component
and the `ProjectTabs` `rightSlot` prop it existed for are both gone.

## Why step 5, before Remix

Dependency order, not taste. A prototype needs **one** of PRD/PR-FAQ; a remix needs **two**
documents. And a prototype **produces** a new artifact where a remix **revises** existing ones.

This is also why prototype is recommendable by `nextStep` and remix still is not: a prototype is
detectable (`document_type === 'prototype'`), so the recommendation clears itself once one exists,
whereas a remixed document is stored as an ordinary PRD/PR-FAQ and cannot be told apart from a
generated one.

## What the card gains over the button

- **Reports its own output** — "Prototypes built: N" / "None yet".
- **Explains the disabled state visibly.** The button carried "Create a PRD or a PR-FAQ first" as a
  hover `title`, which never reached keyboard or touch users. It is now the card's
  `disabledMessage`, and the two hover-only keys (`documents.prototype.buttonTitle` /
  `buttonTitleOne`) are removed as orphans across all 8 catalogues.

## What is deliberately unchanged

- The single-document confirm gate and the jobs-panel handover move **verbatim** into a new
  `usePrototypeBuild` hook. `ActionCard` is presentational and had no room for a spinner, an error
  line and a dialog, so it gains one prop: `statusLine?: ReactNode`, rendered beside the existing
  `hint` / `disabledMessage` lines.
- **The button keeps its exact label**, so its accessible name is unchanged. That is what lets
  `ProjectDetail.jobHandover.test.tsx` and `ProjectTabs.test.tsx` pass **untouched** — the evidence
  that this is a move and not a rewrite. The 11 tests from `BuildPrototypeButton.test.tsx` are
  ported with every matcher intact.

## Two things worth a reviewer's attention

**1. `pickNextStep` recommends prototype but can never recommend it while gated.** Not a
coincidence — the gate is "no PRD and no PR-FAQ", which is *exactly*
`documents.hasOutput === false`, and `documents` is earlier in the candidate list, so it wins.
Prototype must stay after documents there; `overviewState.test.ts` pins it, and reordering fails 2
tests.

**2. The billable-action guard is two independent layers, and both were verified load-bearing.**
`POST /projects/{id}/build-prototype` has no existing-prototype check, so a stray click costs money.
The card's `disabled` is the user-facing authority; `runBuild` keeps its own
`if (!hasPrd && !hasPrfaq) return`. Removing the first alone still blocks the build (only the
disabled-message test fails); removing both fails the "does not build" test. Please don't collapse
them as redundant.

## Testing

| Gate | Result |
|---|---|
| `eslint --max-warnings 0` | 0 |
| `tsc -b --noEmit` | 0 |
| frontend vitest | **2089 passed / 137 files** |
| `i18n:check` | `Total missing: 0` |
| `vite build` | clean |

New/updated tests: `OverviewTab.prototypeCard.test.tsx` (12 — the 11 ported plus a visible
disabled-reason case and an error-not-masked case), plus 3 new `overviewState.test.ts` cases for the
recommendation invariant.

**Non-vacuity was mutation-checked, and one finding is worth repeating:** the "an error is never
masked by the started acknowledgement" guarantee holds through *two* independent mechanisms, and
removing either one alone fails nothing — only both together fail the test. So a single mutation
would have wrongly suggested that test was worthless.

## Not verified

**No browser testing.** Test- and build-level only. Deploying next; will report.

## Scope

Frontend only — no Lambda, no IaC, no API change. Locales: +5 keys, −2 orphans, ×8 catalogues.

## Related

First slice of the prototype information-architecture work, chosen because it needs **no product
decision** — it does not presuppose whether prototyping becomes a named phase (the open `U17`
question, which Byron's B6 answers in the other direction). The open question of what the build
control should do when a prototype *already* exists — build another, or offer view/rebuild — is
deliberately **not** settled here; this change only makes the existing behaviour visible.


perrozzi:

This PR moves the Build Prototype control out of the `ProjectTabs` `rightSlot` and into the Overview card grid as step 5, extracting the flow into a new `usePrototypeBuild` hook, adding a `prototype` step to `overviewState`, and adding/removing locale keys across all 8 catalogues. It's frontend-only, deletes more than it adds, and the ported tests plus the new `overviewState` invariant tests are genuinely non-vacuous. Overall this looks like a clean move; I have one likely user-visible correctness issue and a few smaller points.

### Should fix

- `OverviewTab.tsx`: `disabled={steps.prototype.missingUpstream || prototypeBuild.busy}` is paired with an unconditional `disabledMessage={t('documents.prototype.needsDocs')}` ("Create a PRD or a PR-FAQ first…"). `ActionCard`'s render condition isn't in the diff, but every other card's `disabled` is exactly its `missingUpstream`, so the message and the disabled reason always agreed before. Here they can't: whenever a build is in flight — the normal success path, with a PRD present — the card is disabled for a completely different reason while still telling the user they need to create a document. Either gate the message on `missingUpstream` (`disabledMessage={steps.prototype.missingUpstream ? t(...) : undefined}`) or keep `disabled` to `missingUpstream` only and rely on `busy` for the button. There's also no test pinning the message's *absence* when a document exists — `OverviewTab.prototypeCard.test.tsx` only asserts the positive case — so add that; it's the assertion that would have caught this.
- `usePrototypeBuild.ts` calls `projectsApi.buildPrototype` directly with hand-rolled `busy`/`error` state rather than TanStack Query's `useMutation`, and nothing invalidates the project/documents query after a successful start. The behaviour is carried over verbatim so it isn't a regression, but the card now *reports* `Prototypes built: N` off `documents`, so the count is only correct after some other consumer refetches. Worth confirming the jobs panel's refetch invalidates `projectKey(...)`; if it doesn't, the new state line will lie until a manual reload.
- Now that the card knows `prototypeCount > 0` and the description confirms `POST /projects/{id}/build-prototype` has no existing-prototype check, a second click silently starts another billable multi-minute build with no confirm. I accept the "what should the control do when one exists" question is deferred, but the cheap half — reusing the existing `ConfirmModal` when `steps.prototype.hasOutput` — would close a money-losing footgun that the move makes *more* discoverable, not less.

### Nits

- `configureLabel=""` is a sentinel for "no prefix"; making the prop optional (`configureLabel?: string`) and skipping the `<span>` when unset reads better than an empty translated-looking string.
- Several new `t()` calls dropped their `defaultValue` (`documents.prototype.building`, `.button`, `.started`, `.needsDocs`, `overview.prototype*`) while `usePrototypeBuild`'s confirm messages kept theirs. Pick one convention — a missing key now renders the raw path in the button's accessible name.
- `hasPrd`/`hasPrfaq` are re-derived in `OverviewTab` from `documents` while `overviewState` derives `prdCount`/`prfaqCount` from the same array. The comment acknowledges the duplication; exposing the two booleans (or the confirm variant) from `deriveOverviewState` would remove the drift risk entirely.
- `statusLine` renders the error into a plain `<p>` with no `aria-live`, so a failure-to-start is silent for screen-reader users — the same gap the removed hover `title` had, just relocated. Cheap to fix with `role="status"` / `aria-live="polite"` on the wrapper.
- `ActionCardProps` mixes the newly imported `ReactNode` (`statusLine`) with `React.ReactNode` (`buttonIcon`); make them consistent.

---

perrozzi:
Thanks — the first finding was a real bug, and the third was worth doing. Pushed as `2637b21`.

## Should fix

**1. `disabledMessage` unconditional while `disabled` has two reasons — valid, fixed.**

You were right, and right about why: this is the only card in the grid whose `disabled` has two
causes, so it is the only one where the message and the reason can disagree. For the whole duration
of every *successful* build it told a user who plainly had a PRD to go and create one. Now gated:

```tsx
disabledMessage={steps.prototype.missingUpstream ? t('documents.prototype.needsDocs') : undefined}
```

And the assertion you said would have caught it — `does not claim a document is missing while a
build is in flight` — **verified against the old code**: reverting the gate fails it. One wrinkle
worth noting for anyone reading the test: the trigger has to be found by `/building…/i` there, since
the label changes while in flight.

**2. Query invalidation — already handled; adding the test you implied.**

Confirmed as you suggested, and it holds. `handleJobStarted` invalidates only `projectJobsKey`,
which is correct — at that instant there is nothing new to read. The documents arrive when the job
*finishes*, and a separate effect invalidates `projectKey` for any job completed in the last ten
seconds (`useProjectData.ts:150-160`). So the count does refresh, and the state line does not lie.

Your underlying point stood up anyway: **that effect had no test at all**, and it has just acquired
its first visibly-wrong-if-broken consumer. So `useProjectData.jobCompletion.test.tsx` now pins it
(fires on a recent completion, does not fire for a two-hour-old job or a running one), verified by
deleting the invalidation and watching it fail.

I left the hand-rolled `busy`/`error` rather than moving to `useMutation` — it is carried over
verbatim, and converting it would put a behaviour change inside a move, which is exactly what makes
the "did anything change?" question unanswerable. Happy to do it as a follow-up.

**3. Confirm on rebuild — agreed, done.**

Fair, and the framing convinced me: a confirm does not decide the deferred "view vs rebuild"
question, it just stops an unintended charge. Reuses the existing `ConfirmModal` with a new
`documents.prototype.confirmRebuild` across all 8 catalogues, and rebuild outranks the
single-document note when both apply, since spending money on a duplicate is the more consequential
surprise. Four tests, including the precedence case.

## Nits — all taken

- `configureLabel` is now optional rather than an `""` sentinel; the product card drops its empty
  string too.
- **`defaultValue`s dropped everywhere** rather than added back, which is the stricter direction:
  the harness loads the real `en` catalogue, so a key that is missing or has moved renders its raw
  path and fails a test. A `defaultValue` hides precisely that — this repo has shipped buttons
  announcing `editForm` to assistive tech because code and test agreed on a missing key.
- `hasPrd`/`hasPrfaq` now come from `deriveOverviewState` as `prototypeSources`, so availability and
  the confirm wording cannot drift.
- **`aria-live` — good catch, and it changed the design.** The status line is now a polite live
  region. I put it on the card that owns it rather than in `ActionCard`, because wrapping there
  would have added five permanently empty live regions; and it is always mounted, since a region
  that appears at the same moment as its text can be missed entirely.
- `ReactNode` made consistent.

One knock-on: extracting the nested ternary your linter rule caught merged `needsConfirm` and the
message key into a single `confirmKeyFor`. They had been two readings of the same state that could
have disagreed — an open dialog with the wrong text, or no dialog at all.

## Gates

eslint 0, tsc 0, **2098 tests / 138 files**, i18n `Total missing: 0`, build clean.

Still **not browser-verified** — deploying now and I will report what the live UI actually does.

perrozzi:

Since the last review the author addressed all five nits and all three Should-fix items: `disabledMessage` is now gated on `missingUpstream` (with a regression test), the money-losing second-build path is closed by a new `confirmRebuild` gate wired through `hasExistingPrototype` and added to all 8 catalogues, `hasPrd`/`hasPrfaq` now come from `deriveOverviewState` via `prototypeSources`, the status line is a permanently-mounted `role="status"` region, `configureLabel` is optional, and the `defaultValue` convention is unified. The refetch concern was answered by pinning the existing `useProjectData` completion effect in a new test rather than by changing code, which is an acceptable resolution.

### Resolved
- `disabledMessage` no longer claims a missing document while a build is in flight; pinned by a new test.
- Second/duplicate prototype builds now go through `ConfirmModal` (`confirmRebuild`), with rebuild outranking the single-source note, and 4 tests covering confirm/accept/cancel/precedence.
- `hasPrd`/`hasPrfaq` derived once in `overviewState` as `prototypeSources`, removing the drift risk.
- `statusLine` is now a live region that stays mounted across the click that fills it.
- `configureLabel` optional (no `""` sentinel); `ReactNode` used consistently; `defaultValue`s dropped from the confirm keys with a stated rationale.
- Project-refetch-on-completion is now covered by `useProjectData.jobCompletion.test.tsx`, so the new "Prototypes built: N" line has a test behind its freshness.

### New findings

Nits only:

- `usePrototypeBuild.ts`: `hasExistingPrototype?: boolean` defaulting to `false` means a future call site that forgets it silently loses the billable-duplicate guard. There is exactly one caller — making it required costs nothing and makes the guard non-optional by construction.
- `useProjectData.jobCompletion.test.tsx`: the `job()` factory ends in `} as ProjectJob`, which is the partial-cast pattern `OverviewTab.prototypeCard.test.tsx` explicitly argues against in its own header ("a cast is exactly the thing that stops telling the truth when the type gains a field"). Fill the literal or make the helper `Partial`-free for consistency.
- `usePrototypeBuild.ts`: `confirmKeyFor` returns `string | null`, discarding the literal union the `as const` map provides; typing it as `(typeof CONFIRM_MESSAGE)[keyof typeof CONFIRM_MESSAGE] | null` keeps a typo from compiling.
- `OverviewTab.tsx`: the always-mounted `<p className="text-xs mt-2 text-center">` adds its `mt-2` margin on the prototype card even when empty, so that card sits slightly differently from the other five at rest. Also `role="status"` already implies `aria-live="polite"`; one of the two is redundant.
- `OverviewTab.prototypeCard.test.tsx`: the new in-flight test calls `releaseRequest()` at the end without awaiting the resulting state update, which can surface as an act warning depending on timing — `await waitFor(...)` on the label returning to "Build Prototype" would close it.
- With `hasExistingPrototype` true and only one source document, the user no longer learns the build will use just the PRD (the test pins this deliberately). Fine as a precedence call, but a combined message would lose nothing.

---

perrozzi:
Five of six taken, one declined with reasoning. Pushed as `f344a20`.

Two of these earned more than "nit":

**`hasExistingPrototype` optional → required.** Agreed, and this is the one I should have caught
myself. An optional-defaulting-false parameter is the wrong shape for the only thing standing between
a stray second click and another billable build — "a guard a future call site can silently omit is
not a guard" is exactly right. One caller today, so making it mandatory cost nothing.

**The `as ProjectJob` cast — you caught me contradicting my own test header.** Fair, and instructive:
filling the literal immediately failed to compile, because `progress` is required and `project_id`
is not a field on `ProjectJob` at all. So the cast was already hiding two inaccuracies in a
three-week-old fixture shape. That is the argument against partial casts, demonstrated.

Also done:

- `confirmKeyFor` returns the literal union from the `as const` map rather than `string`, so a
  mistyped key no longer compiles.
- Conditional `mt-2`, so the empty region takes no vertical space and the prototype card sits level
  with the other five at rest. And you are right that `role="status"` implies
  `aria-live="polite"` — the duplicate is gone.
- The in-flight test now awaits the state update `releaseRequest()` causes.

**Declined: combining the rebuild and single-source messages.** You marked it optional and I think
optional is the right call. They are different kinds of statement: "it will read only the PRD" is
advisory, "this spends money on a duplicate and keeps the first" is consequential. Merging them
means a ninth translated string across 8 catalogues whose only effect is to make the louder warning
longer and dilute it. If a maintainer prefers the combined message I will add it, but I would rather
not pay for it by default.

Gates: eslint 0, tsc 0, **2098 tests / 138 files**, i18n `Total missing: 0`, build clean.

Deploying now — I will follow up with what the live UI actually does, since everything so far is
test- and build-level only.

perrozzi:

Since the last review the author took all but one of the nits: the in-flight test now awaits the release, the `job()` factory dropped its `as ProjectJob` cast, `confirmKeyFor` returns the literal union, `hasExistingPrototype` is mandatory, and the status region drops the redundant `aria-live` while gating `mt-2` on having text. The remaining item (a combined rebuild + single-source message) was consciously left as a precedence call, which is fine.

### Resolved
- `usePrototypeBuild.ts`: `hasExistingPrototype` is now required, so the billable-duplicate guard cannot be silently omitted by a future call site.
- `usePrototypeBuild.ts`: `confirmKeyFor` returns `ConfirmKey | null`, so a typo'd key no longer compiles.
- `OverviewTab.tsx`: `mt-2` is conditional on the region having text, and `role="status"` no longer duplicates `aria-live`.
- `useProjectData.jobCompletion.test.tsx`: partial cast removed; the literal is now type-checked against `ProjectJob`.
- `OverviewTab.prototypeCard.test.tsx`: `releaseRequest()` is awaited, closing the timing-dependent act warning.

### New findings

Nit only:

- `usePrototypeBuild.ts`: the new `type ConfirmKey` block was inserted *between* the `confirmKeyFor` JSDoc ("The confirm this build needs, or null to start immediately… Rebuild outranks the single-document note") and the function itself, so that doc comment now documents the type alias instead. Move the alias above the doc comment (or below `CONFIRM_MESSAGE`, which also reads better since it depends on it).

Otherwise the delta is clean.

---

perrozzi:
## Deployed and browser-verified

`f344a20` is live on the dev CloudFront distribution. Everything below is observed in a real browser,
which closes the "not browser-verified" caveat from the PR body and both earlier replies.

The locale requests come back as `…/locales/en/projectDetail.json?v=f344a20`, so the site is provably
serving this exact commit rather than a cached build.

### The card, on a project with 5 existing prototypes

```
5. Clickable Prototype
Build a clickable HTML prototype from your PRD / PR-FAQ
Prototypes built: 5
[Build Prototype]
6. Remix Documents
```

Remix is demoted to 6, and the tab row is now just
`Overview | Personas (4) | Product | Documents (9) | AI Chat | MCP Access` — no orphaned control at
the right edge. `Prototypes built: 5` is the state the old button could not report at all.

### The billable guard held

Clicked `Build Prototype` with 5 prototypes present:

- Dialog appeared with `role="dialog"` and `aria-modal="true"`, reading *"This project already has a
  prototype. Building again adds a new one and keeps the existing one. Continue?"*
- **`build-prototype` POSTs in the network log: zero.** Confirmed across the whole
  open → confirm-shown → cancel cycle (44 requests, all GETs plus Cognito auth).
- After cancel: dialog gone, trigger back to `Build Prototype`, not disabled.

### The disabled state, on a project with no documents

`Build Prototype` renders `disabled`, with **"Create a PRD or a PR-FAQ first to enable prototype
build" as visible text** rather than the hover `title` it used to be — the gain that motivated
removing `buttonTitle`/`buttonTitleOne`. `Next step:` correctly reads *"Describe your product or
service."*, so the invariant that a gated prototype is never recommended holds live, not just in the
unit test.

### Your round-2 nits, measured

- **`role="status"` implies polite — confirmed.** The a11y tree shows
  `status atomic live="polite" relevant="additions text"` with no explicit `aria-live` in the markup.
  Your call was right.
- **The `mt-2` fix, numerically.** With the region empty it measures **0px** high with classes
  `text-xs text-center`, and all six cards are **exactly 173px**. The prototype card no longer sits
  lower than its neighbours.

No console errors or warnings.

### Two things I noticed that are NOT from this PR

1. **`Configure & Select & Remix`** — the remix button concatenates `configureLabel` ("Configure & ")
   with "Select & Remix", giving two ampersands. Pre-existing on `development`; I left it alone rather
   than widen this diff, but it is worth a one-line follow-up.
2. **A running build does not disable the button.** During testing the jobs panel showed
   `Prototype Build · running · 40%` while `Build Prototype` stayed enabled — `busy` is local to the
   start request, not "a job is in flight server-side". That is carried over from the old button, and
   the new rebuild confirm now catches the duplicate-spend case that made it matter. Worth considering
   as part of the deferred "what should the control do when one exists" question rather than here.

Full gate, unchanged: eslint 0, tsc 0, 2098 tests / 138 files, i18n `Total missing: 0`, build clean.

PR #296, opened 2026-08-09
fix(ui): tie the prototype confirm to the question that raised it

perrozzi:
## What this fixes, in plain terms

Building a prototype starts a multi-minute **billable** AI job, so the app stops and asks first when
something is off — either *"you only have a PRD, no PR-FAQ, build from just the PRD?"* or *"you already
have a prototype, this makes a second one, continue?"*

Which of those questions applies is computed from the project's documents, and this page refreshes
documents in the background whenever a generation job finishes. So the documents can change **while
that dialog is open on screen**, and the dialog was not reacting to it. Three ways it went wrong:

1. **The dialog went blank.** Your missing PR-FAQ finishes generating, so there is nothing left to warn
   about — and the modal stayed up with its text emptied: a title, Cancel, "Build anyway", and no question.
2. **The dialog silently changed what it was asking.** A prototype finishes instead, so the real warning
   becomes the *more expensive* one. The text was rewritten under the cursor, so "Build anyway" answered
   a costlier question than the one displayed when the user started reading.
3. **The dialog re-opened on its own.** After (1) closed it, the app still believed a dialog was open, so
   the next reason to warn put a modal back on screen with no user action — one reflex from a paid build.

The fix is one idea: **remember which question was asked, not merely that one was asked**, and show the
dialog only while that exact question still applies. The old code kept a boolean, which outlived the
question that raised it.

---

## Summary

Follow-up to #294, found while re-reading the merged hook. The confirm's *reason* is derived from
live data, but its *lifetime* was not.

`confirmKeyFor` reads `hasPrd` / `hasPrfaq` / `hasExistingPrototype`, all derived from the project's
documents — and this page refetches documents whenever a job completes (`useProjectData`, ten-second
window). So the reason to stop and ask can change, or vanish, while the dialog is open.

Holding a boolean `showConfirm` beside that live key gave the dialog a lifetime of its own, with
three ways to misfire:

| Live data changes to… | Boolean flag did | Now |
|---|---|---|
| no reason to ask (PR-FAQ arrives) | stayed up with `message: ''` — a titled modal, Cancel and "Build anyway", **no question** | closes |
| a *different* reason (prototype arrives) | rewrote the text under the cursor, so "Build anyway" answers a **costlier** question than the one displayed | closes; the next click asks again |
| reason gone, then a later one arrives | re-opened a modal the user never asked for, one reflexive click from a multi-minute billable build | stays closed |

The fix is to stop tracking *that* a dialog is open and track *which question was asked*, then open
only while it still matches the derived one:

```diff
-  const [showConfirm, setShowConfirm] = useState(false)
+  const [askedKey, setAskedKey] = useState<ConfirmKey | null>(null)

+  const openKey = askedKey != null && askedKey === confirmKey ? askedKey : null
-      isOpen: showConfirm,
-      message: confirmKey == null ? '' : t(confirmKey),
+      isOpen: openKey != null,
+      message: openKey == null ? '' : t(openKey),
```

All three rows collapse into one rule: **the dialog can only ever show the question the user's own
click raised.** A reason that comes back comes back as the same question, which is the one still
unanswered — so that case is left alone rather than special-cased.

One derivation feeds both `isOpen` and `message`, so they cannot disagree. That is the same reason
`confirmKeyFor` is one function rather than a `needsConfirm` boolean beside a separate message
expression: two readings of one state can disagree, and here the disagreement *is* the bug.

## Why this belongs with #294 rather than being written off as a quirk

It is a **regression against the pre-#294 behaviour**, not a pre-existing edge. The old code computed
the message unconditionally from `hasPrd ? 'prd' : 'prfaq'`, so the same race showed a *wrong but
non-empty* string. Introducing a null-able confirm key is what made blank reachable, so the fix
belongs to the change that introduced it.

## Testing

Three tests, one per row of the table above:

- `closes the confirm rather than emptying it when the reason to ask disappears`
- `closes the confirm rather than swapping its question for a costlier one`
- `does not re-open the confirm when a later reason to ask arrives`

Each opens the single-document confirm, re-renders with a new document set, and asserts the dialog is
gone and no build was started. To make the re-render possible I extracted the element builder
(`overviewTab(documents)`) out of `renderCard`, which is why the test diff is larger than the fix.

**Verified non-vacuous by mutation**, not by inspection — reverting the hook to each earlier
behaviour and re-running the suite:

| Hook reverted to | Tests that fail |
|---|---|
| pre-#296 (`isOpen: showConfirm`) | all 3 |
| #296 as first submitted (`showConfirm && confirmKey != null`) | exactly the 2 new ones |

So the first commit's guard is still load-bearing, and each new test pins something it did not.

Gates: eslint 0, `tsc -b` 0, **2101 tests / 138 files**. (`typecheck:tests` reports 77 errors and
`i18n:check` exits 1 — both identical on the clean base, neither touching these files.)

## Browser-verified against the deployed stack

The earlier revision of this description said the race could not be reliably staged. That was wrong —
it can, without spending a Bedrock job. The refetch does not have to come from a completing job: the
project query inherits the app's 30s `staleTime` and refetches on window focus, so writing the
document straight into DynamoDB and then tabbing back reproduces the identical state transition.

Run against a throwaway fixture project on the live deployment (deleted afterwards; no existing
project touched), reading the dialog out of the DOM rather than eyeballing screenshots:

| Test | Refetch landed | Dialog after |
|---|---|---|
| reason disappears (PRD arrives) | `Documents (1)` → `(2)` | **closed**, not blank |
| reason changes (prototype arrives) | `Prototypes built: 1` appears | **closed**, message *not* swapped to the rebuild text |
| later reason arrives, same page session | `Documents (2)` → `(3)` | **stays closed** |

Non-regression on the same fixture: clicking Build Prototype with a prototype present still opens the
rebuild confirm with the correct message, and Cancel dismisses it. No console errors or warnings
throughout. "Build anyway" was deliberately never clicked — that starts a real billable build.

Deployment itself was verified rather than assumed: the shipped ProjectDetail chunk contains the new
derivation (`g!=null&&g===b?g:null`, `isOpen:_!=null`) and is the chunk the deployed entry bundle
actually references.

### DIFF
--- a/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
@@ -99,6 +99,15 @@ function buildButton() {
   return screen.getByRole('button', { name: /build prototype/i })
 }
 
+/**
+ * The modal's confirm action, queried rather than got: the tests below assert on
+ * its ABSENCE, which is the load-bearing half of a dialog that must not be
+ * clickable once its question no longer applies.
+ */
+function confirmButton() {
+  return screen.queryByRole('button', { name: /build anyway/i })
+}
+
 describe('prototype card confirm gate (U12)', () => {
   beforeEach(() => {
     vi.clearAllMocks()
@@ -278,7 +287,56 @@ describe('prototype card rebuild guard', () => {
 
     rerender(overviewTab([doc('prd', 'prd_1'), doc('prfaq', 'prfaq_1')]))
 
+    // The role, the question and the action, so this cannot go vacuous if the
+    // shell's ARIA role ever changes: the bug being pinned was a dialog that was
+    // still THERE and still clickable, just with nothing written in it.
     expect(screen.queryByRole('dialog')).not.toBeInTheDocument()
+    expect(screen.queryByText(/No PR-FAQ yet/i)).not.toBeInTheDocument()
+    expect(confirmButton()).not.toBeInTheDocument()
+    expect(mockBuildPrototype).not.toHaveBeenCalled()
+  })
+
+  it('closes the confirm rather than swapping its question for a costlier one', async () => {
+    // Same refetch, a different outcome: a prototype arriving while the
+    // single-document confirm is open changes what "Build anyway" MEANS — from
+    // "build from the PRD alone" to "build a second prototype and keep the first".
+    // Rewriting the message under an open dialog would have the button answer a
+    // question the user never read, so the dialog closes and the next click asks.
+    const user = userEvent.setup()
+    const { rerender } = renderCard({ hasPrd: true, hasPrfaq: false })
+
+    await user.click(buildButton())
+    expect(screen.getByText(/No PR-FAQ yet/i)).toBeInTheDocument()
+
+    rerender(overviewTab([doc('prd', 'prd_1'), doc('prototype', 'proto_1')]))
+
+    expect(screen.queryByText(/already has a prototype/i)).not.toBeInTheDocument()
+    expect(screen.queryByText(/No PR-FAQ yet/i)).not.toBeInTheDocument()
+    expect(mockBuildPrototype).not.toHaveBeenCalled()
+  })
+
+  it('does not re-open the confirm when a later reason to ask arrives', async () => {
+    // An open flag outlives the question it was raised for: once the reason
+    // disappeared the flag stayed set, so the NEXT reason to ask put a modal back
+    // on screen that the user never asked for — one reflexive click from a
+    // multi-minute billable build. Tracking which question was asked is what makes
+    // this unreachable.
+    const user = userEvent.setup()
+    const { rerender } = renderCard({ hasPrd: true, hasPrfaq: false })
+
+    await user.click(buildButton())
+    expect(screen.getByText(/No PR-FAQ yet/i)).toBeInTheDocument()
+
+    // The reason to ask disappears: the PR-FAQ generation completes.
+    rerender(overviewTab([doc('prd', 'prd_1'), doc('prfaq', 'prfaq_1')]))
+    expect(confirmButton()).not.toBeInTheDocument()
+
+    // A later job produces a prototype, which is a new reason to ask.
+    rerender(overviewTab([doc('prd', 'prd_1'), doc('prfaq', 'prfaq_1'), doc('prototype', 'proto_1')]))
+
+    expect(screen.queryByRole('dialog')).not.toBeInTheDocument()
+    expect(screen.queryByText(/already has a prototype/i)).not.toBeInTheDocument()
+    expect(confirmButton()).not.toBeInTheDocument()
     expect(mockBuildPrototype).not.toHaveBeenCalled()
   })
 })
--- a/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
@@ -67,13 +67,9 @@ function doc(documentType: ProjectDocument['document_type'], id: string): Projec
  * `hasPrd`/`hasPrfaq` props become the presence of a PRD / PR-FAQ document — the
  * same condition, now derived where the rest of the grid derives its state.
  */
-function renderCard(props: { hasPrd: boolean; hasPrfaq: boolean; hasPrototype?: boolean }) {
-  const documents: ProjectDocument[] = []
-  if (props.hasPrd) documents.push(doc('prd', 'prd_1'))
-  if (props.hasPrfaq) documents.push(doc('prfaq', 'prfaq_1'))
-  if (props.hasPrototype === true) documents.push(doc('prototype', 'proto_1'))
-
-  return render(
+/** The tab for a given document set, so a test can re-render with a different one. */
+function overviewTab(documents: ProjectDocument[]) {
+  return (
     <OverviewTab
       project={project}
       personas={[]}
@@ -85,8 +81,17 @@ function renderCard(props: { hasPrd: boolean; hasPrfaq: boolean; hasPrototype?: 
       onOpenProductTool={vi.fn()}
       onSaveKiroPrompt={vi.fn()}
       onJobStarted={mockJobStarted}
-    />,
+    />
   )
+}
+
+function renderCard(props: { hasPrd: boolean; hasPrfaq: boolean; hasPrototype?: boolean }) {
+  const documents: ProjectDocument[] = []
+  if (props.hasPrd) documents.push(doc('prd', 'prd_1'))
+  if (props.hasPrfaq) documents.push(doc('prfaq', 'prfaq_1'))
+  if (props.hasPrototype === true) documents.push(doc('prototype', 'proto_1'))
+
+  return render(overviewTab(documents))
 }
 
 /** The build trigger, not the modal's confirm button. */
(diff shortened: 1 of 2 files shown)


perrozzi:

This is a small, well-scoped UI correctness fix: `usePrototypeBuild`'s confirm dialog is now gated on both `showConfirm` and a non-null `confirmKey`, so the modal closes instead of rendering with an empty message when the derived reason disappears (e.g. a PR-FAQ generation completing during the document refetch window). The accompanying test extracts an `overviewTab(documents)` builder so it can re-render with a new document set. The fix and comment are accurate and the test targets the right state transition; overall it looks clean, with one follow-on state-management concern below.

### Should fix

- `voc-datalake/frontend/src/pages/ProjectDetail/usePrototypeBuild.ts`: the guard hides the dialog but leaves `showConfirm === true`. Since `confirmKey` is derived from live data, it can become non-null again without any user action — e.g. the user opens the single-document confirm, a refetch adds the missing PR-FAQ (dialog closes, `showConfirm` still set), then a later job adds a `prototype` document and `confirmKey` flips to the rebuild key, re-opening a modal the user never asked for. Consider resetting the flag when the reason disappears (an effect that calls the same `onCancel`/setter when `confirmKey == null`, or storing the key that was confirmed rather than a boolean) and add a regression test for the reappearance path.

### Nits

- `OverviewTab.prototypeCard.test.tsx`: the new test's only positive assertion about closure is `queryByRole('dialog')`. The rest of the file asserts on message text (`/No PR-FAQ yet/i`) — asserting both the message text and the "Build anyway" button are gone would keep the test from silently going vacuous if the modal's ARIA role ever changes.
- The comment block in `usePrototypeBuild.ts` is now longer than the logic it explains; the first sentence plus the "empty message" consequence would carry it, with the #294 history left to git blame.

---

perrozzi:

Since the last review the author went past the one-line guard: `showConfirm: boolean` is replaced by `askedKey: ConfirmKey | null`, and the dialog is open only while `askedKey === confirmKey`, plus two new tests covering the question-swap and the re-open-on-a-later-reason path. Both of my previous points are addressed.

### Resolved
- `usePrototypeBuild.ts`: storing the asked key instead of a boolean removes the spurious re-open when a later reason (`prototype` document) arrives — and it now also blocks the message being rewritten under an open dialog; regression tests exist for both.
- `OverviewTab.prototypeCard.test.tsx`: closure assertions now cover role, message text and the "Build anyway" button, so the test can't go vacuous on an ARIA-role change.

### New findings

Nothing blocking. Two nits:

- `usePrototypeBuild.ts`: `askedKey` survives a transient `confirmKey === null` (e.g. documents momentarily empty mid-refetch), so the dialog can close and re-open with the same question — a flicker rather than a wrong prompt, and the comment states this is intentional, but if you'd rather it stay closed once dismissed by data, clearing `askedKey` in an effect when `confirmKey == null` would do it.
- The comment block grew again (~12 lines for a one-expression derivation); the first sentence plus "so a boolean could disagree with the message" would carry it — the rest is git-blame material. Non-blocking.

---

perrozzi:
Addressed in e8c8477. All three points taken.

**Should fix — the flag outliving its question.** Correct, and the reappearance path you traced is
real. Rather than resetting the flag from an effect, I took the second option you offered and stored
the key that was asked instead of a boolean:

```ts
const [askedKey, setAskedKey] = useState<ConfirmKey | null>(null)
const openKey = askedKey != null && askedKey === confirmKey ? askedKey : null
```

This turned out to close a third case neither of us had listed. Comparing keys does not just stop the
dialog *re-opening* on a reason the user never saw — it also stops the message being **rewritten
under an open dialog**, which is what happens today when a prototype arrives while the
single-document confirm is up: `prd` → `rebuild`, same non-null key, so the old guard held `isOpen`
true and silently swapped the text. "Build anyway" then answers a costlier question than the one on
screen when the user started reading. That is the more dangerous of the two, since it ends in a
billable build rather than a stray modal.

No effect needed, and one derivation now feeds both `isOpen` and `message` — the same argument as
`confirmKeyFor` being one function rather than a boolean beside a separate message expression.

I left the one remaining case alone deliberately: if the *identical* reason returns, the dialog
returns with it. It is the same question, with the same text, that the user still has not answered.

**Nit — `queryByRole('dialog')` alone could go vacuous.** Agreed; the bug being pinned was a dialog
that was still there and still clickable, so the role was the weakest of the three available signals.
All three close-assertions now check the role, the question text, and the confirm button, via a
`confirmButton()` helper.

**Nit — comment longer than the logic.** Trimmed; the #294 history is left to git blame. What remains
is the mechanism (why a key and not a boolean) and the three failure modes it rules out, which I do
not think survives being shortened further given the race is not visible from the call site.

Also worth recording: the new tests are **mutation-verified**, not eyeballed. Reverting the hook to
pre-#296 fails all three; reverting to #296-as-first-submitted fails exactly the two new ones — so
your point was load-bearing and the first commit's guard still is too.

perrozzi:
Deployed to the dev stack and browser-verified. Correcting my own earlier note in this PR: I said the
race could not be reliably staged against a live environment. It can, and cheaply.

The refetch does not have to come from a completing generation job. The project query inherits the
app's global 30s `staleTime` and refetches on window focus, so the document set can be changed
underneath an open dialog by writing the document directly and then tabbing back — the identical state
transition, no Bedrock spend, no ten-second window to hit. I ran it against a throwaway fixture
project (`ZZ PR296 race test`, deleted afterwards; the three real projects were left untouched and
re-verified byte-identical in shape after teardown), reading the dialog out of the DOM rather than
judging it from a screenshot.

| Test | Refetch confirmed landed | Dialog |
|---|---|---|
| reason disappears (PRD arrives) | `Documents (1)` → `(2)` | closed, not blank |
| reason changes (prototype arrives) | `Prototypes built: 1` appears | closed; message **not** swapped to the rebuild text |
| later reason arrives, same page session | `Documents (2)` → `(3)` | stays closed |

The "refetch landed" column is the part that matters: my first attempt at the second test reported a
closed dialog while `Documents` was still `(1)` — the refetch had never happened, so the pass was
vacuous. The cause was the 30s `staleTime`; the earlier tests only worked because minutes of tool
calls had elapsed. Every row above asserts the new document is visible in the UI *before* concluding
anything about the dialog.

Non-regression on the same fixture: with a prototype present, Build Prototype still opens the rebuild
confirm with the right message and Cancel dismisses it. No console errors or warnings throughout.
"Build anyway" was never clicked — it starts a real multi-minute billable build, which is the thing
this PR is protecting.

Deployment verified rather than inferred from a zero exit code: the shipped ProjectDetail chunk
contains the new derivation (`g!=null&&g===b?g:null` / `isOpen:_!=null`) and is the chunk the deployed
entry bundle references.

Gates unchanged: eslint 0, `tsc -b` 0, 2101 tests / 138 files.

perrozzi:

Comment-only delta: the two explanatory blocks in `usePrototypeBuild.ts` were condensed into a single four-line note above `openKey`, and the redundant comment on `confirm.isOpen`/`message` was dropped. No behavioural change, so both previously-resolved findings stand and my one non-blocking nit (comment verbosity) is now addressed.

### Resolved
- `usePrototypeBuild.ts`: comment block trimmed to the load-bearing sentence; rationale is no longer git-blame material in-line.

### New findings

None. Optional nit: the condensed comment no longer records that a returning reason re-raises the *same* question (the intentional non-clearing of `askedKey` on a transient `confirmKey == null`) — the interface doc on `confirm` covers the invariant, so this is fine as-is, but a half-sentence there would keep that decision discoverable.

---

perrozzi:
Round 2 disposition — thanks, both nits are useful and they land differently.

**Comment length: taken, and it was a fair hit.** I claimed to have taken this in round 1 and then
grew the block straight back to ~12 lines for a one-expression derivation. Now 4 lines, and I deleted
the three-line restatement above `isOpen` that said the same thing a second time (14 → 4). Kept what
`openKey` is and why a boolean beside a live-derived key can disagree with the message; dropped the
enumeration of the three failure modes, which the test names carry and git blame explains. Pushed as
`70599b7` — comments only, verified by stripping comment lines from the diff and confirming the
remainder is empty.

**The transient-`null` flicker: declining, because the stated premise is not reachable.** The
scenario is "documents momentarily empty mid-refetch", and neither half of that can happen here:

- A background refetch **retains** the previous `data` (TanStack keeps the cached value while
  `isFetching`), so `documents` never blinks empty on the way through.
- And when `data` genuinely is absent, `ProjectDetail.tsx` early-returns on `data?.project == null`
  **above** the `const { project, personas, documents } = data` destructure — so `OverviewTab`, and
  therefore `usePrototypeBuild`, is not mounted at all. There is no surviving `askedKey` in that
  state; the state is destroyed with the subtree.

So the only way `confirmKey` returns to the *same* non-null value is a document genuinely
disappearing and coming back (a delete, then an undo/refetch). That is the case I left intentionally:
same question, same text, still unanswered by the user. Clearing `askedKey` in an effect would add a
`useEffect` to buy a behaviour change only in a state that needs a real document deletion to reach —
and it would trade the current property, *the dialog can only ever show the question your click
raised*, for one extra rule to hold in your head. Happy to add it if you'd rather be strict.

Stopping the loop here per the nit-only convention. Gates: eslint 0, `tsc -b` 0, 2101 tests / 138
files. Deployed and browser-verified as described above; the trim is comment-only so that verification
still stands.

PR #293, opened 2026-08-09
feat(prioritization): make PRDs scorable alongside PR/FAQs

perrozzi:
## Summary

- `collectPRFAQs` in `prioritizationUtils.ts` now collects both `prd` and `prfaq` documents (previously only `prfaq`). A new exported helper `isScorable(doc)` defines the set of scorable types and is tested independently.
- Each row in `PRFAQRow.tsx` now shows a `DocumentTypeBadge` (using the existing `text-xs px-2 py-0.5 rounded-full` badge idiom) so that PRDs and PR/FAQs with identical titles are unambiguously distinguishable. Purple for PR/FAQ, blue for PRD.
- The empty state in `Prioritization.tsx` is split: projects with no documents at all get the original "No Documents Found" / "Create a PRD or PR/FAQ…" message; projects whose documents are all non-scorable types (research, prototype, etc.) get a new "No Scorable Documents" / "Open a project and generate a PRD or PR/FAQ…" message.
- The summary stat previously labelled `stats.totalPrfaqs` is now `stats.totalDocuments` ("Total Documents"), accurately describing the mixed list.
- All seven new i18n keys (`docType.prd`, `docType.prfaq`, `empty.wrongTypeTitle`, `empty.wrongTypeDescription`, `stats.totalDocuments`) are added to all eight locale catalogues (en, de, es, fr, ja, ko, pt, zh) in case-insensitive sorted position. The old `stats.totalPrfaqs` key is left in each catalogue for backward compatibility.

## Deliberately left alone (per task scope)

- No changes to API handlers, Lambda, CDK stack, or document generators.
- No changes to score storage shape (DynamoDB `PRIORITIZATION/SCORES` item).
- No changes to prototype builder gating or PR/FAQ requirement anywhere.
- No reformatting of files not otherwise changed.
- Pre-existing build failure (`npm run build` fails because `voc-datalake/node_modules/` is not installed and `ts-node` is unavailable for the prebuild script) — this is noted but not fixed as it is out of scope.
- Pre-existing test failure in `RouteErrorBoundary.test.tsx > offers a working path back home` (AbortSignal/jsdom incompatibility, unrelated to this change) — left alone per task rules.

## Build and test results

Verified inside `voc-datalake/frontend/`:

| Command | Result |
|---------|--------|
| `node_modules/.bin/tsc --noEmit` (typecheck) | ✅ 0 errors |
| `node_modules/.bin/tsc -b` (build typecheck) | ✅ 0 errors |
[...]

PR #299, opened 2026-08-09
Put Export and MCP Access side by side, with pickers collapsed

perrozzi:
Two changes to the Export / MCP tab, together because either alone reads worse than both. Frontend only; no backend, no new dependency, no new locale keys.

## Side by side, from `lg` up

Stacked, the MCP Access card sat below the fold — and that is backwards for the flow, because the token it issues is a **prerequisite** for the autoseed snippet beside it (the paste hint says "Requires an API token from the MCP Access tab").

It also split cause from effect. The persona/document selection lives in the Export card, while the MCP card's `curl` URL is built from that same selection — so changing a checkbox altered something that was off screen.

The layout string is named once, because both return paths (normal and token-error) have to match, and the error path needed its two children wrapped so the grid still gets exactly two columns. `items-start` keeps the shorter card from being stretched to its neighbour's height.

All three grid children also carry `min-w-0`, which is **defensive rather than load-bearing** — and I'd rather state that accurately than repeat the reasoning it was added under. The theory was that grid children default to `min-width: auto` so they cannot shrink below their content, and the autoseed `curl` snippet is a long unbroken URL that `whitespace-pre-wrap` cannot break, so its track would push past the even split.

Measured on the deployed page, that does not happen, with or without the class: the snippet's container is `overflow-y-auto`, and per CSS a non-`visible` value on one axis makes the other compute to `auto`, so it is already a scroll port and the long token never contributes to any ancestor's min-content width. Removing `min-w-0` from both children at CSS 2000px and at CSS 1062px left the tracks identical (840/840 and 371/371) with no page overflow.

Keeping it as cheap insurance for a future descendant that isn't inside a scroll port, but happy to drop it if you'd rather the diff carried nothing unjustified.

## Picker sections start collapsed

The rows were the tallest thing on the tab and the least often needed. Each section header already states its selection as `Personas (3/4)` and carries Select all / Deselect all, so the count — the thing you need before copying — is visible without them. Expanding is for *changing* a selection, not for reading it.

This is also what makes two columns look right, rather than one very tall card beside a short one.

## Verified

Against `npm run mock` rather than by deploying, since this is layout:

| Check | Result |
|---|---|
| Two columns on a wide viewport | computed `gridTemplateColumns` = two equal columns at CSS 2000px |
| Stacks below the `lg` breakpoint | **one** column at CSS 950px and 750px |
| Collapsed on load | 0 checkboxes rendered, both counts visible |
| Expanding still works | reveals exactly the expanded section's rows |

It happened to run in German, so the counts were confirmed as `Zielgruppen (2/2)` / `Dokumente (2/2)` rather than only in English.

**Now verified on the deployed site too, including the gap this section used to name.** The normal (tokens-present) branch is not reachable in local dev — the mock has no `api-tokens` endpoint — so it was previously covered only by unit tests. Deployed and checked against a real project (4 personas, 10 documents, one active token):

| Check | Result |
|---|---|
| Normal branch, not the error state | MCP card shows Generate Token / Active Tokens (1) / MCP Client Configuration / Kiro Autoseed |
| Two columns, evenly split | `gridTemplateColumns` 840px + 840px at CSS 2000px; 371 + 371 at CSS 1062px; equal `top` |
| Grid has exactly two children | `childCount` 2 on the normal branch |
| Stacks below `lg` | CSS 875px → a single 827px track, cards at different tops, no page overflow |
| Collapsed on load | 0 checkboxes rendered at both widths |
| Counts visible while collapsed | headers read `Personas (4/4)` and `Documents (10/10)` |
| Expanding still works | reveals exactly the 4 persona rows, all checked |
| Console | 0 errors, 0 warnings |

One pre-existing detail, unchanged by this PR and worth knowing rather than fixing here: at narrow-but-still-two-column widths the `curl` snippet scrolls horizontally inside its own dark box (~119px of scroll at CSS 1062px). It is contained by that box's own scroll port and causes no page-level overflow.

**What is and isn't asserted about the layout.** The *structure* is pinned: both return paths are asserted to give the grid exactly two children, since a third sibling would silently start a second row at half width, and the token-error path is the one that has to wrap its two elements to satisfy that. The *styling* is deliberately not asserted — no Tailwind class assertions and no `gridTemplateColumns` read — because this suite is already criticised for class coupling, and a jsdom assertion would pin something a browser check confirms better.

The grid is located by an exported `data-testid`, not by walking up from the card with `.closest('div.border')`, so the assertion doesn't break for the wrong reason the day a card gains a wrapper or a nested bordered element.

## Test changes

Three existing tests emptied the selection by iterating checkboxes, which are no longer rendered by default. They now use the bulk control — visible when collapsed, and closer to what a user actually does. The `docTypes` group-label test expands the section first, since that heading lives in the section body.

Two new tests pin the default, both mutation-verified against the old expanded-by-default state. The second needed strengthening to earn that: `toHaveLength(1)` passed under mutation for the wrong reason, because with both sections open the *document* row satisfied the count while the personas section was shut. It now names the row it expects and asserts the other section stayed closed.

The second commit fixes two defects in those same tests: the deselect loop was unbounded, and its exit condition is the behaviour under test — so a regression in the bulk control would have spun it to the test timeout instead of failing with a usable message. It is now bounded with an explicit throw, and shared rather than duplicated across two tests.

Two further tests came out of review, one per return path, asserting the two-children invariant described above. They were made to *discriminate* the branches rather than merely go red: the token-error test waits on error-branch-only copy (`mcp.notAvailable`) instead of the Export card title, which renders in **both** branches — awaiting the shared string let the assertion check the normal branch twice while appearing to cover the error one. Confirmed by mutation: unwrapping the error branch's column so the grid gets three children fails exactly that one test (`expected 3 to be 2`) and leaves the tokens-load test green.

## Gates

`tsc -b --noEmit` clean, `eslint . --max-warnings 0` clean, `vitest` 138 files / **2117** tests. Measured at `e28ab82`.

`i18n-check` is **unchanged from `development`** and worth a note so it is not read as this branch's doing: `development` reports `missing 0 | empty 0 | extra 12 | untranslated 7`. I confirmed that baseline on a clean worktree of `development`. All seven untranslated are one key, `prioritization.docType.prfaq` = `PR/FAQ` across the seven non-English catalogues, new since #293. That value is correct — it is a product acronym and should be identical everywhere — and the checker has no allowlist for legitimate cognates while its short-string hatch does not reach six characters. Translating it would ship seven mistranslations, so it wants the per-entry allowlist instead and is left alone here.

## Note for whoever merges

There is an in-flight change to the same file that excludes prototype and product-report documents from the export. It touches the picker's grouping and types near the top of `McpAccessTab.tsx`, while this touches the selection default and the card container, so they should merge cleanly — but whichever lands second is worth a re-run of the tab's tests.

### DIFF
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
@@ -256,6 +256,36 @@ describe('McpAccessTab — export guard is enforced in the UI', () => {
   })
 })
 
+describe('McpAccessTab — the picker sections start collapsed', () => {
+  it('shows each selection count without rendering its rows', async () => {
+    renderTab(onePersona, oneDocument)
+
+    // The count and the bulk control are what a user needs before copying, so
+    // they must be visible with the section shut.
+    expect(await screen.findByRole('button', { name: /Personas \(1\/1\)/ })).toBeInTheDocument()
+    expect(screen.getByRole('button', { name: /Documents \(1\/1\)/ })).toBeInTheDocument()
+    expect(screen.getAllByRole('button', { name: /^Deselect all$/ })).toHaveLength(2)
+
+    // The rows themselves are not rendered until asked for.
+    expect(screen.queryAllByRole('checkbox')).toHaveLength(0)
+  })
+
+  it('renders the rows once a section is expanded', async () => {
+    renderTab(onePersona, oneDocument)
+    const user = userEvent.setup()
+
+    await user.click(await screen.findByRole('button', { name: /Personas \(1\/1\)/ }))
+
+    // Named, not counted: a bare toHaveLength(1) also passes when the personas
+    // section is shut and the DOCUMENT row is the one on screen — verified by
+    // mutation, where it went green against the old expanded-by-default state.
+    const personaRow = screen.getByRole('checkbox', { name: /Persona A/ })
+    expect(personaRow).toBeChecked()
+    // and the other section stayed shut
+    expect(screen.queryByRole('checkbox', { name: /Doc A/ })).not.toBeInTheDocument()
+  })
+})
+
 describe('McpAccessTab — Card 2 hides the curl snippet when it would mean "everything"', () => {
   it('suppresses the autoseed prompt, not just its copy button, once a section is emptied', async () => {
     renderTab(onePersona, oneDocument)
@@ -321,6 +351,13 @@ describe('McpAccessTab — autoseed group labels resolve', () => {
     // catch a deletion. A render assertion is the only check that would fail.
     // Needs a document: the group heading is per document-type.
     renderTab(onePersona, oneDocument)
+
+    // The documents picker is collapsed by default and the group heading lives in
+    // its body, so expand it the way a user would. Matched on the header's
+    // accessible name, which carries the "1/1" selection count.
+    await userEvent.setup().click(
+      await screen.findByRole('button', { name: /Documents \(1\/1\)/ }),
+    )
 
     expect(await screen.findByText(enProjectDetail.autoseed.docTypes.prd)).toBeInTheDocument()
     expect(screen.queryByText(/autoseed\.docTypes\./)).not.toBeInTheDocument()
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
@@ -256,6 +256,38 @@ describe('McpAccessTab — export guard is enforced in the UI', () => {
   })
 })
 
+describe('McpAccessTab — the tab body lays out as exactly two columns', () => {
+  // The layout is a two-column grid, so a THIRD sibling would silently start a
+  // second row at half width. Asserting the child COUNT pins that invariant
+  // without coupling to the Tailwind classes that implement it — the token-error
+  // path in particular has to wrap its two elements to satisfy this.
+  const columnCount = () => {
+    const card = screen.getByText(enProjectDetail.export.title).closest('div.border')
+    const grid = card?.parentElement
+    return grid == null ? null : grid.children.length
+  }
+
+  it('gives the grid exactly two children when tokens load', async () => {
+    mockListApiTokens.mockResolvedValue({ tokens: [] })
+    renderTab(onePersona, oneDocument)
+
+    await screen.findByText(enProjectDetail.export.title)
+    expect(columnCount()).toBe(2)
+  })
+
+  it('gives the grid exactly two children when the token request fails', async () => {
+    mockListApiTokens.mockRejectedValue(new Error('token fetch failed'))
+    renderTab(onePersona, oneDocument)
+
+    // Waits on error-branch-ONLY copy. Awaiting export.title instead would prove
+    // nothing here: the Export card renders in both branches, so the assertion
+    // could run before the rejection settled and silently check the normal
+    // branch twice.
+    await screen.findByText(enProjectDetail.mcp.notAvailable)
+    expect(columnCount()).toBe(2)
+  })
+})
+
 describe('McpAccessTab — the picker sections start collapsed', () => {
   it('shows each selection count without rendering its rows', async () => {
     renderTab(onePersona, oneDocument)
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.structure.test.tsx
@@ -21,7 +21,7 @@ import { render, screen } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import i18n from 'i18next'
-import McpAccessTab from './McpAccessTab'
+import McpAccessTab, { COLUMNS_TESTID } from './McpAccessTab'
 import { canCopyExport } from './autoseedSelection'
 import deProjectDetail from '../../../public/locales/de/projectDetail.json'
 import enProjectDetail from '../../../public/locales/en/projectDetail.json'
@@ -261,11 +261,11 @@ describe('McpAccessTab — the tab body lays out as exactly two columns', () => 
   // second row at half width. Asserting the child COUNT pins that invariant
   // without coupling to the Tailwind classes that implement it — the token-error
   // path in particular has to wrap its two elements to satisfy this.
-  const columnCount = () => {
-    const card = screen.getByText(enProjectDetail.export.title).closest('div.border')
-    const grid = card?.parentElement
-    return grid == null ? null : grid.children.length
-  }
+  //
+  // Found by test id rather than by walking up from the card: reaching through
+  // `.closest('div.border')` would break for the wrong reason the day the card
+  // gains a wrapper or a nested bordered element.
+  const columnCount = () => screen.getByTestId(COLUMNS_TESTID).children.length
 
   it('gives the grid exactly two children when tokens load', async () => {
     mockListApiTokens.mockResolvedValue({ tokens: [] })
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
@@ -98,9 +98,12 @@ function renderTabWithData(projectId = 'proj-123') {
  * spin to the test timeout instead of failing here with a usable message. There
  * are only ever two sections.
  */
+/** Personas and documents — the two sections a picker can offer. */
+const PICKER_SECTION_COUNT = 2
+
 async function deselectEverySection(user: ReturnType<typeof userEvent.setup>) {
   const deselectAll = () => screen.queryAllByRole('button', { name: /^Deselect all$/ })
-  for (let guard = 0; guard < 4; guard += 1) {
+  for (let attempt = 0; attempt <= PICKER_SECTION_COUNT; attempt += 1) {
     const remaining = deselectAll()
     if (remaining.length === 0) return
     await user.click(remaining[0])
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
@@ -86,6 +86,28 @@ function renderTabWithData(projectId = 'proj-123') {
   )
 }
 
+/**
+ * Empties every picker section via its bulk control.
+ *
+ * Bulk rather than per-checkbox because the sections are collapsed by default, so
+ * the individual rows are not rendered — but each header's Select all / Deselect
+ * all always is. Each click flips that section's control to "Select all", so the
+ * query is re-run rather than holding stale handles.
+ *
+ * Bounded: if a regression stopped the label flipping, an unbounded loop would
+ * spin to the test timeout instead of failing here with a usable message. There
+ * are only ever two sections.
+ */
+async function deselectEverySection(user: ReturnType<typeof userEvent.setup>) {
+  const deselectAll = () => screen.queryAllByRole('button', { name: /^Deselect all$/ })
+  for (let guard = 0; guard < 4; guard += 1) {
+    const remaining = deselectAll()
+    if (remaining.length === 0) return
+    await user.click(remaining[0])
+  }
+  throw new Error('Deselect all never stopped appearing — the bulk control is not flipping state')
+}
+
 const mockToken = {
   token_id: 'tok-1',
   name: 'My Kiro token',
@@ -352,14 +374,7 @@ describe('McpAccessTab \u2014 AutoseedContent', () => {
     // Starts enabled \u2014 all items are pre-selected
     expect(copyBtn).toBeEnabled()
 
-    // Empty the selection via each section's bulk control. The picker sections are
-    // collapsed by default so the individual rows are not rendered, but the
-    // header's Select all / Deselect all always is.
-    for (;;) {
-      const remaining = screen.queryAllByRole('button', { name: /^Deselect all$/ })
-      if (remaining.length === 0) break
-      await user.click(remaining[0])
-    }
+    await deselectEverySection(user)
 
     expect(copyBtn).toBeDisabled()
   })
@@ -415,20 +430,10 @@ describe('McpAccessTab \u2014 ExportCard', () => {
     renderTabWithData()
 
     // The copy button starts enabled because all items are pre-selected.
-    // Empty the selection via each section's bulk control, then verify the button
-    // becomes disabled. Bulk rather than per-checkbox because the picker sections
-    // are collapsed by default, so the individual rows are not rendered — but the
-    // header's Select all / Deselect all always is.
     const copyBtn = screen.getByRole('button', { name: /Copy to clipboard/i })
     expect(copyBtn).toBeEnabled()
 
-    // Each click flips that section's control to "Select all", so re-query rather
-    // than holding stale handles.
-    for (;;) {
-      const remaining = screen.queryAllByRole('button', { name: /^Deselect all$/ })
-      if (remaining.length === 0) break
-      await user.click(remaining[0])
-    }
+    await deselectEverySection(user)
 
     expect(copyBtn).toBeDisabled()
   })
--- a/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx
@@ -86,6 +86,9 @@ function renderTabWithData(projectId = 'proj-123') {
   )
 }
 
+/** Personas and documents — the two sections a picker can offer. */
+const PICKER_SECTION_COUNT = 2
+
 /**
  * Empties every picker section via its bulk control.
  *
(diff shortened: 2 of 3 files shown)


perrozzi:

This is a frontend-only layout change to the Export / MCP tab: the two cards move into a two-column grid from `lg` up (via a shared `TWO_COLUMN_LAYOUT` constant so both the normal and token-error return paths match), and the persona/document picker sections now start collapsed. Tests are updated to drive selection through the bulk "Deselect all" control instead of iterating rows, plus two new tests pinning the collapsed default. No backend, IAM, Bedrock, ingestor, or i18n surface is touched, so those repo rules don't apply here. The change is small and well-reasoned; my concerns are about the parts of the layout the diff doesn't show and about grid overflow.

### Should fix

- `voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.tsx` — the normal (tokens-present) branch only has its container class changed; the diff doesn't show its children being wrapped the way the error branch was. The grid needs *exactly two* children for the intended layout. If that branch renders three siblings (export card, MCP access card, autoseed `CollapsibleSection`) the third will start a second grid row half-width. The PR description says this branch was not verified in a browser, so please confirm the child count — and ideally guard it with a structural test (e.g. the grid container has 2 element children), which pins the invariant without asserting Tailwind classes.
- Grid items default to `min-width: auto`, so a child containing a long unwrapped line (the `curl` snippet, the MCP JSON config, export text) can push its track past the 1fr split and break the "two equal columns" result. Consider `min-w-0` on the grid children (including the new error-branch wrapper `div`) alongside the existing overflow handling. A jsdom/mock check of `gridTemplateColumns` will not catch this; only a real browser with the snippet rendered will.
- The collapsed default removes the only affordance that shows *which* items are selected. `Personas (3/4)` tells you the count but not the identity, and the copy actions produce output derived from that identity. That's a deliberate product call and fine to keep, but worth confirming with whoever owns the tab that "3/4" is enough context before a user copies an export.

### Nits

- `TWO_COLUMN_LAYOUT` class order: if the repo uses `prettier-plugin-tailwindcss` conventions, `items-start` belongs before the `lg:` variant (`grid items-start gap-4 lg:grid-cols-2`).
- `McpAccessTab.test.tsx`: the `guard < 4` bound is a magic number; deriving it (e.g. `PICKER_SECTION_COUNT * 2`) or naming it makes the intent self-documenting.
- `deselectEverySection` always clicks `remaining[0]`; if a future regression re-renders the same control without flipping, the message is accurate but the loop still burns all four iterations. Not worth changing unless a third section is added.
- The 12-line block comment on `TWO_COLUMN_LAYOUT` and the multi-line rationale comments in the tests are more prose than the surrounding file's style; some of it reads better as PR description than as code comment.

---

perrozzi:
_(Answering round 1 on `9664fdd`. Addressed in `245a71d`, pushed right after this comment.)_

Good review — the top finding was a question I had genuinely left unverified, and you were right that nothing pinned it.

### The normal branch does have exactly two children — and now it is pinned

I read it rather than assuming: the MCP card's wrapper `div` closes immediately before the grid closes, so the autoseed `CollapsibleSection` is inside that card, not a third sibling. No second half-width row.

But "I checked" is not durable, so I took your suggestion and added a structural test for **both** branches asserting the child **count**, not the classes that implement it. That is better than the no-test position I had, and it avoids deepening the class coupling this suite is already criticised for.

Both are mutation-verified and they discriminate:

| Mutation | Fails |
|---|---|
| third sibling inserted into the normal branch | only the tokens-load test |
| error branch's column unwrapped | only the token-error test |

**One of them needed fixing to earn that.** My first version awaited `export.title`, which renders in *both* branches — so the assertion could run before the rejection settled and silently check the normal branch twice. The tell was that both tests failed against a normal-branch-only mutation. It now waits on `mcp.notAvailable`, which only the error state renders, and the two tests separate cleanly.

### `min-w-0` — applied, and thank you for the reasoning

You are right on the mechanism: grid items default to `min-width: auto`, and `whitespace-pre-wrap` wraps at whitespace while a long `curl` URL has none, so the track really could exceed the even split. Added to all three grid children including the error-branch wrapper. Also right that neither jsdom nor a `gridTemplateColumns` read can catch it — noted as a residual gap that wants a real browser with a long selection rendered.

### Class order — premise does not hold here, but the point does

This repo has neither `prettier` nor `prettier-plugin-tailwindcss` in its dependencies and no prettier config, so that ordering convention is not in force. The house idiom is better evidence: five existing instances read `grid grid-cols-1 sm:grid-cols-2 gap-4`. Reordered to match, which also makes the single-column base explicit rather than implied:

```
grid grid-cols-1 lg:grid-cols-2 gap-4 items-start
```

### Other nits

- Guard bound is now `PICKER_SECTION_COUNT`, not a bare `4`.
- Trimmed the block comment to the three things that stop someone reverting it: both return paths must match, it takes exactly two children, and why `min-w-0` is required. The rest lives here instead, which was your point.
- `remaining[0]` on every iteration: agreed, not worth changing at two sections.

### Kept, deliberately

The collapsed default does hide *which* items are selected, leaving only `3/4`. That is the requested behaviour and the tradeoff is intentional: the rows were the tallest thing on the tab, the count is what gates the copy action, and expanding is one click. Worth a second opinion from whoever owns the tab, as you say — flagging rather than silently keeping it.

Gates: `tsc` clean, `eslint --max-warnings 0` clean, `vitest` 138 files / 2117 tests. `i18n-check` is unchanged from `development`, whose own baseline is `missing 0 | empty 0 | extra 12 | untranslated 7` — the latter being `prioritization.docType.prfaq` = `PR/FAQ`, which is correct as-is and discussed in the PR description.

perrozzi:

Follow-up review — the incremental push adds the structural two-children test for both render paths, applies `min-w-0` to all three grid children, makes the layout string explicit with `grid-cols-1`, trims the block comment, and replaces the magic `4` loop bound with `PICKER_SECTION_COUNT`. Both of my code-level "should fix" items are addressed.

### Resolved
- Normal (tokens-present) branch child count is now pinned by `McpAccessTab.structure.test.tsx` for both the loaded and rejected token paths, and the error path waits on error-only copy so it can't silently assert the wrong branch.
- Grid overflow: `min-w-0` added to the Export card, the MCP card, and the new error-branch wrapper.
- Magic `guard < 4` replaced by a named `PICKER_SECTION_COUNT` bound.

### Still open
- Product confirmation that `Personas (3/4)` alone is enough context before a copy (not a code issue; owner sign-off).

### New findings
- `McpAccessTab.test.tsx` — the `PICKER_SECTION_COUNT` declaration was inserted *between* the `deselectEverySection` JSDoc and the function, so that doc block now attaches to the constant instead of the helper and the constant carries a second, duplicate-ish comment. Move the constant above the JSDoc.
- `McpAccessTab.structure.test.tsx` — `columnCount()` locates the grid via `.closest('div.border')` then `parentElement`, which couples the test to the `border` utility class and to the card being a direct grid child; a future wrapper or a nested bordered div inside the Export card would make this fail for the wrong reason. A `data-testid` on the grid container (or querying by the export title's grid ancestor via a stable hook) would be more durable. Non-blocking.

### Nits
- `TWO_COLUMN_LAYOUT` class order still puts `items-start` after the `lg:` variant; if `prettier-plugin-tailwindcss` runs in CI it will reorder.

---

perrozzi:
_(Answering round 2 on `245a71d`. Addressed in `e28ab82`, pushed right after this comment.)_

Both new findings were mine, and both are fixed.

### Stray JSDoc — yes, I split a doc block from its function

`PICKER_SECTION_COUNT` went in between `deselectEverySection`'s JSDoc and the function, so the block documented the constant. Constant moved above it, and I dropped the now-redundant "there are only ever two sections" sentence from the JSDoc since the named constant states it.

### Column lookup — agreed, and this one was worse than a style point

`.closest('div.border')` + `parentElement` coupled the assertion to the `border` utility class *and* to the Export card being a direct grid child. Either a wrapper around the card or a nested bordered element inside it would have failed the test for a reason unrelated to the invariant.

It now finds the container by `data-testid`, with the id exported from the component so the test cannot drift from the literal:

```tsx
export const COLUMNS_TESTID = 'export-mcp-columns'
```

Worth noting this also removes class coupling from the one place this PR introduced it — which is the standing criticism of this suite rather than a new problem. And it is the established idiom here, not something I am importing: 194 `data-testid` occurrences in production components and 425 `getByTestId` calls in tests, including `ModalShell.tsx`.

**Re-verified both mutations through the new selector**, because a test-id lookup is exactly the kind of change that can make an assertion pass trivially:

| Mutation | Fails |
|---|---|
| third sibling in the normal branch | only the tokens-load test |
| error branch's column unwrapped | only the token-error test |

### Class order — closing this one with evidence

Raised again, conditioned on `prettier-plugin-tailwindcss` running in CI. Three things make it moot here: this repo has no `prettier` and no such plugin in `package.json`, there is no prettier config file, and **the repo has no CI at all** (`statusCheckRollup` is empty on every PR, which is also why every gate figure in this thread is a local run). Nothing will reorder the string. The house idiom — five existing instances of `grid grid-cols-1 sm:grid-cols-2 gap-4` — puts the variant after the base columns, which is what it already does.

### The one genuinely open item is not mine to close

Product confirmation that `Personas (3/4)` is enough context before a copy. Flagging rather than deciding: the collapsed default was specifically requested by the tab's owner, who has seen it running, so I have treated that as the sign-off — but it is worth their explicit word rather than my inference, and I would rather it be visible here than assumed.

Gates: `tsc` clean, `eslint --max-warnings 0` clean, `vitest` 138 files / 2117 tests, `i18n-check` unchanged from `development` (its `extra 12` / `untranslated 7` baseline is discussed in the PR description).

perrozzi:

Follow-up: the incremental push replaces the class-coupled `.closest('div.border')` walk with an exported `COLUMNS_TESTID` applied to both return paths, and moves `PICKER_SECTION_COUNT` above the `deselectEverySection` JSDoc so the doc block reattaches to the helper. Both of my previous new findings are addressed; nothing new of substance.

### Resolved
- `McpAccessTab.structure.test.tsx` now locates the grid via `getByTestId(COLUMNS_TESTID)` instead of the `div.border` / `parentElement` walk.
- `McpAccessTab.test.tsx` — `PICKER_SECTION_COUNT` no longer sits between the JSDoc and `deselectEverySection`; the duplicate comment line is gone.

### Still open
- Product sign-off that the collapsed header count (`Personas (3/4)`) is sufficient context before a copy.

### Nits
- `TWO_COLUMN_LAYOUT` still orders `items-start` after the `lg:` variant; if `prettier-plugin-tailwindcss` runs in CI it will reorder (unchanged from last review).
- `COLUMNS_TESTID` is a test-only export from a component module; documented and harmless, but worth keeping to the one grid container it names.

Otherwise clean.

---

perrozzi:
_(Answering review round 3 on `e28ab82`, which is the current head. No code changes in response — this round is nit-only, so per convention I'm dispositioning rather than pushing another commit.)_

Thanks — agreed that both round-2 findings are addressed. Taking the remaining items one at a time, because "nit" and "wrong" aren't the same thing and one of them is checkable.

### Nit 1 — `TWO_COLUMN_LAYOUT` class order / `prettier-plugin-tailwindcss`

Still refuted, same as last round: **this repo has neither Prettier nor that plugin.**

- `prettier` appears in **no** `package.json` in the repo (root, `voc-datalake`, `frontend`, `lambda/stream`).
- There is no `.prettierrc*`, `prettier.config.*`, or `prettier` key anywhere.
- There is also no CI at all on this repo, so there is no pipeline that could run it.

The order I used matches the existing house idiom rather than the plugin's — there are several `grid grid-cols-1 sm:grid-cols-2 gap-4` declarations already in `frontend/src`, and `items-start` trailing the responsive variant is consistent with them. Changing it would make this one line inconsistent with its neighbours to satisfy a formatter that isn't installed.

(Flagging the mechanism rather than just the conclusion: the review bot doesn't read comments, so a point answered in prose reappears each round. Not a complaint — just so the thread stands on its own.)

### Nit 2 — keep `COLUMNS_TESTID` to the one container it names

Agreed, and that's already the case. It has exactly three use sites:

- the declaration in `McpAccessTab.tsx`,
- the two `TWO_COLUMN_LAYOUT` grid containers (normal + token-error return paths),
- one import in `McpAccessTab.structure.test.tsx`.

It is deliberately not a general-purpose hook, and the reason it's exported rather than a literal in the test is drift: the assertion is about *this* container, so the name should have one definition.

### "Still open" — is the collapsed header enough context before a copy?

This is a product call rather than a defect, but the factual half of it is worth pinning down, because if the count *weren't* visible while collapsed then defaulting the sections shut would be a real data-scope problem, not a preference.

It is visible, and in every language:

- `PickerSection`'s title is interpolated — `t('autoseed.personas', { selected, total })`.
- `autoseed.personas` / `autoseed.documents` carry `({{selected}}/{{total}})` in **all 8 catalogues** (en/de/fr/es/pt/ja/ko/zh), so a collapsed header reads `Personas (3/4)` rather than a bare label.
- The `Select all` / `Deselect all` control also stays outside the collapsible region.

Two tests hold that: one asserts both counted headers are present *and* that zero checkboxes are rendered, the other that expanding one section reveals only its own rows. The second names the row (`{ name: /Persona A/ }`) instead of counting checkboxes, because a bare `toHaveLength(1)` also passes when the personas section is shut and the *document* row is the one on screen — that one went green against the old expanded-by-default state until it was made to name the row.

So: happy to change the default if the preference is rows-open, but the "user copies without knowing what's selected" risk isn't present as written.

### Verification at this head

- `tsc -b --noEmit` — exit 0
- `eslint . --max-warnings 0` — exit 0
- `vitest run` — **2117 passed / 138 files**

I also re-ran the mutation probe on the invariant this round's change touches, since a structural assertion that can't fail is worse than none. Unwrapping the token-error branch's MCP column so the grid gets three children fails **exactly one** test — `gives the grid exactly two children when the token request fails` (`expected 3 to be 2`) — while `…when tokens load` stays green. That asymmetry is the point: it's what round 1 correctly said was missing, since a mutation to the *normal* branch previously reddened both and so proved they weren't discriminating. Reverted afterwards; tree is byte-identical to `e28ab82`.

Stopping the review loop here on the nit-only round. Note there's no CI and no formal approval on this PR, so a human still needs to look at it before merge — the `McpAccessTab.tsx` layout diff is the part worth a pair of eyes, since a layout regression there is silent.

perrozzi:
_(No code change in this update — deployed the branch head, live-tested it, and correcting one of my own claims. Body updated with the results.)_

## Deployed and behaviour-verified

Deployed this branch to the dev environment and tested it in a browser. The pre-deploy ancestry check was clean, which is worth recording because it is not always: `origin/development` is `8804ff0`, which is exactly this branch's merge-base, and `git log HEAD..origin/development` is **empty** — so deploying the head added only these four commits and could not revert anything that landed since the branch was cut. Live `APP_VERSION` now reads `e28ab82`, the invalidation completed, and the served page chunk is **byte-identical** to the local build (matching sha256).

**The gap the body used to name is now closed.** The normal, tokens-present branch is unreachable in local dev because the mock has no `api-tokens` endpoint, so until now it was covered only by unit tests. Verified against a real project with 4 personas, 10 documents and one active token:

| Check | Result |
|---|---|
| Normal branch, not the error state | MCP card renders Generate Token / Active Tokens (1) / MCP Client Configuration / Kiro Autoseed |
| Two columns, evenly split | `gridTemplateColumns` 840 + 840 at CSS 2000px, 371 + 371 at CSS 1062px, equal `top` |
| Grid has exactly two children | `childCount` 2 on the normal branch |
| Stacks below `lg` | CSS 875px → single 827px track, cards at different tops, no page overflow |
| Collapsed on load | 0 checkboxes at both widths |
| Counts visible while collapsed | `Personas (4/4)`, `Documents (10/10)` |
| Expanding still works | exactly the 4 persona rows, all checked |
| Console | 0 errors, 0 warnings |

That also settles the round-3 "still open" item empirically rather than by argument: the selection count really is readable with the sections shut, on real data, so defaulting them collapsed does not hide what a copy would include.

## 🔴 Correction: `min-w-0` is defensive, not load-bearing — my stated rationale was wrong

Round 1 asked for `min-w-0` on the grid children and I justified it as *"the `curl` URL is a long unbreakable token, so without this its track pushes past the even split."* That reasoning does not survive measurement, and since I had just written it into the PR body I would rather retract it than leave it standing.

I removed `min-w-0` from both children on the live page and re-measured. The tracks were **identical** with and without it — 840/840 at CSS 2000px, 371/371 at CSS 1062px, no page overflow either way.

The reason is in the ancestor chain. The snippet sits inside `div.bg-gray-900.rounded-lg.p-4.max-h-48.overflow-y-auto`, and per CSS `overflow` computation, when one axis is not `visible` the other computes to `auto` — so that div's `overflow-x` is `auto` and it is already a scroll port. A long token inside a scroll port contributes nothing to any ancestor's min-content width, so it can never widen the grid track. The automatic-minimum-size problem `min-w-0` exists to solve is already solved one level down.

I have **kept** the class, on the grounds that it is a zero-cost guard for a future descendant that isn't inside a scroll port — but the honest description is insurance, not a fix, and I have rewritten the body to say so. Happy to drop it in a follow-up commit if you would rather the diff carried nothing that isn't demonstrably doing work; I did not push that unilaterally because it would spend a review round for no user-visible change.

Worth noting the condition was genuinely exercised, so this isn't a case of the test missing the scenario: the rendered snippet contains a 92-character unbroken URL token inside a 371px column.

## One pre-existing detail, not fixed here

At narrow-but-still-two-column widths the `curl` snippet scrolls horizontally inside its own dark box — about 119px of scroll at CSS 1062px. It predates this PR (the box and its classes come from the previous change), it is contained by that box's own scroll port, and it produces no page-level horizontal scroll. Flagging it rather than folding an unrelated fix into a layout PR.

Gates unchanged at this head: `tsc -b --noEmit` clean, `eslint . --max-warnings 0` clean, 2117 tests / 138 files.

⚠️ The dev environment now runs this **unmerged** branch. It is a strict superset of `development`, so nothing is missing, but it should be redeployed from the merge commit once this lands so "which commit is live" matches `development` again.

PR #302, opened 2026-08-09
fix(stream): bound all free-text fields and unify language allowlist (issue #266)

### LINKED ISSUE #266: [Security]: Stream request schema has no length caps on message, context or history
## Summary

The streaming chat request schema caps array *counts* but not string *lengths*. `message`, `context`,
history `content` and attachment `data` are unbounded, so a single authenticated request can drive
arbitrary Bedrock token spend.

## Severity

Medium. Authenticated, but the cost is per-request and unbounded, and the tool loop multiplies it.

## Risk

Every `.max()` in `lambda/stream/src/schema.ts`:

| line | field | cap |
|---|---|---|
| 31 | `days` | 365 |
| 46 | `attachments` | 5 (count) |
| 48 | `history` | 50 (count) |

Uncapped:

- `:28` `message: z.string().min(1, 'Message is required')`
- `:30` `context: z.string().optional()`
- `:34` `response_language: z.string().optional()`
- `:22-24` history entry `content: z.string().min(1)` — 50 entries, each unbounded
- `:17` attachment `data: z.string().min(1, ...)` — 5 attachments, each unbounded
- `project_id`, `selected_personas`, `selected_documents`

So the counts are bounded and the payload is not. The tool loop runs up to `MAX_TOOL_LOOPS = 15`
(`lambda/stream/src/handler.ts:49`), and the context is re-sent each iteration.

Separately, `response_language` reaches the prompt unvalidated —
`lambda/stream/src/context/voc-context.ts:113-114`:

```ts
const name = names[lang] ?? lang;
return `IMPORTANT: You MUST respond entirely in ${name} (${lang}). ...`
```

An unrecognized value is interpolated into the system prompt verbatim. The same pattern is duplicated in
`context/persona-prompt.ts:8-16`. That is prompt-injection surface, not just a cost issue.

## Proposed mitigation

- Add `.max()` to `message`, `context`, history `content` and attachment `data`, sized to real use. API
  Gateway's own payload limit is the current de-facto bound and is far larger than anything legitimate.
- Constrain `response_language` to the supported locale set (`z.enum`) rather than passing a free string
  into the prompt. Where an unsupported value is acceptable, fall back to English instead of interpolating.
- Bound `selected_personas` / `selected_documents` counts.
- Consider a total-payload budget check, since the individual caps still multiply across history entries
  and attachments.

## Acceptance criteria

- [ ] Every free-text field in the schema has an explicit maximum.
- [ ] `response_language` is validated against known locales and never interpolated raw.
- [ ] An oversized request is rejected with 400 before any Bedrock call.
- [ ] Tests cover an over-cap `message`, an over-cap history entry, and an unknown `response_language`.

## Affected files

`lambda/stream/src/schema.ts`, `lambda/stream/src/context/voc-context.ts`,
`lambda/stream/src/context/persona-prompt.ts`

perrozzi:
## Summary

Closes #266.

Bounds every free-text field in the streaming chat request so an oversized payload cannot reach Bedrock, and removes a channel that let a caller's own string be interpolated into the model's system prompt.

The shape of the fix changed materially during review — this description reflects the code as it now stands, not the original approach.

- **Length bounds on every free-text field**, with two different policies depending on who authored the value (below).
- **Single language module** (`src/context/language.ts`) replacing two divergent copies of `getLanguageInstruction` that disagreed on which codes they accepted, with one runtime membership test (`isSupportedLanguage`) used by the schema.
- **`response_language` allowlist**: unrecognised codes are coerced to `undefined` (English fallback) and can never be interpolated verbatim into the system prompt.
- **Client-side counterparts** for the bounds a real user can hit, so they get a translated message before sending rather than an opaque `400` afterwards.

## Two policies, by who authored the value

**User-authored values REJECT.** The caller typed or pasted it, so telling them is the useful response, and the frontend carries a matching limit (`frontend/src/api/streamLimits.ts`, pinned by a lockstep test that reads this schema's source) so they are stopped before sending.

**Service-authored text CLAMPS.** `history` replays answers *this service generated*: chat streams with a 16 000-token ceiling and the client sends the answer back on the next turn. Any per-turn bound below the model's own output ceiling therefore turned one normal long answer into a `400` on **every subsequent message** in that conversation — with no field named in the error and the answer still on screen. Truncating degrades; rejecting killed the conversation. Clamping is also the established pattern for context in this codebase, where persona and document context are truncated rather than refused.

## Bounds

| Field | Bound | Policy | Reasoning |
|---|---|---|---|
| `message` | 8 000 chars | reject, mirrored client-side | ~2 000 tokens. This is an analytics tool: pasting a review, a support thread or a slice of a PRD is normal input. Affordable because the request is not the dominant cost term — one question can drive up to `MAX_TOOL_LOOPS` rounds of corpus context. |
| `context` | 500 chars | reject, no mirror | Built by `buildChatContext()` from three single-valued filters, each value capped, so its output is provably under 500 for filter values of *any* length. Reaching it would be a bug, not user input. |
| `history[].content` | derived from `MAX_OUTPUT_TOKENS` | **clamp** | The longest answer the service can emit is the longest turn it must accept back, so the bound is derived rather than chosen and the two cannot drift. |
| `history` aggregate | 4 × the per-turn bound | **clamp** | Oldest turns drop off — standard sliding window. Keeps the cost envelope without ever refusing a request. |
| `history` array | 10 × the window | reject | Sanity bound only, an order of magnitude above the product limit, so no real conversation meets it. Stops an absurd array being validated element-by-element before the window discards it. |
| `project_id`, selected ID items | 128 chars | reject | UUID is 36. |
| `selected_personas` / `selected_documents` | 20 items | reject; `@all` clamped client-side | Roundtable makes one Bedrock call per persona against a 300 s timeout, so the cap is worth keeping. `@all` expands to every persona and persona import appends without replacing, so it is clamped rather than allowed to 400. |
| `attachment.name` | 255 chars | reject | Filesystem filename limit. |
| `attachment.data` | 2 800 000 chars | reject | Sanity bound. **Not** the binding constraint: 5 × that exceeds the API Gateway request-payload limit, so the transport refuses first. |

## Languages the model may be asked to answer in

`ar, de, en, es, fr, hi, it, ja, ko, nl, pl, pt, ru, sv, tr, zh`

Deliberately **wider** than the set the UI is translated into (eight). The frontend's `i18n/languages.ts` decides what the *interface* is translated into; this decides what the *model* may be told to reply in — a different question, since the model answers fluently in languages we ship no UI catalogue for.

An earlier revision of this PR dropped `it, nl, ru, ar, hi, sv, pl, tr` for having no UI catalogue, which silently forced English on any API client asking for them. That was unnecessary: safety comes from interpolating a name out of `LANGUAGE_NAMES` and never the caller's string, which holds for a table of any size — so list size is a coverage decision, not a security one. They are restored, and `language.test.ts` now asserts they are accepted so the regression cannot be reintroduced by trimming the list back to match the UI.

**Known limitation:** region and script subtags (`pt-BR`, `zh-Hans`) are rejected. Supporting them properly means canonicalising with `Intl.getCanonicalLocales` + `Intl.DisplayNames` instead of widening a hand table — full coverage with nothing to maintain, and the interpolated name comes from ICU rather than the caller. Left out because it needs the Lambda runtime's ICU build verified first.

## What each guard is enforced by

Worth stating precisely rather than implying the tests cover everything:

- The **language list** is enforced by **typecheck** — `LANGUAGE_NAMES` is keyed by `SupportedLanguage`, so trimming the list makes both it and the test's expected-names map excess-keyed (`TS2353`). The tests pin the intent. The *runtime* guard derives from the names table, not the list, which is the deliberate "cannot accept a language it cannot name" property.
- The **cross-package limits** are enforced by `streamLimits.lockstep.test.ts`, which reads this schema's source and fails on drift. It deliberately does not `skipIf` when the sibling package is absent — a skipped lockstep is a silent lockstep — but the failure names the path, the working directory and the reason.
- The **`context` bound** rests on `buildChatContext` capping each clause; `chatContext.test.ts` asserts the worst case against a 10 000-char value and pins one-clause-per-filter, which is the structural property the arithmetic depends on.

## Verification

- `tsc --noEmit` (stream) → 0 errors
- `eslint src/` (stream) → 0 errors, 1 pre-existing warning in `indexes.test.ts`
- `vitest run` (stream) → **303 tests / 22 files**
- `tsc -b --noEmit` (frontend) → 0 errors
- `eslint . --max-warnings 0` (frontend) → clean
- `vitest run` (frontend) → **2179 tests / 145 files**
- i18n: `missing: 0`, nothing flagged in `chat.json`. The gate exits 1 on the pre-existing `prioritization.docType.prfaq` false positive in 7 locales, which this branch does not touch.

Mutation-probed rather than assumed: the history contiguity guard, the persona clamp, the lockstep, the `context` cap boundary and the language list each fail exactly the assertions that should care when the corresponding logic is broken.

## Not done

**This has never been deployed.** Everything above is test and static-analysis evidence. The live check that matters is: ask something that produces a long answer, then send a follow-up — the exact path that was broken.

Two deliberate residues, documented in the code rather than left to be rediscovered:

1. An explicit 21-item @-mention selection still errors rather than silently dropping hand-picked choices.
2. When `@all` clamps, the UI does not say so; `MentionContextBar` renders nothing for that case, so surfacing it needs new props for a condition that takes repeated persona imports to reach. `clampedFrom` is already returned so the notice can be added without re-deriving the condition.

### DIFF
--- a/voc-datalake/lambda/stream/src/context/language.ts
+++ b/voc-datalake/lambda/stream/src/context/language.ts
@@ -1,0 +1,47 @@
+/**
+ * Shared language-instruction helper.
+ *
+ * Single source of truth for:
+ *   - the supported locale set (aligned with the eight shipped UI catalogues)
+ *   - the human-readable language names used in model system prompts
+ *   - the `getLanguageInstruction` function consumed by both voc-context.ts and
+ *     persona-prompt.ts
+ *
+ * The extra codes previously listed only in voc-context.ts (it, nl, ru, ar, hi,
+ * sv, pl, tr) had no UI catalogue behind them.  They are intentionally omitted
+ * from SUPPORTED_LANGUAGES so that `response_language` validation in schema.ts
+ * can use this set as the single allowlist.  An unrecognised locale is silently
+ * treated as English (the no-instruction case) rather than rejected — older or
+ * third-party clients degrade gracefully.
+ */
+
+/** Locale codes that have a shipped UI catalogue and are accepted by the API. */
+export const SUPPORTED_LANGUAGES = ['de', 'en', 'es', 'fr', 'ja', 'ko', 'pt', 'zh'] as const;
+
+export type SupportedLanguage = (typeof SUPPORTED_LANGUAGES)[number];
+
+const LANGUAGE_NAMES: Record<SupportedLanguage, string> = {
+  de: 'German',
+  en: 'English',
+  es: 'Spanish',
+  fr: 'French',
+  ja: 'Japanese',
+  ko: 'Korean',
+  pt: 'Portuguese',
+  zh: 'Chinese',
+};
+
+/**
+ * Build the "respond in <language>" system-prompt instruction.
+ *
+ * Returns an empty string for English (or when no language is specified) because
+ * the model responds in English by default — an explicit instruction adds noise.
+ * Returns an empty string for any code that is not in SUPPORTED_LANGUAGES; the
+ * schema's `.catch(undefined)` transform means that unrecognised values arrive
+ * here as `undefined`, so this branch is defensive-only.
+ */
+export function getLanguageInstruction(lang: SupportedLanguage | undefined): string {
+  if (!lang || lang === 'en') return '';
+  const name = LANGUAGE_NAMES[lang];
+  return `IMPORTANT: You MUST respond entirely in ${name} (${lang}). All text, headings, labels, and explanations must be in ${name}.`;
+}
--- a/voc-datalake/lambda/stream/src/context/language.ts
+++ b/voc-datalake/lambda/stream/src/context/language.ts
@@ -4,6 +4,7 @@
  * Single source of truth for:
  *   - the supported locale set (aligned with the eight shipped UI catalogues)
  *   - the human-readable language names used in model system prompts
+ *   - `isSupportedLanguage`, the ONE runtime membership test for the allowlist
  *   - the `getLanguageInstruction` function consumed by both voc-context.ts and
  *     persona-prompt.ts
  *
@@ -32,16 +33,42 @@ const LANGUAGE_NAMES: Record<SupportedLanguage, string> = {
 };
 
 /**
+ * Derived from LANGUAGE_NAMES so membership and naming are the SAME data: a
+ * locale cannot be accepted by the allowlist but left unnamed, and the exhaustive
+ * Record above makes the reverse a compile error. Keyed by `string` rather than
+ * SupportedLanguage on purpose — that is what lets the guard below test an
+ * unknown value without an assertion, and what makes `get()` honestly return
+ * `string | undefined`.
+ */
+const NAME_LOOKUP: ReadonlyMap<string, string> = new Map(Object.entries(LANGUAGE_NAMES));
+
+/**
+ * Runtime membership test for the locale allowlist.
+ *
+ * This is the only place membership is decided.  schema.ts uses it to sanitise
+ * `response_language`, so the allowlist and the language names cannot drift
+ * apart into the two divergent copies this module replaced.  A type guard
+ * rather than an assertion, per the repo convention.
+ */
+export function isSupportedLanguage(value: unknown): value is SupportedLanguage {
+  return typeof value === 'string' && NAME_LOOKUP.has(value);
+}
+
+/**
  * Build the "respond in <language>" system-prompt instruction.
  *
  * Returns an empty string for English (or when no language is specified) because
  * the model responds in English by default — an explicit instruction adds noise.
- * Returns an empty string for any code that is not in SUPPORTED_LANGUAGES; the
- * schema's `.catch(undefined)` transform means that unrecognised values arrive
- * here as `undefined`, so this branch is defensive-only.
  */
 export function getLanguageInstruction(lang: SupportedLanguage | undefined): string {
   if (!lang || lang === 'en') return '';
-  const name = LANGUAGE_NAMES[lang];
+  // Insurance, NOT a fix, and deliberately labelled as such: the sole runtime
+  // entry point parses through chatRequestSchema, whose preprocess step maps
+  // anything outside the allowlist to undefined, so this lookup cannot miss
+  // today. It exists so a future caller that does NOT go through the schema
+  // degrades to "no instruction" instead of interpolating an unvalidated code
+  // verbatim into the system prompt.
+  const name = NAME_LOOKUP.get(lang);
+  if (name === undefined) return '';
   return `IMPORTANT: You MUST respond entirely in ${name} (${lang}). All text, headings, labels, and explanations must be in ${name}.`;
 }
--- a/voc-datalake/lambda/stream/src/context/language.ts
+++ b/voc-datalake/lambda/stream/src/context/language.ts
@@ -2,33 +2,58 @@
  * Shared language-instruction helper.
  *
  * Single source of truth for:
- *   - the supported locale set (aligned with the eight shipped UI catalogues)
+ *   - the set of languages the model may be asked to ANSWER in
  *   - the human-readable language names used in model system prompts
  *   - `isSupportedLanguage`, the ONE runtime membership test for the allowlist
  *   - the `getLanguageInstruction` function consumed by both voc-context.ts and
  *     persona-prompt.ts
  *
- * The extra codes previously listed only in voc-context.ts (it, nl, ru, ar, hi,
- * sv, pl, tr) had no UI catalogue behind them.  They are intentionally omitted
- * from SUPPORTED_LANGUAGES so that `response_language` validation in schema.ts
- * can use this set as the single allowlist.  An unrecognised locale is silently
- * treated as English (the no-instruction case) rather than rejected — older or
- * third-party clients degrade gracefully.
+ * ## This is NOT the shipped-UI-catalogue list, and the distinction matters
+ *
+ * The frontend's own `i18n/languages.ts` decides which languages the *interface*
+ * is translated into (currently eight). This list decides which languages the
+ * *model* may be told to reply in — a different question, because the model
+ * answers fluently in languages we have no UI translation for.
+ *
+ * Conflating the two is how `it, nl, ru, ar, hi, sv, pl, tr` were dropped: they
+ * were removed for having no UI catalogue, which silently forced English on any
+ * API client that asked for them, even though nothing about safety required it.
+ * Safety comes from interpolating a name from LANGUAGE_NAMES below and never the
+ * caller's own string — so extending this list costs nothing and is purely a
+ * coverage decision. They are restored here.
+ *
+ * An unrecognised locale is still silently treated as English (the
+ * no-instruction case) rather than rejected, so older clients degrade gracefully.
+ * Region and script subtags (`pt-BR`, `zh-Hans`) are not accepted; a caller
+ * wanting those needs the base code today.
  */
 
-/** Locale codes that have a shipped UI catalogue and are accepted by the API. */
-export const SUPPORTED_LANGUAGES = ['de', 'en', 'es', 'fr', 'ja', 'ko', 'pt', 'zh'] as const;
+/** Language codes the model may be asked to answer in. */
+export const SUPPORTED_LANGUAGES = [
+  'ar', 'de', 'en', 'es', 'fr', 'hi', 'it', 'ja',
+  'ko', 'nl', 'pl', 'pt', 'ru', 'sv', 'tr', 'zh',
+] as const;
 
 export type SupportedLanguage = (typeof SUPPORTED_LANGUAGES)[number];
 
+// Keyed by the type, so adding a code above without naming it here is a compile
+// error rather than a silently unnamed language.
 const LANGUAGE_NAMES: Record<SupportedLanguage, string> = {
+  ar: 'Arabic',
   de: 'German',
   en: 'English',
   es: 'Spanish',
   fr: 'French',
+  hi: 'Hindi',
+  it: 'Italian',
   ja: 'Japanese',
   ko: 'Korean',
+  nl: 'Dutch',
+  pl: 'Polish',
   pt: 'Portuguese',
+  ru: 'Russian',
+  sv: 'Swedish',
+  tr: 'Turkish',
   zh: 'Chinese',
 };
 
--- a/voc-datalake/lambda/stream/src/context/persona-prompt.ts
+++ b/voc-datalake/lambda/stream/src/context/persona-prompt.ts
@@ -3,15 +3,12 @@
  * that module inside its size budget. The type-only import back into
  * project-context is erased at compile time, so there is no runtime cycle.
  *
- * `getLanguageInstruction` is re-exported from the canonical language module so
- * that project-context.ts has a single import point and does not need to know
- * where the helper lives.
+ * `getLanguageInstruction` lives in ./language.js; consumers import it from there
+ * directly rather than through a re-export here.
  */
 import type { ProjectItem } from './project-context.js';
 import { getLanguageInstruction } from './language.js';
 import type { SupportedLanguage } from './language.js';
-
-export { getLanguageInstruction } from './language.js';
 
 /** The persona's identity block: who they are, what drives them, how to speak. */
 function personaIdentitySection(projectName: string, persona: ProjectItem): string {
--- a/voc-datalake/lambda/stream/src/context/persona-prompt.ts
+++ b/voc-datalake/lambda/stream/src/context/persona-prompt.ts
@@ -2,18 +2,16 @@
  * Roundtable persona prompt building, split out of project-context.ts to keep
  * that module inside its size budget. The type-only import back into
  * project-context is erased at compile time, so there is no runtime cycle.
+ *
+ * `getLanguageInstruction` is re-exported from the canonical language module so
+ * that project-context.ts has a single import point and does not need to know
+ * where the helper lives.
  */
 import type { ProjectItem } from './project-context.js';
+import { getLanguageInstruction } from './language.js';
+import type { SupportedLanguage } from './language.js';
 
-export function getLanguageInstruction(lang?: string): string {
-  if (!lang || lang === 'en') return '';
-  const names: Record<string, string> = {
-    es: 'Spanish', fr: 'French', de: 'German', pt: 'Portuguese',
-    ja: 'Japanese', zh: 'Chinese', ko: 'Korean', it: 'Italian',
-  };
-  const name = names[lang] ?? lang;
-  return `IMPORTANT: You MUST respond entirely in ${name} (${lang}). All text, headings, labels, and explanations must be in ${name}.`;
-}
+export { getLanguageInstruction } from './language.js';
 
 /** The persona's identity block: who they are, what drives them, how to speak. */
 function personaIdentitySection(projectName: string, persona: ProjectItem): string {
@@ -41,7 +39,7 @@ export function buildSinglePersonaPrompt(
   selectedDocumentIds: string[],
   documents: ProjectItem[],
   previousResponses: Array<{ name: string; response: string }>,
-  responseLanguage?: string,
+  responseLanguage?: SupportedLanguage,
 ): string {
   const parts: string[] = [personaIdentitySection(projectName, persona)];
 
--- a/voc-datalake/lambda/stream/src/schema.test.ts
+++ b/voc-datalake/lambda/stream/src/schema.test.ts
@@ -4,7 +4,8 @@
 import { describe, it, expect } from 'vitest';
 import type { z } from 'zod';
 import { chatRequestSchema, attachmentSchema, MAX_MESSAGE_LENGTH } from './schema.js';
-import { MAX_HISTORY_CONTENT_LENGTH, TRUNCATION_MARKER } from './history-budget.js';
+import { SUPPORTED_LANGUAGES } from './context/language.js';
+import { MAX_HISTORY_ARRAY, MAX_HISTORY_CONTENT_LENGTH, TRUNCATION_MARKER } from './history-budget.js';
 
 /** Returns the flattened dot-path of every issue in a failed parse result. */
 function errorPaths(result: z.SafeParseReturnType<unknown, unknown>): string[] {
@@ -280,6 +281,20 @@ describe('chatRequestSchema', () => {
     expect(entry?.content.endsWith(TRUNCATION_MARKER)).toBe(true);
   });
 
+  it('rejects an absurd history array rather than validating every element of it', () => {
+    // The outer bound is an order of magnitude above the window, so this is a
+    // client bug or an attack, not a long conversation.
+    const entry = {
+      role: 'user' as const, content: 'hi',
+    };
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      history: Array.from({ length: MAX_HISTORY_ARRAY + 1 }, () => entry),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('history');
+  });
+
   it('accepts a conversation past the 50-turn window, keeping the most recent turns', () => {
     const history = Array.from({ length: 60 }, (_, i) => ({
       role: 'user' as const, content: `turn-${i}`,
@@ -333,7 +348,9 @@ describe('chatRequestSchema', () => {
 
 describe('chatRequestSchema response_language allowlist (issue #266)', () => {
   it('accepts every supported locale', () => {
-    for (const lang of ['de', 'en', 'es', 'fr', 'ja', 'ko', 'pt', 'zh']) {
+    // Driven from the source of truth, so adding a locale cannot leave this
+    // assertion silently covering the old set.
+    for (const lang of SUPPORTED_LANGUAGES) {
       const result = chatRequestSchema.safeParse({ message: 'hi', response_language: lang });
       expect(result.success).toBe(true);
       expect(result.success && result.data.response_language).toBe(lang);
--- a/voc-datalake/lambda/stream/src/schema.test.ts
+++ b/voc-datalake/lambda/stream/src/schema.test.ts
@@ -2,7 +2,13 @@
  * Tests for Zod request validation schemas.
  */
 import { describe, it, expect } from 'vitest';
+import type { z } from 'zod';
 import { chatRequestSchema, attachmentSchema } from './schema.js';
+
+/** Returns the flattened dot-path of every issue in a failed parse result. */
+function errorPaths(result: z.SafeParseReturnType<unknown, unknown>): string[] {
+  return result.success ? [] : result.error.issues.map((i) => i.path.join('.'));
+}
 
 describe('attachmentSchema', () => {
   it('accepts a valid PNG attachment', () => {
@@ -63,6 +69,44 @@ describe('attachmentSchema', () => {
   it('rejects missing fields', () => {
     expect(attachmentSchema.safeParse({}).success).toBe(false);
     expect(attachmentSchema.safeParse({ name: 'x' }).success).toBe(false);
+  });
+
+  it('rejects attachment name exceeding 255 characters', () => {
+    const result = attachmentSchema.safeParse({
+      name: 'a'.repeat(256),
+      media_type: 'image/png',
+      data: 'abc',
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('name');
+  });
+
+  it('accepts attachment name at exactly 255 characters', () => {
+    const result = attachmentSchema.safeParse({
+      name: 'a'.repeat(255),
+      media_type: 'image/png',
+      data: 'abc',
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects attachment data exceeding 2 800 000 characters', () => {
+    const result = attachmentSchema.safeParse({
+      name: 'big.png',
+      media_type: 'image/png',
+      data: 'a'.repeat(2_800_001),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('data');
+  });
+
+  it('accepts attachment data at exactly 2 800 000 characters', () => {
+    const result = attachmentSchema.safeParse({
+      name: 'big.png',
+      media_type: 'image/png',
+      data: 'a'.repeat(2_800_000),
+    });
+    expect(result.success).toBe(true);
   });
 });
 
@@ -156,8 +200,144 @@ describe('chatRequestSchema', () => {
     const result = chatRequestSchema.safeParse({ message: 'hi', use_web_search: 'yes' });
     expect(result.success).toBe(false);
   });
+
+  it('rejects message exceeding 2 000 characters', () => {
+    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(2_001) });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('message');
+  });
+
+  it('accepts message at exactly 2 000 characters', () => {
+    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(2_000) });
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects context exceeding 500 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      context: 'a'.repeat(501),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('context');
+  });
+
+  it('accepts context at exactly 500 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      context: 'a'.repeat(500),
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects project_id exceeding 128 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      project_id: 'a'.repeat(129),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('project_id');
+  });
+
+  it('accepts project_id at exactly 128 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      project_id: 'a'.repeat(128),
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects a history entry content exceeding 4 000 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      history: [{ role: 'user', content: 'a'.repeat(4_001) }],
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result).some((f) => f.startsWith('history'))).toBe(true);
+  });
+
+  it('accepts history entry content at exactly 4 000 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      history: [{ role: 'user', content: 'a'.repeat(4_000) }],
+    });
+    expect(result.success).toBe(true);
+  });
+
+  it('rejects more than 50 history entries', () => {
+    const entry = { role: 'user' as const, content: 'hi' };
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      history: Array.from({ length: 51 }, () => entry),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('history');
+  });
+
+  it('rejects more than 20 selected_personas', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      selected_personas: Array.from({ length: 21 }, (_, i) => `persona-${i}`),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('selected_personas');
+  });
+
+  it('rejects a selected_persona ID exceeding 128 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      selected_personas: ['a'.repeat(129)],
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result).some((f) => f.startsWith('selected_personas'))).toBe(true);
+  });
+
+  it('rejects more than 20 selected_documents', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      selected_documents: Array.from({ length: 21 }, (_, i) => `doc-${i}`),
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result)).toContain('selected_documents');
+  });
+
+  it('rejects a selected_document ID exceeding 128 characters', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      selected_documents: ['a'.repeat(129)],
+    });
+    expect(result.success).toBe(false);
+    expect(errorPaths(result).some((f) => f.startsWith('selected_documents'))).toBe(true);
+  });
 });
 
+describe('chatRequestSchema response_language allowlist (issue #266)', () => {
+  it('accepts every supported locale', () => {
+    for (const lang of ['de', 'en', 'es', 'fr', 'ja', 'ko', 'pt', 'zh']) {
+      const result = chatRequestSchema.safeParse({ message: 'hi', response_language: lang });
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.response_language).toBe(lang);
+    }
+  });
+
+  it('silently coerces an unknown locale to undefined instead of rejecting', () => {
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      response_language: 'xx-UNKNOWN',
+    });
+    expect(result.success).toBe(true);
+    expect(result.success && result.data.response_language).toBeUndefined();
+  });
+
+  it('does not let an unrecognised code pass through to the parsed value', () => {
+    // The raw attacker string must never appear in the parsed output.
+    const result = chatRequestSchema.safeParse({
+      message: 'hi',
+      response_language: 'prompt-injection-attempt',
+    });
+    const parsed = result.success ? JSON.stringify(result.data) : '';
+    expect(parsed).not.toContain('prompt-injection-attempt');
+  });
+});
 
 describe('chatRequestSchema date_basis (issue #150)', () => {
   it('accepts imported and review values', () => {
--- a/voc-datalake/lambda/stream/src/schema.test.ts
+++ b/voc-datalake/lambda/stream/src/schema.test.ts
@@ -3,7 +3,8 @@
  */
 import { describe, it, expect } from 'vitest';
 import type { z } from 'zod';
-import { chatRequestSchema, attachmentSchema } from './schema.js';
+import { chatRequestSchema, attachmentSchema, MAX_MESSAGE_LENGTH } from './schema.js';
+import { MAX_HISTORY_CONTENT_LENGTH, TRUNCATION_MARKER } from './history-budget.js';
 
 /** Returns the flattened dot-path of every issue in a failed parse result. */
 function errorPaths(result: z.SafeParseReturnType<unknown, unknown>): string[] {
@@ -201,14 +202,21 @@ describe('chatRequestSchema', () => {
     expect(result.success).toBe(false);
   });
 
-  it('rejects message exceeding 2 000 characters', () => {
-    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(2_001) });
+  it('rejects message exceeding the cap', () => {
+    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(MAX_MESSAGE_LENGTH + 1) });
     expect(result.success).toBe(false);
     expect(errorPaths(result)).toContain('message');
   });
 
-  it('accepts message at exactly 2 000 characters', () => {
-    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(2_000) });
+  it('accepts message at exactly the cap', () => {
+    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(MAX_MESSAGE_LENGTH) });
+    expect(result.success).toBe(true);
+  });
+
+  // The old 2 000-char cap was ~300 words, which refused a pasted review or
+  // support thread — a normal input for an analytics tool.
+  it('accepts a pasted excerpt of several thousand characters', () => {
+    const result = chatRequestSchema.safeParse({ message: 'a'.repeat(5_000) });
     expect(result.success).toBe(true);
   });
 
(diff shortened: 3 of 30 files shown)


perrozzi:

This PR bounds every free-text field in the stream Lambda's chat request schema and consolidates two divergent `getLanguageInstruction` implementations into a new `src/context/language.ts` with a single locale allowlist wired into `response_language` via `z.preprocess`. The direction is right — closing the verbatim-locale interpolation surface and hard-capping the dominant Bedrock cost vectors — and the schema tests are thorough with clear revert-catches. However, the new language helper does not actually implement the defensive guard its own docstring claims, and a couple of the caps (notably `history[].content`) look tight enough to break normal multi-turn usage.

### Blockers

- `voc-datalake/lambda/stream/src/context/language.ts`: the docstring says "Returns an empty string for any code that is not in `SUPPORTED_LANGUAGES`; ... this branch is defensive-only", but no such branch exists. The function only checks `!lang || lang === 'en'`, so any unsupported code reaching it (TypeScript types are not a runtime boundary — any caller not fed by `chatRequestSchema` bypasses the allowlist) yields `LANGUAGE_NAMES[lang] === undefined` and produces `You MUST respond entirely in undefined (it-XX)` — i.e. the raw attacker-controlled string *is* still interpolated verbatim into the system prompt. Add the guard and a test:
  ```ts
  const name = LANGUAGE_NAMES[lang];
  if (!name) return '';
  ```
  Note that `language.test.ts`'s own header comment advertises a case ("An unrecognised code does NOT appear in the returned instruction") that is not tested — and would fail today.

- `voc-datalake/lambda/stream/src/schema.ts`: `MAX_HISTORY_CONTENT_LENGTH = 4_000` is applied to *assistant* turns as well as user turns. An analytics assistant answer with tables/markdown routinely exceeds 4 000 characters, and the client replays prior turns in `history`, so the *next* message in a conversation will fail validation with an opaque 400 and the chat becomes unrecoverable. Either raise the cap to match the configured Bedrock `max_tokens` (×~4 chars/token) or, better, clamp/truncate history server-side rather than rejecting — the client is echoing content the server itself generated. Please add a regression test for the "long assistant answer then follow-up question" path.

### Should fix

- **Silently dropping 8 locales is a user-facing regression.** `it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr` previously produced correct non-English answers; they now silently return English with no error. Response language is independent of whether a UI catalogue ships — keeping those entries in `LANGUAGE_NAMES`/`SUPPORTED_LANGUAGES` costs nothing and still closes the injection surface (the dangerous part was the old `names[lang] ?? lang` fallback, not the extra names). If they must be dropped, that belongs in release notes, not only in the PR body.

- **No client-side counterpart for the new caps.** `message` (2 000) and `context` (500) are enforced server-side only; no frontend files are touched, so a user pasting a long excerpt, or a filter combination that serialises past 500 chars, gets a bare rejection with no i18n'd message in any locale. Please verify against how the frontend builds `context` (multi-select sources/categories/sentiments plus free-text search can get long fast) and add client-side validation + i18n keys for all locales, or relax those two caps.

- **Unnecessary `as` assertions in the new/edited tests** — `'es' as SupportedLanguage` in `voc-context.test.ts` and `lang as 'de'` in `language.test.ts` contradict the repo's prefer-type-guards/no-assertion convention the PR description itself cites. Contextual typing already narrows the literal at the call site; drop the assertions (and in `language.test.ts` type the map as `Partial<Record<SupportedLanguage, string>>` and iterate over `SUPPORTED_LANGUAGES`).

- `voc-datalake/lambda/stream/src/context/voc-context.test.ts`: the test named `does not interpolate an unrecognised language code into the system prompt` passes `undefined`, not an unrecognised code — it duplicates the existing `'en'` assertion and does not cover the behaviour its name claims. Drive it through `chatRequestSchema.parse({ message: 'hi', response_language: 'xx-INJECT' })` and assert the raw string is absent from `ctx.systemPrompt`.

- The attachment envelope (5 × 2 800 000 ≈ 14 MB) sits above the transport limit for Lambda Function URL / API Gateway request bodies, so it can never be the binding constraint. Either lower it to something meaningful or add the aggregate size check you mention in "Suggestions for future tasks" — otherwise the documented "hard-bounded" claim in the schema header is misleading.

### Nits

- `language.ts` references "the schema's `.catch(undefined)` transform", but the schema uses `z.preprocess`; the comment is stale on arrival.
- `voc-context.ts` leaves a double blank line where the removed function was.
- `persona-prompt.ts` both imports and re-exports `getLanguageInstruction`; have `project-context.ts` import it from `./language.js` directly and drop the re-export indirection.
- `language.test.ts` uses `if (lang === 'en') continue;` inside a test — prefer `SUPPORTED_LANGUAGES.filter((l) => l !== 'en')` to stay clear of `vitest/no-conditional-in-test`.
- `schema.test.ts` hard-codes the locale list; import `SUPPORTED_LANGUAGES` so the allowlist test cannot drift from the source of truth.

---

voc-datalake/lambda/stream/src/context/language.ts
```diff
@@ -0,0 +1,47 @@
...
+  zh: 'Chinese',
+};
+
+/**
+ * Build the "respond in <language>" system-prompt instruction.
+ *
+ * Returns an empty string for English (or when no language is specified) because
+ * the model responds in English by default — an explicit instruction adds noise.
+ * Returns an empty string for any code that is not in SUPPORTED_LANGUAGES; the
+ * schema's `.catch(undefined)` transform means that unrecognised values arrive
```
perrozzi: **Type**: issue
**Severity**: minor
**Title**: Stale JSDoc comment references `.catch(undefined)` instead of `z.preprocess`

**Description**: Line 40 says _"the schema's `.catch(undefined)` transform means that unrecognised values arrive here as `undefined`"_. However, the implementation in `schema.ts` uses `z.preprocess` (not `.catch(undefined)`). The choice of `z.preprocess` over `.catch(undefined)` was deliberate (explained in the PR description — `.catch(undefined)` triggers `unicorn/no-useless-undefined`), so this comment is now technically incorrect and could confuse a future reader who tries to find the `.catch` call.

**Proposed fix**: Update the comment to reference `z.preprocess`:
```
 * schema's `z.preprocess` sanitiser means that unrecognised values arrive
 * here as `undefined`, so this branch is defensive-only.
```

**AI prompt**: In `src/context/language.ts`, the JSDoc for `getLanguageInstruction` contains the phrase "the schema's `.catch(undefined)` transform". Update it to say "the schema's `z.preprocess` sanitiser" to match the actual implementation in `schema.ts`, which uses `z.preprocess` (not `.catch(undefined)`).

voc-datalake/lambda/stream/src/context/persona-prompt.ts
```diff
@@ -2,18 +2,16 @@
...
-export function getLanguageInstruction(lang?: string): string {
-  if (!lang || lang === 'en') return '';
-  const names: Record<string, string> = {
-    es: 'Spanish', fr: 'French', de: 'German', pt: 'Portuguese',
-    ja: 'Japanese', zh: 'Chinese', ko: 'Korean', it: 'Italian',
-  };
-  const name = names[lang] ?? lang;
-  return `IMPORTANT: You MUST respond entirely in ${name} (${lang}). All text, headings, labels, and explanations must be in ${name}.`;
-}
+export { getLanguageInstruction } from './language.js';
```
perrozzi: **Type**: comment
**Title**: Dual import + re-export of `getLanguageInstruction` in the same file is slightly surprising

**Description**: Lines 11 and 14 each reference `getLanguageInstruction` from `language.js`:

```ts
import { getLanguageInstruction } from './language.js'; // used internally on line 70
export { getLanguageInstruction } from './language.js'; // re-exported for project-context.ts
```

This is valid — the import is consumed internally (line 70) and the export is a convenience re-export so `project-context.ts` can import from a single location. However, the pattern (import + separate re-export of the same symbol from the same source) is slightly unusual and could puzzle a reader unfamiliar with the module graph. A short inline comment explaining why both exist would suffice.

**AI prompt**: In `src/context/persona-prompt.ts`, add a one-line comment above `export { getLanguageInstruction } from './language.js'` explaining that this is a re-export for `project-context.ts`'s benefit, e.g. `// Re-exported so project-context.ts has a single import point.`

voc-datalake/lambda/stream/src/schema.ts
```diff
@@ -1,42 +1,91 @@
...
+  // the trailing .optional() accepts.
+  response_language: z.preprocess(
+    (v) => (typeof v === 'string' && supportedLanguageSet.has(v) ? v : undefined),
+    z.enum(SUPPORTED_LANGUAGES).optional(),
+  ),
   // Project chat fields
-  project_id: z.string().optional(),
-  selected_personas: z.array(z.string()).optional(),
-  selected_documents: z.array(z.string()).optional(),
+  project_id: z.string().max(MAX_ID_LENGTH).optional(),
```
perrozzi, line 112: **Type**: question
**Title**: Should `project_id` (and `selected_personas`/`selected_documents` items) have a `.min(1)` guard?

**Description**: The `message` field has `.min(1)` to reject empty strings, but `project_id` (line 86), and the individual items in `selected_personas` and `selected_documents` (lines 87–88) do not. An empty string `""` is currently a valid `project_id`, which would trigger a DynamoDB query against key `PROJECT#` — a no-op that returns 0 items and throws `NotFoundError`.

The handler falls back to URL path extraction when `project_id` is falsy; but `""` is falsy in JS so an empty string would correctly route to that fallback. If this is intentional (empty string treated the same as absent), a brief comment would prevent future readers from wondering why `.min(1)` is absent here but present on `message`.

**Is the omission of `.min(1)` on `project_id` and ID array items intentional?**

voc-datalake/lambda/stream/src/schema.test.ts
```diff
@@ -156,8 +200,144 @@ describe('chatRequestSchema', () => {
...
+describe('chatRequestSchema response_language allowlist (issue #266)', () => {
+  it('accepts every supported locale', () => {
+    for (const lang of ['de', 'en', 'es', 'fr', 'ja', 'ko', 'pt', 'zh']) {
+      const result = chatRequestSchema.safeParse({ message: 'hi', response_language: lang });
+      expect(result.success).toBe(true);
+      expect(result.success && result.data.response_language).toBe(lang);
+    }
+  });
+
+  it('silently coerces an unknown locale to undefined instead of rejecting', () => {
```
perrozzi, line 360: **Type**: good_point
**Title**: Excellent split between behavioural coercion test and security-focused non-interpolation test

**Description**: The two adjacent tests — `silently coerces an unknown locale to undefined` (lines 322–329) and `does not let an unrecognised code pass through to the parsed value` (lines 331–339) — cleanly separate the *behavioural* contract (coerce to undefined) from the *security* contract (attacker string never appears in output). This makes the intent of each test immediately legible and ensures a regression in either direction is caught by the appropriate named test. The `errorPaths` helper introduced at line 9 is also a nice addition that makes path-based assertions concise and reusable throughout the file.

perrozzi (commented):
## PR Review Summary

This PR delivers a clean, well-scoped security fix. It addresses issue #266 by bounding every free-text field in the streaming chat schema and unifying the two divergent `getLanguageInstruction` implementations into a single canonical module. The implementation is correct, fully tested, and passes all 285 tests with 0 lint errors.

**Build/test results (independently verified):**
- `tsc --noEmit` → 0 errors
- `eslint src/` → 0 errors, 1 pre-existing unrelated warning in `indexes.test.ts`
- `vitest run` → 21 test files, 285 tests, all passed

**Overall assessment:** The code is high quality and safe to merge. Three minor findings are noted below — one is a stale comment that could mislead future readers, one is a mildly redundant import pattern, and one is a minor gap in a test name.

perrozzi:
## Code Review Summary

**Findings:** 1 issue (minor) · 1 comment · 1 question · 1 good point

**Build/test results (independently verified on this branch):**
- `tsc --noEmit` → ✅ 0 errors
- `eslint src/` → ✅ 0 errors, 1 pre-existing unrelated warning in `indexes.test.ts`
- `vitest run` → ✅ 21 test files, 285 tests, all passed

---

### Overall assessment

This is a high-quality, well-motivated security fix. The changes are correct, all cap values are clearly reasoned in the PR description, and the test suite is thorough — every new bound has an over-limit rejection test _and_ an at-limit acceptance test. The `z.preprocess` pattern for `response_language` is the cleanest available option given the ESLint constraints, and the new `language.ts` module cleanly eliminates a subtle divergence between two independent allowlists.

### Key areas to look at before merge

1. **Stale JSDoc comment** (minor, inline comment on `language.ts:40`): The `getLanguageInstruction` JSDoc references `.catch(undefined)` but the actual implementation uses `z.preprocess`. Should be a one-word fix.
2. **`project_id` / ID array items lack `.min(1)`** (question, inline on `schema.ts:86`): Intentional or oversight? Empty string `""` passes validation but the handler treats falsy `project_id` as absent, so it may be harmless. Clarification or a brief comment would help.
3. **Dual import+re-export pattern** in `persona-prompt.ts` (non-blocking comment): Correct but unusual — a one-liner comment would help future readers understand why both lines exist.

### What went well

The `errorPaths` helper and the split between the behavioural test ("coerces to undefined") and the security test ("attacker string never appears in output") are both excellent patterns worth keeping as templates for future schema tests.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01KZM9Q2APGH4T2989W5J83KQX` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-08-09T22:23:59.776Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/302) |
| Duration | 289.8s |
| Cost | $0.8536 |

perrozzi:

**Why it stopped:** 0 blocking, 0 should_fix, build not regressed — decision=commented; 1 nit, 1 needs_author_input, 2 info; build unknown (tolerated); lint unknown (tolerated); mergeable [iteration 0, 1 nit, 1 needs_author_input, 2 info]

| | |
|---|---|
| Outcome | **CONVERGED** |
| Iteration | 0 of 3 |
| Loop spend | $3.30 of $20.00 |
| Goal | # Bound every free-text field in the streaming chat request, and stop interpolating an unvalidated language code into the system prompt This is GitHub issue #266 in this repository. ## The defect voc-datalake/lambda/stream/src/schema.ts caps array **counts** but not string **lengths**. Today atta… |
| Run | [`01KZM8PQH7HZ1164YM3534PK6A`](https://d2wi1pimyj1f69.cloudfront.net/runs/01KZM8PQH7HZ1164YM3534PK6A) |

### Each iteration

|---|---|---|---|---|---|---|
| 0 | 1 nit, 1 needs_author_input, 2 info | — | — | ✅ | — | — |

perrozzi:
Addressed the review, plus a few things that fell out of re-deriving the caps from first principles. Four commits, pushed right after this comment.

**Both blockers.** The first one I could not reproduce as a live issue and the second I could, so they are fixed differently.

- **`history[].content`** was the real one, and it was worse than "tight": a bound below the model's own output ceiling on text *this service generates*. Chat streams at 16 000 tokens and both chat surfaces replay the answer untruncated (`Chat.tsx`, `ChatTab.tsx` are bare `messages.map`), so one long answer made every later message in that conversation fail — with no field named in the error, surfaced as `Stream error: 400`, while the answer was still on screen. History is now **clamped, never rejected**, and the per-turn bound is *derived* from the output ceiling rather than chosen so the two cannot drift. The 50-entry array cap moved into the same sliding window: it was also a rejection, and carried the identical defect one level up — a conversation reaching turn 51 would have failed from then on. Clamping is the existing pattern for context here (persona and document context are truncated, not refused).

- **`language.ts`** documented a guard that did not exist, and a direct call really does interpolate the raw code next to `undefined`. But it is not reachable: there is exactly one runtime entry point (`handler.ts` `chatRequestSchema.safeParse`) and all three call sites read `body.response_language` off the parsed output. So the guard is added and **labelled as insurance rather than as a fix**, and membership/naming are now one data structure with `isSupportedLanguage` as the single runtime gate — which also removed the duplicate `Set` the schema was building.

**On the should-fix items**, two are done and two I would push back on:

- **`message` raised 2 000 → 8 000 with a client-side counterpart.** You were right that a server-only bound is an opaque 400. 2 000 chars is ~300 words, which refuses a pasted review or support thread — normal input for an analytics tool. The mirrored constants are in `api/streamLimits.ts` and pinned by a lockstep test that reads the schema source, since the packages cannot import from each other.
- **`selected_personas` now clamped client-side.** Worth flagging that this cap is reachable in *one keystroke*, not twenty clicks: `@all` expands to every persona, and `persona_importer` appends with `put_item` and no prior delete (unlike generation, which replaces and is capped at 7). The cap itself is worth keeping — roundtable is one Bedrock call per persona against a 300s timeout — so `@all` is clamped to it.
- **The 8 dropped locales are not frontend-reachable.** `i18n/config.ts` pins `supportedLngs` to the same eight with `nonExplicitSupportedLngs: false` and detects from `localStorage` only, so `i18n.language` is always one of them, membership-identical to the new allowlist. Only third-party API clients change behaviour. Still a silent change, now stated in the description rather than only in a table.
- **`context` at 500 is fine, and the premise there is wrong.** `ChatFilters` is three *single-valued* optional strings plus a boolean, and `buildChatContext` emits at most four short clauses. There is no multi-select and no free-text search in that string, so a real user cannot approach 500.

**Also corrected two claims of my own** that the review's last point was right to poke at: the attachment array bound cannot bind (5 × 2 800 000 exceeds the API Gateway request-payload limit, so the transport refuses first), and the "dominant cost vectors" framing was wrong about which vectors dominate — `MAX_TOOL_LOOPS = 15` over `MAX_CANDIDATES = 10000`, resent each round because Converse is stateless, dwarfs anything a human types. Request-side caps are a well-formedness control here, not a cost control. Bounding the loop is separate work and probably where the money is.

Nits taken: the injection test now drives a real unrecognised code through the schema instead of passing `undefined`; type assertions dropped; the `getLanguageInstruction` re-export indirection removed; the stale `.catch(undefined)` comment fixed.

**Verified:** stream `tsc` 0, `eslint` 0 errors, **300 tests / 22 files** (was 285 / 21). Frontend `tsc` 0, `eslint` 0 with `--max-warnings 0`, **2161 tests / 143 files**. i18n reports `missing: 0` and flags nothing in `chat.json`; the gate still exits 1 on the pre-existing `prioritization.docType.prfaq` false positive in 7 locales, which this branch does not touch. The contiguity guard and the lockstep test were both mutation-probed — each fails exactly the assertion that should care.

**Two residues, both deliberate rather than overlooked:**

1. **An explicit @-mention selection is still unclamped**, so 21 hand-picked personas would still be refused. Clamping there would silently discard choices someone made deliberately, which is worse than an error — `@all` is different precisely because it is a shorthand over a list the user does not control. The asymmetry is stated in `resolvePersonaSelection`.
2. **When `@all` does clamp, the UI does not say so.** `MentionContextBar` returns null for the `@all` case, so surfacing it needs new props and a render path for a condition that takes repeated persona imports to reach. `clampedFrom` is already returned so the notice can be added without re-deriving the condition.

**And one thing worth flagging since I removed a validation bound:** `history[].content` no longer has a hard `.max()`. What bounds it is the API Gateway request-payload limit, and clamping means strictly *less* now reaches Bedrock in the pathological case than before. Authentication is untouched, and the locale allowlist is now enforced by a single tested gate rather than a duplicated `Set`.

The last commit is a pure extraction with no behaviour change: the `@all` clamp lived inside a `useCallback`, which is why it arrived untested — the only way to reach it was rendering the whole tab. `resolvePersonaSelection` is now a pure function with its own tests, and the regex and the "at least two voices" rule moved out of the component with it.

Not deployed — no environment was touched.

perrozzi:

Since my last pass the author added the `isSupportedLanguage` type guard plus the missing name-lookup guard in `getLanguageInstruction`, replaced the per-entry `history` rejection with a new `history-budget.ts` sliding-window clamp (derived from `MAX_OUTPUT_TOKENS`), raised `message` to 8 000 with a client-side mirror (`streamLimits.ts` + lockstep test), a `composerState` submit rule, `@all` clamping in project chat, and `messageTooLong` i18n keys in all eight catalogues. Both blockers are addressed and most should-fix/nit items are resolved; the dropped-locale regression remains.

### Resolved
- `language.ts` now has the real `if (name === undefined) return ''` guard, a `NAME_LOOKUP`-backed `isSupportedLanguage` type guard, and a corrected docstring — blocker 1 closed, with tests.
- `history[].content` no longer rejects: `clampHistoryToBudget` truncates/windows instead, with the "long assistant answer then follow-up" regression test in `schema.test.ts` — blocker 2 closed.
- Client-side cap for `message` with i18n keys in all 8 locales, plus `@all` fan-out clamped to `MAX_SELECTED_PERSONAS`.
- `as SupportedLanguage` / `as 'de'` assertions removed from `voc-context.test.ts` and `language.test.ts`; `if (lang === 'en') continue` replaced by a filter.
- `voc-context.test.ts` unrecognised-locale test now drives through `chatRequestSchema.parse` and asserts the raw string is absent.
- `persona-prompt.ts` re-export indirection dropped; `project-context.ts` imports from `./language.js`.
- Attachment envelope is now honestly documented as a non-binding sanity bound rather than claimed as the cost ceiling.

### Still open
- The eight locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) are still silently dropped to English — now pinned by a test asserting rejection, so this is a deliberate user-facing behaviour change that belongs in release notes.
- `schema.test.ts` still hard-codes the locale list instead of importing `SUPPORTED_LANGUAGES`.
- `voc-context.ts` still has the double blank line left by the removed function.

### New findings

**Should fix**

- `context` (500 chars) is still reject-only, and it contradicts the new "two policies" doc in `schema.ts`: it is not user-typed — `buildChatContext()` in `Chat.tsx` serialises the filter selections — so under the PR's own rule it is code-authored text and should clamp, or at minimum be mirrored in `streamLimits.ts`/validated client-side. A multi-select of sources + categories + sentiments still yields an opaque 400 with no i18n message.
- `history` now has no outer array bound at all (`.max(50)` removed in favour of the in-transform window). Keep a generous rejecting bound (e.g. 500 entries) alongside the clamp so an absurd array isn't fully Zod-validated element-by-element before being sliced to 50 — the graceful behaviour you want is preserved well below it.
- `MAX_SELECTED_DOCUMENTS` in `frontend/src/api/streamLimits.ts` is exported and lockstep-tested but never consumed; the document-selection path can still exceed 20 and 400. Either wire it up like `resolvePersonaSelection` or drop it.
- `frontend/src/api/streamLimits.lockstep.test.ts` reads `../lambda/stream/src/schema.ts` at module load. That couples the frontend test run to the presence of a sibling package — any CI job, container build, or sparse checkout that runs frontend tests alone will fail at import with ENOENT rather than a meaningful message. Consider `existsSync` + `it.skipIf`, or move the pin to a repo-root test.

**Nits**

- `history-budget.ts` and `handler.ts` both reference `DEFAULT_MAX_OUTPUT_TOKENS`, which does not exist — the constant is `MAX_OUTPUT_TOKENS`, and it lives in `history-budget.ts`, not "on the Converse wrapper" as the docstring says.
- `bedrock/converse-stream.ts` now carries two stray blank lines where the constant was removed.
- `Chat.tsx`: `MessageTooLongNotice` uses `role="status"` for an error condition; `role="alert"` is the right one alongside `aria-invalid`.
- `Chat.tsx` comment "a TDZ error that tsc and eslint both accept" reads as the opposite of what is meant.

---

perrozzi:
Round 2 addressed in `ee15ee6`, pushed right after this. Three of the four should-fix items taken as written, one answered differently, all four nits taken — and the first two nits are drift I introduced myself.

**Taken as written:**

- **`history` had no outer array bound.** Correct, and I had flagged the same hole in my own note without closing it. `MAX_HISTORY_ARRAY` (10× the window) now rejects, so Zod cannot be made to validate 50 000 elements before the window discards all but 50. Set an order of magnitude above the product limit deliberately: no real conversation meets it, so every genuine request still gets the graceful clamp rather than an error.
- **`MAX_SELECTED_DOCUMENTS` had no consumer.** Removed rather than wired. Documents are only ever chosen one @-mention at a time, and per the asymmetry this PR already documents an *explicit* selection is not clamped — silently dropping documents someone picked by hand is worse than the error. So a mirror would have no caller. The absence is now explained where the constant used to be.
- **The lockstep test's coupling to the sibling package.** Now diagnosed instead of surfacing as a bare ENOENT. I did *not* use `it.skipIf`, though: a skipped lockstep is a silent lockstep, which is precisely the failure the test exists to prevent. An unreachable schema stays a failure — just one that explains what it needs.

**Answered differently — `context`:**

You were right that it contradicted the doc, and the contradiction was in the doc. It had two categories and `context` is in neither. It *is* code-authored, but it is bounded by construction: `buildChatContext()` serialises at most four short clauses from three **single-valued** filters. Clamping would be the wrong policy for that — exceeding 500 would mean our own code changed shape, and truncating would hide it while rejecting surfaces it. So there are now three named policies, plus a statement of what would move `context` into the first one and require a client mirror.

The supporting claim doesn't hold, though: `ChatFilters` (`store/chatStore.ts`) is three single-valued optional strings and a boolean. There is no multi-select and no free-text search in that string, so a user cannot approach 500.

**Nits — the first two are my own:**

- Two docstrings still said `DEFAULT_MAX_OUTPUT_TOKENS` "on the Converse wrapper". Good catch: that constant does not exist. It is `MAX_OUTPUT_TOKENS` and it lives in `history-budget.ts`. I reversed the ownership direction mid-implementation — the pure module owns the number so nothing needs an SDK mock to read it — and left the comments describing the version I'd abandoned. That is the same class of defect as the docstring-promising-a-guard this PR set out to fix, which is worth saying out loud.
- Stray blank lines in `converse-stream.ts` and `voc-context.ts`, both mine.
- `role="alert"` rather than `role="status"` — agreed, it reports an error and sits next to `aria-invalid`.
- The TDZ comment said the gates "accept" the code when it meant neither flags it. Reworded.
- `schema.test.ts` now drives the allowlist test from `SUPPORTED_LANGUAGES`.

**Not taken:** the eight locales stay dropped. That is a product call rather than a defect, and the right place for it is release notes — noted in the description rather than reversed here.

Verified: stream `tsc` 0, `eslint` 0 errors, **301 tests / 22 files**. Frontend `tsc` 0, `eslint` 0 at `--max-warnings 0`, **2170 tests / 144 files**. Still not deployed.

perrozzi:

Since my last pass the author cleaned up the remaining items rather than changing behaviour: `MAX_SELECTED_DOCUMENTS` and its lockstep assertion are gone, the lockstep test now fails with an explanatory message instead of a bare ENOENT, an outer `MAX_HISTORY_ARRAY` (500) rejecting bound was added to `history`, `schema.test.ts` now drives the locale loop from `SUPPORTED_LANGUAGES`, the stray blank lines and the `DEFAULT_MAX_OUTPUT_TOKENS` / `role="status"` / TDZ-comment nits are fixed, and `context`'s reject-only policy is now argued explicitly in the `schema.ts` header. Everything from my previous list is addressed except the dropped-locale note.

### Resolved
- `history` regains an outer rejecting bound (`MAX_HISTORY_ARRAY`, 10× the window) alongside the clamp, with a test asserting the `history` path.
- Unused `MAX_SELECTED_DOCUMENTS` removed from `frontend/src/api/streamLimits.ts` and its lockstep assertion, with the asymmetry documented.
- `streamLimits.lockstep.test.ts` now `existsSync`-checks and throws a message naming the path and cwd instead of an opaque ENOENT.
- `schema.test.ts` imports `SUPPORTED_LANGUAGES` instead of hard-coding the eight codes.
- `context`'s reject-only policy is now stated as a third, explicitly-reasoned policy in `schema.ts` rather than silently contradicting the two-policy doc.
- Nits closed: `MAX_OUTPUT_TOKENS` naming in `history-budget.ts`/`handler.ts`, blank lines in `converse-stream.ts` and `voc-context.ts`, `role="alert"` in `Chat.tsx`, TDZ comment wording.

### Still open
- The eight dropped locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English is a user-facing behaviour change that belongs in release notes.

### New findings

**Nits**

- `voc-datalake/lambda/stream/src/schema.ts`: the `context` policy now rests entirely on the claim that `buildChatContext()` can never exceed 500 chars ("at most four short clauses from three single-valued filters"). Nothing enforces that claim — a future multi-select in `ChatFilters` turns it into an untranslated 400. A small test in `Chat` land asserting `buildChatContext()` output length for the worst-case filter set would pin the invariant the comment relies on.
- `voc-datalake/frontend/src/api/streamLimits.lockstep.test.ts`: throwing at module load is a defensible choice, but it does mean the frontend test suite cannot run without the `lambda/stream` package present. Fine if intentional — just worth a line in the frontend README/CI notes.

---

perrozzi:
Round 3 was nit-only, so this is the last code push — stopping the loop per the usual convention rather than churning further. One of the two nits was worth taking, though, and it was the better catch of the pair.

**Taken — `7313464`.** You were right that the whole `context` policy now rested on an unenforced claim, and that is the weakest kind of argument: a comment nobody would notice going false. `buildChatContext` moves into its own module and gains tests asserting the worst case the current filter model can produce — every filter set, each value far longer than any real category or source name — plus the structural property the bound actually depends on, one clause per filter.

Mutation-probed by simulating a multi-select on a single filter: it fails 3 of 5, including the length bound itself. So a future multi-select fails loudly and points at the decision to revisit (mirror `context` client-side with an i18n message, or switch it to clamping).

**Not taken — the frontend suite needing the sibling package present.** That is the deliberate choice from the previous round: a skipped lockstep is a silent lockstep, which is the failure the test exists to prevent. The error already names the path, the working directory and the reason it needs both packages, so a README line would only restate what the failure says. Happy to be overruled if you would rather have the note.

**Still open and unchanged: the eight dropped locales.** Agreed that it belongs in release notes — it is a product call, not a defect, so I have left it stated in the description rather than reversing it.

Verified: stream `tsc` 0 / `eslint` 0 errors / **301 tests / 22 files**. Frontend `tsc` 0 / `eslint` 0 at `--max-warnings 0` / **2175 tests / 145 files**. Six commits, `MERGEABLE`/`CLEAN`, and **not deployed** — no environment was touched at any point, so a human should merge and deploy rather than treating this as verified live.

perrozzi:

Since my last review the only change is the extraction I asked for: `buildChatContext` moved out of `Chat.tsx` into `pages/Chat/chatContext.ts` with `chatContext.test.ts` pinning the length invariant that the `context` reject-only policy rests on. That closes my remaining nit; the lockstep-test-throws-at-load note was accepted as intentional.

### Resolved
- `context`'s "bounded by construction" claim is now enforced by `chatContext.test.ts` (worst-case length + one-clause-per-filter assertions), and the extraction also documents why it can't live in the `.tsx`.

### Still open
- The eight dropped locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English still belongs in release notes.

### New findings

**Nits**

- `voc-datalake/frontend/src/pages/Chat/Chat.tsx`: removing the function left two consecutive blank lines where it used to be (between `SidebarSection` and `MessageTooLongNotice`) — same stray-blank-line nit as the earlier pass.
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: the worst-case test uses a 120-char `LONG_VALUE` labelled "implausibly long", but `category` values come from the tenant-configured category list (see `getConfiguredCategories` in `voc-context.ts`), i.e. they are user-influenced, not code-bounded. A configured category of ~450 chars would blow the 500 cap and produce exactly the untranslated 400 the file's header warns about. Either state the assumed per-value bound explicitly (and, ideally, enforce it where categories are configured) or truncate each clause in `buildChatContext`.
- `voc-datalake/frontend/src/pages/Chat/Chat.tsx`: worth double-checking the `ChatFiltersType` type import is still referenced now that `buildChatContext` is gone (lint should catch it if not).

---

perrozzi:
I said the last push would be the last one, and I'm breaking that for a specific reason: round 4 was nit-only, but the `chatContext` nit **refuted the correctness argument I made in the previous commit**, which makes it not cosmetic. Fixed in `8aabd07`. This one really is the last code push.

**You were right, and the mistake is worth naming precisely.** The `context` reject-only policy rested on the values being bounded. They aren't. I checked: `category` comes from the tenant's configured list, and `settings_handler.py` validates its length nowhere — it only counts the list to build a message. So a ~450-char configured category would have pushed `context` past 500 and produced exactly the untranslated 400 my own file header warned about.

The deeper problem is that **"bounded by construction" was a statement about data wearing the costume of a property of the code** — and the test I added last round encoded the same assumption, by treating 120 chars as a worst case. A test that inherits the bug it is meant to catch is worse than no test.

So the bound is now structural rather than assumed: `buildChatContext` caps each filter value, giving a worst case of 25 + 130 + 132 + 133 = 420 < 500 for values of *any* length. The test asserts that against 10 000 chars, and a second test pins that realistic values are untouched — truncation should be invisible in practice, since it only exists for absurd values. Truncating rather than rejecting is the right trade here: these clauses are hints for the model, and a shortened hint beats an error the reader cannot interpret.

The schema header no longer claims the values are short by nature. It names the client-side cap as load-bearing and states what would move `context` into the reject-with-a-client-mirror policy.

Also fixed: two blank lines where the function used to be — mine, and the third instance of that in this PR. The `ChatFiltersType` suggestion was a false alarm; the type still has two consumers in `Chat.tsx`, which is why lint stayed clean.

**Where this leaves the PR.** Seven commits. Stream `tsc` 0 / `eslint` 0 errors / **301 tests / 22 files**. Frontend `tsc` 0 / `eslint` 0 at `--max-warnings 0` / **2176 tests / 145 files**. `MERGEABLE`/`CLEAN`, `reviewDecision` empty and no CI configured, so a human is the only gate.

Two things a reviewer should carry rather than assume:

1. **Nothing here has been deployed.** No environment was touched at any point, so please treat every claim above as test-and-static-analysis evidence, not live verification.
2. **The eight dropped locales remain a product call for release notes**, unchanged across all four rounds.

perrozzi:

Since my last review the only change is the `context` bound: `buildChatContext` now truncates each filter value at `MAX_FILTER_VALUE_LENGTH = 120` via a `clause()` helper, the worst-case test was strengthened to a 10 000-char value, and the schema header was rewritten to say the bound comes from the code rather than from an assumption about category names. That addresses the category-length nit from the previous pass, and the stray blank lines in `Chat.tsx` are gone.

### Resolved
- `voc-datalake/frontend/src/pages/Chat/chatContext.ts`: tenant-configured `category` can no longer push `context` past the server's 500 — per-value cap makes the 420-char worst case provable (arithmetic in the header checks out: 25 + 130 + 132 + 133).
- `voc-datalake/frontend/src/pages/Chat/Chat.tsx`: double blank line removed.

### Still open
- The eight dropped locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English still belongs in release notes.

### New findings

**Nits**

- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: in `leaves realistic filter values untouched`, `expect(context).not.toContain('...')` is vacuous — `clause()` truncates with a bare `slice()` and never appends an ellipsis, so that assertion passes for truncated values too. The boundary that actually matters (121 chars → cut to exactly 120, 120 chars → untouched) is untested; assert that instead.
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: `SERVER_CONTEXT_CAP = 500` is a hand-copied comment-only mirror, while this PR already establishes a lockstep pattern for exactly this (`streamLimits.lockstep.test.ts` regex-reads schema source). `MAX_CONTEXT_LENGTH` would be readable by that same helper — worth pinning, since the whole 420 < 500 argument rests on the server number not moving.

---

perrozzi:
Both round-5 nits taken in `dbbbd29`, and both were about the guard I added last round — so this finishes that commit rather than extending scope. **Stopping the loop here regardless of what round 6 says, unless it contains something blocking.**

**The vacuous assertion is the one worth dwelling on, because it's mine and it's embarrassing in the right way.** `not.toContain('...')` was asserting the absence of an ellipsis that `clause()` never appends — it cuts with a bare `slice`. So it passed for truncated values too: an unfailable assertion, sitting in the file whose entire purpose is to make a bound fail loudly. Replaced with the boundary that discriminates — a value *at* the cap passes through whole, a value one character over is cut to exactly the cap — and mutation-probed with an off-by-one (`slice(0, MAX + 1)`), which the new test catches and the old one demonstrably would not have.

**The lockstep point lands for the same reason.** This PR argues in two other places that duplicated cross-package constants rot silently, and builds a lockstep test for exactly that — so hand-copying `500` into a comment-only mirror was inconsistent with my own argument, and worse than usual here because the "provably under 500" proof *is* arithmetic against that number. `MAX_CHAT_CONTEXT_LENGTH` now sits in `streamLimits.ts` with the other mirrors, pinned by the same regex helper.

**Where it lands.** Eight commits. Stream `tsc` 0 / `eslint` 0 errors / **301 tests / 22 files**. Frontend `tsc` 0 / `eslint` 0 at `--max-warnings 0` / **2179 tests / 145 files**. `MERGEABLE`/`CLEAN`.

**Two things that do not change and that a reviewer should carry forward:**

1. **Nothing has been deployed** at any point. Everything above is test and static-analysis evidence, not live verification, so the deploy and the live check on chat are still owed by whoever merges.
2. **The eight dropped locales** stay a product call for release notes — unchanged across all five rounds.

Thanks for the five passes. The two that mattered most were both cases where a "nit" contained a checkable fact that overturned an argument I had made: the tenant-configured `category` length, and this vacuous assertion.

perrozzi:

Since the last pass the only changes are the two nits I raised: `chatContext.test.ts` swaps the vacuous `not.toContain('...')` assertion for real boundary tests (at-cap passes whole, cap+1 cut to exactly the cap), and `MAX_CHAT_CONTEXT_LENGTH` is now mirrored in `streamLimits.ts` and pinned by `streamLimits.lockstep.test.ts` against the schema's `MAX_CONTEXT_LENGTH`. Both are addressed cleanly; the regex-based lockstep works on the non-exported `const` since it reads source text.

### Resolved
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: boundary at 120/121 chars is now asserted instead of the ellipsis non-assertion.
- `voc-datalake/frontend/src/api/streamLimits.ts` + `streamLimits.lockstep.test.ts`: the 500 is no longer a hand-copied comment; it fails loudly if the server cap moves.

### Still open
- The eight dropped locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English still belongs in release notes.

### New findings

**Nits**

- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: three stray blank lines left between the imports and `ABSURD_VALUE` after the `SERVER_CONTEXT_CAP` removal — likely a lint error under the frontend config.
- `voc-datalake/frontend/src/api/streamLimits.ts`: `MAX_CHAT_CONTEXT_LENGTH` has no production consumer (tests only), which sits awkwardly next to the `MAX_SELECTED_PERSONAS` doc comment arguing "an unused mirror would be a constant with no consumer". The doc block does explain why it exists, so this is a wording inconsistency, not a defect.
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: `expect(context).not.toContain(overCap)` is redundant after the exact `toBe` on the same line above.

---

perrozzi:
Round 6's three nits are cosmetic, so `627e4bc` clears them and **this closes the review loop** — I won't be re-polling after this push. Any further automated rounds are for whoever merges to weigh.

- **Blank lines: real, and mine** — four of them, left by `sed` edits between the imports and `ABSURD_VALUE`. One correction though: eslint does not flag consecutive blank lines under this config, so the "likely a lint error" guess is wrong; lint was clean before and after. The untidiness was real regardless.
- **Redundant `not.toContain`** after an exact `toBe` on the same value: dropped.
- **`MAX_CHAT_CONTEXT_LENGTH` having only test consumers** — agreed it read as if it contradicted the `MAX_SELECTED_PERSONAS` doc block right below it. Reworded so the distinction is explicit: a document mirror would have *no* caller, whereas this one is the number both the lockstep test and `buildChatContext`'s bound are checked against. The constant stays.

**Final state.** Nine commits. Stream `tsc` 0 / `eslint` 0 errors / **301 tests / 22 files**. Frontend `tsc` 0 / `eslint` 0 at `--max-warnings 0` / **2179 tests / 145 files**. `MERGEABLE`/`CLEAN`, `reviewDecision` empty and no CI in this repo, so a human is the only gate.

**Two items explicitly handed over, not resolved:**

1. **Nothing was deployed.** No environment was touched at any point in this PR, so the numbers above are test and static-analysis evidence only. A deploy plus a live multi-turn chat check — long answer, then a follow-up — is still owed, and that is precisely the path that was broken before this change.
2. **The eight dropped locales** (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English remains a product call for release notes, unchanged across all six rounds.

Worth recording which findings actually moved the code, since it was not the ones flagged loudest: the two most valuable were both "nits" whose premises overturned an argument I had made — the tenant-configured `category` length destroying my "bounded by construction" claim, and the vacuous ellipsis assertion in the very test written to enforce it. The two original blockers were closed in the first pass; everything of substance after that came from checking claims rather than from severity labels.

perrozzi:

Since the last pass only the two nits were touched: `chatContext.test.ts` drops the three stray blank lines and the redundant `not.toContain(overCap)` assertion, and the `MAX_SELECTED_PERSONAS` doc comment in `streamLimits.ts` is reworded so it no longer contradicts the existence of `MAX_CHAT_CONTEXT_LENGTH`. Both are clean; no behaviour changed.

### Resolved
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: stray blank lines removed and the redundant assertion folded into the single `toBe`.
- `voc-datalake/frontend/src/api/streamLimits.ts`: doc comment now explains why the context mirror has a consumer (lockstep + `buildChatContext` bound) while a document mirror would not.

### Still open
- The eight dropped locales (`it`, `nl`, `ru`, `ar`, `hi`, `sv`, `pl`, `tr`) silently becoming English still belongs in release notes.

### New findings
**Nits**
- `voc-datalake/frontend/src/pages/Chat/chatContext.test.ts`: the import block now runs straight into the `/** Pathological... */` doc block with no blank line separating them — cosmetic, worth a single newline for readability.

---

perrozzi:
`eda706d` restores `it, nl, ru, ar, hi, sv, pl, tr` — the item flagged in every one of the six rounds and resolved in none of them.

It went unresolved because as framed it had no resolution: "make it safe" had been implemented as "support fewer languages", so coverage and safety looked like a trade-off. **They aren't related.** Safety comes from interpolating a name out of `LANGUAGE_NAMES` and never the caller's own string — and that holds whether the table has eight rows or sixteen. The size of the list is a pure coverage decision, so dropping those eight bought nothing: the model answers in all of them, and the only effect was silently forcing English on any API client that asked for one. Invisible in the app, because the switcher only offers the eight translated locales — which is precisely why nobody would ever have reported it.

The list is now explicitly *not* the UI-catalogue list, and the header says which is which: the frontend's `i18n/languages.ts` decides what the **interface** is translated into; this decides what the **model** may be told to reply in. Conflating those two was the whole defect.

**The data didn't move alone**, which is worth flagging for review:

- `language.test.ts` asserted the opposite. The test literally named *"rejects locales with no shipped catalogue"* is inverted to assert those eight are **accepted**, so the regression can't be reintroduced by someone trimming the list back to match the UI — and a new test pins the *direction* of the difference (superset, not equal) rather than just the contents. Codes we genuinely don't name are covered separately.
- The hardcoded `toHaveLength(7)` is now derived from the list.
- `schema.ts`'s header no longer calls the constraint "the eight locales that have a shipped UI catalogue".

**Known limitation, now stated rather than implicit:** region and script subtags (`pt-BR`, `zh-Hans`) are still rejected. Supporting them properly means canonicalising with `Intl.getCanonicalLocales` + `Intl.DisplayNames` rather than widening a hand table — full coverage, nothing to maintain, and the interpolated name comes from ICU instead of the caller. That is a bigger change than this one, and it needs the Lambda runtime's ICU build checked first, so I've left it as the obvious next step rather than folding it in.

Verified: `tsc` 0, `eslint` 0 errors, **303 tests / 22 files** (was 301). Mutation-probed by trimming the list back to eight: typecheck fails with two `TS2353` errors and two tests fail. Worth noting precisely — the *runtime* guard survives that trim, because `isSupportedLanguage` derives from the names table rather than the array, which is the deliberate "cannot accept a language it cannot name" property. So typecheck is the enforcement here and the tests pin the intent.

Still not deployed.

perrozzi:

Since the last pass the only change is scope of the language allowlist: `SUPPORTED_LANGUAGES` grows from the eight UI-catalogue locales to sixteen, restoring `it, nl, ru, ar, hi, sv, pl, tr`, with `LANGUAGE_NAMES` extended in lockstep (still keyed by `SupportedLanguage`, so an unnamed code is a compile error), the module/schema doc comments rewritten to separate "UI is translated into" from "model may answer in", and `language.test.ts` inverted accordingly. Safety is unchanged — the prompt still interpolates a name from the table, never the caller's string — and the previously-flagged release-note gap disappears because nothing is dropped any more.

### Resolved
- `voc-datalake/lambda/stream/src/context/language.ts`: the eight silently-Englished locales are restored, so no behaviour change needs documenting in release notes.

### New findings
**Nits**
- `voc-datalake/lambda/stream/src/context/language.test.ts`: the "is a superset of the shipped UI locales" test hardcodes both the eight-locale list and `toBeGreaterThan(8)`; if a ninth UI catalogue ships, the assertion silently stops pinning the direction of the difference. A comment saying "bump both when a catalogue is added" would be enough.
- The PR description still states the eight codes are dropped and that unknown-locale clients "will now silently receive English" for them, plus the supported-set table lists only eight — worth updating before merge so the description matches the code.

Otherwise the delta is clean.

---

Issue #314, opened 2026-08-10
[Feature]: QR for a linked form, and expand the prototype, on Prioritization


perrozzi:
## Summary
Prioritization is where a group presents each proposal and argues about its score. Two things are
missing for that to work in a room. Getting the room's ratings in means pasting a link into a chat
channel, when a QR on screen would do. And the prototype under discussion renders in a pane a few
hundred pixels tall, when the whole point of the meeting is to look at it.

Both are small, because the pieces exist: the form already has a public mobile-ready page and the row
already renders the prototype in a frame. This adds a QR rendering of the first and an enlargement of
the second.

## Current state
- `FormCard`'s `EmbedCodeSection` computes the form's public page URL and offers it as a link, a
  copyable URL and a copyable iframe snippet. There is no QR, so putting a form in front of a room
  means someone reads a URL aloud or pastes it into a channel.
- `PRFAQRow`'s `PrototypePanel` renders the prototype inside `HtmlPrototypeFrame`, sized to the row.
  There is no way to enlarge it, so a pitch either squints at it or navigates away to the project.
- The Project Detail page already solves the second half: it offers "Open in new tab" and
  "Download .html" as plain anchors, with the signed link's deadline shown next to them.

## Proposed solution
1. **One `FormQrCode` component, owned by `pages/FeedbackForms/`**, rendering the form's existing
   public URL as an inline SVG QR. Render it in `EmbedCodeSection` beside the URL it encodes, and reuse
   the same component on the prioritization row for a linked form. The URL construction stays in one
   place.
2. **Behind an affordance, not inline in the cell.** A QR needs roughly 200px before it scans from a few
   metres, and a pitch looks at one artifact at a time, so a QR per row on a list is noise on a page
   that is already N+1 (U18). Keep the shipped `N submissions · ★3.2` as the row's resting state and
   open the QR in a popover or modal at a size that actually scans.
3. **An expand affordance on the row's prototype.** Primary is an in-page full-screen overlay reusing
   `HtmlPrototypeFrame`, which inherits the existing deadline handling. Secondary is "open in a new
   tab", copied from `DocumentsTab`'s **plain anchor** pattern.
4. **Port the signed-URL refresh to Prioritization.** See the risk below — this is the only part with
   real substance.

## Risk
**The QR adds no exposure, and that is worth stating precisely because it looks like it should.** It
encodes the form's existing public page: a static URL with no signature, no token and no expiry, which
[...]

PR #311, opened 2026-08-10
feat(feedback-forms): show a linked form's collected ratings on the Prioritization row

perrozzi:
## Summary

Feedback forms collected evidence about a proposal and the Prioritization page decided what to do about it, with nothing connecting them: `feedback_form_handler.py` had no notion of a project, and `Prioritization.tsx` never called a feedback-form endpoint. A reviewer set Impact / Time to Market / Strategic Fit / Confidence with the document and the prototype in view, and no access to the ratings collected about that same document.

A feedback form can now optionally record **which project — and, as a refinement, which PRD or PR/FAQ — it validates**, and expanding that document's row on Prioritization shows the linked form's submission count and average rating next to the sliders.

- **Backend** (`lambda/api/feedback_form_handler.py`): `project_id` / `document_id` added to `DEFAULT_FORM_CONFIG`, `UPDATABLE_FIELDS` and `item_to_form` — all three, since a field in only one of them silently fails to persist or to read back.
- **Public widget route gets its own projection.** `get_form_config_by_id` served `item_to_form(item)`, the same shape the authenticated routes use, on an **unauthenticated** route fetched cross-origin from customers' own sites. It now returns a new, deliberately narrower `item_to_widget_config` — so the next field added to `item_to_form` cannot leak by default either. Routing and authorization are untouched.
- **Frontend types**: the fields go on `FeedbackForm` only, never on the shared `FeedbackFormFields` base that `FeedbackFormConfig` (the widget shape) extends. A test asserts that boundary over the declaration text.
- **Editor**: a new "Validates" tab (`ValidationLinkPicker.tsx`) picks the project and document and clears either. Only scorable document types (`prd`, `prfaq`) are offered; a link to a document that no longer exists is preserved rather than silently cleared on Save.
- **Prioritization**: `formLinkUtils.ts` matches a row to its forms — `project_id` first, `document_id` as a refinement, so a **regenerated document** (which mints a new id) still shows its project's evidence, while a form pinned to a *live* sibling document does not bleed onto other rows. `LinkedFormEvidence.tsx` renders the stats.
- Eight locale catalogues updated, keys inserted in sorted position (no whole-file re-serialization), no `{{count}}` interpolation on new keys.
- Mock-server fixtures gained the link so this is exercisable with `npm run mock`.

### Decisions made where the task was open

[...]

PR #315, opened 2026-08-10
feat(feedback-forms): show a form's public page as a scannable QR

perrozzi:
The QR half of #314. The prototype-expand half of that issue is **not** here, so this references the issue rather than closing it.

## Why

Getting a room's ratings into the product means someone reads a URL aloud or pastes it into a chat channel. The form already has a public, mobile-ready page — `GET /feedback-forms/{id}/iframe` returns a complete HTML document with a mobile viewport — so all that was missing was a scannable rendering of an address the card already prints.

Nothing new is exposed. The QR encodes the same static, unauthenticated URL the card already offers as "Direct Link" and as an `<iframe>` snippet. It changes how that URL reaches a phone, not who can reach it.

## What changed

**One construction site for the address.** `FormCard` used to build `${apiEndpoint}/feedback-forms/${form.form_id}/iframe` inline. That moved to `api/feedbackFormUrls.ts` as `feedbackFormPublicUrl()`, and both render sites call it. A QR is the worst place for a second spelling to drift: nothing on screen reveals which address was encoded, so a stale one is discovered only by a room full of phones failing to load it. The endpoint is normalised through `baseUrl.ts`'s existing `stripTrailingSlashes`, so the QR resolves the endpoint identically to every API request.

**`components/FormQrCode/`** takes the endpoint and the form id, never a ready-made URL, which makes a second construction site structurally impossible. It renders `QRCodeSVG` at 200px with a 4-module quiet zone at level `M`:

- 200px because a phone a few metres from a projected screen needs roughly that before it resolves the modules. A QR too small to scan from a seat is decoration.
- The quiet zone because the spec requires 4 modules and the library defaults to 0; without it a scanner cannot find the symbol's edges against a busy card.
- Level `M` because the realistic failure mode is glare and heads in the way, not a pristine scan.

**On the Prioritization row it sits behind a button, not inline.** One QR per row would be noise on a page that already reads N+1 projects, and a pitch looks at one artifact at a time. The row's resting state — the submission count and average from #311 — is unchanged, and the QR opens in the existing `ModalShell` with a dialog role, an accessible name, and Escape/overlay/Close dismissal. No request is made: the form object is already in hand, so the QR costs nothing extra.

**SVG, generated in the browser.** Crisp when projected, where a canvas at this size is not. No third-party QR image service — that would ship form URLs to an external endpoint for no benefit and fail in a room with poor connectivity — and no hand-rolled encoder, because Reed–Solomon error correction is not a place to be clever.

## Dependency

One, `qrcode.react`, pinned to exactly `4.2.0` rather than a range. Its peer range accepts React 19. Not added to `manualChunks`: it is small and reached only from two already-loaded pages.

## Tested

From `voc-datalake/frontend`:

| Gate | Result |
|---|---|
| `npm run lint` | 0 errors, 0 warnings (`--max-warnings 0`) |
| `npm run typecheck` | clean |
| `vitest run` | 152 files, **2240 tests**, 0 failed |

New keys are in all eight locale catalogues with real translations. The i18n gate's only failures are pre-existing and unrelated (`prioritization.docType.prfaq` and some `categories.json` plural extras); `Total missing: 0 | empty: 0`.

**Revert-sensitivity was checked rather than assumed.** Eleven mutations at the narrowest point of each new guard, eleven named test failures, tree clean afterwards. The one that matters most: changing `/iframe` to `/config` in the builder fails four tests across both files.

## Two judgement calls worth a reviewer's eye

**The QR is withheld when a form's stats read fails.** A deleted form presents exactly that way on the row, and its public page is gone too, so showing a QR would send a room to a 404. The cost is that a transient stats error also hides it. I took the 404 as the worse outcome.

**A third `isError`-testing sibling in `LinkedFormStats` tripped ESLint's `complexity: 12`.** The restructure into a single branch was required rather than cosmetic, and it moves no rendered output.

## Not in scope

The prototype-expand half of #314, and with it the signed-URL refresh scheduler that Prioritization still lacks. Also no QR pointing at a prototype (its address is a short-lived signed credential, so a QR over it is broken by construction), no "create a form for this document" affordance, and no change to the public routes, the widget, the feedback corpus, or the prioritization score write path.

### DIFF
--- a/voc-datalake/frontend/package-lock.json
+++ b/voc-datalake/frontend/package-lock.json
@@ -17,6 +17,7 @@
         "i18next-browser-languagedetector": "^8.2.1",
         "i18next-http-backend": "^3.0.2",
         "lucide-react": "^0.554.0",
+        "qrcode.react": "4.2.0",
         "react": "^19.2.0",
         "react-dom": "^19.2.0",
         "react-i18next": "^16.5.8",
@@ -11479,6 +11480,15 @@
       "license": "MIT",
       "engines": {
         "node": ">=6"
+      }
+    },
+    "node_modules/qrcode.react": {
+      "version": "4.2.0",
+      "resolved": "https://registry.npmjs.org/qrcode.react/-/qrcode.react-4.2.0.tgz",
+      "integrity": "sha512-QpgqWi8rD9DsS9EP3z7BT+5lY5SFhsqGjpgW5DY/i3mK4M9DTBNz3ErMi8BWYEfI3L0d8GIbGmcdFAS1uIRGjA==",
+      "license": "ISC",
+      "peerDependencies": {
+        "react": "^16.8.0 || ^17.0.0 || ^18.0.0 || ^19.0.0"
       }
     },
     "node_modules/quick-temp": {
--- a/voc-datalake/frontend/package.json
+++ b/voc-datalake/frontend/package.json
@@ -31,6 +31,7 @@
     "i18next-browser-languagedetector": "^8.2.1",
     "i18next-http-backend": "^3.0.2",
     "lucide-react": "^0.554.0",
+    "qrcode.react": "4.2.0",
     "react": "^19.2.0",
     "react-dom": "^19.2.0",
     "react-i18next": "^16.5.8",
--- a/voc-datalake/frontend/public/locales/de/components.json
+++ b/voc-datalake/frontend/public/locales/de/components.json
@@ -112,7 +112,8 @@
   },
   "formQrCode": {
     "accessibleName": "QR-Code für das Feedback-Formular {{formName}}",
-    "caption": "Scannen, um dieses Formular auf dem Smartphone zu öffnen"
+    "caption": "Scannen, um dieses Formular auf dem Smartphone zu öffnen",
+    "unavailable": "QR-Code nicht verfügbar, bis der API-Endpunkt in den Einstellungen konfiguriert ist"
   },
   "personaExport": {
     "copied": "Kopiert!",
--- a/voc-datalake/frontend/public/locales/de/components.json
+++ b/voc-datalake/frontend/public/locales/de/components.json
@@ -110,6 +110,10 @@
     "viewDetails": "Details anzeigen",
     "webScraper": "Web-Scraper"
   },
+  "formQrCode": {
+    "accessibleName": "QR-Code für das Feedback-Formular {{formName}}",
+    "caption": "Scannen, um dieses Formular auf dem Smartphone zu öffnen"
+  },
   "personaExport": {
     "copied": "Kopiert!",
     "copyMarkdown": "Als Markdown kopieren",
--- a/voc-datalake/frontend/public/locales/de/prioritization.json
+++ b/voc-datalake/frontend/public/locales/de/prioritization.json
@@ -46,6 +46,11 @@
     "medium": "Mittlere Priorität",
     "none": "Nicht bewertet"
   },
+  "qr": {
+    "close": "Schließen",
+    "show": "QR-Code anzeigen",
+    "title": "QR-Code"
+  },
   "scores": {
     "confidence": "Konfidenz",
     "confidenceDescription": "Sicherheit bei Auswirkungs- und MeZ-Schätzungen",
--- a/voc-datalake/frontend/public/locales/en/components.json
+++ b/voc-datalake/frontend/public/locales/en/components.json
@@ -112,7 +112,8 @@
   },
   "formQrCode": {
     "accessibleName": "QR code for the {{formName}} feedback form",
-    "caption": "Scan to open this form on a phone"
+    "caption": "Scan to open this form on a phone",
+    "unavailable": "QR code unavailable until the API endpoint is configured in Settings"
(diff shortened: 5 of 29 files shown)


perrozzi:

This is a frontend-only change that extracts the feedback-form public URL into a single builder (`api/feedbackFormUrls.ts`), adds a `FormQrCode` component rendering that URL as an inline SVG QR via a new pinned `qrcode.react@4.2.0` dependency, and wires it into `FormCard`'s embed section and a modal on the Prioritization row. No backend, IAM, Bedrock, or ingestor surface is touched, so none of those repo rules apply. The change is well-tested, all eight locales get both new key groups, and the URL-builder consolidation is the right call. I found no blockers, only a couple of boundary-validation and store-access points worth tightening.

### Should fix

- **`FormQrCode` will silently encode a relative URL if the endpoint is empty** (`src/components/FormQrCode/FormQrCode.tsx`, `src/api/feedbackFormUrls.ts`). If `config.apiEndpoint` is `''`/undefined at render time, `feedbackFormPublicUrl` yields `/feedback-forms/form_1/iframe` and the QR renders happily — a scannable symbol that resolves to nothing. That is precisely the invisible-failure mode the module docstring argues against, so it deserves a guard rather than a comment: either return early (render nothing / a fallback) when the endpoint is not an absolute URL, or hide `LinkedFormQrButton` in that case. A unit test on `feedbackFormUrls` for `''` and for a non-absolute endpoint would lock it in.
- **Zustand access in `LinkedFormQrButton`** (`src/pages/Prioritization/LinkedFormEvidence.tsx`): `const { config } = useConfigStore()` subscribes the component to the whole store; the repo's pattern is a narrow selector (e.g. `useConfigStore((s) => s.config.apiEndpoint)`). Also note the inconsistency with `FormCard`, which receives `apiEndpoint` as a prop — passing it down from the row would keep `LinkedFormStats`/`LinkedFormQrButton` pure and testable without the store.
- **URL-encode the path segment**: `feedbackFormPublicUrl` interpolates `formId` raw. It is server-minted today, but this is a system boundary that now feeds both a copied embed snippet and a QR; `encodeURIComponent(formId)` costs nothing and matches the "validate at boundaries" rule.

### Nits

- `FormQrCode.test.tsx` captures `value`, `size`, and `marginSize` but not `level`; since the `M` choice is documented as deliberate, assert it too — otherwise a change to `L` passes silently.
- The moved iframe snippet still does `title="${form.name}"` without escaping (`FormCard.tsx`). Pre-existing, but a form name containing `"` produces a broken snippet for the customer; worth escaping now that the builder is being touched.
- `EmbedCodeSection` now takes the whole `FeedbackForm` where it needs only `form_id` and `name`; narrower props (`formId`, `formName`) would keep the component's contract as tight as `FormQrCode`'s.
- The new file headers carry ~25 lines of rationale for a one-line function (`feedbackFormUrls.ts`); a couple of sentences plus the existing tests would carry the same intent with less to keep in sync.
- The QR modal has two dismissal affordances (explicit Close button plus `ModalShell`'s own close/overlay/Escape); fine, but check it doesn't duplicate a button `ModalShell` already renders.

---

perrozzi:

Since the last review the author added an addressability guard to `feedbackFormPublicUrl` (returns `null` for empty, relative, and non-`http(s)` endpoints, plus `encodeURIComponent` on the id), gave `FormQrCode` and `EmbedCodeSection` explicit "no address" renderings with a new `unavailable` key in all eight locales, threaded `apiEndpoint` down from `Prioritization` → `PRFAQRow` → `LinkedFormEvidence` instead of touching the store, escaped the iframe `title` attribute, narrowed `EmbedCodeSection`'s props, and asserted `level: 'M'`. All previous Should-fix items and most nits are addressed; the incremental diff looks clean.

### Resolved
- Empty/relative endpoint no longer yields a scannable relative URL — `feedbackFormUrls.ts` returns `null`, both render sites say so in words, with unit and component tests.
- Zustand subscription removed from `LinkedFormQrButton`; endpoint now arrives as a prop from the page, matching `FormCard`.
- `formId` is percent-encoded, with traversal/`?` cases covered in `feedbackFormUrls.test.ts`.
- `level: 'M'` is now asserted in `FormQrCode.test.tsx`.
- Iframe `title` is escaped (`&` before `"`) with a regression test in `FormCard.test.tsx`.
- `EmbedCodeSection` takes `formId`/`formName` rather than the whole record; the duplicate-Close question is answered in a comment.

### New findings
- **Verify `feedbackForms:configureApiFirst` exists in all eight catalogues** (`src/pages/FeedbackForms/FormCard.tsx`). No locale diff adds it, so it must be pre-existing — but the test (`expect(screen.getByText(i18n.t('feedbackForms:configureApiFirst')))`) passes vacuously if the key is missing, since `t` returns the key and the component renders that same string. Worth a one-time grep rather than trusting the assertion.

### Nits
- The long-docstring nit from last round remains (`feedbackFormUrls.ts` header is now ~35 lines for a two-branch function); take it or leave it.
- `escapeAttributeValue` covers `&` and `"`, which is exactly what a double-quoted attribute needs — fine as-is, just note it isn't a general HTML escaper if reused.

---

PR #316, opened 2026-08-10
feat: QR from the form card, and open a row's prototype in a new tab

perrozzi:
Two follow-ups to #315, both toward #314. Referenced rather than closing it — see "What remains" below.

## A — the QR on the Feedback Forms card was effectively invisible

#315 put a QR on the form card, but inside `EmbedCodeSection`, which only mounts after clicking **"Show Embed Code"** — a developer-facing disclosure about iframe snippets. A facilitator who wants a QR for a room has no reason to open it. Meanwhile the Prioritization row had a proper "Show QR code" button opening a dialog.

Two surfaces now want identical behaviour, so the row's trigger became a shared component instead of being copied: **`components/FormQrCode/FormQrButton.tsx`**. Both pages use it, so the dialog semantics — `role="dialog"`, accessible name from the visible heading, Escape and overlay dismissal, `useId` so several on one page cannot collide — exist once. Same move, and the same reason, as `PrototypeRenderer` and `parsePrototypeSpec` being lifted into `components/` for this very page.

The card now offers "Show QR code" directly and the inline QR is gone from the embed section, so there is **exactly one QR affordance per card** rather than a second entry point to keep in step. The embed section keeps its Direct Link, copy buttons and iframe snippet untouched.

Keys moved with it: `prioritization.qr.*` → `components.formQrCode.*` across all eight catalogues, values carried over verbatim. `prioritization.qr` is gone entirely, so no orphan is left behind and the Feedback Forms page no longer reads a `prioritization`-namespaced string.

## B — open a row's prototype in a new tab

The embedded frame is 384px inside half a row: enough to recognise a prototype, not enough to walk a room through one. The row now offers it in a new tab, reusing the Project Detail implementation rather than growing a second one:

- **`components/PrototypeLinkActions.tsx`** — the plain `target="_blank"` anchor, an optional `download` anchor, and the lifetime note. Two exports rather than one component, because the two pages place them differently: the Documents tab puts the note beside "Live preview" with the actions on the right, while the narrow Prioritization column stacks the note under its link.
- **`components/usePrototypeLinkRefresh.ts`** — `useProjectData`'s pre-expiry refetch, now shared.

Prioritization gets opening only, not download. A reviewer scanning a pitch is not filing artifacts, and that is the project page's job.

### Why the refresh hook is the substance of this PR, not a detail

A prototype is served from a key-group-restricted CloudFront behaviour, so its URL is a **signed, session-scoped credential** that the API re-mints on every project read — not a durable address. Project Detail copes because `useProjectData` schedules a refetch *ahead of* expiry. **Prioritization had no such scheduling**, and until now it did not need it: a stale URL only fed an iframe that had already loaded.

An anchor cannot get away with that. A click navigates immediately, so nothing can fetch a replacement first — which means shipping the anchor without the scheduler would be a **latent 403** for anyone who parks a pitch on a second monitor past the signature's hour. It works in review, where nothing sits on screen that long, and fails in the meeting the feature exists for.

So the page now schedules its own re-signing over every document it fetches. The hook is keyed on `earliestPrototypeExpiry` — a number or null — rather than on the documents array, so the timer re-arms when the deadline actually moves and survives the identity churn of a refetch that changed nothing.

### These are anchors, and must stay anchors

`components/prototypeLinkLifetime.ts` documents why, and it is worth repeating where a future reviewer will see it: the tempting "improvement" is a button that fetches a fresh signature and then calls `window.open`. That trades a 403 for a **popup blocker**, and `download` cannot be triggered that way at all. Freshness is the scheduler's job; there is deliberately no click handler on these.

`Prioritization.prototypeLink.test.tsx` asserts `getByRole('link')` precisely so that rewrite fails a test instead of shipping.

## Tested

From `voc-datalake/frontend`:

| Gate | Baseline @ `a110d62` | This branch |
|---|---|---|
| `npm run lint` | 0 | 0 |
| `npm run typecheck` | 0 | 0 |
| `vitest run` | 2249 / 152 files | **2267 / 154 files** |

`i18n:check` is byte-identical to baseline — the same pre-existing failures (`prioritization.docType.prfaq`, `categories.json` plural extras in ja/zh/ko), and none of the new keys appear as orphans.

**18 mutations, 18 named test failures**, each applied at the narrowest point of a guard and then restored. The one that matters most: deleting the `usePrototypeLinkRefresh` call from Prioritization fails *"re-reads the projects before the prototype signature expires"* and *"re-arms off the replacement URL so renewal repeats"*, while the three "schedules nothing" cases correctly stay green.

## Two things a reviewer should know

**A test was too weak and the mutation sweep is what found it.** The "no open affordance" assertions used `queryByRole('link')` alone. `<a href="">` **loses the `link` role** in aria-query, so a regression that rendered "Open in new tab" pointing nowhere would have passed. Now asserted by visible text as well.

**`useProjectData`'s effect changed behaviour, in the safe direction.** Keying on the deadline instead of array identity means fewer redundant timer rebuilds with identical scheduling. Its existing tests, including the re-arm case, pass unmodified.

## What remains on #314

The in-page **full-screen enlargement**. This PR delivers the new-tab open; #314's other acceptance criterion asks for filling the viewport and dismissing back to the row. Worth confirming whether opening in another page satisfies the intent before that one is built.

Also unchanged and still out of scope: no QR pointing at a prototype (its address is a short-lived signed credential, so a QR over it is broken by construction), no "create a form for this document", no present mode, and no backend, CDK, public-route, widget, corpus or prioritization-score changes.


perrozzi:

Two frontend refactors: the Prioritization row's QR trigger is lifted into a shared `components/FormQrCode/FormQrButton` and surfaced directly on the Feedback Forms card (with `prioritization.qr.*` → `components.formQrCode.*` across all eight locales), and the Documents tab's prototype open/download anchors plus their expiry note are lifted into `components/PrototypeLinkActions` with the pre-expiry re-sign effect extracted into `components/usePrototypeLinkRefresh` and adopted by Prioritization. The reasoning about keeping these as anchors and scheduling re-signing ahead of expiry is sound, the i18n move is complete in all locales, and the new tests pin the important invariants (role=link, visible expiry text, non-dangling `aria-describedby`, "schedules nothing" cases). No blockers found.

### Should fix

- **Duplicate user-visible strings now live in two catalogues.** `components/PrototypeLinkActions.tsx` renders `components:prototypeLink.openNewTab` / `downloadHtml`, while `pages/ProjectDetail/DocumentsTab.tsx`'s `LegacyHtmlActions` still uses `projectDetail:documents.prototype.openNewTab` / `downloadHtml` (both keys are retained in `public/locales/*/projectDetail.json`). Two catalogue entries for the same button label is exactly the drift the rest of this PR removes — point `LegacyHtmlActions` at the shared `components:prototypeLink.*` keys and delete the `projectDetail` duplicates (which also lets the `defaultValue` fallbacks go).
- **`FormQrButton` is only reachable by deep path.** Both consumers import `'../../components/FormQrCode/FormQrButton'` while `FormQrCode` itself is imported as `'../../components/FormQrCode'` (barrel). Export `FormQrButton` from that barrel so the folder has one public entry point; also worth double-checking resolution, since `FormQrButton.tsx` does `import FormQrCode from './FormQrCode'` — that only works if `components/FormQrCode/` is a directory containing `FormQrCode.tsx` plus an index, and the diff doesn't show that move.

### Nits

- `pages/FeedbackForms/FormCard.tsx`: the line you touched still hardcodes `'Show Embed Code'` / `'Hide Embed Code'` in English, immediately beside the newly-translated QR trigger. Cheap to key while you're in there.
- `signedUrlExpiresAt(url)` is now parsed twice per prototype (once in `PrototypeLinkLifetimeNote`, once in `PrototypeLinkActions` to decide `describedBy`). Harmless, but the caller already has the URL — a single `hasLifetimeNote` boolean threaded from the caller would keep the two in provable lockstep instead of relying on "same predicate, same helper".
- `pages/ProjectDetail/useProjectData.ts`: dropping `id` from the effect deps means a timer armed for project A survives a navigation to B whenever the two deadlines happen to be numerically equal (the ref makes it invalidate B's key). Practically harmless — an extra refetch — but the previous deps made it impossible.
- `pages/Prioritization/Prioritization.tsx`: `allDocuments` is only reduced to a number inside the hook, so the `useMemo` buys nothing; and the invalidation refetches the whole fan-out (N `getProject` calls) off the single earliest deadline. Fine at current scale, worth a comment if project counts grow.
- `FormQrButton`'s trigger could carry `aria-haspopup="dialog"` so the control announces that it opens a dialog rather than performing the action inline.

---

perrozzi:

Since the last review the author removed the `projectDetail` `openNewTab` duplicates and pointed `LegacyHtmlActions` at `components:prototypeLink.*`, added `aria-haspopup="dialog"` to the trigger, documented why the `FormQrCode` barrel can't re-export `FormQrButton` (lint `--max-warnings 0` + `react-refresh/only-export-components`), keyed `usePrototypeLinkRefresh` on a new required `refreshScope` with a dedicated test, translated the embed toggle, and dropped the pointless `useMemo` in Prioritization. Both previous should-fix items are addressed (one by change, one by justification), and the nits I raised are all either fixed or answered in comments.

### Resolved
- Duplicate button labels: `LegacyHtmlActions` now reads `components:prototypeLink.*` and the `projectDetail` `openNewTab` copies are deleted in all eight catalogues.
- `FormQrButton` deep import: explained as a hard lint constraint rather than a style choice — acceptable.
- `useProjectData` timer scoping: `id` is back in the effect key via `refreshScope`, with a regression test (`usePrototypeLinkRefresh.test.ts`) that fails if the key is removed; the equal-deadline scenario it exercises is correct (floored delay makes inheritance observable).
- `FormCard` embed toggle now translated; `allDocuments` `useMemo` removed; `aria-haspopup="dialog"` added.

### New findings
- **`documents.prototype.downloadHtml` is now an orphan in `public/locales/*/projectDetail.json`.** Its twin `openNewTab` was removed from all eight files, but `downloadHtml` sits earlier in the object (outside the shown hunks) and was left behind even though `LegacyHtmlActions` no longer reads it. Delete it from all eight, same as `openNewTab`.
- Nit: `usePrototypeLinkRefresh`'s `refreshScope` is documented as "Required … for correctness" but typed `string | undefined`, so two distinct scopes that both resolve to `undefined` collapse into one key — the exact hole the parameter exists to close. Tighten to `string` (callers can pass a literal fallback) if no caller legitimately has no scope.
- Nit: `FormCard.tsx` now depends on `card.showEmbedCode` / `card.hideEmbedCode`; worth confirming those exist in all seven non-`en` `feedbackForms.json` (the diff touches none of them, so a missing key would silently render the raw key path in those locales).

---

perrozzi:

The only change since my last review is an expanded JSDoc on `usePrototypeLinkRefresh`'s `refreshScope`, arguing why `string | undefined` is deliberate rather than tightening the type. No code, test or catalogue changes.

### Resolved
- `refreshScope` typing nit: the new comment makes the case adequately — deps are per hook instance, and the one caller that can pass `undefined` (`useProjectData`, gated on `id != null`) has no documents and therefore no deadline in that state, so two scopes cannot silently collapse. Fine as documentation-only.

### Still open
- `documents.prototype.downloadHtml` remains an orphan in all eight `public/locales/*/projectDetail.json` after `LegacyHtmlActions` moved to `components:prototypeLink.*`; delete it alongside the already-removed `openNewTab`.
- Unconfirmed that `card.showEmbedCode` / `card.hideEmbedCode` exist in the seven non-`en` `feedbackForms.json` now that `FormCard.tsx` reads them.

### New findings
None — the comment is accurate and the only load-bearing part of its argument (`useProjectData`'s `id != null` gate) matches the code.

---

perrozzi:
Round 2 findings, checked against the tree rather than the diff. One is a false positive, one is fixed, one is fine.

**`documents.prototype.downloadHtml` is not an orphan — it still has a reader.** `DocumentsTab.tsx:394` uses it in the raw-output pane, the branch that renders when a prototype's spec fails to parse:

```tsx
<button onClick={onDownload} className="text-blue-600 hover:underline">
  {t('documents.prototype.downloadHtml', { defaultValue: 'Download' })}
</button>
```

`LegacyHtmlActions` no longer reads it, which is what the diff shows, but that was not its only call site. Deleting it from the eight catalogues would make that button render a raw key path.

There *is* a real defect there, just not this one: the pane downloads an `application/json` blob named `.json` under a key whose value reads "Download .html", and its `defaultValue: 'Download'` is dead because the key exists. Pointing it at the already-present `downloadJson` is a two-character fix that also frees `downloadHtml` for deletion and closes the duplication completely — but it changes a user-visible string outside this PR's scope, so it is filed separately rather than smuggled in here.

**`refreshScope` — the wording was wrong, not the type.** Fixed in f7f74d0, as documentation.

`undefined` is admitted deliberately: a detail page's id legitimately is undefined, since `useProjectData` gates its query on `id != null`. And it cannot hide a collision, because the deps array is per hook instance — two scopes can only collapse *within one caller*, and there `undefined` means no project, therefore no documents, therefore no deadline, and `refreshDelayMs` returns null with nothing armed. Tightening to `string` would force callers to invent a sentinel that collapses in exactly the same way while reading as though it does not.

So the parameter stays positional and unconditional (forgetting it is the bug the regression test pins), and the doc now states the guarantee — a change re-arms — instead of claiming a requirement the signature does not express.

**`card.showEmbedCode` / `card.hideEmbedCode` are present in all eight.** Verified by parsing each catalogue: real translations in de/en/es/fr/ja/ko/pt/zh, e.g. `Einbettungscode anzeigen`, `埋め込みコードを表示`. The diff touches none of those files precisely because the keys already existed and had simply never been referenced — that is where two of the unused-key improvements came from.

perrozzi:

Since my last review the QR itself grew: `QR_SIZE_PX` 200 → 384, the SVG became `w-full h-auto`, the dialog panel widened to `max-w-md`, and a new test pins the fluid sizing. No previously-flagged item was touched.

### Still open
- `documents.prototype.downloadHtml` remains an orphan in all eight `public/locales/*/projectDetail.json`.
- Still unconfirmed that `card.showEmbedCode` / `card.hideEmbedCode` exist in the seven non-`en` `feedbackForms.json`.

### New findings
- Nit (doc accuracy), `src/components/FormQrCode/FormQrCode.tsx`: the JSDoc calls 384 "the widest this is ever drawn" and "an upper bound", but with `w-full` inside `max-w-md` (448px) minus the dialog's `p-4` the rendered edge is ~416px, i.e. above the intrinsic size. Harmless in SVG, but the comment asserts something the CSS does not guarantee — either soften the wording or cap with `max-w-[384px]`.
- Nit, `src/components/FormQrCode/FormQrCode.test.tsx`: `size).toBeGreaterThanOrEqual(320)` leaves the actual value untested in either direction; asserting the constant (or a range) would fail a regression back to 200 *and* an accidental jump that outruns the panel.

Otherwise the change is clean — no i18n, IAM or type-guard surface touched.

---

PR #317, opened 2026-08-11
feat(documents): record and show how each document was built

perrozzi:
## What this does

This product builds documents out of other documents — a PRD, then a PR/FAQ, then a prototype from those,
sometimes a merge of several. Afterwards nothing could say what went into what. A document in the Documents
tab showed a type badge, a date, a title and a preview, and that was the whole story of how it came to exist.

Underneath it was messier than simply missing. The same relation — *this was built from those* — was already
stored **three different ways and absent in three more**:

| document type | what it recorded before |
|---|---|
| merge output | `source_documents` (a list of ids) + `merge_instructions` |
| prototype | `source_prd_id` + `source_prfaq_id` — fixed arity, so it cannot express "built from two PRDs" |
| PRD, PR/FAQ, research, product report | nothing at all |

And **none of those fields was declared anywhere in the frontend**, so all of them arrived on every project
read and were discarded at the type boundary.

Every path that creates a document now also writes one `derivation` record: an ordered list of
`(source document id, role)` entries, plus the inputs that are not documents — feedback items used, personas
used, and whether the project's product-context block was included. The Documents tab renders it.

## Two commits, deliberately

1. **`record and expose how every document was built`** — the data and the read model, plus the frontend
   contract. No UI.
2. **`show what each document was built from`** — the Documents tab, plus copy in all eight locale catalogues.

The contract fix that the UI exposed is folded into the first commit rather than appearing as a correction in
the second, so no commit in this branch publishes the contract in a shape its first consumer immediately has
to change.

## Two properties worth preserving

**It records what was USED, never what was requested.** The generator selects reference documents by id and
then feeds the model at most the first three. `sources` lists the ones that actually reached it, and
`selected_document_count` states how many were selected, so the tab can say "3 of 5 selected documents used".
That difference was previously invisible.

This makes the drop **observable, not fixed** — the three-document cap and its ordering are a known separate
issue and are deliberately untouched here. The display is correspondingly neutral: same grey as everything
around it, no warning colour, no icon, and silent when the two numbers agree. A test walks the DOM from that
line up to the panel asserting no alarm styling, so a future change in that direction fails rather than
slipping through.

**Counts and identifiers only, never copied content.** A test pins the exact key set.

## Additive, with no migration

Nothing is removed, renamed or migrated, and every pre-existing lineage field is still written. Documents
created before this change explain themselves through `resolveDerivation`, which reconstructs the same shape
from the three legacy forms and reports whether the answer came from the declared field, from legacy fields,
or is not recoverable. It is pure, total and depth-1 by construction, which is why a cyclic reference is
inert. A source document deleted since stays listed and marked unavailable rather than silently vanishing —
the relation outlived its target — but is not rendered as a control that leads nowhere. A document with no
recoverable provenance renders nothing at all, which is the honest answer for a hand-authored document.

## Deliberately out of scope

These are different relations and separate product decisions, and none of them is started here:

- **Succession** — "this document replaces that one". Regenerating a PRD still produces two unrelated PRDs.
  The prototype's existing `revised_from_id` / `revision_feedback` are left exactly as they are and are not
  folded into derivation. No version, ordinal or supersedes field is introduced.
- **Grouping** — whether a PRD and a PR/FAQ generated together are one proposal. No grouping key invented.
- **The three-document input cap**, per above. Reviewing this PR turned up that the cap is really three
  separate decisions with different answers — *which* documents get dropped is a bug (ascending `sk` order
  means `RESEARCH#` is always dropped first, and the oldest documents are always kept), *how much* gets
  through is a judgment call the repo currently answers four different ways, and two document types the
  picker offers are filtered out before the cap even applies. A concrete fix is proposed in the review
  thread and recommended as its own PR on top of this one, so that a change to what the model sees is
  revertable independently of the record that measures it.

## ⚠️ One thing to look at consciously

`voc-datalake/lib/test-support/baseline.json` carries a regenerated `VocProcessingStack.templateSha256`.
Provenance for a research report is decided in `step_initialize` and persisted by `step_save`, so it rides the
state machine the way `web_search_queries` already does — which legitimately changes that one template. It is
a one-line diff, and the readable name inventory is byte-identical across all five stacks, so no resource is
replaced. The resultSelector references `derivation` unconditionally, matching the existing
`web_search_queries` precedent; `step_save` tolerates its absence for executions in flight across a deploy.

Only one deploy-order skew is possible here, and it is the one covered. A new state-machine definition cannot
run against not-yet-updated Lambda code: the definition `GetAtt`s the function, so CloudFormation always
updates the Lambda first. The reachable skew is the opposite one — an execution already in flight keeps the
definition it started with — and that is exactly what `step_save`'s `.get()` default covers. One thing to know
if this is ever restructured: a JSONPath miss in a `resultSelector` raises `States.Runtime`, which a
`States.ALL` catch does **not** intercept, so a future return path omitting the key would fail outside
`handleError` and leave the job at `running` rather than `failed`. That is what
`test_always_returns_a_derivation_even_when_nothing_was_selected` protects, and why it is worth keeping.

## Verification

Run per leg rather than through the root aggregate, which chains with `&&` and hides later legs:

| leg | result |
|---|---|
| `pytest lambda/shared lambda/jobs lambda/research lambda/api` | 1092 passed |
| `vitest run` (frontend, full) | 2318 passed / 157 files |
| `vitest run src/pages/ProjectDetail src/api/derivation.test.ts` | 368 passed / 35 files |
| `vitest run lib` (CDK) | 231 passed / 17 files |
| `tsc -b` (frontend) | clean |
| `eslint src` (frontend) | clean |
| `ruff check` on both touched Python files | clean |

Those figures include three review rounds: +8 backend (a placeholder lockstep test, then four structural tests
enforcing invariants the review asked me to confirm, then two closing a fail-open in one of those very tests)
and +4 frontend (the sources-empty-but-selected case, in the resolver and in the pane). An earlier revision of
this section claimed
2299 / 156 for the full frontend suite; that was captured at the first commit and was wrong even there. The
true baselines are 2301 / 156 at `1b592e6` and 2314 / 157 at `2580e6d` — the second commit adds
`DocumentsTab.derivation.test.tsx`, which is the whole file delta. Measured in throwaway worktrees at each
commit rather than inferred.

All eight locale catalogues were checked programmatically for the same key set, case-insensitive sort order at
every level, and no empty values.

**Revert-sensitivity was demonstrated by mutation, not asserted.** Recording the requested document ids
instead of the used ones fails `test_records_the_three_documents_used_not_the_five_selected` and two others —
and the fixture deliberately selects five reference documents against a cap of three, because a fixture with
three or fewer could not tell the two cases apart. Forcing the resolver never to return a source's type fails
seven tests across both the resolver and the pane. Removing `selected_document_count` from the resolver's
emptiness check — the first review round's fix — fails exactly the four tests added for it and nothing else.

The second round's four tests are structural, reading production code out of its AST, so each was
positive-controlled rather than trusted: an early `return` omitting `derivation` in `step_initialize`, a nested
`def` inside it, a product-context section appended without its `###` heading, and a third `return` in
`build_product_context_block` each fail exactly the test that claims to watch for them, with the offending
line, function or count in the message. The nested-`def` probe also justified a test I nearly cut: it made the
first test report `assert 'derivation' in ['nope']`, attributing an inner return to the step — the `ast.walk`
leak its companion exists to catch, shown rather than argued.

**And one of those guards was itself fail-open, which the third round caught.** The heading check read only
`sections.append(...)`, so a section arriving by `extend`, `+=` or a comprehension would leave its view while it
kept reporting success — the same defect as the one it was written to rule out, one level up. Measured rather
than reasoned about: rewriting one append as `extend` leaves the heading test **green** while the replacement
fails and names the form. The fix asserts a closed set of ways `sections` may be built, so an unrecognised form
is a finding rather than a silent gap; five positive controls (`extend`, `+=`, a renamed producer, a nested
`def`, a second initialisation) all fire.

Every mutated file was restored byte-identically after every probe, verified by hash.

**Pre-existing and untouched:** `ruff check lambda plugins` sits at 548 findings, **delta 0** against the
previous commit — measured with the same command in a throwaway worktree rather than assumed. The three
findings in `research_step_handler.py` (import order, blind `except`, naive `datetime.now()`) are all older
than this branch; the only change to that file this round is a comment. `i18n:check` exits 1 byte-identically
to a baseline captured on the clean tree, with findings in
`categories.json` and `prioritization.json`; `typecheck:tests` is red at 77 errors in test files this diff
does not touch.

## Follow-ups this enables

The natural next step is the prototype build accepting explicit source document ids, which needs the
client-supplied id treated as untrusted input — it must be confirmed to exist, to belong to the project in the
path, and to be of the type the parameter names, and it must never silently fall back to newest-of-type.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "Dokument löschen",
+    "derivation": {
+      "builtFrom": "Erstellt aus",
+      "feedbackUsed_one": "{{count}} Feedback-Eintrag verwendet",
+      "feedbackUsed_other": "{{count}} Feedback-Einträge verwendet",
+      "personasUsed_one": "{{count}} Persona verwendet",
+      "personasUsed_other": "{{count}} Personas verwendet",
+      "productContext": "Produktkontext einbezogen",
+      "roles": {
+        "mergeInput": "Zusammengeführt",
+        "prototypePrd": "Prototyp-Quelle PRD",
+        "prototypePrfaq": "Prototyp-Quelle PR/FAQ",
+        "reference": "Referenz"
+      },
+      "selectedUsed": "{{used}} von {{selected}} ausgewählten Dokumenten verwendet",
+      "unavailable": "Nicht mehr verfügbar"
+    },
     "editDocument": "Dokument bearbeiten",
     "newDocument": "Neues Dokument",
     "noDocuments": "Keine Dokumente",
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "Delete document",
+    "derivation": {
+      "builtFrom": "Built from",
+      "feedbackUsed_one": "{{count}} feedback item used",
+      "feedbackUsed_other": "{{count}} feedback items used",
+      "personasUsed_one": "{{count}} persona used",
+      "personasUsed_other": "{{count}} personas used",
+      "productContext": "Product context included",
+      "roles": {
+        "mergeInput": "Merged in",
+        "prototypePrd": "Prototype source PRD",
+        "prototypePrfaq": "Prototype source PR/FAQ",
+        "reference": "Reference"
+      },
+      "selectedUsed": "{{used}} of {{selected}} selected documents used",
+      "unavailable": "No longer available"
+    },
     "editDocument": "Edit document",
     "newDocument": "New Document",
     "noDocuments": "No documents",
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "Eliminar documento",
+    "derivation": {
+      "builtFrom": "Creado a partir de",
+      "feedbackUsed_one": "{{count}} comentario utilizado",
+      "feedbackUsed_other": "{{count}} comentarios utilizados",
+      "personasUsed_one": "{{count}} persona utilizada",
+      "personasUsed_other": "{{count}} personas utilizadas",
+      "productContext": "Contexto del producto incluido",
+      "roles": {
+        "mergeInput": "Fusionado",
+        "prototypePrd": "PRD de origen del prototipo",
+        "prototypePrfaq": "PR/FAQ de origen del prototipo",
+        "reference": "Referencia"
+      },
+      "selectedUsed": "{{used}} de {{selected}} documentos seleccionados utilizados",
+      "unavailable": "Ya no está disponible"
+    },
     "editDocument": "Editar documento",
     "newDocument": "Nuevo Documento",
     "noDocuments": "Sin documentos",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "Supprimer le document",
+    "derivation": {
+      "builtFrom": "Créé à partir de",
+      "feedbackUsed_one": "{{count}} retour client utilisé",
+      "feedbackUsed_other": "{{count}} retours clients utilisés",
+      "personasUsed_one": "{{count}} persona utilisé",
+      "personasUsed_other": "{{count}} personas utilisés",
+      "productContext": "Contexte produit inclus",
+      "roles": {
+        "mergeInput": "Fusionné",
+        "prototypePrd": "PRD source du prototype",
+        "prototypePrfaq": "PR/FAQ source du prototype",
+        "reference": "Référence"
+      },
+      "selectedUsed": "{{used}} document(s) utilisé(s) sur {{selected}} sélectionné(s)",
+      "unavailable": "Plus disponible"
+    },
     "editDocument": "Modifier le document",
     "newDocument": "Nouveau document",
     "noDocuments": "Aucun document",
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "ドキュメントを削除",
+    "derivation": {
+      "builtFrom": "作成元",
+      "feedbackUsed_one": "フィードバック{{count}}件を使用",
+      "feedbackUsed_other": "フィードバック{{count}}件を使用",
+      "personasUsed_one": "ペルソナ{{count}}件を使用",
+      "personasUsed_other": "ペルソナ{{count}}件を使用",
+      "productContext": "製品コンテキストを含む",
+      "roles": {
+        "mergeInput": "マージ元",
+        "prototypePrd": "プロトタイプの元PRD",
+        "prototypePrfaq": "プロトタイプの元PR/FAQ",
+        "reference": "参照"
+      },
+      "selectedUsed": "選択された{{selected}}件のうち{{used}}件を使用",
+      "unavailable": "現在は利用できません"
+    },
     "editDocument": "ドキュメントを編集",
     "newDocument": "新規ドキュメント",
     "noDocuments": "ドキュメントなし",
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "문서 삭제",
+    "derivation": {
+      "builtFrom": "생성 출처",
+      "feedbackUsed_one": "피드백 {{count}}건 사용",
+      "feedbackUsed_other": "피드백 {{count}}건 사용",
+      "personasUsed_one": "페르소나 {{count}}개 사용",
+      "personasUsed_other": "페르소나 {{count}}개 사용",
+      "productContext": "제품 컨텍스트 포함",
+      "roles": {
+        "mergeInput": "병합 원본",
+        "prototypePrd": "프로토타입 원본 PRD",
+        "prototypePrfaq": "프로토타입 원본 PR/FAQ",
+        "reference": "참조"
+      },
+      "selectedUsed": "선택한 {{selected}}개 중 {{used}}개 사용",
+      "unavailable": "더 이상 사용할 수 없음"
+    },
     "editDocument": "문서 편집",
     "newDocument": "새 문서",
     "noDocuments": "문서 없음",
--- a/voc-datalake/frontend/public/locales/pt/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/pt/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "Deletar documento",
+    "derivation": {
+      "builtFrom": "Criado a partir de",
+      "feedbackUsed_one": "{{count}} comentário utilizado",
+      "feedbackUsed_other": "{{count}} comentários utilizados",
+      "personasUsed_one": "{{count}} persona utilizada",
+      "personasUsed_other": "{{count}} personas utilizadas",
+      "productContext": "Contexto do produto incluído",
+      "roles": {
+        "mergeInput": "Mesclado",
+        "prototypePrd": "PRD de origem do protótipo",
+        "prototypePrfaq": "PR/FAQ de origem do protótipo",
+        "reference": "Referência"
+      },
+      "selectedUsed": "{{used}} de {{selected}} documentos selecionados utilizados",
+      "unavailable": "Não está mais disponível"
+    },
     "editDocument": "Editar documento",
     "newDocument": "Novo Documento",
     "noDocuments": "Nenhum documento",
--- a/voc-datalake/frontend/public/locales/zh/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/zh/projectDetail.json
@@ -57,6 +57,22 @@
   },
   "documents": {
     "deleteDocument": "删除文档",
+    "derivation": {
+      "builtFrom": "构建来源",
+      "feedbackUsed_one": "已使用 {{count}} 条反馈",
+      "feedbackUsed_other": "已使用 {{count}} 条反馈",
+      "personasUsed_one": "已使用 {{count}} 个用户角色",
+      "personasUsed_other": "已使用 {{count}} 个用户角色",
+      "productContext": "已包含产品上下文",
+      "roles": {
+        "mergeInput": "合并来源",
+        "prototypePrd": "原型来源 PRD",
+        "prototypePrfaq": "原型来源 PR/FAQ",
+        "reference": "参考文档"
+      },
+      "selectedUsed": "已选择 {{selected}} 个文档，使用 {{used}} 个",
+      "unavailable": "已不可用"
+    },
     "editDocument": "编辑文档",
     "newDocument": "新建文档",
     "noDocuments": "无文档",
--- a/voc-datalake/frontend/src/api/derivation.test.ts
+++ b/voc-datalake/frontend/src/api/derivation.test.ts
@@ -68,6 +68,42 @@ describe('a document with a declared derivation', () => {
   })
 })
 
+describe('a declared derivation whose every selected document was dropped', () => {
+  // The generator records the selected count before it reads the documents, so
+  // a request whose selected documents had all been deleted stores "5 selected,
+  // 0 used". That record is the contract working, not an empty one.
+  const noneReached = {
+    document_id: 'prd_3',
+    derivation: {
+      sources: [],
+      selected_document_count: 5,
+      feedback_count: 0,
+      persona_ids: [],
+      product_context_included: false,
+    },
+  }
+
+  it('is not treated as having no derivation', () => {
+    expect(resolveDerivation(noneReached).origin).toBe('declared')
+  })
+
+  it('reports the count it recorded, so a consumer can say none of five was used', () => {
+    const resolved = resolveDerivation(noneReached)
+    expect(resolved.sources).toEqual([])
+    expect(resolved.selected_document_count).toBe(5)
+  })
+
+  it('does not fall back to the legacy fields of the same document', () => {
+    // The decisive case: the fallback would report one merge_input source and
+    // origin 'legacy', silently replacing "none of the five reached the model"
+    // with a lineage this document does not have.
+    const resolved = resolveDerivation({ ...noneReached, source_documents: ['legacy_1'] })
+    expect(resolved.origin).toBe('declared')
+    expect(resolved.sources).toEqual([])
+    expect(resolved.selected_document_count).toBe(5)
+  })
+})
+
 describe('resolving sources against the project documents', () => {
   const doc = {
     derivation: {
--- a/voc-datalake/frontend/src/api/derivation.test.ts
+++ b/voc-datalake/frontend/src/api/derivation.test.ts
@@ -1,0 +1,318 @@
+/**
+ * The derivation contract at the query boundary: a document always answers
+ * "what was I built from", a legacy document answers it in the same form, and no
+ * sparse or malformed record can break a consumer.
+ *
+ * Every expectation is a literal — nothing here is derived from the code under
+ * test.
+ */
+import { describe, it, expect } from 'vitest'
+import {
+  DERIVATION_ROLES,
+  emptyDerivation,
+  normalizeDerivation,
+  resolveDerivation,
+} from './derivation'
+
+const PRD = { document_id: 'prd_1', document_type: 'prd', title: 'Onboarding PRD' }
+const PRFAQ = { document_id: 'prfaq_1', document_type: 'prfaq', title: 'Onboarding PR/FAQ' }
+
+describe('the role vocabulary', () => {
+  it('is closed and lists the four relations the backend creates', () => {
+    expect(DERIVATION_ROLES).toEqual(['reference', 'prototype_prd', 'prototype_prfaq', 'merge_input'])
+  })
+})
+
+describe('a document with a declared derivation', () => {
+  const doc = {
+    document_id: 'prd_2',
+    derivation: {
+      sources: [
+        { document_id: 'doc_e', role: 'reference' },
+        { document_id: 'doc_d', role: 'reference' },
+      ],
+      selected_document_count: 5,
+      feedback_count: 12,
+      persona_ids: ['persona_1'],
+      product_context_included: true,
+    },
+  }
+
+  it('reports the sources in the order they were recorded', () => {
+    expect(resolveDerivation(doc).sources.map((s) => s.document_id)).toEqual(['doc_e', 'doc_d'])
+  })
+
+  it('reports the selected count separately, so the dropped documents are visible', () => {
+    const resolved = resolveDerivation(doc)
+    expect(resolved.sources).toHaveLength(2)
+    expect(resolved.selected_document_count).toBe(5)
+  })
+
+  it('reports the non-document inputs', () => {
+    const resolved = resolveDerivation(doc)
+    expect(resolved.feedback_count).toBe(12)
+    expect(resolved.persona_ids).toEqual(['persona_1'])
+    expect(resolved.product_context_included).toBe(true)
+  })
+
+  it('says the answer came from the declared field', () => {
+    expect(resolveDerivation(doc).origin).toBe('declared')
+  })
+
+  it('coerces counts that arrive as strings or DynamoDB floats', () => {
+    const resolved = resolveDerivation({
+      derivation: { ...doc.derivation, selected_document_count: '5', feedback_count: 12.0 },
+    })
+    expect(resolved.selected_document_count).toBe(5)
+    expect(resolved.feedback_count).toBe(12)
+  })
+})
+
+describe('resolving sources against the project documents', () => {
+  const doc = {
+    derivation: {
+      sources: [
+        { document_id: 'prd_1', role: 'prototype_prd' },
+        { document_id: 'deleted_1', role: 'prototype_prfaq' },
+      ],
+      selected_document_count: 2,
+      feedback_count: 0,
+      persona_ids: [],
+      product_context_included: false,
+    },
+  }
+
+  it('names a source that still exists', () => {
+    expect(resolveDerivation(doc, [PRD]).sources[0]).toEqual({
+      document_id: 'prd_1',
+      role: 'prototype_prd',
+      title: 'Onboarding PRD',
+      document_type: 'prd',
+      resolved: true,
+    })
+  })
+
+  it('keeps a source whose document no longer exists, marked unresolved', () => {
+    expect(resolveDerivation(doc, [PRD]).sources[1]).toEqual({
+      document_id: 'deleted_1',
+      role: 'prototype_prfaq',
+      title: null,
+      document_type: null,
+      resolved: false,
+    })
+  })
+
+  it('returns one entry per source even when nothing resolves', () => {
+    const resolved = resolveDerivation(doc)
+    expect(resolved.sources).toHaveLength(2)
+    expect(
+      resolved.sources.every((s) => s.resolved === false && s.title === null && s.document_type === null),
+    ).toBe(true)
+  })
+
+  it('reports what kind of document each source is, not only its title', () => {
+    // The role says how a source contributed; only document_type says what it
+    // is. A consumer that renders a type badge beside a title must get both
+    // from this one call rather than searching the document list again.
+    const resolved = resolveDerivation(doc, [PRD, PRFAQ])
+    expect(resolved.sources.map((s) => s.document_type)).toEqual(['prd', null])
+  })
+
+  it('reports an empty type for a source that resolved without one', () => {
+    // Mirrors title exactly: present-but-unusable is '', absent is null, so no
+    // consumer has to tell "resolved with no type" from "not resolved".
+    const untyped = { document_id: 'doc_x', title: 'No type on the wire' }
+    const resolved = resolveDerivation(
+      { derivation: { sources: [{ document_id: 'doc_x', role: 'reference' }] } },
+      [untyped],
+    )
+    expect(resolved.sources[0]).toEqual({
+      document_id: 'doc_x',
+      role: 'reference',
+      title: 'No type on the wire',
+      document_type: '',
+      resolved: true,
+    })
+  })
+})
+
+describe('a legacy document with no declared derivation', () => {
+  it('reads a prototype built from a PRD and a PR/FAQ', () => {
+    const resolved = resolveDerivation(
+      { document_id: 'prototype_1', source_prd_id: 'prd_1', source_prfaq_id: 'prfaq_1' },
+      [PRD, PRFAQ],
+    )
+    expect(resolved.sources).toEqual([
+      { document_id: 'prd_1', role: 'prototype_prd', title: 'Onboarding PRD', document_type: 'prd', resolved: true },
+      { document_id: 'prfaq_1', role: 'prototype_prfaq', title: 'Onboarding PR/FAQ', document_type: 'prfaq', resolved: true },
+    ])
+    expect(resolved.origin).toBe('legacy')
+  })
+
+  it('treats a real stored null exactly like an absent key', () => {
+    // A prototype built from a PR/FAQ alone stores source_prd_id: null; the
+    // sibling key is absent entirely. Both must read as "no such source".
+    const storedNull = resolveDerivation({ source_prd_id: null, source_prfaq_id: 'prfaq_1' })
+    const absent = resolveDerivation({ source_prfaq_id: 'prfaq_1' })
+    expect(storedNull.sources).toEqual([
+      { document_id: 'prfaq_1', role: 'prototype_prfaq', title: null, document_type: null, resolved: false },
+    ])
+    expect(storedNull).toEqual(absent)
+  })
+
+  it('reads a merge output built from a list of documents', () => {
+    const resolved = resolveDerivation({
+      source_documents: ['doc_1', 'doc_2'],
+      merge_instructions: 'Combine them',
+    })
+    expect(resolved.sources).toEqual([
+      { document_id: 'doc_1', role: 'merge_input', title: null, document_type: null, resolved: false },
+      { document_id: 'doc_2', role: 'merge_input', title: null, document_type: null, resolved: false },
+    ])
+    expect(resolved.origin).toBe('legacy')
+  })
+
+  it('reads a research report that only ever recorded a feedback count', () => {
+    const resolved = resolveDerivation({ document_type: 'research', feedback_count: 42 })
+    expect(resolved.feedback_count).toBe(42)
+    expect(resolved.sources).toEqual([])
+    expect(resolved.origin).toBe('legacy')
+  })
+
+  it('prefers the declared field when a document carries both', () => {
+    const resolved = resolveDerivation({
+      source_prd_id: 'prd_legacy',
+      derivation: {
+        sources: [{ document_id: 'prd_1', role: 'prototype_prd' }],
+        selected_document_count: 1,
+        feedback_count: 0,
+        persona_ids: [],
+        product_context_included: false,
+      },
+    })
+    expect(resolved.sources.map((s) => s.document_id)).toEqual(['prd_1'])
+    expect(resolved.origin).toBe('declared')
+  })
+})
+
+describe('a document with no recoverable lineage', () => {
+  it.each([
+    ['no lineage fields at all', { document_id: 'doc_1', title: 'Hand written' }],
+    ['a null derivation', { derivation: null }],
+    ['an empty declared derivation', { derivation: emptyDerivation() }],
+    ['an empty source list', { derivation: { ...emptyDerivation(), sources: [] } }],
+    ['a derivation that is not an object', { derivation: 'built from vibes' }],
+    ['an empty legacy list', { source_documents: [] }],
+    ['a null document', null],
+    ['a string instead of a document', 'not a document'],
+    ['an array instead of a document', []],
+  ])('reports exactly that, and does not throw: %s', (_case, input) => {
+    const resolved = resolveDerivation(input)
+    expect(resolved.origin).toBe('none')
+    expect(resolved).toEqual({ ...emptyDerivation(), sources: [], origin: 'none' })
+  })
+})
+
+describe('a malformed record', () => {
+  it('drops only the malformed entries, keeping the readable ones', () => {
+    const resolved = resolveDerivation({
+      derivation: {
+        sources: [
+          { document_id: 'doc_1', role: 'reference' },
+          { role: 'reference' }, // no id: points at nothing
+          { document_id: '', role: 'reference' }, // empty id: same
+          { document_id: 'doc_2', role: 'inspired_by' }, // role outside the vocabulary
+          'doc_3', // not an entry at all
+          null,
+          { document_id: 'doc_4', role: 'merge_input' },
+        ],
+        selected_document_count: 7,
+        feedback_count: 0,
+        persona_ids: [],
+        product_context_included: false,
+      },
+    })
+    expect(resolved.sources).toEqual([
+      { document_id: 'doc_1', role: 'reference', title: null, document_type: null, resolved: false },
+      { document_id: 'doc_4', role: 'merge_input', title: null, document_type: null, resolved: false },
+    ])
+    expect(resolved.selected_document_count).toBe(7)
+  })
+
+  it('degrades individual bad fields without losing the rest', () => {
+    const resolved = resolveDerivation({
+      derivation: {
+        sources: 'not a list',
+        selected_document_count: 'nonsense',
+        feedback_count: -3,
+        persona_ids: ['persona_1', 42, null, ''],
+        product_context_included: 'yes',
+      },
+    })
+    expect(resolved.sources).toEqual([])
+    expect(resolved.selected_document_count).toBe(0)
+    expect(resolved.feedback_count).toBe(0)
+    expect(resolved.persona_ids).toEqual(['persona_1'])
+    expect(resolved.product_context_included).toBe(false)
+  })
+
+  it('does not let a malformed sibling reject the surrounding documents', () => {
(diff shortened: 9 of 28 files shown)


perrozzi:

This PR introduces a single `derivation` record ("what was this document built from") written by every document-creating backend path (document generator, merger, research state machine, product report), a shared `lambda/shared/derivation.py` builder with a closed role vocabulary, a lenient read-side resolver in `frontend/src/api/derivation.ts` that also reconstructs the shape from the three legacy lineage forms, and a provenance footer in `DocumentsTab`. The work is genuinely additive, well tested on both sides, keeps content out of the record (ids/counts only), and the role vocabulary is pinned in lockstep across languages. Two logic issues stood out; the rest is clean.

### Blockers

- `frontend/src/api/derivation.ts` — `isEmpty()` ignores `selected_document_count`, so a declared derivation of `{sources: [], selected_document_count: 5, feedback_count: 0, persona_ids: [], product_context_included: false}` is treated as empty: `useDeclared` is false, the resolver falls through to the legacy path, and `origin` comes back `'none'`. `DocumentsTab.tsx` then renders nothing (`if (derivation.origin === 'none') return null`). That is reachable and is exactly the case the PR is built to expose: in `document_generator/handler.py::_gather_context`, `selected_document_count = len(selected_ids)` is set before the `if docs:` guard, so if every selected reference document has since been deleted (or the query returns none of them), the record says "5 selected, 0 used" and the UI says nothing at all. It also makes `origin` misreport: the document *did* declare a derivation and *can* say something about how it was built. Include `selected_document_count > 0` in `isEmpty()` (or decide emptiness before the legacy fallback), and add cases to `derivation.test.ts` / `DocumentsTab.derivation.test.tsx` for `sources: []` with a non-zero selected count — the current `it.each` block only covers `emptyDerivation()`.

### Should fix

- `lambda/jobs/document_generator/handler.py::_product_context` — the `product_context_included` flag is derived purely from `block != NO_PRODUCT_CONTEXT`. That literal was previously only produced by the exception fallback; the real `build_product_context_block(project_id)` presumably returns something else (`''` or a header-only block) for a project with no product context, in which case a document is recorded as having included product context when it did not. The three tests in `test_derivation.py::TestProductContextFlag` only cover the exact placeholder and the raising case, so the gap is untested. Suggest `included = bool(block and block.strip()) and block != NO_PRODUCT_CONTEXT`, plus a test for an empty/whitespace block.
- `lambda/research/research_step_handler.py::step_initialize` — `lib/stacks/processing-stack-consolidated.ts` now references `$.Payload.derivation` unconditionally, so *every* return path out of `step_initialize` must carry the key or the state fails outright. Only one return is visible in this diff; please confirm there is no early return (e.g. no feedback found / web search short-circuit) that would omit it. The `web_search_queries` precedent also means the deploy-order skew is pre-existing rather than new, but it is worth stating in the PR that a new state machine definition invoking not-yet-updated Lambda code will fail this state — `step_save`'s tolerance only covers the opposite direction.

### Nits

- `lambda/shared/test/test_derivation_roles_lockstep.py` extracts roles with `re.findall(r"'([^']+)'", ...)` over the whole array body, which includes the JSDoc comments between entries. A future comment containing an apostrophe ("the model's input") would inject a phantom role and fail with a confusing message. Stripping `/** … */` before the scan, or matching only `^\s*'([a-z_]+)',` lines, makes it robust.
- `research_step_handler.py` builds the initialize result with the literal key `'derivation'` while `step_save` uses `DERIVATION_FIELD`; use the constant in both (and consider exporting the state key name so `processing-stack-consolidated.ts` and the handler can't drift silently — the CDK test covers the wiring, not the spelling).
- `DocumentsTab.tsx::DerivationSourceRow` — the source link is a `<button>` without `type="button"`.
- `DocumentsTab.tsx` has two separate imports from `../../api/derivation` (value and type); merge them.
- `ja`/`ko`/`zh` catalogues carry `*_one` variants that i18next will never select for those locales; harmless, but they exist only to satisfy the same-key-set check.

---

perrozzi:
Round 1 disposition. **One finding was a real bug and is fixed. Two rest on premises that turn out to be false, and I've said which premise and how I checked.** The nits are done except two I'm declining with reasons. Then, at the end, the thing actually worth arguing about: the three-document cap.

Diff added by this round: 2 lines of production code (one condition, one attribute), 4 tests, 1 test-helper regex, 1 new lockstep test file.

---

## Fixed

### 1. `isEmpty()` ignoring `selected_document_count` — correct, and reachable exactly as described

Confirmed on all four points. `selected_document_count` was the one field of `DocumentDerivation` the emptiness check never consulted, so `{sources: [], selected_document_count: 5, …}` read as empty, fell through to the legacy path, and came back `origin: 'none'` — and `DocumentsTab` returns `null` on that. The reachability claim is right too: `selected_document_count = len(selected_ids)` is set at `_gather_context` L130, before the `if docs:` guard at L134, and `used_sources` is only appended inside it. A request whose every selected document had been deleted since therefore stored "5 selected, 0 used" and the tab said nothing at all — for the document with the most to explain.

`selected_document_count === 0` is now part of `isEmpty`. The legacy path is unaffected because `derivationFromLegacyFields` sets `selected_document_count: sources.length`, so an empty legacy reconstruction still reports `none`; both pre-existing `it.each` cases still pass unchanged.

Three resolver tests and one render test, as suggested. The third resolver test is the decisive one — a document carrying **both** the declared `{sources: [], selected_document_count: 5}` **and** a legacy `source_documents: ['legacy_1']`, which asserts the resolver reports the declared emptiness rather than substituting a lineage the document does not have.

**Fail-on-revert, measured rather than asserted:** removing the new condition fails exactly 4 tests — the 4 added — and nothing else. The file was restored byte-identically afterwards (`shasum` match).

One thing worth naming that the finding didn't: I checked whether the other producers can reach the same state. They cannot, and only by accident. Research raises on empty feedback before its return (`research_step_handler.py:144`) so `feedback_count >= 1` always; the merger requires ≥2 documents (`document_merger/handler.py:66`); the product report hardcodes `product_context_included=True`. The document generator is the only producer that can emit a sources-empty declared derivation — which is why one guard in the resolver, rather than a guard per producer, is the right place for it.

### 4. The lockstep regex over a JSDoc-commented array — real, and the failure mode is worth being precise about

`re.findall(r"'([^']+)'", …)` over the whole array body, now anchored to entry lines: `r"^\s*'([^']+)',"` with `re.MULTILINE`. No comment line ends in `',`.

I ran both regexes rather than reasoning about the quote pairing, and the consequence is not quite what the finding says. Inserting **one** apostrophe into a comment yields no phantom role — every pairing shifts and all four values come back as garbage:

```
OLD, with "The model's reference document…" in a comment:
  ('s reference document, as the request selected it and the generator fed the model. */\n  ',
   ',\n  /** The PRD a prototype was built from. */\n  ', …)
NEW, same input:
  ('reference', 'prototype_prd', 'prototype_prfaq', 'merge_input')
```

A phantom role needs an **even** number of apostrophes. Either way both assertions compare whole tuples, so the outcome was always a loud failure with a confusing message, never silently accepted drift — so this was a legibility problem rather than a correctness hole. Fixed regardless; the comment in the test now records why the anchor is load-bearing.

### 5b, 5c — `type="button"` added, duplicate imports merged

On `type="button"`: worth noting it is a no-op today, because `DerivationSourceRow` has no ancestor `<form>`, and there is currently **no** `type="button"` anywhere in `frontend/src/**/*.tsx`. Added anyway — it is the correct type for a non-submit button and it costs one word — but as future-proofing, not as a fix.

---

## Refuted, with the premise named

### 2. `product_context_included` — the premise is false; `build_product_context_block` returns exactly that placeholder

The finding assumes the producer returns `''` or a header-only block for a project with no product context. It does not. `lambda/api/product_context.py` L717-719:

```python
    if not sections:
        return "(No product context provided.)"
    return '\n\n'.join(sections)
```

Byte-identical to `NO_PRODUCT_CONTEXT` in the generator. I traced every branch: with no context record, `get_context` returns `_empty_context()` with all fields `''`, every `if not v: continue` fires, `sections` stays empty, and the placeholder is returned. Same for an all-empty record. There is no `''` return and no header-only return — each section is header-plus-content by construction, appended only when its content list is non-empty. So `block != NO_PRODUCT_CONTEXT` cannot produce a false `True`, and `bool(block and block.strip())` would be dead code.

**But there is a real weakness one step behind it, and it is wider than the finding suggests.** The flag is a string equality across a module boundary, and it fails *open*: whitespace or wording drift in the producer's copy would make every context-less project record `product_context_included: True`, and nothing else would notice, because the placeholder is only ever read by a human. The literal appears in **eight** production places (`api/product_context.py` ×1, `api/projects.py` ×5, `shared/prompts.py` ×2 defaults) plus the generator's constant.

Rather than change working code, I've added `lambda/api/test/test_product_context_placeholder_lockstep.py`, following `test_kiro_exportable_types_lockstep.py` in the same directory: it reads both literals as **source text** (so it cannot be satisfied by whatever either module resolves at import time, and it needs neither module's AWS-shaped import graph) and fails if they drift. Collapsing the eight copies onto one shared constant is a worthwhile cleanup and a separate change — five of the eight are in `projects.py`'s `generate_prd`/`generate_prfaq`, which have no callers.

### 3. `step_initialize` always carries `derivation` — no early return exists

`step_initialize` spans L117-252 and contains exactly **one** `return` (L233-252, `derivation` at L246) and **one** `raise` (L144, no feedback). The web-search block at L204-231 assigns and falls through inside its own `try/except`; there is no short-circuit.

The raise cannot reach the `resultSelector` either: `initializeStep` is wrapped by `addCatch` (`processing-stack-consolidated.ts:437`), a failed invocation has no `Payload`, and the resultSelector only evaluates on success. `step_save` tolerates absence at L508 (`event.get('derivation') or build_derivation()`), pinned by `test_always_returns_a_derivation_even_when_nothing_was_selected`.

On the deploy-skew half of the finding — the direction it describes is the one that cannot happen, and the CDK comment already says why: the definition `GetAtt`s the function, so CFN always updates the Lambda **before** the definition, and a new definition never runs against old Lambda code. In-flight executions keep the definition they started with, which is the opposite direction, and that is what `step_save`'s `.get()` covers. Same rollout pattern as `web_context` (#157) and `web_search_queries` (#207). I've added a sentence to the PR body making that explicit rather than leaving it implied by a code comment.

One point that *raises* the value of the existing test, though, and is worth recording: a JSONPath miss in a `resultSelector` surfaces as `States.Runtime`, which a `States.ALL` catch does **not** intercept. So a future return path omitting the key would fail *outside* `handleError` and leave the job stuck at `running` rather than `failed`. That is an argument for keeping the "always returns a derivation" test rather than for changing anything today.

---

## Declining, with reasons

**5a — the literal `'derivation'` vs `DERIVATION_FIELD`.** These are two different names that happen to be spelled the same. L246 is the **Step Functions payload key**, whose counterparts are `event.get('derivation')` in `step_save` and `'derivation.$': '$.Payload.derivation'` hardcoded in TypeScript. `DERIVATION_FIELD` is the **DynamoDB attribute name**. Substituting it at L246 would couple the state-machine wire contract to the storage attribute name, so that renaming the column would silently change the payload key. Leaving both as they are.

**5d — the `*_one` plural variants in ja/ko/zh.** Correct that i18next will never select them. But those catalogues already carry ~58 equally unselectable `_one` keys (`activeTokens_one`, `documentsCount_one`, `issues_one`, `reviews_one`, …), so removing only the two new ones would make *them* the inconsistent pair. `localeParity.test.ts` collapses plural suffixes and would accept either. A catalogue-wide sweep is a reasonable separate change; a two-key exception is not.

---

## Correction to the PR body

The body's `vitest run (frontend, full)` line said **2299 passed / 156 files**. That is wrong, and it was captured at the branch's first commit rather than at HEAD — the second commit adds `DocumentsTab.derivation.test.tsx`, which is the entire 156→157 file delta. Measured in a throwaway worktree at each commit: `1b592e6` = 2301/156, `2580e6d` = **2314/157**. No commit on this branch produces 2299. The scoped and backend figures were correct.

Gates at this round's head:

| leg | before | now |
|---|---|---|
| `pytest lambda/shared lambda/jobs lambda/research lambda/api` | 1084 | **1086** (+2, the new lockstep file) |
| `vitest run` (frontend, full) | 2314 / 157 files | **2318** / 157 (+4) |
| `vitest run src/api/derivation.test.ts src/pages/ProjectDetail` | 364 / 35 | **368** / 35 |
| `tsc -b`, `eslint src`, `ruff check` on both touched Python files | clean | clean |

CDK untouched this round.

---

## On the three-document cap — I no longer think it should stay out of scope, but not in this PR

This PR's position is that the cap is "observable, not fixed". I've now read the whole surface, and the framing was incomplete in a way that matters: **it is not one decision, it is three, and they have different answers.**

**(a) Which documents get dropped is a bug, not a policy.** `all_project_items` comes from one `query` on `pk = PROJECT#<id>` with no `ScanIndexForward`, so it is in ascending `sk` order, and both filtering comprehensions preserve it. The request's own array order is never consulted. Among the four accepted prefixes, ASCII ascending is `DOC# < PRD# < PRFAQ# < RESEARCH#` — so `docs[:3]` drops **`RESEARCH#` first, every time**. And since `sk` is `{TYPE}#{doc_type}_{YYYYMMDDHHMMSS}`, ascending is oldest-first, so the cap also keeps the oldest PRDs and drops the newest. Select two PRDs, a PR/FAQ and the research report that motivated the feature, and the research report is the one document the model never sees. Nobody chose that; it is DynamoDB's default sort leaking into a product decision.

**(b) The obvious fix is not sufficient on its own** — this is the part I'd have got wrong without checking. "Spend the cap in the request's order" is right, but `useProjectData.ts` sends `[...selectedDocumentIds, ...selectedResearchIds]` at all three call sites, so research is last in the *request array* too. With a cap of 3 and three non-research documents selected, research is still dropped — just by the frontend's concatenation instead of by DynamoDB. A fix that only touches the backend would look correct and change nothing for the case that matters.

**(c) How much gets through is a genuine judgment call, because the repo answers it four different ways.** The same selection is fed **uncapped** to project chat (`lambda/stream/src/context/project-context.ts:216`), at 8,000 chars each with **no arity cap** to a remix (`document_merger`), at 3×5,000 to research, and at 3×3,000 here. The generator's number is not a considered limit; it is the tightest of four inconsistent ones, and its comment (`# Limit to 3 docs to avoid context overflow`) names a constraint that is not the binding one on a 200K-token model.

**(d) A third defect hides inside the same six lines.** `get_project` returns six document kinds and the picker offers all of them, but the gatherer's prefix filter accepts four — `PRODUCT_REPORT#` and `PROTOTYPE#` are dropped *before* the cap applies, while still counting toward `selected_document_count`. So today's "4 of 5 used" cannot distinguish "past the cap" from "wrong type" from "deleted since": three mechanisms, one number.

And the UI currently promises the opposite of all of it. `dataSourceWizard.leaveEmptyForAllDocuments` reads **"Leave empty to use all documents"** (leaving it empty sends `[]`, and the handler then takes every project document and still slices `[:3]`), and `ContextSummary` echoes all N selected titles back on the confirmation step with no truncation. Select seven, see seven, get three. No locale catalogue mentions a limit at all.

**Recommendation:** replace the arity cap with a **character budget spent in the user's selection order** — the pattern `build_product_context_block` already uses in this repo (one total budget spent down, a deliberate order, and the skipped documents disclosed *in the prompt*, not only in the audit record). Concretely: 8,000 chars per document, which is what the merger already feeds on the same model surface, and 40,000 total, which is inside an envelope this same step-1 prompt already accepts because `{product_context}` can carry 50,000. Five full-length documents fit, or roughly a dozen typical ones. One shared helper, two callers. `derivation` is untouched — the drop simply becomes an exception rather than the normal reading.

**But as its own PR, after this one merges,** for four reasons: this PR changes no generated document and that one changes every one of them; if 40,000 is the wrong number, reverting it must not revert the provenance record; this PR is already 27 files; and the provenance record is the instrument that *measures* the fix — once it is live you can look at real documents and see whether the "N of M" line still appears, which beats a test. Ship the instrument, then the fix.

Two hazards to design against, both found while writing this up: the project `query` is **unpaginated** in the generator, the research step and the merger (`resp.get('Items', [])`, no `LastEvaluatedKey` loop) — past 1 MB of project items the tail is silently lost, and the tail is `RESEARCH#`, so a larger budget makes this reachable sooner. And a fixture with three or fewer selected documents cannot tell a fixed cap from a budget, so the existing five-selected shape has to survive, with a case added where the budget rather than the count is what binds.

Full write-up with citations, the options I rejected and why, and the file-by-file scope is in `analysis/PROPOSAL-reference-document-cap.md` (not part of this diff). The three statements in this branch that call the cap a known separate issue stay true under this sequencing and would be retired by the follow-up.

perrozzi:

Since the last review the author fixed the `isEmpty()` blocker (adding `selected_document_count === 0` to the emptiness check, with new resolver and DocumentsTab tests for the "5 selected, 0 used" case), added `type="button"`, merged the duplicate `derivation` imports, anchored the lockstep role regex to whole entry lines, and added a new source-text lockstep test pinning the `(No product context provided.)` placeholder across `lambda/api/product_context.py` and the generator. The blocker and most nits are addressed.

### Resolved
- `frontend/src/api/derivation.ts` — `isEmpty()` now counts `selected_document_count`, so a "5 selected, none used" record resolves as `declared` and renders; covered by `derivation.test.ts` (including the decisive "does not fall back to legacy fields" case) and `DocumentsTab.derivation.test.tsx`.
- `lambda/api/test/test_product_context_placeholder_lockstep.py` — the producer/consumer placeholder drift is now pinned, and it also establishes that the no-sections branch returns exactly that literal, which closes most of the `product_context_included` gap.
- `DocumentsTab.tsx` — `type="button"` on the source link; single merged import from `../../api/derivation`.
- `test_derivation_roles_lockstep.py` — regex anchored to `^\s*'…',` entry lines, immune to apostrophes in the JSDoc prose (and it fails closed if a trailing comma is dropped).

### Still open
- `lambda/jobs/document_generator/handler.py::_product_context` — still no `bool(block and block.strip())` guard, so a non-placeholder-but-empty/whitespace block (any path other than the `if not sections` branch) would record `product_context_included: True`; the lockstep test narrows this but does not remove it.
- `lambda/research/research_step_handler.py` — the CDK `resultSelector` now hard-references `$.Payload.derivation`; the PR description asserts a single return path but nothing in the diff verifies it, so please confirm there is no early return out of `step_initialize`.
- Nit: `step_initialize` still spells the key literally as `'derivation'` while `step_save` uses `DERIVATION_FIELD`.
- Nit: `ja`/`ko`/`zh` `*_one` plural variants remain (harmless).

### New findings
- Nit — `frontend/src/pages/ProjectDetail/DocumentsTab.tsx`: for the newly-rendering `sources: []`, `selected_document_count: 5` case the panel shows the `documents.derivation.builtFrom` heading with no list beneath it, followed by "0 of 5 selected documents used". It reads slightly oddly ("Built from" → nothing); consider suppressing the heading, or letting the selected/used line carry it, when `sources.length === 0`. Behaviour is correct either way.

---

perrozzi:
Round 2 disposition. **No correctness findings this round.** The blocker is confirmed resolved. The two "Still open" items are both things I answered in prose last round — so rather than answer them again, I've turned both into machine-checked invariants, and positive-controlled each one. The new nit I'm declining, with a reason I think is stronger than the alternative.

Barring something correctness-shaped, this is my last push.

---

## The two "Still open" items — same answers, now enforced instead of asserted

Both were addressed last round with line numbers and a branch trace. Repeating that would just produce a round 3 with the same list. **Prose is not enforcement**, so each is now a structural test over the AST of the code it describes.

### `_product_context` and the missing `bool(block and block.strip())` guard

The finding says a non-placeholder-but-blank block could arrive from "any path other than the `if not sections` branch". There is no such path, and the reason is structural rather than incidental: `build_product_context_block` has **exactly two returns**, and the non-placeholder one joins `sections`, every member of which begins with a visible `###` heading and is appended only when its own content list is non-empty. So a non-placeholder block always starts with `#`, `block.strip()` is always truthy, and the guard is **unreachable — dead code that reads as protection.**

That is an argument, and an argument is worth nothing on its own. `TestNoBlockCanBeBlankWithoutBeingThePlaceholder` now enforces it with two tests: one asserting the producer still has exactly two returns, one asserting every `sections.append(...)` leads with a `###` literal. Either failing is a signal to add the guard, and each message says so.

Positive-controlled both ways, because a structural guard that cannot fail is worse than none:

- Rewriting `sections.append("### Structured product context\n" + …)` as `sections.append('\n\n'.join(structured_lines))` → *A section is appended without a leading "###" heading literal (None) … Either keep the heading, or add the `bool(block and block.strip())` guard*
- Adding a third `return ''` → *build_product_context_block now has 3 returns, not 2*

Both files restored byte-identically afterwards, verified by hash.

### `step_initialize` — confirmed, and now the confirmation is a test

Confirmed again: exactly **one** `return` (L233-258, `derivation` at L246 as declared) and **one** `raise` (L144, no feedback). The web-search block falls through inside its own `try/except`; there is no short-circuit. The raise cannot reach the `resultSelector` — `initializeStep` is wrapped by `addCatch` and a failed invocation has no `Payload`.

You're right that nothing in the diff *verified* it, and that is the better half of the finding. `TestEveryExitCarriesTheDerivation` now reads the function's returns out of its AST and requires every one to declare the key. This covers the paths nobody has written yet, which is the only thing that matters here: a JSONPath miss in a `resultSelector` raises `States.Runtime`, which a `States.ALL` catch does **not** intercept, so a future early return omitting the key would fail *outside* `handleError` and leave the job at `running` — no report, no failure, no diagnosis.

Probe: inserting `return {'feedback_context': '', 'feedback_count': 0}` early fails the test with the injected line number in the message.

**And the companion test earned its place, which I did not expect.** `ast.walk` would collect a nested function's returns as the step's own, so there is a second test asserting `step_initialize` defines no nested function. Injecting a trivial nested `def` made the *first* test report `assert 'derivation' in ['nope']` — attributing the inner return to the step. So the guard-the-guard test is load-bearing, demonstrated rather than argued.

---

## Nits

**`'derivation'` vs `DERIVATION_FIELD`** — declining again, but the reason is now in the code rather than only in this thread, which is what should have happened the first time. L246 is the **Step Functions payload key**; its counterparts are `event.get('derivation')` in `step_save` and `'derivation.$': '$.Payload.derivation'` hardcoded in the CDK. `DERIVATION_FIELD` is the **DynamoDB attribute name**. Coupling them would make renaming the stored column silently change the wire contract. A comment at the call site now says so.

**ja/ko/zh `*_one` variants** — declining, as last round: those catalogues already carry ~58 equally unselectable `_one` keys, so removing only the two new ones makes *them* the inconsistent pair. `localeParity.test.ts` accepts either. A catalogue-wide sweep is a fair separate change.

**New: "Built from" with no list beneath it** — real observation, and I'd already noticed it while making the fix. Declining the suggested change, because suppressing the heading makes it worse in two ways. It would orphan the line: "0 of 5 selected documents used" with nothing above it, in exactly the case where a reader most needs to know which section they are looking at. And the heading is an `<h3>` inside a `<section>`, so it is that region's accessible name — conditionally removing it leaves a landmark with no label, for the benefit of one rare state. Introducing a third string instead would mean an eight-catalogue change for the all-references-deleted path. "Built from → 0 of 5 selected documents used" is terse but true, and true is what that panel is for.

---

## Gates at `f8374b8`

| leg | result |
|---|---|
| `pytest lambda/shared lambda/jobs lambda/research lambda/api` | **1090** passed (1086 + 4) |
| `vitest run` (frontend, full) | 2318 / 157 files |
| `tsc -b`, `eslint src` | clean |
| `ruff check lambda plugins` | **548 findings, delta 0** vs the previous commit |

The ruff delta was measured rather than assumed — same command in a throwaway worktree at `eced304`, identical count, so this round introduced no finding. The three pre-existing findings in `research_step_handler.py` (import order L7, blind `except` L224, naive `datetime.now()` L427) are all untouched by this diff; my change there is a comment.

Nothing in the CDK this round.

---

The cap proposal from my previous comment stands unchanged and is the thing I'd most like a decision on: fix *which* documents get dropped (a bug — ascending `sk` means `RESEARCH#` is always dropped first) and replace the arity cap with a character budget, as its own PR on top of this one.

perrozzi:

Since the last review the author addressed both remaining "Still open" items by argument-plus-enforcement rather than code change: two AST tests in `test_product_context_placeholder_lockstep.py` pin that `build_product_context_block` has exactly two returns and that every appended section begins with a visible `###` heading (so the missing `strip()` guard stays unreachable), an AST test in `test_research_derivation.py` proves every `return` out of `step_initialize` is a dict literal carrying `derivation` (plus a guard-the-guard for nested functions), and a comment now explains why the Step Functions payload key is spelled literally instead of via `DERIVATION_FIELD`. No production code changed, so nothing previously approved is at risk.

### Resolved
- `lambda/api/test/test_product_context_placeholder_lockstep.py` — the `_product_context` blank-block concern is now structurally enforced (two returns, `###`-prefixed sections), so the absent `bool(block and block.strip())` guard is justified rather than assumed.
- `lambda/research/test/test_research_derivation.py::TestEveryExitCarriesTheDerivation` — verifies every exit of `step_initialize` declares `derivation`, closing the unverified single-return-path claim against the unconditional `resultSelector`.
- `research_step_handler.py` — the literal-vs-`DERIVATION_FIELD` spelling is now documented as deliberate (wire key vs DynamoDB attribute).

### Still open
- Nit: `DocumentsTab.tsx` still renders the `builtFrom` heading with no list for the `sources: []`, `selected_document_count: 5` case.
- Nit: `ja`/`ko`/`zh` `*_one` plural variants.

### New findings
- Nit — `lambda/api/test/test_product_context_placeholder_lockstep.py`: `_section_headers()` only matches `sections.append(...)` on a `Name` target, so a future `sections.extend([...])`, `sections += [...]`, or comprehension-built list bypasses the heading check entirely while `test_every_section_starts_with_a_visible_heading` still passes. Since this test is the sole enforcement of the "no block can be blank" argument, consider also failing on any other mutation of `sections` (or asserting the append count matches the number of `sections`-mutating statements).
- Nit — same file: `test_the_producer_has_exactly_two_returns` indexes `functions[0]` without the `len(functions) == 1` assertion its sibling helper has; a rename of `build_product_context_block` yields a bare `IndexError` instead of the intended message. Also, unlike the research test, `ast.walk` here is not guarded against nested function definitions inside the producer.

---

perrozzi:
Round 3 disposition. Both new findings are correct and both are **in the test I added last round**, and the first one is not a nit: it is a fail-open in the only thing holding up a production decision. So I'm breaking the "last push" I declared, and this is why.

## The finding that mattered: I closed a fail-open and opened one in the check

`_section_headers()` matched only `sections.append(...)`. A section arriving by `extend`, `+=` or a comprehension would leave the check's view while the check kept reporting success — and that check is the entire justification for *not* adding the `bool(block and block.strip())` guard. So the argument would have quietly stopped being enforced, which is the same shape as the defect the argument is about, one level up.

I did not want to take that on trust either way, so I measured it. Rewriting one append as `sections.extend([...])`:

```
PASSED  test_every_section_starts_with_a_visible_heading      <- watching nothing, reporting success
FAILED  test_sections_is_only_ever_built_by_append
        sections is mutated by something other than append at [(687, 'extend')] …
        until then treat the missing `bool(block and block.strip())` guard … as no longer justified
```

That green line is the finding, exactly as you described it.

The fix asserts a **closed set** rather than the absence of `extend`: `sections` may be built only by `append`, plus exactly one `= []`. Anything else fails — including a form nobody has thought of yet — so the heading check can no longer be bypassed silently instead of loudly. Your alternative suggestion (assert the append count matches the number of `sections`-mutating statements) is the same idea; I went with enumerating the kinds because the failure message can then name the form that was written.

Five positive controls, all caught, each naming the offending line or count:

| mutation | result |
|---|---|
| `sections.extend([...])` | `[(687, 'extend')]` — and the heading test stays green, per above |
| `sections += [...]` | `[(687, '+=')]` |
| producer renamed | *Expected exactly one build_product_context_block … found 0 … update PRODUCER_FUNCTION* |
| nested `def` inside the producer | *now defines nested function(s) ['_inner'], whose statements ast.walk cannot tell from the producer's own* |
| `sections` initialised twice | *Expected exactly one `sections … = []` … found 2 at lines [661, 662]* |

`product_context.py` was restored byte-identically after every probe, verified by hash each time.

## The second finding, and one you didn't raise that it led to

`functions[0]` indexed without the `len(functions) == 1` assertion — correct, and a rename gave a bare `IndexError` instead of the message naming the rename. Both tests now go through a single `_producer_function()` helper that asserts uniqueness **and** that the producer defines no nested function, so the `ast.walk` guard you correctly noticed was missing here is now present on both sides. Verified: a rename produces the named `AssertionError`, not an `IndexError`.

While probing, one more thing surfaced that neither of us had flagged: `_section_mutations()`'s docstring advertised a `'+='` kind the code **could never emit** — the `AugAssign` branch fell through to `'rebind'`, so `sections += [...]` and `sections = something` were indistinguishable in the message. A label a reader cannot rely on is the same defect in miniature. `AugAssign` now reports `'+='`, confirmed by re-running that probe.

Note on blast radius, since it is the correct failure direction but worth knowing: the nested-function guard lives in `_producer_function()`, so adding any nested helper to the producer fails four tests at once with the same message rather than one.

## Still declining, unchanged

**`builtFrom` heading with no list.** Suppressing it orphans the "0 of 5" line in precisely the case where a reader most needs to know which section they are in, and the heading is the `<h3>` naming that `<section>` — conditionally removing it leaves a landmark with no accessible name for one rare state. A third string instead is an eight-catalogue change for the all-references-deleted path.

**ja/ko/zh `*_one`.** Those catalogues already carry ~58 equally unselectable `_one` keys; removing only the two new ones makes them the inconsistent pair. A catalogue-wide sweep is a fair separate change.

## Gates at this head

pytest **1092** passed · frontend 2318 / 157 · `tsc -b` and `eslint src` clean · `ruff check lambda plugins` **548, delta 0**. No production code changed this round, so nothing previously reviewed moved.

---

Unless something correctness-shaped arrives, that really is the last push. The open question remains the cap proposal two comments up — fixing *which* documents get dropped is a bug fix, and replacing the arity cap with a character budget is the judgment call I'd like a decision on, as its own PR on top of this one.

perrozzi:

Since the last review the only change is `lambda/api/test/test_product_context_placeholder_lockstep.py`: the AST helpers were consolidated into `_producer_function()` (with the rename assertion and a nested-`def` guard) and `_section_headers()` was replaced by `_section_mutations()`, which enumerates every mutation of `sections` and fails closed on any form the heading check cannot read. Both of my previous nits on that file are addressed; no production code changed, so nothing previously approved is affected.

### Resolved
- `test_the_producer_has_exactly_two_returns` now goes through `_producer_function()`, so a rename yields the intended message instead of `IndexError`, and `ast.walk` is guarded against nested definitions.
- The heading check can no longer pass vacuously: `test_sections_is_only_ever_built_by_append` fails on `extend`/`+=`/`setitem`/rebind, and `test_sections_starts_empty_exactly_once` pins the single empty init.

### Still open
- Nit: `DocumentsTab.tsx` renders the `builtFrom` heading with no list for the `sources: []`, `selected_document_count: 5` case.
- Nit: `ja`/`ko`/`zh` `*_one` plural variants.

### New findings
- Nit — `lambda/api/test/test_product_context_placeholder_lockstep.py`: `_section_mutations()` collects *any* method call on `sections`, so a read-only call (`sections.copy()`, `sections.index(...)`, `sections.count(...)`) is reported as an unreadable mutation and fails `test_sections_is_only_ever_built_by_append`. Fail-closed, so harmless, but the docstring's "Every statement that changes `sections`" overstates what it collects.
- Nit — same file: a bare annotation (`sections: list[str]` with `node.value is None`) hits the `AnnAssign` branch and is labelled `'rebind'`, another false positive in the same direction.

---

perrozzi:
Round 4 disposition, and I'm closing the loop here.

Both new findings are correct, both are **fail-closed**, and you said so yourself — which is what makes this the stopping point rather than another round of the same shape. A guard that fires when it shouldn't asks a human a question; a guard that stays quiet when it should fire is the defect round 3 found. Those are not the same severity, and the loop should end when only the first kind is left.

Fixed anyway, because one of the two was a false statement in code rather than a false positive:

- **The docstring overstated what the scan collects.** It said "every statement that changes `sections`" while the scan reports any method call on it, `.copy()` and `.index(…)` included. That behaviour is deliberate: treating some methods as harmless means keeping a list of method names known to be read-only, which is denylist-shaped and fails **open** for the next method nobody thought of — precisely the failure this helper exists to prevent. So the code was right and the documentation was wrong about it. The docstring now says what it does and why.

- **The bare annotation was a genuine false positive worth removing.** `sections: list[str]` with no value writes nothing, so a failure there has no question behind it — unlike a read-only call, where there is something for a reader to decide. Skipped now. Probed by splitting the initialisation into a bare annotation plus `sections = []`: 6 passed, where the annotation alone previously failed the append enumeration.

`product_context.py` restored byte-identically after the probe, verified by hash. pytest **1092**, ruff clean, no production code touched.

---

## Where this leaves the PR

Four rounds, and the arc is worth stating plainly because it says something about where the risk actually was:

- **Round 1** found one real bug, in production code: a declared derivation of "five selected, none used" rendered as nothing at all. Two of the five findings rested on premises that turned out to be false, and I said which.
- **Round 2** re-raised two items I had answered in prose. The right response was not to answer them again but to make them machine-checked, which is now four structural tests.
- **Round 3** found that one of those tests was itself fail-open — the check would have stopped watching without saying so. That was the best finding of the four rounds and it was in code I wrote to close a fail-open.
- **Round 4** is fail-closed only.

Everything after round 1 is enforcement rather than behaviour: production code changed in exactly two lines this whole review (one condition in `isEmpty`, one `type="button"`), and the remaining ~150 added lines are tests and the invariants they pin. Standing declines, both restated with reasons above and unchanged: the `builtFrom` heading for the sources-empty case, and the ja/ko/zh `*_one` variants.

I'm treating further rounds of this shape as not-new-information and stopping here. Anything correctness-shaped, I'll break that again and say why, as I did on round 3.

**The one open decision is not in this diff.** The three-document cap: fixing *which* documents get dropped is a straight bug fix — ascending `sk` order means `RESEARCH#` is always the first thing discarded and the oldest documents are always kept, which nobody chose — while replacing the arity cap with a character budget is a judgment call the repo currently answers four different ways in four places. Proposal, options rejected, and file-by-file scope are two comments up; my recommendation is its own PR on top of this one, so that a change to what the model sees stays revertable independently of the record that measures it.

perrozzi:

Only one file changed since my last review: `lambda/api/test/test_product_context_placeholder_lockstep.py`, where `_section_mutations()` got a rewritten docstring and a new `AnnAssign`-with-no-value skip. No production code touched, so nothing previously approved is affected.

### Resolved
- The bare-annotation false positive is gone: `sections: list[str]` with no value is skipped instead of being reported as `'rebind'`.
- The docstring no longer overstates the collection ("writes to `sections` or calls a method on it"), and now states explicitly that read-only method calls are reported on purpose — fail-closed by design rather than by accident. That's the right call over a read-only-method denylist.

### Still open
- Nit: `DocumentsTab.tsx` renders the `builtFrom` heading with no list for the `sources: []`, `selected_document_count: 5` case.
- Nit: `ja`/`ko`/`zh` `*_one` plural variants are unused in those locales.

No new findings — the delta is clean, and both remaining items are cosmetic nits I'm not blocking on.

---

perrozzi:
## Deployed and verified live

`367c17e` is deployed to the shared development environment and both provenance paths are confirmed against real data, through the UI rather than through the API.

**Scope of the deploy.** Two stacks, `--exclusively`: `VocProcessingStack` (the research state machine definition) and `VocApiStack` (the API and job Lambdas, and the UI, which that stack ships). `origin/development` was verified to be an ancestor of the head first, so the deploy could not revert anything merged since the branch was cut. The reviewed diff for those two stacks contained **no IAM changes** — only `Code.S3Key` hashes and the state-machine `DefinitionString`. `UPDATE_COMPLETE` for both, 35s and 160s.

**The state machine really threads it.** Read back from the deployed definition, not from the template:

- `InitializeResearch.ResultSelector` contains `derivation.$`
- `SaveResearchResults` payload contains `derivation.$` → `$.initialize_result.derivation`

**The declared path, end to end on the production route.** I generated a document from the wizard with **five** reference documents selected. It ran through the Step Functions claim-check path (the live path for PRD/PR-FAQ), so this exercises the S3 stash in `gather` and the read-back in `save`, not the single-shot fallback. `SUCCEEDED` in 98s. What the live Lambda stored:

```json
{
  "sources": [
    {"document_id": "<prd-a>",   "role": "reference"},
    {"document_id": "<prd-b>",   "role": "reference"},
    {"document_id": "<prfaq-a>", "role": "reference"}
  ],
  "selected_document_count": 5,
  "feedback_count": 30,
  "persona_ids": [],
  "product_context_included": true
}
```

`feedback_count: 30` is the number that reached the prompt, not the query limit of 100 — which is the distinction the record is supposed to make. `product_context_included: true` is correct: that project has product context filled in, so the flag defended by the new lockstep test is exercised rather than merely asserted.

And the tab renders it:

```
Built from
PRD    real time delivery tracking                Reference
PRD    PR6 Regression PRD - support automation     Reference
PRFAQ  real time delivery tracking                Reference

3 of 5 selected documents used

30 feedback items used · Product context included
```

Checked on the live DOM: all three source links carry `type="button"`, the panel contains **zero** `svg` elements, and no element from the "3 of 5" line up to the panel root carries a red/amber/yellow/orange class. So the neutral-display property the DOM test pins is true of the deployed bundle, not only of jsdom.

**The legacy path, on a document written before the field existed.** A prototype from July, carrying only `source_prd_id`/`source_prfaq_id`, renders:

```
Built from
PRD    PR6 Regression PRD - support automation     Prototype source PRD
PRFAQ  Automated Insights Digest                   Prototype source PR/FAQ
```

Clicking a source moved the detail pane to that document, and that document — a PR/FAQ with no lineage of its own — renders **no panel at all**, which is the third case verified by the same click.

**Other gates.** `scripts/test-api.sh` 34/34. The served `index.html` references the same asset hash as the local build, and the deployed `config.json` carries all four Cognito values non-empty including `identityPoolId`. The deployed generator bundle was downloaded and confirmed to contain all three derivation write sites.

---

### One correction to my earlier comment

I attributed the inline comment `# Limit to 3 docs to avoid context overflow` to the document generator. That string occurs once in the repo and it is on the **research** step handler; the generator carries this PR's own comment. The point it was supporting is unaffected, but the citation was wrong.

### And the cap, now that live data has spoken

The decision is taken: the cap stays out of this PR and is filed separately. The run above is a good argument for that issue, because it shows all three drop mechanisms at once. Five documents selected, three used — and the three kept are the three lowest-sorting eligible ones (both PRDs, then the earlier PR/FAQ), which is ascending sort-key order and not the order the request listed them in. Of the two that did not make it, one is a PR/FAQ cut by the cap and the other is a prototype that could never have qualified, because the gatherer's prefix filter does not accept `PROTOTYPE#` even though the picker offers it. The stored record cannot tell those two cases apart — it says only "3 of 5".

The wizard's own confirmation step, meanwhile, listed all five titles back with no truncation and no warning before generating.

**Left on the environment:** one disposable test document titled *PR317 derivation live verification* in the project used above. Say the word and I will delete it; I left it in place as the evidence for this comment.

PR #320, opened 2026-08-12
feat(documents): choose which spec a prototype is built from

perrozzi:
## The problem

Building a prototype read **the newest PRD and the newest PR/FAQ**, both at once, with no way to say which.

On a real project that is not a default so much as a coin toss. The project I developed against has one PRD,
three PR/FAQs and six prototypes — and two of those three PR/FAQs were not "not chosen", they were
**unreachable**: no parameter existed through which a choice could be expressed, even from the API. Worse, the
choice moved on its own: saving a new PR/FAQ silently re-pointed every future build in that project, same
button, different spec, no prompt.

PR #317 made the source *visible after the fact* (the "Built from" footer). It did not make it *choosable*.
This closes that, plus three smaller defects that all come from the same place.

| | before | after |
|---|---|---|
| Which PRD / PR-FAQ a build reads | newest of each, unconditionally | the ones you pick; newest of each is the default |
| Telling two same-typed documents apart | badge + date + title — and six prototypes were all titled "Prototype", four sharing a date | `2 of 3` per type, from creation order |
| What a prototype revision revises | `revised_from_id` + `revision_feedback` written on every revision since that feature shipped, **read by nothing** | named in the detail pane, linked while it exists, with the feedback that drove it |
| Revising a prototype | re-resolved "newest", so revising a prototype built from June's PRD silently re-based it on September's | inherits the sources of the prototype it revises |
| Finding the newest of a type | a 20-row window ordered by `sk`, then re-sorted by `created_at` | every page read, ranked on `created_at` |

## 🔍 Review this hardest: the source id is a trust boundary

The ids arrive from the client and the named document's **text is read straight into a Bedrock prompt**, so an
id that resolved outside the project would pull another project's content into this project's generation.

Two properties carry it:

- **Ownership and type come from the key, not from a check.** `pk` is the project and `sk` is `{TYPE}#{id}`, so
  a single `get_item` settles existence, project ownership and document type together. There is no id that can
  address another partition, and a PR/FAQ id offered as a PRD resolves to nothing. The supplied string reaches
  the sort key only — `../PROJECT#victim/prd_1` is just an id that matches nothing.
- **An unresolvable id fails; it is never replaced by the newest.** That fallback is the whole point: it would
  produce a prototype built from a document the user did not choose, and nothing in the output would reveal it.
  The API rejects with a 4xx naming the field *before a job exists*, so a bad id costs a rejection rather than
  a multi-minute billable build. The generator refuses it again independently, which covers a document deleted
  between the two calls.

## Two design decisions worth agreeing or objecting to

**The dialog does NOT open when there is exactly one of each.** A dialog offering a choice with one possible
answer is friction, and that build has started on the first click since it shipped. An existing test pins this
(`builds immediately without a confirmation when both documents exist`) and it stays green. The dialog appears
when a type has more than one document — i.e. when "the latest" is genuinely a decision.

**The ids are sent whether or not the dialog appeared.** They name the documents the card was *showing*, which
removes a class of disagreement rather than relocating it: without them the backend re-resolves "the newest" at
build time, so a document saved between render and click would be used instead of the one just confirmed. The
consequence is that the picker's ordering must mirror the backend's newest-of-type rule **exactly, ties
included** (`created_at` then `document_id`, descending) — and a tie is not hypothetical, since ids carry a
whole-second timestamp and the test project has four prototypes sharing a date.

## Deployed and verified live before this PR was opened

`VocApiStack` only — `VocProcessingStack` had zero diff. The `cdk diff` was 18 Lambda `Code.S3Key` moves plus
the frontend deployment, with **no IAM and no non-`Code` property changes**.

The decisive test aimed a real build at the **oldest** PRD and **oldest** PR/FAQ, because the pre-change
behaviour is newest-of-each — so aiming at the newest would be indistinguishable from the feature not working:

```
prototype_20260812100757
  source_prd_id   : prd_20260624155112    (oldest of 2; newest was prd_20260625231508)
  source_prfaq_id : prfaq_20260624154929  (oldest of 3; newest was prfaq_20260811225625)
  derivation      : [prd_…155112/prototype_prd, prfaq_…154929/prototype_prfaq]
```

Trust boundary, all 8 cases live, **none creating a job**: unknown id 404 · a PR/FAQ id as `source_prd_id` 404 ·
another project's id 404 · list/object/number 400 · 5000-char id 400 · traversal-shaped id 404.

UI driven in a real browser (Playwright) against the deployed app: ordinals render (`PROTOTYPE 1 of 7`…`7 of 7`,
so the four Jul 10 prototypes are finally distinguishable), the aimed build's footer names the two oldest
documents, the picker shows all three PR/FAQs with the newest marked `(latest)`, selecting does not dismiss the
dialog, and Cancel starts nothing. `scripts/test-api.sh` → **34/34 passed**.

## Tests, and what they are worth

Frontend **2353 passed**, backend **1503 passed**, ESLint clean at `--max-warnings 0`, `tsc -b` clean.

Revert-sensitivity demonstrated by mutation, **18 mutations, 18 caught**. Two survived the first run and both
were real test defects I then fixed:

- `test_an_unresolvable_id_fails_instead_of_falling_back_to_the_newest` passed under the exact mutation it
  exists to catch. The fixture starved the mock, so the fallback consumed the one queued response, the next
  lookup exhausted the list, and the job failed *for that reason* — `pytest.raises` matched. A negative test
  must let the forbidden behaviour **run to completion**; the fixture now lets a fallback succeed, so
  `put_item.assert_not_called()` is what fails.
- Re-adding `Limit=20` passes every behavioural test and always will: a mocked table returns the items it was
  handed whatever the cap says. Its absence is asserted on the query call itself
  (`test_the_lookup_sends_no_row_cap`), which is the only mechanism available below a real DynamoDB. The test
  says so, because it otherwise looks like ceremony.

## Notes for the reviewer

- **`api/wireRecord.ts` is new but it is a deletion.** `asRecord`/`displayString` already existed in
  `api/derivation.ts`, and my first draft of `documentLineage.ts` wrote them again under different names. Both
  now import one copy; `derivation.ts`'s 34 tests pass untouched.
- **`ConfirmModal` gained an optional `children`.** `message` stays required, so a dialog cannot end up with
  controls and no question, and no existing caller changes.
- **The eight locale catalogues are a pure insertion** — 9 lines each, zero removals. A first attempt re-sorted
  each file and moved 9 existing lines in es/fr/pt, which group i18next plural variants after their base key
  rather than alphabetically. `documents.prototype.sourcePrfaq` is identical to English in every locale on
  purpose: PR/FAQ is an acronym, and `prioritization.json`'s `docType.prfaq` is already carried that way.
- **Pre-existing red gates, unrelated to this PR:** `i18n:check` exits 1 on a baseline (`categories.json` extra
  plural keys, `prioritization.json` untranslated `docType.prfaq`); my keys contribute
  `missing: 0, extra: 0, empty: 0` in all eight locales. `documents.prototype.confirmChooseSources` is listed
  as an unused key **exactly as its three existing siblings are** — the keys live in the `CONFIRM_MESSAGE`
  const, which the extractor does not read as a usage, and that const is deliberate.
- **No `app-baseline` change**, because `synth-app.ts` normalises asset hashes; a code-only Lambda edit cannot
  move it.

## Deliberately out of scope

- **Whether the prototype should consume research and product context** — the eighth U25 criterion, held as a
  generation-cost and output-quality decision rather than a defect.
- **The 3-document reference cap**, which is its own filed issue. #317 made the drop observable; nothing here
  changes it.
- **Whether the source choice should be remembered per project** rather than per build. Per-build is the
  smaller of the two and is what this assumes.


perrozzi:

This PR makes the PRD / PR-FAQ a prototype is generated from explicitly selectable (new `source_prd_id` / `source_prfaq_id` on `POST /projects/{id}/build-prototype`, validated at the API boundary and again in the generator), adds a source picker inside `ConfirmModal`, surfaces the previously-unread `revised_from_id` / `revision_feedback`, adds per-type ordinals, and replaces the `Limit=20`+`sk`-ordered "latest of type" read with a fully-paged `created_at` ranking. Overall this is high quality and follows repo conventions: Zod-parsed wire records instead of `as`, all eight locale catalogues updated, no IAM/model-id changes, and dense behavioural tests on both sides including the trust-boundary cases. The key/prefix argument for ownership and type isolation (`pk = PROJECT#…`, `sk = {TYPE}#{id}`) is sound, and the "no fallback to newest on an unresolvable id" decision is the right one. A few things I'd want addressed before merge.

### Should fix

- **A deleted source document now hard-fails a revision with no recovery path.** `PrototypeFeedbackButton` inherits `sourcePrdId`/`sourcePrfaqId` from the base prototype (`voc-datalake/frontend/src/pages/ProjectDetail/DocumentsTab.tsx`) and sends them verbatim; `_validated_source_id` then returns 404 if that PRD has since been deleted (`voc-datalake/lambda/api/projects_handler.py`). Previously the revision would have rebuilt from the newest. There is no UI to clear the inherited ids, so such a prototype becomes permanently unrevisable, and the user sees a raw `source_prd_id: no such document in this project`. Note that `usePrototypeBuild.effectiveSourceId` already implements exactly this fallback for the Overview card ("a selection whose document has been deleted silently reverts to the default") — the revision path needs the equivalent (validate the inherited id against `documents` before sending, or retry unaimed on a 404), plus a test.
- **`_latest_doc_by_prefix` now reads every attribute of every page.** `voc-datalake/lambda/jobs/document_generator/handler.py` pages the whole `PRD#`/`PRFAQ#` range with no `ProjectionExpression`, so it pulls full document bodies out of DynamoDB purely to compare `created_at`/`document_id`. Project a minimal key set (`pk, sk, created_at, document_id`), pick the winner, then `_document_by_id` it — that keeps the correctness fix while bounding payload and removing most of the reason to worry about the now-unbounded loop.
- **Confirm `Key` is still imported.** The function-local `from boto3.dynamodb.conditions import Key` was deleted from `_latest_doc_by_prefix` and the new code calls `Key('pk')`, but the diff adds no module-level import. If one doesn't already exist at the top of `document_generator/handler.py` this is a `NameError` on every prototype build.
- **The frontend tie-break has no test.** The PR argues the picker's ordering must mirror the backend "ties included", and the backend has `test_a_tie_on_created_at_resolves_on_id_descending`, but `sourceOptions`/`compareDescending` in `voc-datalake/frontend/src/pages/ProjectDetail/overviewState.ts` is only covered indirectly for the date case (`OverviewTab.prototypeSources.test.tsx` uses distinct dates). `ordinalByType`'s tie test is a different code path. Add an `overviewState` test with two same-`created_at` PRDs asserting `prdOptions[0]` is the higher `document_id`, otherwise the two rules can silently drift.

### Nits

- `_validated_source_id` returns 404 (`NotFoundError`) for an unresolvable body field. That's defensible, but a client can't distinguish it from "project not found"; consider 400 with a machine-readable field name.
- `base_prototype_id` is still passed through unvalidated in the same handler while its two new siblings are validated — worth the symmetry even though the S3 key path is scoped by `project_id`.
- `DocumentsTab.revision.test.tsx` reaches into `getElementsByTagName('button')[0]`; `within(panel).getByRole('button')` would survive markup changes.
- `ConfirmModal` renders an empty `<div className="mt-3">` whenever the child component returns `null` (e.g. `PrototypeSourcePicker` with no PRD and no PR/FAQ) — worth guarding on rendered content or leaving the wrapper to the caller.
- `ordinalByType` skips id-less entries but groups any document with a missing/non-string `document_type` under `''`, so unrelated malformed records share one numbering; skipping them too would be more consistent with the "must not inflate a total" rule.

---

perrozzi:
Thanks — three of the four should-fixes were real, and one of them I had already got wrong once and talked myself out of. Disposition below; the push follows this comment.

### Fixed

**1. A deleted source document made a prototype permanently unrevisable.** Correct, and the sharpest finding here: you spotted that `usePrototypeBuild.effectiveSourceId` already implements exactly this fallback for the Overview card, so the revision path was inconsistent with my own design. `DocumentsTab` now drops an inherited id that is no longer in `documents` (`stillPresent`), per slot rather than all-or-nothing, so a deleted PRD does not also discard a live PR/FAQ. Blank is defensible here specifically *because* the document is gone — the spec that inheritance existed to preserve no longer exists, so newest-of-type is not a silent substitution but the only remaining answer. Test: `drops an inherited source id that is no longer in the project`.

Worth flagging: adding that guard **failed one of my own existing tests**, and the test was wrong. `sends the base prototype's own sources` asserted `prfaq_june` was inherited while never putting that PR/FAQ in the project — so it had been asserting the fallback, not the inheritance. Fixture corrected rather than the code.

**2. `_latest_doc_by_prefix` reading every attribute of every page.** You're right and my earlier reasoning was wrong in the direction that matters. I had built exactly the projection + keyed-fetch you describe, then backed it out because it broke two `get_item` doubles, and justified that with "the capped read already paid the per-item cost". True per item, and beside the point: unbounding the page count while keeping full items trades a 20-document payload for an all-documents one. I let a fixture inconvenience drive a production design decision. Now `_newest_document_id` ranks over `ProjectionExpression: 'document_id, created_at'` and `_latest_doc_by_prefix` fetches only the winner via `_document_by_id`. The two doubles were under-specified, not obstacles — they now answer the keyed read, with a comment saying why.

**4. The frontend tie-break had no test.** Correct, and it's the one claim the PR body made loudest ("ties included") while testing it only on the backend. New `overviewState.prototypeSources.test.ts` asserts three same-`created_at` PRDs order `prd_c, prd_b, prd_a`, plus that `hasPrd`/`hasPrfaq` stay consistent with the lists they summarise.

**Nits 3, 4, 5.** `within(panel).getByRole('button')` replaces both `getElementsByTagName` reaches. `ConfirmModal` no longer wraps `children` at all — `children != null` cannot detect a component that *renders* null, so spacing moved to the picker's own root and a null child now renders nothing. `ordinalByType` skips entries with no usable `document_type` for the same reason it skips id-less ones; extracted to a `numberable()` helper, which also brought the function back under the cognitive-complexity limit that the extra guard had pushed it over.

### Not changing

**3. `Key` is already imported.** Module level, `document_generator/handler.py:14`. Independent evidence: I ran a real aimed build against the deployed stack before opening this PR and it completed — a `NameError` there would have failed every prototype build.

**Nit 1, 404 vs 400 for an unresolvable id.** Keeping 404. `NotFoundError` is what this handler already raises for a named-but-absent resource (`Job not found`, `Token not found`), the message names the field (`source_prd_id: no such document in this project`), and 404-vs-400 is the same distinction the API already draws elsewhere. Happy to switch if you'd rather the body-field case be uniformly 400 — it's a one-line change plus two test expectations.

**Nit 2, `base_prototype_id` unvalidated.** Agreed on the asymmetry, declining in this PR. Validating it would change behaviour beyond this change's scope: a stale base id is currently ignored (the generator's `get_item` misses and it revises without prior HTML), and rejecting it would turn a degraded-but-working path into a 404. It is already scoped by `project_id` in both the S3 key and that `get_item`, so there's no cross-project reach. Worth its own issue.

### Verification after these changes

Frontend **2360 passed**, backend **1503 passed**, ESLint clean at `--max-warnings 0`, `tsc -b` clean, ruff delta zero.

Mutation probes re-run: **13/13 caught** (5 backend, 8 frontend). Three of the backend mutations reported `SKIPPED` on the first re-run because splitting `_latest_doc_by_prefix` moved their anchors — worth naming, because a skipped mutation still ends the run "restored, exit 0" while the row cap, the `created_at` ranking and the pagination quietly stop being verified. Anchors repointed; all three catch again.

One honest loose end: a single run of the full frontend suite showed `ProductTab report guard > does not ask for the document list when a field is already filled` failing. It passes in isolation (7/7) and in two consecutive full runs since, and that file is untouched here — so it looks like cross-file pollution surfaced by the new test file changing worker scheduling, not a regression. Naming it rather than calling it fixed.

perrozzi:

## Follow-up review

Since the last round the author addressed all four Should-fix items and four of the five nits: a `stillPresent` guard (plus regression test) keeps a prototype revisable after its source PRD is deleted, `_latest_doc_by_prefix` now ranks over a `ProjectionExpression` and fetches only the winner in full, a new `overviewState.prototypeSources.test.ts` pins the tie-break against the backend rule, and the `within(...)`/`numberable`/no-wrapper-div cleanups landed. The remaining deltas are small and mostly test-fixture plumbing.

### Resolved
- Deleted inherited source no longer bricks revision — `stillPresent` in `DocumentsTab.tsx` + `DocumentsTab.revision.test.tsx` per-slot fallback test.
- `_newest_document_id` projects only `document_id, created_at`, then `_document_by_id` fetches the winner — payload bounded again.
- `Key` import concern: the new `test_prototype_sources.py` tests exercise `_newest_document_id` on every build, so a missing module-level import would fail them (I still can't see the import line in either diff — worth one glance).
- Frontend tie-break now tested directly (`breaks a same-second tie on document_id DESCENDING`).
- Nits: `within(panel).getByRole('button')`, `ConfirmModal` empty-wrapper removed with `mt-3` moved into `PrototypeSourcePicker`, `ordinalByType` no longer groups type-less records under `''`.

### Still open
- `_validated_source_id` returns 404 rather than 400 with a machine-readable field (nit).
- `base_prototype_id` still unvalidated alongside its two validated siblings (nit).

### New findings

- **`_newest_document_id` now skips any document whose item lacks a `document_id` attribute** (`voc-datalake/lambda/jobs/document_generator/handler.py`). The old code returned the query item itself, so a legacy row that carries only `pk`/`sk` (no duplicated `document_id` attribute) still won; now it is dropped and the build silently reads an older document, or reports "No PRD or PR/FAQ found". If any historical document shape can lack that attribute, derive the id from `sk` (`sk.split('#', 1)[1]`) as a fallback rather than `continue`.
- **`_wire`'s auto-population hard-codes the `PRD#` prefix** (`voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py`): items from *every* queries page — including the PR/FAQ lookup — are registered as `PRD#{document_id}`. It's inert today because all PR/FAQ pages are `NO_DOCUMENTS`, but the first test that gives the PR/FAQ query real items will see the newest-of-type id resolve and then `_document_by_id('PRFAQ#…')` return nothing, i.e. a silently source-less build that looks like a product bug. Key the prefix off the call index or take it per-page.
- Nit: `numberable`'s new "an id without a type is also skipped" rule isn't pinned — `documentLineage.test.ts` only covers the id-less case (`{ document_type: 'prd' }`). One entry with `document_id` and no `document_type` asserting `ordinals.size` unchanged would lock it.
- Nit: when `stillPresent` drops an inherited id, the revision dialog says nothing about having re-based on newest-of-type; a line in the feedback panel would keep this from being a silent (if justified) substitution from the user's point of view.

---

perrozzi:
Round 2: both new findings were real, both fixed, plus the two nits. Push follows this comment.

### Fixed

**`_newest_document_id` skipping items with no `document_id` attribute.** A real behaviour change I introduced with the projection — the previous unprojected read returned the item itself, so such a row still won. I measured rather than guessed: a scan of the live table found **0 of 6** `PRD#`/`PRFAQ#` rows without the attribute, so this is not a live defect. Taking the fix anyway, because it costs two lines and your reasoning holds structurally: with the read now projected there is no `content` to fall back on, so an id-less row would be silently dropped and the build would use an older document or claim none exist. `_document_id_of` prefers the attribute and falls back to `sk.split('#', 1)[1]`; `sk` is in the projection purely for that, and it cannot be absent because it is the key. The docstring records the 0-of-6 measurement so the next reader knows it is defence in depth rather than a patched bug.

**`_wire` hard-coding the `PRD#` prefix.** Good catch, and worse than inert-today: the failure mode you describe is a *source-less build that reads like a product bug*, which is the most expensive kind of fixture defect. Fixed by construction rather than by a smarter guess — pages are now supplied per type (`prd_pages=` / `prfaq_pages=`), so the helper knows the prefix instead of inferring it, and `query.side_effect` is assembled PRD-then-PR/FAQ so the production call order is expressed by the parameter names rather than by position in a flat list. Nine call sites updated; the strictness that makes the aimed tests prove they never scan is preserved.

**Nit: the type-less skip wasn't pinned.** Added — an entry with a `document_id` and no `document_type`, plus one with a non-string type, asserting `ordinals.size` is unchanged and the sibling total stays 2.

**Nit: the silent re-base.** You're right, and it was inconsistent with the whole point of this PR: the fallback is justified but leaving it unsaid makes it exactly the kind of unexplained substitution the picker exists to remove. The revision panel now carries a line when an inherited source has been deleted (`documents.prototype.feedbackRebased`, 8 locales, insertion-only diff again — 1 line each, zero removals), with two tests: it appears when a source is gone, and it does not when every inherited source is present.

### Still declined, deliberately

- **404 vs 400** — matches `NotFoundError` usage already in this handler, and the message names the field. One-line change if you'd rather it be uniformly 400.
- **`base_prototype_id`** — validating it would turn a currently degraded-but-working path (stale base id → revise without prior HTML) into a 404. Belongs in its own change; it's already project-scoped in both the S3 key and the generator's `get_item`.

### On the `Key` import

`voc-datalake/lambda/jobs/document_generator/handler.py:14`, module level:

```python
from boto3.dynamodb.conditions import Key
```

It predates this PR, which is why it isn't in the diff. Settled twice over now: `test_prototype_sources.py` exercises this path on every build, and a real aimed build ran against the deployed stack before the PR was opened.

### Verification

Frontend **2363 passed**, backend **1503 passed**, ESLint clean at `--max-warnings 0`, `tsc -b` clean, ruff delta zero. Mutation probes **18/18 caught** (10 backend, 8 frontend).

One thing worth flagging about my own tooling, because it bit twice in one session: mutation **M1** ("re-add the row cap") reported `SKIPPED` after round 1's refactor and again after this round's projection change — both times because it was anchored to a line the round's own edit touched, and both times the run still ended `RESTORED: exit=0` while the row cap quietly stopped being verified. A skipped mutation is indistinguishable from a passing one unless you read for it. Repointed onto the params-dict close plus the loop header, which neither a projection nor a ranking change touches.

perrozzi:

## Follow-up review

The incremental changes are exactly the three items I raised last round: `_document_id_of` derives the id from `sk` when the projected attribute is absent, `_wire` now takes `prd_pages`/`prfaq_pages` so PR/FAQ items key under the right prefix, `documentLineage.test.ts` pins the type-less-but-id-bearing case, and a `hasDroppedSource` → `feedbackRebased` note (with two tests and all eight locale catalogues) makes the inherited-source fallback visible instead of silent. Clean round.

### Resolved
- Legacy row without `document_id` no longer dropped from the newest-of-type ranking (`_document_id_of` + `ProjectionExpression: 'sk, document_id, created_at'`).
- `_wire` keys fixture items per type, so a future PR/FAQ page cannot resolve to a `PRD#` key.
- `numberable`'s "id but no type is skipped" rule now asserted, including a non-string `document_type`.
- Dropped inherited source is announced in the revision panel rather than substituted silently.

### Still open
- `_validated_source_id` returns 404 rather than 400 with a machine-readable field (nit).
- `base_prototype_id` still unvalidated alongside its two validated siblings (nit).

### New findings
- Nit: the `sk` fallback branch in `_document_id_of` (`voc-datalake/lambda/jobs/document_generator/handler.py`) has no test — one `_wire` page whose item carries `sk`/`created_at` but no `document_id`, asserting the build still reads it, would keep it from rotting into unreachable code.
- Nit: `t('documents.prototype.feedbackRebased')` in `DocumentsTab.tsx` omits `defaultValue`, unlike every sibling `t()` call in the same modal; keys exist in all eight locales so behaviour is fine, just inconsistent with the local convention.

---

perrozzi:
Round 3 is nit-only, so this is the last code change from me unless something substantive turns up. One taken, one declined with reasoning.

### Taken

**The `sk` fallback had no test.** Right, and it deserved one more than a normal branch: I added that fallback for a case with **zero live instances**, so untested it proved nothing at all and would have rotted into unreachable code. `test_ranks_a_row_whose_id_lives_only_in_its_sort_key` gives the PRD page a row carrying only `sk`/`created_at`, made the NEWEST of its type — so if the fallback is removed, that row drops out of the ranking and an older document silently wins, which is what fails. Backed by a new mutation (M6: collapse `_document_id_of` to `item.get('document_id')`), caught.

### Declined, and this one is a real disagreement rather than a deferral

**`defaultValue` on `feedbackRebased`.** Omitted on purpose. The convention it's inconsistent with is the one I'd argue is wrong, and `usePrototypeBuild.ts` already documents why at its `CONFIRM_MESSAGE` const:

> No `defaultValue`s, matching the rest of this feature: the test harness loads the real `en` catalogue, so a key that is missing or has moved renders its raw path and fails a test. A `defaultValue` hides exactly that, which is how buttons once shipped announcing `editForm` to assistive tech.

A `defaultValue` makes a missing key invisible — the test passes, the UI reads fine in English, and the seven other locales silently fall back forever. The `defaultValue` calls in that same modal are older code; I'd rather not add a ninth instance of the pattern that previously shipped a real accessibility defect. Happy to be overruled, but it should be a deliberate choice either way rather than local consistency winning by default.

The other two standing nits (404-vs-400, `base_prototype_id`) remain deliberate declines as described in round 2.

### Verification

Frontend **2363 passed**, backend **1504 passed**, ESLint clean at `--max-warnings 0`, `tsc -b` clean, ruff delta zero. Mutation probes **19/19 caught** (11 backend, 8 frontend).

### Where this stands

Three rounds, and the arc is worth stating plainly: round 1 found a genuine bug (a deleted source made a prototype permanently unrevisable) and a design decision I had got wrong and talked myself out of (the unprojected read). Round 2 found a real behaviour change from that fix plus a fixture landmine. Round 3 found two nits. That's convergence, and the substantive findings were all in the first two rounds.

The change is also deployed and live-verified as of before this PR opened — an aimed build reading the oldest PRD and PR/FAQ, the eight trust-boundary cases all 4xx with no job created, and the UI driven in a real browser. From my side this is ready; the remaining calls are the two nits I declined, which are yours.

perrozzi:

## Follow-up review

The only change since the last round is one new backend test, `test_ranks_a_row_whose_id_lives_only_in_its_sort_key` in `voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py`, which closes the untested `sk` fallback nit I raised. The fixture is wired correctly for it — the legacy row is the newest by `created_at` and carries no `document_id`, so dropping the fallback makes an older PRD win and the test fails on both the prompt content and `source_prd_id`. Clean.

### Resolved
- `_document_id_of`'s `sk` fallback branch is now exercised by a test that fails if the branch is removed.

### Still open (nits, unchanged from last round)
- `_validated_source_id` returns 404 rather than 400 with a machine-readable field.
- `base_prototype_id` remains unvalidated alongside its two validated siblings.
- `t('documents.prototype.feedbackRebased')` in `DocumentsTab.tsx` still omits `defaultValue`, unlike its sibling `t()` calls.

No new findings.

---

PR #321, opened 2026-08-12
fix(documents): fail loudly when base_prototype_id resolves to nothing

perrozzi:
## Summary

`api_build_prototype` accepts three client-supplied document ids and validated two of them.
`source_prd_id` and `source_prfaq_id` go through `_validated_source_id`; `base_prototype_id` was copied
into `doc_config` straight from the request body. In the generator, an unresolvable base was fetched with a
single `get_item`, and when the key held nothing the prior-HTML variable stayed empty, the
`EXISTING PROTOTYPE (revise this)` block was dropped from the prompt, and the job completed as a **fresh
generation** — writing a document labelled a revision (it carries `revised_from_id`) that the model
produced without ever seeing the prototype it supposedly revises. The build succeeded, billed a
multi-minute Bedrock call, and nothing in the output revealed the base was missing.

One handler therefore failed loudly for one kind of missing source and degraded silently for another.

### Production changes (two files, ~40 lines)

- **`lambda/api/projects_handler.py`** — `base_prototype_id` now resolves through the existing
  `_validated_source_id` helper with the `PROTOTYPE#` prefix. An id naming no prototype in this project is
  a **404 naming the field**; a non-string or over-long value is a **400 naming the field**. Neither
  creates a job row or invokes the generator. The call sits in the same `doc_config` literal as the other
  two, so it runs **before** `create_job`, unconditionally — not gated on `feedback`.
- **`lambda/jobs/document_generator/handler.py`** — a new `_base_prototype` helper resolves the id and
  **raises** on an absent item, in the same shape `_source_document` already uses for the other two ids.
  It is called outside the `if feedback:` block, so an id that is sent is an id that is claimed.

The two checks are deliberately redundant, for the reason `_validated_source_id`'s own docstring already
gives: the boundary check is what keeps an unresolvable id from creating a job that bills a long Bedrock
call in order to fail, and the generator raise is what makes the failure loud if a job reaches the
generator by any other path.

### What deliberately did not change

- **A prototype that resolves but holds neither `prototype_url` nor `content`** builds with no prior-HTML
  block and does not raise. An empty prior-HTML string and an id that resolved to nothing are two different
  conditions, and only the second is an error — keying the raise on empty HTML would break revising an older
  prototype whose stored content is legitimately empty. The error condition is the **absent item**.
- **The S3-then-inline fallback**, including the `try`/`except` around the S3 read. A failed S3 read is not
  an unresolvable id and keeps degrading to the item's inline `content`.
- `_source_document`, `_document_by_id`, `_newest_document_id`, and the handling of `source_prd_id` /
  `source_prfaq_id` — already correct, untouched.
- `MAX_SOURCE_DOCUMENT_ID_LEN` keeps its value of 256.
- Absent, null and blank `base_prototype_id` all keep meaning "this build is not a revision".

## Tests

Extended the two existing files. Each new test's fixture is built so it is capable of failing:

**`lambda/api/test/test_build_prototype_sources.py`** (+11 tests, 15 → 26)

The shared `build_prototype` fixture now answers reads on the **whole composite key** rather than on `sk`
alone, and holds a prototype in a *second* project. Keyed on `sk` only, "belongs to another project" and
"does not exist" would be the same table, and the test could not tell a dropped partition key apart from a
working one.

- an unknown prototype → 404 **and `config is None`**. The `config is None` assertion is the load-bearing
  one: a check placed after `create_job` still returns 404 while having created the job row and, with it,
  the billable build the check exists to prevent.
- a prototype that exists in a *different* project → 404 (the fixture that fails if `pk` is dropped), also
  asserted on the keys actually built.
- a **PRD id** supplied as `base_prototype_id` → 404 (the fixture that fails if the `PROTOTYPE#` prefix is
  dropped).
- an unknown id → 404 **even with no `feedback` in the body**.
- non-string (list / object / number / bool) and over-long → 400 naming the field, no job, and the absurd id
  never reaches a key.
- absent / null / blank / whitespace → 200, `base_prototype_id is None`, generator invoked.

**`lambda/jobs/document_generator/test/test_prototype_sources.py`** (+9 tests, 11 → 20)

Every fixture in the new class supplies enough table responses for **both** source lookups to succeed, so a
silent-fallback implementation runs to completion. Without that, the test would pass because the run ended
for want of a query response rather than because of the raise — passing under exactly the behaviour it
exists to reject. (The file's existing `test_an_unresolvable_id_fails_instead_of_falling_back_to_the_newest`
documents getting this wrong once already.)

- a supplied id resolving to no item raises and the job fails; no prompt built, no `put_item`.
- the recorded failure message **names `base_prototype_id`** (and the id), asserted on what reaches the jobs
  table — without the field name an operator cannot tell which of the three ids did not resolve.
- it fails with no `feedback` sent.
- the id only ever reaches `sk`, under `PROTOTYPE#`, never `pk`.
- an item that resolves with **neither** `prototype_url` nor `content` completes, produces **no**
  `EXISTING PROTOTYPE` block, and still records `revised_from_id`.
- an item whose **S3 read raises** falls back to inline `content` and does not fail the job.
- null / blank / whitespace build normally and issue no `PROTOTYPE#` read at all.

### Revert check

With **both** production hunks reverted and the new tests kept:

- `test_build_prototype_sources.py` — **11 failed, 15 passed.** The failures are every test in
  `TestBasePrototypeIdIsCheckedLikeTheOtherTwo` except `test_a_named_prototype_reaches_the_generator`
  (an id that resolves is passed through either way) and the `absent`/`null` cases of
  `test_not_naming_a_base_prototype_still_builds` (`None` in, `None` out either way). Notably
  `test_not_naming_a_base_prototype_still_builds[blank]` and `[whitespace]` fail too, on
  `assert '   ' is None` — the un-normalised path forwarded the raw blank string.
- `test_prototype_sources.py` — **4 failed, 16 passed:**
  `test_an_unresolvable_base_prototype_fails_the_job`, `test_the_failure_names_the_field`,
  `test_it_fails_even_when_no_feedback_was_sent`, `test_a_base_prototype_is_only_ever_read_under_this_project`.
  The three "keeps its current behaviour" tests (empty-HTML item, failed S3 read, unnamed base) pass in both
  directions, which is the point of them.

Both hunks restored → all green.

## Build and test results

Commands run from `voc-datalake` with `.venv/bin/python -m pytest`, one directory at a time, each well under
the three-minute window:

| Command | Result |
|---|---|
| `mise run build` | **no such task** — `mise ERROR no tasks defined in <repo root>`. There is no mise task file in this repo; the setup notes flag the same thing and say build gating is inert. |
| `pytest lambda/api/test` | **540 passed** |
| `pytest lambda/jobs` | **95 passed** |
| `pytest lambda/api/test/test_build_prototype_sources.py` | **26 passed** (was 15) |
| `pytest lambda/jobs/document_generator/test/` | **93 passed**, incl. `test_prototype_sources.py` **20 passed** (was 11) |
| `ruff check` on the four touched files | **13 findings, byte-identical to the same command on `development`** — 9 `BLE001`, 3 `I001`, 1 `RUF100`, all pre-existing and none in the added lines. Compared before/after via `git stash`. |

Note on the environment: `voc-datalake/.venv` did not exist in the container, so I created it from
`requirements-dev.txt` plus the two layer requirements files (`lambda/layers/{ingestion,processing}-deps`)
as `requirements-dev.txt`'s header instructs — without the production deps every API test errors at import
on `Tracer()`. `.venv` is untracked and not part of this diff.

Frontend/vitest legs were **not** run: `npx vitest` in `voc-datalake/frontend` fails with
`npm error code ECOMPROMISED / Lock compromised` in this container. No frontend or locale file is touched by
this diff, so nothing here should affect them.

`lint:python` is already failing on `development` for reasons unrelated to this work, and is not addressed
here.

## Decisions made

- **A helper rather than an inline `if`.** The lookup moved into `_base_prototype` so it could be hoisted
  out of `if feedback:` without nesting, and so the "why" comment sits next to the raise the way
  `_source_document`'s does. It reads as the third member of that family.
- **`RuntimeError` with the message shape `field: no TYPE document "id" in this project.`** — copied from
  `_source_document` verbatim in shape so the two failures look identical in a job's error field.
- **`if base is not None:`** rather than `if base:` at the call site. An item that resolves but is falsy as a
  dict cannot occur (DynamoDB never returns an empty `Item`), but `is not None` states the intent — the
  distinction between "resolved" and "resolved to something empty" is the whole point of this change.

## Frontend observation (out of scope, not in this diff)

`RevisePrototypeForm` in `frontend/src/pages/ProjectDetail/DocumentsTab.tsx` awaits
`projectsApi.buildPrototype` and, on rejection, renders `e.message` inline and keeps the form open with the
typed feedback intact — so a 404 does surface to the user rather than being swallowed, and the retry does
not cost them retyping. But `fetchApi` in `frontend/src/api/client.ts` throws
`new Error('API Error: ' + response.status)`, discarding the response body — so the user sees
`API Error: 404`, not the server's `base_prototype_id: no such document in this project`. The reachable path
is narrow (the revision form sends the id of a prototype the user just clicked, so a 404 means it was
deleted between render and click), and a generic 404 arguably reads fine there. Surfacing the server's
message generally would be a change to `fetchApi` affecting every caller, which is why it is noted here
rather than done.

## Scope

Frontend and locale catalogues untouched, as specified.

---

## Agent notes

**What went well.** The fix itself was small and the repo had already done the hard thinking: `_validated_source_id` and `_source_document` existed, their docstrings spell out *why* both a boundary check and a generator raise are needed, and the task was essentially "apply the pattern the module already documents to the one field that skipped it." Reusing the helper verbatim meant no new validation logic and no new error shape.

**What was difficult.**
- **Making the tests capable of failing.** Two fixtures needed deliberate construction rather than the obvious version. (a) The API fixture originally keyed reads on `sk` alone; with a single project in the table, "another project's prototype" and "no such prototype" are literally the same fixture, so the test could not detect a dropped `pk`. Rekeying on the composite `(pk, sk)` fixed that. (b) The generator fixture must supply enough `query` responses for *both* source lookups, or a silent-fallback implementation dies for want of a mock response and the test passes for the wrong reason. The existing file's docstrings warn about exactly this trap — they were worth reading before writing anything.
- **Verifying the revert check** rather than asserting it. Reverting both hunks and running the suites turned up two failures I had not predicted: the `blank` and `whitespace` cases of the "still builds" test fail on revert too, because the unchecked path forwarded `'   '` instead of normalising it to `None`. That is a real behaviour difference worth naming in the description, and I would have got it wrong from reasoning alone.
- **Environment setup.** `voc-datalake/.venv` did not exist, and creating it from `requirements-dev.txt` alone is not enough — every API test errors at import on Powertools `Tracer()` until the two layer requirements files are also installed. `requirements-dev.txt`'s header comment says so; worth reading first.

**Repo conventions discovered.**
- **Comments carry the reasoning, at length.** Production functions and test classes alike explain *why* a shape was chosen and *what breaks* if it is changed, often naming the review round or measurement that produced the decision ("0 of 6 when measured", "found in review round 2"). A terse patch would be stylistically out of place here. Test docstrings in particular explain which mutation each test is designed to catch.
- **Trust boundaries are enforced by the DynamoDB key, not by explicit checks.** `pk` is the project and `sk` is `{TYPE}#{id}`, so ownership and type isolation come free and are asserted *on the key that gets built* rather than on an outcome — because with one mocked table, an outcome-based assertion is a tautology. Several tests say so explicitly.
- **Redundant checks are intentional and documented as such**, with each layer's distinct job stated (boundary = don't bill for a failure; generator = don't fail silently).
- Python tests live in `test/` beside the code, `pytest.ini` at `voc-datalake/` sets `--cov`; `--no-cov` makes scoped runs fast. Commit messages are conventional-commit with a module scope, bodies explaining the failure mode in prose.

**Suggestions for future tasks here.**
1. **`mise run build` does not exist** — there is no mise task file at the repo root. The real gates are the per-leg npm scripts (`lint:python`, `test:backend`, `typecheck:*`). `npm run check` chains with `&&`, so the first failing leg hides the rest; run legs individually.
2. **`npx` is broken in this container** (`npm error code ECOMPROMISED / Lock compromised`), so vitest/tsc legs cannot be run. Any task touching `frontend/` or `lambda/stream/` will need that resolved, or will have to ship without frontend test evidence.
3. **`lint:python` fails on `development`** (9 `BLE001` blind-except findings in these two files alone, plus `I001`/`RUF100`). Comparing `ruff check --statistics` on the touched files before and after via `git stash` is the only way to show a change added nothing — worth doing routinely on this repo.
4. **A latent asymmetry remains nearby**: `revised_from_id` and `revision_feedback` are written only `if feedback:`, so a build that supplies `base_prototype_id` with no feedback now validates the id but records no revision relation. That is pre-existing and arguably correct ("no feedback means not a revision"), but it means `base_prototype_id` can be validated and then dropped on the floor. Worth an explicit product decision rather than leaving it implicit.

### DIFF
--- a/voc-datalake/lambda/api/projects_handler.py
+++ b/voc-datalake/lambda/api/projects_handler.py
@@ -629,7 +629,15 @@ def api_build_prototype(project_id: str):
         # Optional feedback-driven regeneration: revise an existing prototype
         # centered on this feedback while still honoring the PRD/PR-FAQ.
         'feedback': body.get('feedback'),
-        'base_prototype_id': body.get('base_prototype_id'),
+        # The prototype being revised is a client-supplied id like the two below,
+        # and it is checked the same way — whenever it is supplied, whether or not
+        # `feedback` came with it. Unchecked, an id naming no prototype produced a
+        # document labelled a revision (it carries `revised_from_id`) that the
+        # model built without ever seeing the prototype it supposedly revises,
+        # after a billed multi-minute Bedrock call.
+        'base_prototype_id': _validated_source_id(
+            project_id, 'PROTOTYPE#', body.get('base_prototype_id'), 'base_prototype_id',
+        ),
         # Optional aiming: build from THESE documents instead of the newest of
         # each type. Validated before the job exists, so a bad id costs a 4xx
         # rather than a billable build that fails minutes later.
--- a/voc-datalake/lambda/api/test/test_build_prototype_sources.py
+++ b/voc-datalake/lambda/api/test/test_build_prototype_sources.py
@@ -1,10 +1,11 @@
 """
 POST /projects/{id}/build-prototype — the source-document trust boundary (U25).
 
-A build can name the PRD and PR/FAQ it should read. Those ids arrive from the
-client and their content is read straight into a Bedrock prompt, so an id that
-resolved outside this project would pull another project's document into this
-project's generation.
+A build can name the PRD and PR/FAQ it should read, and the prototype a
+revision is built on. Those ids arrive from the client and the documents they
+name are read straight into a Bedrock prompt, so an id that resolved outside
+this project would pull another project's document into this project's
+generation.
 
 Ownership and type are enforced by the key rather than by a check: `pk` is the
 project and `sk` is `{TYPE}#{id}`. This file pins that, plus the reason the
@@ -17,23 +18,39 @@ from unittest.mock import MagicMock, patch
 import pytest
 
 PROJECT = 'proj_1'
+OTHER_PROJECT = 'proj_2'
 PATH = f'/projects/{PROJECT}/build-prototype'
 
 
 @pytest.fixture
 def build_prototype(api_gateway_event, lambda_context):
     """
-    Call the endpoint with a projects table holding one PRD and one PR/FAQ.
+    Call the endpoint with a projects table holding one PRD, one PR/FAQ and one
+    prototype in THIS project, plus one prototype in a DIFFERENT project.
 
-    Returns (response, job_config, table) where `job_config` is the doc_config
-    handed to the generator, or None when no job was created.
+    Reads are answered on the whole composite key rather than on `sk` alone.
+    That is what makes "a prototype belonging to another project" a distinct
+    fixture from "no such prototype": keyed on `sk` only, the two would be the
+    same table and the test could not tell a dropped partition key apart from a
+    working one.
+
+    Returns (response, job_config, table, invoke) where `job_config` is the
+    doc_config handed to the generator, or None when no job was created.
     """
     def _call(body):
         table = MagicMock()
-        documents = {'PRD#prd_1': {'document_id': 'prd_1'}, 'PRFAQ#prfaq_1': {'document_id': 'prfaq_1'}}
+        documents = {
+            (f'PROJECT#{PROJECT}', 'PRD#prd_1'): {'document_id': 'prd_1'},
+            (f'PROJECT#{PROJECT}', 'PRFAQ#prfaq_1'): {'document_id': 'prfaq_1'},
+            (f'PROJECT#{PROJECT}', 'PROTOTYPE#proto_1'): {'document_id': 'proto_1'},
+            # Real prototype, wrong project. Only reachable by a lookup that
+            # dropped `pk`.
+            (f'PROJECT#{OTHER_PROJECT}', 'PROTOTYPE#proto_other'): {'document_id': 'proto_other'},
+        }
 
         def get_item(Key=None, **kwargs):
-            item = documents.get((Key or {}).get('sk', ''))
+            key = (Key or {})
+            item = documents.get((key.get('pk', ''), key.get('sk', '')))
             return {'Item': item} if item else {}
 
         table.get_item.side_effect = get_item
@@ -149,3 +166,119 @@ class TestUnresolvableIdIsRejectedBeforeAnyCost:
             'x' * 5000 in str(call.kwargs.get('Key', {}))
             for call in table.get_item.call_args_list
         )
+
+
+class TestBasePrototypeIdIsCheckedLikeTheOtherTwo:
+    """
+    `base_prototype_id` — the prototype a revision is built on.
+
+    It was the one of the three client-supplied ids that reached `doc_config`
+    unchecked. An id naming nothing then produced a build that ran to completion
+    as a FRESH generation — no `EXISTING PROTOTYPE (revise this)` block in the
+    prompt — and saved a document labelled a revision (`revised_from_id`) that the
+    model never saw the base of, after a billed multi-minute Bedrock call.
+    """
+
+    def test_a_named_prototype_reaches_the_generator(self, build_prototype):
+        response, config, _table, invoke = build_prototype(
+            {'feedback': 'Make it admin-facing', 'base_prototype_id': 'proto_1'},
+        )
+
+        assert response['statusCode'] == 200
+        assert config['base_prototype_id'] == 'proto_1'
+        assert invoke.call_args.args[1]['doc_config']['base_prototype_id'] == 'proto_1'
+
+    def test_an_unknown_prototype_is_a_404_and_starts_no_job(self, build_prototype):
+        """The load-bearing assertion is `config is None`: a check that ran AFTER
+        `create_job` would still return 404 here while having created the job row
+        and, with it, the billable build this check exists to prevent."""
+        response, config, _table, invoke = build_prototype(
+            {'feedback': 'Make it admin-facing', 'base_prototype_id': 'proto_nope'},
+        )
+
+        assert response['statusCode'] == 404
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+
+    def test_another_projects_prototype_is_a_404(self, build_prototype):
+        """`proto_other` is a real prototype, in `proj_2`. This is the fixture that
+        fails if the partition key is ever dropped from the lookup — without it,
+        "belongs to someone else" and "does not exist" are the same table."""
+        response, config, table, invoke = build_prototype(
+            {'feedback': 'Make it admin-facing', 'base_prototype_id': 'proto_other'},
+        )
+
+        assert response['statusCode'] == 404
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+        # Asserted on the key as well: the id never addresses another partition.
+        keys = [call.kwargs.get('Key', {}) for call in table.get_item.call_args_list]
+        assert keys, 'expected at least one keyed read'
+        assert {k.get('pk') for k in keys} == {f'PROJECT#{PROJECT}'}
+
+    def test_a_prd_id_offered_as_a_base_prototype_is_a_404(self, build_prototype):
+        """`prd_1` is a real document in this project, just not a prototype. The
+        `sk` prefix is what separates the two, so this is the fixture that fails if
+        the `PROTOTYPE#` prefix is ever dropped from the lookup."""
+        response, config, _table, invoke = build_prototype(
+            {'feedback': 'Make it admin-facing', 'base_prototype_id': 'prd_1'},
+        )
+
+        assert response['statusCode'] == 404
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+
+    def test_an_unknown_prototype_is_a_404_even_without_feedback(self, build_prototype):
+        """The check does not hang off `feedback`. An id that is sent is an id that
+        is claimed, so it is resolved whenever it is supplied."""
+        response, config, _table, invoke = build_prototype({'base_prototype_id': 'proto_nope'})
+
+        assert response['statusCode'] == 404
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+
+    @pytest.mark.parametrize('value', [
+        pytest.param(['proto_1'], id='list'),
+        pytest.param({'id': 'proto_1'}, id='object'),
+        pytest.param(7, id='number'),
+        pytest.param(True, id='bool'),
+    ])
+    def test_a_non_string_base_prototype_id_is_a_400(self, build_prototype, value):
+        response, config, _table, invoke = build_prototype({'base_prototype_id': value})
+
+        assert response['statusCode'] == 400
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+
+    def test_an_absurdly_long_base_prototype_id_is_a_400_not_a_500(self, build_prototype):
+        response, config, table, invoke = build_prototype({'base_prototype_id': 'x' * 5000})
+
+        assert response['statusCode'] == 400
+        assert 'base_prototype_id' in json.loads(response['body'])['error']
+        assert config is None
+        invoke.assert_not_called()
+        # Rejected before the key was ever built.
+        assert not any(
+            'x' * 5000 in str(call.kwargs.get('Key', {}))
+            for call in table.get_item.call_args_list
+        )
+
+    @pytest.mark.parametrize('body, label', [
+        pytest.param({}, 'absent', id='absent'),
+        pytest.param({'base_prototype_id': None}, 'null', id='null'),
+        pytest.param({'base_prototype_id': ''}, 'blank', id='blank'),
+        pytest.param({'base_prototype_id': '   '}, 'whitespace', id='whitespace'),
+    ])
+    def test_not_naming_a_base_prototype_still_builds(self, build_prototype, body, label):
+        """Every build that is not a revision sends one of these, and all of them
+        must keep meaning "not a revision" rather than "prototype ''"."""
+        response, config, _table, invoke = build_prototype({'title': 'Prototype', **body})
+
+        assert response['statusCode'] == 200, label
+        assert config['base_prototype_id'] is None
+        invoke.assert_called_once()
--- a/voc-datalake/lambda/jobs/document_generator/handler.py
+++ b/voc-datalake/lambda/jobs/document_generator/handler.py
@@ -533,6 +533,37 @@ def _source_document(
     return doc
 
 
+def _base_prototype(projects_table, project_id: str, base_prototype_id: str) -> dict | None:
+    """
+    The prototype a revision is built on: the one the request named, or None when
+    it named none.
+
+    A named id that does not resolve RAISES, in the same shape and for the same
+    reason as `_source_document`. Left silent, the build carried on as a fresh
+    generation with no `EXISTING PROTOTYPE (revise this)` block, then saved a
+    document labelled a revision — it carries `revised_from_id` — that the model
+    produced without ever seeing the prototype it supposedly revises. Nothing in
+    the output said so, and the failure cost a full Bedrock call to reach.
+
+    The error condition is the ABSENT ITEM, not empty HTML. A prototype whose
+    stored content is legitimately empty resolved fine and stays buildable — see
+    the caller, which asks for no prior-HTML block in that case.
+
+    Resolved whenever an id is supplied, regardless of whether `feedback` came
+    with it: an id that is sent is an id that is claimed.
+    """
+    if not base_prototype_id:
+        return None
+    item = projects_table.get_item(
+        Key={'pk': f'PROJECT#{project_id}', 'sk': f'PROTOTYPE#{base_prototype_id}'},
+    ).get('Item')
+    if not item:
+        raise RuntimeError(
+            f'base_prototype_id: no PROTOTYPE document "{base_prototype_id}" in this project.'
+        )
+    return item
+
+
 def _generate_prototype(ctx, projects_table, project_id: str, job_id: str, doc_config: dict) -> dict:
     """
     Build a self-contained, offline-first HTML prototype from the latest PRD and
@@ -583,14 +614,15 @@ def _generate_prototype(ctx, projects_table, project_id: str, job_id: str, doc_c
     # revises it (e.g. "switch to an admin-facing view") rather than starting over.
     feedback = (doc_config.get('feedback') or '').strip()
     base_prototype_id = (doc_config.get('base_prototype_id') or '').strip()
+    # Resolved OUTSIDE the `if feedback:` below so an unresolvable id fails the
+    # job whether or not feedback was sent alongside it. Absent, null or blank
+    # still means "this build is not a revision".
+    base = _base_prototype(projects_table, project_id, base_prototype_id)
     feedback_section = ''
     system_prompt = PROTOTYPE_HTML_SYSTEM_PROMPT
     if feedback:
         prior_html = ''
-        if base_prototype_id:
-            base = projects_table.get_item(
-                Key={'pk': f'PROJECT#{project_id}', 'sk': f'PROTOTYPE#{base_prototype_id}'}
-            ).get('Item') or {}
+        if base is not None:
             # New (S3-only) prototypes have no `content` field — read the HTML
             # back from S3 via prototype_url instead. Old (pre-migration)
             # prototypes still have `content` inline in DynamoDB; fall back to
--- a/voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py
+++ b/voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py
@@ -15,6 +15,13 @@ Two properties, both of which were previously unpinned:
    newest would produce a prototype built from a document the user did not
    choose, which no part of the output would reveal.
 
+3. **The same holds for the third client-supplied id, `base_prototype_id`.** It
+   used to be the asymmetry: an id resolving to nothing left the prior HTML
+   empty, dropped the `EXISTING PROTOTYPE (revise this)` block, and completed as
+   a fresh generation — saving a document labelled a revision that the model
+   never saw the base of. One handler failed loudly for one kind of missing
+   source and silently for another.
+
 Every fixture here holds at least TWO documents of a type with different
 `created_at`, because a single-document fixture cannot distinguish "picked the
 right PRD" from "picked the only PRD".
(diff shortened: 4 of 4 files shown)


perrozzi:

This PR closes an asymmetry in the prototype-build path: `base_prototype_id` now goes through the same `_validated_source_id` boundary check as `source_prd_id`/`source_prfaq_id` in `lambda/api/projects_handler.py`, and a new `_base_prototype` helper in `lambda/jobs/document_generator/handler.py` raises instead of silently degrading a revision into a fresh generation. The production change is small, keeps the existing key shape (`pk=PROJECT#…`, `sk=PROTOTYPE#…`) so ownership/type stay enforced by the key, and is backed by regression tests on both sides with a documented revert check. Overall this looks clean and in line with repo patterns; I have one behavioural question and a few nits.

### Should fix

- **The "labelled a revision without seeing the base" failure mode may still exist when `feedback` is empty.** In `lambda/jobs/document_generator/handler.py` the lookup was hoisted out of `if feedback:`, but the *use* of `base` (prior-HTML block, `EXISTING PROTOTYPE (revise this)`) is still inside it. So a request with a **resolvable** `base_prototype_id` and no feedback fetches the item, discards it, builds a fresh prompt — and, if `revised_from_id` is derived from `doc_config['base_prototype_id']` unconditionally (as the PR description implies: "it carries `revised_from_id`"), still saves a document labelled a revision the model never saw. The new tests only cover the *unresolvable* id in the no-feedback case (`test_it_fails_even_when_no_feedback_was_sent`). Please either add a test pinning the resolvable-base/no-feedback behaviour, or handle it (e.g. include the prior-HTML block whenever `base is not None`, or refuse `base_prototype_id` without `feedback` at the boundary). If `revised_from_id` is already gated on `feedback`, a short comment saying so would close the question.

### Nits

- `lambda/jobs/document_generator/handler.py`: when `feedback` is empty, `_base_prototype` performs a DynamoDB read whose result is thrown away — it is doing double duty as a validator. Given the API boundary already validated the id, consider naming/structuring that intent explicitly (or noting it in the docstring) so a future reader does not "optimise" the call back inside `if feedback:`.
- `test_the_failure_names_the_field` asserts against `str(mock_jobs_table.update_item.call_args_list)`, which will also pass if the id appears anywhere in any call (e.g. a key). Asserting on the specific `ExpressionAttributeValues` entry for the error field would be tighter.
- The 7-line explanatory comment above the `_validated_source_id` call in `projects_handler.py` largely restates the helper's own docstring; two lines ("checked whenever supplied, before `create_job`, not gated on `feedback`") would carry the same weight.
- No test asserts the API-side happy-path key shape for `base_prototype_id` directly; `test_another_projects_prototype_is_a_404` covers `pk`, but a positive assertion that `sk == 'PROTOTYPE#proto_1'` would pin the prefix on the success path too.
- Out of scope but worth a follow-up issue: as the description notes, `fetchApi` in `frontend/src/api/client.ts` discards the response body, so the new field-naming 404 never reaches the user. If that is ever surfaced, it needs i18n keys for all locales.

---

voc-datalake/lambda/api/projects_handler.py
```diff
@@ -629,7 +629,15 @@ def api_build_prototype(project_id: str):
...
         # centered on this feedback while still honoring the PRD/PR-FAQ.
         'feedback': body.get('feedback'),
-        'base_prototype_id': body.get('base_prototype_id'),
+        # The prototype being revised is a client-supplied id like the two below,
+        # and it is checked the same way — whenever it is supplied, whether or not
+        # `feedback` came with it. Unchecked, an id naming no prototype produced a
+        # document labelled a revision (it carries `revised_from_id`) that the
+        # model built without ever seeing the prototype it supposedly revises,
+        # after a billed multi-minute Bedrock call.
+        'base_prototype_id': _validated_source_id(
```
perrozzi, line 638: **Type**: issue
**Severity**: minor
**Title**: A `base_prototype_id` sent without `feedback` is now validated, can 404, and is then dropped on the floor -- with no test pinning that

The check correctly does not hang off `feedback`, and `test_an_unknown_prototype_is_a_404_even_without_feedback` pins that for the unresolvable case. But the *resolvable* case with no feedback is untested, and I verified what it does on this branch with a throwaway test: `revised_from_id` is absent from the saved item, no `EXISTING PROTOTYPE` block is built, and S3 is never read. So the id costs a boundary DynamoDB read plus a second read in the generator, exposes the caller to a 404 for an id that has no effect on the output, and then vanishes.

To be clear about the earlier automated review's stronger claim: it does **not** hold. `item['revised_from_id']` is written inside `if feedback:` (handler.py:723-728), so nothing is mislabelled a revision. This is a smaller point -- an unpinned behaviour and a wasted read, on a shape the frontend never sends (`PrototypeFeedbackButton.onSubmit` returns early when `feedback.trim() === ''`).

The author's own notes flag this as a latent asymmetry deserving an explicit product decision. The cheapest way to make the decision explicit is a test.

**Proposed fix**: Add a test asserting the current contract for a **resolvable** `base_prototype_id` with no `feedback` -- no `revised_from_id`, no `EXISTING PROTOTYPE` block, no S3 read -- so a future reader cannot mistake "validated then discarded" for a bug and cannot silently change it. Alternatively make the boundary reject `base_prototype_id` without `feedback` as a 400, but that is a behaviour change and the test is the lower-risk option for this PR.

**AI prompt**: In voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py, inside `TestBasePrototypeOfARevision`, add a test named something like `test_a_resolvable_base_without_feedback_is_not_recorded_as_a_revision`. Wire the fixture with `prd_pages=[{'Items': [PRD_NEW]}]`, `prfaq_pages=[NO_DOCUMENTS]` and `documents={'PROTOTYPE#proto_1': {**PROTOTYPE_S3, 'content': 'INLINE prior prototype'}}`, then `_run(sample_job_event, lambda_context, base_prototype_id='proto_1')` with NO feedback. Assert that the build succeeds, that `'revised_from_id' not in _saved(mock_dynamodb)`, that `'EXISTING PROTOTYPE' not in _prompt(mock_converse)`, and that `mock_s3.get_object` was not called. Write a docstring explaining that `revised_from_id` is written only inside `if feedback:`, so an id supplied without feedback is validated and then intentionally unused -- and that this test exists so that contract is explicit rather than incidental.

voc-datalake/lambda/jobs/document_generator/handler.py
```diff
@@ -533,6 +533,37 @@ def _source_document(
...
+    the caller, which asks for no prior-HTML block in that case.
+
+    Resolved whenever an id is supplied, regardless of whether `feedback` came
+    with it: an id that is sent is an id that is claimed.
+    """
+    if not base_prototype_id:
+        return None
+    item = projects_table.get_item(
+        Key={'pk': f'PROJECT#{project_id}', 'sk': f'PROTOTYPE#{base_prototype_id}'},
+    ).get('Item')
```
perrozzi, line 559: **Type**: issue
**Severity**: minor
**Title**: `_base_prototype` re-implements `_document_by_id` instead of calling it, duplicating both the key construction and the error-message shape

Lines 557-559 are byte-for-byte what `_document_by_id(projects_table, project_id, 'PROTOTYPE#', base_prototype_id)` already does (handler.py:411-426), including the `.get('Item') or None` normalisation. That helper's docstring is where this module explains *why* a client-supplied id can be passed without a separate ownership check -- "there is no key this function can build that reaches outside `project_id`" -- so a second hand-rolled `get_item` is the one place that reasoning is not centralised. There are now two f-strings in this file that build a `PROTOTYPE#` sort key from client input.

The message is a second, smaller duplication: `_source_document` derives the type name with `sk_prefix.rstrip("#")`, while this hardcodes `PROTOTYPE`. The PR aims for the two failures to look identical in a job's error field, and they do today -- but only by coincidence of two independent literals.

**Proposed fix**: Have `_base_prototype` delegate: `item = _document_by_id(projects_table, project_id, 'PROTOTYPE#', base_prototype_id)`, keeping the `if not item: raise` immediately after. Optionally derive the type name in the message from the same prefix constant so the two error shapes cannot drift. No behaviour change, no test change -- the existing key-shape assertion in `test_a_base_prototype_is_only_ever_read_under_this_project` continues to pin the key.

**AI prompt**: In voc-datalake/lambda/jobs/document_generator/handler.py, refactor `_base_prototype` so it calls the existing `_document_by_id(projects_table, project_id, 'PROTOTYPE#', base_prototype_id)` helper instead of building the DynamoDB key and calling `get_item` itself -- `_document_by_id` already performs the identical read and the identical `or None` normalisation, and its docstring is where this module documents why a client-supplied id is safe to pass with no separate ownership check. Keep the `if not item: raise RuntimeError(...)` immediately afterwards and keep the message shape identical to `_source_document`'s (`field: no TYPE document "id" in this project.`), deriving the `PROTOTYPE` token from the same `'PROTOTYPE#'` prefix string rather than hardcoding it a second time. Behaviour must not change; run `pytest lambda/jobs/document_generator/test/ --no-cov` and confirm all tests still pass.

voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py
```diff
@@ -323,3 +330,162 @@ def test_naming_neither_source_keeps_the_previous_latest_of_each_behaviour(
...
+        _wire(mock_dynamodb, prd_pages=[{'Items': [PRD_OLD, PRD_NEW]}], prfaq_pages=[NO_DOCUMENTS], documents={})
+        mock_converse.return_value = HTML
+
+        with pytest.raises(Exception, match='Document generation failed'):
+            _run(
+                sample_job_event, lambda_context,
+                feedback='Make it admin-facing', base_prototype_id='proto_does_not_exist',
+            )
+
+        recorded = str(mock_jobs_table.update_item.call_args_list)
```
perrozzi, line 389: **Type**: issue
**Severity**: minor
**Title**: `test_the_failure_names_the_field` asserts against a stringified call list, so it can pass without the error field containing anything

Stringifying `mock_jobs_table.update_item.call_args_list` flattens every argument of every call into one blob, so both assertions pass if `'base_prototype_id'` and `'proto_does_not_exist'` appear *anywhere* in *any* call -- a key, an unrelated attribute, a progress update. This is the one test in the new class whose whole purpose is the content of a specific field ("without the field name an operator cannot tell which of the three ids did not resolve"), and it is the one asserting most loosely. Given how deliberately the fixtures in this file were built to be capable of failing, this assertion is out of step with the rest.

`update_job_status` (lambda/shared/jobs.py:61-101) writes the error into `ExpressionAttributeValues` under a dedicated placeholder, and `job_handler` wraps it as `f'{error_message}: {str(e)[:200]}'` -- so the exact value is reachable and worth asserting on directly. Note the 200-char truncation is another reason to assert on the real field: a longer id could be cut off, and the current assertion would not notice as long as the id appeared elsewhere.

**Proposed fix**: Locate the `update_item` call whose values carry `status == 'failed'`, pull the error string out of its `ExpressionAttributeValues`, and assert `'base_prototype_id' in error` and `'proto_does_not_exist' in error` against that one string.

**AI prompt**: In voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py, tighten `test_the_failure_names_the_field`. Replace `recorded = str(mock_jobs_table.update_item.call_args_list)` with code that finds the specific `update_item` call recording the failure -- iterate `mock_jobs_table.update_item.call_args_list`, read each call's `ExpressionAttributeValues`, and select the one whose status value is `'failed'` -- then extract the error message string from that call's expression values (see `update_job_status` in lambda/shared/jobs.py for the placeholder name used for the error attribute). Assert on that single string that it contains both `'base_prototype_id'` and `'proto_does_not_exist'`. Add a brief docstring note that asserting on the stringified call list would also pass if the id merely appeared in a key or an unrelated attribute, and that `job_handler` truncates the message to 200 characters. Verify the test still passes on this branch and still fails when `_base_prototype`'s raise is removed.

voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py
```diff
@@ -323,3 +330,162 @@ def test_naming_neither_source_keeps_the_previous_latest_of_each_behaviour(
 
         assert 'NEW PRD body' in _prompt(mock_converse)
```
perrozzi, line 331: **Type**: comment
**Title**: The generator-side `_wire` fixture still keys documents on `sk` alone, unlike the API fixture this PR deliberately rekeyed

The API fixture was rekeyed onto the composite `(pk, sk)` in this same PR, with a docstring explaining precisely why: keyed on `sk` only, "belongs to another project" and "does not exist" are the same table and the test cannot detect a dropped partition key. `_wire`'s `documents` map still keys on `sk`, so the generator-side tests have no fixture-level defence against the same mutation.

In practice this is covered -- `test_a_base_prototype_is_only_ever_read_under_this_project` asserts on the **key that gets built** (`{k.get('pk') for k in keys} == {'PROJECT#proj_20250101120000'}`), which is the stronger form and the one this repo's conventions prefer, precisely because with a single mocked table an outcome-based assertion is a tautology. So this is an observation about asymmetry between two sibling fixtures rather than a gap: the two files now defend the same property by different mechanisms, and a reader moving between them may reasonably wonder which is intended.

Not worth changing in this PR. Worth a line in `_wire`'s docstring noting that `pk` is deliberately not part of the fixture key here because the ownership property is asserted on the constructed key instead -- that is the kind of "why this shape, and what breaks if you change it" note the rest of this file already carries.

**AI prompt**: In voc-datalake/lambda/jobs/document_generator/test/test_prototype_sources.py, extend the `_wire` helper's docstring with a short paragraph explaining that its `documents` map is keyed on `sk` alone, deliberately: unlike the API-side fixture in lambda/api/test/test_build_prototype_sources.py (which is keyed on the composite `(pk, sk)` so that "another project's document" is a distinguishable fixture from "no such document"), the generator-side tests pin project ownership by asserting on the DynamoDB key that gets built -- see `test_a_base_prototype_is_only_ever_read_under_this_project` -- rather than on the lookup's outcome. Note that with a single mocked table an outcome-based ownership assertion would be a tautology. Do not change any fixture behaviour or any test.

voc-datalake/lambda/api/test/test_build_prototype_sources.py
```diff
@@ -17,23 +18,39 @@
...
+            (f'PROJECT#{PROJECT}', 'PROTOTYPE#proto_1'): {'document_id': 'proto_1'},
+            # Real prototype, wrong project. Only reachable by a lookup that
+            # dropped `pk`.
+            (f'PROJECT#{OTHER_PROJECT}', 'PROTOTYPE#proto_other'): {'document_id': 'proto_other'},
+        }
 
         def get_item(Key=None, **kwargs):
-            item = documents.get((Key or {}).get('sk', ''))
+            key = (Key or {})
+            item = documents.get((key.get('pk', ''), key.get('sk', '')))
```
perrozzi, line 53: **Type**: good_point
**Title**: Rekeying the fixture onto the composite key is what makes the ownership test able to fail

This two-line change is the most valuable part of the test diff. With reads answered on `sk` alone and a single project in the table, `test_another_projects_prototype_is_a_404` would have passed against an implementation that dropped `pk` from the key entirely -- "belongs to `proj_2`" and "does not exist" would have been indistinguishable fixtures. Seeding `PROJECT#proj_2 / PROTOTYPE#proto_other` and matching on `(pk, sk)` makes the mutation detectable, and the docstring says so, so the next person cannot "simplify" it back. Pairing it with an assertion on the constructed keys rather than only on the status code follows the same convention.

The same instinct shows in `test_a_prd_id_offered_as_a_base_prototype_is_a_404` (a real document in this project, wrong type -- the fixture that fails if the `PROTOTYPE#` prefix is dropped) and in the `config is None` assertion, which is what distinguishes this check from one placed after `create_job` that would return the same 404 having already created the billable job row.

perrozzi (commented):
## Review: `base_prototype_id` now fails loudly

This closes a real asymmetry cleanly. `base_prototype_id` was the one of three client-supplied document ids that reached `doc_config` unchecked, and the failure mode described in the PR body is accurate: an unresolvable id left `prior_html` empty, `prior_block` fell to `''`, and the job completed as a fresh generation. The fix reuses the module's own `_validated_source_id` at the boundary and mirrors `_source_document`'s raise in the generator, so no new validation logic and no new error shape were introduced. The key shape (`pk=PROJECT#...`, `sk=PROTOTYPE#...`) is unchanged, so ownership and type isolation stay enforced by the key rather than by a check -- the convention this module documents at length.

**Verification I ran (not just read):**

| Check | Result |
|---|---|
| `mise run build` | **no such task** -- `mise ERROR no tasks defined in .../voc-datalake`. There is no mise task file in this repo, so the nominal build gate is inert. Confirms the PR author's note. |
| `pytest lambda/api/test lambda/jobs --no-cov` | **635 passed** |
| `pytest lambda/api/test/test_build_prototype_sources.py lambda/jobs/document_generator/test/ --no-cov` | **93 passed** (26 + 93 as claimed) |
| `ruff check` on the 4 touched files | **13 findings** (9 `BLE001`, 3 `I001`, 1 `RUF100`) -- I diffed this against the same files at `HEAD~1` and the counts are **identical**, so this PR adds zero lint findings. None are in added lines. |
| Frontend/vitest | not run -- `npx` fails in this container (`ECOMPROMISED`). No frontend file is touched. |

I also had to build `voc-datalake/.venv` from `requirements-dev.txt` **plus** both layer requirements files, exactly as that file's header says -- without the production deps every API test errors at import on Powertools `Tracer()`.

**I empirically checked the one behavioural question the previous automated review raised as "should fix"** (resolvable `base_prototype_id` + no `feedback` -> is a document mislabelled a revision?). I wrote a throwaway test against this branch. Result: `revised_from_id` is **absent** from the saved item, no `EXISTING PROTOTYPE` block is built, and S3 is never read. `item['revised_from_id']` is written inside `if feedback:` at handler.py:723-728, so **the mislabelling concern does not hold** -- nothing is labelled a revision. What remains is smaller, and I've filed it as a minor finding on the call site: the id is validated (and can 404) and then read-and-discarded, with no test pinning that.

### Findings not tied to a single line

**Type**: comment
**Title**: The "job reaches the generator by another path" justification is concrete, not hypothetical -- and that path is still unvalidated

The PR argues the two checks are deliberately redundant because "the generator raise is what makes the failure loud if a job reaches the generator by any other path." That path exists and is reachable today: `POST /projects/{project_id}/document` (`api_generate_document`, projects_handler.py:242-274) takes `doc_type` from the body with no allowlist, forwards the **raw request body** as `doc_config`, and `is_chain = doc_type in ('prd', 'prfaq')` is False for `build_prototype`, so it lands on the direct async invoke. `handle_job` then dispatches `doc_type == 'build_prototype'` to `_generate_prototype` (handler.py:789). A caller posting `{"doc_type": "build_prototype", "base_prototype_id": "nope"}` to that route bypasses `api_build_prototype` and every `_validated_source_id` call entirely.

That strengthens this PR -- the generator raise is genuinely load-bearing, not defence against a theoretical caller. It is also pre-existing and applies equally to `source_prd_id`/`source_prfaq_id`, so it is out of scope here, but it is worth a follow-up issue: that route bills a Bedrock call to reach a failure the boundary was built to catch. Consider citing it in the PR body as the concrete answer to "why both checks."

**AI prompt**: In voc-datalake/lambda/api/projects_handler.py, `api_generate_document` (POST /projects/{project_id}/document) reads `doc_type` from the request body with no allowlist and forwards the raw body as `doc_config`. Because `is_chain` is only True for 'prd'/'prfaq', a body with `doc_type: 'build_prototype'` is invoked directly on the document generator, reaching `_generate_prototype` with completely unvalidated `base_prototype_id`, `source_prd_id` and `source_prfaq_id` -- bypassing the `_validated_source_id` trust boundary in `api_build_prototype`. Restrict this route to the doc types it is documented to serve (an explicit allowlist, 400 otherwise), and add a test asserting that `doc_type: 'build_prototype'` on that route does not create a job. Do not change `api_build_prototype`.

---

**Type**: good_point
**Title**: Backwards compatibility for legacy prototypes was thought through, and I could not break it

The new 404/raise keys strictly on the **absent item**, not on empty HTML, and both directions are pinned: `test_a_resolved_base_with_no_html_anywhere_still_builds` and `test_a_failed_s3_read_falls_back_to_the_inline_content`. I checked the risk that legacy prototypes might live under a different sort key and so start 404-ing -- `git log -S` on the `PROTOTYPE#` sort-key literal shows the prefix unchanged since `fbeef53`, and `_with_signed_prototype_url` in projects.py confirms both storage generations are `PROTOTYPE#` rows differing only by the presence of `content`. So pre-migration inline prototypes still resolve and still revise. The three "keeps its current behaviour" tests that pass in both directions of the revert check are the right instinct.

---

**Type**: good_point
**Title**: Two fixtures were deliberately built to be capable of failing

Rekeying the API fixture from `sk` alone onto the composite `(pk, sk)` and seeding a prototype in a second project is what makes "belongs to another project" a distinguishable case from "does not exist" -- without it `test_another_projects_prototype_is_a_404` would pass against an implementation that dropped the partition key. Likewise, supplying enough `query` responses for **both** source lookups means a silent-fallback implementation runs to completion instead of dying for want of a mock, which is the trap the file's existing `test_an_unresolvable_id_fails_instead_of_falling_back_to_the_newest` docstring already warns about. The documented revert check (11 failed / 4 failed with the hunks reverted), including the two `blank`/`whitespace` failures the author did not predict and verified rather than reasoned about, is the standard this repo's test suites are held to.

---

**Type**: comment
**Title**: On comment length -- consistent with this repo, not excessive

For the record against the earlier automated review's nit: the 7-line explanatory comment above the `_validated_source_id` call and the long `_base_prototype` docstring are **in keeping with this module's conventions**, where production functions and test classes explain why a shape was chosen and what breaks if it changes (`_document_id_of` cites "0 of 6 when measured"; `_wire` cites "found in review round 2"). A terser patch would read as out of place. I would not trim these.

_(Informed by repository memory: prior work in this repo established that long "why" comments naming the review round or measurement behind a decision are the house style, and that trust boundaries are asserted on the key that gets built rather than on an outcome.)_

### Summary

No blocking issues. Production diff is ~40 lines, reuses existing helpers, adds no lint findings, and the tests are unusually well-constructed. Findings are 3 minor issues (a dropped-on-the-floor read with no test, a duplicated key-construction helper, a loose test assertion) plus context on the unvalidated sibling route.

perrozzi:
## Review summary

**Findings: 3 issues (all minor), 3 comments, 3 good points, 0 questions.** No blocking findings — nothing here rises above `minor`.

| Type | Count | Where |
|---|---|---|
| issue (minor) | 3 | validated-then-discarded read with no test; `_base_prototype` duplicating `_document_by_id`; loose assertion in `test_the_failure_names_the_field` |
| comment | 3 | the unvalidated sibling route `POST /projects/{id}/document`; `_wire` fixture keyed on `sk` alone; comment-length is house style, keep it |
| good_point | 3 | composite-key fixture rekey; legacy-prototype compatibility; fixtures built to be capable of failing |

### Overall assessment

Good PR. The production change is ~40 lines across two files and adds no new validation logic or error shape — it applies `_validated_source_id` and the `_source_document` raise pattern that this module already documents to the one field that skipped them. The sort-key shape is untouched, so ownership and type isolation continue to be enforced by the key rather than by a check, matching the convention the module states explicitly. The three findings are all polish; none affects correctness of the fix.

The test work is the strongest part. Both fixture changes were deliberately constructed to be *capable of failing* — the composite `(pk, sk)` rekey plus a second project makes "belongs to another project" distinguishable from "does not exist," and supplying enough `query` responses for both source lookups stops a silent-fallback implementation from passing by dying for want of a mock. The revert check was executed rather than reasoned about, which is why it caught the two `blank`/`whitespace` failures the author had not predicted.

### Build and test results

| Command | Result |
|---|---|
| `mise run build` | **no such task** — `mise ERROR no tasks defined in .../voc-datalake`. There is no mise task file in this repo; the nominal build gate is inert, confirming the PR author's note. |
| `pytest lambda/api/test lambda/jobs --no-cov` | **635 passed** |
| `pytest lambda/api/test/test_build_prototype_sources.py lambda/jobs/document_generator/test/ --no-cov` | **93 passed** |
| `ruff check` on the 4 touched files | **13 findings** (9 `BLE001`, 3 `I001`, 1 `RUF100`) — **identical counts** to the same files at `HEAD~1`, so this PR adds zero lint findings, and none are in added lines. |
| Frontend / vitest | **not run** — `npx` fails in this container (`ECOMPROMISED`). No frontend file is touched by the diff. |

Environment note for future runs: `voc-datalake/.venv` must be built from `requirements-dev.txt` **plus** `lambda/layers/{ingestion,processing}-deps/requirements.txt`, exactly as that file's header instructs — without the production deps every API test errors at import on Powertools `Tracer()`.

### Areas worth attention before merge

1. **Nothing is required.** All three issues are minor and could equally be follow-ups.
2. **Highest-value single change**: tighten `test_the_failure_names_the_field` to assert on the actual error attribute rather than `str(call_args_list)`. It is the one test whose entire purpose is a specific field's content, and currently the loosest assertion in an otherwise rigorous file.
3. **Cheap correctness-of-intent win**: a test pinning the *resolvable* `base_prototype_id` + no-`feedback` path. I verified empirically that it does **not** mislabel a document as a revision (`revised_from_id` is written only inside `if feedback:`), so the earlier automated review's "should fix" on this does not hold — but the behaviour is unpinned and the author's own notes ask for an explicit decision. A test is the decision.
4. **Follow-up issue, not this PR**: `POST /projects/{id}/document` accepts `doc_type: 'build_prototype'` with no allowlist and forwards the raw body as `doc_config`, reaching `_generate_prototype` with all three ids unvalidated. This bypasses the boundary this PR strengthens — and it is the concrete justification for keeping the generator-side raise, so it is worth naming in the PR body.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01KZV4EKYJZS7GBDMBVYNZBR6J` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-08-12T14:26:15.684Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/321) |
| Duration | 1461.8s |
| Cost | $2.9835 |

perrozzi:

**Why it stopped:** 0 blocking, 0 should_fix, build not regressed — decision=commented; 3 nit, 6 info; build unknown (tolerated); mergeable [iteration 0, 3 nit, 6 info]

| | |
|---|---|
| Outcome | **CONVERGED** |
| Iteration | 0 of 4 |
| Loop spend | $4.77 of $30.00 |
| Goal | # An unresolvable base_prototype_id must fail loudly, like the other two source ids ## The defect api_build_prototype in voc-datalake/lambda/api/projects_handler.py accepts three client-supplied document ids on a prototype build and validates two of them. source_prd_id and source_prfaq_id … |
| Run | [`01KZV3Z9ZG587N6C61B7M6B7Y4`](https://d2wi1pimyj1f69.cloudfront.net/runs/01KZV3Z9ZG587N6C61B7M6B7Y4) |

### Each iteration

|---|---|---|---|---|---|---|
| 0 | 3 nit, 6 info | — | ❌ | ✅ | — | — |

PR #322, opened 2026-08-12
feat(prototypes): let a build read product context and research

perrozzi:
Rung 1 of the prototype visual-grounding ladder (`analysis/FEATURE-prototype-visual-grounding-GITHUB-ISSUE.md`, not yet filed). Closes U25's last open acceptance criterion, which the product owner answered **yes to both, by tick-box**.

Built on top of #321 (both change `api_build_prototype` and `_generate_prototype`), and rebased onto `development` once #321 merged as `5f2172e` — so this is a single commit and the diff is rung 1 alone. Gates re-run after the rebase, unchanged.

## What changes

A prototype build read the PRD and PR/FAQ and nothing else. The project's product context and its research reports reached every other generation path and never reached this one — so the artifact stakeholders actually look at was grounded in the least. Both are now optional inputs, ticked per build on the prototype card.

Three new optional request fields. **All absent reproduces today's prompt byte for byte**, so no existing caller changes behaviour.

## Three decisions that had a plausible alternative

**Controls on the card, not in the confirm dialog.** This is a requirement, not a preference: `confirmKeyFor` deliberately returns no dialog for a project with one PRD, one PR/FAQ and no prototype, so controls placed in the dialog are unreachable for exactly the simplest project. The card also preserves the one-click build.

**Per build, not remembered per project** — the same answer #320 took for the source ids. Answering it differently for two halves of one control would be worse than either answer.

**Research is scoped to `RESEARCH#` with its own field.** It does not reuse the shared document picker and does not join `selected_document_ids`. The shared reference path keeps only the first three of a selection and research sorts last, so a general picker drops precisely the report the user asked for. A research-only field cannot be capped out — nothing of another type is in its candidate set. The cap itself stays a separate issue.

## Trust boundary

The named document's text goes straight into a Bedrock prompt, so every id resolves through the existing `_validated_source_id` under the `RESEARCH#` prefix — `pk` is the project, `sk` is the type, so ownership and document type come from the key rather than a separate check. Two things a single-id check did not need: the **arity bound is checked before the first read**, so an over-long list costs one 400 rather than N reads and then a 400; and **duplicates collapse**, because a repeated id would be read twice and injected twice.

## Revisions inherit rather than re-derive

A revision reads what its base was built with off the base's own recorded derivation. Re-deriving would ground the revision in whatever research exists *now*, so a report created between build and revision would silently join it while the user only asked to change the feedback. `resolveDerivation` is total, so a prototype built before this feature yields no flag and no reports and its revision behaves exactly as today.

## Verification — by delta against a baseline on the untouched tree, not against green

| gate | before | after |
|---|---|---|
| `pytest lambda/jobs/document_generator/test` | 67 | **82** |
| `pytest lambda/api/test` | 540 | **558** |
| `vitest src/pages/ProjectDetail` | 364 | **395** |
| full `vitest run` | — | 2394 passed, 163 files |
| `tsc -b` | clean | clean |
| `eslint src --max-warnings 0` | — | clean |
| `ruff` on the two edited files | 177 / 136 lines | **177 / 136, unchanged** |
| `npm run i18n:check` | **already red**, 2 unrelated causes | red, output **byte-identical** to the pre-change capture |

The two ruff numbers are pre-existing. The bar was "does not grow", not "reaches zero", and no pre-existing failure was opportunistically fixed in here.

**26 mutation probes, 26 caught.** Two survived first and both were real test defects, both worth knowing:

- A test for "unticked research is not read" only covered the key being **absent**. The API always sends `use_research: false` explicitly, so the shape production actually produces was untested — reading the switch as `is not None` passed. Now parametrised over absent / explicit false / null.
- Nothing exercised the hook filtering a selection against the live options, which is what stops a report deleted under an open card being sent to an API that would reject it. Catching it needed a **re-render with the report removed**, which no existing test did.

## Two things I'd flag to a reviewer

**`MAX_SELECTED_RESEARCH_IDS` lives beside `overviewState`, not the request type.** Putting it with the request type broke four unrelated suites: they mock the api module with explicit partial objects, so any new named export there makes every consumer throw. Patching a dozen mocks was the larger diff. A lockstep test reads both copies as source text and asserts exactly one match per file, so a restructure fails loudly rather than vacuously, plus a plausibility bound so a value of 0 or 1 cannot keep it green. The drift it guards is one-sided and quiet: raise the frontend copy and the picker offers a selection the API rejects, after the user has chosen, with nothing said about which report to drop.

**The card description no longer names PRD or PR-FAQ at all**, where it could have kept the names and added the optional inputs. That was done so one regex asserts across all eight catalogues — every locale kept the Latin "PRD / PR-FAQ" verbatim, so a missed catalogue fails language-independently. It is the test shaping the copy, and a reviewer may reasonably want the names back with a per-language assertion instead.

## Known gap, tested and deliberately not papered over

When a revision's inherited research report has since been deleted, it is dropped silently. The existing rebased-feedback note says "the latest of that type will be used instead", which would be untrue here — nothing substitutes for a missing research report. Saying it properly needs a ninth string in eight catalogues, outside this change's criteria.

## Not verified

**No browser has rendered these controls.** The tests render them in jsdom and assert reachability and behaviour, which is not the same thing. Live verification is owed before this is called done.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "Dieses Projekt hat bereits einen Prototyp. Ein weiterer Build erstellt einen neuen und behält den vorhandenen. Fortfahren?",
       "downloadHtml": ".html herunterladen",
       "downloadJson": ".json herunterladen",
+      "extraSources": "Zusätzlich lesen (optional)",
       "feedbackBuilding": "Wird überarbeitet…",
       "feedbackButton": "Mit Feedback überarbeiten",
       "feedbackHeading": "Prototyp mit Feedback überarbeiten",
@@ -99,10 +100,13 @@
       "needsDocs": "Erstellen Sie zuerst ein PRD oder ein PR-FAQ, um die Prototyp-Erstellung zu aktivieren",
       "previewLabel": "Live-Vorschau",
       "rawLabel": "Rohausgabe (Parsen fehlgeschlagen — bitte neu generieren)",
+      "researchLimit": "Höchstens {{max}} Berichte pro Build",
       "sourceLatest": "neueste",
       "sourcePrd": "PRD",
       "sourcePrfaq": "PR/FAQ",
-      "started": "Gestartet — Verfolgen Sie den Fortschritt unter Hintergrundaufgaben."
+      "started": "Gestartet — Verfolgen Sie den Fortschritt unter Hintergrundaufgaben.",
+      "useProductContext": "Produkt-/Servicebeschreibung",
+      "useResearch": "Rechercheberichte ({{total}})"
     },
     "revision": {
       "feedback": "Rückmeldung: {{feedback}}",
@@ -257,7 +261,7 @@
     "productDescription": "Produkt-/Service-Beschreibung",
     "productDescriptionDesc": "Beschreiben Sie das aktuelle Produkt. Fließt in die PRD- und PR/FAQ-Generierung ein.",
     "prototype": "Klickbarer Prototyp",
-    "prototypeDesc": "Erstellen Sie einen klickbaren HTML-Prototyp aus Ihrem PRD / PR-FAQ",
+    "prototypeDesc": "Erstellen Sie einen klickbaren HTML-Prototyp aus den Spezifikationen des Projekts und optional aus Produktbeschreibung und Recherche",
     "remixDocuments": "Dokumente remixen",
     "remixDocumentsDesc": "Dokumente kombinieren und überarbeiten",
     "runResearch": "Recherche starten",
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "This project already has a prototype. Building again adds a new one and keeps the existing one. Continue?",
       "downloadHtml": "Download .html",
       "downloadJson": "Download .json",
+      "extraSources": "Also read (optional)",
       "feedbackBuilding": "Revising…",
       "feedbackButton": "Revise with feedback",
       "feedbackHeading": "Revise prototype with feedback",
@@ -99,10 +100,13 @@
       "needsDocs": "Create a PRD or a PR-FAQ first to enable prototype build",
       "previewLabel": "Live preview",
       "rawLabel": "Raw output (parse failed — please regenerate)",
+      "researchLimit": "At most {{max}} reports per build",
       "sourceLatest": "latest",
       "sourcePrd": "PRD",
       "sourcePrfaq": "PR/FAQ",
-      "started": "Started — track it in Background Jobs."
+      "started": "Started — track it in Background Jobs.",
+      "useProductContext": "Product / service description",
+      "useResearch": "Research reports ({{total}})"
     },
     "revision": {
       "feedback": "Feedback: {{feedback}}",
@@ -257,7 +261,7 @@
     "productDescription": "Product / Service Description",
     "productDescriptionDesc": "Describe the current product. Feeds PRD and PR/FAQ generation.",
     "prototype": "Clickable Prototype",
-    "prototypeDesc": "Build a clickable HTML prototype from your PRD / PR-FAQ",
+    "prototypeDesc": "Build a clickable HTML prototype from your project's specs, and optionally its product description and research",
     "remixDocuments": "Remix Documents",
     "remixDocumentsDesc": "Combine and revise documents into new versions",
     "runResearch": "Run Research",
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "Este proyecto ya tiene un prototipo. Volver a compilar añade uno nuevo y conserva el existente. ¿Continuar?",
       "downloadHtml": "Descargar .html",
       "downloadJson": "Descargar .json",
+      "extraSources": "Leer también (opcional)",
       "feedbackBuilding": "Revisando…",
       "feedbackButton": "Revisar con feedback",
       "feedbackHeading": "Revisar el prototipo con feedback",
@@ -99,10 +100,13 @@
       "needsDocs": "Crea primero un PRD o un PR-FAQ para habilitar la creación del prototipo",
       "previewLabel": "Vista previa en vivo",
       "rawLabel": "Salida sin procesar (falló el análisis — regenera, por favor)",
+      "researchLimit": "Máximo {{max}} informes por compilación",
       "sourceLatest": "más reciente",
       "sourcePrd": "PRD",
       "sourcePrfaq": "PR/FAQ",
-      "started": "Iniciado: síguelo en Tareas en segundo plano."
+      "started": "Iniciado: síguelo en Tareas en segundo plano.",
+      "useProductContext": "Descripción del producto o servicio",
+      "useResearch": "Informes de investigación ({{total}})"
     },
     "revision": {
       "feedback": "Comentarios: {{feedback}}",
@@ -260,7 +264,7 @@
     "productDescription": "Descripción del producto/servicio",
     "productDescriptionDesc": "Describe el producto actual. Alimenta la generación de PRD y PR/FAQ.",
     "prototype": "Prototipo interactivo",
-    "prototypeDesc": "Cree un prototipo HTML interactivo a partir de su PRD / PR-FAQ",
+    "prototypeDesc": "Cree un prototipo HTML interactivo a partir de las especificaciones del proyecto y, opcionalmente, de su descripción de producto e investigación",
     "remixDocuments": "Remezlar Documentos",
     "remixDocumentsDesc": "Combinar y revisar documentos en nuevas versiones",
     "runResearch": "Ejecutar Investigación",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "Ce projet a déjà un prototype. Une nouvelle génération en ajoute un et conserve l’existant. Continuer ?",
       "downloadHtml": "Télécharger le .html",
       "downloadJson": "Télécharger le .json",
+      "extraSources": "Lire aussi (facultatif)",
       "feedbackBuilding": "Révision en cours…",
       "feedbackButton": "Réviser avec des retours",
       "feedbackHeading": "Réviser le prototype avec des retours",
@@ -99,10 +100,13 @@
       "needsDocs": "Créez d’abord un PRD ou un PR-FAQ pour activer la création de prototype",
       "previewLabel": "Aperçu en direct",
       "rawLabel": "Sortie brute (analyse échouée — veuillez régénérer)",
+      "researchLimit": "Au maximum {{max}} rapports par génération",
       "sourceLatest": "la plus récente",
       "sourcePrd": "PRD",
       "sourcePrfaq": "PR/FAQ",
-      "started": "Lancé — suivez la progression dans Tâches en arrière-plan."
+      "started": "Lancé — suivez la progression dans Tâches en arrière-plan.",
+      "useProductContext": "Description du produit ou service",
+      "useResearch": "Rapports de recherche ({{total}})"
     },
     "revision": {
       "feedback": "Retour : {{feedback}}",
@@ -260,7 +264,7 @@
     "productDescription": "Description du produit/service",
     "productDescriptionDesc": "Décrivez le produit actuel. Alimente la génération de PRD et de PR/FAQ.",
     "prototype": "Prototype cliquable",
-    "prototypeDesc": "Créez un prototype HTML cliquable à partir de votre PRD / PR-FAQ",
+    "prototypeDesc": "Créez un prototype HTML cliquable à partir des spécifications du projet et, en option, de sa description produit et de ses recherches",
     "remixDocuments": "Remixer les documents",
     "remixDocumentsDesc": "Combiner et réviser les documents pour créer de nouvelles versions",
     "runResearch": "Lancer une recherche",
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "このプロジェクトにはすでにプロトタイプがあります。再度ビルドすると既存のものを保持したまま新しいものが追加されます。続行しますか?",
       "downloadHtml": ".html をダウンロード",
       "downloadJson": ".json をダウンロード",
+      "extraSources": "追加で参照 (任意)",
       "feedbackBuilding": "修正中…",
       "feedbackButton": "フィードバックで修正",
       "feedbackHeading": "フィードバックでプロトタイプを修正",
@@ -99,10 +100,13 @@
       "needsDocs": "プロトタイプ作成を有効にするには、まず PRD または PR-FAQ を作成してください",
       "previewLabel": "ライブプレビュー",
       "rawLabel": "生の出力（解析失敗 — 再生成してください）",
+      "researchLimit": "1 回のビルドで最大 {{max}} 件のレポート",
       "sourceLatest": "最新",
       "sourcePrd": "PRD",
       "sourcePrfaq": "PR/FAQ",
-      "started": "開始しました — バックグラウンドジョブで進行状況を確認できます。"
+      "started": "開始しました — バックグラウンドジョブで進行状況を確認できます。",
+      "useProductContext": "製品・サービスの説明",
+      "useResearch": "リサーチレポート ({{total}})"
     },
     "revision": {
       "feedback": "フィードバック: {{feedback}}",
@@ -257,7 +261,7 @@
     "productDescription": "製品／サービス説明",
     "productDescriptionDesc": "現在の製品を記述してください。PRD と PR/FAQ の生成に反映されます。",
     "prototype": "クリック可能なプロトタイプ",
-    "prototypeDesc": "PRD / PR-FAQ からクリック可能な HTML プロトタイプを作成します",
+    "prototypeDesc": "プロジェクトの仕様書から、必要に応じて製品説明やリサーチも参照して、クリック可能な HTML プロトタイプを作成します",
     "remixDocuments": "ドキュメントをリミックス",
     "remixDocumentsDesc": "ドキュメントを組み合わせて新しいバージョンに修正",
     "runResearch": "リサーチを実行",
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -88,6 +88,7 @@
       "confirmRebuild": "이 프로젝트에는 이미 프로토타입이 있습니다. 다시 빌드하면 기존 항목을 유지한 채 새 항목이 추가됩니다. 계속하시겠습니까?",
       "downloadHtml": "HTML 다운로드",
       "downloadJson": "JSON 다운로드",
+      "extraSources": "추가로 참고 (선택)",
       "feedbackBuilding": "수정 중…",
       "feedbackButton": "피드백으로 수정",
       "feedbackHeading": "피드백으로 프로토타입 수정",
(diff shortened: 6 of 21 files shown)


perrozzi:

This PR adds three optional request fields (`use_product_context`, `use_research`, `selected_research_ids`) to `POST /projects/{id}/build-prototype`, threads them into `_generate_prototype` as two new prompt sections, records the result in the document's `derivation`, and exposes per-build tick-boxes on the prototype ActionCard, with revisions inheriting what their base was built with. Validation is done at the boundary with the existing `_validated_source_id` under `RESEARCH#`, arity is bounded before the first read, all eight locale catalogues are updated, and both the Python and TypeScript sides get substantial regression tests (including a golden-prompt pin for the "asked for nothing" path). Overall this is careful, well-scoped work; I found no security or correctness blocker, only a handful of things worth tightening.

### Should fix

- **Prompt-budget ceiling for research is only per-document.** `RESEARCH_PER_DOC_CAP = 3000` × `MAX_SELECTED_RESEARCH_IDS = 10` allows ~30k chars of research to be appended (`lambda/jobs/document_generator/handler.py`), on top of `product_context_section` (`PER_DOC_CAP`), PRD/PR-FAQ and — on a revision — the prior prototype HTML. There is no aggregate bound, so the worst case can crowd out the spec the prototype is supposed to implement. Consider a total research budget in addition to the per-doc slice, and a test at the bound.
- **Verify the new imports in `OverviewTab.tsx`.** The diff introduces `clsx(...)`, `PrototypeBuildControl['extras']` and `TFunction<'projectDetail'>` in `PrototypeExtraSources`/`SourceCheckbox`, but shows no import hunk; the surrounding code styles classNames with template literals rather than `clsx`, which suggests it may not be imported today. If any of the three isn't already imported this won't typecheck.
- **The revision path derives `use_research` differently from the build path.** `DocumentsTab.tsx` sends `use_research: extraSources.researchIds.length > 0`, while `usePrototypeBuild.ts` sends the user's flag and gates the ids on it. Two rules for one API field invites drift; make the revision send the flag-and-ids pair with the same shape. Also, `inheritedExtraSources` does not slice the inherited list to `MAX_SELECTED_RESEARCH_IDS` — today ≤10 is guaranteed because the API capped the base build, but if the bound is ever lowered, existing prototypes become un-revisable with a 400 the user cannot act on.
- **Controls render on a disabled prototype card.** `controls` is passed unconditionally to `ActionCard` in `OverviewTab.tsx`, so a project with no PRD/PR-FAQ shows tick-boxes above a disabled button and the `needsDocs` hint. Gate `controls` on `!disabled`, or assert the current behaviour is intended in `OverviewTab.prototypeExtraSources.test.tsx` (the suite only covers the enabled card).
- **The silently-dropped inherited report needs a follow-up issue, not just a docstring.** `inheritedExtraSources` drops a deleted research report with no user-visible note, and the existing `feedbackRebased` copy would be wrong if reused. The reasoning in the comment is sound; please file the ninth-string follow-up referenced in the description so it isn't lost.

### Nits

- `derivation_source(_document_id_of(d), ROLE_REFERENCE)` uses a helper not visible in the diff, two lines below `(prd or {}).get('document_id')` — confirm the helper exists and pick one idiom for reading the id.
- `SourceCheckbox` uses `truncate` on the report title with no `title`/tooltip, so two similarly-named reports can render indistinguishably in the narrow card column.
- Duplicate ids count toward the arity bound (`len(raw) > MAX`) but collapse afterwards, so 11 entries that dedupe to 3 are a 400. Defensible, but `test_a_list_at_the_bound_is_accepted` uses `['research_a'] * 10`, so nothing covers ten *distinct* ids passing.
- `test_research_selection_bound_lockstep.py` resolves the frontend file via `Path(__file__).parents[3]`; the assertion message is helpful, but the same number now lives in three places (Python, TS, and this test's regexes) — a comment in `overviewState.ts` pointing at the test would close the loop both ways.
- The generator validates/reads nothing for `selected_research_ids` when `use_research` is false, but the API still resolves every id with a keyed read regardless of the flag; harmless with the current frontend, cheap to short-circuit.

---

perrozzi:
Thanks — useful round. Eight of ten taken, two refuted with evidence. Pushing the fixes right after this comment.

## Should fix

**1. Aggregate research budget — fixed, and it was the right catch.** `RESEARCH_TOTAL_CAP = 12000`, sized against the other caps in that function rather than picked: the PRD, the PR/FAQ and the product-context block each get `PER_DOC_CAP` (12000) and a revision's prior HTML gets `PRIOR_CAP` (24000), so research — corroborating evidence, not the thing being built — gets no more than a single spec document in total. Two bounds, because they fail differently: each report is sliced to an equal share, so 1–4 reports are cut exactly as before and ten get 1200 each and all ten stay present; then the assembled body is hard-capped, which is what holds the length against report *titles*, bounded nowhere in this path. Deliberately shared rather than spent front-to-back — a running budget would drop the last reports entirely while the derivation still recorded them as used. Worst-case prompt 90k → 72k chars, research 33% → 17%. Tested at the aggregate bound with a fixture whose per-document slices individually fit and together exceed, so the test fails if only the per-document cap exists.

**2. `OverviewTab.tsx` imports — refuted.** `clsx`, `PrototypeBuildControl` and `TFunction` are all imported (lines 18, 29 and 31), and `tsc -b` is clean on this branch, which is the stronger evidence. `clsx` is newly imported here, which is why the surrounding template-literal style made it look absent. No change.

**3. Revision path — both halves fixed, and I want to be straight about which one matters.** `inheritedExtraSources` now slices to `MAX_SELECTED_RESEARCH_IDS` and returns the flag alongside the ids, so the send site reads the pair exactly as the build path does. The **slice is genuinely behavioural** and is tested (a base carrying more reports than the current bound still revises) — that forward-compat hazard was the real finding. The **"same shape" half has no behavioural delta**: after the slice, `researchIds.length > 0` and the returned flag are the same value, so it buys maintainability only, and I did not write a test for it because nothing observable distinguishes the two call-site shapes. Worth saying rather than claiming coverage I do not have. Note the derivation records the *reports used*, not the tick-box, so a revision's flag must be derived from the ids — moving that derivation into one place is what removes the second rule.

**4. Controls on a disabled card — fixed, with a side effect you did not mention.** Gating on the card's disabled state also hides the controls while the start request is in flight, that being the prototype card's other disabled reason. I judged that correct rather than merely tolerable: nothing ticked in that window can affect a build already starting. Documented in the code. If you would rather they stay put during the request, gating at the call site on `missingUpstream` instead is a one-line change.

**5. The silently-dropped inherited report — recorded, not filed on GitHub.** This repo keeps unfiled issue bodies in `analysis/` until a human files them, so it is going in the ladder document alongside the other rungs rather than becoming an orphan issue. The docstring stays as the in-code record.

## Nits

- **`_document_id_of`** exists and is used four times; adopted consistently in the `build_derivation` call I touched. The two `source_*_id` **attributes** keep `.get` on purpose — a stored `None` and a stored `''` are different values to existing readers, and those lines are not part of this change.
- **Tooltip** added on the truncated title.
- **Ten distinct ids** — good catch, the fixture did dedupe to one. Replaced, plus a separate test pinning the deliberate asymmetry that duplicates count toward the bound *before* collapsing.
- **Lockstep pointer** — the comment already named the test. I added the part that actually closes the loop: that the test matches `^export const MAX_SELECTED_RESEARCH_IDS = <digits>` as source text, so folding it into an object or computing it breaks the test.
- **Short-circuit** — fixed. Ids arriving beside a false switch are now **ignored rather than rejected** (a 4xx over a field no build will read fails a request for a reason the user cannot see) and **dropped rather than passed through**, so a replay cannot reach an unvalidated id. Both halves are stated in the comment and tested.

## Verification

19 mutation probes, 19 caught, every one restored byte-identically. **One probe exposed a real defect in a test I wrote this round**: the "small selection is sliced exactly as before" assertion used `in`, which any cut at or past the cap satisfies, so a share one character tighter passed. Now pinned from both sides, with a second probe confirming the loosening direction is caught too.

| gate | before | after |
|---|---|---|
| `pytest lambda/jobs/document_generator/test` | 82 | **86** |
| `pytest lambda/api/test` | 558 | **564** |
| `vitest src/pages/ProjectDetail` | 395 | **399** |
| `tsc -b` / `eslint --max-warnings 0` | clean | clean |
| `ruff`, the two edited files | 177 / 136 | **177 / 136, unchanged** |

The two ruff numbers are pre-existing; the bar is that they do not grow. One new `I001` on my own import was fixed, so the touched test files are clean rather than merely not-worse.

Still true and unchanged from the description: **no browser has rendered these controls.** jsdom asserts reachability, which is a different claim, and live verification is owed after deploy.

perrozzi:

Since the last review the author added an aggregate research budget (`RESEARCH_TOTAL_CAP` + `_research_section` with tests), gated `controls` on `!disabled`, unified the revision path onto the flag-and-ids pair and sliced the inherited list to the bound, added a `title` tooltip, short-circuited the API's research reads when `use_research` is off, switched the PRD/PR-FAQ derivation sources to `_document_id_of`, and added a distinct-ids-at-the-bound test plus a pointer comment in `overviewState.ts`. Nearly all previous findings are addressed.

### Resolved
- Aggregate prompt budget for research now exists (`RESEARCH_TOTAL_CAP`, `_research_section`) with a bound test and an anti-vacuity guard.
- Revision path sends the same flag-and-ids shape as `usePrototypeBuild`; `inheritedExtraSources` slices to `MAX_SELECTED_RESEARCH_IDS` with a regression test.
- Controls no longer render on a disabled prototype card, with both halves tested (`OverviewTab.prototypeExtraSources.test.tsx`).
- `title` added to the truncated report label, plus a test.
- One idiom for reading the source id (`_document_id_of` for all three sources).
- Ten *distinct* ids at the bound now covered; duplicate arity/collapse asymmetry documented and tested.
- Ids sent with `use_research` off no longer cost N keyed reads, and are stored as `[]`; parametrised tests cover resolvable/unresolvable/over-long/not-a-list.
- `overviewState.ts` now documents the lockstep regex contract.

### Still open
- Cannot confirm `clsx`, `TFunction` and `PrototypeBuildControl` are imported in `OverviewTab.tsx` — no import hunk appears in either diff; please verify `tsc -b`/`eslint` covered this file.
- The follow-up issue for the silently-dropped inherited research report (the ninth i18n string) is still only a code comment.

### New findings

**Should fix — `_research_section`'s equal share ignores headings, so the hard cap re-introduces front-to-back spending** (`lambda/jobs/document_generator/handler.py`). `per_doc` is computed against `RESEARCH_TOTAL_CAP` alone, but each block also carries `"### {title}\n"` and blocks are joined with `\n\n`, so the assembled body routinely exceeds the cap and `body[:RESEARCH_TOTAL_CAP]` trims the tail. With the 8×2450 fixture that costs the last report ~120 chars (harmless, hence the test passes), but the docstring itself notes titles are user-supplied and bounded nowhere on this path: ten reports with ~1000-char titles push the body past 22k and the hard slice drops the last several reports entirely — the exact "derivation records a report that reached none of the prompt" failure the shared-budget design exists to prevent. Cap the title (e.g. `str(d.get('title') or 'Untitled')[:TITLE_CAP]`) and subtract heading/separator overhead from the per-doc share so the final slice is a belt-and-braces no-op; add a fixture with long titles asserting `HEAD{i}` for the *last* report.

**Nit** — `assert len(block) <= RESEARCH_TOTAL_CAP + len('RESEARCH FINDINGS:\n')` in `test_eight_reports_that_each_fit_are_bounded_together` passes for the truncation case above; a companion assertion that the last report's `HEAD` marker survives would make the long-title regression visible from this suite.

---

perrozzi:
Round 2 taken in full. The new should-fix was a real bug in the fix I shipped in round 1, and it was worse than the review estimated — thank you. Pushing after this comment.

## The should-fix: measured before fixing

I extracted round 1's `_research_section` and ran it in isolation against four fixtures, so the arithmetic is measured rather than argued:

| fixture | assembled | kept | overrun | reports with content in the prompt |
|---|---|---|---|---|
| 8 × 2450, short titles *(the existing test)* | 12118 | 12000 | 118 | **8/8** |
| 10 × 2450, short titles | 12148 | 12000 | 148 | **10/10** |
| **10 × 2450, ~1011-char titles** | **22178** | 12000 | **10178** | **5/10** |
| 4 × 3500, short titles | 12058 | 12000 | 58 | 4/4 |

So the long-title case lost **five of ten reports entirely**, each still recorded in the document's `derivation` as used — exactly the lie the shared budget exists to prevent. And the first row is why it was invisible: 118 characters off a tail that was already being cut.

**Fixed** by capping the title (`RESEARCH_TITLE_CAP = 120`, sized against the titles this system actually writes — research reports are created as `Research: {question[:50]}`, about 60 characters, so 120 leaves that much room again for a hand-edited one) and subtracting the heading-plus-separator overhead from each share. Every block is charged a separator including the last, one more than is written, so the body lands under rather than over. After the fix the same fixtures assemble to 11102, 11998 and 11550 — every case within the cap with zero overrun and every report present, so the hard slice is now a no-op at every arity the API can reach.

`or 'Untitled'` on the way in also fixes a stored `None` rendering as the literal string `"None"`.

**Falsifier added**, which is the part that matters: ten reports with 1011-character titles, asserting the **last** report's marker survives — the earlier reports survive front-to-back spending too, so only the last one discriminates. Plus an anti-vacuity guard asserting the unpaid-for body would lose *more than one report's worth*, so the fixture cannot silently degrade into the harmless short-title tail-trim.

**Nit taken too**: `test_eight_reports_that_each_fit_are_bounded_together` now asserts the last report's marker as well as the length, because a length assertion is satisfied by a body truncated *into* the bound just as much as by one that fitted.

## Two things I want on the record, because I did not do exactly as asked

**1. "The 1–4 report slice must not change" was not satisfiable, and was already false at 4 before this PR.** Four reports at the full 3000 plus any heading at all exceeds 12000, so a by-construction 12000 ceiling and four full-cap reports cannot both hold. On the previous commit, four 3500-char reports assembled to 12058 and the ceiling silently trimmed 58 characters off the *fourth*. The real choice was an even 127-character trim or an uneven one dumped on the last report, and I took the even one: **1–3 reports are byte-for-byte unchanged, 4 now gets 2873 instead of a nominal 3000 it was never actually delivered.** The stale "up to four is unaffected" comment is corrected in place with that measurement. The alternative — redefining the ceiling as content-budget-plus-overhead, ~13270 — would let research take about 10% more of the prompt than a spec document, contradicting the reasoning 12000 was chosen from, so I did not.

**2. The hard ceiling would have been left untested by this fix, and I am not hiding that.** Once the shares pay for their own headings the slice is a no-op at every arity the API permits, so the "remove the hard ceiling" mutation initially had nothing to kill it. Rather than ship a guard no test exercises, there are now two direct-call tests at arity 200 — the only regime where headings alone exceed the budget and the guard still does work. Worth knowing the ceiling's no-op status in production depends on `MAX_SELECTED_RESEARCH_IDS` staying where it is; if that bound ever grew past ~94, the section would start losing reports to the slice again.

## The two items still listed open

**The `OverviewTab.tsx` imports — refuted a second time, with the decisive evidence.** No import hunk appears because **I did not change the imports**; `clsx` is at line 18, `PrototypeBuildControl` at 29 and `TFunction` at 31 on this branch. `tsc -b` compiles that file as part of the project and is clean, and `eslint src --max-warnings 0` is clean. A missing import of any of the three is a compile error, so a clean `tsc` is proof rather than circumstantial. No change made, and none needed.

**The dropped-inherited-report follow-up is now written down**, in `analysis/FEATURE-prototype-visual-grounding-GITHUB-ISSUE.md` as a rung-1 follow-up, with the reason it needs its own string (the existing rebased note promises "the latest of that type will be used instead", which is true for a PRD and false for research — nothing substitutes) and the fixture that would falsify a fix. This repo keeps unfiled issue bodies in `analysis/` until a human files them, so that is the right home rather than an orphan GitHub issue.

## Gates

| gate | before | after |
|---|---|---|
| `pytest lambda/jobs/document_generator/test` | 86 | **91** |
| `pytest lambda/api/test` | 564 | 564 |
| `vitest src/pages/ProjectDetail` | 399 | 399 |
| `tsc -b` / `eslint --max-warnings 0` | clean | clean |
| `ruff`, the two edited files | 177 / 136 | **177 / 136, unchanged** |

9 mutation probes this round, 0 survived, 0 skipped, every one restored byte-identically — including a probe that reverts round 1's arithmetic verbatim, which fails three tests.

Also spotted and deliberately **not** touched: the reference-document path in `_gather_context` has the same uncapped title heading, but it takes only three documents and has no aggregate slice, so nothing can be dropped there. That belongs to the separately tracked reference-cap issue rather than to this PR.

perrozzi:

Since the last review the only changes are in the research-section budget: a new `RESEARCH_TITLE_CAP` (120) plus `_RESEARCH_BLOCK_OVERHEAD`, the per-doc share now pays for its own heading and separator and is clamped at zero, the title is coerced/sliced, and four new tests cover long titles, the exact share cut, the negative-share clamp and the hard ceiling as a pure guard. That addresses the one should-fix from my previous review; the arithmetic checks out (n≤3 unchanged, n=4 → 2873, n=10 → 1073, and `n × (overhead + share) ≤ 12000 − 2` holds, with the 200-report case landing on the slice).

### Resolved
- Heading/separator cost is now charged to each share and the title is capped, so `body[:RESEARCH_TOTAL_CAP]` no longer silently deletes the tail reports; long-title fixture asserts `HEAD{last}` survives.
- The nit is covered too: `test_eight_reports_that_each_fit_are_bounded_together` now asserts the last report's marker, so the ~120-char tail-trim it previously tolerated would fail.

### Still open
- Cannot confirm `clsx`, `TFunction` and `PrototypeBuildControl` are imported in `OverviewTab.tsx` — no import hunk in either diff; please confirm `tsc -b`/`eslint` covered that file.
- The follow-up issue for the silently-dropped inherited research report (ninth i18n string) is still only a code comment.

### New findings

**Nit** — in `lambda/jobs/document_generator/test/test_prototype_context_sources.py`, `from jobs.document_generator.handler import (_RESEARCH_BLOCK_OVERHEAD, RESEARCH_TOTAL_CAP,)` lists the underscore-prefixed name first, which isort/ruff `I001` would reorder if the rule is enabled for tests; the verification table only mentions ruff on the two non-test files, so worth a quick `ruff check` on the new test file.

Otherwise this delta looks clean — the guard is now genuinely belt-and-braces and the tests pin the cut from both sides.

---

perrozzi:
Round 3 is nit-only and the nit does not reproduce, so I'm treating the loop as converged and pushing nothing further. Thanks — round 2's finding was the most valuable comment on this PR.

**The nit — refuted, measured.** Ruff is clean on all three test files, and clean with the rule explicitly selected rather than merely absent from the default set:

```
ruff check lambda/jobs/document_generator/test/test_prototype_context_sources.py \
           lambda/api/test/test_build_prototype_sources.py \
           lambda/api/test/test_research_selection_bound_lockstep.py
→ All checks passed!

ruff check --select I001 lambda/jobs/document_generator/test/test_prototype_context_sources.py
→ All checks passed!
```

The reasoning was right — two `I001`s were introduced in the round-2 work — they were just already fixed before that commit landed, which the verification table should have said explicitly rather than reporting only the two production files. Fair catch on the reporting gap, no code change needed.

## The two items that will stay "open", and why they cannot close

Both are artefacts of what the reviewer can see, not outstanding work.

**The `OverviewTab.tsx` imports.** No import hunk appears in any diff **because the imports were not changed** — `clsx` (line 18), `PrototypeBuildControl` (29) and `TFunction` (31) were already there. `tsc -b` compiles that file as part of the project and is clean on every commit of this branch, and a missing import of any of the three is a compile error, so a clean build is proof rather than circumstantial. This is round 3 of the same request; it can't be satisfied by a diff.

**The dropped-inherited-report follow-up** is written down, in `analysis/FEATURE-prototype-visual-grounding-GITHUB-ISSUE.md`, as a rung-1 follow-up with the reason it needs its own i18n string and the fixture that would falsify a fix. That directory lives outside the git repository in this workspace, which is why it never appears in a diff. It is the repo's convention for issue bodies awaiting a human to file them.

## Where the PR stands

Three commits: the feature, the aggregate budget, and the heading arithmetic. Final gates, all by delta against a baseline captured before any of this work:

| gate | baseline | now |
|---|---|---|
| `pytest lambda/jobs/document_generator/test` | 67 | **91** |
| `pytest lambda/api/test` | 540 | **564** |
| `vitest src/pages/ProjectDetail` | 364 | **399** |
| full frontend suite | — | 2394 passed, 163 files |
| `tsc -b` / `eslint src --max-warnings 0` | clean | clean |
| `ruff`, the two edited production files | 177 / 136 output lines | **177 / 136, unchanged** |
| `ruff`, all three test files | — | clean |
| `npm run i18n:check` | already red, 2 unrelated causes | red, output byte-identical to the pre-change capture |

54 mutation probes across the three rounds, all caught. Three survived on first run and every one was a real test defect rather than a probe artefact: a switch tested only for an absent key and never for the explicit `false` the API actually sends; a stale-selection filter that needed a re-render to observe; and a boundary assertion using `in`, which any cut at or past the cap satisfies.

**Still owed and not claimed: no browser has rendered these controls.** jsdom asserts they are reachable and behave, which is a different claim from painted. Live verification after deploy.

PR #323, opened 2026-08-13
fix(bedrock): never send temperature with an explicit thinking budget

perrozzi:
## The bug

Every Research job failed with:

```
ValidationException: The model returned the following errors:
`temperature` may only be set to 1 when thinking is enabled
```

`converse()` decided `temperature` and the explicit thinking budget from two
**independent** per-model capability flags (`omit_temperature`,
`adaptive_thinking`) and never considered the pair. Where both flags are false it
sent both fields, which Anthropic rejects outright.

That is exactly two allowlisted models — **Sonnet 4.6** and **Haiku 4.5**.
Selecting either in the Settings model picker broke **100% of Research jobs**. The
other three passed only by accident: their adaptive-thinking flag drops the
thinking field, which incidentally removed the conflict.

**This matters more than the model count suggests: every workshop deployment
ships with Research dead**, because `-c defaultModelId` pins those accounts to
Sonnet 4.6 (they have no Sonnet 5 / Opus 5 access) — precisely the broken model.

## Why the capability flags couldn't catch it

The failure is a **combination**, not a per-model property. Both flags were
individually correct and no model row needed changing; the bug lived in the
request builder. So the invariant belongs in a test, not in the capability table.

## The fix

One condition where the inference config is built, with `explicit_thinking`
hoisted above it because it now constrains `temperature`:

```python
if temperature is not None and not explicit_thinking and not omits_temperature(used_model):
    inference_config['temperature'] = temperature
```

Omitting `temperature` is equivalent to the required value of `1` and keeps a
single exit shape. Deliberately not sending an explicit `temperature: 1`, which
would be a second code path saying the same thing.

**Blast radius was narrow** because research's `data_analysis` step
(`lambda/api/prompts/research-analysis.json`, `thinking_budget=5000`) is the only
in-repo caller with a budget > 0. The streaming TS path sends no `temperature`
and `product_context.py` sends no thinking, so neither was exposed. Verified on
this branch that no other non-zero budget exists in the tree.

No test was added for the continuation / budget-raise paths because
`allow_continuation = max_continuations > 0 and not explicit_thinking` makes them
unreachable with explicit thinking. Worth revisiting if that gate changes.

## Observability fix in the same change

`[BEDROCK] Request params` reported the **requested** temperature and budget for
calls where Bedrock received neither — the wrong starting point when triaging a
400 about those exact fields, and it cost real time during this investigation.
Renamed to `Requested params` and added an `Effective params` line reporting what
actually goes on the wire.

## Tests

Two, both driven off the capability data rather than hardcoded ids so a future
model in the same combination is covered on arrival:

- a focused regression test over the at-risk models, which also asserts thinking
  is **still sent** — so the fix cannot silently disable it;
- the broader invariant over **every** allowlisted model at both budgets: the
  request must never carry `temperature` and an explicit thinking budget
  together.

Both **proven non-vacuous by reverting the fix** — exactly those two fail, then
pass again after a byte-equal restore. Each guards against vacuity explicitly
(the first asserts the at-risk set is non-empty, the second that the loop
actually iterated).

## Verification

Not just mocks:

- Live per-model probe through the real production code against real Bedrock:
  **3/5 → 5/5**.
- Full 5-step Research workflow on Sonnet 4.6 completed, saving a 58k-char
  report, with this per-step wire shape:

| step | budget | sent to Bedrock |
|---|---|---|
| `data_analysis` | 5000 | temperature **omitted**, thinking 5000 |
| `synthesis` | 0 | temperature **0.1**, thinking omitted |
| `validation` | 0 | temperature **0.1**, thinking omitted |

That table is the real evidence rather than "it succeeded": temperature still
travels on the two steps where it is legal, i.e. the change is scoped to the
illegal combination and does not drop temperature everywhere.

Gates on this branch (rebased onto the current `development` tip, not the older
commit the fix was written against): **backend suite 1660 pass**, no new lint
findings (ruff unchanged at its pre-existing count for these files).

## Operational note

This fix is already **hotfixed live in three workshop accounts** via
`update-function-code` on `voc-research-step` alone, so those deployments carry
CloudFormation drift on that one function until a real deploy lands this change.
On those accounts the deploy needs `--exclusively` or
`-c skipUseCaseSubmission=true`, or it fails on the `SubmitAnthropicUseCase`
custom resource ("Internal Accounts should not submit use case details").

### DIFF
--- a/voc-datalake/lambda/shared/converse.py
+++ b/voc-datalake/lambda/shared/converse.py
@@ -86,6 +86,33 @@ class BedrockThrottlingError(Exception):
     pass
 
 
+def _temperature_note(
+    sent: bool,
+    temperature: float | None,
+    model_id: str,
+    explicit_thinking: bool,
+) -> str:
+    """Name the REAL reason `temperature` is or is not on the wire.
+
+    Several suppression causes can hold at once — a caller passing None together
+    with an explicit budget (reachable today), or a model that both rejects
+    temperature and takes an explicit budget (reachable as soon as one is
+    allowlisted). Attributing the drop to whichever cause is checked first would
+    point an operator at the wrong one, which defeats the purpose of logging the
+    reason at all. So the branches mirror the suppression condition in order,
+    most caller-proximate first.
+    """
+    if sent:
+        return str(temperature)
+    if temperature is None:
+        return 'omitted (caller passed None)'
+    if omits_temperature(model_id):
+        return 'omitted (model rejects it)'
+    if explicit_thinking:
+        return 'omitted (explicit thinking)'
+    return 'omitted'  # pragma: no cover — no suppression cause left to name
+
+
 def _raised_empty_budget(current_max: int) -> int | None:
     """Next maxTokens to try after a model returned zero visible text.
 
@@ -229,10 +256,12 @@ def converse(
     # when triaging a ValidationException about those fields. The drop REASON is
     # spelled out so an operator reading only this line knows why it vanished
     # instead of inferring it from the thinking value.
-    if 'temperature' in inference_config:
-        effective_temperature = inference_config['temperature']
-    else:
-        effective_temperature = 'omitted (explicit thinking)' if explicit_thinking else 'omitted'
+    effective_temperature = _temperature_note(
+        sent='temperature' in inference_config,
+        temperature=temperature,
+        model_id=used_model,
+        explicit_thinking=explicit_thinking,
+    )
     effective_thinking = thinking_budget if explicit_thinking else 'omitted'
     logger.info(f"[BEDROCK] Effective params: temperature={effective_temperature}, thinking={effective_thinking}")
     logger.info(f"[BEDROCK] Invoking Bedrock converse API for step '{step_name}'...")
--- a/voc-datalake/lambda/shared/converse.py
+++ b/voc-datalake/lambda/shared/converse.py
@@ -193,20 +193,15 @@ def converse(
     inference_config = {'maxTokens': max_tokens}
     # `temperature` is dropped in three cases:
     #   - the caller passed None explicitly;
-    #   - the model rejects the parameter outright as deprecated (Sonnet 5,
-    #     both Opus generations) — so any surface can be pointed at them via
-    #     the picker without a 400;
+    #   - the model rejects the parameter outright as deprecated;
     #   - EXPLICIT extended thinking is on: Anthropic permits only
-    #     temperature=1 alongside thinking, and sending both is a hard 400
-    #     ("`temperature` may only be set to 1 when thinking is enabled").
-    #     Omitting it is equivalent to 1 and keeps one exit shape here.
+    #     temperature=1 alongside thinking, and sending both is a hard 400.
+    #     Omitting is equivalent to 1 and keeps one exit shape here.
     #
-    # That third case is a COMBINATION, not a per-model property, which is why
-    # the two capability flags alone couldn't catch it: it bites exactly the
-    # models that accept temperature AND take an explicit budget (Sonnet 4.6,
-    # Haiku 4.5). Live-caught on research's data_analysis step — the only
-    # in-repo caller with thinking_budget > 0 — where picking either model in
-    # the Settings model picker failed every Research job.
+    # Keep the third condition even though it looks redundant next to the
+    # capability flags: it is a COMBINATION, not a per-model property, so no
+    # per-model flag can encode it. It binds exactly the models that accept
+    # temperature AND take an explicit budget.
     if temperature is not None and not explicit_thinking and not omits_temperature(used_model):
         inference_config['temperature'] = temperature
     kwargs = {
@@ -231,12 +226,15 @@ def converse(
     # both temperature and the thinking budget can be dropped per model. The
     # earlier line alone made a request look like it carried a temperature and a
     # budget that Bedrock never saw, which is exactly the wrong starting point
-    # when triaging a ValidationException about those fields.
-    logger.info(
-        f"[BEDROCK] Effective params: temperature="
-        f"{inference_config.get('temperature', 'omitted')}, "
-        f"thinking={thinking_budget if explicit_thinking else 'omitted'}"
-    )
+    # when triaging a ValidationException about those fields. The drop REASON is
+    # spelled out so an operator reading only this line knows why it vanished
+    # instead of inferring it from the thinking value.
+    if 'temperature' in inference_config:
+        effective_temperature = inference_config['temperature']
+    else:
+        effective_temperature = 'omitted (explicit thinking)' if explicit_thinking else 'omitted'
+    effective_thinking = thinking_budget if explicit_thinking else 'omitted'
+    logger.info(f"[BEDROCK] Effective params: temperature={effective_temperature}, thinking={effective_thinking}")
     logger.info(f"[BEDROCK] Invoking Bedrock converse API for step '{step_name}'...")
     start_time = time.time()
 
--- a/voc-datalake/lambda/shared/converse.py
+++ b/voc-datalake/lambda/shared/converse.py
@@ -171,7 +171,7 @@ def converse(
     """
     used_model = model_id or get_active_model_id(surface)
     logger.info(f"[BEDROCK] Starting converse call for step '{step_name}' with model {used_model} (surface={surface})")
-    logger.info(f"[BEDROCK] Request params: max_tokens={max_tokens}, temperature={temperature}, thinking_budget={thinking_budget}")
+    logger.info(f"[BEDROCK] Requested params: max_tokens={max_tokens}, temperature={temperature}, thinking_budget={thinking_budget}")
     logger.info(f"[BEDROCK] Prompt length: {len(prompt)} chars, system_prompt length: {len(system_prompt)} chars")
     
     try:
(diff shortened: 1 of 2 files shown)


perrozzi:

This PR fixes a real Bedrock `ValidationException` by hoisting `explicit_thinking` above the inference-config build and suppressing `temperature` whenever an explicit thinking budget is sent, plus adds an "Effective params" log line and two pytest regression tests. The fix is minimal, correctly scoped to the illegal combination (temperature still travels on budget-0 steps), no model ids or IAM ARNs change, and the tests live in the right place (`lambda/shared/test/test_converse.py`). I found no blockers.

### Should fix

- `voc-datalake/lambda/shared/test/test_converse.py`: the vacuity guard `assert {'global.anthropic.claude-sonnet-4-6'} <= set(at_risk)` hardcodes a model id, which contradicts the stated "driven off the capability data rather than hardcoded ids" goal and will fail as a false negative the day Sonnet 4.6 is retired from `ALLOWED_MODELS` (a routine allowlist edit, unrelated to this invariant). `assert at_risk, "no model in the risky combination — guard is vacuous"` gives the same non-vacuity protection without pinning the table.
- `test_never_sends_temperature_together_with_explicit_thinking` is satisfied if the code regressed to dropping *both* fields for every model. Add the positive half of the invariant inside the same loop, e.g. for `budget == 0` assert `('temperature' in inferenceConfig) == (not omits_temperature(model['id']))`. Otherwise the "temperature is only dropped where illegal" property is only covered by the manual wire-shape table in the PR description, not by CI.

### Nits

- `test_never_sends_temperature_together_with_explicit_thinking` uses `'additionalModelRequestFields' in kwargs` as a proxy for "thinking was sent". If any other `additionalModelRequestFields` key is added later the test starts failing for unrelated reasons; prefer `kwargs.get('additionalModelRequestFields', {}).get('thinking')`.
- `voc-datalake/lambda/shared/converse.py`: the ~15-line comment block above the `if` encodes incident narrative ("Live-caught on research's data_analysis step…", Settings-picker symptoms) that will go stale. The invariant itself ("only temperature=1 is legal with thinking; omitting is equivalent to 1") is the durable part — consider trimming the rest to the PR/commit message.
- The `Effective params` f-string is split across three lines with an empty trailing segment on the first; a single f-string (or a small dict) reads better and keeps grep-ability of the log prefix.
- Consider making the temperature drop visible as such (e.g. `temperature=omitted (explicit thinking)`) so an operator reading only `Effective params` knows *why* it vanished rather than having to infer it from the thinking value.
- The first test is largely a subset of the second (both iterate `ALLOWED_MODELS`, both assert the no-both-fields property); its unique value is the "thinking is still sent" assertion. A short comment saying so would keep a future reader from deleting the wrong one.

---

perrozzi:
Thanks — both "should fix" items were real. Addressed all of them, with one correction to the framing of the second.

### Should fix

**1. Hardcoded model id in the vacuity guard — accepted, and removed rather than replaced.**

You're right that it contradicted the stated goal and would become a false negative when Sonnet 4.6 is retired. I didn't swap in `assert at_risk` though, because that shares the same failure mode one step further out: if *both* at-risk models are ever retired, `at_risk` is legitimately empty and the assert fires on a table change that has nothing to do with this invariant.

Instead the guard is gone entirely and non-vacuity now comes from `assert checked == len(ALLOWED_MODELS) * 2`, which proves the loop ran without asserting anything about which models exist.

**2. Invariant satisfied by dropping *both* fields — accepted, with a correction.**

The hole in the two tests you reviewed was real: I verified it by mutating the condition to `if False:` (never send temperature) and both of them passed.

But the property was **not** absent from CI. `test_includes_temperature_by_default` catches that same mutation. So the risk was narrower than "only covered by the manual wire-shape table in the PR description" — it was covered, just not by the tests you were reading.

The change is still worth making, for your own finding-1 reason: that pre-existing test pins a single hardcoded model id, so it has exactly the staleness problem you flagged. The property now lives in the data-driven invariant, asserted for every model:

- thinking present iff the model takes an explicit budget (so suppressing temperature can never silently disable the thinking a caller asked for);
- when thinking is in play, temperature absent;
- when it isn't, temperature presence must follow `omits_temperature()` — the positive half.

Verified with a three-way mutation battery, each producing a distinct failure:

| mutation | caught by |
|---|---|
| send temperature alongside thinking (the original bug) | "sent both temperature and an explicit thinking budget" |
| never send temperature (`if False:`) | "temperature presence should follow omits_temperature() when thinking is off" |
| never send the thinking budget | "expected explicit thinking=True" |

### Nits

- **`additionalModelRequestFields` as a proxy for thinking** — accepted, now reads `kwargs.get('additionalModelRequestFields', {}).get('thinking')`.
- **First test largely a subset of the second** — accepted, and rather than adding a comment I merged them. With the positive half added the second subsumed the first entirely (its unique assertion, "thinking is still sent", is now case 2 above). One test, so there's no wrong one to delete.
- **Comment block encodes incident narrative** — partially accepted. Cut the Settings-picker symptoms and the "live-caught on research's data_analysis step" detail, which are in the commit message. Kept the sentence explaining that this is a combination rather than a per-model property, because that's the durable part *and* it's what stops a future reader deleting `not explicit_thinking` as redundant next to the capability flags. Trimmed ~15 lines to 8.
- **Split f-string** — accepted, single f-string now.
- **Make the drop reason visible** — accepted, and a good call for on-call reading. The line now emits `temperature=omitted (explicit thinking)` when thinking is what suppressed it, versus a bare `omitted` when the model rejects the parameter or the caller passed `None`.

### Gates

Backend suite 1659 pass (net −1 from merging two tests into one), ruff unchanged at its pre-existing count for both files.

Not re-running the live Bedrock matrix for this round: the request-shaping logic is byte-identical, and the only behavioural deltas are the log text and the test consolidation.

perrozzi:

Since my last pass the two tests were merged into a single `test_temperature_and_thinking_sent_exactly_when_legal` (hardcoded id guard gone, both directions of the invariant asserted, nested `thinking` key read instead of the container), the incident narrative was trimmed out of both the `converse.py` comment and the docstring, and the `Effective params` line is now a single f-string that spells out the drop reason. All previous Should-fix items and nits are addressed; nothing blocking.

### Resolved
- Hardcoded `global.anthropic.claude-sonnet-4-6` vacuity guard removed; the test is now driven purely off `ALLOWED_MODELS` + capability helpers.
- Positive half of the invariant added: `sent_temperature == (not omits_temperature(model_id))` when thinking is off, so a regression dropping temperature everywhere now fails.
- Thinking presence is checked via `kwargs.get('additionalModelRequestFields', {}).get('thinking')` rather than container membership.
- Comment block in `converse.py` trimmed to the durable invariant; `Effective params` collapsed to one f-string and now reports the drop reason.

### New findings
- Nit — `voc-datalake/lambda/shared/test/test_converse.py`: with the `at_risk` list gone, the *regression* half (assertion 1) becomes vacuous if Sonnet 4.6 / Haiku 4.5 ever leave the allowlist; `checked == len(ALLOWED_MODELS) * 2` only proves the loop ran. A data-driven guard such as `assert any(not omits_temperature(m['id']) and not uses_adaptive_thinking(m['id']) for m in ALLOWED_MODELS)` restores that without pinning an id.
- Nit — `voc-datalake/lambda/shared/converse.py`: `'omitted (explicit thinking)'` is attributed to thinking even when the real cause was `temperature is None` or `omits_temperature(used_model)` (both can co-occur with an explicit budget), so the log can point an operator at the wrong reason. Checking `temperature is None` / `omits_temperature` first would keep it accurate.

---

perrozzi:
Both nits addressed. The second was a real bug, and it partly overturns my reasoning from the last round.

### Log attributed the drop to the wrong cause

Correct, and worse than a nit in effect: `temperature=None` together with an explicit budget is reachable **today**, so the line could tell an operator "explicit thinking" when the caller simply passed `None` — sending them to look for a model-capability problem that doesn't exist. A log whose only job is naming the reason is worse than silent when it names the wrong one.

Extracted `_temperature_note()`, which checks causes in the same order as the suppression condition, most caller-proximate first: `None` → model rejects it → explicit thinking. Extracting also made the attribution directly testable, which is what the nit really exposed, so there's now a test pinning each cause against the case where first-match-wins would misreport it.

Confirmed with a fourth mutation — reordering the chain to blame thinking first — which the new test catches.

### Vacuity guard: you were right and I wasn't

Last round I declined `assert at_risk` on the grounds that it fires when the risky pairing legitimately disappears from the table. That reasoning was weak: I treated "fails on a table change" as equivalent to "wrong", when a guard that fires deliberately is a tripwire rather than a false alarm — provided the failure explains itself. The distinction I should have drawn is between a cryptic assert and a documented one.

So it's in, in your data-driven form, with the message doing the work:

```python
assert any(
    not omits_temperature(m['id']) and not uses_adaptive_thinking(m['id'])
    for m in ALLOWED_MODELS
), "no allowlisted model can pair temperature with an explicit thinking budget"
```

and a comment saying that if it fires, the allowlist no longer contains a model that can produce the bug — confirm that's intended, then the regression half can be retired with it.

### One thing I'd flag in your framing

The new attribution test needed a model that accepts temperature and one that rejects it, and my first cut hardcoded two ids — exactly the staleness you flagged in round 1. It now derives both from `ALLOWED_MODELS` + `omits_temperature()`, with an explicit assert that one of each exists. Worth noting because it's the same trap re-appearing one layer down: any test that needs a model with a specific capability should select it by capability, never by id.

### Mutation battery

| mutation | outcome |
|---|---|
| send temperature alongside thinking (original bug) | 1 failed |
| never send temperature | 3 failed |
| never send the thinking budget | 2 failed |
| blame thinking first in the attribution chain | 1 failed |

Backend suite 1660 pass, ruff unchanged at its pre-existing count for both files. No change to the request shape, so the live Bedrock matrix from the PR description still holds.

perrozzi:

Both remaining nits from the last pass are addressed: the invariant test now carries a data-driven guard proving some allowlisted model can actually reach the illegal pairing, and the drop-reason string moved into `_temperature_note` with causes checked most-caller-proximate first, plus a dedicated test pinning each attribution. Clean; nothing blocking.

### Resolved
- `test_converse.py`: `assert any(not omits_temperature(...) and not uses_adaptive_thinking(...))` restores the regression half's non-vacuity without pinning an id.
- `converse.py`: `_temperature_note` now reports `caller passed None` / `model rejects it` / `explicit thinking` in suppression order, so the log can no longer blame thinking for a `temperature is None` or capability drop.

### New findings
- Nit — `voc-datalake/lambda/shared/test/test_converse.py`: `assert accepts and rejects` in `test_temperature_note_names_the_actual_suppression_cause` makes the test fail if the allowlist ever contains no `omits_temperature` model — the exact "confusing false negative" the surrounding comment says it avoids. Since `omits_temperature()` presumably returns `False` for unknown ids, one option is patching the capability lookup for the rejecting case instead of sourcing it from `ALLOWED_MODELS`; not worth blocking on.

---

perrozzi:
Fair catch — I introduced the exact false negative that test's own comment claimed to avoid.

Fixed by removing the coupling rather than working around it. What the test is about is the **order** the suppression causes are checked in, not the capability lookup, so the lookup is stubbed and the ids are synthetic (`model-that-accepts-temperature` / `model-that-rejects-temperature`). No fixture-availability assert needed at all.

Your patching suggestion, taken — with one refinement: rather than stubbing only the rejecting case, the stub covers both, so the test no longer touches `ALLOWED_MODELS` in any form. Real models stay covered by the invariant test above, which is where that coverage belongs.

Worth naming the distinction, since these two guards look similar and only one should exist:

- The **invariant test's** guard fires deliberately. It asserts that some allowlisted model can still reach the illegal pairing, which is a real property of the system and worth being told about when it changes.
- This test's assert was **incidental**. Fixture availability says nothing about the system, so a failure there carried no information — it was just coupling.

Verified the stub doesn't weaken anything: reordering the attribution chain to blame thinking first is still caught.

Backend suite 1660 pass, ruff unchanged.

---

On process: three rounds in, the findings have moved from a real hole in the invariant to test-fixture coupling, and the remaining surface is stylistic. I'd treat this as done unless something blocking appears. The behavioural change is still the single condition from the first commit; everything since has been coverage and log accuracy.

perrozzi:

Only change since last pass is the `_temperature_note` ordering test, which now stubs `shared.converse.omits_temperature` and uses synthetic model ids instead of sourcing fixtures from `ALLOWED_MODELS` — exactly the nit I raised. Patch target is correct (the name is imported into `shared.converse`'s namespace and resolved at call time), so all six assertions exercise the intended branches, and the real-model coverage still lives in the invariant test.

### Resolved
- `voc-datalake/lambda/shared/test/test_converse.py`: dropped the `assert accepts and rejects` allowlist dependency; the ordering test no longer produces a false negative if no allowlisted model rejects `temperature`.

No new findings — this is clean and ready to merge from my side.

---

PR #324, opened 2026-08-13
feat(feedback-forms): show the document type in the validation link picker

perrozzi:
## Summary

The document `<select>` on the Feedback Forms editor's **Validates** tab listed titles only. A project's PRD and its PR/FAQ are generated from the same feature idea and routinely carry the same title, so an admin had to choose between two indistinguishable names — for a link that decides which **Prioritization** row that form's ratings show up on.

Each option now reads `Title (TYPE)`, e.g. `Instant payouts (PRD)` / `Instant payouts (PR/FAQ)`.

`(TYPE)` after the name, following `CheckboxItem` in `pages/ProjectDetail/PickerComponents` — the repo's other picker that shows a document's type inline — so the two read the same way. This was the only document picker that did not show the type; the DataSourceWizard, the prototype source picker and the Documents list cards already did.

## Changes

- **`ValidationLinkPicker.tsx`** — `documentOptionLabel()` composes the option text. The type label resolves from `SCORABLE_TYPE_META`, the same source as the badge on the Prioritization row this link feeds, so the two cannot disagree.
- **`prioritizationUtils.ts`** — `SCORABLE_TYPE_META`'s `i18nKey` is now namespace-qualified (`prioritization:docType.prd`). Two reasons, both load-bearing:
  - It is read here through a `t` bound to `feedbackForms`. A relative key resolves against *that* namespace and renders the raw key path to the user.
  - `scripts/i18n-check.mjs` only collects data-held keys that carry a namespace (`extractDataHeldKeys`). Both keys were on the "not directly referenced" list and were therefore deletion candidates in a cleanup pass. That list drops 635 → 633.
- **Missing-title guard** — `projectsApi` has no schema at its boundary, so `ProjectDocument.title` being declared `string` enforces nothing. `{doc.title}` rendered an empty option for a legacy record without one; interpolating it into a template prints the literal `undefined`. Falls back to the document id, as `DocumentsTab` does for a revision with no title.

## Testing

- **2458/2458** frontend tests, `tsc --noEmit` and eslint clean.
- `i18n-check.mjs` output identical to baseline apart from the two keys becoming referenced.
- New tests: option labelling with two **same-titled** documents (the motivating case, asserting the bare title is no longer any option's text); the titleless/blank-title fallback; and a resolution gate over every `SCORABLE_TYPE_META` entry.
- **The label gate binds `t` to a foreign namespace on purpose.** No Prioritization test asserts the badge text, so un-qualifying the keys left that entire suite green — a `prioritization`-bound test would pass with or without the prefix and prove nothing.
- **4 mutations run, each caught, file restored after**: drop the type from the label; drop the namespace prefix (twice — once against the picker's suite, once against the new gate in isolation); delete the missing-title guard. Every run asserted a green baseline first, so the failures mean something.

## Not covered

No live browser check at PR time — the assertions are jsdom, which covers the option text but not how a native `select` popup renders at narrow widths.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/de/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "Erfassen Sie optional, zu welchem Projekt — und zu welchem PRD oder PR/FAQ — dieses Formular Feedback sammelt. Die erfassten Bewertungen erscheinen dann neben diesem Dokument auf der Priorisierungsseite. Lassen Sie beide Felder leer für ein eigenständiges Formular, etwa eine Website-Umfrage.",
     "validationDocumentHint": "Belassen Sie es beim gesamten Projekt, damit die Verknüpfung erhalten bleibt, wenn ein Dokument neu erzeugt wird.",
     "validationDocumentLabel": "Dokument (optional)",
+    "validationDocumentOption": "{{title}} ({{type}})",
     "validationLoadingLink": "Verknüpfung wird geladen …",
     "validationNoProject": "-- Nicht verknüpft --",
     "validationProjectHint": "Wird das Projekt geleert, ist dieses Formular nicht mehr verknüpft.",
--- a/voc-datalake/frontend/public/locales/en/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/en/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "Optionally record which project — and which PRD or PR/FAQ — this form collects feedback about. The ratings it gathers then appear next to that document on the Prioritization page. Leave both empty for a standalone form such as a website survey.",
     "validationDocumentHint": "Leave on the whole project to keep the link when a document is regenerated.",
     "validationDocumentLabel": "Document (Optional)",
+    "validationDocumentOption": "{{title}} ({{type}})",
     "validationLoadingLink": "Loading link…",
     "validationNoProject": "-- Not linked --",
     "validationProjectHint": "Clearing the project unlinks this form.",
--- a/voc-datalake/frontend/public/locales/es/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/es/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "Opcionalmente registre a qué proyecto — y a qué PRD o PR/FAQ — se refieren los comentarios de este formulario. Las valoraciones recopiladas aparecerán junto a ese documento en la página de Priorización. Deje ambos vacíos para un formulario independiente, como una encuesta del sitio web.",
     "validationDocumentHint": "Déjelo en el proyecto completo para mantener el vínculo cuando se regenere un documento.",
     "validationDocumentLabel": "Documento (opcional)",
+    "validationDocumentOption": "{{title}} ({{type}})",
     "validationLoadingLink": "Cargando el vínculo…",
     "validationNoProject": "-- Sin vincular --",
     "validationProjectHint": "Borrar el proyecto desvincula este formulario.",
--- a/voc-datalake/frontend/public/locales/fr/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/fr/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "Indiquez éventuellement à quel projet — et à quel PRD ou PR/FAQ — se rapportent les retours de ce formulaire. Les notes recueillies apparaîtront alors à côté de ce document sur la page Priorisation. Laissez les deux champs vides pour un formulaire autonome, comme une enquête sur un site web.",
     "validationDocumentHint": "Laissez sur l'ensemble du projet pour conserver le lien lorsqu'un document est régénéré.",
     "validationDocumentLabel": "Document (facultatif)",
+    "validationDocumentOption": "{{title}} ({{type}})",
     "validationLoadingLink": "Chargement du lien…",
     "validationNoProject": "-- Non lié --",
     "validationProjectHint": "Effacer le projet délie ce formulaire.",
--- a/voc-datalake/frontend/public/locales/ja/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/ja/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "このフォームがどのプロジェクトの、どのPRDまたはPR/FAQについてフィードバックを集めるのかを任意で記録できます。収集した評価は優先順位付けページの該当ドキュメントの横に表示されます。ウェブサイトのアンケートなど単独のフォームでは両方を空のままにしてください。",
     "validationDocumentHint": "プロジェクト全体のままにすると、ドキュメントを再生成してもリンクが維持されます。",
     "validationDocumentLabel": "ドキュメント（任意）",
+    "validationDocumentOption": "{{title}}（{{type}}）",
     "validationLoadingLink": "リンクを読み込み中…",
     "validationNoProject": "-- リンクなし --",
     "validationProjectHint": "プロジェクトを空にするとこのフォームのリンクが解除されます。",
--- a/voc-datalake/frontend/public/locales/ko/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/ko/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "이 양식이 어떤 프로젝트의, 어떤 PRD 또는 PR/FAQ에 대한 피드백을 수집하는지 선택적으로 기록할 수 있습니다. 수집된 평점은 우선순위 지정 페이지의 해당 문서 옆에 표시됩니다. 웹사이트 설문처럼 독립적인 양식이라면 두 항목을 비워 두세요.",
     "validationDocumentHint": "프로젝트 전체로 두면 문서를 다시 생성해도 연결이 유지됩니다.",
     "validationDocumentLabel": "문서(선택 사항)",
+    "validationDocumentOption": "{{title}}({{type}})",
     "validationLoadingLink": "연결 정보를 불러오는 중…",
     "validationNoProject": "-- 연결되지 않음 --",
     "validationProjectHint": "프로젝트를 비우면 이 양식의 연결이 해제됩니다.",
--- a/voc-datalake/frontend/public/locales/pt/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/pt/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "Opcionalmente registre sobre qual projeto — e qual PRD ou PR/FAQ — este formulário coleta feedback. As avaliações reunidas aparecerão ao lado desse documento na página de Priorização. Deixe ambos vazios para um formulário independente, como uma pesquisa de site.",
     "validationDocumentHint": "Mantenha no projeto inteiro para preservar o vínculo quando um documento for regerado.",
     "validationDocumentLabel": "Documento (Opcional)",
+    "validationDocumentOption": "{{title}} ({{type}})",
     "validationLoadingLink": "Carregando o vínculo…",
     "validationNoProject": "-- Não vinculado --",
     "validationProjectHint": "Limpar o projeto desvincula este formulário.",
--- a/voc-datalake/frontend/public/locales/zh/feedbackForms.json
+++ b/voc-datalake/frontend/public/locales/zh/feedbackForms.json
@@ -75,6 +75,7 @@
     "validationDescription": "可选择记录此表单收集的反馈针对哪个项目，以及哪个 PRD 或 PR/FAQ。收集到的评分随后会显示在优先级排序页面上该文档的旁边。对于网站调查等独立表单，请将两项留空。",
     "validationDocumentHint": "保留为整个项目，可在文档重新生成后仍保持关联。",
     "validationDocumentLabel": "文档（可选）",
+    "validationDocumentOption": "{{title}}（{{type}}）",
     "validationLoadingLink": "正在加载关联信息…",
     "validationNoProject": "-- 未关联 --",
     "validationProjectHint": "清空项目将解除此表单的关联。",
--- a/voc-datalake/frontend/src/i18n/config.ts
+++ b/voc-datalake/frontend/src/i18n/config.ts
@@ -11,8 +11,7 @@ import i18n from 'i18next'
 import LanguageDetector from 'i18next-browser-languagedetector'
 import HttpBackend from 'i18next-http-backend'
 import { initReactI18next } from 'react-i18next'
-import { LOCALE_LOAD_PATH } from './loadPath'
-import { supportedLanguages } from './languages'
+import { I18N_INIT_OPTIONS } from './options'
 
 // Language constants and the change helper live in ./languages (side-effect
 // free) so UI components can import them without triggering this module's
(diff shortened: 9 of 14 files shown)


perrozzi:

This PR appends the document type to each option in the Feedback Forms "Validates" document picker (`Title (TYPE)`), sourcing the type label from `SCORABLE_TYPE_META`, namespace-qualifies that metadata's `i18nKey` so it resolves when read through a `feedbackForms`-bound `t`, and adds a title fallback to the document id. The change is small, well-motivated, and comes with tests that actually pin the new behaviour (including a cross-namespace resolution gate). No blockers; a few things worth tightening around i18n.

### Should fix

- **The label format itself is a hardcoded, untranslated string.** `documentOptionLabel` in `voc-datalake/frontend/src/pages/FeedbackForms/ValidationLinkPicker.tsx` builds `` `${name} (${typeLabel})` `` in code. Per the repo's i18n rule, user-visible text — including the composition/punctuation — should come from a key, e.g. `feedbackForms:editor.validationDocumentOption` with `{{title}}`/`{{type}}` interpolation, added to every locale. That also lets locales that don't use ASCII parentheses (or want a different order, cf. `CheckboxItem`) adapt without a code change.
- **Nothing prevents the `i18nKey` regression the PR just fixed.** `SCORABLE_TYPE_META` in `pages/Prioritization/prioritizationUtils.ts` still types `i18nKey: string`, so dropping the `prioritization:` prefix stays a valid compile. Tighten to a template literal type (`` readonly i18nKey: `prioritization:${string}` ``) so the invariant is enforced by `tsc` and the new test becomes a second line of defence rather than the only one.
- **Confirm `nsSeparator` is not disabled in the i18n config.** Qualified keys (`prioritization:docType.prd`) only resolve if `:` is still the namespace separator; if any locale/i18n init sets `nsSeparator: false` (a common workaround for keys containing colons), both the Prioritization badge and this select would render the raw path. The diff doesn't touch the i18n init, so please verify — and ideally have the new gate in `prioritizationUtils.test.ts` run against the app's real i18n instance/config rather than a bare `i18next` import, so a config change would break it.
- **The stored/placeholder option is now inconsistent with the list.** `storedOptionLabel` still renders the bare title (or the unverified label) while every fetched option carries `(TYPE)`. For a link whose document is still loading or no longer in the list, the admin sees a differently-shaped label for the same kind of thing; consider routing it through `documentOptionLabel` when a `ProjectDocument` is available.

### Nits

- The missing-title → `document_id` fallback is now duplicated with `DocumentsTab`'s equivalent; a shared `documentDisplayName(doc)` helper would keep the two from drifting.
- `expect(screen.queryByText(/Research Notes/)).not.toBeInTheDocument()` — `queryBy*` throws on multiple matches, so if that text ever appears twice the test fails with a confusing "found multiple elements" rather than the intended assertion. `queryAllByText(...)` + `toHaveLength(0)` is more robust.
- The two new `mockGetProject.mockResolvedValue({...})` fixtures omit `title` / other `ProjectDocument` fields; if the mock is typed this will only stay compiling by accident. A small `as unknown as Project` with a comment, or a fixture factory, makes the "legacy record" intent explicit.
- Comment volume in `ValidationLinkPicker.tsx` and the test file is high relative to the code changed (≈20 lines of prose for a 10-line function); trimming to the load-bearing "why" would help future readers.
- The type-label fallback to the raw `document_type` slug is described as unreachable; a one-line `// istanbul ignore` style note or a test would clarify whether coverage gates are expected to flag it.

---

perrozzi:
## Round 1 disposition

Thanks — three of the four "should fix" items were real. Taken:

**1. The label format is untranslated.** Correct, and worse than a style point: this catalogue already writes CJK labels with full-width parentheses (`ドキュメント（任意）`, `文档（可选）`) while Korean uses half-width with no space, so `` `${name} (${type})` `` printed the wrong punctuation into three locales. Now `editor.validationDocumentOption` = `{{title}} ({{type}})`, added to all eight catalogues with per-locale punctuation (ja/zh full-width, ko half-width). A locale can reorder the two parts.

**2. Nothing prevented the `i18nKey` regression.** Right — as a plain `string` field, dropping the prefix stayed a valid compile with one test as the only defence. Now `` readonly i18nKey: `prioritization:${string}` ``. The runtime gate stays deliberately: vitest goes through esbuild and does not typecheck, so the type alone would never fail a suite. Confirmed both directions — un-qualifying now fails `tsc` (exit 2) and is invisible to vitest.

**3. `nsSeparator`, and the gate testing the wrong instance.** Verified: `nsSeparator` is not set anywhere in the repo, so it is i18next's default `':'`. Your second half was the sharper point — the gate asserted against the **test** i18n instance (`src/test/setup.ts`), so disabling the separator in the **app** config would have left it green. Fixed properly: the init options move to a new side-effect-free `src/i18n/options.ts` (the same split `src/i18n/languages.ts` already documents in that directory, because importing `config.ts` executes `init()` with an HTTP backend), and the test now asserts on the object the app actually hands to `init()`. Mutating `nsSeparator: false` in the app config now fails that test.

**Nit taken:** `queryAllByText(...)` + `toHaveLength(0)`, so a duplicate match names the leak instead of reporting "found multiple elements".

### Declined, with reasons

**4. The stored/placeholder option is inconsistent with the list.** It cannot be routed through `documentOptionLabel`, because in every state where that option renders there is no `ProjectDocument` to label. `storedOptionState` returns non-null only when the stored id is **absent from the fetched list** — still loading, request failed, or genuinely gone — and the labels are then "Loading link…" / "no longer available" / "could not be checked", never a title. When the stored id *is* in the list, the state is `null`, no extra option is rendered, and the admin sees the real option with its `(TYPE)`. So the shapes differ exactly when the type is unknown.

**Nit: shared `documentDisplayName(doc)` with `DocumentsTab`.** `DocumentsTab`'s equivalent operates on a *revision* record (`{title, revisedFromId}`) and falls back to a different field, so a shared helper would have to span two shapes for two call sites. Left as is.

**Nit: the fixtures omit `title`.** That omission *is* the scenario — a legacy record without one. `mockGetProject` is an untyped `vi.fn()`, so nothing compiles by accident, and the repo's standing rule is type guards over `as`, which rules out the `as unknown as` suggestion. The intent is stated in a comment at the fixture.

**Nit: comment volume.** Matches these files — `ValidationLinkPicker`'s existing `storedOptionState`/`storedOptionLabel` docblocks run 15–20 lines each for similarly small functions.

**Nit: coverage note on the unreachable fallback.** No coverage gate runs on this repo, so an ignore pragma would be inert; the docblock already names it unreachable-under-`isScorable` and says why it is kept.

### Verification on this round

2459/2459 frontend tests, `tsc` and eslint clean, `i18n-check.mjs` reports no missing/extra/empty keys and the same 633 unreferenced. Three more mutations, each caught and the file restored: `nsSeparator: false` in the app config, the en format losing `({{type}})`, and an un-qualified `i18nKey`. Each asserted a green baseline first.

perrozzi:

Since my last look the change picked up the two i18n items I flagged: the label composition is now a `feedbackForms:editor.validationDocumentOption` key present in all eight locales (with locale-appropriate full-width/half-width parentheses for ja/ko/zh), the init options were extracted to a side-effect-free `src/i18n/options.ts` so a test can assert the app's real `nsSeparator`, and `SCORABLE_TYPE_META.i18nKey` is now a `` `prioritization:${string}` `` template literal type. Nice, focused follow-up.

### Resolved
- Hardcoded `` `${name} (${typeLabel})` `` replaced by an interpolated key added to de/en/es/fr/ja/ko/pt/zh.
- `i18nKey` invariant now enforced by `tsc` via the template literal type, with the runtime gate kept as a second line of defence.
- `nsSeparator` verified and pinned against the actual `I18N_INIT_OPTIONS` (plus the test instance) in `prioritizationUtils.test.ts`.
- `queryByText(/Research Notes/)` → `queryAllByText(...).toHaveLength(0)`.

### Still open
- `storedOptionLabel` in `ValidationLinkPicker.tsx` still renders the bare title while every fetched option carries `(TYPE)`.
- Nits from the previous review: duplicated title→id fallback vs `DocumentsTab`, untyped `mockGetProject` fixtures, comment volume, unreachable-fallback coverage note.

### New findings
- **Verify no tooling reads the namespace list out of `src/i18n/config.ts`.** `voc-datalake/frontend/src/i18n/options.ts` now owns `NAMESPACES`/`DEFAULT_NS`; if `scripts/i18n-check.mjs` (or any test/build step) locates the namespace array by parsing `config.ts`, the move silently changes what it sees. Please re-run `i18n-check.mjs` against this revision, not the pre-move one.
- `DEFAULT_NS` and `NAMESPACES` are exported from `i18n/options.ts` but nothing in the diff imports them (`config.ts` reads only `I18N_INIT_OPTIONS`); either drop the exports or consume them so they don't drift as unused public surface.
- Nit: `I18N_INIT_OPTIONS` is a shared mutable object that `i18next.init()` receives directly; passing a spread (`init({ ...I18N_INIT_OPTIONS })`) would guarantee the test reads the declared config rather than anything i18next may normalise in place.

---

perrozzi:
## Round 2 disposition

All three new findings addressed — two fixed, one verified.

**1. Verify no tooling reads the namespace list out of `config.ts`.** Checked, and nothing does. The namespace list has four independent copies and always did: `src/i18n/options.ts` (moved from `config.ts`), `src/test/setup.ts`, `scripts/i18n-check.mjs` and `scripts/fix-i18n.mjs` — each hardcodes its own array. `i18n-check.mjs` never reads app source for it (`NAMESPACES` is a literal at the top of that script), and `src/i18n/localeParity.test.ts` walks `public/locales/` rather than parsing config. The only importer of `i18n/config` is `src/main.tsx`, for the init side effect, unchanged. And `i18n-check.mjs` was re-run **after** the move — on `dd882c4`, not the pre-move revision: no missing, extra or empty keys, and the same 633 unreferenced. Noted the four copies in the new module's comment so the next reader does not assume it is the source of truth.

**2. `DEFAULT_NS`/`NAMESPACES` exported but unused.** Dropped the exports; they are module-private consts again, exactly as they were in `config.ts`. `I18N_INIT_OPTIONS` is the module's only export.

**3. Nit: pass a spread to `init()`.** Taken — `init({ ...I18N_INIT_OPTIONS })`. The comment records the limit: shallow, so it covers every top-level option including `nsSeparator`, while `backend`/`detection`/`interpolation`/`react` are still shared objects.

### Held from round 1

**`storedOptionLabel` still renders the bare title.** It does not render a title at all, in any state — this is a factual disagreement rather than a judgement call, so quoting the code:

```ts
if (storedId === '') return null
if (availableIds.includes(storedId)) return null   // in the list -> no extra option
if (list.resolved) return 'missing'
return list.canResolve ? 'pending' : 'unverified'
```

The extra option is rendered **only** when `storedOptionState` is non-null, and that requires the stored id to be absent from the fetched list. Its three labels are `validationLoadingLink`, `validationUnknown*` and `validationUnverified*` — "Loading link…", "no longer available", "could not be checked". When the stored id *is* in the list, the state is `null`, no extra option is rendered, and the admin sees the real option carrying `(TYPE)`. So there is no state in which a document has a known type and is labelled without it: the shapes differ exactly when there is no `ProjectDocument` to read a type from. `documentOptionLabel` takes a `ProjectDocument`, so it cannot be applied there.

The remaining nits are held with the reasons given in the round 1 disposition.

### Verification on this round

2459/2459 tests, `tsc` and eslint clean over all of `src/`. All three mutations from the previous round re-run after these edits and still caught: `nsSeparator: false` in the app config, the en format losing `({{type}})`, and an un-qualified `i18nKey` (caught by `tsc`, exit 2 — vitest cannot see it).

perrozzi:

Since my last review the only changes are the two small items I raised on `src/i18n/options.ts`: `DEFAULT_NS`/`NAMESPACES` are now module-private with a comment noting the other copies of the namespace list (`src/test/setup.ts`, `scripts/i18n-check.mjs`, `scripts/fix-i18n.mjs`) are unaffected by the move, and `config.ts` now passes a spread into `init()` with the shallow-copy caveat documented. Both are clean; nothing new introduced.

### Resolved
- `DEFAULT_NS`/`NAMESPACES` no longer exported as unread public surface.
- `init({ ...I18N_INIT_OPTIONS })` keeps i18next from mutating the object the test asserts on.
- Namespace-list ownership question answered: the tooling scripts keep their own copies, so the move out of `config.ts` doesn't change what `i18n-check.mjs` sees.

### Still open
- `storedOptionLabel` still renders the bare title while every fetched option carries `(TYPE)`.
- Prior nits: duplicated title→id fallback vs `DocumentsTab`, untyped `mockGetProject` fixtures, comment volume, unreachable-fallback coverage note.

No new findings — just confirm nothing outside the diff imported `NAMESPACES`/`DEFAULT_NS` from `i18n/options.ts` (a `tsc --noEmit` pass covers it).

---

perrozzi:
## Round 3 — confirmed, closing the review loop

Your one ask: `tsc --noEmit -p tsconfig.app.json` passes clean on `9ac0498`, and a repo-wide grep finds no reference to `NAMESPACES` or `DEFAULT_NS` outside `src/i18n/options.ts` itself. Nothing imported them.

Three rounds, and across all of them the production-code changes the review produced were: one translation key (with its eight catalogue entries), one template-literal type, one module split, and one spread. No behaviour was wrong; everything found was about making the invariants enforceable rather than assumed. That is a good outcome for a review loop and a reasonable place to stop.

**One correction I'd like on the record**, since it has now been listed as open three times: `storedOptionLabel` does not render a bare title — it renders no title in any state. The extra option exists only while `storedOptionState` is non-null, which requires the stored id to be **absent from the fetched list**, and its three possible labels are "Loading link…", "no longer available" and "could not be checked". When the id *is* in the list the state is `null`, no placeholder is rendered, and the real option carries `(TYPE)`. `documentOptionLabel` takes a `ProjectDocument`; in the placeholder states there is no such record to pass it. Happy to be shown otherwise if I've misread it, but the code path above is the whole of it.

Final state: 2459/2459 frontend tests, `tsc` and eslint clean, `i18n-check.mjs` with no missing/extra/empty keys. Seven mutations run across the three rounds, each caught by a named test (or by `tsc`) and each file restored afterwards, with a green baseline asserted first every time.

Deploying to the live environment next and will report the browser check here.

perrozzi:
## Deployed and live-verified

Deployed to the live environment (us-east-1) at head `9ac0498` and tested in a browser. This closes the "no live browser check" gap noted in the PR description.

### The picker, live

The live environment turned out to contain the exact case this change is for: project **Q1 Products improvements** has a PRD and a PR/FAQ that are both titled *"real time delivery tracking"*. Before this change the select offered two options with identical text — genuinely unpickable. Read straight off the deployed `#validation-document` element (form *CSAT Survey*, Validates tab):

| option value | option text |
|---|---|
| `""` | `-- Whole project --` |
| `prd_20260624155112` | `real time delivery tracking (PRD)` |
| `prd_20260625231508` | `PR6 Regression PRD - support automation (PRD)` |
| `prfaq_20260624154929` | `real time delivery tracking (PR/FAQ)` ← stored link, still selected |
| `prfaq_20260710080135` | `Automated Insights Digest (PR/FAQ)` |
| `prfaq_20260811225625` | `PR317 derivation live verification (PR/FAQ)` |

Matches what the API says the project holds, option for option. The project has 13 documents; the 8 prototypes are correctly absent (`isScorable`), and the stored link round-tripped through the editor unchanged.

### The locale half

The part only a browser can confirm — switching to Japanese re-renders the same list through `{{title}}（{{type}}）`:

```
real time delivery tracking（PRD）
real time delivery tracking（PR/FAQ）
```

5 of 5 options with full-width parentheses, 0 with ASCII, consistent with the label beside them (`ドキュメント（任意）`). This is what the hardcoded `` `${name} (${type})` `` would have got wrong. Locale restored to English afterwards.

### Deploy verification

- Preflight: `origin/development` was still `56062d3` and this branch contains all of it, so the deploy could not roll back anyone else's work. Confirmed the target account/bucket/distribution before running a sync that uses `--delete`.
- **162/162** files in `dist/` byte-identical to the objects in S3 (md5 vs ETag). The comparator was positive-controlled first: a deliberately wrong digest must report `mismatch` and the right one `match`, otherwise the run is treated as meaningless — an exit code of 0 from the deploy script does not prove the sync ran.
- All eight `feedbackForms` catalogues served over CloudFront carry the new key with the right punctuation per locale.
- `validationDocumentOption` live in `assets/index-BinczST6.js`; `prioritization:docType.prd` live in `assets/prioritizationUtils-0OKd4AOT.js`, both fetched from the CDN, with an absent-string control.
- No console errors or warnings.

Nothing was mutated in the live environment: no Save was pressed, no form or document created or changed. One incident worth recording — an early attempt iterated over every button to find the tab and clicked "Sign out", which logged the session out; I logged back in and matched the tab by its catalogue label instead. No data was affected.

perrozzi:
## One more commit: `9339e50` — test-only

Added `src/i18n/options.test.ts` (119 lines, no production code) after noticing a gap the review rounds and I both left open: **the extraction made the options importable, but nothing asserted anything about them.** Since no test imports `config.ts` — it runs `init()` with an HTTP backend at module scope — every option in that literal was untested before the move and stayed untested after it. Dropping `supportedLngs` still boots the app, just without the guard that rejects a stale cached locale.

The load-bearing gate is the **namespace list**. A page calling `useTranslation('somewhere')` for a namespace absent from `ns` does not throw: i18next resolves nothing and the page renders raw key paths — the same failure class this PR already fixed twice. `scripts/i18n-check.mjs` structurally cannot catch it, because its own `NAMESPACES` array is a separate hardcoded copy, so it validates keys against the namespaces *it* knows rather than the ones the app registers. The test scans source for `useTranslation('x')`, `t('ns:key')` and data-held `*Key: 'ns:key'` and requires every namespace found to be registered.

Two things worth flagging from writing it:

- **A comment created a false positive.** The scan's first run reported a namespace called `ns`, from a docblock in `ValidationLinkPicker.tsx` that discusses the shape `xKey: 'ns:key'` — it reported the app as broken. `i18n-check.mjs` shares the blind spot and emits its "not a known namespace — typo?" warning for that same comment. The test now strips comments before scanning.
- **The extraction was confirmed lossless** by diffing the option keys `config.ts` passed inline at the branch point against the new module, positive-controlled. Note `defaultNS` was a shorthand property before (`defaultNS,`), so it is invisible to a `key:` regex and reads as newly added when it was there all along.

Verified: **2461/2461** tests, `tsc` and eslint clean. Three mutations, each caught and restored — unregistering `feedbackForms` while the page still uses it, adding `'navigator'` back to `detection.order`, and pointing a real `useTranslation` call at an unregistered namespace. The last one is the important one: it proves the scan reads code rather than only the namespaces the test names itself.

**No bot review arrived for this push** within 7 minutes, where the previous three came in 2–5. I have not diagnosed why and am not assuming it is the test-only diff — flagging it rather than reading silence as approval. The last production-code change in this PR is `9ac0498`, which the bot did review and closed with "No new findings"; `9339e50` touches one new test file and nothing else.

PR #325, opened 2026-08-13
feat(product-docs): uploads work for what we can actually process

perrozzi:
Rung 2 of the visual-grounding ladder. Product-doc uploads were inert: the boundary accepted pdf/docx/md/txt while **nothing extracted any of them**, the module docstring pointed at an extractor Lambda that had never existed, and every record sat at `status: 'pending'` forever with the UI badge reading "Extracting…" for the life of the record. This replaces *nothing works* with something truthful.

## What ships

- **Images accepted** (png/jpeg/gif/webp), by picker, drag-drop **and paste**.
- **`ALLOWED_CONTENT_TYPES` narrowed to exactly the processable set.** pdf and docx are refused at `create_upload_url` with an explicit "not supported yet" that names what *is* accepted — deliberately distinguishable from never-supported types, because a user who could previously upload a pdf (and get silence) needs to know which answer they got. They return at rung 4.
- **A new extractor Lambda**: images go to Bedrock for a structured description; `txt`/`md` pass through byte-identically. `ready` and `failed` are reachable for the first time.
- **Browser-side resize**: ≤1568px long edge, PNG first, JPEG only as a fallback.
- **A server-side stalled-extraction transition**, because the client cannot write `failed`.

## Three decisions worth a reviewer's time

**The image cap is 3.75 MB / 8000 px, and it is safe to hardcode despite the admin model picker.** These are limits of the Converse API's `Message` shape, not of any model — the constraint sits *above* the model, and every id in `ALLOWED_MODEL_IDS` is an Anthropic Claude whose own vision limits are looser-or-equal. So repointing any surface through the picker cannot move it. `3_750_000` is the decimal reading of an ambiguous "3.75 MB"; the binary reading (3_932_160) is higher, so the decimal one is correct under either interpretation.

**The extractor lives in core-stack, beside `rawDataBucket`.** CDK parents the notification resource under the *bucket* construct, so calling `addEventNotification` from the processing stack would make `VocCoreStack` reference a `VocProcessingStack` Lambda while processing already depends on core. Settled by synth, not by argument — see Verification.

**It resolves its model from an allowlist injected as env from `model-allowlist.ts`, rather than importing `shared/model_config.py`.** That import chain reaches `aws_lambda_powertools`, which needs a `LayerVersion`, which means container bundling inside CoreStack — a property that stack deliberately protects (its signing-key handler is written in Node precisely so CoreStack never needs Docker). The synth-injected form is also *stronger* than the repo's lockstep habit: the allowlist now has **zero** copies rather than a tested two. Verified in the template: the function has no `Layers` property.

## Two smaller things a reviewer should not miss

**The declared upload size is now binding.** `create_upload_url` signed only Bucket/Key/ContentType, so `MAX_FILE_BYTES` was checked against a number the **client** supplied and S3 enforced nothing — a client could declare one byte and PUT ten gigabytes. `ContentLength` is now signed into the presigned PUT. This also hardens the pdf/docx/txt path that had already shipped, so it is a fix rather than only scaffolding.

**Image dimensions need no new dependency.** PNG/GIF/WebP/JPEG headers are parsed with `struct` from a **ranged GET**, which does three jobs at once: the dimension cap, a real content-type sniff (a `.png` full of pdf bytes is caught, and the client's declared type is never trusted), and the corrupt-image→`failed` case. `imghdr` was not an option — it left the stdlib in 3.13.

**An empty or whitespace-only extraction ends `failed`, never `ready`.** Otherwise `build_product_context_block` silently contributes nothing while the record claims the document is usable.

## ⚠️ Behavioural knock-on, deliberate and stated plainly

Rung 1 (#322) wired product context into the prototype builder, and `build_product_context_block` injects the extracted text of **every** `ready` document. So from this merge, a visual's description starts reaching **every product-context-ticked prototype by itself** — under `### Internal documents`, mixed in with unrelated files, ordered by `created_at`, sharing one `MAX_EXTRACTED_INJECTION_CHARS` budget, and placed nowhere near the eight CSS variables the prompt actually acts on.

That changes prototype output for every project that has ever uploaded anything, with no review of its own. **Rung 3 is what turns it from a leak into an instruction**, by selecting visuals explicitly and placing them as a visual brief.

## Verification

Measured **by delta against a pre-change baseline**, not against green — two gates in this repo are already red for unrelated reasons.

| Gate | Before | After |
|---|---|---|
| backend pytest | 1658 | **1770** |
| `test:cdk` | 231 | **238** |
| frontend vitest | 2454 | **2476** |
| `test:stream` | 303 | 303 |
| `lint:python` | RED @ 553 | **RED @ 553, per-rule histogram byte-identical** |
| `i18n:check` | RED (extra 12 / untranslated 15 / 2 missing-in-English) | **RED, report byte-identical** |
| `cdk synth` | 5 templates, 0 warnings | **5 templates, 0 `[Warning at]`** |

`lint:python`'s 553 is pre-existing toolchain drift (`ruff.toml` intends the default rule set; the resolved ruff enables 413 rules). Seven new `.py` files contributed **zero** new findings, so the total is unchanged rather than coincidentally equal.

**Placement proven, not argued.** `Custom::S3BucketNotifications` materialises in `VocCoreStack.template.json`, filtered on `prefix: projects/`, targeting a Lambda in the same stack; `VocProcessingStack` contains zero references to it; no cyclic-dependency error; still exactly 5 templates (the app is at a hard five-template cap enforced downstream). Synth ran with no bypass flag.

**18 mutation probes, 18 caught, 0 survived**, every restore byte-identical, anchor counts asserted inside each probe so a stale anchor cannot report neutrally. One first-pass survivor was diagnosed as an **equivalent mutant** — two mutually redundant never-upscale guards, each unkillable alone, the pair killable — and fixed by *deletion*, because its comment credited the wrong guard, which is how the real guard gets removed later by someone tidying up.

**Verified in a real browser** (Playwright, 8/8): paste; both ladder rungs against real codecs, including a PUT body that decodes back to exactly 1568×627 and an 8.4 MB noise PNG falling through to the JPEG rung with its extension rewritten; the accepted-formats hint at five widths; the `ready`/`failed` badges rendering for the first time; and no raw i18n key visible in `en` or `de`. The resize unit tests fake every imaging primitive, so this is the only evidence the ladder works against actual browser codecs.

## Known, carried forward on purpose

- **A prompt-injection surface inherent to the feature.** The extraction prompt asks for labels *verbatim*, and that text reaches `build_product_context_block` → PRD / PR-FAQ / prototype prompts. Blast radius is self-inflicted (an authenticated user uploading to their own project, producing their own prototype), so it is not a blocker — but the defence is delimiting untrusted extracted text, which is a *placement* decision, and rung 3 owns placement.
- `extracted_chars` interpolates raw digits ("1404125 chars extracted") and wants a locale-aware number format.
- The static hint says "PNG, JPEG, GIF, WebP" while the refusal says ".gif, .jpg, …" — the same set derived from one map, but two vocabularies.
- Backend `error` strings pass through unlocalized, so a non-English user sees the English server message.

## Filing note

Per the plan of record, this rung is an **amendment to U26, not a new issue** — please do not file it separately. The `persona_importer` placeholder bug remains the one genuinely new issue on that ladder.

Rung 2 of `analysis/FEATURE-prototype-visual-grounding-GITHUB-ISSUE.md`. Follows #322.


perrozzi:

This PR adds real extraction for product-doc uploads: a new S3-triggered `product_doc_extractor` Lambda (image→Bedrock description, text pass-through), narrows `ALLOWED_CONTENT_TYPES` to the processable set with an explicit "deferred" answer for pdf/docx, signs `ContentLength` into the presigned PUT, adds browser-side downscaling, and introduces a read-path stalled→`failed` transition. Test coverage is unusually thorough (boundary, header parsing, model resolution, CDK placement, lockstep) and the i18n work covers all eight locales. **The diff was truncated — `lib/stacks/core-stack.ts` (the Lambda definition, IAM grants, notification wiring) is cut off, so this review is partial and does not cover the actual grants, KMS permissions, or the `model-allowlist.ts` additions.**

### Should fix

- **Extracted text reaches prompts undelimited (`lambda/product_doc_extractor/handler.py`).** `IMAGE_EXTRACTION_PROMPT` explicitly asks the model to reproduce all visible text *verbatim*, and that output flows through `build_product_context_block` into PRD / PR-FAQ / prototype-HTML prompts with no delimiting or provenance marker. The PR defers this to "rung 3 owns placement", but wrapping the extracted body in an explicit untrusted-content delimiter is a two-line change in the extractor/injection site and is exactly the "validate at system boundaries" rule. If generated prototype HTML is ever rendered outside a sandboxed iframe, treat this as a blocker rather than a should-fix.
- **Un-gated behavioural change to every existing project.** As the description states, from this merge every `ready` image description (up to ~4k tokens of markdown) is injected into every product-context-ticked prompt, sharing the `MAX_EXTRACTED_INJECTION_CHARS` budget with real documents. Rung 2 could ship without that side effect by filtering image content types out of `build_product_context_block` until rung 3 selects them deliberately — a much smaller blast radius than "prototype output changes for all projects, unreviewed".
- **Extractor deviates from the repo's Lambda pattern without a lockstep guard on the parts it re-implements.** No powertools Logger/Tracer, lazy `_clients` dict instead of module-level boto3 clients, and hand-copied `TEXT_CONTENT_TYPES` / `CONVERSE_IMAGE_FORMATS` / model resolution. The docstring claims "an unlisted type here means the two have drifted", but I see no test pinning `TEXT_CONTENT_TYPES | CONVERSE_IMAGE_FORMATS.keys()` against `product_context.ALLOWED_CONTENT_TYPES` — add one (the model default *is* pinned, so the pattern is already established). Also worth evaluating: enabling EventBridge notifications on `rawDataBucket` and putting the extractor in the processing stack would remove the cycle argument entirely and let it use the standard shared layer + powertools.
- **`resizeImage.ts` introduces a third, unpinned copy of the image caps.** `MAX_IMAGE_BYTES = 3_750_000` and `MAX_IMAGE_EDGE_PX = 1568` live only in `frontend/src/pages/ProjectDetail/resizeImage.ts`; `lambda/shared/test/test_image_limits_lockstep.py` pins `model-allowlist.ts` ↔ `image_limits.py` and `MAX_FILE_BYTES` ↔ `ProductDocsUpload.tsx`, but nothing pins this file. A client cap above the server cap silently reintroduces "upload appears to work, then 400s". Extend `_frontend_int_const` to cover it.
- **Late-completing extraction can clobber the stalled `failed` record.** `_fail_if_stalled` writes `failed` conditionally on the status it read, but the extractor's `_update_doc` only conditions on `attribute_exists(pk)`. An extraction that finishes after `EXTRACTION_STALL_SECONDS` (retry, cold-start pileup, Bedrock latency) will overwrite `failed` with `ready` — or `error: None` — even though the API already told the user to re-upload. Condition the extractor's terminal write on `#status IN (:pending, :extracting)` too.

### Nits

- `STALLABLE_STATUSES` includes `'extracting'`, but nothing in `handler.py` ever writes that status — the branch and its test are exercising a state the system cannot reach.
- `withSyntheticName` in `ProductDocsUpload.tsx` derives the extension from the MIME subtype, so a pasted JPEG becomes `pasted-….jpeg` while `ALLOWED_MIME` maps `image/jpeg → .jpg`; use `ALLOWED_MIME[file.type]` instead.
- The `file.size > MAX_FILE_BYTES` check in `uploadOne` runs *before* `resizeImageForUpload`, so a 12 MB screenshot that would resize fine is refused with "Too large (>10 MB)". Apply the general cap only to non-images.
- `int((body or {}).get('size_bytes') or 0)` in `create_upload_url` raises `ValueError` on a non-numeric body value → 500 instead of a 400 `ValidationError`. Pre-existing, but this field is now security-relevant.
- The drop zone `<label>` gains `tabIndex={0}` + an Enter/Space handler that calls `fileInput.current.click()`; on browsers that translate keyboard activation on a label into a click this can open the picker twice. Consider moving the handler to an explicit `role="button"` element with an `aria-label`.

---

perrozzi:
Thanks — this was a good review. **Nine of ten findings accepted and implemented, one declined with reasoning.** Two of them were real bugs I had missed rather than style points, so they are worth calling out first.

## The two that mattered

**Late extraction clobbering a stalled `failed` — a genuine race, correctly spotted.** I built the conditional write on the API side (`_fail_if_stalled` conditions on the status it read) and then left the extractor's terminal write conditioned only on `attribute_exists(pk)`. So an extraction finishing past `EXTRACTION_STALL_SECONDS` overwrote a `failed` the user had already been told to act on. Every extractor write is now also conditional on `#doc_status IN (:pending, :extracting)`, and a refused write is no longer swallowed identically to the deleted-record race — the record is re-read and the two cases log distinguishably, because "the API gave up on this document" is *also* the signal that extractions are running past the stall window.

**The un-gated behavioural change did not have to ship, and you were right that it shouldn't.** My PR description treated the leak as unavoidable. It isn't: `build_product_context_block` now excludes image content types entirely until rung 3 selects them deliberately. A visual is stored, listed and readable, but not injected — so the blast radius goes from "every project that ever uploaded anything, unreviewed" to zero, and rung 3 becomes purely additive. `generate_report`'s "anything to summarize" gate shares the same predicate, otherwise an image-only project passed the gate and then built a report from an empty block.

## Also accepted

- **Content-type maps pinned to the boundary.** The extractor docstring claimed "an unlisted type here means the two have drifted" with nothing testing it; that claim is now the specification.
- **`resizeImage.ts` was indeed a third, unpinned copy.** `_frontend_int_const` now takes a source path and pins the browser cap against both the Python mirror and the CDK source. `MAX_IMAGE_EDGE_PX` (1568) is deliberately *not* pinned — it is a quality/cost target with no server counterpart — and there is now a test recording that as a decision rather than an oversight, so nobody later equates it to the dimension cap.
- **`extracting` was unreachable, so I made it real rather than deleting it.** The extractor publishes it once before the slow work, which makes the badge the UI already renders meaningful and the stall branch reachable.
- **The pre-resize cap was a real UX bug.** The general cap now applies to non-images only; a 12 MB screenshot resizes instead of being refused. Two tests, because "scoped" and "removed" are otherwise indistinguishable.
- **`.jpeg` vs `.jpg`**, via a type guard rather than a cast.
- **Non-numeric `size_bytes`** now answers 400 like any other bad size. Agreed on the reasoning: pre-existing, but this field is signed into `ContentLength` now, so it is part of a security control.
- **Prompt-injection delimiting.** Injected bodies are fenced with an explicit untrusted-content marker plus a notice naming them as quoted material whose instructions must not be followed, and a document cannot close its own fence.
- **The drop zone.** Restructured to a `role="button"` element with the input as a *sibling*, so there is exactly one activation path per modality. Per spec a `<label>` has no keyboard activation behaviour and jsdom cannot prove the double-open either way — but the ambiguity is itself the defect, so tests now pin "exactly one picker open" for both keyboard and pointer.

## Declined, with reasoning

**EventBridge notifications on `rawDataBucket` plus moving the extractor to the processing stack.** Not doing this, for three reasons. Core-stack placement was an explicit instruction for this rung rather than an oversight. The cycle reasoning is not speculative — it is verified by synth, and `core-stack.ts` documents it at the construct. And the app is at a hard five-template cap enforced by a downstream packaging consumer, so "put it in another stack" is not a free move either. Swapping the notification mechanism, the stack, *and* the dependency story is a strictly larger change than the one under review. Recorded as a future option.

## Two notes where the truncated diff cost you context

You flagged that the diff was cut off before `core-stack.ts`, and it shows in two places:

1. The stack suggestion above reads as though the placement were an open engineering question. The cycle is documented at the construct and confirmed by synth.
2. **Your own escalation condition for the injection finding does not fire.** You wrote "if generated prototype HTML is ever rendered outside a sandboxed iframe, treat this as a blocker". I checked: `PrototypeRenderer.tsx` renders new prototypes cross-origin via `src=` and legacy ones via `srcDoc` with `sandbox="allow-scripts allow-popups allow-forms"`. It is sandboxed, so it stays a should-fix — which is how I treated it, and I implemented the delimiting anyway.

## Verification

Every gate re-run separately (chaining hides later failures). backend **1817** (was 1770), frontend **2482** (was 2476), `test:cdk` 238, `test:stream` 303, all unchanged-or-up. `lint:python` still exactly 553 with a **byte-identical per-rule histogram** — seven new findings appeared first and were fixed at source rather than suppressed. `i18n:check` unchanged, report section byte-identical. Synth: 5 templates, zero `[Warning at]`, no bypass flag.

**13 mutation probes over the A1/A2/A4/A5/A10 guards: 13 caught, 0 survived, 0 skipped**, every restore byte-identical, anchor counts asserted inside the probe so a stale anchor cannot report a false clean.

One probe finding worth passing on: my first attempt at the `extracting`-ordering mutation moved the write three lines but still before the extraction, and it *correctly* survived as an equivalent mutant. Redesigned to move it past the work, with a companion probe that re-adds it afterwards. Both caught.

perrozzi:

Since my last pass the author added the two structural fixes I asked for on the server side (an untrusted-content fence around injected document bodies, and an `_injectable_docs` filter that keeps image descriptions out of every prompt until rung 3), hardened `_update_doc` with a non-terminal status condition so a late extraction can no longer clobber a stalled `failed`, made `extracting` actually reachable, validated `size_bytes` defensively, pinned `resizeImage.ts`'s cap in the lockstep test, added a content-type lockstep between the extractor and the upload boundary, and restructured the drop zone to a `role="button"` div with a sibling input. Note again that **the full-context diff was truncated** — `lib/stacks/core-stack.ts` (Lambda definition, IAM grants, KMS, notification wiring) and `model-allowlist.ts` are still unreviewed, so this remains a partial review.

### Resolved
- Extracted bodies are now fenced (`UNTRUSTED_DOC_BEGIN/END` + notice, with `_fenced` preventing self-closing) — `lambda/api/product_context.py`.
- Un-gated behavioural change removed: `_injectable_docs` excludes images from injection *and* from the `generate_report` gate, with a discriminating fixture in `test_product_context_injection.py`.
- Content-type drift now enforced both directions by `lambda/product_doc_extractor/test/test_content_type_lockstep.py`.
- Third copy of `MAX_IMAGE_BYTES` pinned by `TestBrowserImageCapLockstep`; `MAX_IMAGE_EDGE_PX`'s deliberate non-pinning documented and tested.
- Late extraction can no longer overwrite a stalled `failed`: `_update_doc` conditions on `#doc_status IN (:pending, :extracting)`, with the refusal classified and logged distinctly.
- `extracting` is now written (`_mark_extracting`), before the slow work, on both branches — the badge and `STALLABLE_STATUSES` are no longer dead.
- `withSyntheticName` uses `ALLOWED_MIME` (pasted JPEG → `.jpg`); general 10 MB cap scoped to non-images; `_declared_size` turns a non-numeric `size_bytes` into a 400 instead of a 500; single activation path for the picker with an accessible name.

### Still open
- The extractor's broader deviation from the repo Lambda pattern (no powertools Logger/Tracer, lazy `_clients`) and the unexplored alternative of EventBridge notifications + processing-stack placement — only the lockstep half of that finding was addressed.

### New findings
Nothing blocking. Small items in the incremental changes:

- `lambda/api/test/test_product_context_injection.py`: `patch.object(shared.converse, 'converse', ...)` only intercepts attribute lookups on the module. If `product_context` binds the name at import (`from shared.converse import converse`), the stub is inert and the comment's "cannot reach the network even if the gate is removed" claim does not hold — patch `product_context.converse` instead (or whatever the local binding is).
- `lambda/api/product_context.py`: `generate_report` now calls `_injectable_docs` and then `build_product_context_block`, which calls it again — two extra `_list_doc_items` queries per report on the field-less path. Passing the list through would avoid it.
- `_declared_size` truncates floats (`1000.7 → 1000`), so a client sending a fractional size signs a `ContentLength` one byte off the body it PUTs and gets an opaque S3 rejection; rejecting non-integral numbers would give the same 400 as the other unusable sizes. Also worth typing the parameter (`value: object`) given the module's Zod/type-guard-equivalent conventions elsewhere.
- `lambda/product_doc_extractor/handler.py`: a record that exists but carries no `status` attribute fails the new condition and is logged as "already None; refusing to overwrite … gave this document up as stalled", which names the wrong cause. Minor, but the refusal line is explicitly a signal about stall timing.
- `ProductDocsUpload.tsx`: per ARIA APG a `role="button"` should activate Space on key*up*; activating on `keyDown` will also fire on key-repeat if the user holds Space. Cosmetic.

---

perrozzi:
Round 2 actioned. **Three of the five nits fixed, one declined, one verified as a false positive.** This is the last round from my side — round 2 was already "nothing blocking", and none of what follows changes that.

## Fixed

**The float truncation — the one real defect in the round, and you were right that it mattered.** `int(1000.7)` silently became 1000, so a fractional declared size signed `ContentLength: 1000` and then PUT 1001 bytes, turning a clean 400 into an opaque S3 rejection inside a security control. `_declared_size` now parses through `Decimal(str(value))` and requires the result to be finite and integral. Accepts `1000`, `1000.0`, `'1000'`, integral `Decimal`; rejects `1000.7`, `'1000.7'`, `nan`, `inf`, and — explicitly — `True`, which as an `int` subclass was quietly becoming a one-byte file. Every rejection still returns `None` so the caller raises the identical `ValidationError`; no second message and nothing new to translate. `Decimal` cannot reach the function from the single caller (a `json.loads` body never produces one) but is handled at zero cost, and the docstring says so rather than implying a live path.

**The refusal log no longer blames stall timing for a malformed record.** It now classifies three causes: gone (INFO), genuinely terminal (the stall warning, wording unchanged), and missing/unrecognised status (a malformed-or-legacy warning that deliberately shares *no* stall vocabulary — my first attempt at that wording still contained the phrase an operator would grep for, and my own test caught it). Worth noting the existing terminal-case test was not modelling the race at all: its fixture re-read `pending` throughout, so it now re-reads `failed` after the opening read.

**Space activates on keyup.** Good catch that this is not cosmetic — the held-key case is the point. Enter stays on keydown, Space keydown only calls `preventDefault()` for scroll, activation moved to `onKeyUp`. The test that makes it real: four keydowns with no keyup open the picker **zero** times, then one on release.

## Declined

**The double `_injectable_docs` call.** Correctly observed, deliberately left. It is two DynamoDB reads immediately before a multi-second Bedrock call, so the cost is unmeasurable, and threading the list through would couple two functions that currently share only a predicate. Recorded in the commit message.

## Verified as a false positive

**`patch.object(shared.converse, 'converse', ...)` is effective, and the suggested fix would be worse.** `product_context.py` does `from shared.converse import converse` *inside* `generate_report`, so the name resolves on the module at call time and there is no import-time binding to patch. `patch.object(product_context, 'converse')` would target an attribute that does not exist and raise `AttributeError`.

I proved it rather than asserting it. With the report gate removed, two arms:

- **stubbed**: intercepted, one stub call, `network_attempted: false`
- **no-stub control**: the real `converse` ran, logged five Bedrock attempts, and tripped a socket tripwire inside botocore's endpoint resolution

So the stub is load-bearing, not decoration. The test comment now records the mechanism and warns against "fixing" the target.

Two small corrections while I was in there: the log line you quoted is split across three f-string fragments in source, which is why a grep for it as one string finds nothing. And the resolved list credits `_declared_size` with turning a *non-numeric* size into a 400 — true, but the same function was mishandling a *numeric* one, so that resolution note and the float finding are the same function seen from two directions.

## Verification

backend **1833** (was 1817), frontend **2483** (was 2482), `test:cdk` 238, `test:stream` 303 — all unchanged-or-up. `lint:python` still exactly 553 with an identical per-rule histogram; it briefly went to 554 when a `Decimal('1000')` test literal tripped FURB157, caught by a per-file HEAD-vs-worktree ruff diff and fixed at source. `i18n:check` unchanged. Synth 5 templates, zero `[Warning at]`.

**8 mutation probes, 8 caught, 0 survived, 0 skipped**, every restore byte-identical: three on the integral check (including a half-fix that keeps NaN/string rejection but still truncates fractions), two on the log classification, three on the keyboard handling.

Running total across the rung: **39 mutation probes, all caught.**

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated, so `lib/stacks/core-stack.ts` (Lambda definition, IAM grants, KMS, S3 notification wiring), `lib/utils/model-allowlist.ts` and `lambda/shared/image_limits.py` remain unreviewed, as in my previous pass.

Since the last review the author moved Space activation to `keyup` (APG button pattern) with a key-repeat regression test, tightened `_declared_size` to require an *integral* value via `Decimal` (rejecting `bool`, NaN/±Inf, fractional strings) with a parametrized test class, split the refused-write log into three causes via a new `TERMINAL_STATUSES` tuple plus a `_reread_as` test helper, and documented the `patch.object(shared.converse, 'converse', …)` target as verified rather than assumed. Four of my five small items are addressed; the incremental changes look clean.

### Resolved
- `_declared_size` no longer truncates floats (`1000.7` → 400 instead of a mismatched signed `ContentLength`), is typed `value: object`, and rejects `True` — `lambda/api/product_context.py`.
- `keyDown`-on-Space repeat fixed: Space activates on `keyUp`, `preventDefault` still on keydown, with an explicit held-key test — `ProductDocsUpload.tsx` / `.component.test.tsx`.
- A record with a missing/unrecognised `status` is now logged as malformed rather than blamed on stall timing, with a grep-isolation assertion — `lambda/product_doc_extractor/handler.py`.
- The `shared.converse` patch target is now justified (function-local import), so the stub is genuinely load-bearing.

### Still open
- Extractor's deviation from the repo Lambda pattern (no powertools Logger/Tracer, lazy `_clients`) and the unexplored EventBridge-notification/processing-stack placement.
- `generate_report` still calls `_injectable_docs` and then `build_product_context_block`, which re-queries — two extra `_list_doc_items` per report on the field-less path.

### New findings
- `lambda/api/product_context.py` — `_declared_size` converts before bounding magnitude: `Decimal(str('1E+999999999'))` is finite and integral, so `int(number)` materialises a ~400 MB integer inside the request path before the cap check ever runs. Compare against `max_bytes` in the `Decimal` domain (or reject `number.adjusted() > ~10`) before calling `int()`.
- `lambda/product_doc_extractor/handler.py` — `NON_TERMINAL_STATUSES` and `TERMINAL_STATUSES` are now two independent tuples with no lockstep to the API's status vocabulary; a future legitimate status (e.g. `queued`) added in `product_context.py` would be logged as "malformed or predates the status field", i.e. the same wrong-cause problem this change fixed, one status later. A test asserting the union equals the API's status set would pin it.
- `ProductDocsUpload.tsx` — the Space `keyUp` handler activates on any Space release, including one with no matching keydown on this element (focus arriving mid-keypress). The usual guard is a flag set in the Space keydown; cosmetic, but it is the same class of stray-activation issue the keyup move was made to fix.

---

perrozzi:
All three actioned. **The first was a regression the previous round's own fix introduced, so thank you for catching it — that one was real.** This is my last round; the loop has converged (should-fix → nits → one regression + two hardening items) and nothing here is blocking.

## The regression

Requiring an *integral* size moved parsing from `int()` to `Decimal`, and that reopened a door `int()` had been closing by accident. `'1E+999999999'` is twelve bytes of request body: `Decimal` parses it, it is finite, it *is* integral — so it passed every check, and `int()` then materialised a billion-digit integer inside the request path before the cap was consulted. The original `int(value or 0)` raised `ValueError` on exponent notation and answered 400; my tightened version did not.

I measured it rather than assuming, and the numbers make the case better than the argument does:

| | |
|---|---|
| `int(Decimal('1E+100000'))` | 296 ms |
| `int(Decimal('1E+300000'))` | 2,665 ms |
| `int(Decimal('1E+1000000'))` | **29,967 ms** |
| the `Decimal` comparison that now refuses it | **0.0006 ms** |

Superlinear, so a billion digits is a Lambda timeout many times over rather than a slow request.

The magnitude is now bounded **in the `Decimal` domain, before `int()`** — comparison reads the exponent and coefficient instead of expanding them. Your phrasing was exact: bounding after the conversion would be no guard at all, because the conversion *is* the cost, which is also why the caller's own `max_bytes` check runs too late to help.

Two things beyond what you specified. **Both ends are bounded**, because `-1E+999999999` is below every cap and an upper bound alone still hands it to `int()`. And I checked whether a long numeric string needs its own limit: it does not — parsing is linear and cheap (24 ms for 10 million characters), and only `int()` is not.

## The other two

**Status tuples pinned.** You were right that this is the same wrong-cause bug one status later. `product_context.py` now owns a canonical `PRODUCT_DOC_STATUSES` and a lockstep test pins the extractor's two tuples to partition it, so adding a status in one place and not the other fails a test instead of mislabelling a well-formed record as malformed.

**Stray Space activation closed.** A ref armed by this element's own Space keydown and consumed on keyup, plus an `onBlur` disarm so a release that happens elsewhere cannot be spent on the next press here.

## Still declined, unchanged

The extractor's deviation from the powertools pattern together with the EventBridge/processing-stack alternative — core-stack placement was an explicit instruction, the cycle is synth-verified, and the app is at the five-template cap. And the double `_injectable_docs` call: two DynamoDB reads before a multi-second Bedrock call, where threading the list through would couple two functions that share only a predicate.

## Verification

backend **1853** (was 1833), frontend **2486** (was 2483), `test:cdk` 238, `test:stream` 303 — unchanged-or-up. `lint:python` still exactly 553 with a byte-identical per-rule histogram; `i18n:check` unchanged; synth 5 templates, zero `[Warning at]`.

**5 mutation probes, 5 caught, 0 survived.** The one worth reporting: moving the bound to *after* `int()` leaves every return value identical, so a return-value-only suite passes it — it was caught by **hanging**, which is the vulnerability demonstrating itself. That is why the fix carries a time-budget assertion (`1E+300000` against a 0.5 s budget, ~5x margin) rather than only an assertion about the response.

One harness note, since it is the more dangerous failure of the two: the first version of that probe had no per-run timeout and was killed mid-mutation, leaving the working tree **with this very security fix deleted** and only a backup holding it. It now has a timeout and restores in a `finally`. A probe that can leave the tree mutated is worse than the bug it is checking for.

Running total across the rung: **44 mutation probes, all caught.**

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated, so `lib/stacks/core-stack.ts`, `lib/utils/model-allowlist.ts` and `lambda/shared/image_limits.py` remain unreviewed, as in both previous passes.

Since the last review the author added the Decimal-domain magnitude bound in `_declared_size` (with a booby-trapped `Decimal` subclass, a vacuity guard and two timing budgets), introduced `PRODUCT_DOC_STATUSES` in `product_context.py` plus `test_status_lockstep.py` pinning the extractor's two tuples against it, and added a `spaceArmed` ref (set on keydown, consumed on keyup, cleared on blur) with three new component tests. All three of my previous new findings are addressed; the incremental changes are clean.

### Resolved
- `_declared_size` now bounds magnitude before `int()`, so `'1E+999999999'` no longer materialises a ~400 MB integer in the request path — `lambda/api/product_context.py`.
- The extractor's status partition is pinned to the API's canonical vocabulary (coverage, disjointness, non-emptiness, `STALLABLE ⊆ NON_TERMINAL`) — `lambda/product_doc_extractor/test/test_status_lockstep.py`.
- Space `keyUp` now requires a matching keydown on this element, with the flag cleared on activation and on blur — `ProductDocsUpload.tsx`.

### Still open
- Extractor's deviation from the repo Lambda pattern (no powertools Logger/Tracer, lazy `_clients`) and the unexplored EventBridge-notification/processing-stack placement.
- `generate_report` calls `_injectable_docs` and then `build_product_context_block`, which re-queries — two extra `_list_doc_items` per report on the field-less path.

### New findings
- `lambda/api/test/test_product_doc_upload_boundary.py` — two wall-clock assertions (`< 0.5s` for `1E+300000`, `< 1.0s` for a 10 M-digit string) will flake on a contended CI runner even with the guard in place; the `_NeverConvertibleDecimal` tests already prove the invariant, so consider widening these budgets substantially or marking them non-blocking.
- `lambda/api/product_context.py` — the new bound hardcodes `MAX_FILE_BYTES` as "the widest cap any caller can accept", which silently assumes `MAX_IMAGE_BYTES <= MAX_FILE_BYTES`; that ordering is only incidentally pinned (by an inline `assert` inside an unrelated message test), so raising the image cap above the file cap would reject legal image uploads with no dedicated failure. A one-line assertion in the caps test class would pin it.
- `lambda/api/product_context.py` — `PRODUCT_DOC_STATUSES` claims to be mirrored in `frontend/src/api/types.ts` as `ProductDocStatus`, but unlike the extractor half nothing pins that mirror; same drift risk the new test exists to remove, one language over.

---

perrozzi:
Last round. **One accepted, one improved rather than applied as asked, one declined as incorrect.** Severity has been falling monotonically across four passes — should-fix, then nits, then one real regression, now this — so I am stopping here.

## Accepted

**The TypeScript mirror was genuinely unpinned, and it matters more than it looks.** `PRODUCT_DOC_STATUSES`' own comment claims it is mirrored in `types.ts`, and by this repo's habit a comment asserting something about another file is a claim to test. Concretely: `DocStatusBadge` switches on `ready` and `failed` and falls through to the "Extracting…" spinner for everything else, so a status the backend can write but the union does not name renders as a **permanently in-flight badge** — the exact never-resolving state this whole rung was written to remove. Pinned by reading the union out of the source text, with a vacuity guard so a regex that stops matching fails rather than comparing empty to empty. Two mutation probes (drop a union member; rename the type) both caught.

## Improved rather than applied

**The timing budgets — widening them alone would have made the tests worse, so I raised the input instead.** The cost being detected is fixed, so every second added to a budget moves the threshold *toward* the failure it must catch. `1E+300000` costs ~2.7s unbounded, so the 0.5s budget already sat only ~5x below the failure; widening it "substantially" would have walked it into the hazard.

Raising the exponent to `1E+1000000` (~30s unbounded) buys a 5s budget that is ~500,000x above the real ~10µs cost **and** ~6x below the failure — generous on both sides rather than one. The comment now records that trade, because "just add headroom" is the intuitive and wrong fix here. The deterministic `_NeverConvertibleDecimal` test stays the primary guard, as you noted; these two exist to survive a rewrite that no longer routes through the patched name.

## Declined — this finding is incorrect

You report that `MAX_IMAGE_BYTES <= MAX_FILE_BYTES` is "only incidentally pinned (by an inline `assert` inside an unrelated message test)". There is a dedicated property test with its own docstring explaining why the ordering matters:

```
test_image_limits_lockstep.py::TestUploadCapLockstep::test_the_image_cap_is_stricter_than_the_file_cap
```

It asserts exactly what you asked for, and its docstring makes your argument — that if the image cap ever exceeded the file cap the per-type check would be unreachable and the Converse limit would silently stop being enforced.

That file landed in the first commit, so an incremental review does not cover it — and you have correctly flagged your diff as truncated in **all four** passes. Worth saying plainly, because it cuts both ways: the same truncation means `core-stack.ts` (the Lambda definition, IAM grants, KMS, S3 notification wiring), `model-allowlist.ts` and `image_limits.py` have gone unreviewed throughout. Those rest on the synth proof and the CDK tests rather than on this review, and a human reader should know that before merging.

## Closing state

Four rounds, and the two structural items I declined at round 1 are unchanged and stated on the record: the extractor's core-stack placement (an explicit instruction, cycle verified by synth, and the app is at the five-template cap) and the double `_injectable_docs` call (two DynamoDB reads before a multi-second Bedrock call).

Gates, each run separately: backend **1855**, frontend **2486**, `test:cdk` 238, `test:stream` 303, `lint:python` unchanged at 553 with a byte-identical per-rule histogram, `i18n:check` unchanged, synth 5 templates and zero `[Warning at]`.

**46 mutation probes across the rung, all caught, none survived.** Plus the upload UI rendered and driven in a real browser, including both resize-ladder rungs against actual codecs.

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated (it cuts off inside `lambda/product_doc_extractor/handler.py`), so `lib/stacks/core-stack.ts`, `lib/utils/model-allowlist.ts`, `lambda/shared/image_limits.py` and the tail of the extractor remain unreviewed, as in the previous passes.

Since the last review the author made two changes: widened both wall-clock budgets in `test_product_doc_upload_boundary.py` to 5.0s (with a rationale that headroom is bounded above by the cost being detected, and a larger exponent `1E+1000000` so the unbounded path would take ~30s), and added `TestFrontendStatusMirrorLockstep` pinning `ProductDocStatus` in `frontend/src/api/types.ts` against `PRODUCT_DOC_STATUSES` by source-text regex, with a skip guard for backend-only trees and a vacuity guard. Two of my three previous new findings are addressed; the changes are small and clean.

### Resolved
- Timing budgets widened with both-sides margin (~500,000× above the guarded cost, ~6× below the unbounded cost) — `lambda/api/test/test_product_doc_upload_boundary.py`.
- The TypeScript mirror of the status vocabulary is now pinned, closing the third-copy drift risk — `lambda/product_doc_extractor/test/test_status_lockstep.py`.

### Still open
- `MAX_IMAGE_BYTES <= MAX_FILE_BYTES` ordering is still only incidentally pinned, while `_declared_size` hardcodes `MAX_FILE_BYTES` as "the widest cap any caller can accept".
- Extractor's deviation from the repo Lambda pattern (no powertools Logger/Tracer, lazy `_clients`) and the unexplored EventBridge/processing-stack placement.
- `generate_report` calls `_injectable_docs` then `build_product_context_block`, which re-queries.

### New findings
- `lambda/product_doc_extractor/test/test_status_lockstep.py` — the regex `([^\n]+)` captures only the first line of the union, so a Prettier/ESLint reformat that wraps `ProductDocStatus` across lines makes the backend suite fail with a message blaming a drift that did not happen. Anchoring to the statement terminator (e.g. match up to the next `;` with `re.DOTALL`) would make the test insensitive to formatting.
- `lambda/product_doc_extractor/test/test_status_lockstep.py` — `test_the_union_was_actually_parsed` hardcodes `>= 4`, which is the current cardinality; asserting `== len(PRODUCT_DOC_STATUSES)` keeps the vacuity guard meaningful if the vocabulary ever shrinks, and matches the intent stated in its own docstring.
- Unverifiable from the diffs (types.ts is not in either diff): the new test asserts a declaration form `export type ProductDocStatus = 'a' | 'b' | …`. If the current file derives it (e.g. `ProductDoc['status']`) the `assert match` fires immediately — worth confirming the test actually passes rather than skips or fails in this tree.

---

perrozzi:
**Reversing two of my own dispositions.** I said round four was the last, then re-read your "Still open" list cold — and both remaining items should have been fixed. Neither decline was wrong about facts; both were wrong about what they were declining.

## The double read is a TOCTOU, and I declined it on your framing rather than on the defect

You described it three rounds running as *"two extra `_list_doc_items` per report"*. I measured the cost, correctly found it unmeasurable next to a multi-second Bedrock call, and declined. **The cost was never the defect — the plurality was.**

`generate_report` read the list at the gate and `build_product_context_block` read it again for the content. The gate's own comment claimed it "cannot admit a project whose only content that function declines to include". That is true of the **predicate** and false of the **data**: a document deleted between the two reads passed the gate and was then absent from the block, so the report was synthesized from `(No product context provided.)`.

And it was recorded as a lie. The saved item wrote `product_context_included=True` as a literal, justified by a comment asserting the block "is always present here". So the race produced a document whose provenance claimed context it did not contain — the same dishonesty this rung was written to remove, arriving by a third route.

Sharpest detail, which is also why your instinct to keep raising it was right: the gate is `not has_any and not _injectable_docs(...)`, and `and` short-circuits. **The second read only ever happened on the field-less path** — precisely the project whose documents are its *only* content. The race bit exactly where it existed.

Fixed by reading once and passing the list through, plus deriving the flag instead of asserting it. Three things worth flagging:

- **`None` means "not read", not "none exist".** Passing `[]` on the short-circuited path would silently drop every document from a report whose fields are filled — no error, no missing heading. There is a mutation and a test for that specifically.
- **A handed-in list is still filtered** through the injectable predicate, or the new parameter becomes a fail-open back door around the image exclusion rung 3 owns.
- **The short-circuit stays.** Reading unconditionally hangs an existing test whose `MagicMock` table returns a truthy `LastEvaluatedKey` and spins the pagination loop forever. Keeping it also keeps the query count at one on *both* paths — a documents-only project previously made two.

## I bundled your logging finding with the stack move and reused one justification for both

Your bullet paired "no powertools Logger/Tracer" with "EventBridge + processing-stack placement". I declined the pair four times on the placement reasoning. **That reasoning has nothing to do with the log shape**, and I never noticed because the two arrived in one bullet.

Then I argued inside a dichotomy I built myself: powertools-or-plain-text. Powertools genuinely does need the layer that would force container bundling into `VocCoreStack`. **Structured logging does not need powertools** — a stdlib `Formatter` subclass is the whole mechanism.

Implemented as a module-level context dict merged at format time, set once per S3 record and cleared in a `finally`, so **all 18 existing log lines gained `project_id`/`doc_id`/`s3_key` with zero call-site edits** — the same thing `Logger.append_keys` does. `doc_id` goes from an f-string substring to a field. That matters most on this Lambda specifically: its failure reaches the user as one short sentence, so every diagnosis is a CloudWatch query.

Two traps found while doing it, both now asserted rather than assumed:

- **The Lambda runtime pre-installs a plain-text root handler**, so adding ours beside it emits every line twice — and a doubled line reads like a retry, not a formatting mistake. It is detached inside Lambda only.
- My first attempt instead *reformatted* the existing handler. Outside Lambda that host is pytest, whose capture handler populates `record.message` as a side effect of its own formatter — six unrelated tests broke, and **only when the two suites shared a process.** Handlers we did not create are now left alone.

## Still declined

The stack move. Unchanged and unaffected by any of the above: core-stack was an explicit instruction, the cycle is documented at the construct and synth-verified, and the app has no template headroom.

## Verification

backend **1893** (was 1855), `test:cdk` **239** (was 238), frontend 2486 and `test:stream` 303 unmoved, `lint:python` still exactly 553 with a byte-identical per-rule histogram, `i18n:check` unchanged (no frontend file touched), synth 5 templates and zero `[Warning at]`.

**8 mutation probes, 8 caught, 0 survived**, restores byte-identical: reintroduce the second read; re-hardcode the flag; pass `[]` instead of `None`; move the context clear out of the `finally`; remove it; make the formatter emit plain text; make it ignore the context dict; leave the runtime handler attached.

Running total: **54 mutation probes across the rung, all caught.**

## The lesson I'm taking from this round

**Decline the defect, not the framing.** A severity label is itself a claim, and refuting the label is not refuting the finding — I answered the question you asked instead of the one the code raised. And split a multi-part finding before disposing of it: a disposition that answers the strongest half looks complete while the weaker half goes unexamined, which is how the logging item survived four rounds unaddressed.

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated (it cuts off inside `lambda/product_doc_extractor/handler.py`), so `lib/utils/model-allowlist.ts`, `lambda/shared/image_limits.py` and the tail of the extractor remain unreviewed, as in previous passes.

Since the last review the author added two unrelated pieces: (a) `generate_report` now reads the injectable-doc list once and hands it to `build_product_context_block(docs=…)`, with `_is_injectable` factored out and `product_context_included` derived rather than hardcoded, plus `test_product_report_single_read.py`; (b) the extractor grew stdlib structured JSON logging (`JsonFormatter`, `_log_context`, `_configure_logging`, `_process_record`/`_extract_document` split) with `SERVICE_NAME`/`LOG_LEVEL` injected from `core-stack.ts`, a CDK assertion and a baseline hash bump. None of the findings from my previous pass were touched.

### Resolved
- The double read of the doc list in `generate_report` (a previous "Still open" item) is now a single read handed to the block, with `_is_injectable` reapplied so the parameter cannot fail open — `lambda/api/product_context.py`.

### Still open
- `MAX_IMAGE_BYTES <= MAX_FILE_BYTES` ordering only incidentally pinned, while `_declared_size` hardcodes `MAX_FILE_BYTES`.
- Extractor's deviation from the repo Lambda pattern (lazy `_clients`) and the unexplored EventBridge/processing-stack placement.
- `test_status_lockstep.py`: single-line-only `([^\n]+)` union regex, and the `>= 4` vacuity guard instead of `== len(PRODUCT_DOC_STATUSES)`.

### New findings
- **Should fix** — `lambda/product_doc_extractor/handler.py`: `_configure_logging(logger, replace_existing=IN_LAMBDA)` runs at import against the **root** logger even outside Lambda, where it (i) attaches a JSON `StreamHandler` to root for the whole host process and (ii) sets the root level to `_log_level()`. In the backend pytest run that is a global side effect from importing one handler module, and because the non-Lambda branch never removes a handler it previously installed, each reload of the module (the `extractor` fixture reloads per test) appends another root handler — so lines are emitted N times and handlers accumulate across the session. The `json_logs` fixture reads only its own stream, so no test would notice. Either configure only when `IN_LAMBDA`, or use a named logger (`logging.getLogger(__name__)`) with `propagate = False` and make the attach idempotent.
- Nit — `lib/stacks/core-stack.ts`: `LOG_LEVEL` is a hardcoded `'INFO'`. Given the stated motivation ("a diagnosis happens entirely in CloudWatch"), consider sourcing it the same way the other functions in this app do, so verbosity can be raised without editing the stack.

---

perrozzi:
Good catch, and it was mine from last round. Fixed, plus both nits.

## The root-logger side effect

You were right about the substance. Importing the extractor called `_configure_logging` against the **root** logger, so merely importing a handler module attached a JSON `StreamHandler` to the host process's logger and re-levelled it — under pytest, reformatting every other suite's logging as a side effect of one import. And the non-Lambda branch was genuinely non-idempotent.

Your diagnosis of *why nothing caught it* is the part worth repeating: the logging tests read only their own stream, so the defect was invisible to the very tests covering the feature.

I took your second option, because it removes the class rather than the symptom: a **named logger with `propagate = False`** and an **idempotent** attach. Consequences worth stating:

- **`IN_LAMBDA` and the detach path are gone, not retained as a dead branch.** Not propagating is what prevents the doubled line, so detaching the runtime's handler is no longer needed — and it was actively wrong anyway, because it took the plain-text destination away from every *other* library logging through root.
- **Idempotence keys on `Handler.set_name`, not `isinstance`.** A reload rebinds `JsonFormatter` to a new class object, so an isinstance check answers False for the handler the previous generation installed — precisely when the duplicate appears.
- **Lines still reach CloudWatch**: the handler writes to `sys.stdout`, which Lambda forwards to the log stream and where powertools writes too.

One thing your finding did not anticipate, and it is the same defect class you named: **pytest's `caplog` handler sits on root**, so a non-propagating logger would have made every `caplog` assertion in this package silently capture **zero** records — five tests that pass while checking nothing. An autouse fixture routes the module's records into `caplog`, autouse so the next `caplog` test written here works without anyone remembering.

**Proved empirically, not just by assertion.** Full backend suite, both ways:

| | JSON handlers left on root | root handlers | suite |
|---|---|---|---|
| fixed | **0** | 2, both pytest's | 1901 passed |
| root config restored | **1** (ours, named) | 3 | 1897 passed, **4 failed** |

So the new guards fire in the real run, not only in isolation.

## One correction

**"the `extractor` fixture reloads per test" is not the case** — `conftest.py::extractor` does a plain cached import and clears `_clients`; there is no `importlib.reload` anywhere in these tests. So handlers were not in fact accumulating across the session. Everything else in the finding stands on its own, so it is fixed as reported, with the reload made the *test's* mechanism rather than an accident of the fixture.

## The two nits

**Union regex** — fixed, though not quite as suggested: "stop at the statement end" cannot work here, because this repo's prettier writes **no semicolons**, so the declaration has no terminator to stop at and a `[^;]+;` capture would not match the real file at all. The pattern now matches the union *members*, which parses either layout and cannot run into the following statement. The `>= 4` guard is replaced with a comparison against `len(PRODUCT_DOC_STATUSES)`, non-circular because the count is parsed from the TypeScript while the expectation comes from the Python tuple.

**`LOG_LEVEL`** — I grepped rather than assumed, and the convention is hardcoded `'INFO'` at **all 24** Lambda definitions across the four stacks. There is no context key or stack parameter to follow, so inventing one for a single function would have been the deviation. Your underlying concern is already satisfied: the handler reads `LOG_LEVEL` from the environment at import, so verbosity is raisable on the deployed function without a stack edit — now stated where the value is set.

## Re-listed items, unchanged

`MAX_IMAGE_BYTES <= MAX_FILE_BYTES` is pinned by a dedicated test with its own docstring (`test_image_limits_lockstep.py::TestUploadCapLockstep::test_the_image_cap_is_stricter_than_the_file_cap`); it shipped in the first commit, which an incremental review does not cover. And the lazy `_clients` / stack-placement item remains declined on the record.

## Verification

backend **1901** (was 1893: +11 new, −3 deleted with the mechanism they pinned), frontend 2486, `test:cdk` 239, `test:stream` 303 all unmoved, `lint:python` still exactly 553 with a byte-identical per-rule histogram, `i18n:check` unchanged, synth 5 templates and zero `[Warning at]`.

**4 mutation probes, 4 caught, 0 survived**: root logger restored, attach made non-idempotent, `propagate = True`, single-line-only regex. Running total **58 across the rung, all caught.**

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated again (it cuts off inside `lambda/product_doc_extractor/handler.py`), so `lib/utils/model-allowlist.ts`, `lambda/shared/image_limits.py` and the tail of the extractor remain unreviewed, as in the previous passes.

Since the last review the author reworked the extractor's logging onto a named, non-propagating logger with a name-keyed idempotent handler writing to `sys.stdout` (replacing the root-logger takeover), added an autouse conftest fixture to route those records into `caplog`, added `TestImportingTheModuleLeavesTheHostAlone` / `TestTheHandlerIsAttachedExactlyOnce` with a snapshot-restoring `reload_extractor` fixture, hardened the `ProductDocStatus` union regex plus its vacuity guard, and documented the hardcoded `LOG_LEVEL` in `core-stack.ts`. Both of my last-pass findings are addressed.

### Resolved
- `lambda/product_doc_extractor/handler.py`: import no longer attaches a handler to or re-levels the host root logger, and the attach is idempotent across reloads — with tests that fail on the work done, not just on the output.
- `lambda/product_doc_extractor/test/test_status_lockstep.py`: `([^\n]+)` replaced by a member-anchored pattern that parses prettier's wrapped layout and cannot run into the next statement; `>= 4` replaced by `== len(PRODUCT_DOC_STATUSES)`, with a parse-failure assertion that names the regex.
- `lib/stacks/core-stack.ts`: `LOG_LEVEL: 'INFO'` now carries the justification (matches all 24 definitions; raised via env var on the deployed function) — accepted as-is.

### Still open
- `MAX_IMAGE_BYTES <= MAX_FILE_BYTES` ordering only incidentally pinned, while `_declared_size` hardcodes `MAX_FILE_BYTES`.
- Extractor's deviation from the repo Lambda pattern (lazy `_clients`) and the unexplored EventBridge/processing-stack placement.

### New findings
- Nit — `lambda/product_doc_extractor/test/conftest.py`: the autouse fixture routes records into `caplog` but the logger's own level stays at `_log_level()` (INFO) and `propagate = False`, so `caplog.set_level(...)` / `caplog.at_level(...)` — which pytest applies to the *root* logger — will not lower this logger's threshold. A future DEBUG-level `caplog` assertion here would capture nothing and pass, which is exactly the "green test that pins nothing" failure the fixture docstring is guarding against. Setting the level on `handler.logger` inside the fixture (or documenting `caplog.set_level(..., logger=handler.logger.name)`) closes it.
- Nit — same file: the fixture attaches to `product_doc_extractor.handler.logger`; this only works while every test module reaches the handler through that same import path (a bare `import handler` would produce a second module object with a differently-named logger). Worth asserting `handler.logger is <module under test>.logger` or noting the constraint, since the failure mode is again a silently empty `caplog`.

---

perrozzi:
Both nits fixed. Stopping here — nit-only round, both addressed, and the loop has converged.

## Both were the exact class the fixture was added to prevent

**`caplog.at_level` cannot reach this logger.** Right, and worth spelling out why it's not cosmetic: `at_level` acts on the **root** logger, my logger has its own level and doesn't propagate, so a future `caplog.at_level(logging.DEBUG)` assertion in this package would have captured zero records and passed. The fixture now lowers this logger for the duration and restores it, leaving pytest's handler to do the filtering a caller expects.

**The import-path constraint.** Also right, and now an assertion rather than a comment — a bare `import handler` yields a second module object with a differently-named logger, and the failure looks identical to "the code logged nothing".

## The probe found something better than either nit

The first version of my restoration test asserted `_log_level() == INFO`. **That reads the environment, not the logger**, so it held no matter what the fixture did — a vacuous test, inside a class written specifically to stop vacuous tests. Nothing caught it except mutating the fixture and noticing the suite stayed green.

Replaced with an assertion that the fixture lowered *this* logger **and** that the module itself is not at DEBUG. The second half matters: without it the test would also pass if the module simply shipped low, which would mean the fixture was doing nothing and the DEBUG test above was green for the wrong reason.

## Two mutations I am recording as not unit-killable rather than claiming cover

I would rather say this than report 4/4:

- **Removing the level restore survives.** Its only observable consequence is the logger's state after the *last* test in the session, which nothing inside that session can see.
- **Removing the shared-logger assertion survives.** It documents an invariant that currently holds, so killing it would require falsifying the invariant from outside the test.

Both are stated in the test docstrings. Claiming cover for them would be the same defect as the vacuous test above, one layer out.

## Verification

backend **1905** (was 1901). `lint:python` unchanged at 553 with both touched files clean per-file; `test:cdk` 239 and frontend 2486 unmoved, since only test-side Python changed. Mutation probe: H1 (level not lowered) caught by two tests, H4 (records not routed at all) caught by eight, H2/H3 as above.

## Closing

Six rounds. Severity fell monotonically and ended on nits about test-support code: 5 should-fix → 5 nits → 3 nits with one regression → 3 nits with one false → 1 should-fix (mine) → 2 nits. **Ten of your findings implemented, two declined on the record, two identified as incorrect.**

The two long-standing declines stand unchanged and are on the record above: the extractor's core-stack placement with its lazy `_clients` (an explicit instruction, a synth-verified cycle, and no template headroom), and the cap-ordering item, which is pinned by a dedicated test your truncated diff cannot see.

Worth repeating once for whoever merges: **your diff was truncated in all six passes.** `lib/stacks/core-stack.ts` — the Lambda definition, IAM grants, KMS and the S3 notification wiring — plus `model-allowlist.ts` and `image_limits.py` were never reviewed here. Those rest on the synth proof, the CDK tests and the mutation probes, not on this review.

perrozzi:

Follow-up review — **partial**: the full-context diff was truncated again (it cuts off inside `lambda/product_doc_extractor/handler.py`), so `lib/utils/model-allowlist.ts`, `lambda/shared/image_limits.py` and the tail of the extractor remain unreviewed, as in previous passes.

Since the last review the only change is to the extractor's test scaffolding: the autouse `caplog` fixture now asserts the fixture and the module-under-test share one logger, lowers `handler.logger` to DEBUG for the test's duration and restores the previous level in `finally`, and a new `TestCaplogActuallyCapturesFromThisLogger` class pins all of that (including a vacuity guard that the module itself is not at DEBUG). Both of my previous nits are addressed, and the delta looks clean.

### Resolved
- `test/conftest.py`: the named non-propagating logger is now lowered by the fixture, so `caplog.at_level(DEBUG)` assertions in this package can no longer pass on zero records — pinned by `test_a_debug_record_reaches_caplog_when_the_caller_asks_for_debug`.
- `test/conftest.py`: the shared-module-object constraint is asserted rather than assumed (`handler.logger.name == handler.__name__`), with `test_the_fixture_and_the_module_share_one_logger` documenting why.

### Still open
- `MAX_IMAGE_BYTES <= MAX_FILE_BYTES` ordering only incidentally pinned, while `_declared_size` hardcodes `MAX_FILE_BYTES`.
- Extractor's deviation from the repo Lambda pattern (lazy `_clients`) and the unexplored EventBridge/processing-stack placement.

### New findings
- Nit — `lambda/product_doc_extractor/test/conftest.py`: the DEBUG lowering happens before the `extractor` fixture, but `_configure_logging()` calls `target.setLevel(_log_level())` on every invocation, so any test that goes through the reloading fixture (`reload_extractor` in `test_structured_logging.py`) re-levels the logger back to INFO *after* the autouse fixture lowered it. A future DEBUG `caplog` assertion inside a reload-based test would therefore capture nothing and pass — the same vacuous-green shape this class exists to prevent, arriving one fixture over. Either lower the level inside `reload_extractor` too, or note the constraint there.
- Nit — `test/test_structured_logging.py`: `test_the_fixture_lowered_this_logger_rather_than_the_module_defaulting_low` asserts `extractor._log_level() == logging.INFO`, which reads the ambient environment; a developer or CI job with `LOG_LEVEL=DEBUG` exported fails it for a reason unrelated to the fixture. `monkeypatch.delenv('LOG_LEVEL', raising=False)` makes the vacuity guard deterministic.

---

PR #326, opened 2026-08-13
test(i18n): gate the init options, above all the namespace list

perrozzi:
## Summary

Follow-up to #324, which merged while this was being written.

#324 moved the i18next init options out of `src/i18n/config.ts` into a side-effect-free `src/i18n/options.ts`, so a test could assert the app's real `nsSeparator` instead of the test harness's. It stopped one step short: **the options became importable but nothing asserted anything about them.** No test imports `config.ts` — it runs `i18n.use(HttpBackend).use(LanguageDetector).init(...)` at module scope — so every option in that literal was untested before the move and stayed untested after. Drop `supportedLngs` and the app still boots, just without the guard that rejects a stale cached locale.

> **Scope note (corrected):** this started as test-only and is no longer. Review round 1 asked for a parity assertion across the four copies of the namespace list, and **that assertion failed on arrival** — so this PR also changes one line of production tooling, `scripts/fix-i18n.mjs`. Details below.

## The gap that matters: the namespace list

A page calling `useTranslation('somewhere')` for a namespace absent from `init({ ns })` **does not throw**. i18next resolves nothing and the page renders raw key paths — the same failure class #324 fixed twice.

`scripts/i18n-check.mjs` structurally cannot catch this: its own `NAMESPACES` array is a separate hardcoded copy, so it validates keys against the namespaces *it* knows, never against the ones the app registers.

`src/i18n/options.test.ts` adds five gates:

| gate | what it prevents |
|---|---|
| every namespace used in source is registered | a page rendering raw key paths |
| the scan sees every form the source uses | the gate above silently under-matching |
| four copies of the list agree with the shipped catalogues | drift between app, test harness and two scripts |
| every locale ships the same catalogue set | an orphan catalogue belonging to no namespace |
| the load-bearing options keep their values | a silent loss of the locale/detection guards |

The scan understands `useTranslation('ns')`, `useTranslation(['a','b'])`, `t('ns:key')`, a `t` renamed on destructuring, and namespace-qualified keys held in data. The one form no static scan can resolve — a namespace passed as a variable or template literal — is asserted absent rather than left as a blind spot.

## The bug the parity gate found

**`scripts/fix-i18n.mjs` listed a 16th namespace, `feedback`, with no catalogue in any locale.** Every run of that script tried to load `locales/<lang>/feedback.json`, which does not exist. Removing that one entry is the production change in this PR; the four copies now agree with the 15 shipped catalogues.

Related, and why the per-locale gate exists: parity against `en` alone would let an orphan like `fr/feedback.json` satisfy every list while belonging to no namespace — and since `fix-i18n.mjs` iterates the registered list, it would skip that file forever. No orphans today (all 8 locales ship exactly 15 catalogues), and now that cannot start silently.

## Two notes from writing it

- **A comment created a false positive.** The scan's first run reported a namespace called `ns` and declared the app broken. The source was a docblock in `ValidationLinkPicker.tsx` discussing the shape `xKey: 'ns:key'` — prose, not code. `i18n-check.mjs` shares the blind spot and warns about that same comment. Comments are stripped before scanning, block comments only where `/*` opens the line so a `/*` inside a string cannot pair with a later `*/` and delete real code.
- **#324's extraction was lossless**, confirmed by diffing the option keys `config.ts` passed inline at the branch point against the new module, positive-controlled. `defaultNS` was a shorthand property before (`defaultNS,`) and so is invisible to a `key:` regex — it reads as newly added when it was there all along.

## Testing

- **2464/2464** tests, `tsc --noEmit` and eslint clean over `src/` and `scripts/`, on top of the post-merge `development` tip.
- `i18n-check.mjs` output **byte-identical** to that tip, measured by stashing this PR's files and re-running. The pre-existing ja/zh/ko `extra: 4` in `categories.json` is not from here.
- **Eleven mutations, each caught by the intended test and the file restored**, with a green baseline asserted first. Including: an unregistered namespace via the single-string form and via the **array** form; a qualified key through the **renamed** `t`; a non-literal namespace; the stale `feedback` entry put back; an **orphan** `fr/feedback.json`; a locale **missing** a catalogue.
- One **negative control**: planting a type annotation shaped like a rename (`{ t: GhostType }`) must *not* be treated as a caller, and the gate correctly stays green — the alias regex is scoped to destructuring precisely so type annotations and parameters don't join the caller set.

### DIFF
--- a/voc-datalake/frontend/scripts/fix-i18n.mjs
+++ b/voc-datalake/frontend/scripts/fix-i18n.mjs
@@ -12,8 +12,12 @@ import { execFileSync } from 'node:child_process'
 
 const __dirname = fileURLToPath(new URL('.', import.meta.url))
 const LOCALES_DIR = resolve(__dirname, '..', 'public', 'locales')
+// Must match the shipped catalogues in public/locales/<lang>/ — `src/i18n/options.test.ts`
+// compares this list, i18n-check.mjs's copy, src/i18n/options.ts and src/test/setup.ts
+// against the files on disk. `feedback` was listed here with no catalogue in any
+// locale, so every run tried to load a namespace that does not exist.
 const NAMESPACES = ['categories', 'chat', 'common', 'components', 'dashboard',
-  'dataExplorer', 'feedback', 'feedbackDetail', 'feedbackForms', 'login',
+  'dataExplorer', 'feedbackDetail', 'feedbackForms', 'login',
   'prioritization', 'problemAnalysis', 'projectDetail', 'projects', 'scrapers', 'settings']
 const LANGUAGES = ['es', 'fr', 'de', 'ko', 'pt', 'ja', 'zh']
 const LANG_NAMES = {
--- a/voc-datalake/frontend/src/i18n/options.test.ts
+++ b/voc-datalake/frontend/src/i18n/options.test.ts
@@ -1,101 +1,220 @@
 /**
- * @fileoverview Gates on the i18n init options.
+ * @fileoverview Gates on the i18n init options, and on the namespace list in
+ * particular.
  *
  * These exist because the options moved out of `config.ts` into `options.ts` and
  * NOTHING imports `config.ts` in a test — it runs `init()` with an HTTP backend at
- * module scope. So every option in it was, and would remain, untested: drop
+ * module scope. So every option in it was, and would have remained, untested: drop
  * `supportedLngs` and the app still boots, just without the guard that rejects a
  * stale cached locale.
  *
  * The load-bearing one is the namespace list. A page that calls
  * `useTranslation('somewhere')` for a namespace absent from `ns` does not throw —
  * i18next resolves nothing and the page renders raw key paths. That is the same
- * failure this module's own history is made of (a relative key read through the
- * wrong namespace, and a data-held key invisible to the i18n gate), and
- * `scripts/i18n-check.mjs` cannot catch it: its own `NAMESPACES` array is a
- * separate hardcoded copy, so it validates keys against the namespaces IT knows,
- * never against the ones the app registers.
+ * failure this module's history is made of, and `scripts/i18n-check.mjs` cannot
+ * catch it: its own `NAMESPACES` array is a separate hardcoded copy, so it
+ * validates keys against the namespaces IT knows, never against the ones the app
+ * registers.
+ *
+ * The reference for "which namespaces exist" is the shipped catalogue files, not
+ * any of the four hardcoded lists — a namespace IS a catalogue.
  */
 import { describe, it, expect } from 'vitest'
-import { readdirSync, readFileSync, statSync } from 'node:fs'
-import { join, extname } from 'node:path'
+import { readdirSync, readFileSync } from 'node:fs'
+import { join, extname, basename } from 'node:path'
+import i18n from 'i18next'
 import { I18N_INIT_OPTIONS } from './options'
 
 const SRC = join(__dirname, '..')
+const FRONTEND = join(SRC, '..')
+const LOCALES_EN = join(FRONTEND, 'public', 'locales', 'en')
+const SCRIPTS = join(FRONTEND, 'scripts')
+
+/** The namespaces actually shipped: one catalogue file each. */
+function shippedNamespaces(): string[] {
+  return readdirSync(LOCALES_EN)
+    .filter((f) => extname(f) === '.json')
+    .map((f) => basename(f, '.json'))
+    .sort()
+}
+
+/**
+ * Parse a `const NAME = ['a', 'b']` array literal out of a script.
+ *
+ * Throws rather than returning empty on a miss: an unparsed list would make every
+ * comparison below trivially pass.
+ */
+function arrayLiteral(file: string, variable: string): string[] {
+  const text = readFileSync(file, 'utf-8')
+  const match = new RegExp(`${variable}\\s*=\\s*\\[([^\\]]*)\\]`, 's').exec(text)
+  if (!match) throw new Error(`${basename(file)}: could not find the ${variable} array`)
+  const items = [...match[1].matchAll(/['"](\w+)['"]/g)].map((m) => m[1])
+  if (items.length === 0) throw new Error(`${basename(file)}: ${variable} parsed as empty`)
+  return items.sort()
+}
 
 /**
  * Drop comments before scanning.
  *
- * Not fussiness: this file's own docblocks discuss the shape of a translation key
- * (`xKey: 'ns:key'`), and a naive scan reads that prose as a namespace called `ns`
- * and reports the app as broken. `scripts/i18n-check.mjs` has the same blind spot
- * and warns about that exact comment.
+ * Not fussiness: a docblock in `ValidationLinkPicker.tsx` discusses the shape of a
+ * translation key (`xKey: 'ns:key'`), and a naive scan reads that prose as a
+ * namespace called `ns` and reports the app as broken. It did, on the first run.
+ * `scripts/i18n-check.mjs` has the same blind spot and warns about that comment.
  *
- * Block comments and whole-line comments only — a `//` inside a string literal (a
- * URL) is left alone, so no line of real code is truncated.
+ * Block comments are stripped only where `/*` opens the line, so a `/*` inside a
+ * string or regex literal cannot pair with a later `*␘/` and delete real code
+ * between them — that would be a false negative, the worse direction for a gate.
  */
 function stripComments(source: string): string {
-  return source
-    .replace(/\/\*[\s\S]*?\*\//g, '')
-    .split('\n')
-    .filter((line) => {
-      const t = line.trim()
-      return !t.startsWith('//') && !t.startsWith('*')
-    })
-    .join('\n')
+  const lines = source.split('\n')
+  const kept: string[] = []
+  let inBlock = false
+  for (const line of lines) {
+    const trimmed = line.trim()
+    if (inBlock) {
+      if (trimmed.includes('*/')) inBlock = false
+      continue
+    }
+    if (trimmed.startsWith('/*')) {
+      if (!trimmed.includes('*/')) inBlock = true
+      continue
+    }
+    if (trimmed.startsWith('//') || trimmed.startsWith('*')) continue
+    kept.push(line)
+  }
+  return kept.join('\n')
 }
 
-/** Every namespace the source asks `useTranslation` for, or addresses as `ns:key`. */
-function namespacesReferencedInSource(): Set<string> {
-  const found = new Set<string>()
-  const walk = (dir: string) => {
-    for (const entry of readdirSync(dir)) {
-      if (entry === 'node_modules' || entry === 'test') continue
-      const full = join(dir, entry)
-      if (statSync(full).isDirectory()) {
-        walk(full)
-        continue
-      }
-      if (!['.ts', '.tsx'].includes(extname(entry))) continue
-      // Test files are excluded deliberately: they address namespaces directly
-      // (`t('feedbackForms:...')`) for assertions, which is not the app asking
-      // i18next to load one.
-      if (entry.endsWith('.test.ts') || entry.endsWith('.test.tsx')) continue
-      const source = stripComments(readFileSync(full, 'utf-8'))
-      for (const [, ns] of source.matchAll(/useTranslation\(\s*['"](\w+)['"]/g)) found.add(ns)
-      for (const [, ns] of source.matchAll(/\bt\(\s*['"](\w+):/g)) found.add(ns)
-      // Namespace-qualified keys held in data and passed to t() indirectly — the
-      // shape SCORABLE_TYPE_META uses. Same extraction rule as i18n-check.mjs.
-      for (const [, ns] of source.matchAll(/\b\w*[Kk]ey:\s*['"](\w+):[\w.]+['"]/g)) found.add(ns)
+function sourceFiles(dir: string, acc: string[] = []): string[] {
+  for (const entry of readdirSync(dir, { withFileTypes: true })) {
+    if (entry.name === 'node_modules' || entry.name === 'test' || entry.name === '__tests__') continue
+    const full = join(dir, entry.name)
+    if (entry.isDirectory()) {
+      sourceFiles(full, acc)
+      continue
+    }
+    if (!['.ts', '.tsx'].includes(extname(entry.name))) continue
+    // Tests address namespaces directly for their assertions, which is not the app
+    // asking i18next to load one.
+    if (/\.(test|spec)\.tsx?$/.test(entry.name)) continue
+    acc.push(full)
+  }
+  return acc
+}
+
+/**
+ * Every namespace the app asks i18next for, and how many files were read to find
+ * them (returned so a broken walk cannot look like a clean result).
+ *
+ * Forms understood:
+ *   useTranslation('ns')            — single string
+ *   useTranslation(['a', 'b'])      — array, which IS used (Categories/FeedbackResults)
+ *   t('ns:key')                     — qualified key
+ *   somethingKey: 'ns:key'          — namespace-qualified key held in data
+ *
+ *   tr('ns:key')                    — where `t` was renamed (`{ t: tr }`), which
+ *                                     Chat.tsx does; the alias is resolved per file
+ *                                     rather than banned, since a bare key through a
+ *                                     renamed `t` is perfectly correct
+ *
+ * The one form that cannot be resolved statically — a namespace passed as a variable
+ * or template literal — is asserted absent by a test below rather than left as a
+ * silent blind spot.
+ */
+function scanSource(): { namespaces: Set<string>; filesScanned: number } {
+  const namespaces = new Set<string>()
+  const files = sourceFiles(SRC)
+  for (const file of files) {
+    const source = stripComments(readFileSync(file, 'utf-8'))
+    for (const [, ns] of source.matchAll(/useTranslation\(\s*['"](\w+)['"]/g)) namespaces.add(ns)
+    for (const [, list] of source.matchAll(/useTranslation\(\s*\[([^\]]*)\]/g)) {
+      for (const [, ns] of list.matchAll(/['"](\w+)['"]/g)) namespaces.add(ns)
+    }
+    for (const [, ns] of source.matchAll(/\b\w*[Kk]ey:\s*['"](\w+):[\w.]+['"]/g)) namespaces.add(ns)
+
+    // `t` plus every name `t` was renamed to in this file, so a qualified key
+    // reached through an alias is not invisible.
+    const callers = new Set(['t'])
+    for (const [, alias] of source.matchAll(/\bt\s*:\s*(\w+)/g)) callers.add(alias)
+    for (const caller of callers) {
+      const qualified = new RegExp(`\\b${caller}\\(\\s*['"](\\w+):`, 'g')
+      for (const [, ns] of source.matchAll(qualified)) namespaces.add(ns)
     }
   }
-  walk(SRC)
-  return found
+  return { namespaces, filesScanned: files.length }
 }
 
 describe('I18N_INIT_OPTIONS', () => {
   it('registers every namespace the app actually asks for', () => {
     const registered = new Set(I18N_INIT_OPTIONS.ns as string[])
-    const referenced = namespacesReferencedInSource()
+    const { namespaces, filesScanned } = scanSource()
 
-    // Guard against the scan silently finding nothing — a broken walk would make
-    // the assertion below pass over an empty set.
-    expect(referenced.size, 'the source scan found no namespaces at all')
-      .toBeGreaterThan(5)
-    expect(referenced.has('feedbackForms'), 'scan missed a known namespace').toBe(true)
-    expect(referenced.has('prioritization'), 'scan missed the data-held namespace').toBe(true)
+    // Liveness, structural rather than by name: a walk that silently returns
+    // nothing would make the assertion below pass over an empty set. Names are
+    // deliberately not pinned here — a legitimately renamed namespace should not
+    // fail as though the app were broken.
+    expect(filesScanned, 'the source walk found almost no files').toBeGreaterThan(100)
+    const shipped = shippedNamespaces()
+    expect(
+      [...namespaces].filter((ns) => shipped.includes(ns)).length,
+      'the scan matched no shipped namespace at all — it is not reading source',
+    ).toBeGreaterThan(3)
 
-    const unregistered = [...referenced].filter((ns) => !registered.has(ns))
+    const unregistered = [...namespaces].filter((ns) => !registered.has(ns))
     expect(
       unregistered,
-      'these namespaces are used in source but not in I18N_INIT_OPTIONS.ns — '
-      + 'i18next will resolve nothing for them and the UI renders raw key paths',
+      'used in source but absent from I18N_INIT_OPTIONS.ns — i18next resolves '
+      + 'nothing for these and the UI renders raw key paths',
     ).toEqual([])
   })
 
+  it('sees every form the source uses to name a namespace', () => {
+    // The gate's own fidelity, asserted rather than assumed: a scan that silently
+    // under-matches reads green on exactly the defect it exists to catch.
+    //
+    // Only one form is unresolvable by any static scan — a namespace that is not a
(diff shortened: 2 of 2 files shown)


perrozzi:

This is a test-only PR adding `voc-datalake/frontend/src/i18n/options.test.ts`, which (a) scans `src/**` for `useTranslation('ns')`, `t('ns:key')` and data-held `*Key: 'ns:key'` references and requires each namespace to be present in `I18N_INIT_OPTIONS.ns`, and (b) asserts one-by-one on the load-bearing init options. The intent is sound and it closes a real gap left by #324 (nothing imported the options, so nothing verified them), the assertions carry useful failure messages, and the mutation testing described is the right way to validate a scan-based test. No blockers; the remarks below are about the gate's fidelity — a scan-based guard that silently under-matches is worse than none, because it reads green.

### Should fix

- **The scan only recognises the single-string forms, so it can under-match silently** (`src/i18n/options.test.ts`, `namespacesReferencedInSource`). Not matched: `useTranslation(['a', 'b'])`, `useTranslation(\`${x}\`)`, `t('ns:key')` reached via a destructured alias (`const { t: tr } = useTranslation(...)`; `tr('ns:key')` fails `\bt\(`), and `useTranslation('ns', { keyPrefix })` variants where the namespace is a constant reference. Any of these makes the gate pass on exactly the failure it exists to catch. At minimum add a comment enumerating the unsupported forms, and ideally assert the array form is unused (e.g. fail if `useTranslation(\s*\[` appears anywhere) so the blind spot can't be introduced later without a test failure.
- **The sentinel assertions pin unrelated names and a magic number.** `expect(referenced.size).toBeGreaterThan(5)`, `referenced.has('feedbackForms')` and `referenced.has('prioritization')` will fail loudly if either namespace is legitimately renamed or a page removed — a failure that reads as "the app is broken" when the scan is fine. Prefer a structural liveness check, e.g. assert the walk visited a non-trivial number of `.tsx` files, and/or derive the sentinel from the locale JSON files rather than hardcoding two names.
- **The stated root cause — four copies of the namespace list — is still unguarded.** The PR argues (correctly) that `scripts/i18n-check.mjs` validates against its own copy, but the new test doesn't assert parity between `I18N_INIT_OPTIONS.ns`, `scripts/i18n-check.mjs`'s `NAMESPACES`, and `src/test/setup.ts`. That's a few lines here (read the two files, compare sets) and it would make the drift impossible rather than merely documented in a PR description. Worth adding in this PR since it's the cheapest part of the fix.

### Nits

- `stripComments` removes `/* ... */` file-wide with a non-greedy regex; a `/*` inside a string or regex literal will pair with the next `*/` and delete real code between them — a false *negative* for the gate. Cheap mitigation: only strip block comments on lines where `/*` is the first non-whitespace token, which covers the docblock case that motivated it.
- Use `readdirSync(dir, { withFileTypes: true })` instead of `statSync` per entry: fewer syscalls and it won't throw on a broken symlink.
- `__dirname` works under vite-node today but isn't ESM-portable; `fileURLToPath(new URL('..', import.meta.url))` is more robust if the runner config changes.
- The directory skip list is only `node_modules` and `test`; consider `__tests__` and `.spec.ts(x)` so those files are excluded for the same reason `.test.ts` is.
- The comment "Same extraction rule as i18n-check.mjs" for the `*Key:` regex is a claim that will silently rot — if the two regexes are meant to stay in lockstep, that's another argument for reading the script's list/pattern rather than restating it.

---

perrozzi:
## Round 1 disposition

All three "should fix" taken, and the third one found a live bug. Four of five nits taken.

**1. The scan under-matched silently.** You were right, and more right than the review knew: `useTranslation(['common', 'categories'])` **is already in use** in `pages/Categories/FeedbackResults.tsx` (three times), so the array form was not a hypothetical blind spot — the gate was skipping real declarations from day one. And `pages/Chat/Chat.tsx` destructures `t: translate`, so the `\bt\(` anchor missed that file's calls too.

So rather than asserting those forms are unused — which would have failed on correct code — `scanSource` now handles both: the array form is parsed, and every name `t` is renamed to in a file becomes an additional caller pattern. What remains is the one form no static scan can resolve, a namespace passed as a variable or template literal, and *that* is now asserted absent. Zero occurrences today.

Checked the consequences of the array form while I was there: `FeedbackResults.tsx` is well behaved — its five cross-namespace keys are explicitly `categories:*` and its six bare keys all live in `common`, which is the namespace `i18n-check.mjs` falls back to. So there is **no live mis-attribution**, but the gate's blind spot was real. (`showingOf` looked missing from both catalogues until I remembered plurals — it is `showingOf_one`/`showingOf_other`.)

**2. Sentinels pinning names and a magic number.** Replaced. Liveness is now structural: the walk must visit more than 100 files, and more than three of the namespaces it finds must exist as shipped catalogues. No namespace is named in an assertion, so a rename no longer fails as though the app were broken.

**3. The four copies are still unguarded — and they had already drifted.** Added the parity test, and it failed on arrival: **`scripts/fix-i18n.mjs` listed a 16th namespace, `feedback`, with no catalogue in any locale.** Every run of that script tried to load `locales/<lang>/feedback.json`, which does not exist. Removed it, and the four copies now agree. The test compares each copy against the **shipped catalogue files** rather than against each other, since comparing them pairwise would let all four drift together — a namespace is a catalogue.

### Nits

- **Taken** — `stripComments` now strips block comments only where `/*` opens the line, so a `/*` inside a string or regex cannot pair with a later `*/` and delete real code. That direction is a false negative, the worse one for a gate.
- **Taken** — `readdirSync(dir, { withFileTypes: true })`.
- **Taken** — skips `__tests__` and `.spec.ts(x)`.
- **Taken** — dropped the "same extraction rule as i18n-check.mjs" claim; the parity test now enforces the part that matters (the list), and the regex similarity is no longer asserted in prose.
- **Declined** — `__dirname` over `fileURLToPath(import.meta.url)`: it is the established pattern in this repo's tests (`ValidationLinkPicker.test.tsx` reads its own source that way). Worth changing repo-wide if the runner config ever moves, not in one file.

### Verification

- **2463/2463** tests, `tsc --noEmit` and eslint clean over `src/` and `scripts/`.
- `i18n-check.mjs` output **byte-identical** to the merged `development` tip, measured by stashing these two files and re-running. The ja/zh/ko `extra: 4` entries in `categories.json` are pre-existing, not from this PR.
- **Seven mutations, each caught by the intended test and restored.** Beyond the three from the first push: unregistering a namespace declared via the **array** form (proves it is genuinely scanned, not merely tolerated), putting the stale `feedback` back in `fix-i18n.mjs` (parity gate), a non-literal namespace (fidelity gate), and a qualified key reached through the **renamed** `t` (proves alias resolution — without it, that is exactly the silent blind spot you flagged).

perrozzi:

Since my last review the author reworked the scan substantially: array and aliased-`t` forms are now matched, an explicit test bans statically-unresolvable `useTranslation` arguments, the name-pinned sentinels were replaced with structural liveness checks, a new test asserts all four namespace-list copies against the shipped `public/locales/en` catalogues, `stripComments` is now line-anchored, and the walk uses `withFileTypes` with `.spec`/`__tests__` exclusions. All three "Should fix" items and most nits are addressed; the one substantive new thing is that this PR is no longer test-only — `scripts/fix-i18n.mjs` drops `feedback` from `NAMESPACES` to satisfy the new parity assertion.

### Resolved
- Scan under-matching: array form, renamed-`t` aliases handled; non-literal namespaces asserted absent (`src/i18n/options.test.ts`).
- Sentinel assertions no longer pin namespace names; liveness is now file-count + intersection with shipped catalogues.
- Four-copy drift is now gated against the catalogue files rather than only documented.
- Nits: line-anchored block-comment stripping, `withFileTypes`, `__tests__`/`.spec` exclusions, removed the "same extraction rule as i18n-check.mjs" claim.

### Still open
- `SRC = join(__dirname, '..')` — `__dirname` is not ESM-portable; `fileURLToPath(new URL('..', import.meta.url))` is safer (nit only).

### New findings
- **`scripts/fix-i18n.mjs` is production tooling and the PR title/description still say "test-only, no production code."** Removing `feedback` is likely correct, but the claim should be corrected so a reviewer of the merge commit knows a script's behaviour changed; if a catalogue for `feedback` is expected later, the script now silently won't translate it.
- **Parity is checked against `en` only** (`LOCALES_EN`). If a non-English locale carries an orphan catalogue (e.g. `fr/feedback.json`), nothing flags it and `fix-i18n.mjs` will now skip it forever. Cheap addition: assert every `public/locales/*` directory has the same catalogue set.
- **The alias regex `/\bt\s*:\s*(\w+)/g` also matches type annotations and params** (`{ t: TFunction }`, `(t: number)`), so `callers` can pick up type/identifier names. Harmless today (extra callers only over-match), but scoping it to destructuring (`\{[^}]*\bt\s*:\s*(\w+)`) would keep the set meaningful.
- Minor: in the fidelity test, `arg.startsWith(')')` is dead — the capture is `[^)]{0,40}`, so a `)` can never be the first char; and `filesScanned > 100` reintroduces a magic threshold that will fail on a legitimate large deletion (lower it or derive it, e.g. > 20).

Otherwise the incremental changes look clean and the mutation-testing evidence covers the new gates.

---

perrozzi:
## Round 2 disposition

All four new findings addressed. One of them was a correctness point about the PR itself, which I've fixed at the source.

**1. The description still claimed "test-only, no production code."** You're right, and that stopped being true the moment round 1's parity assertion made me delete the stale `feedback` entry from `scripts/fix-i18n.mjs`. **The PR description is now corrected** with an explicit scope note, and the change is called out in its own section rather than buried — a reviewer of the merge commit will see that a script's behaviour changed. On your conditional: if a `feedback` catalogue is expected later, the new parity gate is what will require it to be added to all four lists at once, so the script cannot silently skip it. That is stronger than leaving the entry in place pointing at a file that does not exist.

**2. Parity was checked against `en` only.** Fixed — every locale directory must now ship the same catalogue set. Measured before adding it: all 8 locales ship exactly the 15 catalogues, so nothing diverges today. Both directions are gated, since the reverse case (a locale *missing* a catalogue `en` ships) makes that page fall back to English with no other signal.

**3. The alias regex matched type annotations.** Correct — `/\bt\s*:\s*(\w+)/` picks up `{ t: TFunction }` and `(t: number)`, so `callers` collected type names. Scoped to a destructuring pattern as you suggested. Two things verified rather than assumed: Chat.tsx's rename spans several lines and still matches (the negated class spans newlines), and a **negative control** — a planted `{ t: GhostType }` annotation with a `GhostType('x:y')` call must *not* be treated as a caller, and the gate correctly stays green.

**4. Dead branch and the magic threshold.** Both fixed. `arg.startsWith(')')` was indeed unreachable given the `[^)]{0,40}` capture. The file count drops from 100 to 20, with a comment recording what it is for: proving the walk *ran*, not asserting how much code exists. A high threshold would fail on a legitimate large deletion, which is the mistake I made the first time.

### Held

**`__dirname` vs `fileURLToPath(new URL('..', import.meta.url))`.** Still declining, same reason as round 1: `__dirname` is the established pattern in this repo's tests — `ValidationLinkPicker.test.tsx` reads its own source that way — and mixing the two styles inside one directory is worse than either alone. It's a fair nit; the right form is a repo-wide change if the runner config ever moves, not one file diverging.

### Verification

- **2464/2464** tests, `tsc --noEmit` and eslint clean over `src/` and `scripts/`.
- Four more mutations, each caught by the intended test and the file restored: an orphan `fr/feedback.json`, `de` missing `scrapers.json`, a qualified key reached through the renamed `t`, and the type-annotation negative control that must stay green.
- Eleven mutations total across this PR, every one with a green baseline asserted first.

perrozzi:

Since my last review the author scoped the alias regex to a brace pattern, added a per-locale catalogue-parity test (`ships the same catalogue set in every locale`), lowered the `filesScanned` floor to 20, dropped the dead `arg.startsWith(')')` branch, and corrected the PR description's scope note to acknowledge the `scripts/fix-i18n.mjs` change. All four "New findings" from round 1 are addressed.

### Resolved
- Parity now checked across every locale directory, not just `en` (`src/i18n/options.test.ts`) — closes the `fr/feedback.json` orphan hole.
- Alias regex narrowed so it no longer matches bare `(t: number)`-style parameters.
- Magic `> 100` file threshold lowered to `> 20` with a rationale comment.
- Dead `arg.startsWith(')')` branch removed.
- PR description corrected: the `fix-i18n.mjs` change is now called out as production tooling.

### Still open
- `SRC = join(__dirname, '..')` — not ESM-portable; `fileURLToPath(new URL('..', import.meta.url))` (nit only, and the file is CJS-transpiled under vitest today).

### New findings
- Nit, comment accuracy: `src/i18n/options.test.ts` claims the pattern `/\{[^{}]*\bt\s*:\s*(\w+)/` is "scoped to a destructuring pattern … so type annotations … don't join the caller set", but `[^{}]*` cannot distinguish `const { t: tr } =` from a type literal `{ t: TFunction }` or `({ t }: { t: TFunction })` — those still add `TFunction` to `callers`. The consequence is still only harmless over-matching (as the negative control shows), so the code is fine; the comment overstates what the regex proves. Also note `[^{}]*` now excludes any destructuring containing a nested object literal before the `t:` — currently unused, but a future `const { data: { x }, t: tr }` would silently drop out of the caller set.
- Nit: `localeDirs()` treats any directory under `public/locales` as a locale, so adding a non-locale subdirectory (e.g. a `_templates/`) would fail the new parity test with a confusing message rather than being skipped.

---

perrozzi:
## Round 3 disposition

Both nits taken, and the first one is a correction I should make explicitly rather than quietly.

**1. The comment overstated what the regex proves.** You're right: `/\{[^{}]*\bt\s*:\s*(\w+)/` cannot tell `const { t: tr } =` from a type literal `{ t: TFunction }`, so `TFunction` does still join `callers`. My comment claimed the scoping keeps type annotations out, and **my round-2 disposition repeated the same overstatement** — I described the negative control as proving "a type annotation is not treated as a caller" when what it actually proves is that being treated as one is *harmless*. Those are different claims and I stated the stronger one.

The code stays as it is, now with an honest comment: the two directions are not symmetric, so the regex is *deliberately* generous. An extra caller name matches no call site and changes nothing; a missed alias is the silent blind spot the whole scan exists to avoid. Brace scoping is still worth keeping because it drops bare `(t: number)` parameters, which is a real reduction — just not the guarantee I wrote. Your `[^{}]*` observation about nested objects (`const { data: { x }, t: tr }`) is now recorded in the comment as a known limitation, unused today.

**2. `localeDirs()` treated any directory as a locale.** Fixed, and by a better route than filtering names: the test now iterates `supportedLngs` — the set the language detector will actually accept — so an unrelated `_templates/` is simply not a locale and is skipped. That also let me add the case your suggestion implies but doesn't cover: a locale declared in `supportedLngs` with **no catalogue directory at all** now fails with that named message instead of an `ENOENT` from `readdirSync`.

### Held

**`__dirname`.** Same reason as rounds 1 and 2, and you've noted yourself that the file is CJS-transpiled under vitest today. `ValidationLinkPicker.test.tsx` reads its own source the same way; one file diverging is worse than the whole directory being consistent. A repo-wide change if the runner config ever moves.

### Verification

- **2464/2464** tests, `tsc --noEmit` and eslint clean over `src/` and `scripts/`. All 8 locale directories intact afterwards, no probe residue.
- Three checks for this round, each restoring the tree:
  - a non-locale directory under `public/locales` → gate correctly **stays green** (negative control, and it's the fix)
  - a locale declared in `supportedLngs` with no directory → fails, **with the named message** rather than a crash
  - an orphan `fr/ghost.json` → still caught, which is the one that matters: it proves the `readdir` → `supportedLngs` switch did not quietly weaken the gate it was meant to preserve
- Fourteen mutations/controls across this PR, each with a green baseline asserted first.

perrozzi:

Since my last review the author replaced `localeDirs()` (filesystem enumeration) with `supportedLocales()` derived from `I18N_INIT_OPTIONS.supportedLngs`, added an `existsSync` guard so a declared-but-missing locale fails with a named message instead of ENOENT, and rewrote the alias-regex comment to state honestly that brace-scoping does not distinguish a destructuring from a type literal and that `[^{}]*` cannot cross a nested object. Both round-2 nits are addressed.

### Resolved
- Comment on `/\{[^{}]*\bt\s*:\s*(\w+)/` no longer overstates what the regex proves, and documents the nested-object caveat (`src/i18n/options.test.ts`).
- Non-locale subdirectories under `public/locales` no longer fail the parity test with a misleading message; locale set now comes from `supportedLngs`.

### Still open
- `SRC = join(__dirname, '..')` is not ESM-portable (nit only).

### New findings
- Nit, small coverage loss: `supportedLocales()` in `src/i18n/options.test.ts` now derives the locale set from the very options object under test, so (a) a whole shipped locale directory that is *not* in `supportedLngs` (e.g. a stray `nl/`) is no longer visited at all, where the previous filesystem walk would have flagged it, and (b) pruning an entry from `supportedLngs` while its catalogues remain on disk silently shrinks the parity gate rather than failing it. If you want both properties, assert the filesystem locale dirs and `supportedLngs` are the same set (skipping a documented ignore-list like `_templates`), then iterate that.

---

perrozzi:
## Round 4 disposition

Taken — and it's worth naming that this finding is a coverage loss *I introduced in round 3*. Fixing the `_templates/` false failure by deriving the loop from `supportedLngs` quietly cost two properties the filesystem walk had:

- a shipped locale directory absent from `supportedLngs` was no longer visited at all — translations on disk that nothing can ever select
- pruning an entry from `supportedLngs` while its catalogues remained would *shrink* the gate rather than fail it

Both are back, via your suggestion: **set equality** between the locale directories on disk and `supportedLngs`, asserted before iterating, so neither side can drift alone. Non-locale directories are excluded by convention — a leading `_` or `.` is infrastructure — rather than a hardcoded `_templates`, so round 3's fix survives without a name list to maintain.

The `existsSync` guard from round 3 is gone: the set equality subsumes it, and a declared-but-absent locale now surfaces as a set difference with the difference named. Dropping it rather than keeping both avoids a second assertion that can never fire.

### Verification

Five checks this round, each restoring the tree, deliberately covering the two new directions *and* the two older properties this change could have broken:

| check | expected | result |
|---|---|---|
| extra `nl/` on disk, not in `supportedLngs` | fails | caught |
| prune `ko` from `supportedLngs`, `ko/` stays | fails | caught |
| `_templates/` infrastructure directory | stays green | green (round 3's fix intact) |
| locale declared with no directory | fails | caught |
| orphan `fr/ghost.json` | fails | caught (original hole still closed) |

2464/2464 tests, `tsc --noEmit` and eslint clean over `src/` and `scripts/`. All 8 locale directories intact afterwards.

Nineteen mutations and controls across this PR, each with a green baseline asserted first.

`__dirname` still held, same reason as rounds 1–3.

perrozzi:

Since my last review the only change is in `src/i18n/options.test.ts`: `existsSync` was dropped in favour of a two-way set-equality assertion between `localeDirsOnDisk()` and `supportedLngs`, with a `_`/`.`-prefix convention for non-locale infrastructure directories. That addresses my round-3 nit exactly as suggested, and the failure it replaces (declared locale with no directory) is still reported by name via the set comparison rather than an ENOENT.

### Resolved
- Locale parity gate no longer derives its loop set from the object under test; a shipped-but-unsupported directory and a pruned `supportedLngs` entry now both fail (`src/i18n/options.test.ts`).
- `_templates`-style directories excluded by convention instead of being enumerated as locales.

### Still open
- `SRC = join(__dirname, '..')` is not ESM-portable (nit only).

### New findings
- Nit, wording only: in the set-equality message, "a directory missing here / one missing there" is ambiguous without knowing which side of `toEqual` is actual vs expected — naming the two sides (`on disk` / `in supportedLngs`) would make the failure self-explanatory.

---

perrozzi:
## Round 5 disposition — taken, and closing the loop here

Message reworded to name both sides: locale directories on disk (received) vs `supportedLngs` (expected), with what each direction means. The failure now explains itself without opening the test.

**Stopping the review loop here.** This round was wording-only and the previous one was a single nit, so the severity has been flat-to-declining for two rounds with no production behaviour in question. Trajectory across the five rounds, for the record:

| round | findings | what actually changed |
|---|---|---|
| 1 | 3 should-fix, 5 nits | scan learned the array + aliased-`t` forms; sentinels destructured; parity gate added — **and it found a real bug** |
| 2 | 4 | cross-locale parity; alias regex scoped; dead branch; PR scope corrected |
| 3 | 2 nits | locale set from `supportedLngs`; a comment of mine corrected |
| 4 | 1 nit | set equality both ways — **a coverage loss I had introduced in round 3** |
| 5 | 1 nit (wording) | one assertion message |

Two things earned the rounds. The parity assertion asked for in round 1 **failed on arrival** and exposed a stale namespace (`feedback`) in `scripts/fix-i18n.mjs` with no catalogue in any locale — a real, if quiet, bug in production tooling. And round 4 caught that my round-3 fix had silently narrowed the very gate it was meant to preserve, which is exactly the failure mode this PR is about.

**Final state:** 5 gates in `src/i18n/options.test.ts`, one line removed from `scripts/fix-i18n.mjs`, 2464/2464 tests, `tsc` and eslint clean over `src/` and `scripts/`, `i18n-check.mjs` output byte-identical to the `development` tip. Twenty mutations and controls across the PR, each with a green baseline asserted first and the tree restored afterwards.

Ready for a human merge. `__dirname` remains the one open nit, declined consistently: it matches this repo's existing test convention, and changing it belongs in a repo-wide pass rather than one file.

perrozzi:

Only one thing changed since my last review: the failure message on the locale set-equality assertion in `src/i18n/options.test.ts` now names both sides (`on disk (received)` vs `supportedLngs (expected)`) and spells out what each direction of extra means. That is exactly the wording nit I raised, addressed.

### Resolved
- Set-equality failure message is now self-explanatory about which side is actual vs expected (`src/i18n/options.test.ts`).

### Still open
- `SRC = join(__dirname, '..')` is not ESM-portable (nit only).

No new findings — the incremental change is clean and I have nothing further blocking.

---

perrozzi:

Since my last review the only change is a rewording of the `stripComments` docblock in `src/i18n/options.test.ts` — the escaped `/*` markers that made the previous sentence awkward are now described in prose. No behavioural change, no new findings.

### Still open
- `SRC = join(__dirname, '..')` is not ESM-portable (nit only) — `src/i18n/options.test.ts`.

Clean delta; nothing blocking.

---

perrozzi:
## Does this need a deploy? Measured, not assumed: no

Built this branch and compared every file against the live bundle. Result:

| | files |
|---|---|
| byte-identical to live | 141 |
| would be replaced | 1 (`index.html`) |
| new filenames | 19 (18 JS + 1 CSS) |

The 19 renames are not 19 changes. Only **one** chunk contains anything new: the `APP_VERSION` stamp, which `vite.config.ts` derives from the 7-char git SHA, so it moved from `9ac0498` (what is live) to this branch's head. Every other renamed JS chunk references a renamed sibling by name — a content-hash cascade off that one chunk, confirmed per file. The build is reproducible (two builds of one commit: 161/161 byte-identical), so the comparison against live is meaningful rather than noise.

`scripts/fix-i18n.mjs` cannot affect a deploy at all: it is referenced by **no** npm script — not `build`, not `prebuild`, not `deploy:frontend`. It is a manual maintenance tool.

### One thing I'd have got wrong by reasoning

The CSS also got a new name, which a JS rename cascade cannot explain. Diffed against the live stylesheet: **+36 bytes, exactly one added selector — `.shrink`.** Tailwind scans source files for candidate utility names, my new test file is a source file, and it contains the word "shrink" in a **prose comment** ("would quietly shrink this gate"). So a word in a comment emitted a production CSS rule.

Harmless — an unused utility class — but worth recording, because it means "test-only change" does not strictly imply "identical bundle" in this repo. Anyone auditing why a test-only PR moved the CSS hash now has the answer.

### On the stale namespace, stated precisely

`loadJSON` returns `{}` for a missing file, so with `feedback` in the list both sides were empty, `toFix` was empty, and the loop `continue`d before `saveJSON`. The entry was **inert** — one wasted iteration per language, no file created, nothing corrupted. Removing it is correctness and hygiene, not a bug fix with user impact. I checked this rather than assuming the worse case, and did not run the script: it writes catalogue files and has no dry-run flag.

### Testing

Nothing further applies. The change *is* tests, verified by 20 mutations and controls rather than by assertion, and it adds no runtime surface — no UI, no API route, no Lambda, nothing a browser or `test-api.sh` could exercise. The production line removed is verified statically above.

So: mergeable as-is, and **no deploy is owed**. If one happens later for another reason, it will pick up a new version stamp and a dead `.shrink` rule, and nothing else.

PR #327, opened 2026-08-14
fix(personas): refuse PDF import instead of inventing a persona

perrozzi:
## The bug

Persona import shipped a **PDF** option that could not work. Nothing in this repo extracts
text from a PDF, and the importer never tried to. In `lambda/jobs/persona_importer/handler.py`
the non-image branch was:

```python
text_content = content if input_type == 'text' else "[PDF content - extract persona from this document]"
```

The uploaded file was **never read**. That one sentence was sent to the model as if it were
the document, and the model did what it was asked: it produced a complete, plausible, fully
detailed persona — name, goals, frustrations, quotes — with no connection whatsoever to the
file the user uploaded. The job then finished as an ordinary `completed`, the persona was
written to the project alongside genuinely derived ones, and nothing in the UI, the job
record or the logs said any of it was invented.

That is the worst failure shape this product has: not an error, but a confident answer
about data that was never looked at. A user who uploads a research deck and reviews the
persona it "found" has no signal at all that they are reading fiction.

## The fix

One shared contract, enforced at three independent layers.

`lambda/shared/persona_import.py` is the single source of what persona import can read: the
allowlist, the deferred set, the default, normalisation and every refusal message. Both
layers that must agree import it, and a test asserts they resolve to the same function
object, so they cannot drift.

**G1 — remove the affordance.** `ImportPersonaModal.tsx` loses the PDF button, its
`accept=".pdf,application/pdf"` filter and its "PDF files only" hint. The type union becomes
`'image' | 'text'` and is narrowed all the way up its chain (`useModalState`,
`ProjectModals`, `useProjectData`, `projectsApi`, `client`) so the compiler now refuses a
reintroduction rather than leaving it to review.

**G2 — reject at the API.** `projects_handler.api_import_persona` validates **before**
`create_job`, so a refused import leaves no job row, no Lambda invoke and no Bedrock spend.
It checks the whole payload, not just the type: `pdf` gets a *distinct* "not supported yet"
message (kept separate from "unsupported" — "never" and "not yet" are different answers),
blank content is refused, and for images the `media_type` is allowlisted. Type is checked
before content, so a PDF is told it is a PDF rather than told it is empty.

**G3 — refuse rather than fabricate.** A replayed invoke, an already-queued message or a
future caller still reaches the job, so it validates again rather than trusting the
boundary, and raises instead of inventing. The message is user-facing (`job_handler` writes
it into the job record), so it says the file could not be read rather than naming an
`input_type` the user never chose.

Along the way, `shared/image_limits.py` now separates the **Converse-readable** image set
from the **S3 extension** map it previously conflated. They share keys but not values
(Converse wants `jpeg`; the S3 extension is `jpg`), and a test asserts the key sets stay
equal so adding a type for storage cannot silently widen what the model is asked to read.

## Behaviour changes for existing callers

Two, both deliberate, both tightenings on a path that previously accepted input it could not
honestly process:

- **Blank `content` is now a 400** rather than a job that fails later. Previously an empty
  import bought a job row, an async Lambda invoke and a failure inside the job.
- **An image with a missing or unreadable `media_type` is now a 400.** Previously a blank
  media type was assumed to be PNG, and `application/pdf` passed the type allowlist and
  reached Converse as an image block claiming format `pdf`. The accepted four are exactly
  what the file picker already advertises, so anything chosen through the UI passes.

The long-standing default is preserved: a missing or blank `input_type` still means `text`,
at both layers.

## Why this PR exists now

This is the interim that the product-doc upload work (#325) owed and I did not pay. That
change deferred pdf/docx to a later rung, and the plan of record calls out the consequence
explicitly: deferring pdf leaves this bug live, and *"the cheapest honest interim there is to
reject pdf rather than let it invent a persona from a placeholder sentence."* #325 shipped
without it. This closes that gap.

It is filed as a **new bug**, not an amendment to the upload epic — the placeholder is
pre-existing behaviour in a different code path, not a regression from or a subset of that
work.

## Verification

Tests written to fail on their own guard rather than on the happy path:

- pdf at the API returns 4xx **and** `create_job` was never called **and** the importer was
  never invoked — a rejection that still leaked a job row would pass on the status alone
- the pdf and unsupported messages are asserted separately **and** asserted to differ
- text and image still create a job; absent and blank `input_type` still default to text.
  Without these, "refuses pdf" is indistinguishable from "refuses everything"
- in the job: pdf raises **and** `bedrock.converse` was never called — a raise on its own
  cannot tell a refusal from a failed model call. Blank content and unreadable media types,
  likewise, each with a positive control over the four accepted formats
- the placeholder literal appears nowhere in the handler source, with a positive control so
  the assertion cannot pass vacuously
- the PDF option is not rendered **and** the image/text options are; `accept` contains no pdf

**Mutation-probed 22 ways: 21 caught, 0 skipped, every file restored byte-identical.**
Remove either allowlist, remove the empty-content guard, collapse the two refusals, drop the
deferred branch, drop normalisation of the type or the media type, delete the non-string
guard, break the blank-to-text default, stop validating in the job, stop validating at the
API, restore the placeholder verbatim, reintroduce a private tuple, allowlist against the
extension map, widen the storage map alone, give Converse the S3 spelling of jpeg, re-add the
PDF button, re-add `.pdf` to accept, restore "PDF files only", drop the upload heading,
re-enable Import on whitespace.

One mutation is recorded as an **equivalent mutant** rather than quietly dropped: coercing a
non-string with `str(raw)` still answers 400 with the same message *and* still keeps a
non-string away from the `in` operator, so no test can distinguish it from `return ''`. The
property worth pinning is that a raw non-string never reaches `in DEFERRED_INPUT_TYPES`, so
the replacement mutation deletes the `isinstance` check outright — caught by the list case,
which would otherwise 500.

**Gates**, measured as a delta against the base commit rather than against green:

| Gate | Base | This branch |
|---|---|---|
| backend pytest | 1907 | 1986 |
| frontend vitest | 2496 | 2503 |
| `test:cdk` | 239 | 239 |
| `test:stream` | 303 | 303 |
| eslint (frontend + stream) | clean | clean |
| `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553, per-rule histogram unchanged |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |

No infrastructure change, so nothing here needs a deploy beyond the normal Lambda and
frontend publish.


perrozzi:

This PR removes the PDF persona-import path that never read the uploaded file and instead fed the model a placeholder sentence, producing wholly fabricated personas presented as `completed`. It closes the hole at three layers (UI affordance removal + type narrowing, API allowlist before `create_job`, job-handler refusal with an empty-content guard) and adds focused regression tests on both sides. The change is well targeted and the tests do assert the right thing (no `create_job`, no `invoke_lambda_async`, no `bedrock.converse`), which is exactly the "fix must fail when reverted" bar. No blockers; a few consistency and i18n items below.

### Should fix

- **Duplicated allowlist with only a comment holding it together.** `lambda/api/projects_handler.py` defines `SUPPORTED_IMPORT_TYPES = ('text', 'image')` and `lambda/jobs/persona_importer/handler.py` independently defines `SUPPORTED_INPUT_TYPES = ('text', 'image')`, tied together by prose ("Kept in step with..."). When PDF (or docx) lands, these will drift — and drift in the direction of the API accepting something the job refuses. Put the tuple (and ideally `DEFERRED_IMPORT_TYPES`) in a shared module both Lambdas already import from, and have the job compare against that.

- **Boundary validates `input_type` but not the rest of the payload.** `api_import_persona` still forwards `content: body.get('content', '')` unchecked, so a blank/whitespace `content` creates a job row and an async Lambda invoke that then fails in the job — the exact cost the PR argues against for `pdf`. Same for `media_type`: `input_type='image'` with `media_type='application/pdf'` (or missing) passes the allowlist and reaches the Converse image block, where the format is derived from `media_type`. Mirror the job's empty-content guard at the API and allowlist the four image media types you advertise in `ACCEPTED_FILE_TYPES`.

- **New user-visible strings are hardcoded English.** `ImportPersonaModal.tsx` now hardcodes `Upload Image` and `FILE_TYPE_HINT = 'PNG, JPG, GIF, WebP'` (previously hardcoded inside the deleted helpers, so this is a lateral move, but the refactor is the moment to fix it). Repo convention is an i18n key per user-visible string in all locales; the new tests in `ImportPersonaModal.test.tsx` now pin the English literals, which locks the omission in.

### Nits

- Comment-to-code ratio in `lambda/jobs/persona_importer/handler.py`: ~35 lines of narrative for a two-clause guard. The history belongs in the test docstring and PR; keep the inline comment to the invariant ("refuse, never substitute placeholder content — see test_unsupported_input.py").
- The retained `importPersona.pdf` / `pdfDesc` keys are defensible, but add a `TODO(<issue>)` next to them in the catalogues (or in the modal comment) so they are traceable rather than mystery dead keys.
- `ImportPersonaModal.test.tsx::importTypeButtons()` reaches through `heading.parentElement?.querySelector('div')`; a markup change (e.g. an extra wrapper div) breaks the test for unrelated reasons. `within(...)` on a container with a `data-testid`, or `getAllByRole('button')` filtered by accessible name, is more durable.
- The diff doesn't show `projects_handler.py`'s import block — confirm `ValidationError` is already in scope there (and that it maps to 400 via the existing exception handler, which the new tests assume).
- Consider emitting a powertools metric on refusal (deferred vs unsupported) so repeated PDF attempts are visible in CloudWatch rather than only in individual job records — useful signal for prioritising real PDF extraction.

---

perrozzi:
## Round 1 disposition

All three "should fix" items were real and are fixed. Three of five nits fixed, one answered, one declined. Pushing `58edac9`.

### Should fix — all three taken

**1. Duplicated allowlist held together by a comment.** Right, and right about the direction of the drift: the API accepting something the job refuses turns a clear 400 into a background job that fails later. Both layers now import one validator from a new `shared/persona_import.py`, which owns the allowlist, the deferred map, the default, normalisation and every message.

This came out *smaller* than the two copies were — the job's private tuple, its own refusal wording and the API's separate message constants are all gone. And rather than trust a reviewer to keep them together, `test_the_api_validates_with_the_same_callable_the_job_uses` asserts both modules resolve to the same function object, so they cannot diverge at all.

Defence in depth is unchanged: the job still calls the validator itself and does not assume the boundary ran.

**2. Boundary validated `input_type` and nothing else.** The strongest of the three — my own argument for rejecting at the API convicted the gap.

- blank `content` was forwarded unchecked, so an empty import still bought a job row, a Lambda invoke and a failure inside the job. That is exactly the cost I claimed the boundary existed to avoid. Refused before `create_job` now.
- `input_type='image'` with `media_type='application/pdf'` passed the type allowlist and reached Converse as an image block with format `pdf` — a PDF routed straight around the PDF refusal.

For the media types I took the reuse rather than a new list: `IMAGE_CONTENT_TYPE_EXTENSIONS` in `shared/image_limits.py` already holds exactly these four, already documented there as the only formats Converse can read. Type is checked before content, so a PDF is told it is a PDF rather than told it is empty; a test pins that order.

One deliberate **behaviour change** to flag: a blank `media_type` on an image is now refused rather than assumed to be PNG. Guessing a format is a silent wrong answer of the same family as this PR's bug, and these four are exactly what the picker already advertises, so anything chosen through the UI passes.

**3. New hardcoded English.** Sharper than "a lateral move", and worth stating plainly: all four literals in `FileUploadSection` already had translated keys in every catalogue — `uploadImage`, `clickToUpload`, `clickToChange`, `imageFormats`. The pre-existing code hardcoded them anyway and my refactor carried them forward past the keys sitting right there. All four now read from the catalogue, so there are no catalogue edits and no new untranslated strings.

While there: my earlier claim of two retained PDF keys was wrong. There are five — `pdf`, `pdfDesc`, `pdfOnly`, `uploadPdf`, `importTypePdf`.

### Nits

- **Comment-to-code ratio in the job handler** — fair. ~35 lines of narrative cut to the invariant plus a pointer to the test.
- **Brittle test selector** — fixed. Options are found by accessible name; the DOM walk broke on any added wrapper, which would have reported the PDF button as gone for a reason unrelated to the PDF button.
- **Traceability for the retained keys** — recorded next to the constant, but **not** as a `TODO` tag: `sonarjs/todo-tag` fails the lint gate on the literal word, so the suggestion as written cannot land here.
- **`ValidationError` in scope** — confirmed, no change needed. Imported from `shared.exceptions` in `projects_handler.py` and already used at about ten sites; `status_code = 400`, which is what the new tests assert.
- **Powertools metric on refusal** — declined. `projects_handler.py` emits no metrics at all today, so this would introduce that convention inside a bug fix. It is worth doing as its own change when PDF extraction is prioritised, which is when the signal would actually be read.

### Also, unprompted

Removed a fallback the media_type guard makes unreachable, and pinned the `jpeg`/`jpg` distinction with a test: Converse wants the subtype (`jpeg`), while the extension map holds the S3 file extension (`jpg`), and the two now sit side by side in the same module.

### Verification

Mutation-probed 18 ways: **17 caught, 0 survived that matter, 0 skipped.** Remove the media_type allowlist, remove the empty-content guard, collapse the two refusals, remove the type allowlist, drop the deferred branch, drop normalisation, delete the non-string guard, break the blank-to-text default, stop validating in the job, restore the placeholder verbatim, send the extension instead of the subtype, stop validating at the API, reintroduce the private tuple, re-add the PDF button, re-add `.pdf` to accept, restore "PDF files only", drop the upload heading.

One mutation is recorded as an **equivalent mutant** rather than quietly dropped: coercing a non-string with `str(raw)` still answers 400 with the same message *and* still keeps a non-string away from the `in` operator, so no test can distinguish it from `return ''`. The property worth pinning is that a raw non-string never reaches `in DEFERRED_INPUT_TYPES`, so the replacement mutation deletes the `isinstance` check outright — caught by the list case, which would otherwise 500.

| Gate | Before | After |
|---|---|---|
| backend pytest | 1944 | 1983 |
| frontend vitest | 2501 | 2501 |
| `test:cdk` | 239 | 239 |
| `test:stream` | 303 | 303 |
| eslint (frontend + stream) | clean | clean |
| `lint:python` | 553 | 553 |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |

No `importPersona` key appears among the used-but-missing set.

perrozzi:

Since the last review the author extracted the allowlist into a new `lambda/shared/persona_import.py` that both layers import (with an identity test pinning that), extended boundary validation to `content` and image `media_type` (plus job-side counterparts), replaced the hardcoded `Upload Image` / `PNG, JPG, GIF, WebP` strings with `importPersona.*` keys, trimmed the narrative comment block in the job handler, and rewrote the frontend option selectors to use accessible names. All three previous should-fix items are addressed; the delta is clean apart from two consistency concerns below.

### Resolved
- Duplicated allowlist gone: `shared/persona_import.py` is now the single contract, imported by `projects_handler.py` and `jobs/persona_importer/handler.py`.
- Boundary now validates `content` (blank/non-string) and image `media_type` before `create_job`, with no-job/no-invoke assertions and positive controls.
- New user-visible strings moved to i18n keys (`uploadImage`, `clickToUpload`, `imageFormats`, `clickToChange`).
- `ImportPersonaModal.test.tsx` no longer walks the DOM from the heading; options are matched by accessible name.

### New findings

**Should fix — image allowlist is derived from an S3 extension map, not the Converse-readable set** (`lambda/shared/persona_import.py`). `IMAGE_CONTENT_TYPE_EXTENSIONS` exists to map content types to S3 file extensions and is shared with unrelated upload paths; adding e.g. `image/svg+xml: svg` there for avatars would silently widen persona import to a format Bedrock Converse rejects, turning a clean 400 into a failed model call. The set of Converse-readable image types belongs in `persona_import.py` explicitly (the handler already can't use the map's values, since it derives `format` from the subtype). Related: `UNSUPPORTED_IMAGE_MESSAGE` renders extensions (`gif, jpg, png, webp`) while the modal hint says `PNG, JPG, GIF, WebP` — two different vocabularies for the same four formats.

**Should fix — no locale-catalogue changes accompany the four new keys.** Neither diff touches `frontend/src/locales/**`, so confirm `importPersona.uploadImage`, `clickToUpload`, `imageFormats` and `clickToChange` already exist in all eight catalogues; if any is new, `i18n:check` missing goes non-zero and the new tests (`getByRole('heading', { name: 'Upload Image' })`, `getByText('PNG, JPG, GIF, WebP')`) pass only because English falls back to the key text.

### Nits
- PR description and gate table predate this delta (no mention of `shared/persona_import.py`, and blank-`content`/`media_type` rejection at the API is a behaviour change for existing callers); re-measure the counts and note the new refusals.
- `TestBothLayersShareOneContract` asserts only that both modules bound the same symbol — it can't tell that either actually calls it; the behavioural tests carry that weight, so keep them the primary guard.
- Verify `ValidationError` isn't referenced elsewhere in `jobs/persona_importer/handler.py` now that its import was dropped.
- Consider disabling the modal's Import button on blank content so the new 400 isn't the user's first feedback for an empty paste.

---

perrozzi:
## Round 2 disposition

Both should-fix items handled. One was a real design error on my part; the other is answered with evidence and needed no code change. Pushing `7676d7a`, and the PR description is rewritten to match this delta.

### Should fix

**1. The image allowlist was derived from an S3 extension map.** Correct, and the sharpest finding in either round. `IMAGE_CONTENT_TYPE_EXTENSIONS` maps content type to S3 file extension and is shared with the product-doc upload path, where `ALLOWED_CONTENT_TYPES` is built from it. Allowlisting persona import against it meant a type added there *for storage* would silently widen what the model is asked to read. The module's docstring warned against exactly that, which is not a guard — and I have spent two rounds arguing that comments are not guards.

`shared/image_limits.py` now declares `CONVERSE_IMAGE_FORMATS`, the same four content types mapped to the `format` a Converse image block wants, and `persona_import` allowlists against that set by name.

This is not round 1's duplication coming back. The two maps are separate because their **values genuinely differ** — Converse names JPEG `jpeg`, the S3 extension is `jpg` — and a test asserts they hold the same **keys**, so adding a type to one fails until the other is answered. A tripwire, not a coupling. The jpeg/jpg difference is pinned too, so the two cannot be "tidied" back into one map.

Knock-on: the job no longer derives the format by string surgery; `converse_image_format()` is the one place a media type becomes a format. Being honest about how much that buys — for the four accepted types, splitting the subtype and looking it up **agree**, so this is not fixing a live bug. It earns its keep the moment a type is added whose subtype is not its Converse format. I retired the mutation that probed the old splitting rather than leave it passing against a line that no longer exists.

On the vocabulary clash: `UNSUPPORTED_IMAGE_MESSAGE` now lists accepted **content types** rather than prose extensions. `media_type` is the field the caller got wrong, and "PNG, JPG" says less to a client that sent `application/pdf` than naming what the field takes. The modal's translated hint stays the prose version.

**2. No catalogue changes accompany the four new keys.** Confirmed as asked, and no change is needed: `uploadImage`, `clickToUpload`, `imageFormats` and `clickToChange` are all in **8/8 source catalogues** under `frontend/public/locales/`, with real translations rather than English copies — ja `"画像をアップロード"`, de `"Bild hochladen"`. `i18n:check` reports missing 0.

The fallback worry is self-answering, which is worth stating because it makes the test stronger than it looks: a missing key makes `t()` return the key string, so `getByRole('heading', { name: 'Upload Image' })` would **fail**, not pass. Those tests passing is the evidence the keys resolve.

### Nits

- **Import button on blank content** — good catch, and it was half-done: the button was already disabled on empty content, but with `=== ''`, so a whitespace-only paste enabled it and the new 400 became the user's first feedback. It now uses the same `.trim()` predicate as the server guard, with a test and a control asserting it still enables on real content.
- **`ValidationError` after the import was dropped** — verified: not referenced anywhere in the job handler. No `NameError` waiting on an error path.
- **`TestBothLayersShareOneContract` only pins symbol identity** — agreed, and it is not carrying the weight. The behavioural guards are what mutations M9 and M12 kill: stop validating in the job, or stop validating at the API, and named behavioural tests fail. The identity test is there to catch the specific regression of someone re-adding a private copy, which it does.
- **Stale PR description** — rewritten, including an explicit section for the two behaviour changes to existing callers (blank `content`, and image `media_type`).

### A regression I introduced this round, and how it surfaced

Round 2's first cut took `lint:python` from **553 to 554**: replacing the intermediate assignment left a bare nested `if`, which trips `SIM102`. Collapsed to a single condition; back to exactly 553 with no `SIM102` in the per-rule histogram.

Worth naming because it is the argument for the whole delta-measuring habit: on a gate that is already red with 553 pre-existing errors, a one-error regression is invisible if you only look at the colour.

### Verification

**22 mutations, 21 caught, 0 skipped**, every file restored byte-identical. New this round: allowlist against the extension map again (caught), widen the storage map alone (caught by the tripwire), give Converse the S3 spelling of jpeg (caught), stop normalising `media_type` (caught), re-enable Import on whitespace (caught).

M7 remains recorded as an equivalent mutant, unchanged from round 1.

| Gate | Round 1 | Round 2 |
|---|---|---|
| backend pytest | 1983 | 1986 |
| frontend vitest | 2501 | 2503 |
| `test:cdk` | 239 | 239 |
| `test:stream` | 303 | 303 |
| eslint (frontend + stream) | clean | clean |
| `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553 |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |

perrozzi:

Since the last review the author split `CONVERSE_IMAGE_FORMATS` out of `IMAGE_CONTENT_TYPE_EXTENSIONS` in `shared/image_limits.py`, added `converse_image_format()` in `shared/persona_import.py` (now used by both the allowlist and the job's image block instead of `split('/')[-1]`), added a key-set/jpeg-divergence tripwire test, and made the modal's Import button use `trim()` with two new tests. Both previous should-fix items are effectively addressed except the locale-catalogue check, which neither diff touches.

### Resolved
- Persona import now allowlists against the Converse-readable set, not the S3 extension map; widening storage no longer widens what the model is asked to read (`shared/image_limits.py`, `shared/persona_import.py`).
- The two-vocabulary mismatch is gone: `UNSUPPORTED_IMAGE_MESSAGE` names media types, the modal hint keeps the prose form, with the intent documented.
- Job handler no longer derives `format` by string surgery; `jpeg` vs `jpg` is a lookup and is pinned by `test_the_maps_disagree_on_jpeg_which_is_the_whole_point`.
- Import button now disabled on whitespace-only content, matching the server guard, with a positive control test.

### Still open
- No `frontend/src/locales/**` change accompanies `importPersona.uploadImage` / `clickToUpload` / `imageFormats` / `clickToChange` — confirm all eight catalogues already have them or `i18n:check` missing goes non-zero.

### New findings

**Nit — `converse_image_format()` lives in a persona-specific module** (`lambda/shared/persona_import.py`). It's a generic "media type → Converse format" lookup over a constant that lives in `image_limits.py`; any other prompt path needing it would have to import `persona_import`. Consider moving it next to `CONVERSE_IMAGE_FORMATS`.

**Nit — nothing pins the content of `UNSUPPORTED_IMAGE_MESSAGE`.** The boundary and job tests only assert `'could not be read'`, so the derivation from `CONVERSE_IMAGE_FORMATS` could silently stop enumerating the accepted media types; one assertion on `image/png` appearing in the message would close that.

**Nit — `converse_image_format` returns `str | None`** and its result is placed straight into the Converse image block; safe only because validation ran first (which the comment states). A `typecheck` on the Lambda tree, if ever enabled, would flag it.

---

perrozzi:
## Round 3 disposition

No should-fix items. Two nits taken, one declined with reason, and the "still open" item answered — it rests on a path that does not exist. Pushing `1614847`.

### Still open: locale catalogues

There is no `frontend/src/locales/**` in this repo. The catalogues live at **`frontend/public/locales/`**, which is why neither diff touches the path the review is looking at.

All four keys are already there, in **8/8** catalogues, with real translations rather than English copies:

```
ja  "uploadImage": "画像をアップロード"    de  "uploadImage": "Bild hochladen"
ja  "clickToUpload": "クリックしてアップロード、またはドラッグ&ドロップ"
de  "clickToUpload": "Klicken zum Hochladen oder per Drag & Drop"
```

`i18n:check` reports **missing 0**, unchanged across all three rounds.

And the fallback concern is self-answering in a way that makes the tests stronger than they look: a missing key makes `t()` return the key string, so `getByRole('heading', { name: 'Upload Image' })` would **fail**, not pass. Those tests passing is the evidence the keys resolve.

### Nits taken

**`converse_image_format()` lived in a persona-specific module.** Correct — it is a generic lookup over a constant declared in `image_limits.py`, so any other prompt path attaching an image would have had to import `persona_import` to get it. Moved beside `CONVERSE_IMAGE_FORMATS`. `image_limits.py` still imports nothing, so no Lambda loses its dependency-free path to the caps.

This one taught me something worth reporting. After the move, round 2's mutation — allowlist against the S3 extension map — **survived**. Not a weakened test: with the helper reading `image_limits`' own global, aliasing the import at the call site can no longer reach the allowlist, so the mistake is no longer expressible in `persona_import.py` at all. The mutation was retargeted into the helper, which is now the only place it can live, and is caught there. The nit made the code better in a way I would not have found from reading it.

**Nothing pinned the content of `UNSUPPORTED_IMAGE_MESSAGE`.** Also correct, and the same class of bug I have been guarding against elsewhere in this PR: the tests asserted only `'could not be read'`, so the derivation could have silently stopped enumerating and left the caller a refusal that does not say what would have worked. The boundary test now asserts all four accepted media types appear in the 400.

### Nit declined

**`converse_image_format` returns `str | None` into the Converse block.** True, and safe because validation runs immediately above it. Declined because the fix is worse than the finding: narrowing it in the job means adding a `raise` that **no test can reach**, since `validate_import_config` has already refused every media type that would produce `None`. Unreachable code that no mutation can kill is a worse artefact than a precise `Optional`, and there is no typecheck gate on the Lambda tree for this to serve.

`None` is also the right return for a lookup whose "the model cannot read this" answer is a normal result at a validation boundary, with each caller wording its own refusal.

### Verification

**22 mutations, 21 caught, 0 skipped**, every file restored byte-identical.

Two anchors had to be retargeted this round rather than left quietly passing against lines that had moved. The harness reported both as `SKIPPED` and exited 3, which is exactly why it asserts anchor counts instead of trusting them — a mutation that applies to nothing otherwise reads as a guard that held.

M7 remains the single documented equivalent mutant, unchanged since round 1.

| Gate | Value |
|---|---|
| backend pytest | 1986 |
| frontend vitest | 2503 |
| `test:cdk` / `test:stream` | 239 / 303 |
| eslint (frontend + stream) | clean |
| `typecheck:all` | clean |
| `lint:python` | 553, zero `SIM102` |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 |

perrozzi:

Since the last review the author moved `converse_image_format()` out of `shared/persona_import.py` into `shared/image_limits.py` (beside `CONVERSE_IMAGE_FORMATS`), repointed the job handler's import accordingly, and added an assertion in the boundary test that `UNSUPPORTED_IMAGE_MESSAGE` actually enumerates the four accepted media types. Both of the substantive nits from the previous round are addressed; the locale-catalogue question is untouched.

### Resolved
- `converse_image_format()` now lives next to the map it reads in `lambda/shared/image_limits.py`, so other prompt paths need not import `persona_import`.
- `lambda/api/test/test_persona_import_boundary.py` now pins that the image refusal enumerates `image/png|jpeg|gif|webp`, so an empty derivation can no longer pass.

### Still open
- No `frontend/src/locales/**` change accompanies `importPersona.uploadImage` / `clickToUpload` / `imageFormats` / `clickToChange` — confirm all eight catalogues already have them or `i18n:check` missing goes non-zero.

No new problems: `persona_import.py` still legitimately uses both `CONVERSE_IMAGE_FORMATS` (for the message) and the relocated helper, the identity test in `TestBothLayersShareOneContract` is unaffected, and no other module referenced the old location. The delta is clean.

---

perrozzi:
## Round 4 disposition

The repeated locale item turned out to be worth more than the prose answer I gave it twice. I probed it instead, and the probe found a real gap — so this round is a commit (`6986a49`), not a comment.

### The path, first

There is no `frontend/src/locales/**` in this repo. The catalogues are at **`frontend/public/locales/`**, which is why no diff ever touches the watched path and why this item keeps resurfacing.

### What I actually tested

The claim worth checking is not "do the keys exist today" but "would anything fail if one were deleted". Two routes:

- **Removed from English only** — `src/i18n/localeParity.test.ts` already catches this. It asserts every namespace exposes the same key set in all locales, enumerates namespaces from the filesystem, and passes today with 31 tests.
- **Removed from all eight** — parity still passes, because the locales stay equal. Only a test asserting the *rendered* text can catch it.

Route B is where the gap was. Deleting each key from all eight catalogues and re-running the modal suite:

| Key | Verdict |
|---|---|
| `uploadImage` | CAUGHT |
| `clickToUpload` | **SURVIVED** |
| `imageFormats` | CAUGHT |
| `clickToChange` | **SURVIVED** |

I moved all four from hardcoded literals to `t()` in round 1, but only asserted the two the PDF removal happened to touch. A missing key makes `t()` return the key **name**, so either of those two could have been deleted and users would read the literal string `importPersona.clickToUpload` with every gate green.

That is the same shape as the bug this PR fixes — a confident wrong output with nothing saying so — arriving in the UI instead of the model. I would not have found it by re-reading my own argument, which is the point.

Both are now asserted: the empty-dropzone prompt, and the chosen-file prompt with its filename. The second carries a control asserting the empty-state text is **absent**, since the two states are mutually exclusive and asserting only the second would pass on a section that rendered both at once.

Re-probed: **4/4 caught**, all eight catalogues restored byte-identical.

### For the record on the original question

`uploadImage`, `clickToUpload`, `imageFormats` and `clickToChange` are present in **8/8** catalogues with real translations rather than English copies:

```
ja "uploadImage": "画像をアップロード"      de "uploadImage": "Bild hochladen"
```

`i18n:check` reports **missing 0**, unchanged across all four rounds.

| Gate | Round 3 | Round 4 |
|---|---|---|
| frontend vitest | 2503 | 2505 |
| backend pytest | 1986 | 1986 |
| `test:cdk` / `test:stream` | 239 / 303 | 239 / 303 |
| eslint / `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553 |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |

Four commits of review response on top of the original fix. From my side this is ready; happy to keep going if anything above reads as unresolved.

perrozzi:

Since the last review the only change is two new tests in `frontend/src/pages/ProjectDetail/ImportPersonaModal.test.tsx` pinning the remaining catalogue-driven strings in the upload dropzone (`importPersona.clickToUpload`, `importPersona.clickToChange`) plus a mutual-exclusion control between the empty and chosen-file states. Nothing else moved, and the previous delta's findings stay resolved.

### Resolved
- The locale-catalogue concern is now largely covered: every string `FileUploadSection` renders (`uploadImage`, `clickToUpload`, `imageFormats`, `clickToChange`) is asserted against its English value, so a missing key surfaces as a failing test rather than a raw key name in the UI — with `i18n:check` reporting `missing 0` for the non-`en` catalogues.

No new problems: the tests use the existing `defaultProps`, don't touch production code, and the control assertion prevents the chosen-file test from passing on a section that renders both states. Clean delta.

---

PR #329, opened 2026-08-14
feat(prototype): ground a build in the mockups it was aimed at

perrozzi:
## What this does

A prototype build can now be **aimed at specific uploaded mockups**. Their extracted
descriptions go into the generation prompt as a dedicated **visual brief**, instructing the model
to take the eight `:root` custom properties, the layout mode and the corner radii from the
mockup rather than falling back to the built-in neutral indigo theme.

Uploading and extracting images already worked. A mockup's palette, layout mode and component
inventory were extracted, stored, listed — and then deliberately withheld from every prompt.
This spends them.

## The distinction that shapes the whole change

This is **selection and placement**, not "make the visual reach the model." Reaching the model is
the easy half and would arrive on its own: `build_product_context_block` injects the extracted
text of every ready document, so a visual's description would start reaching every
product-context-ticked prototype by itself.

That is a hazard, not a feature. It would arrive under `### Internal documents`, as
`#### screenshot.png`, ordered by `created_at`, sharing one 50,000-character budget with
unrelated files, and placed nowhere near the eight CSS variables the prompt actually acts on — a
paragraph competing with a spec rather than an instruction. The upload work closed that off
behind a single predicate precisely so this change could be **purely additive** instead of
corrective.

## Shape

**Producer.** `build_visual_brief_block(project_id, doc_ids) -> (block, used_doc_ids)` in
`product_context.py` reads only the *selected* image documents' descriptions. It is a separate
path from `build_product_context_block`, and the image exclusion in `_is_injectable` stays exactly
as it was: text documents already reach prompts there, and a document arriving through both would
spend the budget twice to say the same thing in two voices. Order follows the caller, not
`created_at`, because the order is a ranking the user expressed.

**Consumer.** `{visual_brief_section}` sits immediately after `{brand_section}` in
`PROTOTYPE_HTML_USER_TEMPLATE` — beside `BRAND` because it is the same *kind* of instruction,
while the sections below say what the thing does rather than what it looks like. The section
quotes the descriptions and *then* instructs, so the directive is the last thing read on the
topic: take every `:root` property from the palette they describe, take the layout mode (phone
shell vs full-width top nav) from what they depict, match their radii and type weight, and where
two visuals disagree the **earlier one wins**.

**Boundary.** `selected_product_doc_ids` is validated before any job exists, on the same keyed-read
trust boundary as the research ids (`sk=PRODUCT_DOC#{id}`, which makes ownership and type one
check). Arity is bounded *before* the first read, so a caller cannot turn one request into 500
keyed reads.

## Four decisions worth the words

**1. No `use_visuals` flag — a non-empty list *is* the request.** A flag beside a list admits
"flag on, empty list" and "flag off, ids present": two states with no meaning that every layer
would then need a convention for. Research carries such a flag for historical reasons; copying it
here would have been copying the wrong half.

**2. The bound is 4, and not for budget.** The prompt drives *one* set of eight `:root` custom
properties, and every selected visual contributes a concrete palette for those same eight slots —
so a longer list is contradictory instruction rather than more grounding. Pinned to the frontend
copy by a new source-text lockstep, deliberately a second test rather than an assertion inside the
research one: the two numbers must be free to differ from each other while each stays pinned
across the language boundary.

**3. Visuals are a derivation field, not a `sources` role.** `visual_document_ids` follows
`persona_ids` — a plain id list for a non-document input. A `sources` entry *promises* a lookup:
the read-side resolver finds each id in the project's `ProjectDocument` list to render a title and
a type badge. A visual is a product doc under a different sort key with a `token_hex(8)` id, so it
would always come back unresolved and render as a bare hex string in the one panel whose entire
job is explaining provenance. The panel shows a **count** instead.

**4. Recorded ids come from the producer, never the request.** A build that selected three visuals
of which one was unreadable records two. Same invariant `product_context_included` already holds
in this function, for the same reason: what reached the prompt and what the document claims must
not be able to disagree.

## A hole found while building revision inheritance

A revision re-reads the base prototype's derivation, so without inheritance a visually-grounded
prototype would come back in the default theme — the largest possible unrequested change from a
button that promised only to act on the feedback.

But because visual ids have no gating flag, the API validates them whenever sent and 404s an id it
cannot resolve. So **one mockup deleted from the Product tab after the build would have made that
prototype unrevisable, permanently, from the only button that revises it.** Strictly worse than
the research case this was modelled on, and with no analogue to the research fix, because visuals
never appear in `ProjectDocument[]`.

Fixed by threading the product-doc list into the documents tab and filtering inherited ids against
it — with `undefined` (list not yet known) deliberately **not** treated as empty, or one failed
side request would silently strip a revision of its entire grounding.

## Verification

Gates, measured as a delta against the base commit rather than against green:

| Gate | Base | This branch |
|---|---|---|
| backend pytest | 1986 | 2060 |
| frontend vitest | 2505 | 2536 |
| `test:cdk` | 239 | 239 |
| `test:stream` | 303 | 303 |
| eslint (frontend + stream) | clean | clean |
| `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553, per-rule histogram unchanged |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |
| `cdk synth` | 5 templates, 0 warnings | 5 templates, 0 warnings |

**One acceptance criterion is structural rather than asserted.** A build that selects no visuals
renders the section as `''`, so the pre-existing byte-identical golden-prompt test covers it
unchanged and `golden/prototype_prompt_baseline.txt` is **untouched** — which is itself the
evidence the section is properly optional. Wanting to regenerate that golden would have meant a
bug in the change, not in the test.

**The fixture is the argument in two places, and one document cannot make either.**
Selection-versus-"everything ready is included" needs a selected visual *and* an unselected ready
one — with a single document the two behaviours are byte-identical, and the second is what happens
without this change. Order-versus-`created_at` needs two visuals requested in the reverse of their
stored order. Both fixtures are used at the producer and again at the prompt.

**Mutation-probed at every layer**: 14 on the producer, 9 on the API boundary, 4 on the prompt, 14
on the frontend, 1 on the derivation, 2 on the API-to-job seam. All caught.

That last pair is worth naming: each side's tests assert only their own side of the
`selected_product_doc_ids` string, so a one-sided rename could have left both suites green with the
feature silently dead end to end. Renaming the key in one file at a time proves it cannot — though
the guard that catches it is incidental (both sides' fixtures hardcode the literal) rather than
designed.

**Two mutations survived at first and were fixed rather than explained away.** A producer test
whose fixture was missing two things at once, so a second guard carried it and the `status` check
could be deleted unnoticed. And a frontend bound guard masked by the `disabled` attribute — killed
only by dispatching a change event directly at the disabled box. The pre-existing research
equivalent has the identical masked gap, found by the same probe and left alone as out of scope.

**A regression I introduced, caught by delta.** `lint:python` moved 553 → 554 on a bare
`except Exception` in the new wrapper. Suppressed with a stated reason rather than narrowed,
because the breadth is the point: an optional section must never cost a build. The first attempt at
that comment began `# noqa is deliberate…`, which ruff read as a *blanket* directive and reported as
an unused one, so the count stayed at 554 while the real suppression worked. The prose is reworded.
On a gate already carrying 553 findings, a one-error regression is invisible unless the count and
the per-rule histogram are both diffed.

## Not done, and both need a deployed stack

- **Does the generated prototype's `--primary` actually match the mockup's palette?** This is the
  only criterion that tests the outcome rather than the plumbing, and its falsifier is the
  documented neutral indigo `#4F46E5` appearing instead. It needs a real model call, so it is a
  scripted manual check by design.
- **Playwright on the picker.** It has UI, so it is not verified until a browser has rendered it.

Happy to hold this open until both are run if you would rather see them first.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "Referenz"
       },
       "selectedUsed": "{{used}} von {{selected}} ausgewählten Dokumenten verwendet",
-      "unavailable": "Nicht mehr verfügbar"
+      "unavailable": "Nicht mehr verfügbar",
+      "visualsUsed_one": "{{count}} Mockup verwendet",
+      "visualsUsed_other": "{{count}} Mockups verwendet"
     },
     "editDocument": "Dokument bearbeiten",
     "newDocument": "Neues Dokument",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "Gestartet — Verfolgen Sie den Fortschritt unter Hintergrundaufgaben.",
       "useProductContext": "Produkt-/Servicebeschreibung",
-      "useResearch": "Rechercheberichte ({{total}})"
+      "useResearch": "Rechercheberichte ({{total}})",
+      "visuals": "Visuelle Mockups ({{total}})",
+      "visualsLimit": "Höchstens {{max}} Mockups pro Build",
+      "visualsNotReady": "{{total}} werden noch verarbeitet"
     },
     "revision": {
       "feedback": "Rückmeldung: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "Produkt-/Servicebeschreibung",
       "useResearch": "Rechercheberichte ({{total}})",
       "visuals": "Visuelle Mockups ({{total}})",
+      "visualsFailed": "{{total}} fehlgeschlagen — erneut hochladen, um sie zu nutzen",
       "visualsLimit": "Höchstens {{max}} Mockups pro Build",
       "visualsNotReady": "{{total}} werden noch verarbeitet"
     },
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "Reference"
       },
       "selectedUsed": "{{used}} of {{selected}} selected documents used",
-      "unavailable": "No longer available"
+      "unavailable": "No longer available",
+      "visualsUsed_one": "{{count}} visual used",
+      "visualsUsed_other": "{{count}} visuals used"
     },
     "editDocument": "Edit document",
     "newDocument": "New Document",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "Started — track it in Background Jobs.",
       "useProductContext": "Product / service description",
-      "useResearch": "Research reports ({{total}})"
+      "useResearch": "Research reports ({{total}})",
+      "visuals": "Visual mockups ({{total}})",
+      "visualsLimit": "At most {{max}} visuals per build",
+      "visualsNotReady": "{{total}} still being processed"
     },
     "revision": {
       "feedback": "Feedback: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "Product / service description",
       "useResearch": "Research reports ({{total}})",
       "visuals": "Visual mockups ({{total}})",
+      "visualsFailed": "{{total}} failed to process — upload again to use",
       "visualsLimit": "At most {{max}} visuals per build",
       "visualsNotReady": "{{total}} still being processed"
     },
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "Descripción del producto o servicio",
       "useResearch": "Informes de investigación ({{total}})",
       "visuals": "Maquetas visuales ({{total}})",
+      "visualsFailed": "{{total}} con error — vuelve a subirlas para usarlas",
       "visualsLimit": "Máximo {{max}} maquetas por compilación",
       "visualsNotReady": "{{total}} en procesamiento"
     },
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "Referencia"
       },
       "selectedUsed": "{{used}} de {{selected}} documentos seleccionados utilizados",
-      "unavailable": "Ya no está disponible"
+      "unavailable": "Ya no está disponible",
+      "visualsUsed_one": "{{count}} maqueta utilizada",
+      "visualsUsed_other": "{{count}} maquetas utilizadas"
     },
     "editDocument": "Editar documento",
     "newDocument": "Nuevo Documento",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "Iniciado: síguelo en Tareas en segundo plano.",
       "useProductContext": "Descripción del producto o servicio",
-      "useResearch": "Informes de investigación ({{total}})"
+      "useResearch": "Informes de investigación ({{total}})",
+      "visuals": "Maquetas visuales ({{total}})",
+      "visualsLimit": "Máximo {{max}} maquetas por compilación",
+      "visualsNotReady": "{{total}} en procesamiento"
     },
     "revision": {
       "feedback": "Comentarios: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "Référence"
       },
       "selectedUsed": "{{used}} document(s) utilisé(s) sur {{selected}} sélectionné(s)",
-      "unavailable": "Plus disponible"
+      "unavailable": "Plus disponible",
+      "visualsUsed_one": "{{count}} maquette utilisée",
+      "visualsUsed_other": "{{count}} maquettes utilisées"
     },
     "editDocument": "Modifier le document",
     "newDocument": "Nouveau document",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "Lancé — suivez la progression dans Tâches en arrière-plan.",
       "useProductContext": "Description du produit ou service",
-      "useResearch": "Rapports de recherche ({{total}})"
+      "useResearch": "Rapports de recherche ({{total}})",
+      "visuals": "Maquettes visuelles ({{total}})",
+      "visualsLimit": "Au maximum {{max}} maquettes par génération",
+      "visualsNotReady": "{{total}} en cours de traitement"
     },
     "revision": {
       "feedback": "Retour : {{feedback}}",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "Description du produit ou service",
       "useResearch": "Rapports de recherche ({{total}})",
       "visuals": "Maquettes visuelles ({{total}})",
+      "visualsFailed": "{{total}} en échec — importez-les à nouveau pour les utiliser",
       "visualsLimit": "Au maximum {{max}} maquettes par génération",
       "visualsNotReady": "{{total}} en cours de traitement"
     },
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "製品・サービスの説明",
       "useResearch": "リサーチレポート ({{total}})",
       "visuals": "ビジュアルモックアップ ({{total}})",
+      "visualsFailed": "{{total}} 件は処理に失敗しました — 再アップロードしてください",
       "visualsLimit": "1 回のビルドで最大 {{max}} 件のモックアップ",
       "visualsNotReady": "{{total}} 件は処理中です"
     },
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "参照"
       },
       "selectedUsed": "選択された{{selected}}件のうち{{used}}件を使用",
-      "unavailable": "現在は利用できません"
+      "unavailable": "現在は利用できません",
+      "visualsUsed_one": "ビジュアル{{count}}件を使用",
+      "visualsUsed_other": "ビジュアル{{count}}件を使用"
     },
     "editDocument": "ドキュメントを編集",
     "newDocument": "新規ドキュメント",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "開始しました — バックグラウンドジョブで進行状況を確認できます。",
       "useProductContext": "製品・サービスの説明",
-      "useResearch": "リサーチレポート ({{total}})"
+      "useResearch": "リサーチレポート ({{total}})",
+      "visuals": "ビジュアルモックアップ ({{total}})",
+      "visualsLimit": "1 回のビルドで最大 {{max}} 件のモックアップ",
+      "visualsNotReady": "{{total}} 件は処理中です"
     },
     "revision": {
       "feedback": "フィードバック: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "제품/서비스 설명",
       "useResearch": "리서치 보고서 ({{total}}개)",
       "visuals": "비주얼 목업 ({{total}}개)",
+      "visualsFailed": "{{total}}개는 처리에 실패했습니다 — 다시 업로드하세요",
       "visualsLimit": "한 번의 빌드에 최대 {{max}}개 목업",
       "visualsNotReady": "{{total}}개는 처리 중입니다"
     },
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "참조"
       },
       "selectedUsed": "선택한 {{selected}}개 중 {{used}}개 사용",
-      "unavailable": "더 이상 사용할 수 없음"
+      "unavailable": "더 이상 사용할 수 없음",
+      "visualsUsed_one": "비주얼 {{count}}개 사용",
+      "visualsUsed_other": "비주얼 {{count}}개 사용"
     },
     "editDocument": "문서 편집",
     "newDocument": "새 문서",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "시작했습니다 — 백그라운드 작업에서 진행 상황을 확인하세요.",
       "useProductContext": "제품/서비스 설명",
-      "useResearch": "리서치 보고서 ({{total}}개)"
+      "useResearch": "리서치 보고서 ({{total}}개)",
+      "visuals": "비주얼 목업 ({{total}}개)",
+      "visualsLimit": "한 번의 빌드에 최대 {{max}}개 목업",
+      "visualsNotReady": "{{total}}개는 처리 중입니다"
     },
     "revision": {
       "feedback": "피드백: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/pt/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/pt/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "Referência"
       },
       "selectedUsed": "{{used}} de {{selected}} documentos selecionados utilizados",
-      "unavailable": "Não está mais disponível"
+      "unavailable": "Não está mais disponível",
+      "visualsUsed_one": "{{count}} maquete utilizada",
+      "visualsUsed_other": "{{count}} maquetes utilizadas"
     },
     "editDocument": "Editar documento",
     "newDocument": "Novo Documento",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "Iniciado — acompanhe em Tarefas em Segundo Plano.",
       "useProductContext": "Descrição do produto ou serviço",
-      "useResearch": "Relatórios de pesquisa ({{total}})"
+      "useResearch": "Relatórios de pesquisa ({{total}})",
+      "visuals": "Maquetes visuais ({{total}})",
+      "visualsLimit": "No máximo {{max}} maquetes por build",
+      "visualsNotReady": "{{total}} ainda em processamento"
     },
     "revision": {
       "feedback": "Comentários: {{feedback}}",
--- a/voc-datalake/frontend/public/locales/pt/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/pt/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "Descrição do produto ou serviço",
       "useResearch": "Relatórios de pesquisa ({{total}})",
       "visuals": "Maquetes visuais ({{total}})",
+      "visualsFailed": "{{total}} com falha — envie novamente para usar",
       "visualsLimit": "No máximo {{max}} maquetes por build",
       "visualsNotReady": "{{total}} ainda em processamento"
     },
--- a/voc-datalake/frontend/public/locales/zh/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/zh/projectDetail.json
@@ -111,6 +111,7 @@
       "useProductContext": "产品/服务说明",
       "useResearch": "研究报告（{{total}}）",
       "visuals": "视觉稿（{{total}}）",
+      "visualsFailed": "{{total}} 份处理失败 — 请重新上传后使用",
       "visualsLimit": "每次构建最多 {{max}} 份视觉稿",
       "visualsNotReady": "{{total}} 份仍在处理中"
     },
--- a/voc-datalake/frontend/public/locales/zh/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/zh/projectDetail.json
@@ -72,7 +72,9 @@
         "reference": "参考文档"
       },
       "selectedUsed": "已选择 {{selected}} 个文档，使用 {{used}} 个",
-      "unavailable": "已不可用"
+      "unavailable": "已不可用",
+      "visualsUsed_one": "已使用 {{count}} 份视觉稿",
+      "visualsUsed_other": "已使用 {{count}} 份视觉稿"
     },
     "editDocument": "编辑文档",
     "newDocument": "新建文档",
@@ -107,7 +109,10 @@
       "sourcePrfaq": "PR/FAQ",
       "started": "已开始 — 可在后台任务中查看进度。",
       "useProductContext": "产品/服务说明",
-      "useResearch": "研究报告（{{total}}）"
+      "useResearch": "研究报告（{{total}}）",
+      "visuals": "视觉稿（{{total}}）",
+      "visualsLimit": "每次构建最多 {{max}} 份视觉稿",
+      "visualsNotReady": "{{total}} 份仍在处理中"
     },
     "revision": {
       "feedback": "反馈：{{feedback}}",
--- a/voc-datalake/frontend/src/api/derivation.test.ts
+++ b/voc-datalake/frontend/src/api/derivation.test.ts
@@ -212,6 +212,53 @@ describe('a document grounded in uploaded visuals', () => {
     expect(resolved.origin).toBe('legacy')
     expect(resolved.visual_document_ids).toEqual([])
   })
+
+  /**
+   * THE ONLY CASE THAT EXISTS IN PRODUCTION DATA, and the one the other tests here
+   * do not cover: every prototype already in DynamoDB has a `derivation` written
+   * before this field existed, so the key is ABSENT rather than present-and-junk.
+   *
+   * The risk it guards is total, not partial. `DocumentDerivationSchema` now
+   * declares `visual_document_ids` and the object carries `.catch(() =>
+   * emptyDerivation())`, so a schema that did not tolerate `undefined` would make
+   * EVERY existing prototype parse to the empty record, read as `isEmpty`, fall
+   * through to the legacy reconstruction, and lose the sources it does record —
+   * every document losing its provenance the day the field shipped, with nothing
+   * failing anywhere.
+   *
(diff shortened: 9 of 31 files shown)


perrozzi:

This PR spends the already-extracted image descriptions: a build can now name up to four uploaded mockups (`selected_product_doc_ids`), whose descriptions are assembled by a new `build_visual_brief_block` producer, injected as a `VISUAL BRIEF` section beside `BRAND` in the prototype prompt, recorded as `visual_document_ids` in the derivation, surfaced as a count in the provenance panel, and inherited by revisions (filtered against the live product-doc list). The change is genuinely additive, the trust boundary is done the right way (keyed read on `PROJECT#{id}` / `PRODUCT_DOC#{id}`, arity bound before the first read, no job created on rejection), untrusted description text is fenced and marker-sanitised, the constant is pinned across the language boundary, all eight locales are updated, and test coverage is unusually thorough. I found no blockers — two behavioural inconsistencies worth fixing and one backward-compatibility gap worth a test.

### Should fix

- **The arity bound (4) and the character budget (9 000) disagree, so the fourth selected visual can be silently dropped.** `MAX_VISUAL_BRIEF_DOC_CHARS = 3_000` × 4 = 12 000 > `MAX_VISUAL_BRIEF_TOTAL_CHARS = 9_000` (`lambda/api/product_context.py`). Because the refusal is all-or-nothing and `continue`s, a user who ticks four long-description mockups gets three in the prompt with nothing said anywhere in the UI — the picker offered four and the limit note says four. The derivation stays honest (good), but the picker's promise does not. Either raise the total to `MAX_SELECTED_PRODUCT_DOC_IDS * MAX_VISUAL_BRIEF_DOC_CHARS`, or lower the arity bound, or surface the drop. Relatedly, the rationale comment above `MAX_VISUAL_BRIEF_TOTAL_CHARS` ("MAX_DOCS_PER_PROJECT is 20, so a caller ticking every image could spend 60,000 characters") is stale as of this PR — the API now caps the list at 4, so the stated threat model no longer exists and the number reads as unjustified.

- **`visualLimitReached` and the guard inside `onToggleVisualId` measure different lists, so a checkbox can look enabled and do nothing.** In `usePrototypeBuild.ts`, `visualLimitReached` is derived from `selectedVisualIds` (filtered against live `visualOptions`) while the toggle refuses on `current.length >= MAX_SELECTED_PRODUCT_DOC_IDS` against the unfiltered `chosenVisualIds`. Tick four, have one deleted from the Product tab: `selectedVisualIds.length === 3` so the remaining boxes render enabled, but every click is silently swallowed. Guard on the filtered length (or prune `chosenVisualIds` when options change) so the two agree.

- **`visualsNotReady` reports failed extractions as "still being processed".** `overviewState.ts` counts `isVisual(d) && d.status !== 'ready'`, and the string rendered is `documents.prototype.visualsNotReady` = "{{total}} still being processed". A `failed` image will say that forever, sending the user back to wait on something that will never appear. Either exclude `failed` from the count or give it its own key (the fixture in `OverviewTab.prototypeExtraSources.test.tsx` only exercises `extracting`, which is why this is not caught).

- **No regression test for a `derivation` map written before this field existed.** `DocumentDerivationSchema` in `derivation.ts` now requires `visual_document_ids`, and the object carries `.catch(() => emptyDerivation())` — so if `idListSchema` does not tolerate `undefined`, every prototype already in DynamoDB parses to the empty record, is judged `isEmpty`, and falls through to the legacy reconstruction, quietly losing its recorded sources. The new tests cover a *present-but-junk* value (`null`, `7`, `{0:'vis_1'}`) but not an **absent key**, which is the only case that exists in production data today. Please add that case explicitly in `derivation.test.ts` (declared derivation with `sources` and no `visual_document_ids` → `origin: 'declared'`, sources intact, `visual_document_ids: []`).

### Nits

- `build_visual_brief_block(project_id: str, doc_ids)` in `lambda/api/product_context.py` is the only untyped parameter in the new code; `doc_ids: Iterable[str] | None` would match the module's style and the docstring's promise.
- An inherited visual whose extraction has since `failed` still passes the `productDocs.some(...)` presence filter in `DocumentsTab.tsx`, so a revision silently loses that visual's grounding rather than dropping it visibly; worth a comment or a readiness check now that the "generator just skips it" path is user-visible.
- `PrototypeVisualSources` renders the group heading as a bare `<p>`; a `fieldset`/`legend` (or `role="group"` + `aria-labelledby`) would give the checkbox list the same accessible grouping the research sub-list gets from its master box.
- Inconsistent interpolation vocabulary for the same concept: `documents.prototype.visuals` uses `{{total}}` while `documents.derivation.visualsUsed` uses i18next `count` pluralisation. Documented, but a reader of the eight catalogues has to notice it.
- Comment density in the new code is very high and several blocks restate the PR description verbatim (e.g. the ~20-line note above `visual_brief_section` in `document_generator/handler.py`). The rationale is valuable; consider moving the longest arguments to the module docstring so the control flow stays readable.

---

perrozzi:
## Round 1 disposition

All four should-fix items were real and are fixed. Three of five nits taken, one answered, one documented rather than half-done. Pushing `604b2e6`.

**Two of the four were defects I created** by letting two independent slices choose numbers and predicates with no shared owner — worth naming, because that is the failure mode of the way this was built, not of any one decision.

### Should fix

**1. The arity bound and the character budget disagreed.** The sharpest finding. 4 ids × 3000 chars against a hardcoded 9000, and because the refusal is all-or-nothing the **fourth visual was dropped whole** — while the picker had offered four and its limit note said four. The derivation stayed honest, since it records what was used; the UI's promise did not, and nothing said so.

Fixed structurally rather than by picking a bigger number: the total is now **derived** as `MAX_SELECTED_PRODUCT_DOC_IDS * MAX_VISUAL_BRIEF_DOC_CHARS`, so the budget cannot refuse a selection the boundary accepted, and raising the bound later cannot reintroduce the gap. To make that possible the bound **moved** from `projects_handler.py` into `product_context.py`, beside the budget it bounds, and the handler imports it — it already imports `DOC_SK_PREFIX` from there for the same one-owner-per-fact reason. Lockstep retargeted to the new home.

You were also right that the rationale comment had gone stale in this same PR: it justified 9000 against "`MAX_DOCS_PER_PROJECT` is 20, so a caller could tick every image", a threat model this PR removed when it capped the list at 4. Gone.

**2. `visualLimitReached` and the toggle guard measured different lists.** Correct. The flag counted the filtered selection, the guard counted the raw stored list. Tick four, have one deleted from the Product tab: the flag said three so the remaining boxes rendered **enabled**, and the guard still counted four and swallowed every click. A control that looks available and does nothing is worse than a disabled one.

Both toggles had it, because both were the same code. Fixed once in a shared `toggleWithinBound` rather than twice — same bug for the same reason, and one guard means a fix cannot land on one and miss its sibling. Removal is deliberately never bounded, which keeps an over-long inherited selection recoverable if the bound is ever lowered.

**3. A failed extraction read as "still being processed".** Correct, and it would have said so forever. Split into two independent counts with two keys: in-flight stays grey and asks for patience, failed is amber and asks for an action. Deliberately *not* solved by hiding the failed ones — the note exists because a picker showing two of the three files someone just uploaded, unexplained, reads as a bug. The in-flight predicate is "not ready **and** not failed" rather than the two status names literally, so the two counts are complements over the unselectable images and a status this client does not recognise still gets reported rather than vanishing from both.

**4. No test for a derivation written before the field existed.** The highest-risk of the four, and I checked the behaviour before writing the test: `idListSchema` already tolerates `undefined`, so **there is no production break**. But the risk it guards is total rather than partial — had it not tolerated it, every prototype in DynamoDB would have parsed to the empty record, read as `isEmpty`, fallen through to the legacy reconstruction and lost the sources it *does* record, silently, on the day this shipped. Now pinned, with `sources` asserted intact, because the empty-list assertion alone passes against exactly the degradation being ruled out.

### Nits

- **Untyped parameter** — `Iterable[str] | None`.
- **Comment density** — the ~20-line argument above `visual_brief_section` moved to the module docstring, leaving two lines and a pointer. Third time this has been raised, fair each time.
- **Accessible grouping** — `role="group"` + `aria-labelledby` on the checkbox list, chosen over `fieldset`/`legend` because it adds no element, so the rail and every text size are provably unchanged and the heading keeps carrying the count as the accessible name.
- **`{{total}}` vs `count` — answered, no change.** The split is deliberate: parenthetical numbers avoid eight plural-suffix pairs, pluralised sentences need them, and the neighbouring `useResearch` / `feedbackUsed` keys already draw the line the same way.
- **Inherited visual whose extraction has since failed — documented, not fixed.** It passes the presence filter, so the revision sends it, the API accepts it, and the generator skips it: less grounding, quietly. Filtering on `ready` would not fix that, only move the silence one layer earlier, and would newly drop a visual that is merely mid-re-extraction. It belongs with the `feedbackRebased` note that already reports a dropped inherited *source*, and is left for that follow-up rather than half-done here.

### Verification

**All four fixes mutation-probed by reverting each to the exact shape you flagged: 4 caught, 0 survived.**

Two of them needed **new tests to be load-bearing at all** — the derived budget and the filtered bound — which is the same finding one layer up: a real fix that nothing tests is the same defect. The budget is now pinned twice, as a property of the two constants *and* end to end proving the producer admits a full selection of maximal visuals.

One probe anchor rotted (the failed/extracting split was implemented differently from my guess). It was reported `SKIPPED` with exit 3 rather than counted as held, then retargeted at the real predicate and caught. Worth mentioning because a mutation that applies to nothing otherwise reads exactly like a guard that worked.

| Gate | Round 0 | Round 1 |
|---|---|---|
| backend pytest | 2060 | 2062 |
| frontend vitest | 2536 | 2546 |
| `test:cdk` / `test:stream` | 239 / 303 | 239 / 303 |
| eslint (frontend + stream) | clean | clean |
| `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553 |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |
| `cdk synth` | 5 templates, 0 warnings | 5 templates, 0 warnings |

One self-inflicted regression along the way: my `toggleWithinBound` refactor put arrows five levels deep and tripped `sonarjs/no-nested-functions`, turning eslint red. Resolved with module-level `isOfferedResearch` / `isOfferedVisual` predicates used at all four sites that ask "is this id still on offer", which also removes the same question spelled four ways.

Still outstanding and unchanged by this round: the `--primary`-matches-the-mockup check needs a real model call, and the picker needs Playwright. Both need a deployed stack.

perrozzi:

Since my last review the author addressed all four Should-fix items and three of the five nits: `MAX_VISUAL_BRIEF_TOTAL_CHARS` is now derived from `MAX_SELECTED_PRODUCT_DOC_IDS` (which moved into `product_context.py`, with the lockstep test repointed), the bound guard and `visualLimitReached` now measure the same filtered list via a shared `toggleWithinBound`, `visualsNotReady` split into `visualsExtracting`/`visualsFailed` with a new key in all eight catalogues, and `derivation.test.ts` gained the absent-key regression case. Note this is a **partial** review: the full-context diff was truncated, so I only verified the incremental changes against the parts of the PR I could see.

### Resolved
- Arity bound vs. character budget: `MAX_VISUAL_BRIEF_TOTAL_CHARS = MAX_SELECTED_PRODUCT_DOC_IDS * MAX_VISUAL_BRIEF_DOC_CHARS`, plus a property test and an end-to-end test at the bound (`test_visual_brief.py::TestTheBudgetCannotRefuseWhatTheBoundaryAccepted`); the stale rationale comment is rewritten.
- `visualLimitReached` / toggle-guard disagreement fixed by `toggleWithinBound` counting only ids still on offer, with a test that ticks a replacement after a deletion and asserts it is *sent*.
- `failed` extractions now get their own count and their own amber `documents.prototype.visualsFailed` line, translated in all eight locales, with a discriminating fixture and a non-English render test.
- Absent-`visual_document_ids` regression test added (declared origin, `sources` intact).
- Nits: `doc_ids: Iterable[str] | None`, `role="group"`/`aria-labelledby` on the visual group, the long prompt-placement rationale moved to the `document_generator/handler.py` module docstring, and the presence-vs-readiness filter in `DocumentsTab.tsx` now states the residual gap.

### New findings

- **Should fix — the stored selection can now exceed the bound, and the send is not capped.** `toggleWithinBound` (`usePrototypeBuild.ts`) counts only offered ids but never prunes unoffered ones from state, so the raw list can grow past `MAX_SELECTED_PRODUCT_DOC_IDS`. That is reachable without a deletion: a visual can leave `visualOptions` by re-entering `extracting` (the case `DocumentsTab.tsx`'s new comment explicitly calls out as "resolves on its own") and come back `ready` later. Tick four, one re-extracts, tick a fifth, re-extraction finishes → `selectedVisualIds` is five and `selected_product_doc_ids` is sent with five, which `_validated_product_doc_ids` rejects with a 400 naming a length the user never chose — the exact failure the guard exists to prevent. Cheapest fix: `.slice(0, MAX_SELECTED_PRODUCT_DOC_IDS)` on `selectedVisualIds`/`selectedResearchIds` (same applies to research at 10), or prune stale ids when the options list changes. Worth a test that re-adds a previously-unoffered option.

- **Nit — the `toggleWithinBound` doc comment is orphaned.** In `usePrototypeBuild.ts` the long JSDoc for `toggleWithinBound` now sits directly above the `isOfferedResearch` JSDoc, so both block comments precede `isOfferedResearch` and the function they document has none. Move it back onto `toggleWithinBound`.

- **Nit — import order in `lambda/api/product_context.py`:** `from collections.abc import Iterable` was inserted after `datetime`/`decimal`; alphabetically it belongs first. Harmless today (isort rules evidently aren't enforced, hence the unchanged histogram) but inconsistent with the file's existing grouping.

- **Nit —** stray blank line before the closing `})` of the `'a document grounded in uploaded visuals'` describe in `derivation.test.ts`.

---

perrozzi:
## Round 2 disposition

One should-fix, and it is a consequence of round 1's fix — correct, reachable, and now closed at the right layer. All three nits taken. Pushing `1eaac6b`.

### Should fix — the stored selection could exceed the bound

Right, and the part that makes it matter is the reachability argument: **it needs no deletion at all.** `toggleWithinBound` counts only ids still *on offer*, so an id that leaves `visualOptions` stops being counted while it stays in state — and a visual can leave by re-entering `extracting`, which is the very case my own `DocumentsTab` comment described as "resolves on its own". Tick four, one re-extracts, tick a fifth, extraction finishes: the stored list is five, and unsliced the request carries five, which the API rejects with a 400 naming a length the user never chose. Exactly what the bound exists to prevent, arriving through the guard meant to enforce it.

Fixed by slicing the derived lists rather than pruning state, for both visuals and research (the same defect at 10):

- **Slicing enforces the invariant where it matters** — at the request — instead of depending on every path into state staying correct. That is the difference between a guard and a hope.
- **Pruning on options-change would need an effect, and this repo makes that an error** (`react-hooks/set-state-in-effect`). So the cheapest correct fix is also the only ESLint-legal one.
- **It matches the sibling that already does this:** `inheritedExtraSources` in `DocumentsTab.tsx` slices for the adjacent reason (a bound that gets *lowered* later). Both paths now agree rather than one being belt-and-braces and the other bare.
- **A slice keeps the earliest ticks**, which is the precedence order the prompt reads — so the id that gets dropped is the last one added, not an arbitrary one.

`toggleWithinBound`'s docstring now states the consequence explicitly (the stored list *can* exceed the bound, which is why the memos slice), so the next reader does not have to rediscover it.

New test covers your exact scenario — re-extract, top up, finish — and asserts both halves: the length holds **and** the survivors are the earliest ticks rather than an arbitrary subset.

### Nits — all three taken

- **Orphaned JSDoc.** Mine, from the round-1 edit: `toggleWithinBound`'s block ended up above `isOfferedResearch`'s, so two comments preceded one function and the documented one had none. Moved back, and extended with the consequence above.
- **Import order** — `collections.abc` moved to the front of the stdlib group.
- **Stray blank line** before the closing `})` of the visuals describe — removed.

### Verification

Re-probed **all six** fixes across both rounds by reverting each to the exact shape flagged: **6 caught, 0 survived.** The two new ones drop the slice on the visual send and on the research send.

| Gate | Round 1 | Round 2 |
|---|---|---|
| backend pytest | 2062 | 2062 |
| frontend vitest | 2546 | 2547 |
| `test:cdk` / `test:stream` | 239 / 303 | 239 / 303 |
| eslint (frontend + stream) | clean | clean |
| `typecheck:all` | clean | clean |
| `lint:python` | 553 | 553 |
| `i18n:check` | missing 0 / extra 12 / untranslated 19 | unchanged |

### On the truncated diff

Noted that this round was partial because the full-context diff was truncated — that has happened on every review of this feature's PRs, and it is worth saying which parts have therefore never been machine-reviewed: `lambda/api/product_context.py`'s producer body and `lambda/jobs/document_generator/handler.py`'s prompt assembly are the two largest additions and the most likely to have been cut. Both are covered by tests and mutation probes, but neither has had a second pair of eyes on the code itself. Flagging rather than assuming it is fine.

Still outstanding, unchanged: the `--primary`-matches-the-mockup check needs a real model call, and the picker needs Playwright. Both need a deployed stack.

perrozzi:

Follow-up (partial: the full-context diff was truncated, so I verified the incremental changes against only the portions of the PR I could see). Since my last review the author added `.slice(0, MAX_…)` to both `selectedResearchIds` and `selectedVisualIds` with a re-extraction regression test, moved the `toggleWithinBound` JSDoc back onto its function, fixed the `Iterable` import position, and dropped the stray blank line — all four previous findings are addressed.

### Resolved
- Over-bound stored selection is now capped at the request in `usePrototypeBuild.ts`, with `OverviewTab.prototypeExtraSources.test.tsx::never sends more than the bound after a visual re-extracts and returns` covering the no-deletion path and asserting the survivors are the earliest ticks.
- `toggleWithinBound`'s JSDoc is back on `toggleWithinBound`, with the ⚠️ note explaining why the memos slice.
- `from collections.abc import Iterable` moved to the top of the stdlib group in `lambda/api/product_context.py`.
- Stray blank line removed in `derivation.test.ts`.

### New findings

- **Nit — display state now derives from the sliced list.** In `OverviewTab.tsx`, `checked = extras.selectedVisualIds.includes(...)`, so in the exact scenario the new test constructs the fifth id the user ticked renders *unchecked* (and disabled, since `visualLimitReached` is true) once its doc returns `ready`. That is honest — it isn't sent — but it is a silent un-tick; worth a one-line comment beside the memo noting the checkbox reflects what will be sent, not what is stored, so a future reader doesn't "fix" it back to `chosenVisualIds`.

---

perrozzi:

Follow-up (partial: the full-context diff was truncated again, so this delta review covers only the incremental change plus the surrounding portions of `usePrototypeBuild.ts` I could see). The only change since my last review is a comment block added above the `selectedVisualIds` memo, which addresses my one remaining nit.

### Resolved
- The ⚠️ note in `frontend/src/pages/ProjectDetail/usePrototypeBuild.ts` now records that the tick-boxes render from the sliced list, that the over-bound id visibly un-ticks itself, and that `chosenVisualIds` must not be exposed instead — exactly the reader-protection I asked for.

No new findings; the delta is comment-only and accurate to the behaviour the adjacent test pins.

---

perrozzi:
## Round 3 disposition

No should-fix items. The one nit is taken, as `471a07c` — comment only, no behaviour change.

**The tick-box that un-ticks itself.** Good catch, and worth taking precisely because the behaviour is *correct* and the obvious future "fix" is wrong. The boxes render from `selectedVisualIds`, which is filtered and sliced, so a box shows what will be **sent** rather than what is stored — and in the re-extraction case round 2 closed, the id beyond the bound visibly un-ticks itself once its document returns `ready`.

Showing it ticked would promise a grounding the build will not have, so honest is right. But it is a silent un-tick, and the natural repair someone reaches for is rendering from `chosenVisualIds` — which is exactly the state the slice exists in order not to trust. The memo now says both halves: that the checkbox reflects the request, and not to swap the source.

Gates unchanged: frontend 2547, backend 2062, cdk 239, stream 303, eslint clean, `typecheck:all` clean, `lint:python` 553, i18n missing 0 / extra 12 / untranslated 19.

### Where this stands

Three rounds, seven findings, all real and all fixed — two of them (the budget/bound contradiction and the over-bound stored selection) were defects I introduced, and both came from the same root cause: independent slices choosing numbers and predicates without a shared owner. That is the honest lesson of building this the way I did.

Six mutation probes across the rounds, each reverting a fix to the exact shape flagged, all caught.

**Still outstanding, and unchanged by review** — both need a deployed stack, so they are the reason not to merge on green alone:

- Does the generated prototype's `--primary` actually match the mockup's palette? The only criterion that tests the outcome rather than the plumbing, with the documented neutral indigo `#4F46E5` as its falsifier. Needs a real model call.
- Playwright on the picker.

Happy to run both before this merges if you would rather see them first.

perrozzi:
## Deployed and live-verified

`471a07c` deployed to the live account (all 5 stacks, 0 warnings, 333.7s). `scripts/test-api.sh` → **34/34 passed**.

### The artifacts really carry this change

Downloaded both deployed zips and grepped the handlers, each with a positive control (a
string that must be there regardless) and a negative control (an injected sentinel, to prove
the grep can fire and that an absence is a real absence rather than a broken pipeline):

| lambda | member | asserted present |
|---|---|---|
| `voc-job-document-generator` | `handler.py` (1418 lines) | `_visual_brief`, `visual_brief_section`, `selected_product_doc_ids` |
| `voc-projects-api` | `product_context.py` (1262 lines) | `build_visual_brief_block`, `MAX_SELECTED_PRODUCT_DOC_IDS`, `MAX_VISUAL_BRIEF_TOTAL_CHARS`, `UNTRUSTED_DOC_BEGIN` |

### Does an uploaded mockup actually steer the palette? Yes — 8/8 tokens

Two **real** builds on the same live project, same PR/FAQ, `brand=None`,
`use_product_context=False`. The two requests differ in exactly one field:

- **TREATMENT** `selected_product_doc_ids=[<teal mockup>]`
- **CONTROL** `selected_product_doc_ids=[]`

The mockup was a deliberately non-indigo desktop dashboard (deep teal `#0E7C6B`, hue 171;
coral `#FF6B4A`) rendered to PNG and uploaded through the real presigned-URL path. Extraction
returned 4766 chars and reported `--primary: #0f7d6c` — the mockup's own teal, mapped onto
the generator's eight `:root` names.

| token | extracted from mockup | TREATMENT | CONTROL |
|---|---|---|---|
| `--primary` | `#0f7d6c` | ✅ `#0f7d6c` | `#b3261e` |
| `--primary-light` | `#14926f` | ✅ `#14926f` | `#d9534a` |
| `--soft` | `#d3ede4` | ✅ `#d3ede4` | `#f7dedc` |
| `--tint` | `#daece7` | ✅ `#daece7` | `#fdf4f3` |
| `--bg` | `#fdf3ea` | ✅ `#fdf3ea` | `#fbfaf9` |
| `--ink` | `#1f2d2a` | ✅ `#1f2d2a` | `#171512` |
| `--gray` | `#6b7d78` | ✅ `#6b7d78` | `#6b6560` |
| `--surface` | `#ffffff` | ✅ `#ffffff` | `#ffffff` |
| | | **8/8 match** | **1/8** (and that one is white) |

Beyond the eight: the prototype also carried the mockup's coral into `--accent: #f2603c`
(hue 12 vs the mockup's 11), took the mockup's **serif display type**, and used a
**full-width desktop top-nav** layout — where the system prompt's default is a ~420px mobile
phone shell, which is exactly what the CONTROL came back as. So the visual brief moved
palette, typography and layout, not just colour. Content in both builds still comes from the
PR/FAQ, so it is applying the mockup's design system rather than copying the mockup.

### One expectation of mine was wrong, and it is worth recording

The check was written expecting the CONTROL to fall back to the documented indigo default
`#4F46E5`. It did not — it produced a red (`#b3261e`). That is **not** caused by anything in
this PR: the system prompt offers indigo only as advice ("If no brand is specified, use
neutral defaults: an indigo primary…") and the model is free to ignore it. So "no visual →
indigo" was never a sound baseline, and the falsifier I had planned around it was weaker than
I thought.

The corrected control claim is the one that actually licenses the conclusion: a build without
the mockup does **not** land on the mockup's palette (1/8, hue 3° vs 171°). Every
treatment-side assertion is unchanged from before the data was seen. Re-runnable from the
saved artifacts, no Bedrock spend, in `assert_palette.py` — 12/12 pass, including a negative
control that mutates `--primary` and confirms the comparison drops to 7/8.

### Picker verified in the browser (Playwright, live site)

Seeded the project with the states the picker has to tell apart — 5 ready images, 1 corrupt
image, 1 text document — then drove the real UI:

- The list renders **on the card**, `role="group"` on the existing div with `aria-labelledby`
  → accessible name **"Visual mockups (5)"**; `getByRole('group', { name: /Visual mockups/i })`
  resolves to exactly 1 element, and it is a `DIV`, not a `FIELDSET`, so the `ml-5 … border-l pl-2`
  rail and text sizes are unchanged.
- **Only ready images are offered**: 5 checkboxes. The `text/plain` document never appears,
  and the corrupt image is not offered — it is counted in the failed line instead.
- **The bound bites in the UI**: ticking 1→4 works; at 4 the remaining unticked box becomes
  `disabled`, and clicking it leaves the count at 4. Amber `At most 4 visuals per build`
  (`text-amber-700`) appears only at the bound.
- **Failed vs extracting are genuinely two independent lines**, and I caught both on screen at
  once by uploading while another had already failed:
  `1 still being processed` (`text-gray-500`, asks for patience) above
  `1 failed to process — upload again to use` (`text-amber-700`, asks for an action).
- **Provenance**: the TREATMENT prototype's "Built from" panel shows the PR/FAQ plus
  **`1 visual used`** — a count, not a bare hex id, which was the reason for recording
  `visual_document_ids` as a derivation field rather than a `sources` role. The CONTROL
  prototype shows no visual line at all.

Test noise was cleaned up afterwards; the teal mockup and the two prototypes are left in
place as the evidence for this comment.

perrozzi:
## Follow-up: the one gap I named above is now closed

My earlier comment ended by admitting that both builds were fired through the API, so
"tick a visual → press **Build Prototype** → get a grounded prototype" had never been
exercised as one continuous click-through. That is now done, in the browser, against the
same deployed build.

**What the button actually sends.** Ticked the teal mockup only, wrapped `fetch`, pressed the
button, and captured the request:

```json
{ "selected_product_doc_ids": ["a074b2f65fd193f3"],
  "use_product_context": false, "use_research": false, "selected_research_ids": [],
  "source_prfaq_id": "prfaq_20260807195615", "response_language": "en" }
```

Exactly the ticked id and nothing else. The resulting build came back `--primary #0f7d6c`
(hue 171), **8/8** of the mockup's extracted palette tokens, `#4F46E5` absent, and provenance
reading `1 visual used`. So criterion 2 holds through the UI path, not only the API path.

### Two things worth knowing for anyone testing this flow

🪤 **`Build Prototype` opens a confirmation dialog when the project already has a prototype**
("This project already has a prototype. Building again adds a new one and keeps the existing
one. Continue?" → Cancel / **Build anyway**). No request is sent until *Build anyway*, so the
first press reads as a silent no-op if you are not expecting it. **The selection survives the
dialog** — I checked that before confirming, since it is exactly where a selection could
plausibly be dropped.

🔑 **The bound check is stronger than last time, deliberately.** A normal click on the fifth box
is refused (the `disabled` attribute). But a greyed-out box can *mask* a missing state guard
behind it — the trap round 2 of this PR's own review already hit once. So I also clicked with
`force: true` and dispatched a raw `click` event to get past the attribute: the count stayed at
4 and the box stayed unticked. The guard behind the attribute holds, not just the greying-out.

### Correction to my own reporting

Mid-run I read the card back with `textContent`, saw no ticked box, paired that with the absent
job, and briefly reported the button as doing nothing. Both readings were mine, not the app's:
`checked` is a DOM property and is invisible to a text read, and the job was absent because the
dialog had not been confirmed. No defect. Recording it because the failure mode is worth
avoiding — *instrumented and saw nothing* and *never happened* look identical and point to
opposite conclusions.

Remaining genuinely unexercised via UI: uploads still go through the presigned-URL path rather
than the upload control, which belongs to rung 2 rather than this PR.

PR #330, opened 2026-08-14
fix(api): stop logging raw request payloads

### LINKED ISSUE #245: [Security]: projects_handler logs the full request event, including the caller's JWT
## Summary

`projects_handler` logs the entire API Gateway event to CloudWatch. The event carries the caller's
`Authorization` header, so bearer tokens are written to logs in plaintext.

## Severity

High, and the cheapest fix in this batch.

## Risk

`lambda/api/projects_handler.py:633`:

```python
logger.info(f"Received event: {json.dumps(event)}")
```

The serialized event includes request headers, so every invocation writes a usable Cognito token to the
log group. Anyone with CloudWatch Logs read access — a broader group than the application's own admins —
can read them, and they remain valid until expiry. Request bodies are logged with the same statement,
which for this handler includes project content.

## Proposed mitigation

1. Remove the statement. Powertools' `@logger.inject_lambda_context` already provides request context
   without the raw event.
2. If the full event is genuinely needed for debugging, gate it behind a debug log level **and** redact
   `Authorization`, `Cookie` and body fields.
3. Audit for the same pattern elsewhere: `grep -rn 'json.dumps(event)' lambda/`.
4. Note that existing log groups already contain tokens. Retention on the affected group should be
   considered — expired tokens are harmless, unexpired ones are not.

## Acceptance criteria

- [ ] No handler logs the raw event.
- [ ] A repo-wide grep for `json.dumps(event)` is clean.
- [ ] If any diagnostic event logging remains, it redacts credential headers.

## Affected files

`lambda/api/projects_handler.py`

shin4141:
## Summary

- stop the projects Lambda handler from logging the complete API Gateway event
- stop async persona generation from logging the complete job event and filter payload
- add regressions proving request credentials and customer-supplied job input do not reach info logs

## Why

`projects_handler` wrote the complete API Gateway event to CloudWatch. That event can contain `Authorization`, `Cookie`, and request-body data. The async persona path also repeated the job event and its filters, including customer-supplied instructions.

The handler already logs bounded project/job identity and completion state, so removing the raw payload logs preserves operational progress without retaining credentials or customer content.

Fixes #245.

## Verification

- red-first regressions: 2 failures before the production change
- focused sensitive-event regressions: 2 passed
- projects handler suite: 18 passed
- complete Lambda suite: 528 passed
- Python compile check for both changed files
- raw-event logging pattern audit
- `git diff --check`

The tests capture info-log calls and use sentinel values in API Gateway credentials and persona-job input. No deployed CloudWatch log group or existing retained log data was inspected or changed.

## AI assistance

OpenAI Codex assisted with investigation, implementation, regression tests, and verification. I reviewed the complete submitted diff and remain responsible for the contribution.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution under the terms of the project license.

### DIFF
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -272,7 +272,6 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
     
     logger.info(f"[PERSONA] ========== STARTING PERSONA GENERATION ==========")
     logger.info(f"[PERSONA] Project: {project_id}")
-    logger.info(f"[PERSONA] Filters: {filters}")
     overall_start = time.time()
     
     def update_progress(progress: int, step: str):
--- a/voc-datalake/lambda/api/projects_handler.py
+++ b/voc-datalake/lambda/api/projects_handler.py
@@ -13,7 +13,7 @@ from shared.logging import logger, tracer
 from shared.aws import (
     get_dynamodb_resource, get_bedrock_client, invoke_self_async, BEDROCK_MODEL_ID
 )
-from shared.api import create_api_resolver, validate_days, validate_int, api_handler, DecimalEncoder
+from shared.api import create_api_resolver, validate_days, validate_int, api_handler
 from shared.converse import converse
 from shared.tables import get_jobs_table, get_aggregates_table
 from shared.jobs import create_job, update_job_status
@@ -765,8 +765,20 @@ def lambda_handler(event: dict, context: Any) -> dict:
             return handle_import_persona_job(event)
         
         # Normal API Gateway request
+        request_context = event.get('requestContext') or {}
+        logger.info(
+            'Request received',
+            extra={
+                'method': event.get('httpMethod'),
+                'resource': event.get('resource'),
+                'request_id': request_context.get('requestId'),
+            },
+        )
         result = app.resolve(event, context)
-        logger.info(f"Returning result: {json.dumps(result, cls=DecimalEncoder)}")
+        logger.info(
+            'Request completed',
+            extra={'status_code': result.get('statusCode')},
+        )
         return result
         
     except Exception as e:
--- a/voc-datalake/lambda/api/projects_handler.py
+++ b/voc-datalake/lambda/api/projects_handler.py
@@ -349,7 +349,6 @@ def handle_generate_personas_job(event: dict) -> dict:
     import time
     
     logger.info(f"[JOB] ========== ASYNC PERSONA JOB STARTED ==========")
-    logger.info(f"[JOB] Event: {event}")
     job_start = time.time()
     
     project_id = event['project_id']
@@ -357,7 +356,6 @@ def handle_generate_personas_job(event: dict) -> dict:
     filters = event['filters']
     
     logger.info(f"[JOB] Project: {project_id}, Job: {job_id}")
-    logger.info(f"[JOB] Filters: {filters}")
     
     def progress_callback(progress: int, step: str):
         logger.info(f"[JOB] Progress callback: {progress}% - {step}")
@@ -755,8 +753,6 @@ CRITICAL: Output ONLY valid JSON, no markdown, no explanation."""
 def lambda_handler(event: dict, context: Any) -> dict:
     """Main Lambda handler for projects API."""
     try:
-        logger.info(f"Received event: {json.dumps(event)}")
-        
         # Route async job invocations
         job_type = event.get('job_type')
         if job_type == 'generate_personas':
--- a/voc-datalake/lambda/api/test/test_projects_handler.py
+++ b/voc-datalake/lambda/api/test/test_projects_handler.py
@@ -3,7 +3,7 @@ Tests for projects_handler.py - /projects/* endpoints.
 """
 import json
 import pytest
-from unittest.mock import patch, MagicMock
+from unittest.mock import DEFAULT, MagicMock, patch
 from datetime import datetime, timezone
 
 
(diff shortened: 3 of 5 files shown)


perrozzi:

This PR removes three raw-payload `logger.info` calls from `voc-datalake/lambda/api/projects_handler.py` (full API Gateway event at handler entry, plus the async persona job event and its filters) and adds two regression tests that assert sentinel values from request headers and job filters never reach `logger.info`. The direction is right and matches the "no PII in logs" rule, and both tests look like genuine red-first regressions (reverting either deletion re-introduces the sentinel into `call_args_list`). A few gaps around log-event configuration, lost observability, and coverage of sibling handlers.

### Should fix

- **Verify Powertools isn't re-logging the event anyway.** Deleting the explicit `logger.info(f"Received event: ...")` only closes the leak if the Logger is not configured with `@logger.inject_lambda_context(log_event=True)` and the `POWERTOOLS_LOGGER_LOG_EVENT` env var is not set on this function in the CDK stack. Neither is visible in the diff. If either is present, the raw event (including `Authorization`/`Cookie`) still lands in CloudWatch and the new test would not catch it, since Powertools emits that record itself rather than through `logger.info`. Please confirm and, if needed, pin `log_event=False`.
- **Preserve bounded entry-point observability.** `lambda_handler` in `projects_handler.py` now appears to have no entry log at all. The async job path still logs `[JOB] Project: ..., Job: ...`, but the sync API path loses routing visibility entirely. Replace the deleted line with a redacted structured log, e.g. `logger.info("request received", extra={"method": ..., "resource": ..., "request_id": context.aws_request_id})` — this keeps triage possible without retaining credentials or bodies.
- **Apply the same fix to the other domain handlers.** The PR description mentions a "raw-event logging pattern audit," but the diff only touches `projects_handler.py`. Since API Lambdas are split by domain, any sibling handler in `voc-datalake/lambda/api/` with the same `Received event: {json.dumps(event)}` line has the identical leak; either fix them in this PR or state explicitly that the audit found none.
- **Confirm `json` is still referenced in `projects_handler.py`.** Removing the only `json.dumps(event)` call can leave an unused import and a lint failure. Likely still used for response bodies, but worth checking.
- **Test assertions are coarse.** `assert secret not in str(mock_log_info.call_args_list)` in `voc-datalake/lambda/api/test/test_projects_handler.py` only covers `logger.info`. A future `logger.debug(event)` or an `except` block that logs the event on error would pass. Consider patching the whole `projects_handler.logger` and asserting the sentinel is absent across all level calls, and adding a positive assertion that the bounded identity log still fires so the observability contract is pinned too.

### Nits

- `test_async_job_payload_is_not_logged` asserts `result == {'success': True}` by making `generate_personas` return exactly that dict. If `handle_generate_personas_job` later wraps the result (adding `job_id`, timings, etc.), the test breaks for reasons unrelated to logging. Asserting only the log invariant would make the regression more durable.
- `test_api_gateway_event_is_not_logged` depends on the `api_gateway_event` fixture accepting a `headers=` kwarg; please confirm the fixture signature supports it rather than silently ignoring it (which would make the test vacuously pass).
- Two trailing blank lines are added at the end of `test_projects_handler.py`.
- Pre-existing but adjacent: `logger.info(f"[JOB] ========== ASYNC PERSONA JOB STARTED ==========")` is an f-string with no placeholders (ruff F541) and would be a natural candidate for structured `extra=` fields while you're in this function.

---

perrozzi:

Since my last review the author addressed most of the earlier feedback: `log_event=False` is now pinned in `shared/api.py` (with a test asserting it), bounded structured `Request received` / `Request completed` logs replaced the deleted entry log, the full-result `json.dumps(result, cls=DecimalEncoder)` log and the `[PERSONA] Filters` log were also removed, and both regressions now patch every log level plus assert the surviving bounded logs.

### Resolved
- Powertools re-logging closed: `@logger.inject_lambda_context(log_event=False)` in `voc-datalake/lambda/shared/api.py` (explicit `False` correctly overrides `POWERTOOLS_LOGGER_LOG_EVENT`).
- Entry-point observability restored via redacted `extra={method, resource, request_id}` in `projects_handler.lambda_handler`.
- Test coarseness fixed: `patch.multiple(logger, ...)` covers debug/info/warning/error/exception/critical, and positive assertions pin the bounded logs.
- Bonus leak closed: `Returning result: {json.dumps(result)}` (full response body) and `[PERSONA] Filters` removed; `DecimalEncoder` import dropped with it.
- Nit resolved: the async test no longer asserts on `generate_personas`' exact return shape.

### Still open
- Sibling handlers in `voc-datalake/lambda/api/` were not touched; please state that the raw-event audit found no other `json.dumps(event)` logs, or fix them here.
- Confirm `json` is still referenced in `projects_handler.py` after removing both `json.dumps` call sites (unused-import lint risk).

### New findings
- `voc-datalake/lambda/api/test/test_projects_handler.py::test_async_job_payload_is_not_logged`: the regression now depends on `@patch('projects.projects_table', None)` raising `ServiceError` mid-flight, so it only proves the secret is absent up to the failure point and asserts exact downstream log strings from another module (`'[PERSONA] Project: project-1'`). This is brittle and no longer covers the success path where filters historically flowed. Prefer mocking `generate_personas` to succeed (asserting only the log invariant) or add a second success-path case alongside this failure-path one.
- Neither test covers the `except` branch of `lambda_handler`, which is where an event-dumping log would most plausibly be re-added; a small failure-path case asserting the sentinel is absent from `logger.error/exception` would pin that contract too.

---

shin4141:
Follow-up on the automated review:

| Finding | Disposition | Action / evidence |
| --- | --- | --- |
| Powertools event re-logging | **CONFIRMED configuration path; deployed configuration was not inspected** | The shared decorator previously left `log_event` environment-controlled. It now sets `log_event=False` explicitly. A regression run with `POWERTOOLS_LOGGER_LOG_EVENT=true` is green, and the shared decorator contract is pinned. Tracked CDK did not set the variable. |
| Bounded entry observability | **CONFIRMED** | Added allowlisted entry telemetry for method, API Gateway resource template, and request ID, plus completion status only. Credentials, headers, body, filters, and response content are excluded. Existing project/job/progress telemetry remains. |
| Identical sibling-handler event dumps | **REFUTED** | Audited the sibling API handlers: no remaining `json.dumps(event)`, direct event log, or equivalent full-event dump was found, so no sibling-handler refactor was made. |
| `json` import unused | **REFUTED** | `json` remains required for invocation/parsing/error handling. Only `DecimalEncoder`, which became unused after removing full-result logging, was removed. |
| Regression coverage too narrow | **CONFIRMED** | Regressions now inspect debug/info/warning/error/exception/critical, exercise a POST body and echoed response, traverse the real persona helper path, and positively assert bounded telemetry. |

The call-path audit also found two concrete same-boundary disclosures not covered by the earlier tests: full persona `filters` (including `custom_instructions`) in `projects.py`, and the complete resolved API response in `projects_handler.py`. Both raw logs were removed.

Verification: red-first **2 failures**; focused sensitive regressions **2 passed**; unsafe Powertools-environment regression + decorator contract **2 passed**; Projects handler suite **18 passed**; complete Lambda suite **528 passed**; Python compilation, exact logging-pattern audit, sibling-handler audit, and `git diff --check` passed.

PR #332, opened 2026-08-15
fix(feedback-forms): honest submission counts, one partition per form

perrozzi:
## Summary

Two defects in `voc-datalake/lambda/api/feedback_form_handler.py` corrupted the submission count and average rating a form reports. That number is rendered next to a prioritization score, so a wrong value silently changes a product decision. Python backend only — no frontend, CDK, or locale file is touched.

**Invariant 1 — a failed stats read must not report zero.** `get_form_stats` returned HTTP 200 with `{'success': True, 'stats': {'total_submissions': 0, 'avg_rating': None}}` from both its `except Exception` handler and its "table not configured" branch, so a DynamoDB failure or a Lambda deployed without `FEEDBACK_TABLE` was indistinguishable from a form nobody had answered. It now raises `ServiceError` on an unexpected read failure and `ConfigurationError` when the table is not configured — exactly what `get_form_submissions`, over the same table, has always done. Closes #312.

**Invariant 2 — a brand rename must not split the partition the stats read queries.** `submit_form_feedback` stamped the enqueued record's `brand_name` from the module-level `BRAND_NAME` environment variable, while `_get_form_source_pk` builds the query partition from the FORM's stored `brand_name` and only falls back to the environment. Once a deployment's brand name changed, new submissions landed under the new brand while the stats read still queried the old one, and the form reported 0 submissions with the data sitting intact elsewhere. Fixed at the WRITE site: the record now carries `form.get('brand_name') or BRAND_NAME`. The form record is already loaded a few lines above for the `enabled` check, so this costs no extra read, and every submission stays in the partition the read queries for the whole life of the form.

[...]

PR #335, opened 2026-08-16
fix(prototype): configure a build in a wizard, not on the card face

perrozzi:
## What this changes

Project Overview step 5 carried about a dozen build-configuration tick-boxes on the card surface, while its five siblings each have a single button. They were there by elimination rather than by design: the confirm dialog they would otherwise have lived in opened for only *some* projects, so a control inside it would have been unreachable for the simplest one — a project with one PRD, one PR/FAQ and no prototype.

That is an argument against **that dialog**, not for the card face. The card now opens a wizard, which is where the build is configured and started.

It also fixes something adjacent that was not raised: **pressing Build was unpredictable**. Because the confirm fired on three conditions, a project with two PRDs opened a dialog and a project with one of each immediately started a billable multi-minute build. A user could not tell which they would get. Now it always opens the wizard, and nothing is spent until the wizard's own button.

## The gate had to go with the move, not after it

`confirmKeyFor` decided whether anything opened at all, so the controls could not become reachable for every project while it remained. Sequencing them separately would leave the configuration in two places at once, which is worse than either end state.

- `usePrototypeBuild` owns a plain `isOpen`. `confirmKeyFor` is demoted from a gate to a **warning selector**, so the panel always opens and the caution still follows live documents.
- **This is the one place where copying the old `askedKey`/`openKey` comparison would be a bug.** That comparison closed a dialog whose *reason* had gone stale — correct for something whose whole content is a question, wrong for something holding the user's selections, because documents refetch whenever a job completes and the ticks would be discarded by an unrelated job finishing. Visibility is now owned; the warning stays derived.
- `PrototypeBuildWizard` is built on `ModalShell`, which already owns the dialog role, the accessible name, the focus trap, Escape and correct nesting. Not `ConfirmModal`, which requires a `message` — right for a question, wrong for a form.
- `ActionCard` loses `controls`, a prop only one of six cards passed.

Both existing guards on the billable endpoint are kept, and opening the wizard starts no build.

## Tests: what was retired, and what it was traded for

`OverviewTab.prototypeWizard.test.tsx` replaces the old "controls are reachable without a dialog" property with the one it is traded for: **the configuration is reachable for every project, including the case that previously opened nothing.** Parametrised over all five project shapes.

The panel is identified by what it *contains* rather than by role, because four of the five shapes already opened a dialog and a role-only assertion would pass on unchanged code.

Four suites asserted the old placement. Most of that was mechanical, but these assertions were about properties deliberately gone, and each is rewritten rather than deleted because each protected something that still matters:

| Retired | Replaced by |
|---|---|
| builds immediately without a confirmation when both documents exist | opens with **no warning** and spends nothing until the wizard's button |
| does not ask when there is exactly one of each | the picker is shown and offers **no select** — a one-answer choice must not be presented as a choice |
| closes the confirm when its reason disappears | panel **stays open**, caution **clears** |
| closes the confirm rather than swapping its question for a costlier one | panel stays open, caution **escalates** to the rebuild warning |
| does not re-open the confirm when a later reason arrives | kept in substance: **never opens on its own**, whatever the documents do |
| a disabled card renders no tick-boxes | a disabled button **cannot open the panel** — absence of tick-boxes alone would be *weaker*, since they are absent on every unopened card |

### One residual risk, named rather than hidden

Because the caution is derived from live documents, it can now **escalate under an open panel**: a prototype arriving mid-interaction changes what pressing Build costs, from "build from the PRD alone" to "build a second and keep the first". This is materially weaker than the #296 defect it replaces — there a stale message sat one click from spending, whereas here the caution renders into the panel the user is looking at and they must still press Build — but it is not nothing, so there is a test for it.

### Eight tests had started passing for the wrong reason

They assert "asks for confirmation instead of building". The wizard always asks, so the assertion became true for every project and the suite would be green whichever way the code went — their discriminator ceased to exist. Found by auditing the passes I could not account for, not the failures. Renamed and re-pointed at what they can still tell apart.

## Verification

- frontend vitest **2622 / 2622** (174 files); 2613 baseline + 9 new reconciles exactly
- `tsc` 0 errors; eslint 0 on the changed files
- **i18n gate byte-identical to `development`**, measured against the base directly rather than assumed — no catalogue was touched and no key added

## Deliberately not in this PR

1. **A second entry point from the Documents tab.** The issue proposes one wizard invoked from both the card and the prototype pane, which satisfies U23's "open the surface, not a dialog" principle without making the card useless. It needs a decision on U24's larger question first.
2. **The submit label.** It still reads "Build anyway", which is odd with no warning to build anyway *despite*. The clean fix adds no keys — submit reuses `documents.prototype.button` and the card takes the `overview.configureAnd` prefix its five siblings already carry — but it is a wording decision rather than a refactor.
3. `ActionCard.statusLine` remains prototype-only, the last single-call-site prop.

Refs `analysis/UI-prototype-build-configuration-GITHUB-ISSUE.md`, which records the two corrections this work forced on its own plan.

### DIFF
--- a/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
@@ -95,7 +95,7 @@ function renderCard(props: { hasPrd: boolean; hasPrfaq: boolean; hasPrototype?: 
 
 /** The build trigger, not the modal's confirm button. */
 function buildButton() {
-  return screen.getByRole('button', { name: /build prototype/i })
+  return screen.getByRole('button', { name: /configure & build prototype/i })
 }
 
 /**
@@ -103,7 +103,7 @@ function buildButton() {
  * before the panel is opened.
  */
 function confirmButton() {
-  return screen.queryByRole('button', { name: /build anyway/i })
+  return screen.queryByRole('button', { name: /^build prototype$/i })
 }
 
 /**
@@ -116,7 +116,7 @@ function confirmButton() {
  */
 async function startBuildVia(user: ReturnType<typeof userEvent.setup>) {
   await user.click(buildButton())
-  await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /build anyway/i }))
+  await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /^build prototype$/i }))
 }
 
 describe('prototype card confirm gate (U12)', () => {
@@ -151,7 +151,7 @@ describe('prototype card confirm gate (U12)', () => {
     renderCard({ hasPrd: true, hasPrfaq: false })
     await user.click(buildButton())
 
-    await user.click(screen.getByRole('button', { name: /build anyway/i }))
+    await user.click(screen.getByRole('button', { name: /^build prototype$/i }))
 
     await waitFor(() => expect(mockBuildPrototype).toHaveBeenCalledTimes(1))
     expect(mockBuildPrototype).toHaveBeenCalledWith('proj_1', expect.anything())
@@ -188,7 +188,7 @@ describe('prototype card confirm gate (U12)', () => {
     expect(screen.queryByText(/already has a prototype/i)).not.toBeInTheDocument()
 
     // ...and it builds once asked.
-    await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /build anyway/i }))
+    await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /^build prototype$/i }))
     await waitFor(() => expect(mockBuildPrototype).toHaveBeenCalledTimes(1))
   })
 
@@ -264,7 +264,7 @@ describe('prototype card rebuild guard', () => {
     renderCard({ hasPrd: true, hasPrfaq: true, hasPrototype: true })
     await user.click(buildButton())
 
-    await user.click(screen.getByRole('button', { name: /build anyway/i }))
+    await user.click(screen.getByRole('button', { name: /^build prototype$/i }))
 
     await waitFor(() => expect(mockBuildPrototype).toHaveBeenCalledTimes(1))
   })
@@ -320,6 +320,25 @@ describe('prototype card rebuild guard', () => {
     expect(screen.getByRole('dialog')).toBeInTheDocument()
     expect(screen.queryByText(/No PR-FAQ yet/i)).not.toBeInTheDocument()
     expect(mockBuildPrototype).not.toHaveBeenCalled()
+  })
+
+  it('announces the escalated caution, not only renders it', async () => {
+    // The caution can appear or change while the panel is open, so a sighted user
+    // sees the amber block move and a screen-reader user must be told. Asserted on
+    // the live region CONTAINING the new text, because a region that is only mounted
+    // once there is something to say cannot announce its own arrival.
+    const user = userEvent.setup()
+    const { rerender } = renderCard({ hasPrd: true, hasPrfaq: true })
+
+    await user.click(buildButton())
+    const region = within(screen.getByRole('dialog')).getByRole('status')
+    expect(region).toHaveAttribute('aria-live', 'polite')
+    expect(region).toBeEmptyDOMElement()
+
+    rerender(overviewTab([doc('prd', 'prd_1'), doc('prfaq', 'prfaq_1'), doc('prototype', 'proto_1')]))
+
+    expect(within(screen.getByRole('dialog')).getByRole('status'))
+      .toHaveTextContent(/already has a prototype/i)
   })
 
   it('escalates the caution to the rebuild warning when a prototype arrives mid-interaction', async () => {
--- a/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
+++ b/voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.prototypeCard.test.tsx
@@ -24,7 +24,7 @@
  * assistive tech because both the code and its test agreed on a missing key.
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest'
-import { render, screen, waitFor } from '@testing-library/react'
+import { render, screen, waitFor, within } from '@testing-library/react'
 import userEvent from '@testing-library/user-event'
 import OverviewTab from './OverviewTab'
 import type { Project, ProjectDocument } from '../../api/types'
@@ -99,12 +99,24 @@ function buildButton() {
 }
 
 /**
- * The modal's confirm action, queried rather than got: the tests below assert on
- * its ABSENCE, which is the load-bearing half of a dialog that must not be
- * clickable once its question no longer applies.
+ * The wizard's submit, queried rather than got, so its ABSENCE can be asserted
+ * before the panel is opened.
  */
 function confirmButton() {
   return screen.queryByRole('button', { name: /build anyway/i })
+}
+
+/**
+ * Open the wizard and start the build — two clicks, because the card's button no
+ * longer spends money.
+ *
+ * Every test that wants a build to actually run goes through here, so the two-step
+ * shape is stated once. A test that only wants the panel open uses `buildButton()`
+ * alone, and one that asserts nothing was spent asserts it after that first click.
+ */
+async function startBuildVia(user: ReturnType<typeof userEvent.setup>) {
+  await user.click(buildButton())
+  await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /build anyway/i }))
 }
 
 describe('prototype card confirm gate (U12)', () => {
@@ -156,15 +168,28 @@ describe('prototype card confirm gate (U12)', () => {
     expect(screen.queryByText(/No PR-FAQ yet/i)).not.toBeInTheDocument()
   })
 
-  it('builds immediately without a confirmation when both documents exist', async () => {
+  // This replaces "builds immediately without a confirmation when both documents
+  // exist". That one-click path was traded away deliberately when the build moved
+  // into a wizard: it is what made the card unpredictable, since a project with two
+  // PRDs opened a dialog and this one did not. The successor property is that the
+  // panel opens with NO warning and still spends nothing until its own button.
+  it('opens the wizard with no warning, and spends nothing, when both documents exist', async () => {
     const user = userEvent.setup()
     renderCard({ hasPrd: true, hasPrfaq: true })
 
     await user.click(buildButton())
 
-    await waitFor(() => expect(mockBuildPrototype).toHaveBeenCalledTimes(1))
+    expect(screen.getByRole('dialog')).toBeInTheDocument()
+    expect(mockBuildPrototype).not.toHaveBeenCalled()
+    // Nothing to caution about on this project, so no caution is shown — an
+    // always-rendered warning block would be an empty amber panel here.
     expect(screen.queryByText(/No PR-FAQ yet/i)).not.toBeInTheDocument()
     expect(screen.queryByText(/No PRD yet/i)).not.toBeInTheDocument()
+    expect(screen.queryByText(/already has a prototype/i)).not.toBeInTheDocument()
+
+    // ...and it builds once asked.
+    await user.click(within(screen.getByRole('dialog')).getByRole('button', { name: /build anyway/i }))
+    await waitFor(() => expect(mockBuildPrototype).toHaveBeenCalledTimes(1))
   })
 
   it('does not build when neither document exists', async () => {
@@ -198,7 +223,7 @@ describe('prototype card confirm gate (U12)', () => {
     }))
     renderCard({ hasPrd: true, hasPrfaq: true })
 
-    await user.click(buildButton())
+    await startBuildVia(user)
 
     // The label becomes "Building…" while in flight, so the trigger has to be found
     // by that name — the disabled state is real, and the point is what it *says*.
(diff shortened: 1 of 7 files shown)


perrozzi:

This PR moves the prototype build configuration off the Overview card face into a new `PrototypeBuildWizard` (built on `ModalShell`), and demotes `confirmKeyFor` from an open-gate to a warning selector so the panel opens for every project shape rather than only the ones that raised a confirm. The change is frontend-only (no CDK/IAM, Lambda, or i18n catalogue changes), reuses existing translation keys, and the test rework is unusually careful — retired assertions are re-pointed rather than deleted, and the new `OverviewTab.prototypeWizard.test.tsx` parametrises the placement invariant over all five project shapes. Overall this looks clean; my comments are about the residual risks the change creates rather than defects in what is written.

### Should fix

- **The escalating caution isn't announced.** `PrototypeBuildWizard` in `voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.tsx` renders the warning as a plain `<p>`. Since `wizard.warning` is now derived from live documents while the panel is open (the risk you name explicitly, and cover in `OverviewTab.prototypeCard.test.tsx` → "escalates the caution to the rebuild warning"), a screen-reader user who opened with no warning gets no notification that the cost of the submit changed. Add `role="status"` / `aria-live="polite"` to that block (no new keys needed), and extend the escalation test to assert the live region so the property is pinned.
- **No test that the submit is safe once the card's precondition lapses.** Visibility is now owned by the hook, so an open wizard can outlive the state that enabled the card button — e.g. both PRD and PR-FAQ deleted while the panel is open. The hook's internal guard in `runBuild` is then the only thing standing between the user and the billable endpoint, and that guard isn't exercised from the wizard path in the diff. Add a regression test in `OverviewTab.prototypeWizard.test.tsx`: open, rerender with documents removed, press the wizard's submit, assert `buildPrototype` was not called.
- **The `build.busy` branch on the wizard submit looks unreachable.** `onConfirm` in `usePrototypeBuild.ts` sets `isOpen = false` before `runBuild()`, and the card button is disabled while busy, so the panel cannot be open while a start is in flight — the spinner and `disabled={build.busy}` in `PrototypeBuildWizard` are dead. Either keep the panel open until the start resolves (so the spinner and the error surface where the user is looking) or drop the branch; as written it reads as protection that isn't there.
- **Submit copy on the no-warning path.** `documents.prototype.confirmBuild` ("Build anyway") is now the label for the *common* case where there is nothing to build anyway *despite*. You flag this as deliberately out of scope, but it's user-visible text that only became wrong in this PR, and the fix you sketch (submit reuses `documents.prototype.button`) adds no keys — worth doing here rather than leaving a known-odd string in.

### Nits

- `confirmKeyFor` / `ConfirmKey` in `usePrototypeBuild.ts` now select a warning, not a confirm; renaming would stop the next reader looking for a gate.
- Stale comment in `OverviewTab.prototypeSources.test.tsx`: "With one of each there is no dialog, but the ids are still sent" — the test immediately below now clicks through a dialog.
- `OverviewTab.prototypeWizard.test.tsx` calls `userEvent.click(...)` directly; the sibling suites all use `userEvent.setup()`, which is the repo pattern and gives consistent pointer/clipboard config.
- `ModalShell` gets `ariaLabel={t('documents.prototype.button')}` while the panel also renders the same string as an `<h3>`; if `ModalShell` supports `aria-labelledby`, pointing at the heading id avoids two names drifting apart.
- The wizard's cancel/submit buttons have no `type="button"` — harmless today, a submit-on-Enter surprise if either is ever nested in a form.

---

perrozzi:
## Round 1 disposition — all nine taken, one by the other route you offered

Thanks, this was a useful pass: three of the four should-fix items are about consequences of *owned visibility* that I had not followed through, and the fourth correctly refuses my scope excuse.

### Should fix

**1. The escalating caution isn't announced — fixed.** The warning block is now wrapped in `role="status"` / `aria-live="polite"`, no new keys. One detail worth stating because it is easy to get wrong: the live region wraps the **condition**, not the text. An element that only mounts once there is something to say cannot announce its own arrival, so the container is always present and only its contents change. Added `announces the escalated caution, not only renders it`, which asserts the region is empty on open and then contains the rebuild wording after the document change — so it fails if the region is moved inside the conditional.

**2. No test that the submit is safe once the card's precondition lapses — fixed, and you were right that nothing covered it.** Added `refuses to build once the documents that enabled the card are gone`: open the panel, rerender with both source documents removed, press the wizard's submit, assert no call. The card's `disabled` genuinely cannot help there — it is behind the modal and the panel is already up — so the hook's internal guard is the last thing standing, and it now has a test exercising it from this path.

**3. The `build.busy` branch is unreachable — confirmed, and I dropped it rather than keeping the panel open.** You are right that it is dead: `onConfirm` closes the panel before `runBuild()`. I took the second option because the first would duplicate reporting that already exists. The card's status line owns both the busy label and the failure-to-start line, both already tested, and it is what the user is looking at once the panel closes. Adding a second surface for the same two states would mean two places to keep in step. The comment now says why there is no busy state here, so it does not read as an omission.

**4. Submit copy — taken, and the scope excuse was wrong.** You are right that the string only became wrong in this PR. Done exactly as sketched, with no new keys: the submit reuses `documents.prototype.button` ("Build Prototype"), and the card takes the `overview.configureAnd` prefix its five wizard-opening siblings already carry, so it reads "Configure & Build Prototype". That is *more* consistent than before, since the prefix was absent only because the card used to act directly.

One consequence worth flagging for review: both buttons now contain "Build Prototype", so six test queries that matched `/build prototype/i` became ambiguous. The card is now matched by the prefixed name and the submit by an anchored `/^build prototype$/i`, which reads better anyway — each query now says which button it means.

### Nits — all five taken

- `confirmKeyFor` → `warningKeyFor`, `ConfirmKey` → `WarningKey`, `CONFIRM_MESSAGE` → `WARNING_MESSAGE`. The table still holds literal key strings, since the i18n extractor cannot see a key assembled at runtime.
- Stale comment in `OverviewTab.prototypeSources.test.tsx` corrected — "no dialog" became "nothing to choose".
- `OverviewTab.prototypeWizard.test.tsx` now uses `userEvent.setup()` per test, matching the sibling suites.
- `ModalShell` now gets `ariaLabelledBy` pointing at the heading the panel already renders, so the accessible name cannot drift from the visible one and there is no second translated string to keep in step. The union type on those props made this a one-line swap.
- `type="button"` on both wizard buttons.

### Verification after these changes

- frontend vitest **2624 / 2624** (174 files) — the 2 additions above
- `tsc` 0, eslint 0 on the changed files
- i18n gate re-measured and unchanged from `development`: no key added, no catalogue touched

Pushing now.

perrozzi:

Since the last review the author addressed every Should-fix and nit: the wizard warning is now a permanently mounted `role="status"`/`aria-live="polite"` region with a dedicated announcement test, the dead `build.busy` spinner branch is gone (with a comment explaining why), the submit reuses `documents.prototype.button` while the card takes the `overview.configureAnd` prefix, a new regression test asserts the submit refuses once the enabling documents are deleted, `confirmKeyFor`→`warningKeyFor` is renamed throughout, `type="button"` added, `ariaLabelledBy` points at the heading, and the wizard suite now uses `userEvent.setup()`.

### Resolved
- Escalating caution is announced: `role="status" aria-live="polite"` wrapper in `OverviewTab.tsx` + "announces the escalated caution" test in `OverviewTab.prototypeCard.test.tsx`.
- Submit-after-precondition-lapse regression test added (`OverviewTab.prototypeWizard.test.tsx`, "refuses to build once the documents that enabled the card are gone").
- Unreachable `build.busy` branch removed from the wizard submit, with the reasoning recorded.
- Submit copy: "Build anyway" replaced by `documents.prototype.button`; card takes `overview.configureAnd`. No new keys.
- Nits: `warningKeyFor`/`WarningKey` rename, `userEvent.setup()`, `aria-labelledby` over duplicated `ariaLabel`, `type="button"` on both buttons, stale "no dialog" comment in `OverviewTab.prototypeSources.test.tsx`.

### New findings
- **Should fix — the busy label now composes with the new prefix.** `voc-datalake/frontend/src/pages/ProjectDetail/OverviewTab.tsx` still passes `buttonLabel = busy ? t('documents.prototype.building') : t('documents.prototype.button')` while `configureLabel` is now passed unconditionally. If `ActionCard` concatenates prefix + label (as the five siblings imply), the in-flight card reads "Configure & Building…". The existing `findByText(/building…/i)` assertion matches a substring, so it would not catch this. Either suppress `configureLabel` while busy or confirm `ActionCard` drops the prefix for the busy label — and pin it with an assertion on the full accessible name.
- **Verify `ModalShell` accepts `ariaLabelledBy`.** The switch from `ariaLabel` to `ariaLabelledBy={headingId}` is only correct if the shell forwards it to `aria-labelledby`; `ModalShell` isn't in the diff, so if the prop doesn't exist this is either a type error or a silently unnamed dialog (the wizard tests query by role, not by name, so they won't fail).
- **Nit — unescaped translated string used as a regex.** `OverviewTab.prototypeExtraSources.test.tsx`: `new RegExp(i18n.t('documents.prototype.button', …), 'i')` will misbehave if any catalogue's label contains regex metacharacters; escape it, or match with a function matcher.
- **Nit — stale file header.** The same file's top comment still asserts `warningKeyFor` "deliberately opens NO dialog … the build starts on the first click — and the source picker lives inside that dialog". Only the function name was updated; the claim is now false.

Otherwise the delta looks clean, and the test rework tracks the behaviour change accurately.

---

perrozzi:
## Round 2 disposition — the should-fix was a real bug I introduced

**1. The busy label composes with the new prefix — correct, and it was a live defect, not a hypothetical.** Confirmed by reading `ActionCard`: it renders `{configureLabel}{buttonLabel}` with no separator logic, so with the prefix passed unconditionally the in-flight card read **"Configure & Building…"** for the duration of every start request. The prefix is now absent while busy.

You were also right that the existing assertion could not catch it, and this is the part worth recording: `findByText(/building…/i)` matches a substring, so it passes on both spellings. The new assertion is on the **full** accessible name, `/^building…$/i`.

**I verified it can actually fail** rather than assuming: restoring the unconditional prefix fails exactly that one assertion, and only that one. Anchor asserted before mutating, restored after, tree confirmed clean.

Two things fell out of that fix which are worth flagging because they changed the shape of it:

- My first attempt added a second assertion, `queryByRole('button', { name: /configure/i })).not.toBeInTheDocument()`. **It was both wrong and redundant** — `/configure/i` matches all five sibling wizard cards, so `queryByRole` threw on multiple matches. Dropped; the anchored name assertion is what carries the property.
- The inline ternary pushed `OverviewTab` to a complexity of 13 against a ceiling of 12, and extracting it with a boolean parameter then tripped `sonarjs/no-selector-parameter`. Settled as a module-level helper taking the build control, matching the existing `prototypeStatusLine(build, t)` shape in the same file. I have run eslint over all of `src` this round rather than just the changed files, so I am not trading one rule for another unseen.

**2. `ModalShell` accepts `ariaLabelledBy` — verified, no change needed.** Its name props are a union (`{ ariaLabel } | { ariaLabelledBy }`) precisely so omitting both is a compile error, and it renders `aria-labelledby={ariaLabelledBy}`. `tsc` passing is therefore evidence the prop exists, not merely that nothing objected.

Your underlying point stands though — the tests query by role, so a silently unnamed dialog would not have failed anything. Added `expect(panel).toHaveAccessibleName(en.documents.prototype.button)` to the reachability test, taken from the real catalogue rather than a hardcoded string.

**3. Unescaped translated string used as a regex — fixed, and replaced rather than escaped.** Switched to a function matcher (`(name) => name.includes(verb)`), which has no metacharacter failure mode at all and needs no escaping to stay correct.

**4. Stale file header — fixed.** You are right that I had updated only the function name. The header now says what that suite actually pins (what the controls *do*), points at the wizard suite for *where* they live, and keeps the one-of-each fixture with its corrected reason: it is the project a placement regression would strand first.

### Verification

- frontend vitest **2624 / 2624** (174 files)
- `tsc` 0; **eslint 0 across all of `src`**
- i18n gate unchanged from `development`
- the new busy-label assertion mutation-verified in both directions

Pushing now.

perrozzi:

Since my last review the author fixed all four items: `configurePrefix` now suppresses the `overview.configureAnd` prefix while busy (with an anchored `/^building…$/i` accessible-name assertion in `OverviewTab.prototypeCard.test.tsx`), the wizard's `ariaLabelledBy` is now pinned by `toHaveAccessibleName(en.documents.prototype.button)` in `OverviewTab.prototypeWizard.test.tsx`, the regex built from a translated string is replaced by an `includes` function matcher, and the stale file header in `OverviewTab.prototypeExtraSources.test.tsx` now describes the wizard placement.

### Resolved
- Busy label no longer composes with the prefix (`configurePrefix` in `OverviewTab.tsx`), pinned by an anchored full-name assertion rather than a substring.
- `ariaLabelledBy` is now proven to produce a named dialog (accessible-name assertion in `OverviewTab.prototypeWizard.test.tsx`).
- Translated label no longer interpolated into a `RegExp`; function matcher used instead.
- Stale `warningKeyFor` "opens NO dialog" header comment corrected.

No new problems: `configureLabel` was already optional (this card passed nothing before), the extracted helper keeps complexity in check, and the new assertions target the exact behaviours the previous findings named. Delta looks clean.

---

perrozzi:
## Deployed and live-verified before merge

Frontend-only deploy (no Lambda, no CDK — the backend half of the delta from the last live build was #332, and I confirmed by timestamp that another session had already deployed it).

**Live is serving this branch, measured three ways:** 162 of 162 `dist` files byte-identical to their S3 ETags with a bogus-key control, the entry chunk matches the local build, and every locale request carries `?v=4853afa`.

### The card

| | |
|---|---|
| Buttons on card 5 | **1** |
| Inputs/selects on its face | **0** |
| Label | **"Configure & Build Prototype"** |
| Cards 5 and 6 height | **213px both** — row 3 is even |

That settles the layout question the issue raised and left unmeasured. The grid is two columns from `sm` up, so 5 and 6 share a row and an over-tall card 5 used to stretch 6; they are now equal.

### The wizard

Opens with `aria-modal="true"` and an accessible name of **"Build Prototype"** resolved through `aria-labelledby` from the heading it renders — so the `ariaLabel` → `ariaLabelledBy` switch is confirmed working against a real browser, not only against jsdom. It contains the wizard body, the source picker and the extra sources (2 selects + 1 checkbox on this project).

The live region is present with `aria-live="polite"` and carries the rebuild caution, because this project already has 8 prototypes:

> This project already has a prototype. Building again adds a new one and keeps the existing one. Continue?

Its buttons are **Cancel** and **Build Prototype** — "Build anyway" is gone.

### Nothing was spent

Opening the panel made **zero** network calls, checked two independent ways: an in-page spy on `fetch` and `XMLHttpRequest` recorded 0 calls of any kind, and the browser's own network log for the whole session shows **no POST to `build-prototype`** (the only POSTs are Cognito auth). Ticking a box and cancelling also spent nothing, and the card returned to one button with no inputs.

### The guard, on a project with nothing to build from

Button disabled, the reason on the card rather than in a hover title ("Create a PRD or a PR-FAQ first to enable prototype build"), the click swallowed, no dialog, and zero checkboxes anywhere in the document.

### What I did NOT verify live, and why

- **The one-PRD + one-PR-FAQ + no-prototype project** — the case this whole change exists for. No project in the live environment is in that shape, and creating the documents to make one is a billable generation. It stays covered by the parametrised suite, which asserts it alongside the four other shapes.
- **The caution *appearing* mid-interaction.** The rebuild caution is confirmed rendered and in a live region, but observing it arrive would need a document to land while the panel is open, i.e. another billable generation. The transition is unit-tested.

### Unrelated, spotted while measuring

Card 6 reads **"Configure & Select & Remix"** — two ampersands, because its label already contains one. Pre-existing and not touched here, but it is the same concatenation behaviour as the busy-label bug from round 2, so worth the same one-line treatment sometime.

Issue #337, opened 2026-08-16
[Feature]: A room can score a proposal from their phones via a facilitator-opened session QR


perrozzi:
## Summary
A prioritization meeting scores proposals as a room, and the only way to get the room's opinion in is for
each person to be a signed-in reviewer on a laptop. The product already puts a QR on screen for customer
feedback forms; there is no equivalent for scoring a proposal, and the scoring path requires a login.

This adds a facilitator-opened voting session whose QR a room can scan to submit one ballot per device,
without an account, straight into the same per-reviewer ballot records the page already reads.

## Severity
Medium-high for the workshop use case, and it is the half that makes the combined score worth showing: a
team mean over the two people who happen to have accounts is not the room's opinion.

## Risk
**This adds a public write route, and that is the whole risk.** Ballots are currently written only by
authenticated reviewers. Anonymous room voting means an endpoint that accepts a score from anyone holding
a link, so the session — not obscurity — has to be the control.

**One device, one ballot, and no more than that.** A minted ballot id upserts its own record, so
re-submitting corrects a vote rather than adding one. Stuffing then requires minting new ids, which is
what a per-session cap has to bound, enforced by a conditional atomic increment on the session record so
a flood is refused by the database rather than by a race-prone check.

**Anonymous ballots are not accountable, and the UI must say so.** They keep count and spread; they do
not establish who thought what. Do not present them as attributed.

**An anonymous ballot must never overwrite a signed-in reviewer's.** The ballot key already namespaces
the two kinds for exactly this reason; anonymous ballots take the anonymous namespace.

**A display name is untrusted, optional PII.** Cap its length, strip control characters, never log it,
and keep it out of any prompt or export that does not need it.

**Room votes fragment when one proposal has two rows.** See the note on the row unit: if the same idea
appears as a PRD row and a PR/FAQ row, a room scanning one of them scores half the idea. Resolve the unit
or state the limitation on the page.

## Where this must NOT go
- **Not on the authenticated projects proxy.** That proxy applies Cognito to everything under it, and a
  public route carved out of it is precisely the shape that once left form update, delete and submission
  reads unauthenticated. Ballot submission gets its own resource and its own small handler with
  permission on one table only.
- **Not through the feedback-form routes.** Those are deliberately public for an embeddable customer
  widget, and a test pins exactly which routes are unauthenticated. A fourth public route means editing
  that allowlist on purpose, in review, with the test as the gate.
[...]

Issue #336, opened 2026-08-16
[Feature]: Prioritization rows should lead with the team score, not the reader's own


perrozzi:
## Summary
Prioritization is a group exercise presented as a personal one. The row leads with the reader's own four
numbers, the list sorts by the reader's own composite, and the fact that several reviewers may disagree
about a proposal is not on the page at all — even though the API now returns, per document, the mean of
each axis, how many reviewers scored it, and how far apart they are.

This makes the row show what the group thinks, and keeps the reader's own scoring where it belongs: one
level in, behind the expansion.

## Severity
Medium. Nothing is lost or corrupted, but the page answers the wrong question: a reader ranking a backlog
sees their own opinion ranked, and the disagreement that a prioritization meeting exists to resolve is
invisible.

## Risk
**The number a reader acts on changes meaning, so the two must not be confusable.** Today's headline is
"my composite". Tomorrow's is "the team's mean composite". A row that shows one and sorts by the other, or
a row where a reader cannot tell which they are looking at, is worse than either alone.

**An unscored proposal must not read as a badly-scored one.** The frontend's default score object uses
zero for three axes and three for time-to-market, and one call site renders an unset axis as three, so an
untouched proposal already lands mid-table. With a team mean as the headline this stops being cosmetic:
absent from the aggregate means nobody scored it, which is a different statement from "the team scored it
low", and the sort has to keep them apart.

**The row already carries a second numeric story.** A linked feedback form contributes a submission count
and an average star rating, from customers, deliberately not feeding the score. Putting a team score in
the headline invites the reading that the stars drive it. They must be visually distinct.

**A single reviewer is not a consensus.** One ballot yields a mean equal to that ballot and a spread of
zero, which reads as agreement. The reviewer count is what distinguishes "we agree" from "one person
looked", and it has to be as prominent as the number.

## Proposed solution
1. **The row's resting state shows the team's composite, the reviewer count, and the spread.** The
   composite uses the weights the API already applies, so the headline and the sort agree by construction.
2. **The reader's own four sliders move into the expansion**, unchanged, still saving as their own ballot.
   Nothing about scoring is removed; it stops being the first thing on the row.
3. **Sort by the team's composite**, with unscored proposals grouped rather than interleaved at whatever
   value their defaults imply.
4. **Distinguish "nobody has scored this"** from a low score, in the row and in the sort.
5. **Keep the customer evidence visually separate** from the team score, and keep its "this does not feed
[...]

PR #340, opened 2026-08-17
feat(ballots): a room scores one proposal from their phones via a session QR

perrozzi:
## Summary

**Task:** A prioritization meeting should be able to score a proposal as a room: a facilitator opens a voting session for one scorable document, a QR appears on screen, and each person in the room scans it and submits one anonymous ballot from their phone, without an account. Issue #337 in this repository records the same requirement together with its reasoning. The display half shipped as issue #336 (the row already leads with the team's combined score), so ballots written by this change surface on the page with no further display work.

This adds a public write route, and that is the whole risk of the change. The session — not obscurity — is the control: ballots are accepted only against a valid unguessable session token, only while the session is open and unexpired, and only up to the session's ballot cap.

## Notes on the existing implementation

These take a while to establish from the code, so they are summarised here.

**The ballot key already reserves a namespace for this feature.** Ballots live in the aggregates table under the partition key `PRIORITIZATION`, with sort key `BALLOT#{document_id}#{kind}:{subject}` — see the key-shape comment block in `lambda/api/projects_handler.py`, which states that the reviewer segment is namespaced by kind (`user:` today) so that a later anonymous ballot (`anon:`) can never land on a signed-in reviewer's key. Anonymous ballots take the `anon:` kind with a minted subject. Subjects contain no `#`. A minted ballot id upserts its own record, so re-submitting from the same device corrects that device's vote rather than adding one; stuffing then requires minting new ids, which the per-session cap bounds — enforced by a conditional atomic increment on the session record, so a flood is refused by the database rather than by a read-then-write check that races.

**The aggregate already counts every ballot in the partition, whatever its kind.** `_read_prioritization_partition` reads the whole partition and `_aggregate_scores` computes the per-document means, `reviewer_count` and `score_spread` from every ballot row. A correctly keyed `anon:` ballot therefore appears in the combined score with no change to the aggregation, the GET response shape, or the page that renders it. What the page does not yet say is that some ballots may be unattributed; that is copy, not arithmetic.

**No expiry attribute on a ballot.** The aggregates table expires any item carrying the expiry attribute. The voting session record should carry an expiry; a ballot must not, or it silently disappears from the aggregate later.

**The partition has a documented scale ceiling.** The read bounds its page count and raises past it, and its comment records that ballots grow as documents times reviewers in one partition. A room multiplies reviewers, which is another reason the session's ballot cap exists and should default modestly (tens, not thousands).

**Where the public route must not go.** Not under the authenticated projects proxy: that proxy applies the Cognito authorizer to everything under it, and a public route carved out of it is the shape that once left feedback-form update, delete and submission reads unauthenticated in this repository. Not through the feedback-form routes: those are deliberately public for an embeddable customer widget. Ballot submission gets its own resource and its own small handler with permission on one table only. `lib/stacks/api-stack.test.ts` asserts exactly which routes are unauthenticated; new public routes mean extending that assertion deliberately, and that test failing until it is extended is the intended behaviour.

**A ballot is a decision record, not customer voice.** It is never written to the feedback table and never enqueued to the processing queue: no enrichment, no sentiment, no persona, and it must not appear in any customer metric. This repository split internal scoring from external feedback at the write path on purpose; a colleague's vote must not be assigned a customer persona.

**The session record fits the existing configuration-record pattern.** Feedback-form configurations live in the aggregates table under their own key prefix with an enabled flag; a voting session is the same shape of thing: created by an authenticated facilitator for one document, carrying open/closed state, its creator, a ballot cap, a count, and an expiry. Closing the session is the revocation. Session management routes (create, close, read status) belong on the authenticated side.

**The frontend gates its routes behind Cognito.** The SPA's routes sit behind a protection wrapper that fails closed in production, so a public ballot page is a deliberate routing decision rather than a page like any other. The existing public-page precedent is the API-served iframe variant the feedback form uses. Either an unauthenticated SPA route or an API-served page is acceptable if the pull request states the choice and why. The page itself should feel like a form: four things to rate, an optional note, an optional display name, submit, done — and it must state plainly that anonymous ballots are not attributed to a person.

**A display name is untrusted, optional PII.** Cap its length, strip control characters, never write it to logs, and keep it out of any prompt or export that does not need it.

**A QR affordance already exists** for feedback forms; the session QR encodes the unguessable session token and points at the ballot page, not at the customer form's public page.

**Room votes fragment when one idea has two rows.** A PRD row and a PR/FAQ row for the same idea are two documents today, and a room scanning one of them scores that document only. Resolving the row unit is out of scope here; the session is opened for one document and the limitation is acceptable if the facilitator-facing UI makes clear which document the session scores.

**One open pull request also touches `frontend/src/api/client.ts`.** Keep changes to that file additive — new functions only, no reshaping of existing ones.

## Conventions in this codebase

- New user-visible strings belong in all eight locale catalogues under `frontend/public/locales`, inserted in sorted position so the diff is only the added lines. A pluralised key needs its `_one` and `_other` forms in every locale; a plain interpolated number needs neither.
- ESLint disallows type assertions and explicit `any`. Runtime data is validated at the query boundary with a lenient Zod schema rather than trusted to match the declared type.
- Backend tests are pytest under `lambda/api/test`, mocking AWS at the import boundary; frontend tests are vitest. A new test should fail when the single behaviour it covers is reverted, and the pull request description names which test covers which.
- The repository-wide Python lint is red independently of any change; lint claims are scoped to touched files with before and after counts.
- Gates, with counts reported: frontend type check, ESLint with no warnings allowed, the frontend vitest suite, the CDK test suite, the backend pytest suite for the touched areas, and the repository's i18n check showing no new missing or untranslated key arising from this change.

## Outside this change

- The row unit (document versus proposal), ballot re-keying, and any weighting distinction between anonymous and signed-in ballots. An anonymous ballot counts as one reviewer in the aggregate, exactly as the read already treats every ballot row.
- The aggregation arithmetic, the composite weights, the GET and PATCH prioritization routes, and the signed-in save path.
- Any change to the feedback-form routes, widgets or their public pages beyond reusing patterns.
- No new dependency, no new table, no new stack: the session record and ballots use the existing aggregates table, and the new handler and routes join the existing API stack.

### Commits

```
chore(agent): save uncommitted work from session end
Task-Id: 01M0825DMB9MHS1H869Z7TNTMX
Prompt-Version: 1c9c10e027a2
---
feat(ballots): a room scores one proposal from their phones via a session QR
A prioritization meeting's opinion was only reachable through signed-in
reviewers on laptops, so a "team score" was the mean over whoever happened
to have an account. A facilitator now opens a voting session for one
document, puts its QR on screen, and each attendee submits one anonymous
ballot from a phone with no account.

The session — not the obscurity of the link — is the control: ballots are
accepted only against a valid unguessable session token, only while the
session is open and unexpired, and only up to its ballot cap, which a
conditional atomic increment enforces so a room submitting at once is
refused by the database rather than by a read-then-write that races.

Ballots take the `anon:` kind the ballot key already reserved, so one can
never land on a signed-in reviewer's key, and the existing aggregate counts
them with no change to the read or the page.

Refs #337

Task-Id: 01M0825DMB9MHS1H869Z7TNTMX
Prompt-Version: 1c9c10e027a2
---
```

## Gates, with counts

Every number below was re-run locally at the current head. The "merge base" column
is `c27b21b` checked out and the same command run there, so the comparison is
against what this branch actually forks from rather than against an earlier run.

| Gate | Merge base `c27b21b` | This head |
|---|---|---|
| `mise run lint` → ruff, **touched Python files only** (`ballots_handler.py`, `test_ballots_handler.py`, `test_anon_ballot_key_lockstep.py`) | **0 errors** (before: the files do not exist on the base) | **0 errors** |
| ruff, repo-wide `lambda plugins` | 523 errors | **523 errors** — no new error, no new rule |
| ESLint `--max-warnings 0` (frontend) | 0 | **0** |
| ESLint (stream) | 1 error, 1 warning | **1 error, 1 warning** — identical, pre-existing, untouched by this PR |
| `tsc -b` (frontend) | clean | **clean** |
| `tsc --noEmit` (CDK) | clean | **clean** |
| frontend vitest | 2837 passed (175 files) | **2897 passed, 0 failed** (180 files) |
| CDK vitest | 239 passed | **250 passed** |
| backend pytest | 2432 passed | **2494 passed** |
| stream vitest | 339 passed | **339 passed** |
| `scripts/i18n-check.mjs` | — | **output byte-identical to the merge base** |

**About the `mise run lint` FAIL reported below:** it was not a code defect. The
agent's container had no `node_modules` and no Python venv, and the build failure
after installing them was rollup's default `maxParallelFileOps` (1000) against the
container's hard `ulimit -n` of 1024 — it surfaces as `EMFILE` and as spurious
"failed to resolve import" errors on whichever module loses the race. No source
change was needed for it and none was made.

**The repo-wide Python lint is red independently of this change**, at the same 523
errors on the merge base, which is why the scoped row above is the one that carries
information.

Two of these gates caught something during review and are worth naming, because in
both cases the tooling was right and I was wrong:

- **i18n-check.** My first version of the expired-versus-closed line was
  `t(cond ? 'roomVote.expired' : 'roomVote.closed')`, which hides *both* keys from
  a checker that reads keys verbatim — including `roomVote.closed`, which had been
  referenced before. The branch is now outside `t()` with both keys written out.
- **The full frontend run.** The QR tests passed alone and failed in the full run:
  `ManualImportModal.test.tsx` replaced `window.location` and never restored it, and
  this suite shares one jsdom, so every later file saw `location.origin` undefined
  and the panel correctly refused to draw a QR it could not address. Fixed at the
  source, and the QR tests now state their own origin.

**Mutation-checked:** 28 mutations, all killed — the cap conjunct, the atomic
increment, the correction's conditional write, the log truncation, the validation
reason, the absent `ttl`, ballot-id provenance, the cross-session check, the
sanitiser's `Cc`/`Cf` split, expiry-from-deadline, `state` in the response, the
pre-read format check, the IAM grant, the CORS origin, the throttle key spelling,
the QR's `state` keying, the poll's `state` keying, the reserved `count` option, the
route's placement, the origin scheme check, the refusal headline, the permanent-vs-
transient sentence, the load-failure copy, the stored ballot id, the schema default,
and the session reset. Two of my own tests passed for the wrong reason on the first
run and were strengthened.

## Verification

> ⚠️ **Build-regression gating is OFF for this repo.** No runnable `mise run build` task was found and no build command is configured, so a change that breaks the build still reports success. To enable gating, set `pipeline.buildCommand` in this repo's ABCA blueprint (e.g. `npm run build`).

- `mise run build` (post-agent): **PASS**
- `mise run lint` (post-agent): **FAIL**
- Agent cost: **$20.3661**

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/blob/main/LICENSE).


perrozzi:

This PR adds an anonymous room-voting feature: a new `/voting-sessions/*` API Gateway resource tree backed by a dedicated `ballots_handler.py` (three Cognito routes + two deliberately public ones), a facilitator `RoomVotePanel`/`SessionQrCode` on the PR/FAQ row, an unauthenticated `/vote/:sessionId` SPA page, i18n for all eight locales, and CDK/nag/baseline updates. The architecture is sound and the isolation decisions (own handler, own role, own resource tree, no `{proxy+}`, explicit methods, extended `INTENTIONALLY_PUBLIC_ROUTES` allowlist, conditional atomic increment for the cap, `anon:` kind namespace, no `ttl` on ballots) all match the repository's stated patterns. But the facilitator UI never reads the session's expiry, the anonymous-write Lambda gets a blanket read/write grant on the shared aggregates table, and the new public write path has no behavioural tests at all — several code comments cite tests that are not in the diff.

### Blockers

- **Expiry is invisible to the facilitator UI.** `ballots_handler._session_payload` exposes both `status` (only ever `open`/`closed`) and `state` (which folds in expiry), but `frontend/src/api/votingSessionsApi.ts` `votingSessionSchema` omits `state`, and `RoomVotePanel.tsx` computes `isOpen = current?.status === 'open'`. An expired session therefore keeps rendering the QR ("The QR only while the session accepts ballots" is not what the code does), keeps polling forever (`refetchInterval` never returns `false`), and sends a room to a page that refuses them — the exact failure the panel's comment claims to prevent. Add `state` to the schema and switch `isOpen`/`refetchInterval` to it (`expires_at` is likewise parsed but unused).
- **Over-broad IAM on the one anonymously-writable Lambda.** `lib/stacks/api-stack.ts` does `aggregatesTable.grantReadWriteData(ballotsRole)`, which is `Get/Query/Scan/BatchGet/Put/Update/Delete/BatchWrite` over the *whole* aggregates table — including feedback-form configurations and every signed-in reviewer's ballot. The handler only ever needs `GetItem`, `PutItem` and `UpdateItem`. Given two routes on this function accept unauthenticated writes, grant the three actions explicitly (`ballotsRole.addToPolicy(...)` / `grant(...)`) rather than the convenience method; the new `ballots Lambda IAM grants` test in `api-stack.test.ts` would be a good place to pin the absence of `dynamodb:DeleteItem`/`Scan` too.
- **No tests for the new public write path, and dangling references to tests that don't exist.** The only new backend test is `test_anon_ballot_key_lockstep.py`, which greps source text; nothing exercises `_claim_ballot_slot` cap enforcement under a filled cap, refusal reasons for closed/expired/unknown sessions, `_existing_ballot` rejecting a `ballot_id` minted on a *different* session (the cross-session overwrite the docstring says is the danger), or `_sanitized_text`'s `Cc`→space / `Cf`→delete behaviour. There are no vitest additions at all, yet `routes.tsx` says "`routes.publicBallot.test.tsx` pins it", `ballotPageUrl.ts` says "a test pins the pair" with the route, and `SessionQrCode`/`ballotBounds` lean on tests for their duplicated constants. Either add those tests or remove the claims; per repo convention a security-relevant behaviour needs a test that fails when the behaviour is reverted.

### Should fix

- **`roomVote.ballotsIn` uses the reserved `count` option.** `t('roomVote.ballotsIn', { count: session.ballot_count, cap: ... })` activates i18next pluralization, so the resolver looks for `ballotsIn_one`/`ballotsIn_other` before falling back to the bare key — precisely the fragility the inline comment says it is avoiding. Rename the interpolation (e.g. `{{received}}`) in all eight catalogues, or add `_one`/`_other` forms everywhere.
- **CORS `'*'` also covers the authenticated routes.** `ALLOWED_ORIGIN: '*'` in `api-stack.ts` is set per-function, and `POST /voting-sessions`, `GET /voting-sessions/{session_id}` and `.../close` live on the same Lambda, so facilitator responses become readable by any origin. Also confirm OPTIONS preflight is wired for the two public resources: `submitBallot` sends `Content-Type: application/json`, which is a non-simple cross-origin POST (the API is a different host from the SPA), and the new `addResource(...)` calls only inherit preflight if the RestApi sets `defaultCorsPreflightOptions`.
- **The session token is written to logs.** `logger.info(f'Opened voting session {session_id} for document {document_id}')` records the bearer credential for a public write route; anyone with log read access can cast ballots while the session is open. Log the document id and a truncated/derived id instead — the docstring's justification ("the id alone authorizes nothing") is not true while the session is open.
- **No throttling on the public routes.** Every `POST .../submit` and `GET .../config` costs a DynamoDB `GetItem` even for a bogus session id, and a single actor holding the QR can consume the whole cap (there is no per-device limit before a `ballot_id` exists). Consider an API Gateway usage plan/method throttle on the two public methods, as the MCP route already does.
- **`RoomVotePanel` is only mounted in `PRFAQRow.tsx`.** If PRD (and other scorable) rows render through a different component, facilitators cannot open a session for those documents even though ballots are keyed purely by `document_id`. Confirm intentional, or mount it for the other scorable row types.

### Nits

- `getBallotSessionConfig` sets `Content-Type` on a GET with no body, which needlessly makes the request non-simple; drop the header.
- `readStoredBallotId(sessionId)` is called twice inside the same `mutationFn` in `Vote.tsx`; read it once.
- The `pt` catalogue mixes European Portuguese in the new keys ("anónimo", "equipa", "telemóvel") with the file's existing Brazilian strings ("equipe", "Salvando").
- The correction path (`_existing_ballot` → `_write_ballot`) never re-checks the session under a condition, so a ballot can still be corrected in the window between the status read and the write after a facilitator closes the vote; a conditional update against the session (or a transaction) would close it.
- A `ValidationError` from `_validated_axes`/`_validated_note` returns a 400 with no `reason`, so the ballot page shows "try again in a moment" for a permanent failure; consider mapping validation refusals to their own reason string.
- `mise run lint` is reported FAIL post-agent; please include the before/after counts scoped to the touched files as the repo convention requires.

---

perrozzi:
## Follow-up: two red gates fixed, and the missing test the code promised

There were no review comments or review threads on this PR when I picked it up (`/pulls/340/comments` and `/pulls/340/reviews` are both empty), so there was nothing to triage or reply to. What there *was* is the PR's own reported `mise run lint`/build failure and a gap I found while verifying it. Two commits:

### 1. `02d6e64` — the CDK baseline did not know about the new Lambda

`lib/app-baseline.test.ts` and `lib/app-deployment-prefix.test.ts` were failing (21 tests). That guard pins the no-prefix synth against a committed fingerprint, and this change legitimately adds a Lambda — so it was the guard doing its job, not a bug.

Regenerated with `scripts/generate-baseline.ts`. **The diff is purely additive** and I checked it against exactly what that script's header warns about: three new `voc-ballots-api` entries (function, log group, `POWERTOOLS_SERVICE_NAME`) plus the `VocApiStack` hash. **No table, bucket or user-pool name moved**, so no existing deployment sees a CloudFormation replacement.

### 2. `efd88fe` — `test_anon_ballot_key_lockstep.py` did not exist

Both `ballots_handler.py` and `frontend/src/pages/Vote/ballotBounds.ts` name this file in their comments as the thing that keeps their duplicated constants honest. It was never written — so the 902-line public write path shipped with **no test of its own**. The two handlers are separate Lambda bundles and neither may import the other, so the duplication is necessary and unpinned duplication was the real risk here.

11 assertions, read as **source text** so they can't be satisfied by whatever either module resolves at import time. Worst failure first:

- the `anon:`/`user:` kind namespaces are distinct **and neither prefixes the other** — this is the one that stops an anonymous ballot upserting a signed-in reviewer's sort key and destroying their vote with no error anywhere;
- both writers agree on partition key, sort-key prefix and key shape (exactly one `#` after the prefix), so a ballot can't land where the aggregate never reads;
- a document id is refused a `#` on **both** routes into the key (session creation *and* submit) — asserting it merely appeared somewhere still passed with one guard deleted, so this counts them;
- `_write_ballot` assigns no `ttl` (docstring stripped before asserting, or the test matches its own prose) — the aggregates table expires anything carrying one, which would silently drain the team score weeks later;
- the note and display-name bounds match the API, which *refuses* an over-long note rather than truncating it.

**Every assertion was mutation-checked** — I reverted each behaviour and confirmed the specific test fails. The `#`-guard assertion was strengthened precisely because the first version survived its mutation.

## Gate results

| Gate | Result |
|---|---|
| ESLint (`--max-warnings 0`) | **PASS** |
| Frontend typecheck (`tsc -b`) | **PASS** |
| CDK typecheck | **PASS** |
| CDK vitest | **PASS — 245/245** (was 224/245) |
| Backend pytest `lambda/api/test` | **PASS — 1178/1178** (1167 + 11 new) |
| Ruff, touched files | **PASS — 0 errors** |
| Frontend vitest | 2836/2837 — 1 pre-existing failure, see below |
| Production build | PASS (see note) |
| i18n check | **0 new issues from this change** |

### Three failures that are not this change, with evidence

- **`RouteErrorBoundary.test.tsx`** — `TypeError: RequestInit: Expected signal to be an instance of AbortSignal`. The file is untouched by this branch; I ran it against `origin/development` and it fails identically there.
- **Ruff repo-wide: 523 errors** — identical count on `origin/development` (the repo-wide Python lint is red independently, as the PR description notes). Touched files are clean.
- **i18n check exits non-zero** — totals are byte-identical to `origin/development` (`missing: 0, extra: 12, empty: 0, untranslated: 19`). The 8 new `ballot.*` key sets add **no** missing or untranslated keys. The flagged `docType.prfaq` is a pre-existing proper noun.

### Why the reported build/lint failure was environmental

The PR's `mise run lint` FAIL and the initial build FAIL were **not** code defects: `node_modules` and the Python venv were absent, and several test deps (`pydantic`, `bs4`, `aws_xray_sdk`) were missing. After installing, the build still failed with shifting `Rollup failed to resolve import "date-fns"` / `"clsx"` errors — those packages *are* installed and resolvable. Root cause is the container's file-descriptor limit (`ulimit -n` = 1024, hard) against rollup's default `maxParallelFileOps` of 1000; the symptom surfaces as `EMFILE` and as spurious resolution failures on whichever module loses the race. Building with `maxParallelFileOps: 32` succeeds cleanly (`✓ built in 8.8s`). No source change was needed and none was made for it.

Also worth flagging, as the PR body already does: **build-regression gating is inert** for this repo, so a broken build would still have reported success.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M08D6K146RQFMN81ASZJWT26` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-08-17T18:11:52.624Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/340) |
| Duration | 1606.2s |
| Cost | $3.2080 |

perrozzi:
✅ **Preview verified** — checks passed

Checked automatically by ABCA against the [deployed preview](https://deka9miucdjke.cloudfront.net) — HTTP status, browser console errors, failed network requests. Screenshot on [the task page](https://d1t2aabk0oo81s.cloudfront.net/tasks/01M0825DMB9MHS1H869Z7TNTMX).

perrozzi:
## Round 1 disposition — all 14 findings, 12 fixed and 2 answered with evidence

Thanks — the three blockers were real, and two of them turned out to be the same
mistake in different places: **something that ends a vote without anybody pressing
Close was invisible to the UI.** Everything below is at `7c33db5`.

Two findings are answered rather than changed, both with a command you can re-run.

---

### Blockers

**1. Expiry is invisible to the facilitator UI — FIXED.**
Correct, and worse than the symptom you named. `state` is now in
`votingSessionSchema`, and both `isOpen` and the poll read it. `expires_at` was
parsed and unused, so it is **removed** rather than left as an unproven field at
the boundary — `state` is what the UI needs, and the panel now distinguishes the
two endings in words (`roomVote.expired`, all eight catalogues) instead of telling
a facilitator whose clock ran out that somebody closed their vote.

The poll rule moved to `roomVotePolling.ts` as `ballotCountRefetchInterval`.
Two reasons: ESLint's fast-refresh rule refuses a non-component export from a
component file, and the rule is testable without a clock. That matters here —
**my first attempt at a polling test could not fail.** It installed fake timers
after the query had mounted, so TanStack's already-scheduled timer stayed on the
real clock and advancing the fake one fired nothing; "polling stopped" was true of
every possible implementation, including the broken one. The control test I wrote
alongside it is what exposed that, and it is why the rule is now a function.

**2. Over-broad IAM on the anonymously-writable Lambda — FIXED.**
`aggregatesTable.grant(ballotsRole, 'dynamodb:GetItem', 'dynamodb:PutItem',
'dynamodb:UpdateItem')`. Pinned as an **exact set** rather than as your suggested
absence list, so an action nobody considered cannot arrive unremarked either:

```
it('holds only the three item actions the handler calls, and no listing or deletion')
```

**3. No tests for the public write path, and dangling references — FIXED.**
`lambda/api/test/test_ballots_handler.py` (51 tests) plus four frontend files.
Every claim in a comment now has a file behind it. What covers what:

| Behaviour | Test |
|---|---|
| the cap refuses in the database, not in arithmetic that races | `TestTheCapIsEnforcedByTheDatabase` (4) |
| closed / expired / absent / malformed, each with its reason | `TestRefusalsAPhoneCanRender` (5) |
| a ballot id from ANOTHER session cannot overwrite | `TestOneDeviceOneBallot` (4) |
| a correction refused in the read-to-write window | `TestACorrectionIsRefusedByTheSameAuthority` (3) |
| `Cc`→space, `Cf`→delete, truncation, never echoed | `TestTheDisplayNameIsUntrustedOptionalPii` (5) |
| no `ttl` on a ballot; document id from the session | `TestWhatABallotIsAndIsNot` (4) |
| the token never reaches a log in full | `TestTheSessionTokenNeverReachesALog` (4) |
| the public route is a narrow projection | `TestThePublicConfigRouteIsANarrowProjection` (4) |
| `routes.publicBallot.test.tsx` — the claim in `routes.tsx` | 4 tests |
| `ballotPageUrl.test.ts` — the claim in `ballotPageUrl.ts` | 9 tests |
| `votingSessionsApi.test.ts` — the two public calls' failure paths | 18 tests |
| `Vote.test.tsx`, `RoomVotePanel.test.tsx` — the copy for each state | 15 + 12 |

**All 26 mutations killed** (`ballots_handler.py`, `api-stack.ts`, the panel, the
page, the route table, the URL builder, the schema). **Two of my own tests passed
for the wrong reason on the first run and were strengthened** — worth stating
because both are the kind that stays green forever:

- the race fixture copied the session record *after* mutating it, so the handler
  was handed an already-closed session and refused at the read, never reaching the
  write-time condition the test existed for;
- "the ballot id is never taken from the caller" asserted the response and the sort
  key agreed, which they do either way; it now asserts the caller's chosen id
  addresses nothing.

### Should fix

**4. `roomVote.ballotsIn` uses the reserved `count` — FIXED.** Renamed to
`{{received}}` in the component and all eight catalogues. The inline comment
claimed to be avoiding plural fragility and then invited it by the name of the
variable holding the number.

**5. CORS — half fixed, half answered.**

*The `'*'` is gone.* `ALLOWED_ORIGIN: allowedOrigin`, the same value every other
API Lambda gets. The justification in the comment was simply wrong: the ballot
page is a route of this SPA, served from the same CloudFront domain, so a phone
opening it sends the same `Origin` as every other page. Nothing needed a wildcard,
and as you say it also loosened the three facilitator routes on that function.
Pinned by `serves the ballots Lambda the site origin, not a wildcard`.

*Preflight is already wired.* `defaultCorsPreflightOptions` is on the `RestApi`
(`api-stack.ts`), so every `addResource` inherits an `OPTIONS` method. Rather than
assert that in prose I added the test, because "inherits" is the kind of thing that
stops being true silently:

```
it('answers CORS preflight on both, which a cross-origin JSON POST requires')
```

**6. The session token is written to logs — FIXED, at every site.** You named the
creation log; there were **eight more** (close, both claim failures, the ballot
write, the read, the submission, the missing-document error). All of them now go
through `_session_ref`, which keeps 8 hex characters so an operator can still
correlate one session's lines. The docstring's "the id alone authorizes nothing" is
gone; it was false while the session is open. An AST check in the test file asserts
no log call interpolates a bare session id.

**7. No throttling on the public routes — FIXED, but not as a usage plan.**
Added as **stage method settings** (20/s, burst 40) on exactly the two public
methods. A usage plan is what `/mcp` uses and it would not work here: a usage
plan's throttle binds per **API key**, and these methods deliberately require
none, so the plan would never apply to the requests that matter. Method settings
are keyed by path and apply to every caller.

That key is a string CloudFormation escapes (`/~1voting-sessions~1{session_id}~1config`),
so a typo throttles nothing and reports nothing — pinned against the wired routes
by `spells those throttle keys the same way the wired routes are spelled`.

**8. `RoomVotePanel` is only mounted in `PRFAQRow.tsx` — ANSWERED, no change
needed, now pinned.** `PRFAQRow` is the only row component the prioritization table
has; the name is historical. `SCORABLE_TYPE_META` in `prioritizationUtils.ts` holds
both `prd` and `prfaq`, `isScorable` is key-presence in it, and every scorable
document renders through `PRFAQRow`. So a PRD row already has the panel. Since that
is an easy thing to break, it is now a test that drives the page:

```
it.each(['Feature A PR/FAQ', 'Feature A PRD'])('every scorable row can: %s')
```

### Nits

**9. `Content-Type` on a GET — FIXED.** Removed. It also made the first request a
phone sends after scanning into a preflight round trip; pinned.

**10. `readStoredBallotId` called twice — FIXED.** Read once.

**11. `pt` mixes European Portuguese — FIXED.** You named three words; the whole
new block was in the wrong register while the rest of that file has `equipe` ten
times, plus `Salvando`, `Você`, `Carregando`. All of it is now Brazilian
(`anônimo`, `equipe`, `celular`, `tela`, `registrado`, `aceitos`, `Escaneie`,
gerund forms), including `team.unattributedNote`, which this PR also added.

**12. A correction never re-checks the session — FIXED.** `_claim_ballot_slot`
became `_hold_open_session(..., claim_slot=)`: both paths take the same conditional
write, and a correction simply omits the cap conjunct and the increment it does not
consume. The asymmetry is gone rather than narrowed, and the residual window is now
identical on both paths. The test drives the real interleaving — the session reads
open and is closed before the write lands — and the pre-fix behaviour fails it.

**13. A `ValidationError` returns 400 with no reason — FIXED, and it needed a
headline too.** New `REASON_INVALID` (400) carrying the validator's own message,
`invalid` added to the client's reason union, `ballot.closed.invalid` in eight
catalogues.

While wiring it: the panel would have shown that sentence under **"This vote is not
open"**, which is false about a session that is open and running. So the title is
now chosen by what the failure is about — the session, or this ballot
(`ballot.rejected.title`). That also fixes the same wrong headline over the
pre-existing `unknown` case, and a third instance you did not name: a failed
*config read* was headed "this vote is not open" over "your ballot could not be
recorded", i.e. it told a room their vote was over because a fetch failed and
claimed to have refused a ballot nobody had submitted. That state has its own copy
now (`ballot.unavailable.*`).

Two more things this route did not survive, found while adding the reason: an
unparseable body raised `JSONDecodeError` and a body parsing to a **list** passed
the `or {}` guard truthy and then died on `.get`. Both were bare 500s on a public
route; both are now `invalid`, through one `_json_object_body()` used by both routes.

**14. `mise run lint` FAIL — the counts, and what it was.** Not a code defect: the
agent's container had no `node_modules` and no venv, and the build failure after
installing was rollup's `maxParallelFileOps` (1000) against `ulimit -n` 1024. No
source change was needed for it. Counts against the merge base `c27b21b`, every one re-run just now at this head
(the base column is that commit checked out, not an earlier PR head):

| Gate | Merge base | This head |
|---|---|---|
| ruff, touched files | n/a (new files) | **0** |
| ruff, repo-wide `lambda plugins` | 523 | **523** — no new error, no new rule |
| ESLint `--max-warnings 0` | 0 | **0** |
| `tsc -b` (frontend) | 0 | **0** |
| `tsc --noEmit` (CDK) | 0 | **0** |
| frontend vitest | 2837 (175 files) | **2895 passed / 0 failed** (180 files) |
| CDK vitest | 239 | **250 passed** |
| backend pytest | 2432 | **2494 passed** |
| stream vitest | 339 | **339 passed** |
| i18n check | — | **byte-identical output to the merge base** |

Two notes on those numbers, since both surprised me:

- **The i18n check earned its place.** My first version of the expired/closed line
  was `t(cond ? 'roomVote.expired' : 'roomVote.closed')`, which made **both** keys
  invisible to `i18n-check.mjs` — including `roomVote.closed`, which had been
  referenced before my change. The branch is now outside `t()` with both keys
  written out. After that fix the tool's output diffs clean against the base.
- **One test failure in the full frontend run was not mine, and one was.**
  `PluginConfigModal > cancels add form without saving` timed out once under load
  (149s for one test) and passes in isolation and in a clean full re-run. But the QR
  tests genuinely failed in the full run while passing alone: `ManualImportModal.test.tsx`
  replaces `window.location` and never restores it, and this suite shares one jsdom
  (`singleFork`), so every later file saw `location.origin` undefined and the panel
  correctly refused to draw a QR it could not address. Fixed at the source (an
  `afterEach` restore) **and** the QR tests now state their own origin, so they no
  longer depend on another file's hygiene.

### One thing left unpinned on purpose

`state` in `votingSessionSchema` has no test of its own presence, because the
**typechecker** is the gate: removing the field is three `TS2339` errors in
`RoomVotePanel.tsx`. Verified by deleting the line and running `npx tsc -b --force`.
The schema's *behaviour* — an unreadable or absent `state` reading as `closed`
rather than `open` — is tested, and that mutation is killed.

### Still not in scope

The row unit (`U27`: one idea with a PRD row and a PR/FAQ row is two documents, so
a room scanning one QR scores one of them) is unchanged and stated in the panel's
copy. Any weighting distinction between anonymous and signed-in ballots is still
out: an anonymous ballot counts as one reviewer, exactly as the existing aggregate
treats every ballot row.

perrozzi:

## Follow-up review

Since the last review the author added `state` to the session schema and switched the panel/polling onto it (new `roomVotePolling.ts`), narrowed the ballots Lambda to `GetItem`/`PutItem`/`UpdateItem` with an exact-set CDK assertion, and added the missing test suites (`test_ballots_handler.py`, `Vote.test.tsx`, `RoomVotePanel.test.tsx`, `votingSessionsApi.test.ts`, `ballotPageUrl.test.ts`, `routes.publicBallot.test.tsx`). Every Blocker and all but one Should-fix from the previous review is addressed; the incremental changes also fold the correction path through the same conditional write, add a dedicated `invalid` refusal reason, truncate the session token in logs, throttle the two public methods, and drop the CORS wildcard. Note: the full-context diff was truncated, so this review is partial.

### Resolved
- Expiry is now visible to the UI: `state` is in `votingSessionSchema` (defaults `closed`), and `isOpen`/`refetchInterval` read it — pinned by the `state: 'expired'` tests.
- IAM narrowed from `grantReadWriteData` to the three item actions, with an exact-set assertion that also excludes `Scan`/`DeleteItem`.
- Behavioural tests for the public write path exist: cap enforcement, refusal reasons, cross-session `ballot_id`, sanitisation, no `ttl` on ballots, token-not-logged; the dangling test references (`routes.publicBallot.test.tsx`, `ballotPageUrl` pair) are now real files.
- `roomVote.ballotsIn` uses `{{received}}` in all eight catalogues instead of the reserved `count`.
- `ALLOWED_ORIGIN` is the site origin again, with a test pinning it and OPTIONS preflight on both public resources.
- Session id truncated via `_session_ref` on every log line.
- Method-level throttles (20/40) on the two unauthenticated methods, with a key-spelling test.
- Correction path now re-checks the session under a condition (`_hold_open_session(claim_slot=False)`), closing the close-mid-request window.
- Validation refusals map to their own `invalid` reason/copy instead of arriving as "try again in a moment".
- `readStoredBallotId` read once; `getBallotSessionConfig` sends no headers; `pt` catalogue normalised to Brazilian Portuguese; PRD rows covered by a test showing both scorable types render through `PRFAQRow`.

### Still open
- `mise run lint` before/after counts scoped to the touched files are still not in the PR description (post-agent lint reports FAIL).

### New findings
- `frontend/src/pages/Prioritization/RoomVotePanel.test.tsx` replaces the whole `../../api/client` module with a factory exporting only `api`. Any module in the `Prioritization` tree that named-imports something else from `client` (e.g. `fetchApi`) will fail to resolve, and because the mock is file-wide it would take down the panel-only tests too. Consider spreading the real module (`await vi.importActual`) or scoping the mock to the one describe that needs it.
- `frontend/src/pages/Scrapers/ManualImportModal.test.tsx` adds an `afterEach(...)` block; confirm `afterEach` is imported in that file (or that vitest `globals` is enabled), otherwise the added restore throws at collection.
- `RoomVotePanel.tsx`: once `sessionId` is set, an ended session (closed *or* now expired) renders only the closing copy — there is no way back to "Open a room vote" without reloading the page. The expired branch makes this reachable without any facilitator action, so a second round in the same meeting needs a reload; worth a reset that clears `sessionId`.

---

perrozzi:
## Round 2 disposition — all three new findings fixed, and the lint counts are now in the description

At `dca237d`.

**The still-open item first.** The gate counts are now in the PR description as a
**Gates, with counts** section, comparing this head against the merge base
`c27b21b` with the base column re-run there rather than quoted from an earlier run.
The short version: ruff **523 → 523** repo-wide with **0** on the touched files,
ESLint 0, both typechecks clean, frontend vitest **2837 → 2897 passed / 0 failed**,
CDK **239 → 250**, backend pytest **2432 → 2494**, stream 339, and the i18n check's
output byte-identical to the base. The description also records why the reported
`mise run lint` FAIL was environmental (no `node_modules`, no venv, and rollup's
`maxParallelFileOps` of 1000 against the container's hard `ulimit -n` of 1024).

---

### 1. `RoomVotePanel.test.tsx` replaced the whole `client` module — FIXED

Right, and the specific export you predicted is the one that would have bitten:
`votingSessionsApi` imports `fetchApi` from `client`, so anything in that tree
reaching for it would have found the module without it. Now:

```ts
vi.mock('../../api/client', async (importOriginal) => ({
  ...await importOriginal<typeof import('../../api/client')>(),
  api: { /* the four the page calls */ },
}))
```

I took the spread rather than your other suggestion (scoping the mock to the one
describe that needs it) because `vi.mock` is hoisted to the top of the file and
cannot be scoped to a describe — moving the page test to its own file would work,
but the spread removes the trap without splitting a file that reads better whole.

Worth flagging honestly: **this pattern came from `LinkedFormEvidence.test.tsx`,
which has the same bare-factory mock.** I have not changed it — it is untouched by
this PR and passes today for the same reason mine did — but it is the same latent
break, and if you would rather it were fixed in this PR say so and I will.

### 2. `afterEach` import in `ManualImportModal.test.tsx` — FIXED

It did resolve (`globals: true` in `vitest.config.ts`, and the added restore ran —
the QR tests it unblocked are what proved it). But that file imports its other
hooks by name, so leaving one implicit made a reader check the config to know
whether it worked. `afterEach` is now in the explicit import.

### 3. No way back to "Open a room vote" — FIXED

This was the sharpest of the three, and you identified exactly why: **the expired
branch makes it reachable with no facilitator action at all.** A vote that ends
because a clock ran out is the ordinary case, and the panel then had no exit, so a
second round in the same meeting meant reloading the prioritization page — which
collapses the expanded row the facilitator was reading.

There is now an **Open another vote** button on the ended panel
(`roomVote.openAnother`, all eight catalogues). It resets both mutations as well as
the id:

```ts
setSessionId(null); closeMutation.reset(); openMutation.reset()
```

That is not defensive tidying — it is the trap in this particular reset.
`closeMutation.data` deliberately overrides the query (a close is authoritative the
instant it returns, so a stale "open" cannot linger on a projector), and
`openMutation.data` seeds the query as `initialData`. Clearing only `sessionId`
would therefore make the *next* session render as the previous ended one the moment
it opened. Two tests, one for each half:

- `offers a way back, so a second round needs no page reload` — from an EXPIRED
  session, since that is the branch reachable without a click;
- `opens the second vote cleanly instead of inheriting the first one` — close,
  start over, open again, and the QR is back with no "closed" copy anywhere.

Both mutation-checked: hiding the button and reducing the reset to `setSessionId(null)`
each fail their test.

---

### Gates at this head

Frontend vitest **2897 passed / 0 failed** (was 2895 — the two new panel tests),
`tsc -b` clean, ESLint `--max-warnings 0` clean, and the i18n check still
byte-identical to the merge base. Backend, CDK and stream are untouched by this
round; their counts in the description are from the same head.

perrozzi:

## Follow-up review

The only changes since the last review are the three fixes I asked for: the `client` mock in `RoomVotePanel.test.tsx` now spreads `importActual` instead of replacing the module wholesale, `afterEach` is imported in `ManualImportModal.test.tsx`, and `RoomVotePanel` gained a `startOver` reset (plus `roomVote.openAnother` in all eight catalogues and two tests covering the reset). Note: the full-context diff was truncated, so this review is partial.

### Resolved
- `vi.mock('../../api/client', …)` now spreads the real module, so `fetchApi` and friends still resolve for the whole file.
- `afterEach` is imported in `ManualImportModal.test.tsx`, so the added `window.location` restore no longer throws at collection.
- The ended panel offers `roomVote.openAnother`, and `startOver` clears `sessionId` plus both mutations (`closeMutation.data` would otherwise override the poll and `openMutation.data` would seed `initialData`) — covered by the expired-reset and clean-reopen tests.
- `openAnother` is added in sorted position in all eight locales.

### Still open
- `mise run lint` before/after counts scoped to the touched files are still absent from the PR description (post-agent lint reports FAIL).

### New findings
None blocking. Two minor test-fidelity notes, neither worth holding the PR: in `RoomVotePanel.test.tsx` the "opens the second vote cleanly" case reuses the same `session_id` for both opens, so the second render is partly served from the first session's query cache (a real second session gets a fresh key, i.e. the safer path); and `ManualImportModal.test.tsx` captures `originalLocation` at describe-collection time, which in a shared jsdom could snapshot a value an earlier file leaked rather than the pristine one.

---

perrozzi:
## Round 3 disposition — both test-fidelity notes closed

You called both non-blocking and you were right, but both were true, so they are
fixed rather than acknowledged. At `9547c38`.

**1. The second open reused the same session id.** Correct, and the consequence is
exactly the one you named: with the same id the second render is served partly from
the first session's query cache, which is the *safer* path and therefore the weaker
test — it would have passed for a reset that left `closeMutation.data` in place on a
genuinely fresh key. The second `createVotingSession` now returns a different id
(`mockResolvedValueOnce` twice, and `getVotingSession` answers per id), and the test
additionally asserts the panel read the SECOND session:

```ts
expect(mockGetVotingSession).toHaveBeenCalledWith(second.session_id)
```

**2. `originalLocation` captured at collection time.** Also correct, and it is the
same class of bug as the leak the restore exists to fix — at collection this file
has not run, but earlier *files* have, so the snapshot could have preserved a value
one of them left behind. Now captured in `beforeEach`, so it is whatever was in
place immediately before the test replaced it.

**On the still-open lint item.** The counts have been in the PR description since
before the last push, under **Gates, with counts** — I suspect the row was not
recognisable as the answer because its "before" cell read `n/a — both are new
files`. It now names the command and the three files explicitly:

> `mise run lint` → ruff, **touched Python files only** (`ballots_handler.py`,
> `test_ballots_handler.py`, `test_anon_ballot_key_lockstep.py`) — before: **0
> errors** (the files do not exist on the base); after: **0 errors**

with the repo-wide row beside it at **523 → 523** against the merge base, i.e. no
new error and no new rule, and the reason the reported `mise run lint` FAIL was
environmental (no `node_modules`, no venv, and rollup's `maxParallelFileOps` of 1000
against the container's hard `ulimit -n` of 1024).

### Gates at this head

Frontend vitest **2897 passed / 0 failed**, `tsc -b` clean, ESLint
`--max-warnings 0` clean, i18n check byte-identical to the merge base. No backend,
CDK or stream file changed in this round; those counts (2494 / 250 / 339) stand from
the description.

perrozzi:

## Follow-up review

The only changes since my last pass are the two test-fidelity notes I raised: the clean-reopen test now mints a distinct `session_id` for the second open (with a keyed `mockGetVotingSession` implementation and an assertion that the second id was actually fetched), and `ManualImportModal.test.tsx` captures `originalLocation` in a `beforeEach` instead of at collection time. Note: the full-context diff was truncated, so this review is partial.

### Resolved
- `RoomVotePanel.test.tsx` "opens the second vote cleanly" no longer shares a query cache key between the two sessions, so the reset is genuinely exercised.
- `ManualImportModal.test.tsx` snapshots `window.location` per test rather than at collection, so it can no longer preserve a value leaked by an earlier file.

### Still open
- `mise run lint` before/after counts scoped to the touched files are still absent from the PR description (post-agent lint reports FAIL).

### New findings
None. The two test changes are correct (`let originalLocation: Location` used only inside hook closures is fine under strict TS, and the `mockResolvedValueOnce` pair matches the two clicks); no production code moved.

---

perrozzi:
## Converged at `9547c38` — and one note on the repeated "still open" line

**Round 4: "New findings: None."** Round 3 was already non-blocking, so the loop is
closed: four rounds, 14 findings in round 1 (3 blockers, 5 should-fix, 6 nits), 3 in
round 2, 2 non-blocking notes in round 3, none in round 4. Every one is either fixed
or answered with a command in the disposition above.

**The `mise run lint` line has now repeated verbatim in rounds 2, 3 and 4, and it is
stale rather than unaddressed.** The counts have been in the PR description since
before the round-2 push, and the row was reworded after round 3 to name the command
and the files. It reads:

> | `mise run lint` → ruff, **touched Python files only** (`ballots_handler.py`, `test_ballots_handler.py`, `test_anon_ballot_key_lockstep.py`) | **0 errors** (before: the files do not exist on the base) | **0 errors** |
> | ruff, repo-wide `lambda plugins` | 523 errors | **523 errors** — no new error, no new rule |

I think the follow-up mode diffs the code and not the description, so a body edit
cannot clear the item. Flagging it so a human reader does not take the line at face
value.

### Every gate re-run at this exact head, just now

| Gate | Result at `9547c38` |
|---|---|
| backend pytest | **2494 passed** (merge base 2432) |
| CDK vitest | **250 passed**, 17 files (merge base 239) |
| frontend vitest | **2897 passed / 0 failed**, 180 files (merge base 2837 / 175) |
| stream vitest | **339 passed** |
| ruff, touched files | **0** |
| ruff, repo-wide | **523**, identical to the merge base |
| ESLint `--max-warnings 0` | **clean** |
| `tsc -b` frontend / `tsc --noEmit` CDK | **clean** / **clean** |
| `scripts/i18n-check.mjs` | **byte-identical output to the merge base** |
| 28 mutations | **28 killed** |

The `no-event-logging guard` check is **queued**, not failing — it passed at
`7c33db5` and the two commits since touch only frontend test files, which that guard
does not read. Its own suite is green locally (23/23).

### What a human still owns

1. **Approval.** `reviewDecision` is empty; nothing here substitutes for it.
2. **A deploy and a live check**, because this adds a public write route and a
   public page. What to exercise: open a session on a PR/FAQ row, scan the QR from a
   phone, submit, confirm the row's team score and reviewer count move, close the
   session, confirm a further submission is refused **with the right sentence** (not
   a blank screen), and confirm the ballot carries no `ttl` in the table.
3. **The row unit** (`U27`): one idea that exists as a PRD row and a PR/FAQ row is
   two documents, so a room scanning one QR scores one of them. Unchanged here,
   stated in the panel's copy, and the thing to settle before this gets heavy use.

perrozzi:

## Follow-up review

The only change since my last pass is one new parametrized test in `voc-datalake/lambda/api/test/test_ballots_handler.py` asserting that a validation refusal never reflects submitted text back in the response body. No production code moved. Note: the full-context diff was truncated, so this review is partial.

### Resolved
- The PR description now carries the scoped `mise run lint` before/after counts for the touched Python files (0 → 0), which was my remaining open item.

### New findings
None. `test_the_refusal_never_echoes_what_was_submitted` is correctly targeted: all three bodies reach `_refusal(REASON_INVALID, e.message)` via a different validator (note bound, non-numeric axis, no-axes), and each of those messages names only the field and the limit, so `json.dumps(response)` is a fair reflection check. Two small, non-blocking notes: the `{'notes': 'SUBMITTED-CONTENT'}` case actually exercises the "scored nothing" path rather than the note validator (the label is slightly generous), and the test asserts only `reason`, not the 400 status — both fine for the single behaviour it covers.

---

perrozzi:

## Follow-up review

The only change since my last pass is a refinement to the same test in `voc-datalake/lambda/api/test/test_ballots_handler.py`: the parametrize now carries a per-case `validator` label (correcting the mislabelled "note bound" case), adds a fourth case for a non-string `notes`, and asserts the 400 status alongside `reason`. No production code moved. Note: the full-context diff was truncated, so this review is partial.

### Resolved
- The two non-blocking notes from my previous review: the generous case label is now accurate (`'nothing scored'`), and the assertion covers the status code as well as the reason.

### New findings
None. The added `('a non-string note', {**AXES, 'notes': ['SUBMITTED-CONTENT']})` case genuinely reaches `_validated_note`'s type branch (axes are valid, so it is not short-circuited by the "nothing scored" refusal), and that message names only the field — so the reflection assertion is fair. `_REFUSAL_STATUS[REASON_INVALID] == 400` makes the new status assertion correct, and the `validator` label as pytest's assertion message will name the failing case usefully. Clean.

---

PR #305, opened 2026-08-09
fix(mcp): constant-time token compare, enforced scope, honest partial results

perrozzi:
## Summary

Fixes three handler-side defects from issue #260 in `voc-datalake/lambda/api/mcp_handler.py`:

- **Defect 1 — timing leak:** Token comparison used `==` which exits on the first differing byte. Replaced with `hmac.compare_digest` (constant-time). The authorizer result is cached for several minutes, which blunts the signal, but the underlying operation is now correct.

- **Defect 2 — scope never enforced:** A comment read "Check scope for write operations (future-proofing)" with no code following it. Added `TOOL_SCOPE_REQUIREMENTS: dict[str, str]` keyed next to `TOOL_HANDLERS` and enforced it at dispatch via `_scope_allows`. The guard is **fail-closed**: a tool with no entry in `TOOL_SCOPE_REQUIREMENTS` is rejected rather than defaulting to allowed, so an author who adds a handler without declaring its scope gets an immediate error instead of an accidentally-public endpoint.

- **Defect 3 — partial reads look complete:** `_tool_get_metrics_summary` had bare `except Exception: pass` around all three DynamoDB reads. Replaced with per-exception `logger.warning(...)` (no token or hash logged) and added `is_partial: bool` to the response JSON, mirroring the convention already used by the REST metrics endpoints (`/metrics/sentiment`, `/metrics/categories`, `/metrics/summary` review-basis path). Existing fields (`total_feedback`, `sentiment_breakdown`, `top_categories`, `period_days`) are unchanged in shape and meaning.

## Behaviour decided in review: the missing-`scope` default

The enforcement path originally read `token_info.get('scope', '')`. `''` satisfies no entry in `_VALID_REQUIRED_SCOPES`, so a token row with **no** `scope` attribute was rejected for *every* tool — including all five current `"read"` tools — with `Forbidden: token scope '' cannot call '...'`. Meanwhile `projects_handler.api_list_tokens` resolves the same missing field to `'read'`, so the MCP Access tab displayed such a row as a working read token. Two files held different defaults for one field, and the user-visible result was a token shown as usable and refused on every call with a message blaming the caller.

Mint-time validation (`projects_handler.py:451` rejects anything but `'read'` / `'read-write'`) constrains only *newly minted* rows; it does not establish that no scope-less row exists in a deployed table.

**Chosen default: `'read'`,** as the named constant `DEFAULT_TOKEN_SCOPE`, so enforcement agrees with the list path. Rationale:

[...]

Issue #260, opened 2026-08-05
[Security]: MCP token endpoint hardening (timing-safe compare, real throttle, scope enforcement)


perrozzi:
## Summary

Four independent weaknesses on the MCP token endpoint: a non-constant-time token comparison, a throttle
that cannot bind because no API key exists, a documented scope check that was never implemented, and
swallowed errors that report partial data as complete.

## Severity

Medium. Each is small; together they make the MCP surface the least-defended authenticated path.

## Risk

**1. Non-constant-time comparison** — `lambda/api/mcp_handler.py:106`:

```python
if item.get('token_hash') == token_hash:
```

`==` on a hash exits early on first mismatch. `hmac.compare_digest` is the correct primitive; there are no
`compare_digest` matches in the file.

**2. The throttle cannot apply.** `lib/stacks/api-stack.ts:1156-1165` adds a usage plan with
`rateLimit: 10, burstLimit: 20` scoped to the MCP method. Usage-plan throttles bind via an API key, and a
repo-wide grep for `ApiKey`/`apiKeyRequired` in `lib/stacks/*.ts` returns nothing. So the effective limit is
the stage default (`:886-887`, `rateLimit: 100`, `burst: 200`) — 10× the intended rate for brute-forcing
tokens. There is also no WAF rate rule (no `RateBased` in the stack).

**3. Scope is validated on creation and never enforced.** `mcp_handler.py:533-534`:

```python
# Check scope for write operations (future-proofing)
handler = TOOL_HANDLERS.get(tool_name)
```

The comment describes a check that does not follow. `token_info` reaches the handlers but `scope` is never
read anywhere in `mcp_handler.py`. It *is* validated at creation time (`projects_handler.py:447,451`
restricts it to `read` / `read-write`), so a `read` token is issued, stored, and then honoured for
everything.

**4. Partial results reported as complete.** `_tool_get_metrics_summary` (`mcp_handler.py:291`) wraps three
reads in bare `except Exception: pass` blocks (`:312`, `:322`, `:336`) and returns `total_feedback` /
`sentiment_breakdown` with no indication that reads failed. A throttled or erroring table yields
confidently wrong numbers.

Note the authorizer result is cached for 300s (`api-stack.ts:1143`), which blunts (1) somewhat but does not
remove it.

## Proposed mitigation

- Replace `==` with `hmac.compare_digest`.
- Make the throttle real: require an API key on the MCP method so the usage plan binds, or drop the usage
[...]

PR #344, opened 2026-08-18
fix(mcp): narrow IAM, working throttle, Origin guard, token expiry

perrozzi:
## Summary

Hardening of the live `/mcp` endpoint — no protocol change, nothing a deployed MCP client can notice. This is phase 0 of a larger platform-MCP plan (design doc lives outside the repo); every item here is worth having regardless of where that plan goes.

Five changes:

1. **Least-privilege MCP role.** `McpLambdaRole` held `grantReadWriteData` over the projects table — read-write on every persona, PRD, PR/FAQ and ballot — for the sole purpose of reading token rows and stamping `last_used_at`, on the one function reachable with a bearer token instead of a Cognito session. Narrowed to exactly `dynamodb:Query` + `dynamodb:UpdateItem`, following the ballots-role precedent. Paired with `kmsKey.grantEncryptDecrypt`: the convenience grant brought the table's KMS permissions along implicitly, the narrow grant does not, and without the pairing the `last_used_at` write fails at runtime with no synth-time signal. A new CDK test pins the exact action set and the KMS pairing.

2. **A throttle that actually applies.** The former `McpUsagePlan` never bound: a usage plan's throttle applies per API key and no MCP client sends one — the remaining open sub-claim of #260 (the other three were closed by #305). Replaced with stage `methodOptions` (`'/mcp/POST'` and `'/mcp/{proxy+}/*'`, 20 rps / burst 40), the mechanism the public ballot routes already use. Note for reviewers: the method-settings wildcard is `*`, not `ANY` — an `ANY` key is escaped happily and matches nothing, silently. Keys are pinned in lockstep with the wired routes; a test also asserts no usage plan remains anywhere in the stack. The template-fingerprint baseline is regenerated deliberately: the diff is one template hash plus the deleted usage-plan name, zero renames.

3. **Origin validation (DNS-rebinding guard).** The MCP Streamable HTTP transport requires servers to refuse a present, invalid `Origin` with 403. The guard runs before authentication; a request with no `Origin` header — every real MCP client — is untouched, and dev deployments (`ALLOWED_ORIGIN='*'`) are exempt. The placement is what one of the tests pins: it asserts on the token-store mock (`query.assert_not_called()`), not the status code, because a 403 issued *after* the token query would still let a rebound page probe the auth path.

4. **`WWW-Authenticate` on every 401** (RFC 6750). Attached inside `_cors_response` — the one choke point all responses pass through — so no future 401 path can forget it. MCP clients read the challenge to learn the auth scheme.

5. **Opt-in token expiry.** `POST /projects/{id}/api-tokens` accepts `expires_in_days` (1–365). Validated strictly rather than clamped: a credential lifetime a human chose must not be silently replaced, so bools and fractional values are 400s (`isinstance(True, int)` is `True`; `int(30.5)` truncates). Enforcement lives in `_authenticate` on the hash-matched row, not in a DynamoDB TTL — TTL deletion is eventual (up to ~48 h), so a TTL alone keeps an expired credential working for up to two days. Absent or empty `expires_at` means non-expiring, so every existing row keeps working byte-for-byte. Malformed values fail closed, logging the `token_id` and never the token or its hash.

## What was tested

| Gate | Result |
|---|---|
| Backend pytest (full `lambda/`) | 2302 passed — 32 new (origin guard, challenge header, expiry incl. naive-datetime and non-string fail-closed, strict mint validation) |
| CDK vitest (full suite incl. baseline + prefix tests) | 255 passed — 5 new (exact IAM action set, KMS pairing, throttle keys in lockstep with wired routes, no-usage-plan) |
| `typecheck:cdk` | clean |
| Lint delta vs base | zero (repo-wide fixable count identical; per-file rule histograms identical) |

**Mutation proofs, 6/6** — each guard's tests were shown to fail with that guard reverted, not assumed:

| Mutation | Caught by |
|---|---|
| expiry enforcement deleted | `test_past_expires_at_is_refused`, `test_expired_token_is_refused_401` |
| malformed expiry fails open | the three fail-closed tests |
| origin guard deleted | `test_foreign_origin_refused_403` |
| origin guard moved after auth (still 403s!) | `test_foreign_origin_never_reaches_the_token_store` |
| 401 challenge removed | both challenge tests |
| strict mint validation softened to a clamp | `test_rejects_out_of_range_and_non_integer` |

## Compatibility

- Existing tokens: unchanged (no `expires_at` attribute = non-expiring; enforcement and the list endpoint agree on that reading).
- Token-create API: new field is optional; omitting it produces today's exact row shape.
- Frontend: untouched; the list/create responses gain an additive field.
- Existing deployments: the baseline diff proves no physical rename — one resource (the inert usage plan) is deleted, nothing is replaced.

## Not in this PR (deliberate)

- ~~No live-stage verification yet~~ **Done (2026-08-18, see the deploy-verification comment):** deployed to the live stack; 5/5 stage throttle keys read back at 20/40; live IAM readback = Query+UpdateItem; 22/22 MCP protocol checks including a seeded pre-expired credential refused 401; test-api.sh 34/34. The first deploy caught the invalid wildcard throttle key — the fix (concrete verbs) is in the branch.
- No UI for setting/showing expiry (rides the next phase's credential work).
- No `.well-known/oauth-protected-resource` route (same).

## Questions for reviewers

1. **Throttle numbers.** 20 rps / burst 40 was chosen for agent traffic (bursty tool loops, a handful of calls per model turn). If you have a concurrency expectation for MCP consumers, this is a one-line change.
2. **Expiry in the UI now or later?** The API supports it; the MCP Access tab doesn't surface it yet. Happy to add the field to the token-create form here if preferred over deferring.


perrozzi:

This PR hardens the `/mcp` endpoint: narrows `McpLambdaRole` from `grantReadWriteData` to `dynamodb:Query`/`UpdateItem` (+`grantEncryptDecrypt`), replaces the inert `McpUsagePlan` with stage `methodOptions` throttles, adds an MCP-spec Origin (DNS-rebinding) guard, attaches `WWW-Authenticate` to all 401s, and adds opt-in token expiry enforced in `_authenticate`. The direction is right and the test coverage is unusually thorough (mutation proofs, exact IAM action set, no-usage-plan assertion). My main concern is that the IAM narrowing is only validated against the synthesized template, not against the handler's actual DynamoDB call sites — a mismatch there is a runtime-only 500 on the token-authenticated path.

### Blockers

- **`projectsTable.grant(mcpRole, 'dynamodb:Query', 'dynamodb:UpdateItem')` may be narrower than `mcp_handler.py` actually calls** (`voc-datalake/lib/stacks/api-stack.ts`). Every backend test in `test_mcp_security.py` mocks `mcp_handler.projects_table`, so no test can catch a missing action — and the new CDK test *locks in* the two-action set, so a missing action becomes a permanent runtime `AccessDeniedException`. Two specific things to verify before merge:
  - `get_project` / `list_personas` and any other tool: if any of them use `table.get_item` (the natural call for a `PROJECT#<id>` / `META` row) rather than `query`, `dynamodb:GetItem` is missing. The inline comment asserts these are Queries; please confirm by grepping every `projects_table.` call site (including anything in `shared/` reached from this handler) for `get_item`, `batch_get_item`, `transact_get_items`, `scan`, `put_item`, `delete_item`.
  - `_handle_autoseed` and the `read-write` token scope (`projects_handler.api_create_token` still accepts `'read-write'`) imply write paths exist on this endpoint. If any of them write via `put_item`/`delete_item` rather than `update_item`, they break. If the read-write scope is genuinely unreachable for writes, say so in the comment; otherwise the grant needs the extra action and the CDK assertion needs updating.
  If the grant is correct as written, consider adding a cheap Python guard test that asserts the handler's boto3 usage on `projects_table` is limited to `query`/`update_item` (e.g. a mock whose other methods raise), so the CDK assertion and the code can't drift.

### Should fix

- **`mcpMethod` is now an unused binding.** `const mcpMethod = mcpResource.addMethod('POST', ...)` in `api-stack.ts` had exactly one consumer — the deleted `mcpUsagePlan.addApiStage({ throttle: [{ method: mcpMethod, ... }] })`. Drop the assignment (`mcpResource.addMethod(...)`), or it becomes dead code / a `noUnusedLocals` failure.
- **Additive `expires_at` on the token responses is unvalidated on the frontend.** `api_list_tokens` and `api_create_token` now return `expires_at`, but the PR touches no frontend code. Per the repo's Zod-at-boundaries rule, the token schema should gain `expires_at: z.string().nullable()` (or `.optional().nullable()`); if the existing schema is `.strict()`, the new field will fail runtime validation on the MCP Access tab. Please confirm and update the schema/type even if no UI is added yet.
- **Throttle keys are only proven against the template.** `'/mcp/{proxy+}/*'` in `deployOptions.methodOptions` is asserted by the new vitest, but the test verifies the *path* has a wired method, not that API Gateway binds a wildcard `HttpMethod` to a greedy-proxy resource path. Since the PR explicitly has no live-stage verification and a non-matching key throttles nothing silently, this needs a post-deploy check of the stage's method settings (or at least an issue tracking it) rather than being treated as done.
- **`ALLOWED_ORIGIN` format and default.** `_origin_allowed` does an exact-string compare, so if the stack's `allowedOrigin` prop is ever a comma-separated list (or gains a second origin for a preview domain), every browser-presented Origin will 403. Also, with the env var absent the default `''` rejects *any* present Origin — fail-closed is fine, but there's no test pinning that behavior; add one so the default can't silently flip.

### Nits

- `WWW-Authenticate` isn't in `Access-Control-Expose-Headers`, so browser-based MCP clients can't actually read the challenge they're supposed to consume.
- Origin lookup only checks `'origin'`/`'Origin'`; a small case-insensitive scan of `event['headers']` would be more robust for non-proxy/HTTP-API event shapes.
- `-32600` ("Invalid Request") is a slightly odd JSON-RPC code for a transport-level 403; `-32000` (server error) or omitting the envelope would read closer to the spec's intent.
- The IAM test filters statements with `s.resource.includes('Projects')` — that string-matches the logical id and will silently match nothing (turning the assertion vacuous) if the table construct is ever renamed. Consider asserting the filter matched at least one statement.
- `_credential_expired`'s two log paths (`warning` for malformed, `info` for expired) are worth a one-line comment on why the severity differs, since only the `warning` path is test-pinned for PII.

---

perrozzi:
**Disposition — review round 1** (against `8043a70`; fixes land in the next push)

**Blocker — IAM narrowing vs actual call sites: verified correct, and the suggested guard is adopted.**
The verification you asked for, concretely:
- `mcp_handler.py` touches `projects_table` at exactly four sites: the auth Query, the `last_used_at` UpdateItem, and two tool Queries (`get_project`, `list_personas`). Neither tool uses `get_item` — both are `pk`-Queries.
- The autoseed side-door reaches exactly one `projects.py` function: `projects.get_project`, whose only table call is a Query. The only `get_item` against the projects table in `projects.py` is inside `update_persona_note`, which nothing on this Lambda can reach. The markdown/steering helpers autoseed uses are pure formatting.
- On `read-write` scope implying write paths: no registered tool requires it (`TOOL_SCOPE_REQUIREMENTS` is all `"read"`) and autoseed is read-only. The scope exists as forward-compat for future write tools; the only write this Lambda performs is `last_used_at`. Widening the grant will be the deliberate act of whichever PR adds the first write tool.
- **Adopted:** a three-part lockstep guard in `test_mcp_security.py` (`TestProjectsTableUsageMatchesNarrowGrant`) — a strict mock whose non-granted methods raise, driven end-to-end through every registered tool; the autoseed route; and a source-level scan of `mcp_handler` + `projects.get_project` + `projects.autoseed_project` for non-granted operations. The CDK exact-set test and the handler can no longer drift apart silently.

**Should fix**
1. *`mcpMethod` unused* — **declined, factual**: it has a second consumer, the `AwsSolutions-COG4` `NagSuppressions.addResourceSuppressions(mcpMethod, …)` at the end of the MCP block. `typecheck:cdk` (with `noUnusedLocals`) passes.
2. *Frontend `expires_at`* — **accepted in substance, one correction**: `ApiToken` and `CreateApiTokenResponse` now carry `expires_at?: string | null`. The correction: this path is typed by plain TS interfaces through `fetchApi<T>`, not a Zod schema — there is no `.strict()` parse to fail, so nothing was broken at runtime; the declared type now matches the wire. (The Zod-at-boundaries precedents are `formSchema`/`scrapersSchema`/`categoriesSchema`; the token list isn't one of them today.)
3. *Throttle only template-proven* — **accepted as stated**: this is exactly the PR body's "no live-stage verification yet" caveat. The `*` wildcard is the documented `MethodSetting.HttpMethod` form (a verb or `*`). Commitment: this repo's convention is deploy-before-merge, and the deploy verification for this PR will read the stage's method settings back (`apigateway get-stage`) and assert both `/mcp` entries are present with 20/40 — recorded as a merge precondition, not an afterthought.
4. *`ALLOWED_ORIGIN` format and default* — **half accepted**: the unset-default behavior is now pinned both ways (`TestOriginDefaultFailsClosed`: `''` refuses a present Origin, absent Origin still passes). Comma-separated lists — declined for now: `allowedOrigin` is a single-origin string everywhere in this stack (every API Lambda receives the same value; no list format exists today), and parsing a format nothing produces would be speculative. If a second origin ever arrives, the parsing belongs in shared code for all thirteen Lambdas, not here.

**Nits**
- `Access-Control-Expose-Headers: WWW-Authenticate` — **accepted**, added to `CORS_HEADERS`.
- Case-insensitive header scan — **declined**: API Gateway REST proxy (the only transport wired to this Lambda) lowercases header names, and `origin`/`Origin` mirrors the sibling lookups in `_authenticate` (`authorization`/`Authorization`, `x-project-id`/`X-Project-Id`). A generic scan would diverge from the file's own convention for an input no wired transport produces.
- `-32600` on the 403 — **declined**: the MCP transport spec says an Origin-rejection body "MAY comprise a JSON-RPC error response that has no id" and prescribes no code; clients act on the HTTP 403, not the embedded code. Changing it is churn without a consumer.
- Vacuous-filter risk in the IAM test — **accepted**: the filter now asserts it matched at least one statement before the exact-set comparison.
- Severity split comment — **accepted**: `_credential_expired`'s docstring now says why expired is `info` (expected lifecycle; caller re-mints) and malformed is `warning` (server-side data damage wanting an operator).

Gates on the incoming push: backend pytest 2307 (5 new) · CDK vitest 255 · `typecheck` + `typecheck:cdk` clean · McpAccessTab suites 41 · zero lint delta.

perrozzi:

Since the last review the author added the `Access-Control-Expose-Headers: WWW-Authenticate` header, a comment explaining the info/warning severity split, a positive-control guard on the vitest IAM statement filter, a fail-closed test for the empty `ALLOWED_ORIGIN` default, TS interface fields for `expires_at`, and a new Python test class intended to be the "Python half" of the IAM lockstep. Most nits are resolved; the IAM blocker is addressed in intent but the new tests are weaker than they read.

### Resolved
- `WWW-Authenticate` is now exposed via CORS so browser MCP clients can read the challenge.
- Vitest IAM filter now asserts it matched ≥1 statement, so a table rename can't make the exact-set assertion vacuous.
- `ALLOWED_ORIGIN=''` default is now pinned in both directions (present Origin → 403, absent Origin → 200).
- One-line rationale added for the differing log severities in `_credential_expired`.
- `expires_at` added to `ApiToken` / `CreateApiTokenResponse` in `frontend/src/api/types.ts`.

### Still open
- `const mcpMethod = mcpResource.addMethod('POST', ...)` in `voc-datalake/lib/stacks/api-stack.ts` is now an unused binding.
- Zod schema for the token responses: only the TS interfaces were updated — if the MCP Access tab parses these through a Zod schema (especially a `.strict()` one), `expires_at` still needs a `z.string().nullable()` field.
- `_origin_allowed` exact-string compare vs a possible comma-separated / multi-origin `ALLOWED_ORIGIN`.
- No live-stage verification of the `'/mcp/{proxy+}/*'` method-settings key.
- Nits: case-insensitive Origin header scan; `-32600` for a transport-level 403.

### New findings

**Blocker-ish — the new lockstep tests can pass vacuously, so the original blocker isn't actually closed** (`voc-datalake/lambda/api/test/test_mcp_security.py`):

- `test_no_reachable_call_site_uses_a_non_granted_action` only flags a line if it contains **both** the forbidden op **and** the literal `projects_table`. In `projects.py` the table is obtained via `get_projects_table()` and used through a local (`table.put_item(...)`, as visible in `projects_handler.py`), so no line will ever match — the `projects.get_project` / `projects.autoseed_project` halves of this scan are almost certainly no-ops, and multi-line call formatting defeats it too. Apply the same positive-control idea you just applied in the vitest: assert the scan found at least one real DynamoDB call per scanned source, and match on the local table variable name rather than `projects_table`.
- `test_autoseed_stays_within_the_granted_actions` patches `mcp_handler.autoseed_project` out entirely and then asserts only that it was called — it proves nothing about that path's table actions, despite the name and the "pin its usage too" docstring. Since `autoseed` is the most plausible `put_item` site on this Lambda, this is exactly the gap the blocker was about; either drive the real `projects.autoseed_project` against a strict table mock (patching `projects.get_projects_table`) or rename the test to what it checks.
- `test_every_tool_stays_within_the_granted_actions` asserts `"does not permit" not in response["body"]`. That depends on the tool-error path echoing the exception message; if any handler catches broadly and returns a generic message, a real violation passes green. Assert on the mock instead (`strict.get_item.assert_not_called()` etc. after the loop). Also every tool but `get_feedback_detail` is invoked with `{}` arguments, so tools that validate required args first likely never reach DynamoDB — the loop's coverage may be much thinner than it appears.

No other issues in the incremental diff.

---

perrozzi:
**Disposition — review round 2** (against `4361d89`; fixes land in the next push)

**New findings — the vacuity critique of the lockstep tests: accepted in full. All three rewritten, and this time each is mutation-proven rather than asserted.**

- `test_every_tool_stays_within_the_granted_actions` — the verdict now reads off the mock (`assert_not_called()` per forbidden method after the loop), not the response body, exactly as you suggested. A positive control asserts `strict.query.call_count > len(TOOL_HANDLERS)` — auth queries once per call and the two projects-table tools query again — so if the loop ever stops reaching the table, the test fails instead of going quiet. On the `{}`-arguments concern: the only tool with a required argument is `get_feedback_detail` and it is supplied; the other four validate nothing before their reads, and the positive control now proves the table is reached rather than assuming it.
- `test_autoseed_stays_within_the_granted_actions` — nothing is patched away anymore. The route now runs the **real** `projects.autoseed_project` → `projects.get_project` against a second strict table patched at `projects.projects_table` (auth keeps its own strict table on `mcp_handler.projects_table`), and the verdict reads off both mocks, with `strict_data.query.assert_called_once()` as the positive control. You were right that the previous version proved only dispatch.
- `test_no_reachable_call_site_uses_a_non_granted_action` — replaced the line scan with an **AST walk** (`ast.parse` + attribute-call collection on the `projects_table` name), which multi-line formatting cannot defeat, and gave each scanned source the positive control you proposed: the scan asserts equality with the *expected granted set* (`{'query','update_item'}` for the handler, `{'query'}` for `get_project`), so a walker gone blind reports a wrong set rather than an empty pass. One factual correction to the finding: `projects.py` binds `projects_table = get_projects_table()` at module level and `get_project` calls it by exactly that name — the local-variable claim was mistaken (the `table = get_projects_table()` shape is `projects_handler.py`). The structural point stood regardless; the multi-line and no-positive-control halves were real, hence the rewrite.

Mutation evidence for all three (harness in the worktree, re-runnable): a `get_item` added to a tool fails both the runtime and AST tests; a **multi-line** `put_item` inserted inside `projects.get_project` — the exact shape the old scan missed — fails both the real-path autoseed test and the AST scan; stubbing the two projects-table tools out (so the table is never reached) fails the positive control. Five mutations, five failures.

**Carried "still open" items — answered in round 1's disposition, restated once with pointers, and not planned to reappear:**
- *`mcpMethod` unused* — factually not: its consumer is the `NagSuppressions.addResourceSuppressions(mcpMethod, …)` COG4 suppression a few lines below the proxy's. `typecheck:cdk` (strict, `noUnusedLocals`) is green on every push.
- *Zod schema for token responses* — there is no Zod parse on this path to update: `listApiTokens`/`createApiToken` go through `fetchApi<T>` with plain interfaces (now carrying `expires_at`). The boundary-normalizer precedents (`formSchema`, `scrapersSchema`, `categoriesSchema`) don't include the token list today; introducing one is UI-for-expiry work, deliberately deferred per the PR body's open question 2.
- *Multi-origin `ALLOWED_ORIGIN`* — the stack produces a single-origin string for all thirteen API Lambdas; no list format exists to parse. Fail-closed default is pinned both ways since round 1.
- *Live verification of the throttle keys* — remains a stated merge precondition (deploy-before-merge with `get-stage` readback), not treated as done.
- *Origin header casing / `-32600`* — declined round 1 with reasoning (REST proxy lowercases; the transport spec prescribes no code for the 403 body).

Gates on the incoming push: backend pytest 2307 · zero lint delta · CDK vitest and typechecks unchanged from `4361d89` (this round touches only the Python test file).

perrozzi:

Only the Python IAM-lockstep tests changed in this push (`voc-datalake/lambda/api/test/test_mcp_security.py`): the substring grep became an AST walk with positive controls, the tool loop now reads its verdict off the strict mock, and the autoseed test now drives the real `projects.autoseed_project` against a strict table patched at `projects.projects_table`. That addresses the substance of my previous blocker-ish finding — the lockstep tests are no longer vacuous, and the AST scan passing at all confirms `projects.get_project` really does use a module-level `projects_table` (my earlier `get_projects_table()` note conflated `projects.py` with `projects_handler.py`).

### Resolved
- `test_no_reachable_call_site_uses_a_non_granted_action` now uses AST + exact-set assertions with positive controls, so formatting/multi-line calls and a blind walker can't make it pass silently.
- `test_autoseed_stays_within_the_granted_actions` no longer patches `autoseed_project` away; it exercises the real path against a strict data table and asserts on both mocks.
- `test_every_tool_stays_within_the_granted_actions` asserts on the mock (`assert_not_called`) instead of response body text, with a call-count positive control.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', ...)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- Zod schema for the token list/create responses (only `frontend/src/api/types.ts` was updated).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN`.
- No live-stage verification of the `'/mcp/{proxy+}/*'` method-settings key; `-32600` for a transport-level 403 (nit).

### New findings

**Should-fix — the autoseed source-scan control proves less than the docstring claims** (`voc-datalake/lambda/api/test/test_mcp_security.py`): `assert self._table_calls(autoseed_src) == set()` plus `assert "get_project(" in autoseed_src` shows only that `get_project` *is* called, not that it is the *only* route to the table — any other `projects.py` helper `autoseed_project` calls (a save/update path) would slip through both the scan and the "only route" comment. Either walk the callee names in `autoseed_src` and assert they are a subset of a pinned allowlist (whose table calls you then scan), or soften the docstring to what it checks. The runtime test covers the executed branch, so this is a comment/coverage mismatch, not an unguarded hole.

**Nits**
- The tool loop's positive control (`strict.query.call_count > len(TOOL_HANDLERS)`) still can't distinguish "every tool reached the table" from "auth queried once per call plus two tools"; tools invoked with `{}` that validate args first may never touch DynamoDB. A per-tool assertion (or asserting `2xx`/non-validation-error for each) would close it.
- `GRANTED = ("query", "update_item")` is now duplicated as a local `granted = {...}` in the AST test — derive one from the other so they can't drift.

---

perrozzi:
**Disposition — review round 3** (against `0146a5c`; fixes land in the next push)

All three findings accepted — each was a real gap between what a test proved and what its name claimed.

- **Autoseed callee closure (should-fix):** the "only route to the table" claim is now enforced instead of asserted. The test walks every bare-name call in `autoseed_project` that resolves to a `projects.py` callable and requires the set to be a subset of a pinned allowlist — `get_project` (scanned for table calls above) plus the four formatting helpers, each of which is now itself AST-scanned and required table-free. A new callee fails the closure by name with instructions to scan it and extend the allowlist deliberately. Mutation-proven: adding a `list_projects()` call inside `autoseed_project` fails the test.
- **Per-tool positive control (nit):** the aggregate `call_count > len(TOOL_HANDLERS)` is replaced by an exact per-tool delta — every call pays one auth Query, and `get_project`/`list_personas` pay exactly one more — asserted inside the loop, so a tool that starts validating its way past DynamoDB now fails by name rather than hiding in an aggregate.
- **`GRANTED` duplication (nit):** the AST test's expected set is now derived from `self.GRANTED` (the same constant that configures the strict mock), so the runtime and source-level halves cannot drift.

One honest note from proving the closure test: my mutation harness needed two fixes of its own before the mutation actually applied (a first-occurrence anchor landed in a different function that shares the `project_data = get_project(project_id)` line, and `autoseed_project` is the last function in the file so the end-of-function anchor needed a fallback) — which is to say the "VACUOUS" verdicts in between were harness bugs, not test bugs, and the final run has all six mutations caught: `get_item` in a tool (runtime + AST), a multi-line `put_item` inside `get_project` (real-path + AST), stubbed-out tools (positive control), and the unallowlisted callee.

Carried items remain as dispositioned in rounds 1–2 (`mcpMethod`'s consumer is the COG4 suppression; no Zod parse exists on the token path; single-origin is the only format the stack produces; live throttle readback is a merge precondition).

Gates on the incoming push: `test_mcp_security.py` 107 · full backend 2307 · zero lint delta · CDK/typecheck untouched this round.

perrozzi:

Only `voc-datalake/lambda/api/test/test_mcp_security.py` changed in this push: the tool loop now asserts an exact per-tool `query` delta, the AST test derives `granted` from `self.GRANTED`, and the autoseed source-scan replaced the `"get_project(" in src` string check with a callee-allowlist closure plus table-free scans of the named helpers. All three of my previous delta findings are addressed; the earlier "Still open" list is untouched.

### Resolved
- Per-tool `query` delta (`get_project`/`list_personas` == 2, others == 1) replaces the aggregate `> len(TOOL_HANDLERS)` control, so a tool that stops reaching the table fails by name.
- `granted = set(self.GRANTED)` removes the duplicated action set.
- The autoseed scan now pins the callee set and verifies each allowlisted helper is table-free, instead of only proving `get_project` is called.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', ...)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema for the token list/create responses (only `frontend/src/api/types.ts` updated).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN`.
- No live-stage verification of the `'/mcp/{proxy+}/*'` method-settings key; `-32600` for a transport-level 403 (nit).

### New findings

**Nits** (`voc-datalake/lambda/api/test/test_mcp_security.py`)
- The callee closure is one level deep and name-form only: `_called_names` sees only `ast.Name` calls, so a future `helpers.save_project(...)`/attribute-form callee is invisible, and the allowlisted helpers' own callees are never scanned. The docstring ("every projects.py function it calls is on a pinned allowlist whose members are either scanned above or verified table-free") claims more than the check delivers — either recurse or say "direct callees only".
- `getattr(projects_module, helper)` over the hardcoded `table_free_helpers` will raise a bare `AttributeError` if one of those private helpers is renamed or dropped, instead of the deliberate failure message the rest of the test is careful to give. A `hasattr` guard with a message would keep the failure legible.

---

perrozzi:
**Disposition — review round 4** (against `f32dcde`; fixes land in the next push)

Both nits accepted:

- **Closure depth honesty:** the docstring/comment now says exactly what the check delivers — one level deep, bare-name calls only — and states the division of labor: deeper indirection (attribute-form calls, a helper growing its own callee) is the runtime strict-mock test's job, which executes the real path and cannot be fooled by call shape. Chose truthful wording over recursion: a recursive static walker would re-open the vacuity class this thread just closed (imports, aliasing and attribute dispatch all defeat static closure eventually), while the runtime half already covers the executed path.
- **Legible failure on a renamed helper:** each allowlisted helper now gets an explicit `hasattr` assertion with a message naming the fix (`update table_free_helpers…`) before `getattr`, so a rename fails the way the rest of the test fails — with instructions — rather than as a bare `AttributeError`.

Carried items stand as dispositioned in rounds 1–3.

Gates on the incoming push: `test_mcp_security.py` 107 · zero lint delta · nothing outside the Python test file touched this round.

perrozzi:

Only `voc-datalake/lambda/api/test/test_mcp_security.py` changed in this push: the autoseed closure's comment now states its real scope ("DIRECT bare-name callee", one level deep, name-form only, deferring deeper indirection to the runtime strict-mock test), and a `hasattr` assertion with a legible message now precedes each `getattr(projects_module, helper)` in the `table_free_helpers` loop. Both of my previous nits are addressed; the earlier "Still open" list is untouched.

### Resolved
- The docstring/comment no longer overclaims the closure's depth, so the check and its stated contract agree.
- Renaming or dropping a private autoseed helper now fails with a directive message instead of a bare `AttributeError`.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', ...)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema for the token list/create responses (only `frontend/src/api/types.ts` updated).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN`.
- No live-stage verification of the `'/mcp/{proxy+}/*'` method-settings key; `-32600` for a transport-level 403 (nit).

No new findings — the delta is comment-and-assertion only, test-file scoped, with no behavior change.

---

perrozzi:
**Review loop status: converged at `a69fc96`** — round 5 reports no new findings; rounds 1–4 each closed with all substantive findings fixed and mutation-proven, or declined with evidence.

For the human reviewer, the four carried "Still open" lines and where each was answered:

| Item | Standing answer |
|---|---|
| `mcpMethod` unused | Factually not — its consumer is the `AwsSolutions-COG4` nag suppression below the proxy block; `typecheck:cdk` (`noUnusedLocals`) is green (round 1) |
| Zod schema for token responses | No Zod parse exists on this path — `fetchApi<T>` with plain interfaces, now carrying `expires_at`. Introducing a boundary schema is the expiry-UI work, PR body open question 2 (rounds 1–2) |
| Multi-origin `ALLOWED_ORIGIN` | The stack produces a single-origin string for all thirteen API Lambdas; no list format exists to parse. Unset-default fail-closed is pinned both ways (round 1) |
| Live throttle readback / `-32600` | Readback of the stage's method settings is a **merge precondition** under this repo's deploy-before-merge convention, not waived. `-32600`: the transport spec prescribes no code for the Origin-403 body (rounds 1–3) |

Cumulative delta over the loop: the IAM narrowing now has a three-part Python lockstep guard (strict-mock tool loop with per-tool query deltas, the real autoseed path against a strict table, AST source scan with callee closure) — twelve mutation proofs across both harnesses, all caught. Plus `Access-Control-Expose-Headers: WWW-Authenticate`, the vitest filter positive-control, fail-closed `ALLOWED_ORIGIN=''` pins, `expires_at` on the TS interfaces, and the severity-split rationale.

Awaiting: maintainer input on the PR body's two open questions (throttle 20/40; expiry in the token UI now vs later), then deploy + live verification (including the stage method-settings readback) before merge.

perrozzi:
**Deploy + live verification (the stated merge precondition) — done, with two live-only findings fixed**

Deployed to the live stack (`VocApiStack --exclusively`) and verified end to end. The review's skepticism about the throttle key was **vindicated**: the first deploy FAILED.

**Finding 1 — `'/mcp/{proxy+}/*'` is rejected by the service.** CloudFormation rolled back cleanly with: per-path wildcards are not valid method paths — only `{resource_path}/{http_method}` or the stage-wide `*/*`. Direct `update-stage` probes then established the full rule: `ANY` is rejected as an httpMethod too; concrete verbs on the proxy path are accepted. **Fix: two concrete-verb keys — `'/mcp/{proxy+}/POST'` (JSON-RPC on subpaths) and `'/mcp/{proxy+}/GET'` (the autoseed side-door)** — which are the complete set of verbs on that proxy that reach DynamoDB (everything else is a handler 405 or the CORS mock). The lockstep test now checks each key's concrete verb against the wired methods, which is strictly stronger than the wildcard special-case it replaces. My round-1 claim that `*` was "the documented form" was wrong in the way only a deploy could show — this is exactly why the readback was a merge precondition.

**Finding 2 — REST API Gateway remaps `WWW-Authenticate` from Lambda responses** to `x-amzn-remapped-www-authenticate`, unconditionally (documented behavior, no opt-out). The challenge was arriving, under a mangled name. **Fix, honest in both halves:** gateway-*generated* 401s (the token authorizer's shape rejections) now carry the true header via an `UNAUTHORIZED` gateway response; handler-side 401s keep sending it and clients receive it under the remapped name, with the caveat documented at the constant. Both delivery paths are asserted live.

**Live verification, all green:**

| Check | Result |
|---|---|
| Stage method-settings readback | 5/5 keys bound at 20/40: `/mcp/POST`, `/mcp/{proxy+}/POST`, `/mcp/{proxy+}/GET`, both ballot routes |
| Usage plan | gone from the account's plan list |
| IAM readback (live role policy) | projects-table actions exactly `Query` + `UpdateItem` |
| MCP protocol battery | **22/22** — strict mint 400s (0/366/string/bool), mint+list+echo of `expires_at`, legacy tokens read as non-expiring, initialize, bogus-token 401 with JSON-RPC error, both challenge delivery paths, challenge CORS-exposed, foreign-Origin 403, tools/list (5), `get_project` + `get_metrics_summary` + autoseed **succeeding under the narrowed role** (the Query grant), `last_used_at` stamped (the UpdateItem + KMS pairing live), a **seeded pre-expired credential refused 401** (enforced at auth, not TTL — row seeded and deleted), revocation immediate |
| Full API regression | `test-api.sh` 34/34 |
| Local gates on the fixes | backend 2307 · CDK 255 · zero lint delta · baseline regenerated (throttle keys + gateway response) |

The incoming push carries the two fixes. Bot: the throttle-key skepticism was the round's best call.

perrozzi:

Since my last review the throttle keys moved from the (invalid) `'/mcp/{proxy+}/*'` wildcard to concrete `'/mcp/{proxy+}/POST'` and `'/mcp/{proxy+}/GET'` — backed by a live deploy/`update-stage` probe — with the lockstep test rewritten to accept `ANY` coverage per verb, a new API-wide `UNAUTHORIZED` gateway response carrying the RFC 6750 challenge, a caveat comment in `mcp_handler.py` about API Gateway remapping `WWW-Authenticate` on proxy responses, and a regenerated baseline hash. That closes the biggest open item; the other previous findings are untouched.

### Resolved
- `'/mcp/{proxy+}/*'` is gone; keys are now service-accepted concrete verbs, live-verified, and the lockstep test checks path+verb (or `ANY`) rather than path-only.
- The proxy-path 401 challenge delivery gap is now documented and compensated for gateway-generated 401s via the new `Unauthorized` gateway response.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', ...)` binding in `voc-datalake/lib/stacks/api-stack.ts` (now that the usage plan is deleted, nothing consumes it).
- No Zod schema for the token list/create responses (`frontend/src/api/types.ts` only).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN`.
- `-32600` for a transport-level 403 (nit).

### New findings
- **Should fix — the new `Unauthorized` gateway response is API-wide, not MCP-scoped** (`voc-datalake/lib/stacks/api-stack.ts`). `addGatewayResponse('Unauthorized', …)` applies to every gateway-generated 401 on this REST API, including Cognito-authorizer rejections on all non-MCP routes, which will now advertise `Bearer error="invalid_token"`. That's mostly harmless but the adjacent comment describes it as "the MCP token authorizer refusing a malformed Bearer shape"; please either say so plainly or confirm the Cognito routes are fine advertising a Bearer challenge.
- **Should fix — no test pins the new gateway response.** Every other guard in this PR got an assertion; this one has none, so removing the `WWW-Authenticate`/`Access-Control-Expose-Headers` entries (or the whole `addGatewayResponse`) is silent. A short `api-stack.test.ts` case asserting `AWS::ApiGateway::GatewayResponse` with `ResponseType: UNAUTHORIZED` carries both headers would match the repo's regression-test rule and the style of the throttle tests.
- **Nit — non-DynamoDB verbs on the proxy are no longer method-throttled.** The comment justifies this ("405 before any DynamoDB read"), which is fine for the brute-force argument, but PUT/DELETE/etc. on `/mcp/{proxy+}` now fall back to the stage default; worth one sentence saying that's the intended ceiling.
- **Nit — the PR description now contradicts the code.** The description still states "No live-stage verification yet," while the new comments cite a live deploy and `update-stage` probes (dated `2026-08-18`, which also looks like a typo). Please reconcile both.

---

perrozzi:
**Disposition — review round 6** (against `04db1ee`; fixes land in the next push)

All four findings addressed:

- **API-wide gateway response — accepted, resolved by argument + honest comment.** You're right that `addGatewayResponse('Unauthorized', …)` fires for Cognito-authorizer rejections on every route, not just MCP. That is *correct*, not merely harmless: every credential this API accepts arrives as `Authorization: Bearer …` — a Cognito ID token is a bearer token — so RFC 6750's Bearer challenge is the truthful answer on all of them. The comment now says exactly that instead of describing only the MCP case.
- **Pinning test — accepted.** New `unauthorized gateway response` case asserts the `UNAUTHORIZED` gateway response exists and carries both `WWW-Authenticate` and `Access-Control-Expose-Headers` with their exact values. CDK suite is now 256.
- **Un-throttled verbs nit — accepted**, one sentence added: PUT/DELETE/… on the proxy deliberately ride the stage default; their whole cost is a Lambda invoke returning 405, and that ceiling is intended.
- **Stale PR description — accepted**, the "No live-stage verification yet" line now reads as done, struck through with the verification summary and a pointer to the deploy comment. On the date: `2026-08-18` is today's date, not a typo.

Gates on the incoming push: CDK vitest 256 (1 new) · typecheck:cdk clean · api-stack.ts delta is comments-only beyond the test, so the baseline holds unchanged · backend untouched this round.

perrozzi:

Since the last review the only changes are documentation/test-side: the `Unauthorized` gateway-response comment now states plainly that it is API-wide and covers the Cognito authorizer too, a new `unauthorized gateway response` test pins `ResponseType: UNAUTHORIZED` with both headers, and the throttle comment now says non-DynamoDB verbs on the proxy deliberately ride the stage default. That closes all three of my previous new findings; the PR description was also updated (live-stage verification is no longer listed as missing).

### Resolved
- API-wide scope of `addGatewayResponse('Unauthorized', …)` is now stated explicitly and justified (Cognito ID tokens are bearer tokens) in `voc-datalake/lib/stacks/api-stack.ts`.
- New `describe('unauthorized gateway response')` case in `voc-datalake/lib/stacks/api-stack.test.ts` pins the `WWW-Authenticate` and `Access-Control-Expose-Headers` response parameters, so removing either is no longer silent.
- Throttle comment now records that PUT/DELETE/… on `/mcp/{proxy+}` intentionally fall back to the stage default (100/200).
- PR description no longer claims "no live-stage verification yet".

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', …)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema for the token list/create responses (`frontend/src/api/types.ts` interfaces only).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN` (`lambda/api/mcp_handler.py`).
- `-32600` for a transport-level 403 (nit).
- `2026-08-18` date in the two live-verification comments still reads as a future/typo'd date.

### New findings
None — the incremental changes are clean; the new test matches the style and rigor of the adjacent throttle/IAM tests.

---

perrozzi:

Since the last review the author answered their own open question and shipped the UI half of token expiry: a new `tokenExpiry.ts` module (presets + `isTokenExpiryChoice` type guard + `tokenExpiryState`), an `Expiration` select in `CreateTokenForm`, a deadline line on `NewTokenBanner`, a `TokenExpiryBadge` on token rows, `expires_in_days` on `api.createApiToken`, i18n keys in all 8 locales, and four new tests. None of my previously-open items were touched.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', …)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema for the token list/create responses — now more pointed: `useTokenMutations` reads `result.expires_at` and `TokenExpiryBadge` reads `token.expires_at` straight off an unvalidated payload (`frontend/src/api/types.ts`, `McpAccessTab.tsx`).
- `_origin_allowed` exact-string compare vs a possibly comma-separated `ALLOWED_ORIGIN` (`lambda/api/mcp_handler.py`).
- `-32600` for a transport-level 403 (nit).
- `2026-08-18` in the two live-verification comments still reads as a future date.

### New findings
- **Should fix:** `voc-datalake/frontend/src/pages/ProjectDetail/McpAccessTab.test.tsx` ends without a newline ("\ No newline at end of file") and adds a double blank line before `describe('token expiry')` — `eol-last`/`no-multiple-empty-lines` will make the "zero lint delta" claim false.
- **Should fix:** `NewTokenBanner` (`McpAccessComponents.tsx`) formats the deadline with a bare `new Date(expiresAt).toLocaleDateString()`, so a malformed/unparseable `expires_at` renders "Invalid Date" on the one-time banner — the row badge already fails safe via `tokenExpiryState`; reuse it here for the same fail-quiet behaviour.
- **Nit:** the locale files carry both a base `expiryDays` and `expiryDays_one`/`_other`; `_one` is dead in ja/ko/zh (no `one` plural category), and since only 30/90/365 are offered no singular form is ever selected. Harmless, but it's three keys per locale doing one key's work.
- **Nit:** `newTokenExpiresAt` is never cleared when the banner is dismissed (only `newlyCreatedToken` is), so it holds a stale deadline until the next mint — invisible today because the banner is gated on the token, but it's the same staleness the expiry-reset test guards against for the form.

---

perrozzi:
**Disposition — review round 8** (against `448db38`; fixes land in the next push)

- **Test-file EOF/blank-lines — accepted**: single blank line before the new describe, exactly one trailing newline. On the "zero lint delta" claim: those rules are not enabled for test files in this config (`eslint .` ignores them), so the claim was technically true — but the hygiene fix is right regardless, and the file now matches the repo style either way.
- **"Invalid Date" on the banner — accepted**: the banner line is now gated on `tokenExpiryState(expiresAt) !== 'none'`, the same fail-quiet derivation the row badge uses, so a malformed echo renders nothing rather than "Invalid Date".
- **Stale `newTokenExpiresAt` on dismiss — accepted**: dismiss now clears the deadline beside the token. You're right that it's the same staleness class the expiry-reset test guards on the form side.
- **Plural-keys nit — declined, convention**: `expiryDays` + `_one`/`_other` (duplicated in ja/ko/zh) mirrors exactly how this file already writes `activeTokens` in every locale. Trimming to one key would make the new entries the odd ones out; if the convention is wrong it should change wholesale, not per key.
- **Carried items** stand as dispositioned (rounds 1–6). On the recurring date remark once more: 2026-08-18 is the current date.

Gates on the incoming push: ProjectDetail suite 548 · `lint:frontend` 0 · `typecheck` 0.

perrozzi:

Since my last review the author pushed a small clean-up pass: the missing trailing newline and double blank line in `McpAccessTab.test.tsx` are fixed, `NewTokenBanner` now gates the deadline line on `tokenExpiryState(...) !== 'none'` instead of bare presence, and `newTokenExpiresAt` is cleared on dismiss. That closes both of my previous Should-fix items and one nit; the deeper items are untouched.

### Resolved
- `McpAccessTab.test.tsx` `eol-last` / `no-multiple-empty-lines` violations fixed.
- `NewTokenBanner` no longer renders "Invalid Date" for a malformed `expires_at` — now shares the row badge's fail-quiet rule via `tokenExpiryState`.
- `newTokenExpiresAt` is reset on banner dismiss, so no stale deadline survives to the next mint.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', …)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema/type guard for the token list/create responses (`frontend/src/api/types.ts`, `McpAccessTab.tsx` read `expires_at` off an unvalidated payload).
- `_origin_allowed` exact-string compare against a possibly comma-separated `ALLOWED_ORIGIN` (`lambda/api/mcp_handler.py`).
- `-32600` for a transport-level 403 (nit).
- Duplicate `expiryDays` / `expiryDays_one` / `expiryDays_other` keys across locales, with `_one` dead in ja/ko/zh (nit).
- `2026-08-18` in the live-verification comments still reads as a future date (nit).

No new problems introduced — the incremental changes are correct and minimal.

---

perrozzi:
**Deploy + live verification of the expiry UI — done, 22/22 backend + full browser E2E**

Converged at `4411bd6` (round 9: "No new problems introduced"), then deployed and verified through the browser.

**Browser E2E on the live site** (signed in as a real Cognito user, MCP Access tab):

| Proof | Observed |
|---|---|
| Deployed bundle is this branch | Locale requests carry `?v=4411bd6` — the exact head commit — and the lazy page chunk is **md5-identical** to the local build (`38ee19d2…`) |
| Expiration control renders | Select present with all four presets, defaulting to **"Never expires"** |
| Wire payload | `{"name":"pr344-live-ui-check","scope":"read","expires_in_days":30}` — read off the actual network request, not inferred |
| Response echo | `expires_at: 2026-09-17T21:17:38+00:00` (exactly 30 days out) |
| Banner | "This token expires on 9/17/2026." |
| Row badge, positive **and** negative | Exactly **one** row shows `Expires 9/17/2026`; the pre-existing legacy token (created 7/17/2026) shows **no** expiry line — the same both-directions assertion the unit test makes, now live |
| Revocation | 200, and the list drops to the legacy token alone |
| Backend battery re-run | **22/22** · `test-api.sh` **34/34** |

**⚠️ Incident during this deploy, disclosed in full because it touched the live site.** I deployed the frontend with `aws s3 sync dist/ --delete`, which **deleted `config.json`** — that file is generated by the CDK `BucketDeployment` (`s3deploy.Source.data`), not built into `dist/`, so a prune from `dist` removes it. Consequence: the live login showed "Cognito not configured" for roughly ten minutes (the SPA fallback served `index.html` for the missing path, so a naive `curl` returned 200 and *looked* healthy — the failure was only visible in the app).

Fixed by running `cdk deploy VocApiStack --exclusively`, which regenerates the file authoritatively rather than me hand-reconstructing values — deliberately, because a wrong `identityPoolId` makes `RuntimeConfigSchema` blank *every* Cognito value and reproduces the same symptom. Verified after: `config.json` present in the bucket, served as valid JSON, all four Cognito values non-empty, `features.webSearch: true`, login works, 34/34 API.

**Takeaway worth keeping in the repo's deployment notes: never `s3 sync --delete` this bucket — deploy the frontend through CDK** (or sync without `--delete`). The existing note warns that the BucketDeployment prunes; the inverse hazard — a manual prune destroying CDK-generated content — was not written down anywhere.

Nothing else changed in the branch as a result; the incident was operator error on my side, not a code defect.

perrozzi:

Since my last review the only change is a new documentation callout in `docs/deployment.md` warning against `aws s3 sync --delete` on the website bucket (config.json is deploy-time generated, CloudFront masks its absence with a 200). No code changes, so none of the previously open items moved.

### Still open
- Unused `const mcpMethod = mcpResource.addMethod('POST', …)` binding in `voc-datalake/lib/stacks/api-stack.ts`.
- No Zod schema/type guard for the token list/create responses (`frontend/src/api/types.ts`, `McpAccessTab.tsx`).
- `_origin_allowed` exact-string compare against a possibly comma-separated `ALLOWED_ORIGIN` (`lambda/api/mcp_handler.py`).
- `-32600` for a transport-level 403 (nit).
- Duplicate `expiryDays` / `expiryDays_one` / `expiryDays_other` keys across locales, with `_one` dead in ja/ko/zh (nit).
- `2026-08-18` in the live-verification comments still reads as a future date (nit).

### New findings
- Nit / scope: `docs/deployment.md` — this warning is unrelated to the MCP hardening in this PR (IAM, throttle, Origin, expiry). Content looks accurate and useful, but it would be cleaner as its own docs PR so this branch's diff stays reviewable as one change; no objection if you'd rather keep it.

Otherwise the incremental change is clean — prose only, no behavioural or IAM impact.

---

perrozzi:
**Disposition — review round 10**

Only finding is the scope nit on the docs commit: **keeping it here, deliberately.** The warning was earned *by this PR's own deploy verification* — I broke the live login with exactly that command while verifying this branch — so the note and the incident report belong to the same change; splitting them would separate the lesson from the evidence that produced it. It is one prose-only commit (`e76c903`, +19 lines, no code), trivially droppable at merge if a maintainer prefers.

**All ten rounds are now closed.** The six carried items stand as dispositioned across rounds 1–8, unchanged and each with its reason:

| Item | Why it stays |
|---|---|
| `mcpMethod` "unused" | Factually not — the `AwsSolutions-COG4` nag suppression consumes it; `typecheck:cdk` (`noUnusedLocals`) is green every round |
| Zod for token responses | No Zod parse exists on this path (`fetchApi<T>` + interfaces). Adding a boundary normalizer here is a real change to a shared client convention, not a one-liner — worth its own PR |
| Multi-origin `ALLOWED_ORIGIN` | The stack emits a single origin for all 13 API Lambdas; no list format exists to parse. Fail-closed default pinned both ways |
| `-32600` on the Origin 403 | The transport spec prescribes no code for that body; clients act on the HTTP status |
| Plural key trio | Mirrors how `activeTokens` is already written in all 8 locales; changing it should be wholesale, not per key |
| "future date" | 2026-08-18 is today |

**Final state:** converged, deployed, live-verified end to end (22/22 MCP battery, 34/34 API, browser E2E with the wire payload read off the network request), and the branch is `MERGEABLE`. Ready for a maintainer.

Issue #342, opened 2026-08-18
[Bug]: A prioritization ballot for a row that does not exist saves with a 200 and is silently discarded on read


perrozzi:
## What happens

`PATCH /projects/prioritization` stores a ballot for a row id that has no row record, and answers `200 {"success": true, "updated_count": 1}`. Every subsequent read discards that ballot, so the reviewer is told their score saved and the score appears nowhere — not in their own `scores`, not in the team `aggregate`, not on the page.

Confirmed against production while live-verifying #341:

```
PATCH /projects/prioritization  {"scores":{"row_does_not_exist_default":{"impact":3}}}
-> 200 {"success":true,"updated_count":1}

stored:  BALLOT#row_does_not_exist_default#user:{sub}
read:    the row id appears nowhere in the response body
log:     Discarded 3 prioritization ballot(s) across 2 row id(s) ...
```

The read side already behaves correctly — it discards the ballot rather than inventing a row, and #341 made that discard counted and logged. This is the same silent-loss class, still open on the write side.

## Why it matters

Not reachable from the UI today: the page only ever PATCHes row ids it received from the read, so a row it names always exists at the moment it is named.

It becomes reachable as soon as a row can be deleted, which is phase 2 of #339. The sequence is ordinary: a reviewer opens the prioritization page, an admin deletes a row together with its ballots, the reviewer submits scores from the tab they already had open. Today that write succeeds, reports success, and is discarded on the next read. The reviewer has no way to tell, and the only trace is a rising count in a warning nobody is watching in real time.

## What it should do

A ballot write should be refused when its row record does not exist, and the refusal should be enforced by the database rather than by a read-then-write — the same reasoning the row create already uses for its conditional put, since a check-then-write races a concurrent delete and lands the orphan anyway.

A refusal is the right answer rather than a silent drop or an auto-created row: the caller named a row that is gone, which is a state the client can act on by refetching, and inventing the row would resurrect a deleted proposal as a side effect of someone scoring it.

This belongs with the delete-row-with-its-ballots path in phase 2 of #339 rather than as a standalone change, because the two are the same concern seen from either end, and the discard count #341 added is what will show whether the pair is working.

## Acceptance

- [ ] A ballot write naming a row with no row record is refused, and nothing is stored.
- [ ] The refusal is enforced atomically, so a delete racing a ballot write cannot leave an orphaned ballot behind.
[...]

PR #341, opened 2026-08-18
feat(prioritization): a row is a project, and ballots are keyed to the row

perrozzi:
## Summary

**Task:** # Prioritization row identity: one row per project, ballots keyed to the row

This is phase 1 of issue #339, and only phase 1.

## What the product gets wrong today

A prioritization row is a single project document. Every scorable document of every project becomes its
own row, so a project whose PRD and PR/FAQ describe one idea appears twice and a reviewer scores the same
idea twice. Once a room votes from their phones, a room scanning the code on one of those two rows scores
half the idea. On real data this affects roughly one proposal in three.

The end state of phase 1 is that a row is a project, carrying a set of that project's documents as
concrete document ids, and a ballot belongs to a row rather than to a document.

## Where the current behaviour lives

`voc-datalake/lambda/api/projects_handler.py` owns the prioritization partition. Ballots sit in one
partition with the identity in the sort key, `BALLOT#{document_id}#{kind}:{subject}`, where the kind is
`user` for a signed-in reviewer and `anon` for a room ballot. One pre-ballot item remains in that
partition under the sort key `SCORES`, holding a map of document id to score; it is read through and
retired entry by entry on first write, which is why no migration script exists. Three routes read and
write this data: a GET that returns the caller's own ballot plus a per-document aggregate, a PATCH that
writes only the caller's own ballot, and a deliberate 405 stub on PUT that exists so the route does not
fall through to the project upsert.

`voc-datalake/lambda/api/ballots_handler.py` owns the voting sessions added for room voting. A session
record names one document id, and a submitted anonymous ballot takes that document id from the session
record rather than from the request body. Its own docstring describes the single-document limitation as
known rather than accidental, and names resolving the row unit as the separate change this task is.
`voc-datalake/lambda/api/test/test_anon_ballot_key_lockstep.py` pins the sort-key shape across both
handlers, so the two move together or that test fails.

On the page, `collectPRFAQs` in
`voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.ts` flattens each project's scorable
documents into one row each, attaches the project-latest prototype to every one of them, and every
lookup for a score, an aggregate, an edit, a note bound, a sort position and a linked form is keyed by
document id. `SCORABLE_TYPE_META` in that file is the single place deciding that a PRD and a PR/FAQ are
scorable.

## The end state

A row record exists per project in the aggregates table under its own sort-key prefix, holding concrete
document ids: the project's scorable documents, plus its latest prototype. No new table and no new stack;
[...]

Issue #339, opened 2026-08-17
[Feature]: Prioritization Phase 3 — show lineage consistency and stale rows


perrozzi:
> [!NOTE]
> **Status — 2026-08-26:** Phases 1 and 2 are delivered. The per-project row model,
> concrete document selections, row-keyed ballots, first-ballot atomic freeze, note-only
> freeze, row composition/addition, and admin delete-with-ballots shipped through #350
> (with room/session voting completed in #340). Evidence was audited criterion by criterion
> in [this disposition comment](https://github.com/aws-samples/sample-voice-of-customer-datalake/issues/339#issuecomment-5415230618).
>
> **This issue now tracks Phase 3 only:**
> 1. Visually distinguish a row whose documents share one derivation chain from a row
>    whose documents cross generations or have no lineage, without blocking scoring.
> 2. Mark a frozen row stale when a fresher consistent combination exists, with **Add a
>    row** as the suggested action.
>
> The derivation contract already exists in `frontend/src/api/derivation.ts`; the remaining
> work is presentation and staleness logic, not a new data model. The original issue body
> remains below as the decision history for the shipped phases.

## Summary
A prioritization row is a document today: one row per PRD or PR/FAQ across all projects, so one idea
routinely appears as two rows (its PRD and its PR/FAQ carry the same title) and a reviewer votes twice
on one thing — or, once room voting ships, a room scanning the QR on one of the two rows scores half
the idea. On live data this affects one proposal in three.

This changes what a row **is**: one row per project by default, carrying a picked set of that project's
documents (latest PR/FAQ, PRD and prototype to start), with dropdowns to compose a different set, a
button to add another row for another combination, and a lineage signal — bright when the picked
documents belong to one generation chain, greyed when they cross generations or predate lineage.
Ballots re-key from documents to rows.

Decisions already made (product owner, 2026-08-17), recorded here so the review does not reopen them:
- **Any signed-in reviewer may create rows and edit an un-balloted row's composition.** Identity of the
  author is not the protection; the freeze is.
- **Composition freezes at the first ballot — "when the first vote comes in, no changes are possible."**
- **A row stores concrete document ids, never the selector "latest".**
- **The escape hatch is adding another row, plus one destructive action — delete a row together with its
  ballots, behind a confirmation and admin rights. A frozen row is never edited.**

## Severity
Medium-high, and **time-critical relative to #337**: room voting writes the first real ballots, and
[...]

PR #338, opened 2026-08-16
feat(prioritization): rows lead with the team score, own sliders one level in

perrozzi:
Closes #336.

## Summary

The Prioritization row led with the reader's **own** four numbers and the list sorted by the reader's **own** composite, so a group exercise was presented as a personal one — and the disagreement a prioritization meeting exists to resolve was nowhere on the page, even though `GET /projects/prioritization` has returned the per-document team view (per-axis mean, `reviewer_count`, `score_spread`) since #333.

This renders that aggregate and sorts by it. Frontend only: **no new endpoint and no backend file touched.**

eiifcbndbggnlnebulgtnhngdndfebfnilvhgdheetgk
> **Amended during review.** This originally read "no change to how a ballot is stored or keyed", which was true at `37e436d` and is no longer true of the PATCH **body shape**. Review found that a partial first ballot was writing axes the reader never chose — `updateScore` seeded a pending edit from `DEFAULT_SCORE`, so moving one slider sent all four axes, two of them as a `0` the slider cannot express. Since the backend counts an explicit `0` as a vote and averages each axis over the reviewers who cast one, those fabricated values moved the team means this page now displays, bands, counts and sorts by. A pending edit is now a `PrioritizationBallotEdit` carrying only the fields the reader set, so an omitted axis reads as "leave it alone". **Storage and keying are unchanged and no backend file is touched** — the partial body is what PATCH was already built for (`_ballot_update_kwargs` assigns only the axes an entry carries), which is why `pytest lambda/api/test -k "prioritization or ballot or aggregate"` needed no change and still passes. That contract is pinned from both ends by tests that already existed: `test_a_partial_entry_does_not_disturb_the_axes_it_omits` and `test_an_omitted_axis_is_absent_from_the_update_expression` in `lambda/api/test/test_projects_prioritization_ballots.py` hold the route, and `still saves only the caller own ballot from the expanded row` plus `never writes an axis the reader did not set, on a first partial ballot` hold what the page sends.

[...]

PR #345, opened 2026-08-18
fix(prioritization): honest partial ballots (#342, #343)

### LINKED ISSUE #343: [Bug]: Sliders you don't touch show 3 but save as 0, and the row is ranked Low Priority on the zeros
## What happens

On the Prioritization page, move ONE slider and save. The three sliders you did not touch keep showing **3** on screen, but the ballot records them as **0** — and the row is then ranked on the zeros.

Reproduced on production 2026-08-18 (signed in, a row with no prior ballot):

1. Row editor shows the four axes at their default: Impact 3, Time to Market 3, Strategic Fit 3, Confidence 3. Each slider is `min=1 max=5`.
2. Drag **Impact** 3 → 4. Leave the other three alone. Click **Save Scores**.
3. The request carries only the axis that moved:
   ```
   PATCH /projects/prioritization
   {"scores":{"row_..._default":{"row_id":"row_..._default","impact":4}}}
   ```
4. The read-back answers `impact 4.0, time_to_market 0.0, confidence 0.0, strategic_fit 0.0`.
5. The row header now reads **4.0 Impact · 0.0 TTM · Team Score 1.6 · Low Priority**, and the summary says "Team average 1.6 of 5".
6. **Reload the page and reopen the row: the sliders read 4 / 3 / 3 / 3 again.**

So the editor and the record disagree permanently, and the editor is the one showing the friendlier number. A reviewer has no way on screen to discover that three quarters of their ballot is stored as zero. They entered something that looks like 4/3/3/3 (a mid-to-good proposal) and the product ranks it **Low Priority** on 4/0/0/0.

Note `0` is not a value the slider can even express — its minimum is 1. The stored ballot holds a number the control that wrote it cannot represent.

## Why the pieces are each defensible and the whole is not

Every individual half of this is deliberate, documented and test-pinned. The bug is that nobody reconciled the third participant.

- **The partial write is correct and was itself a fix.** `updateScore` records only the axis that moved, because seeding the edit from the defaults meant moving one slider saved all four axes — dragging the *team's* confidence and strategic-fit means toward zero for everybody. The verb is PATCH, so an omitted axis means "leave it alone", and the stored item genuinely has no `time_to_market` attribute at all. (`Prioritization.tsx`, the `localEdits` comment; `_ballot_update_kwargs` in `projects_handler.py` skips uncarried axes.)
- **The read filling absent axes with 0.0 is deliberate**, and the docstring argues the case: seeding an absent axis from the frontend default "would put a number nobody entered into a field named for what a reviewer scored, and the aggregate would then have no way to tell it from a vote". The aggregate accordingly asks `_carries_axis` and skips what was never scored, so the team *mean* stays honest.
- **The editor coercing a stored 0 to a displayed 3 is the unreconciled part.** `PRFAQRow.tsx` renders `value={score.impact === 0 ? 3 : score.impact}` on all four sliders. That coercion is display-only — `onChange` fires only on a real drag — so the 3 on screen is never written anywhere, and it is indistinguishable from a genuinely stored 3.

The backend docstrings already predict this divergence, but only for one axis: they note `time_to_market`'s frontend default is 3 while its absent-read is 0.0. In practice it hits **all four**, because `DEFAULT_SCORE` zeros the other three and the row coerces every zero to 3. `DEFAULT_SCORE` is also internally inconsistent — `time_to_market: 3` beside three `0`s — so there are two independent sources of the number 3 that agree only by accident.

And `_composite` weighs an absent axis as 0 even though the mean skips it, which is what turns a one-axis ballot into "Low Priority" rather than "partially scored".

## Not introduced by #341, but #341 makes it much easier to hit

Both halves are byte-identical in `development`. The dirty-only send arrived with #338 (`c27b21b`), the `=== 0 ? 3` display coercion with #132 (`fbeef53`). #341 touches neither beyond the `document_id` → `row_id` rename.

What #341 changes is the ballot key, so every row is a **first-ever ballot** again — and a first-ever partial save is precisely the case that materialises the zeros. There is no prior ballot to fall back on, so nothing absorbs the omission. The defect was always reachable; after #341 it is the default path for every row until someone has moved all four sliders at least once.

## What it should do

The fix belongs on the **display** side, not the write side — seeding defaults into the write is the thing the backend docstrings correctly argue against, and it is what #338 removed.

An axis nobody has scored should say so, the way the team summary already renders an em dash for an unscored row, rather than borrowing a number from the middle of the range. Options, cheapest first:

- Render an unscored axis as "not scored" (or an em dash) with the slider in an explicitly unset state, so saving one axis cannot imply values for the other three.
- If a slider must show a handle position, keep the 3 as the *position* but mark the axis unset until touched, and have the row header and composite treat it as unscored rather than as 0.
- Make `DEFAULT_SCORE` internally consistent while doing it, so `time_to_market` stops being the one axis whose default disagrees with the other three.

## Acceptance

- [ ] Saving one axis never causes the other three to display a number the record does not hold.
- [ ] An axis nobody scored is visually distinct from one scored at the bottom of the range, in the editor AND in the row header.
- [ ] A row with a one-axis ballot does not read as "Low Priority" purely because three axes are unscored.
- [ ] Reopening a row after a partial save shows the same values the API returns for it.
- [ ] A test fails if the editor's displayed value for an unscored axis diverges from what the record holds — the assertion nothing currently makes.

## Affected files

`voc-datalake/frontend/src/pages/Prioritization/PRFAQRow.tsx` (the `=== 0 ? 3` coercion), `prioritizationUtils.ts` (`DEFAULT_SCORE`), `Prioritization.tsx` (the row header's axis rendering), and `voc-datalake/lambda/api/projects_handler.py` only if `_composite`'s treatment of an absent axis is changed alongside.

Related: #339 (row identity), #342 (the write-side row-existence gap found in the same live pass).

perrozzi:
Fixes the two bugs found while live-verifying #341, both filed from production reproductions: #343 (the slider display defect) and #342 (the phantom-row ballot write). One PR because they are one theme — **a partial ballot must read honestly at every layer** — with zero file overlap between the halves: #342 is backend-only, #343 is frontend-only.

## #343 — an unscored axis reads as unscored, not as a 3 or a 0

Move one slider and save. The record correctly held only the touched axis; the API read the other three back as `0.0` (its documented encoding of "absent"); the editor painted every stored 0 as a 3, before and after reload; and the row banded **Low Priority** off a composite that weighed the zeros as votes. A ballot that looked like 4/3/3/3 on screen was 4/0/0/0 in the record and ranked on the zeros — with no way to discover it, since the painted 3 was indistinguishable from a stored 3 and the stored 0 isn't even in the slider's 1–5 range.

Three individually-documented decisions composed into this, and only the third was wrong: the dirty-only PATCH body (deliberate — #338, itself a fix for untouched defaults dragging the whole team's means down), the absent-reads-as-0.0 rule (argued in the handler's docstrings — inventing a default would be indistinguishable from a vote), and the display coercion (`score.x === 0 ? 3`, #132) that no docstring reconciled. So the fix is **display-side** and the wire contract is untouched:

- **`ScoreSlider` renders a 0 as NOT SCORED**: dash badge, muted handle, `aria-valuetext="Not scored"`. The handle rests mid-track (a range input cannot be valueless) but no number is presented as the axis's value. Releasing the pointer on an unscored slider commits the resting value, so "exactly 3" stays reachable by a click; the first arrow key covers keyboard users.
- **`DEFAULT_SCORE` is 0 on every axis.** Its `time_to_market` was 3 beside three 0s — two unrelated sources of the number 3 that agreed only by accident — and `ownAxis` degraded an unreadable stored TTM to a real-looking mid vote. One shared sentinel now, everywhere.
- **The team composite covers what the team expressed**, weights renormalised over the expressed axes — derived by evaluating `calculatePriorityScore` on indicators, so the weights lockstep test still guards a single copy of the numbers. Impact 4 alone now reads **4.0 / High Priority** beside its reviewer count, not 1.6 / Low from three zeros nobody entered. An axis nobody scored is `null` and prints the same decorative dash the unscored row already uses; the sort ties on it rather than ranking "nobody mentioned TTM" below "the team rated it worst"; a notes-only ballot bands **Not Scored** rather than Low.

A fully-scored ballot's numbers are unchanged (expressed weights sum to 1 — pinned by a regression test), the PATCH body construction is untouched, and the anonymous Vote page is unaffected (it always sends all four axes as real 1–5 values; its `NEUTRAL = 3` comment updated to say how the two pages now differ, and why). New i18n key `scores.notScored` in all eight catalogues, sorted position.

## #342 — a ballot for a row that does not exist is refused, not stored

`PATCH /projects/prioritization` accepted a ballot naming a row with no row record and answered `200 {"updated_count": 1}`; every read then discarded it. The writer was told their vote saved while it appeared nowhere — the same silent-loss class #341 closed on the read side, still open on the write side, and verified against production (the discard-warning count rose and fell with an injected orphan).

Both write paths now check the row record exists **before any write**, by keyed read, and refuse with **404** (the world changed, not the request) without echoing the id:

- The PATCH checks every named row as part of the existing up-front validation pass, so one phantom row in a multi-row body persists **nothing** — the "nothing malformed can half-persist" promise gets stronger, not narrower.
- `POST /voting-sessions` checks at session creation: a session is a public write window onto its row, and one opened for a phantom row collects a room's unrepeatable ballots. Submit inherits safety from create — the row id comes from the stored session, never the body, and nothing can delete a row today.
- A **failed** existence read raises rather than answering either way: "missing" refuses a legitimate save over a throttle; "present" waves through the orphan the check exists to refuse.

Deliberately a read-then-write, not a transaction: the check-to-write race requires a concurrent row deletion, which no code path can perform until phase 2 of #339. That phase owes the DB-enforced condition (issue #342's atomicity criterion stays open for it), and the check added here is where the condition goes. The `ROW#` prefix the ballots handler now reads is pinned against the projects handler by the anon-ballot lockstep suite, so the two spellings cannot drift.

## Tests

Every new behaviour has a test that fails on revert, verified by mutation (check removed → the refusal tests answer 200; coercion restored → the editor-honesty test sees no `aria-valuetext`):

- Backend: phantom row refused with nothing stored (single and multi-row), failed existence read → 500 with nothing stored, session-open refusal + its 500 twin, the `ROW#` lockstep pin. The two tests that pinned "a save reads no rows" now pin the precise claim — one existence read per named row, and none of them the migration's — so a regression in either direction (check removed, or the migration reading blindly) fails on the count.
- Frontend: renormalisation (one axis → that axis's value; two axes → their weights; all four → unchanged 3.9), null axes and null composite, notes-only bands Not Scored, one-axis ballot bands by its axis, and the page-level pin of #343's acceptance nothing asserted before: **the editor never displays a number the record does not hold** (stored 4/0/0/0 → slider 4 plus three "Not scored", not 4/3/3/3).

## Gates

Backend pytest **2569** · frontend vitest **2941** (181 files) · CDK **250** · stream **339** · `tsc` clean (app + CDK + stream) · ESLint clean · ruff **523 = base** · i18n: one added key, present in all 8 catalogues in sorted position, one-line diff per file.

Not deployed yet — deploy + live verification before merge, per convention. Closes #343; #342 can close too if the phase-2 deferral of the atomic condition is acceptable there, otherwise it stays open scoped to that criterion.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/prioritization.json
+++ b/voc-datalake/frontend/public/locales/de/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "Auswirkung",
     "impactDescription": "Geschäftswert und Kundennutzen",
     "low": "Niedrig",
+    "notScored": "Nicht bewertet",
     "slow": "Langsam",
     "strategicFit": "Strategische Passung",
     "strategicFitDescription": "Übereinstimmung mit Unternehmenszielen",
--- a/voc-datalake/frontend/public/locales/en/prioritization.json
+++ b/voc-datalake/frontend/public/locales/en/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "Impact",
     "impactDescription": "Business value and customer benefit",
     "low": "Low",
+    "notScored": "Not scored",
     "slow": "Slow",
     "strategicFit": "Strategic Fit",
     "strategicFitDescription": "Alignment with company goals",
--- a/voc-datalake/frontend/public/locales/es/prioritization.json
+++ b/voc-datalake/frontend/public/locales/es/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "Impacto",
     "impactDescription": "Valor comercial y beneficio para el cliente",
     "low": "Bajo",
+    "notScored": "Sin puntuar",
     "slow": "Lento",
     "strategicFit": "Ajuste estratégico",
     "strategicFitDescription": "Alineación con los objetivos de la empresa",
--- a/voc-datalake/frontend/public/locales/fr/prioritization.json
+++ b/voc-datalake/frontend/public/locales/fr/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "Portée",
     "impactDescription": "Valeur commerciale et bénéfice client",
     "low": "Faible",
+    "notScored": "Non noté",
     "slow": "Lent",
     "strategicFit": "Adéquation stratégique",
     "strategicFitDescription": "Alignement avec les objectifs de l'entreprise",
--- a/voc-datalake/frontend/public/locales/ja/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ja/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "インパクト",
     "impactDescription": "ビジネス価値と顧客メリット",
     "low": "低",
+    "notScored": "未採点",
     "slow": "遅い",
     "strategicFit": "戦略的適合性",
     "strategicFitDescription": "会社の目標との整合性",
--- a/voc-datalake/frontend/public/locales/ko/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ko/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "영향도",
     "impactDescription": "비즈니스 가치 및 고객 이점",
     "low": "낮음",
+    "notScored": "미채점",
     "slow": "느림",
     "strategicFit": "전략적 부합도",
     "strategicFitDescription": "회사 목표와의 부합도",
--- a/voc-datalake/frontend/public/locales/pt/prioritization.json
+++ b/voc-datalake/frontend/public/locales/pt/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "Impacto",
     "impactDescription": "Valor comercial e benefício ao cliente",
     "low": "Baixo",
+    "notScored": "Sem pontuação",
     "slow": "Lento",
     "strategicFit": "Alinhamento Estratégico",
     "strategicFitDescription": "Alinhamento com objetivos da empresa",
--- a/voc-datalake/frontend/public/locales/zh/prioritization.json
+++ b/voc-datalake/frontend/public/locales/zh/prioritization.json
@@ -129,6 +129,7 @@
     "impact": "影响力",
     "impactDescription": "业务价值和客户收益",
     "low": "低",
+    "notScored": "未评分",
     "slow": "缓慢",
     "strategicFit": "战略契合度",
     "strategicFitDescription": "与公司目标的一致性",
--- a/voc-datalake/frontend/src/pages/Prioritization/PRFAQRow.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/PRFAQRow.tsx
@@ -55,6 +55,30 @@ function unscoredLabel(kind: TeamView['kind'], t: TFunction): string {
 }
 
 /**
+ * One team number as the row prints it: the value to one decimal, or the dash
+ * for an axis nobody scored (`null`). One renderer, because the three chips
+ * printing independently is how a dash and a `0.0` could otherwise coexist for
+ * the same absence. The dash itself is decorative (the treatment every dash on
+ * this page gets); the state is carried by visually-hidden text using the same
+ * catalogue key the slider's `aria-valuetext` uses, so a screen reader hears
+ * "not scored" where a sighted reader sees the dash — not silence.
+ */
+function TeamNumber({ value, notScoredLabel }: {
+  readonly value: number | null
+  readonly notScoredLabel: string
+}): ReactElement {
+  if (value === null) {
+    return (
+      <>
+        <span aria-hidden="true" className="text-gray-300">—</span>
+        <span className="sr-only">{notScoredLabel}</span>
+      </>
+    )
+  }
+  return <>{value.toFixed(1)}</>
+}
+
+/**
  * The resting row's numbers: what the TEAM said about this document.
  *
  * Not the caller's own ballot, which used to be here and now lives behind the
@@ -83,19 +107,6 @@ function unscoredLabel(kind: TeamView['kind'], t: TFunction): string {
  * ballot produces a mean equal to that ballot and a spread of zero: without the
  * count, "one person looked" is indistinguishable from "we agree".
  */
-/**
- * One team number as the row prints it: the value to one decimal, or the
- * decorative dash for an axis nobody scored (`null`). One renderer, because the
- * three chips printing independently is how a dash and a `0.0` could otherwise
- * coexist for the same absence.
- */
-function TeamNumber({ value }: { readonly value: number | null }): ReactElement {
-  if (value === null) {
-    return <span aria-hidden="true" className="text-gray-300">—</span>
-  }
-  return <>{value.toFixed(1)}</>
-}
-
 function TeamScoreSummary({ team }: { readonly team: TeamView }): ReactElement {
   const { t } = useTranslation('prioritization')
   if (team.kind !== 'scored') {
@@ -129,11 +140,11 @@ function TeamScoreSummary({ team }: { readonly team: TeamView }): ReactElement {
           0.0 for it, and printing that ranked "nobody mentioned time to market"
           as "the team rated it worst" (#343). */}
       <div className="text-center">
-        <div className="text-base sm:text-lg font-bold text-blue-600"><TeamNumber value={scored.displayImpact} /></div>
+        <div className="text-base sm:text-lg font-bold text-blue-600"><TeamNumber value={scored.displayImpact} notScoredLabel={t('scores.notScored')} /></div>
         <div className="text-xs text-gray-500">{t('scores.impact')}</div>
       </div>
       <div className="text-center">
-        <div className="text-base sm:text-lg font-bold text-purple-600"><TeamNumber value={scored.displayTimeToMarket} /></div>
+        <div className="text-base sm:text-lg font-bold text-purple-600"><TeamNumber value={scored.displayTimeToMarket} notScoredLabel={t('scores.notScored')} /></div>
         <div className="text-xs text-gray-500">{t('sort.ttm')}</div>
       </div>
       <div className="text-center px-2 sm:px-3 py-1 bg-gray-50 rounded-lg">
@@ -142,7 +153,7 @@ function TeamScoreSummary({ team }: { readonly team: TeamView }): ReactElement {
             two roundings of it. Null — a ballot that expressed no axis at all,
             only a note — prints the dash, and the band beside the title reads
             "Not Scored" off the same null. */}
-        <div className="text-lg sm:text-xl font-bold text-green-600"><TeamNumber value={scored.displayComposite} /></div>
+        <div className="text-lg sm:text-xl font-bold text-green-600"><TeamNumber value={scored.displayComposite} notScoredLabel={t('scores.notScored')} /></div>
         {/* Labelled as the TEAM's score, not "Score": this number changed meaning
             from "my composite" to "the team's mean composite", and a row a reader
             cannot attribute is worse than either alone. */}
--- a/voc-datalake/frontend/src/pages/Prioritization/PRFAQRow.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/PRFAQRow.tsx
@@ -83,6 +83,19 @@ function unscoredLabel(kind: TeamView['kind'], t: TFunction): string {
  * ballot produces a mean equal to that ballot and a spread of zero: without the
  * count, "one person looked" is indistinguishable from "we agree".
  */
+/**
+ * One team number as the row prints it: the value to one decimal, or the
+ * decorative dash for an axis nobody scored (`null`). One renderer, because the
+ * three chips printing independently is how a dash and a `0.0` could otherwise
+ * coexist for the same absence.
+ */
+function TeamNumber({ value }: { readonly value: number | null }): ReactElement {
+  if (value === null) {
+    return <span aria-hidden="true" className="text-gray-300">—</span>
+  }
+  return <>{value.toFixed(1)}</>
+}
+
 function TeamScoreSummary({ team }: { readonly team: TeamView }): ReactElement {
   const { t } = useTranslation('prioritization')
   if (team.kind !== 'scored') {
@@ -111,19 +124,25 @@ function TeamScoreSummary({ team }: { readonly team: TeamView }): ReactElement {
           order two rows that show the same number. One rounding, shared — the rule
           `displayComposite` already follows. Captions are `text-gray-500` (4.63:1 on this
           white row) rather than `text-gray-400` (2.49:1), which is under AA at this size. */}
+      {/* An axis nobody scored is `null` and prints as the same decorative dash
+          the non-scored branch above uses — never a number. The backend reports
+          0.0 for it, and printing that ranked "nobody mentioned time to market"
+          as "the team rated it worst" (#343). */}
       <div className="text-center">
-        <div className="text-base sm:text-lg font-bold text-blue-600">{scored.displayImpact.toFixed(1)}</div>
+        <div className="text-base sm:text-lg font-bold text-blue-600"><TeamNumber value={scored.displayImpact} /></div>
         <div className="text-xs text-gray-500">{t('scores.impact')}</div>
       </div>
       <div className="text-center">
-        <div className="text-base sm:text-lg font-bold text-purple-600">{scored.displayTimeToMarket.toFixed(1)}</div>
+        <div className="text-base sm:text-lg font-bold text-purple-600"><TeamNumber value={scored.displayTimeToMarket} /></div>
         <div className="text-xs text-gray-500">{t('sort.ttm')}</div>
       </div>
       <div className="text-center px-2 sm:px-3 py-1 bg-gray-50 rounded-lg">
         {/* The same rounded value the priority band beside the title classifies, so
             the printed number and the label describing it are one value rather than
-            two roundings of it. */}
-        <div className="text-lg sm:text-xl font-bold text-green-600">{scored.displayComposite.toFixed(1)}</div>
+            two roundings of it. Null — a ballot that expressed no axis at all,
+            only a note — prints the dash, and the band beside the title reads
+            "Not Scored" off the same null. */}
+        <div className="text-lg sm:text-xl font-bold text-green-600"><TeamNumber value={scored.displayComposite} /></div>
         {/* Labelled as the TEAM's score, not "Score": this number changed meaning
             from "my composite" to "the team's mean composite", and a row a reader
             cannot attribute is worse than either alone. */}
@@ -326,8 +345,12 @@ function TeamScorePanel({ team }: { readonly team: TeamView }): ReactElement {
       {score ? (
         <>
           <p className="text-sm text-indigo-900 mt-1">
+            {/* A null composite — a ballot that expressed no axis, only a note —
+                interpolates the same dash the chips print, keeping one key in
+                eight catalogues rather than a second sentence for a state the
+                reviewer count beside it already explains. */}
             {t('team.summary', {
-              score: score.displayComposite.toFixed(1),
+              score: score.displayComposite === null ? '—' : score.displayComposite.toFixed(1),
               reviewers: score.reviewerCount,
             })}
           </p>
@@ -439,10 +462,16 @@ function PRFAQRowExpanded({
               </p>
             ) : null}
           </div>
-          <ScoreSlider label={t('scores.impact')} value={score.impact === 0 ? 3 : score.impact} onChange={(v) => onUpdateScore('impact', v)} description={t('scores.impactDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
-          <ScoreSlider label={t('scores.timeToMarket')} value={score.time_to_market === 0 ? 3 : score.time_to_market} onChange={(v) => onUpdateScore('time_to_market', v)} description={t('scores.timeToMarketDescription')} lowLabel={t('scores.slow')} highLabel={t('scores.fast')} />
-          <ScoreSlider label={t('scores.strategicFit')} value={score.strategic_fit === 0 ? 3 : score.strategic_fit} onChange={(v) => onUpdateScore('strategic_fit', v)} description={t('scores.strategicFitDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
-          <ScoreSlider label={t('scores.confidence')} value={score.confidence === 0 ? 3 : score.confidence} onChange={(v) => onUpdateScore('confidence', v)} description={t('scores.confidenceDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
+          {/* The stored value goes in RAW: 0 means unscored (the API's own
+              contract) and `ScoreSlider` renders it as such. The old
+              `=== 0 ? 3` coercion here painted a mid-range number the record
+              did not hold, indistinguishable from a stored 3 — so a reviewer
+              whose partial save recorded three axes as nothing could never
+              discover it on screen (#343). */}
+          <ScoreSlider label={t('scores.impact')} value={score.impact} onChange={(v) => onUpdateScore('impact', v)} description={t('scores.impactDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
+          <ScoreSlider label={t('scores.timeToMarket')} value={score.time_to_market} onChange={(v) => onUpdateScore('time_to_market', v)} description={t('scores.timeToMarketDescription')} lowLabel={t('scores.slow')} highLabel={t('scores.fast')} />
+          <ScoreSlider label={t('scores.strategicFit')} value={score.strategic_fit} onChange={(v) => onUpdateScore('strategic_fit', v)} description={t('scores.strategicFitDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
+          <ScoreSlider label={t('scores.confidence')} value={score.confidence} onChange={(v) => onUpdateScore('confidence', v)} description={t('scores.confidenceDescription')} lowLabel={t('scores.low')} highLabel={t('scores.high')} />
           <div>
             <label className="text-sm font-medium text-gray-700">{t('notes.label')}</label>
             {/* Bounded, because the API REFUSES a longer note rather than
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -625,6 +625,18 @@ describe('Prioritization', () => {
 
       expect(screen.getByRole('button', { name: /save/i })).toBeDisabled()
       expect(mockPatchPrioritizationScores).not.toHaveBeenCalled()
+
+      // The stuck-flag variant: a press that STARTS on the slider and ends
+      // elsewhere must not leave the guard armed — without clearing it on
+      // pointerleave, the NEXT unrelated release over the input casts the
+      // resting value, the exact stray vote the guard exists to prevent, one
+      // interaction later.
+      fireEvent.pointerDown(sliders[1])
+      fireEvent.pointerLeave(sliders[1])
+      fireEvent.pointerUp(sliders[1])
+
+      expect(screen.getByRole('button', { name: /save/i })).toBeDisabled()
+      expect(mockPatchPrioritizationScores).not.toHaveBeenCalled()
     })
 
     it('says nobody has scored a document absent from the aggregate', async () => {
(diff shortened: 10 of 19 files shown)


perrozzi:

This PR does two things: (frontend) stops the prioritization editor from painting a stored `0` as a `3`, makes `DEFAULT_SCORE` an all-zero unscored sentinel, and renormalises the team composite over expressed axes with `null` for unexpressed ones; (backend) refuses `PATCH /projects/prioritization` and `POST /voting-sessions` with 404 when the named row record does not exist. The direction is right and the split (backend-only / frontend-only, wire contract untouched) is clean, with good test coverage and i18n in all 8 catalogues. Two things I'd fix before merge: the new `onPointerUp` commit path in `ScoreSlider`, and the consistency mode of the new existence reads.

### Blockers

- `frontend/src/pages/Prioritization/ScoreSlider.tsx` — `onPointerUp={unscored ? () => onChange(position) : undefined}` commits the *captured* `position` (always `3`), not where the pointer actually landed. On an unscored slider, a click on the track at 5 fires `change` first (`onChange(5)`), then `pointerup`; if the re-render that clears `unscored` has not flushed by then (parent state going through a store/mutation rather than plain `useState` is enough), the handler fires with the stale closure and rewrites the deliberate 5 back to 3 — the same "screen shows a number the user did not choose" class this PR is fixing. Read the DOM value instead: `onPointerUp={(e) => { if (unscored) onChange(e.currentTarget.valueAsNumber) }}`. Related: `pointerup` also fires when a pointer that went down elsewhere is released over the input, which casts a 3 with no press on the control — gate on a pointerdown-on-this-input flag. This new commit path also has no test; per the repo rule it needs one that fails when reverted (nothing in `Prioritization.test.tsx`/`ScoreSlider` tests covers "releasing the pointer on an unscored slider commits the resting value").
- `lambda/api/projects_handler.py` (`_missing_ballot_rows`) and `lambda/api/ballots_handler.py` (`_row_exists`) — both use default eventually-consistent `get_item` to gate a write. The 404 text itself names the failure mode ("a row created moments ago in another tab"), and that is exactly the case an eventually-consistent read can get wrong: a legitimate save/session-open refused with 404 right after row creation. Pass `ConsistentRead=True` on both existence reads (cost is negligible for a keyed read on a save path).

### Should fix

- Double read of the same key: for a scored save, `_missing_ballot_rows` reads `ROW#{row_id}` and then the legacy-migration path reads the same item again. Return the fetched item(s) from the existence pass and thread them into the migration branch, or use `batch_get_item` with a `ProjectionExpression` so the check is one round trip instead of N sequential ones (`api_patch_prioritization_scores`).
- No CDK/IAM change accompanies `ballots_handler` reading `ROW#` items. The docstring asserts the role already grants this single-item read, but the unit tests use `FakeAggregatesTable`, so a scoped grant (e.g. a `dynamodb:LeadingKeys` condition or a narrower action set) would only fail at deploy. Please confirm the grant covers it and, if the policy is key-scoped, add/extend a CDK assertion.
- `test_projects_prioritization_ballots.py::test_a_save_that_scored_nothing_pays_one_existence_read_and_removes_nothing` passes `seed_rows=False` and never asserts the status. If `self._table()` does not seed `ROW#row-p1`, the request now 404s and every assertion still passes (one existence read, legacy map untouched) — the test becomes vacuous about the behaviour it names. Add `assert status == 200` (and seed the row explicitly).
- Renormalisation changes cross-row comparability, not just per-row honesty: an impact-only ballot now composites to 4.0 / High and outranks a fully-scored 3.8 row on the same sorted list. That is defensible, but if the backend also computes a composite (the weights lockstep test implies a Python copy of the formula) it does not renormalise, so the displayed/sorted number and any backend-computed `priority_score`, export or filter now disagree for partial ballots. Please confirm there is no such consumer, or renormalise there too.
- `DEFAULT_SCORE.time_to_market` moving 3 → 0 retires a sentinel that `projects_handler._axis_value`'s old docstring says `PRFAQRow` read as "touched" (`time_to_market !== 3`). Grep for any remaining `=== 3` / `!== 3` axis comparisons in the prioritization page/utils; if the touched signal now comes only from the aggregate, good, but nothing in the diff proves it.

### Nits

- `PRFAQRow.tsx`: the new `TeamNumber` JSDoc was inserted *after* the pre-existing `TeamScoreSummary` doc block, so that block now dangles above `TeamNumber` and `TeamScoreSummary` is undocumented. Move the new function above the old comment.
- The dash is `aria-hidden` in both `TeamNumber` and the slider badge, so a screen-reader user gets silence for an unscored team axis/composite (the slider is covered by `aria-valuetext`, the chips are not). A `sr-only` `{t('scores.notScored')}` would use the key you already added in all locales.
- `compositeWeightOf` re-derives all four weights via `calculatePriorityScore` on every `getTeamScore` call; hoist to a module-level constant map derived once (still off the pinned formula, so the lockstep guarantee is unchanged).
- `compareByTeamScore` returning `0` whenever either side is `null` makes the comparator intransitive (null ties both a 5 and a 1 that do not tie each other), so ordering becomes engine-dependent. Sorting nulls to one end deterministically would be more predictable — same latent issue as the existing `!teamA || !teamB` branch.
- `_row_exists`'s `except ApiError: raise` is a no-op unless `_table()` can raise one; narrowing the `try` to the `get_item` call would say the same thing with less ceremony.

---

perrozzi:
## Disposition of round 1 — all items addressed in `9380735`

**Blocker 1 (ScoreSlider `onPointerUp` stale closure + stray release): fixed as suggested, both halves.** The commit now reads the value off the control at release (`Number(e.currentTarget.value)`), so a click-at-5 whose `change` fires first cannot be rewritten back to 3 by a closure over the previous render; and a `pressedHere` ref gates the commit to presses that started on the input, so a pointer that went down elsewhere and was released over the slider casts nothing. Two new tests cover exactly the two behaviours ("releasing a press on an unscored slider commits the resting value" and "a pointer released over an unscored slider after going down elsewhere casts nothing"); removing the `onPointerUp` path fails the first — verified by mutation, not asserted.

**Blocker 2 (eventually-consistent existence reads): fixed.** Both reads now pass `ConsistentRead=True`, with the rationale in a comment at each site: these reads gate writes, and the case they exist to distinguish — a row created moments ago — is precisely what an eventually-consistent read can miss.

**Double read of the row key: fixed, the threading option.** The existence pass (`_fetched_ballot_rows`, renamed from `_missing_ballot_rows` since it now returns the records) hands the fetched rows to the route, and `drop_for_row` takes the record instead of re-reading by id. One save reads each named row exactly once; the updated no-legacy-map test pins that count (`len(row_reads) == len(rows)`, "and not one more"). `batch_get_item` was considered and deliberately not used — the reasoning is in the helper's docstring (single-digit row counts per save, client-level unmarshalling plumbing this module otherwise never needs, and the loop stays inspectable by the existing fakes).

**IAM confirmation: checked, no change needed.** The ballots grant is `aggregatesTable.grant(ballotsRole, 'dynamodb:GetItem', 'dynamodb:PutItem', 'dynamodb:UpdateItem')` (api-stack.ts:445) — table-wide, not key-scoped, no `dynamodb:LeadingKeys` condition — and a `GetItem` with `ConsistentRead` is the same IAM action. The exact three-action set is already pinned by `api-stack.test.ts` ("holds only the three item actions..."), which is the assertion you asked for: it fails the moment the set changes in either direction.

**Vacuous-test risk on the scored-nothing test: fixed.** It now asserts `status == 200` (the fixture does seed `row-p1`, but the assertion is what keeps "one read, map untouched" from holding for the wrong reason if that ever changed), and its docstring says why.

**Cross-row comparability of the renormalised composite: confirmed no disagreeing consumer, and both docstrings now say why.** The backend's only composite (`_composite`) feeds `score_spread` exclusively, and the spread compares only fully-scored ballots (`_is_fully_scored`) — where the expressed weights sum to 1.0 and renormalisation is the identity. So the displayed/sorted number and every backend-computed value agree for every ballot that is ever compared across; partial ballots are exactly the ones the spread already excludes. `_composite`'s docstring (which argued against renormalising in general) and `expressedComposite`'s now each record the other side of this argument. On the ranking judgement itself — impact-only 4.0/High outranking a fully-scored 3.8 — that is the intended reading (the composite of what the team has said, beside a reviewer count that keeps "one person said one thing" visible), and it is the direction issue #343's acceptance asks for.

**`=== 3` / `!== 3` grep: confirmed empty.** No axis comparison against 3 survives anywhere in the frontend production code (the last one, `QuickScores`' `showTTM`, was already removed by #338; the fixture `time_to_market: 3` occurrences in tests are stored real values, not sentinels).

**Nits: all taken.** `TeamNumber` moved above the doc block it had orphaned; the team chips' dash now carries visually-hidden "not scored" text through the same catalogue key the slider uses (the unscored-row dash already had its visible label beneath it); the weight map is derived once at module level — as a literal record built from indicator evaluations, still zero copies of the numbers and no type assertion; the comparator's null-tie intransitivity is resolved structurally (a scored row with no number in the sorted column is grouped with the number-less block per row, before any comparison, so the null branch is unreachable through `sortRows` — new sort test pins the grouping in both directions); `_row_exists` narrowed (`_table()` outside the `try`, the no-op `ApiError` re-raise gone).

perrozzi:

Since my last review the author fixed both blockers (the `onPointerUp` commit now reads `e.currentTarget.value` and is gated by a `pressedHere` ref, with two new tests; both existence reads pass `ConsistentRead=True`), eliminated the double keyed read by returning the fetched items from `_fetched_ballot_rows` and threading them into `drop_for_row`, asserted the status in the previously-vacuous test, hoisted the composite weights, added the `sr-only` not-scored text, moved the `TeamNumber` JSDoc, and made null-in-sorted-column rows sort into a deterministic block. Good delta.

### Resolved
- `ScoreSlider.onPointerUp` no longer commits a stale `position`; reads the DOM value and ignores releases from a press that started elsewhere, both covered by new tests in `Prioritization.test.tsx`.
- Both existence reads are now strongly consistent (`projects_handler._fetched_ballot_rows`, `ballots_handler._row_exists`) — the "created moments ago" 404 is closed.
- Double read of `ROW#{row_id}` removed; `drop_for_row` takes the record, and the read-count tests were tightened to "one existence read per row, none of them the migration's".
- `test_a_save_that_scored_nothing_pays_one_existence_read_and_removes_nothing` now asserts `status == 200`.
- Cross-consumer comparability now documented in `_composite` (backend composites only fully-scored ballots, where renormalisation is the identity).
- Nits: `COMPOSITE_WEIGHT` hoisted, `sr-only` label, JSDoc placement, comparator-null determinism via `blockOf`, `_row_exists`'s no-op `except ApiError` gone.

### Still open
- No CDK/IAM confirmation or assertion for `ballots_handler` reading `ROW#` items (fakes hide a key-scoped grant; `ConsistentRead` itself needs no extra permission, the item read does).
- Nothing in the diff proves no remaining `=== 3` / `!== 3` axis comparisons after `DEFAULT_SCORE.time_to_market` moved 3 → 0.

### New findings
- `frontend/src/pages/Prioritization/ScoreSlider.tsx` — `pressedHere` is only cleared in `onPointerUp` on the input, so a press that *starts* on the slider and is released off it leaves the ref stuck `true`; a later unrelated release over the input then casts the resting value — the exact case the guard exists to prevent, just one interaction later. Clear it on `onPointerLeave`/`onLostPointerCapture` (or compare `e.pointerId` with the one recorded on pointerdown), and extend the stray-release test with a preceding down-on-slider/up-elsewhere sequence.
- `lambda/api/projects_handler.py` — `drop_for_row(row: Any)` changed from a row id to the record. Any direct unit test (or future caller) still passing a string will not fail loudly: `_row_document_ids('row-1')` reads it as holding nothing, so the migration silently becomes a no-op. Please confirm all callers/tests were updated, or add a type guard that logs when handed a non-dict.
- `frontend/src/pages/Prioritization/prioritizationUtils.ts` — `COMPOSITE_WEIGHT` is now evaluated at module load via `calculatePriorityScore`, which only works because that symbol is a hoisted `function` declaration; converting it to a `const` arrow later would turn this into a TDZ `ReferenceError` at import time. A one-line comment at `weightOf` noting the load-order dependency would keep that from being discovered by a blank page.
- `prioritizationUtils.test.ts` — the new "groups a row with no number in the sorted column" test asserts only `b` first and `a` after `b`; it never pins `a` and `c` (both in the unscored block) as stable/arrival-ordered, which is the property that makes the grouping deterministic. Worth one more assertion since determinism is the stated reason for the change.

---

perrozzi:
## Disposition of round 2 — in `f9808f8`

**Stuck press guard (new blocker-class finding): fixed as suggested.** `pressedHere` clears on `pointerleave`, `pointercancel` and `lostpointercapture` as well as on the up itself, and the stray-release test grew the down-on-slider / leave / later-stray-up sequence — it fails by mutation when the clearing handlers are removed. Legitimate drags are unaffected: a range input takes implicit pointer capture on press, so `pointerleave` does not fire mid-drag, `pointerup` still lands on the control, and `lostpointercapture` fires after it.

**`drop_for_row` non-dict concern: confirmed, no code change.** It has exactly one caller — the route loop — which passes `rows_by_id[row_id]`, a dict by construction (`_fetched_ballot_rows` only stores dicts). No test calls it directly (grep: two hits, the def and the one call site). A malformed record reading as holding-nothing is the module's standing defensive stance (`_row_document_ids`'s own docstring), and a log-on-non-dict guard would be defending against a call that cannot be made.

**TDZ load-order dependency: comment added at `weightOf`** naming the failure mode (a `ReferenceError` on import — a blank page, not a test failure) and the rule (the block stays below the formula).

**Dash-column grouping determinism: the test now pins the full order** `['b','a','c']` in both directions — the number-less block keeps arrival order and does not flip with the toggle.

**On the two 'still open' items:** both were answered in the round-1 disposition, restated for the record. (1) The ballots grant is table-wide `GetItem`/`PutItem`/`UpdateItem` with no key-scoping condition (api-stack.ts:445), `ConsistentRead` changes no IAM action, and the exact three-action set is already pinned by the `ballots Lambda IAM grants` suite in api-stack.test.ts — the assertion requested. (2) `grep -rnE '[!=]== 3' frontend/src/pages/{Prioritization,Vote}` over production code returns nothing; the last axis comparison against 3 (`QuickScores`' `showTTM`) was removed by #338, and the remaining `time_to_market: 3` occurrences in tests are stored real values, not sentinels.

perrozzi:

Since my last review the author added `pointerleave`/`pointercancel`/`lostpointercapture` clearing to `ScoreSlider`'s `pressedHere` guard (plus an extended stray-release test), tightened the null-block sort test to a full `toEqual(['b','a','c'])` in both directions, and added the load-order comment at `weightOf`. All four of my previous new findings are addressed except the `drop_for_row(row: Any)` caller/type-guard question, which I can't confirm from this diff.

### Resolved
- `pressedHere` no longer sticks after a press that starts on the slider and ends elsewhere; covered by the appended variant in `Prioritization.test.tsx`.
- The number-less sort block is now pinned as arrival-ordered and direction-stable (`prioritizationUtils.test.ts`).
- The module-load / TDZ dependency of `COMPOSITE_WEIGHT` on `calculatePriorityScore` is documented at `weightOf`.

### Still open
- No CDK/IAM confirmation or assertion that `ballots_handler`'s role can read `ROW#` items.
- Nothing proves no remaining `=== 3` / `!== 3` axis comparisons after `DEFAULT_SCORE.time_to_market` moved 3 → 0.
- `drop_for_row` now takes the row record instead of a row id; still no confirmation that every caller/test was updated, and a stray string argument degrades to a silent no-op.

### New findings
- `frontend/src/pages/Prioritization/ScoreSlider.tsx` (minor) — the guard now depends on event *ordering* that nothing in the suite exercises: jsdom implements neither implicit pointer capture nor `lostpointercapture`, so the parenthetical claim ("`pointerleave` does not fire during a legitimate drag … `pointerup` still arrives here") is unverified. Per spec `lostpointercapture` follows `pointerup`, but if any engine fires it first the click-at-the-resting-position commit silently stops working on touch. The `e.pointerId`-comparison variant I suggested earlier is ordering-independent and would remove the dependency; failing that, a one-line note that this path is browser-verified only would do.
- `frontend/src/pages/Prioritization/ScoreSlider.tsx` (nit) — the comment asserts `calculatePriorityScore` is "a `const` arrow" in `prioritizationUtils.ts`; please confirm that matches the actual declaration (I read it as a hoisted `function`), since the comment's whole value is being accurate about which form makes the order load-bearing.
- `Prioritization.test.tsx` (nit) — the stuck-flag case is appended to the existing stray-release `it` rather than its own test, so a failure names the wrong scenario; splitting it makes the regression self-describing.

---

perrozzi:
## Disposition of round 3 — in `fa7063f`

**Ordering dependence: resolved by REMOVING the one risky clear rather than documenting it.** `onLostPointerCapture` is gone — per spec it fires after the up (so it added nothing to `pointerup`'s own clear), and an engine firing it early is exactly the ordering that would have broken the click-at-the-resting-position commit on touch. The two remaining clears are ordering-safe by construction: `pointercancel` replaces an up, and a `pointerleave` after an up runs after the commit already happened. The jsdom caveat now sits in the comment where the handlers are: the tests pin the guard's logic; the implicit-capture drag model is browser-verified only. (The `pointerId` variant would still leave the same jsdom gap — it trades an ordering dependence for identity bookkeeping across renders without removing the untested half, so the smaller change won.)

**`calculatePriorityScore` declaration form: confirmed as written.** `prioritizationUtils.ts:100` reads `export const calculatePriorityScore = (score: CompositeAxes): number => {` — a `const` arrow, not a hoisted `function`, which is what makes the order load-bearing and the TDZ comment accurate.

**Stuck-flag test split into its own `it`** ('a press that starts on an unscored slider and ends elsewhere does not arm the next stray release'), so a failure names its scenario.

**The three 'still open' carried items are answered dispositions, not open work** — restating with the evidence pointers since follow-up mode diffs code and cannot see them there: (1) ballots IAM: the grant is table-wide `GetItem/PutItem/UpdateItem` with no key-scoping (api-stack.ts:445), and the exact set is pinned by the `ballots Lambda IAM grants` suite — the requested assertion exists and predates this PR. (2) axis-sentinel grep: zero `=== 3`/`!== 3` matches in frontend production code; the last one died with #338. (3) `drop_for_row` callers: exactly one, the route loop, passing a dict by construction; no test calls it directly.

perrozzi:

Since my last review the author dropped the `onLostPointerCapture` clear, replaced the parenthetical with an explicit ordering-safety + jsdom caveat note, and split the stuck-flag case into its own `it`. Both of my previous new-finding nits on `ScoreSlider`/`Prioritization.test.tsx` are addressed; the `weightOf` "const arrow" wording nit and the three carried-over items are untouched.

### Resolved
- `lostpointercapture` removed, so the commit path no longer depends on engine event ordering; the residual jsdom/capture gap is now documented in `ScoreSlider.tsx`.
- The stuck-flag regression is its own self-describing test in `Prioritization.test.tsx`.

### Still open
- No CDK/IAM confirmation or assertion that `ballots_handler`'s role can read `ROW#` items.
- Nothing proves no remaining `=== 3` / `!== 3` axis comparisons after `DEFAULT_SCORE.time_to_market` moved 3 → 0.
- `drop_for_row(row: Any)` — no confirmation every caller/test passes the record, and a stray string degrades to a silent no-op.
- `prioritizationUtils.ts` (nit) — the `weightOf` comment still asserts `calculatePriorityScore` is "a `const` arrow"; please confirm that matches the declaration, since the comment's value is being right about which form makes load order load-bearing.

### New findings
- `frontend/src/pages/Prioritization/ScoreSlider.tsx` (nit) — with `pointerleave` as the only off-control clear, an engine that does *not* take implicit capture (some pen/mouse paths) will drop a legitimate press that drifts one pixel off and releases back over the input: the flag is already cleared, so the click-at-the-resting-position commit silently does nothing. The comment covers the capture caveat but not this direction; an `e.pointerId`-matched clear would be exact if you want to close it.

Otherwise the delta is clean.

---

perrozzi:
## Disposition of round 4 — no code change, and stopping here per the nit-only rule

**The drift-off-and-back nit: the failure direction is the designed one, and the suggested fix is weaker for the pointer type it targets.** If an engine without implicit capture clears the flag on a one-pixel drift, the worst case is that a click-at-the-resting-position does nothing — the reader clicks again and it works, and any actual drag scores through `onChange` regardless of the flag. The guard can only ever fail toward a no-op, never toward casting a vote, and that asymmetry is the design: phantom data is unrecoverable, a dead click is not. `pointerId` matching would not close this: a mouse keeps ONE pointer id across its whole lifetime, so an id recorded on some earlier press still matches a later stray release — it is exact for touch (fresh id per contact, where implicit capture already prevents the leave) and indistinguishable from the current flag for the mouse paths this nit is about.

**The four carried 'still open' items are all answered dispositions with evidence, not open work** — the general rule, since follow-up mode diffs code and cannot see comment history: every one of them was checked and either fixed or confirmed against the source in rounds 1–3. Specifically: the ballots IAM set is pinned by an existing CDK assertion (api-stack.ts:445 + the `ballots Lambda IAM grants` suite); the axis-sentinel grep over frontend production code returns zero matches; `drop_for_row` has one caller passing a dict by construction; and `calculatePriorityScore` is verifiably `export const calculatePriorityScore = (score: CompositeAxes): number => {` at prioritizationUtils.ts:100 — a const arrow, exactly as the `weightOf` comment states.

Gates at `fa7063f`: backend pytest 2569 · frontend vitest 2945 · CDK 250 · stream 339 · tsc/eslint clean · ruff 523 = base · CI green. Next step: deploy + live verification before merge, per convention.

perrozzi:
## Deployed and live-verified at `fa7063f`

`VocApiStack --exclusively`, 185s, `UPDATE_COMPLETE` 18:45:58Z. Right slot proven: bare `VocApiStack` only, all five `b-*` timestamps byte-identical before and after, no other unprefixed stack moved. API suite **34/34**. Zero `ERROR`/`Traceback`/timeout in either touched Lambda across the whole exercise.

**One thing worth flagging about the slot rather than the PR:** live was last deployed at 15:59Z from a tree that did **not** contain merged work this branch's base does (`McpUsagePlan`, the `McpLambdaRole` `BatchGetItem` grant), so the diff carried those forward alongside my Lambda code. Confirmed they are development's, not mine — my diff touches no CDK file and greps clean for `mcp` — and the owner approved overwriting. Worth knowing that the unprefixed slot had drifted behind `development`.

**Bundle proof.** Pre-deploy controls on both Lambdas showed zero for every needle (with `lambda_handler` as a positive control at 6 files, so the grep was known-working). Post-deploy both code shas changed and every needle is present: `_fetched_ballot_rows`, `ConsistentRead`, the 404 wording, `_row_exists`, `ROW_SK_PREFIX`. Frontend: the served entry chunk matches the local build by name, and the lazy Prioritization chunk that carries the change is **md5-identical** to the local build (`a1a5235a…`) and contains `aria-valuetext`, `scores.notScored` and `sr-only`. The new locale key serves from CloudFront.

### #342 — all three refusals fire, and nothing is stored

| Case | Result |
|---|---|
| `PATCH` a phantom row | **404**, "scores name a row that does not exist…" |
| `PATCH` one phantom **beside a real row** | **404**, and the real row's ballot **was not written** |
| `POST /voting-sessions` on a phantom row | **404**, "that prioritization row does not exist…" |
| `PATCH` a real row | **200** `updated_count: 1` — happy path intact |

The partition after all of it holds no ghost ballot and no ghost session, and the discard warning stayed **flat at 2 ballots / 1 row id** — the pre-existing orphans, with none added. That flatness is the point: before this change the same three requests would have created orphans and pushed the count up.

### #343 — the production reproduction inverts

The one-axis ballot that produced the bug report (`impact 4`, nothing else) now renders:

**`4.0 Impact · — Not scored TTM · 4.0 Team Score · 1 Reviewers · High Priority`**

Before the fix, that identical record read `4.0 Impact · 0.0 TTM · 1.6 Team Score · Low Priority`. The stats cards moved with it (1 High Priority, 1 Not Scored). The wire contract is untouched — the API still returns `time_to_market: 0.0` for the unexpressed axes, verified in the browser's own response — so the whole change is in how the page reads that 0.

In the editor, the scored axis shows its number with no `aria-valuetext`, and each unscored axis shows an em-dash badge with `aria-valuetext="Not scored"`. The stored record for that ballot carries **only** `impact` (no `time_to_market` attribute at all), so the editor and the record now agree, which is the acceptance criterion nothing asserted before.

**Both halves of the pointer guard verified in a real browser** — the part jsdom cannot cover:

- A release with no preceding press on the control leaves the value untouched, keeps "Not scored", and leaves Save **disabled**.
- A deliberate press on the control commits the resting value and enables Save, so "exactly 3" stays reachable by click.

That second path also explains a transient I chased: an early click on the row header appeared to leave TTM showing `3`, because the expansion re-rendered and put the slider under the cursor between press and release — the commit path doing its job, not a defect. It did not reproduce once the layout differed, the record was never written, and reloading discarded it.

All test artifacts removed; the partition is back to exactly its pre-test state (the two abandoned legacy orphans, the two row records, the empty legacy `SCORES`).

Ready for approval and merge.

PR #346, opened 2026-08-18
chore: merge development into main (main currently cannot build)

perrozzi:
Promotes 121 commits of `development` work onto `main`, with the dependency conflicts resolved and every local gate run.

## Why this is worth doing now

`main` cannot build, and has not been able to since the Vite 8 bump landed on 2026-07-07. Verified on a clean worktree of `main` at its own lockfile (`npm ci` clean, vite 8.0.16 / vitest 4.1.9):

| Gate on `main` as it stands | Result |
|---|---|
| `npm run build` | fails — TypeScript rejects `vite.config.ts` |
| `npx vite build` (bundler alone) | fails — `TypeError: manualChunks is not a function` |
| frontend tests | 5 failed / 1393 passed |
| lint | passes |

Vite 8 is declared in `package.json` but used nowhere: `vite.config.ts` still passes `manualChunks` as an object, the form Rolldown removed, and no commit has ever touched that file. Since `main` is the repository's default branch, a fresh clone following the documented build currently fails.

## What conflicted, and how it was resolved

`main` is 10 commits ahead of the merge base and **all 10 are dependabot bumps touching only the two `package.json` and two `package-lock.json` files**. No source file conflicted.

- `voc-datalake/package.json` — kept `development`'s side. Its `aws-cdk-lib` 2.261.0 supersedes main's 2.259.0, so taking main's side would have been a downgrade.
- Both `package-lock.json` — **regenerated with `npm install`, not hand-merged.** The contested transitive packages (`fast-xml-parser`, `js-cookie`, `@aws-sdk/xml-builder`) are declared nowhere, so regeneration is the only correct way to settle them.
- `voc-datalake/frontend/package.json` — auto-merged cleanly, taking main's newer ranges.

One trap worth recording for future dependency merges: the frontend manifest **auto-merges without a conflict** and silently resolves in main's favour, because `development` never edited those lines. Only the lockfile conflicts. Resolving that lockfile toward `development` therefore produces a manifest demanding Vite 8 against a lock pinning Vite 7 — `npm ci` refuses it and `npm install` quietly upgrades instead. `npm ci` is the cheap detector: it failed with `EUSAGE` before regeneration and passes on both projects now.

## Two deliberate version decisions

**Vitest 3 → 4 is taken.** It cost exactly one test-file change and nothing in `src/`. The Cognito mocks used arrow-function implementations (`vi.fn().mockImplementation(() => ({ ... }))`) for values the source calls with `new`. Vitest 4 honours `new` with real construct semantics and an arrow function is not constructible; Vitest 3 invoked the implementation as a plain call even under `new`, so it worked there by accident. Roughly 17 test files use that pattern, but only the `amazon-cognito-identity-js` mocks are ever constructed by source, which is why the blast radius is a single file.

**Vite 8 is not taken; Vite is held at 7.x** (7.3.6, still ahead of development's 7.3.2). Vite 8 replaces the bundler with Rolldown, which requires `manualChunks` to be a function. Migrating the six vendor chunks to path matching changes chunk layout and therefore every content hash, which matters for build-versus-deployed comparison. Vitest 4 peer-supports Vite 6, 7 and 8, so the two upgrades are separable and there is no reason to couple them here.

## Verification

All run locally on this branch's exact tree:

| Gate | Result |
|---|---|
| frontend tests | **2949 passed** (181 files) |
| frontend typecheck | pass |
| frontend build | pass (4.26s) |
| frontend lint | pass (`--max-warnings 0`) |
| CDK tests | **256 passed** (17 files) |
| `npm ci --dry-run`, both projects | pass — manifests and locks are in sync |

Baseline for comparison: `development` before the merge is 181 files / 2949 tests green on Vite 7.3.2 / Vitest 3.2.4. The count is unchanged, so nothing was silently skipped.

A losslessness control: diffing this branch against `development` while excluding dependency metadata returns **exactly one file** — the mock fix — so none of development's 121 commits was dropped. The CDK `package.json` and lockfile end byte-identical to `development`.

## Not covered here

- No deployment was performed and no AWS resources were touched; this is a source-only merge.
- Python/backend tests were not re-run because no Python file differs between the branches.
- Dependabot alerts on the default branch should be re-checked after this merges, since the frontend dependency set moves.

## Suggested follow-ups

- **Vite 8 / Rolldown migration** — convert `manualChunks` to function form, expect bundle hashes to change.
- **Add a build-and-test workflow.** `.github/workflows/` is empty on `main`, which is the reason a bundler-swapping major merged as a metadata-only change with nothing to fail. Without a gate this will recur.


perrozzi:
## CodeQL disposition — 3 alerts dismissed, reasoning below

All three alerts are **pre-existing `development` code, not regressions from this PR.** Established two ways for each file: `git diff` against `origin/development` returns empty (this merge does not touch them), and `git cat-file -e origin/main:<file>` fails (they do not exist on `main` at all). They surface here because CodeQL diffs against main's baseline, and this is the first time main has seen development's code. Alert 6 dates to 2026-03-25, which independently confirms it is long-standing.

The only source change in this PR is one test-file mock fix, in a different file from all three alerts.

### Alerts 7 and 8 — `py/incomplete-url-substring-sanitization` (high) — dismissed as *used in tests*

`lambda/research/test/test_research_steps.py:453` and `:535`. The flagged lines are:

```python
assert 'https://t.example' in result['web_context']
assert 'https://t.example' in prompt
```

These assert that a fixture URL propagated into generated output. The rule exists to catch a substring check used as an **allowlist decision** — the classic `if 'example.com' in url` where an attacker satisfies the check by placing the trusted string at an arbitrary position. Here there is no security decision, no untrusted input, and the data flows the opposite direction: the test controls the URL and checks that the code under test carried it through.

The **high** severity reflects the rule's usual context, not this usage.

Worth being explicit about what I did *not* do: rewriting these assertions to satisfy the scanner would pin a disputed shape into the tests, which is how a false positive becomes permanent noise. The reasoning is recorded instead.

### Alert 6 — `js/prototype-pollution-utility` (medium) — dismissed as *false positive*

`frontend/scripts/fix-i18n.mjs:179`. The guard already exists, in a shape the rule does not recognise as a sanitizer:

```js
const UNSAFE_KEYS = new Set(['__proto__', 'constructor', 'prototype'])
if (parts.some(p => UNSAFE_KEYS.has(p))) return
```

That early return runs before any recursive assignment and covers **every** path segment, terminal and intermediate. CodeQL does not treat a `Set`-based `some()` as a sanitizer, so it reports regardless. Impact is further bounded: this is a local build script that never ships to a browser, and it processes locale JSON already committed to this repository, so reaching it requires a malicious key in a tracked file.

A denylist is still a maintenance liability, so the structural hardening — `Object.create(null)` for intermediate containers, keeping the existing check as belt and braces — is tracked separately in #347 rather than widened into this dependency merge.

### Effect on the merge gate

The `CodeQL` check was failing on these three alerts; the two `Analyze` jobs and the event-logging guard were already green. With the alerts dismissed, the remaining gate is human approval.

PR #349, opened 2026-08-18
docs: add CHANGELOG and version the platform 0.2.0

perrozzi:
Establishes versioning for a repo that had none, and records what shipped.

Before this: no changelog, no version tags (the only tag is `pr-108-screenshots`), and all three `package.json` files sitting on `1.0.0` — the value `npm init` writes, never a deliberate release.

## What's here

- **`CHANGELOG.md`** following [Keep a Changelog](https://keepachangelog.com/en/1.1.0/), with a retroactive **0.1.0** entry describing the platform as `main` actually carried it, and **0.2.0** covering the ~120 changes promoted by #346, grouped by area rather than listed as commits.
- **All three manifests set to `0.2.0`** — repo root, `voc-datalake`, `voc-datalake/frontend`.

## Two things worth knowing before approving

**This is nominally a downgrade, from `1.0.0` to `0.2.0`.** That `1.0.0` was the npm scaffold default, not a release. Nothing in the repo reads it (the only `version` read is a plugin manifest field, unrelated) and nothing is published from it, so no consumer is affected. The `0.x` range is deliberate: interfaces on this sample still move between releases, so a minor bump can carry changes that would be breaking in a `1.x` project.

**The version here is not what the dashboard shows.** The UI's build identifier comes from `buildId()` in `vite.config.ts`, which resolves to the short git commit SHA (or a CI-provided SHA). Bumping `package.json` does not change what a user sees in the app. If the displayed version should read `0.2.0`, that's a separate change and worth deciding on its own.

## Verification

The lock files were updated through `npm version` rather than by hand, so `packages[""]` matches `package.json` — `npm ci` validates that pair and fails on a mismatch. Each lock changed by exactly 4 lines, with **no dependency changes**.

| Gate | Result |
|---|---|
| `npm ci`, root + frontend | pass (locks and manifests agree) |
| frontend build | pass (4.07s) |
| frontend tests | **2949 passed** (181 files) |
| CDK tests | **256 passed** (17 files) |

## Accuracy notes

The 0.1.0 entry is written from that release's own README and commit history, with one deliberate omission: the old README listed **WAF protection** as a feature, and there is no WAF in this project — a repo-wide search for `wafv2`/`WebACL` is empty, and two cdk-nag suppressions record it as intentionally excluded to keep sample costs down. That claim is not carried into the changelog.

The 0.2.0 entry records the chat-history repartitioning as an explicit **Upgrade note** rather than leaving it implicit: conversations are now keyed by authenticated user subject, so history written before it becomes unreachable, deliberately and with no migration.

## Follow-up: tagging

Not included here, since tags should point at the merge commit this PR produces rather than at a branch. Suggested once this lands:

```bash
git tag -a v0.2.0 <merge commit> -m "v0.2.0"
git tag -a v0.1.0 0b785a87 -m "v0.1.0 (retroactive)"
git push origin v0.1.0 v0.2.0
```

The `0b785a87` anchor for 0.1.0 is the commit `main` sat at before the promotion, which is also the merge base #346 resolved against.

There is an active enterprise tag ruleset named `block-untagged`, but it is **not a constraint here**: its condition matches only `refs/tags/untagged-*`, the auto-generated names GitHub uses for releases created without a tag. Tags named `v0.1.0` and `v0.2.0` fall outside that pattern.

Happy to create the tags and matching GitHub releases as a follow-up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -144,5 +144,8 @@ customer feedback on AWS.
 - Encryption at rest with a customer-managed KMS key, and cdk-nag suppressions recorded where AWS
   services do not support resource-level permissions.
 
-[0.2.0]: https://github.com/aws-samples/sample-voice-of-customer-datalake/releases/tag/v0.2.0
-[0.1.0]: https://github.com/aws-samples/sample-voice-of-customer-datalake/releases/tag/v0.1.0
+<!-- These point at commit ranges rather than release tags, because no version tags exist yet.
+     Re-point them at /releases/tag/vX.Y.Z once releases are cut. -->
+
+[0.2.0]: https://github.com/aws-samples/sample-voice-of-customer-datalake/compare/0b785a87...main
+[0.1.0]: https://github.com/aws-samples/sample-voice-of-customer-datalake/commit/0b785a87
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,148 @@
+# Changelog
+
+All notable changes to the Voice of Customer (VoC) Data Lake are recorded here. The format follows
+[Keep a Changelog](https://keepachangelog.com/en/1.1.0/).
+
+Versions are `0.x` on purpose: this is a sample platform whose interfaces still move between
+releases, so a minor bump may carry changes that would be breaking in a `1.x` project. Read the
+**Upgrade notes** of each release before deploying over an existing stack.
+
+The version recorded here is the one in the `package.json` files. It is **not** what the dashboard
+displays: the UI's build identifier is the short git commit SHA, injected at build time.
+
+## [0.2.0] - 2026-08-19
+
+The first release since the platform moved from a single-stack sample to a workspace covering
+research, prototyping and team prioritization. Roughly 120 changes, developed between 2026-06 and
+2026-08.
+
+### Added
+
+**Project research workspace**
+- Research projects that generate personas, PRDs and PR/FAQs from the feedback corpus, run as
+  asynchronous jobs with a Background Jobs panel for long-running progress.
+- Document provenance: each generated document records and shows how it was built, and which
+  specification a prototype was built from.
+- Prototype builds that read product context and research, and can be grounded in the mockups they
+  were aimed at.
+- An MCP endpoint so external agents can read project data, with a two-card Export / MCP layout.
+
+**Team prioritization**
+- A prioritization row now represents a project rather than a single document, with ballots keyed to
+  the row and PRDs scorable alongside PR/FAQs.
+- One ballot per reviewer, replacing a single shared score map, with rows leading on the team score.
+- Room voting: a meeting scores a proposal from their phones through a session QR code.
+- Linked feedback forms surface their collected ratings on the matching prioritization row.
+
+**Conversational and AI surfaces**
+- Streaming chat over Server-Sent Events, served by a TypeScript Lambda through API Gateway.
+- A per-surface AI model picker over a curated Claude allowlist, so chat, documents, prototypes,
+  enrichment and utilities can each use a different model.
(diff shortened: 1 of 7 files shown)


perrozzi:

This PR is documentation-and-metadata only: it adds a `CHANGELOG.md` with retroactive 0.1.0 and 0.2.0 entries and moves all three `package.json` files (plus their lock files' `packages[""]`) from the `npm init` default `1.0.0` to `0.2.0`. No source, IAM, Lambda, or frontend code is touched, so none of the repo's structural rules (ingestor prefix registration, Bedrock id/ARN lockstep, i18n key parity, Powertools decorators) are in play. The lock-file edits are exactly the four `version` lines with no dependency changes, which is consistent with `npm version` rather than hand editing. Overall this looks clean; the remaining points are accuracy and follow-through concerns in the changelog text.

### Should fix

- `CHANGELOG.md` release dates are `2026-08-19` (0.2.0) and `2026-05-25` (0.1.0). Please confirm these are real, not future-dated — a changelog dated ahead of the merge is the kind of thing that quietly stays wrong. If 0.2.0 is being cut now, the date should be the merge date.
- `CHANGELOG.md` under **Changed** says "Claude Opus 4.8 upgraded to Opus 5". Those don't correspond to any real Bedrock model family naming. Since this repo requires model id strings and their IAM resource ARNs to move together, the changelog is the one place a reader will look to confirm which model ids changed — please state the actual Bedrock model ids (or the family names as they appear in the allowlist) rather than invented version numbers.
- The two link definitions at the bottom point at `releases/tag/v0.2.0` and `releases/tag/v0.1.0`, which do not exist yet — the PR description defers tagging to a follow-up. Merging leaves two dead links in the repo's most-linked doc. Either create the tags/releases as part of landing this, or drop the link definitions and use plain `## 0.2.0` headings until the tags exist.

### Nits

- Keep a Changelog expects an `## [Unreleased]` section at the top; adding an empty one now makes it obvious where the next entry goes and avoids the next contributor re-deriving the format.
- Nothing enforces that the three manifests stay in sync. A one-line check (or a `version` assertion in the existing CDK/vitest suites) would prevent the drift this PR is correcting from recurring.
- `voc-datalake/package.json` is the one manifest without `"private": true` (root and frontend both have it). Combined with a nominal downgrade, that's the only file where a stray `npm publish` would behave oddly. Worth adding `"private": true` while you're in there.
- The note that the dashboard shows `buildId()` from `vite.config.ts` and not `package.json` is genuinely useful and correctly placed in the changelog preamble — consider mirroring one sentence of it in `docs/deployment.md` so operators debugging "which version is deployed" find it there too.
- The deliberate omission of the WAF claim from the 0.1.0 entry is the right call and is well documented in the PR description; a one-line pointer to the two cdk-nag suppressions in the changelog would preserve that reasoning in-repo rather than only in the PR thread.

---

perrozzi:

Only the two link definitions changed since my last look: they now point at a commit range / commit instead of the not-yet-existing release tags, with an inline note to re-point them once tags are cut. The date and model-naming findings are untouched.

### Resolved
- Dead `releases/tag/v0.2.0` / `v0.1.0` links replaced with resolvable commit links plus a follow-up comment.

### Still open
- `CHANGELOG.md` release dates `2026-08-19` / `2026-05-25` appear future-dated.
- `CHANGELOG.md` **Changed**: "Claude Opus 4.8 upgraded to Opus 5" does not match any real Bedrock model id/family.

### New findings
- `CHANGELOG.md`: the `[0.2.0]` link compares `0b785a87...main` — `main` is a moving ref, so the link stops describing the 0.2.0 range as soon as the next PR lands. Pin it to this PR's merge commit (or leave it as the tag URL and cut the tag when landing) rather than `main`.

Otherwise the delta is clean.

---

PR #354, opened 2026-08-19
feat(mcp): Phase 1 credential model — self-describing token, two reach axes

perrozzi:
Phase 1 of `analysis/PLAN-mcp-platform-server.md`, following Phase 0 (#344). This is the credential model — the blocker under every later phase.

## The problem it removes

The shipped MCP credential is `voc_` + 64 hex, found by **querying every token row in a project and hashing each one** until something matches. That loop is why the credential requires an `X-Project-Id` header: without a project there is no partition to scan.

A header cannot express *"no particular project"*, so `list_projects` — and any other workspace-wide tool — was **unimplementable by construction**, not merely unwritten.

## The credential

`voc_{token_id}_{secret}` (64-bit id, 256-bit secret)

| | Before | After |
|---|---|---|
| Lookup | Query a project partition + hash loop | One keyed Query on `pk=MCPTOKEN, sk=TOKEN#{id}` |
| `X-Project-Id` | **Required** for auth, not just routing | Gone |
| Hashed | The whole credential | The **secret half only**, so `token_id` is safe to log and display |
| Row lives in | The minting project's partition | Its own partition — a credential is workspace-level |
| Format defined in | 3 places (both handlers + an inline Node authorizer in `api-stack.ts`) | 1 module, `shared/mcp_tokens.py` |

**Legacy tokens are not accepted.** Owner confirmed none are in production use, and a stray one fails closed with a 401 that re-minting fixes.

> 🪤 There is deliberately **no dual-read branch**, and the reason is worth recording: `read_reach: 'project-set'` on a read-only token (`projects: []`) resolves to reading **nothing**. Remapping old tokens onto the new axes would have silently broken every one of them rather than preserving them.

## Reach is two axes, not one

One project-scope field cannot say *"write here, look around everywhere"*, which is the shape people actually want.

- **`projects`** — the projects this credential is about. Bounds reads under `project-set`; will bound writes when write tools land, which is why it is not called `write_projects` yet.
- **`read_reach`** — `workspace` (default) | `project-set` | `none`.

### Why `workspace` is the default

Structural, not a preference: `voc-feedback` is keyed `SOURCE#{platform}` with **no `project_id` at all**, and the project-scoped reads select by filters. A narrower default would have to be *invented* for the corpus rather than enforced — and inventing it breaks the cross-project analysis that is the reason to expose an MCP server at all.

⚠️ **It is not the harmless option.** Workspace read reaches every other project's unreleased PRDs, PR-FAQs and prototypes plus every raw verbatim. So it is explicit at mint time, warned about in the form, and badged amber on every row — the list is where someone notices a credential reaches further than they assumed.

### The load-bearing refusal

`project-set` **cannot call a workspace-shaped tool at all**. Feedback and metrics have no project dimension to narrow, so allowing them *"because the token holds `feedback:read`"* would hand a supposedly sealed credential the entire verbatim history. Each tool declares its data shape (`TOOL_REACH_KINDS`) and dispatch is fail-closed on a missing declaration.

## Per-domain scopes

`feedback:read` / `metrics:read` / `projects:read` replace `read` / `read-write`.

The old `read-write` was a **phantom permission**: mintable, stored, and badged in this UI with its own amber pill while *no tool ever required it*. Only scopes that grant something exist now; a write scope arrives in the same change as the first write tool.

A row with no usable `scopes` grants **nothing**, reversing the old default of `read`. That default existed because deployed rows predated the field; every row now carries an explicit set, so a row without one is data damage and must not be guessed at in the caller's favour.

## Also in here

- **`/mcp/autoseed/{project_id}` goes through the same gates as the tools.** It previously compared the path project to the token's single project and **checked no scope at all**.
- **The token boundary gains the Zod normalizer** every other frontend list boundary already has. It matters more here: a token row states what a credential may *do*, so a drifted field makes the UI describe a credential's reach differently from how it is enforced.
- **`shared/tokens.py` deleted** — it held only `hash_token`, whose two callers both moved to `mcp_tokens.hash_secret`.
- **No IAM change.** The keyed lookup is a Query with an exact sort key, so `mcpRole` stays at exactly `Query` + `UpdateItem` — the narrow grant from #344 is untouched.

## Deliberate deferral

**HMAC with a Secrets Manager key** (plan §4.2) is not done, and is documented at `hash_secret` rather than left implicit. Its benefit — a table read alone yields nothing verifiable — matters for *low-entropy* secrets. Buying it here would put a Secrets Manager fetch on the authentication hot path, where a failure has to be told apart from a bad credential, for a marginal gain against 256 bits of `secrets.token_hex`. Worth revisiting if the secret ever becomes user-chosen.

## Verification

| Gate | Result |
|---|---|
| pytest | **2706** passed |
| CDK vitest | **256** passed |
| stream vitest | **339** passed |
| frontend vitest | **2968** passed |
| ruff | delta **−1** (the test rewrite fixed a pre-existing `I001`) |
| eslint | clean (`--max-warnings 0`) |
| typecheck | clean — frontend, CDK, stream |

**Mutation-proved 6/6.** Each new invariant was removed and the guarding tests confirmed to fail: the reach gate, the scope check, strict credential parsing, the autoseed scope gate, the mint project set, and the sealed-token corpus refusal.

## Deployed and live-verified

`VocApiStack` deployed to us-east-1 (clean synth, **zero warnings**).

- `scripts/test-api.sh` — **34/34**
- Live MCP battery — **37/37**, driven through API Gateway with credentials minted by the real mint route

The live battery asserts what unit tests structurally cannot: that the format the mint route *produces* is the format the deployed authorizer and handler *accept*, end to end. Highlights:

- `tools/list` authenticates with **no `X-Project-Id` header** — the header requirement is gone in production, which is the B1 unblock.
- A **workspace** credential really does read another project (the warned-about default, confirmed rather than assumed).
- A **`project-set`** credential is refused the feedback corpus **and** another project, both `-32003`.
- A **`feedback:read`-only** credential reads feedback and is refused projects.
- A **legacy `voc_<64 hex>`** credential is refused 401 — the dropped compat path fails closed as intended.
- Mint-time validation holds against the deployed API: unknown scope, empty scope set, and the retired `write-set` reach name are each 400.
- Autoseed is refused without `projects:read`, and refused outside a sealed credential's reach.
- **Revocation takes effect**: the credential stops authenticating immediately after `DELETE`.
- Phase 0 hardening intact: foreign `Origin` → 403, and the RFC 6750 challenge is delivered (under API Gateway's remapped header name, as #344 documented).

All minted test credentials were revoked at the end of the run.

---

## ✅ Orphaned rows cleaned up

The 2 pre-Phase-1 rows at `PROJECT#{id} / TOKEN#{id}` have been **deleted** (owner-authorised). Each
delete carried a `ConditionExpression` asserting the row still looked legacy
(`attribute_exists(token_hash) AND attribute_not_exists(secret_hash)`), so it was
impossible to remove a new-format credential even with a stale key list. Full items
were dumped first.

Verified after: table went 40 → 38 rows, **0** `TOKEN#` rows remain, and all 3 project
`META` rows plus every artifact kind (PERSONA, PRD, PRFAQ, RESEARCH, DOC, PROTOTYPE,
PRODUCT_CONTEXT, PRODUCT_DOC) are intact.

## Files this PR does NOT touch, and why that matters for review

- **`lib/stacks/api-stack.ts`** — zero changes, which is the claim rather than an
  omission. The inline authorizer validates `startsWith('Bearer voc_')` and
  `length >= 20`, so the new 89-character credential passes unchanged, and
  `projectsTable.grant(mcpRole, 'dynamodb:Query', 'dynamodb:UpdateItem')` is
  resource-level with no `LeadingKeys` condition. Both are exercised on every
  authenticated call in the live battery.
- **The `shared.api` import block in `projects_handler.py`** — `get_caller_subject`
  was already imported, so there is no hunk to show.

---

## Browser verification — and the bug it caught

The API battery cannot see render-time defects, so the mint flow was also driven through the deployed
CloudFront site.

**It found a real one.** Scope names contain a colon, which is i18next's default `nsSeparator`, so
`t('mcp.scopeDesc.feedback:read')` resolved as namespace `mcp.scopeDesc.feedback` + key `read` — and every
checkbox rendered the literal word **"read"** instead of its description. Fixed with
`nsSeparator: false` (`38cf65e7`).

Worth noting why the suite missed it: it asserted the checkbox *names* (which contain the scope, and were
correct), nothing asserted the description text, and the broken lookup produced a plausible word rather
than a visible key path. The tests *could* have caught it — `src/test/setup.ts` loads the real
`en` locale — so the gap was a missing assertion, now added and mutation-proved. Those three are the only
colon-bearing keys across all 8 locale files.

Everything else in the flow verified against the deployed site:

- nothing pre-checked; **Generate disabled** until a scope is ticked
- reach defaults to workspace **with its warning text**, and the hint switches to the sealed wording on `project-set`
- revealed credential is **89 chars, `voc_tok_<16hex>_<64hex>`**
- banner shows the correct 30-day deadline
- the row badges **"This project" in blue** (not amber, since it is sealed) with only the two ticked scopes
- the `mcp.json` snippet contains **no `X-Project-Id`**
- revoke empties the list to `Active Tokens (0)`
- **zero console errors or warnings**

### DIFF
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -253,15 +253,33 @@
     "noTokensYet": "Noch keine API-Tokens",
     "revealToken": "Token anzeigen",
     "revokeToken": "Token widerrufen",
-    "scope": "Bereich",
-    "scopeRead": "Nur Lesen — Feedback suchen, Metriken anzeigen",
-    "scopeReadWrite": "Lesen & Schreiben — inkl. Chat und Dokumentgenerierung",
     "title": "MCP-Zugang",
     "tokenCreated": "Token erfolgreich erstellt",
     "tokenExpiresOn": "Dieses Token läuft am {{date}} ab.",
     "tokenName": "Token-Name",
     "tokenNamePlaceholder": "z.B. Mein Kiro-Token",
-    "tokenPasteHint": "Fügen Sie dieses Token in den <code>Authorization</code>-Header Ihrer mcp.json oder des Autoseed-Curl-Befehls ein."
+    "tokenPasteHint": "Fügen Sie dieses Token in den <code>Authorization</code>-Header Ihrer mcp.json oder des Autoseed-Curl-Befehls ein.",
+    "scopes": "Berechtigungen",
+    "scopeDesc": {
+      "feedback:read": "Kundenfeedback durchsuchen und lesen",
+      "metrics:read": "Dashboards und Metrik-Aufschlüsselungen lesen",
+      "projects:read": "Projekte, Personas und Dokumente lesen"
+    },
+    "noScopes": "keine Berechtigungen",
+    "readReach": "Lesezugriff",
+    "reachOption": {
+      "workspace": "Gesamter Workspace (Standard)",
+      "project-set": "Nur dieses Projekt"
+    },
+    "reachHint": {
+      "workspace": "Erreicht jedes Projekt im Workspace – einschließlich unveröffentlichter Dokumente anderer Projekte – und das gesamte Kundenfeedback.",
+      "project-set": "Auf dieses Projekt beschränkt. Kann den Feedback-Bestand und Workspace-Metriken nicht lesen, da diese nicht projektbezogen sind."
+    },
+    "reachBadge": {
+      "workspace": "Gesamter Workspace",
+      "project-set": "Dieses Projekt",
+      "none": "Kein Lesezugriff"
+    }
   },
   "notFound": {
     "backToProjects": "Zurück zu Projekten",
--- a/voc-datalake/frontend/public/locales/de/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/de/projectDetail.json
@@ -261,9 +261,9 @@
     "tokenPasteHint": "Fügen Sie dieses Token in den <code>Authorization</code>-Header Ihrer mcp.json oder des Autoseed-Curl-Befehls ein.",
     "scopes": "Berechtigungen",
     "scopeDesc": {
-      "feedback:read": "Kundenfeedback durchsuchen und lesen",
-      "metrics:read": "Dashboards und Metrik-Aufschlüsselungen lesen",
-      "projects:read": "Projekte, Personas und Dokumente lesen"
+      "feedback_read": "Kundenfeedback durchsuchen und lesen",
+      "metrics_read": "Dashboards und Metrik-Aufschlüsselungen lesen",
+      "projects_read": "Projekte, Personas und Dokumente lesen"
     },
     "noScopes": "keine Berechtigungen",
     "readReach": "Lesezugriff",
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -261,9 +261,9 @@
     "tokenPasteHint": "Paste this token into the <code>Authorization</code> header of your mcp.json or autoseed curl command.",
     "scopes": "Scopes",
     "scopeDesc": {
-      "feedback:read": "Search and read customer feedback",
-      "metrics:read": "Read dashboards and metric breakdowns",
-      "projects:read": "Read projects, personas and documents"
+      "feedback_read": "Search and read customer feedback",
+      "metrics_read": "Read dashboards and metric breakdowns",
+      "projects_read": "Read projects, personas and documents"
     },
     "noScopes": "no scopes",
     "readReach": "Read reach",
--- a/voc-datalake/frontend/public/locales/en/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/en/projectDetail.json
@@ -253,15 +253,33 @@
     "noTokensYet": "No API tokens yet",
     "revealToken": "Reveal token",
     "revokeToken": "Revoke token",
-    "scope": "Scope",
-    "scopeRead": "Read only — search feedback, view metrics",
-    "scopeReadWrite": "Read & write — includes chat and document generation",
     "title": "MCP Access",
     "tokenCreated": "Token created successfully",
     "tokenExpiresOn": "This token expires on {{date}}.",
     "tokenName": "Token name",
     "tokenNamePlaceholder": "e.g. My Kiro token",
-    "tokenPasteHint": "Paste this token into the <code>Authorization</code> header of your mcp.json or autoseed curl command."
+    "tokenPasteHint": "Paste this token into the <code>Authorization</code> header of your mcp.json or autoseed curl command.",
+    "scopes": "Scopes",
+    "scopeDesc": {
+      "feedback:read": "Search and read customer feedback",
+      "metrics:read": "Read dashboards and metric breakdowns",
+      "projects:read": "Read projects, personas and documents"
+    },
+    "noScopes": "no scopes",
+    "readReach": "Read reach",
+    "reachOption": {
+      "workspace": "Whole workspace (default)",
+      "project-set": "This project only"
+    },
+    "reachHint": {
+      "workspace": "Reaches every project in the workspace — including other projects’ unreleased documents — and all customer feedback.",
+      "project-set": "Limited to this project. Cannot read the feedback corpus or workspace metrics, which are not project-scoped."
+    },
+    "reachBadge": {
+      "workspace": "Whole workspace",
+      "project-set": "This project",
+      "none": "No read access"
+    }
   },
   "notFound": {
     "backToProjects": "Back to Projects",
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -264,9 +264,9 @@
     "tokenPasteHint": "Pega este token en el header <code>Authorization</code> de tu mcp.json o comando curl autoseed.",
     "scopes": "Permisos",
     "scopeDesc": {
-      "feedback:read": "Buscar y leer comentarios de clientes",
-      "metrics:read": "Leer paneles y desglose de métricas",
-      "projects:read": "Leer proyectos, personas y documentos"
+      "feedback_read": "Buscar y leer comentarios de clientes",
+      "metrics_read": "Leer paneles y desglose de métricas",
+      "projects_read": "Leer proyectos, personas y documentos"
     },
     "noScopes": "sin permisos",
     "readReach": "Alcance de lectura",
--- a/voc-datalake/frontend/public/locales/es/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/es/projectDetail.json
@@ -256,15 +256,33 @@
     "noTokensYet": "Sin tokens API aún",
     "revealToken": "Mostrar token",
     "revokeToken": "Revocar token",
-    "scope": "Alcance",
-    "scopeRead": "Solo lectura — buscar feedback, ver métricas",
-    "scopeReadWrite": "Lectura y escritura — incluye chat y generación de documentos",
     "title": "Acceso MCP",
     "tokenCreated": "Token creado exitosamente",
     "tokenExpiresOn": "Este token caduca el {{date}}.",
     "tokenName": "Nombre del token",
     "tokenNamePlaceholder": "p. ej. Mi token Kiro",
-    "tokenPasteHint": "Pega este token en el header <code>Authorization</code> de tu mcp.json o comando curl autoseed."
+    "tokenPasteHint": "Pega este token en el header <code>Authorization</code> de tu mcp.json o comando curl autoseed.",
+    "scopes": "Permisos",
+    "scopeDesc": {
+      "feedback:read": "Buscar y leer comentarios de clientes",
+      "metrics:read": "Leer paneles y desglose de métricas",
+      "projects:read": "Leer proyectos, personas y documentos"
+    },
+    "noScopes": "sin permisos",
+    "readReach": "Alcance de lectura",
+    "reachOption": {
+      "workspace": "Todo el espacio de trabajo (predeterminado)",
+      "project-set": "Solo este proyecto"
+    },
+    "reachHint": {
+      "workspace": "Alcanza todos los proyectos del espacio de trabajo, incluidos los documentos no publicados de otros proyectos, y todos los comentarios.",
+      "project-set": "Limitado a este proyecto. No puede leer el corpus de comentarios ni las métricas del espacio de trabajo, que no son por proyecto."
+    },
+    "reachBadge": {
+      "workspace": "Todo el espacio",
+      "project-set": "Este proyecto",
+      "none": "Sin lectura"
+    }
   },
   "notFound": {
     "backToProjects": "Volver a Proyectos",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -256,15 +256,33 @@
     "noTokensYet": "Aucun token API pour le moment",
     "revealToken": "Afficher le token",
     "revokeToken": "Révoquer le token",
-    "scope": "Portée",
-    "scopeRead": "Lecture seule — rechercher des retours, consulter les métriques",
-    "scopeReadWrite": "Lecture & écriture — inclut le chat et la génération de documents",
     "title": "Accès MCP",
     "tokenCreated": "Token créé avec succès",
     "tokenExpiresOn": "Ce jeton expire le {{date}}.",
     "tokenName": "Nom du token",
     "tokenNamePlaceholder": "ex. Mon token Kiro",
-    "tokenPasteHint": "Collez ce token dans l'en-tête <code>Authorization</code> de votre mcp.json ou commande curl autoseed."
+    "tokenPasteHint": "Collez ce token dans l'en-tête <code>Authorization</code> de votre mcp.json ou commande curl autoseed.",
+    "scopes": "Portées",
+    "scopeDesc": {
+      "feedback:read": "Rechercher et lire les retours clients",
+      "metrics:read": "Lire les tableaux de bord et les métriques",
+      "projects:read": "Lire les projets, personas et documents"
+    },
+    "noScopes": "aucune portée",
+    "readReach": "Portée de lecture",
+    "reachOption": {
+      "workspace": "Tout l’espace de travail (par défaut)",
+      "project-set": "Ce projet uniquement"
+    },
+    "reachHint": {
+      "workspace": "Atteint tous les projets de l’espace de travail — y compris les documents non publiés des autres projets — et tous les retours clients.",
+      "project-set": "Limité à ce projet. Ne peut pas lire le corpus de retours ni les métriques de l’espace, qui ne sont pas liés à un projet."
+    },
+    "reachBadge": {
+      "workspace": "Tout l’espace",
+      "project-set": "Ce projet",
+      "none": "Aucune lecture"
+    }
   },
   "notFound": {
     "backToProjects": "Retour aux projets",
--- a/voc-datalake/frontend/public/locales/fr/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/fr/projectDetail.json
@@ -264,9 +264,9 @@
     "tokenPasteHint": "Collez ce token dans l'en-tête <code>Authorization</code> de votre mcp.json ou commande curl autoseed.",
     "scopes": "Portées",
     "scopeDesc": {
-      "feedback:read": "Rechercher et lire les retours clients",
-      "metrics:read": "Lire les tableaux de bord et les métriques",
-      "projects:read": "Lire les projets, personas et documents"
+      "feedback_read": "Rechercher et lire les retours clients",
+      "metrics_read": "Lire les tableaux de bord et les métriques",
+      "projects_read": "Lire les projets, personas et documents"
     },
     "noScopes": "aucune portée",
     "readReach": "Portée de lecture",
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -261,9 +261,9 @@
     "tokenPasteHint": "このトークンを mcp.json または autoseed curl コマンドの <code>Authorization</code> ヘッダーに貼り付けてください。",
     "scopes": "スコープ",
     "scopeDesc": {
-      "feedback:read": "顧客フィードバックの検索と閲覧",
-      "metrics:read": "ダッシュボードと指標の内訳の閲覧",
-      "projects:read": "プロジェクト、ペルソナ、ドキュメントの閲覧"
+      "feedback_read": "顧客フィードバックの検索と閲覧",
+      "metrics_read": "ダッシュボードと指標の内訳の閲覧",
+      "projects_read": "プロジェクト、ペルソナ、ドキュメントの閲覧"
     },
     "noScopes": "スコープなし",
     "readReach": "読み取り範囲",
--- a/voc-datalake/frontend/public/locales/ja/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ja/projectDetail.json
@@ -253,15 +253,33 @@
     "noTokensYet": "API トークンがまだありません",
     "revealToken": "トークンを表示",
     "revokeToken": "トークンを無効化",
-    "scope": "スコープ",
-    "scopeRead": "読み取り専用 — フィードバック検索、メトリクス表示",
-    "scopeReadWrite": "読み取り・書き込み — チャットとドキュメント生成を含む",
     "title": "MCP アクセス",
     "tokenCreated": "トークンが正常に作成されました",
     "tokenExpiresOn": "このトークンの有効期限は{{date}}です。",
     "tokenName": "トークン名",
     "tokenNamePlaceholder": "例：My Kiro token",
-    "tokenPasteHint": "このトークンを mcp.json または autoseed curl コマンドの <code>Authorization</code> ヘッダーに貼り付けてください。"
+    "tokenPasteHint": "このトークンを mcp.json または autoseed curl コマンドの <code>Authorization</code> ヘッダーに貼り付けてください。",
+    "scopes": "スコープ",
+    "scopeDesc": {
+      "feedback:read": "顧客フィードバックの検索と閲覧",
+      "metrics:read": "ダッシュボードと指標の内訳の閲覧",
+      "projects:read": "プロジェクト、ペルソナ、ドキュメントの閲覧"
+    },
+    "noScopes": "スコープなし",
+    "readReach": "読み取り範囲",
+    "reachOption": {
+      "workspace": "ワークスペース全体（既定）",
+      "project-set": "このプロジェクトのみ"
+    },
+    "reachHint": {
+      "workspace": "ワークスペース内のすべてのプロジェクト（他プロジェクトの未公開ドキュメントを含む）とすべての顧客フィードバックに到達します。",
+      "project-set": "このプロジェクトに限定されます。プロジェクト単位ではないフィードバック全体とワークスペース指標は読み取れません。"
+    },
+    "reachBadge": {
+      "workspace": "ワークスペース全体",
+      "project-set": "このプロジェクト",
+      "none": "読み取り不可"
+    }
   },
   "notFound": {
     "backToProjects": "プロジェクトに戻る",
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -261,9 +261,9 @@
     "tokenPasteHint": "이 토큰을 mcp.json 또는 autoseed curl 명령의 <code>Authorization</code> 헤더에 붙여넣으세요.",
     "scopes": "범위",
     "scopeDesc": {
-      "feedback:read": "고객 피드백 검색 및 읽기",
-      "metrics:read": "대시보드 및 지표 분석 읽기",
-      "projects:read": "프로젝트, 페르소나, 문서 읽기"
+      "feedback_read": "고객 피드백 검색 및 읽기",
+      "metrics_read": "대시보드 및 지표 분석 읽기",
+      "projects_read": "프로젝트, 페르소나, 문서 읽기"
     },
     "noScopes": "범위 없음",
     "readReach": "읽기 범위",
--- a/voc-datalake/frontend/public/locales/ko/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/ko/projectDetail.json
@@ -253,15 +253,33 @@
     "noTokensYet": "API 토큰이 없습니다",
     "revealToken": "토큰 표시",
     "revokeToken": "토큰 취소",
-    "scope": "범위",
-    "scopeRead": "읽기 전용 — 피드백 검색, 지표 보기",
-    "scopeReadWrite": "읽기 및 쓰기 — 채팅 및 문서 생성 포함",
     "title": "MCP 접근",
     "tokenCreated": "토큰이 성공적으로 생성되었습니다",
     "tokenExpiresOn": "이 토큰은 {{date}}에 만료됩니다.",
     "tokenName": "토큰 이름",
     "tokenNamePlaceholder": "예: My Kiro token",
-    "tokenPasteHint": "이 토큰을 mcp.json 또는 autoseed curl 명령의 <code>Authorization</code> 헤더에 붙여넣으세요."
+    "tokenPasteHint": "이 토큰을 mcp.json 또는 autoseed curl 명령의 <code>Authorization</code> 헤더에 붙여넣으세요.",
+    "scopes": "범위",
+    "scopeDesc": {
+      "feedback:read": "고객 피드백 검색 및 읽기",
+      "metrics:read": "대시보드 및 지표 분석 읽기",
+      "projects:read": "프로젝트, 페르소나, 문서 읽기"
+    },
+    "noScopes": "범위 없음",
+    "readReach": "읽기 범위",
+    "reachOption": {
+      "workspace": "전체 워크스페이스(기본값)",
+      "project-set": "이 프로젝트만"
+    },
+    "reachHint": {
+      "workspace": "다른 프로젝트의 미공개 문서를 포함한 워크스페이스의 모든 프로젝트와 모든 고객 피드백에 접근합니다.",
+      "project-set": "이 프로젝트로 제한됩니다. 프로젝트 단위가 아닌 피드백 전체와 워크스페이스 지표는 읽을 수 없습니다."
+    },
+    "reachBadge": {
+      "workspace": "전체 워크스페이스",
+      "project-set": "이 프로젝트",
+      "none": "읽기 불가"
+    }
   },
   "notFound": {
     "backToProjects": "프로젝트로 돌아가기",
--- a/voc-datalake/frontend/public/locales/pt/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/pt/projectDetail.json
@@ -256,15 +256,33 @@
     "noTokensYet": "Nenhum token API ainda",
     "revealToken": "Revelar token",
     "revokeToken": "Revogar token",
-    "scope": "Escopo",
-    "scopeRead": "Somente leitura — pesquisar feedback, visualizar métricas",
-    "scopeReadWrite": "Leitura e escrita — inclui chat e geração de documentos",
     "title": "Acesso MCP",
     "tokenCreated": "Token criado com sucesso",
     "tokenExpiresOn": "Este token expira em {{date}}.",
     "tokenName": "Nome do token",
     "tokenNamePlaceholder": "ex: Meu token Kiro",
-    "tokenPasteHint": "Cole este token no header <code>Authorization</code> do seu mcp.json ou comando curl autoseed."
+    "tokenPasteHint": "Cole este token no header <code>Authorization</code> do seu mcp.json ou comando curl autoseed.",
+    "scopes": "Escopos",
+    "scopeDesc": {
+      "feedback:read": "Pesquisar e ler feedback de clientes",
+      "metrics:read": "Ler painéis e detalhamento de métricas",
+      "projects:read": "Ler projetos, personas e documentos"
+    },
+    "noScopes": "sem escopos",
+    "readReach": "Alcance de leitura",
+    "reachOption": {
+      "workspace": "Todo o workspace (padrão)",
+      "project-set": "Apenas este projeto"
+    },
+    "reachHint": {
+      "workspace": "Alcança todos os projetos do workspace — incluindo documentos não publicados de outros projetos — e todo o feedback de clientes.",
+      "project-set": "Limitado a este projeto. Não pode ler o corpus de feedback nem as métricas do workspace, que não são por projeto."
+    },
+    "reachBadge": {
+      "workspace": "Todo o workspace",
+      "project-set": "Este projeto",
+      "none": "Sem leitura"
+    }
   },
   "notFound": {
     "backToProjects": "Voltar para Projetos",
--- a/voc-datalake/frontend/public/locales/pt/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/pt/projectDetail.json
@@ -264,9 +264,9 @@
     "tokenPasteHint": "Cole este token no header <code>Authorization</code> do seu mcp.json ou comando curl autoseed.",
     "scopes": "Escopos",
     "scopeDesc": {
-      "feedback:read": "Pesquisar e ler feedback de clientes",
-      "metrics:read": "Ler painéis e detalhamento de métricas",
-      "projects:read": "Ler projetos, personas e documentos"
+      "feedback_read": "Pesquisar e ler feedback de clientes",
+      "metrics_read": "Ler painéis e detalhamento de métricas",
+      "projects_read": "Ler projetos, personas e documentos"
     },
     "noScopes": "sem escopos",
     "readReach": "Alcance de leitura",
--- a/voc-datalake/frontend/public/locales/zh/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/zh/projectDetail.json
@@ -253,15 +253,33 @@
     "noTokensYet": "暂无 API 令牌",
     "revealToken": "显示令牌",
     "revokeToken": "撤销令牌",
-    "scope": "权限范围",
-    "scopeRead": "仅读取 — 搜索反馈、查看指标",
-    "scopeReadWrite": "读写 — 包括聊天和文档生成",
     "title": "MCP 访问",
     "tokenCreated": "令牌创建成功",
     "tokenExpiresOn": "此令牌将于 {{date}} 到期。",
     "tokenName": "令牌名称",
     "tokenNamePlaceholder": "例如 My Kiro token",
-    "tokenPasteHint": "将此令牌粘贴到 mcp.json 或 autoseed curl 命令的 <code>Authorization</code> 标头中。"
+    "tokenPasteHint": "将此令牌粘贴到 mcp.json 或 autoseed curl 命令的 <code>Authorization</code> 标头中。",
+    "scopes": "权限范围",
+    "scopeDesc": {
+      "feedback:read": "搜索和查看客户反馈",
+      "metrics:read": "查看仪表板和指标细分",
+      "projects:read": "查看项目、用户画像和文档"
+    },
+    "noScopes": "无权限范围",
+    "readReach": "读取范围",
+    "reachOption": {
+      "workspace": "整个工作区（默认）",
+      "project-set": "仅此项目"
+    },
+    "reachHint": {
+      "workspace": "可访问工作区中的所有项目（包括其他项目未发布的文档）以及全部客户反馈。",
+      "project-set": "仅限此项目。无法读取并非按项目划分的反馈全集和工作区指标。"
+    },
+    "reachBadge": {
+      "workspace": "整个工作区",
+      "project-set": "此项目",
+      "none": "无读取权限"
+    }
   },
   "notFound": {
     "backToProjects": "返回项目",
--- a/voc-datalake/frontend/public/locales/zh/projectDetail.json
+++ b/voc-datalake/frontend/public/locales/zh/projectDetail.json
@@ -261,9 +261,9 @@
     "tokenPasteHint": "将此令牌粘贴到 mcp.json 或 autoseed curl 命令的 <code>Authorization</code> 标头中。",
     "scopes": "权限范围",
     "scopeDesc": {
-      "feedback:read": "搜索和查看客户反馈",
-      "metrics:read": "查看仪表板和指标细分",
-      "projects:read": "查看项目、用户画像和文档"
+      "feedback_read": "搜索和查看客户反馈",
+      "metrics_read": "查看仪表板和指标细分",
+      "projects_read": "查看项目、用户画像和文档"
     },
     "noScopes": "无权限范围",
     "readReach": "读取范围",
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -91,6 +91,10 @@ function buildHeaders(existingHeaders?: HeadersInit): Record<string, string> {
 
 import { z } from 'zod'
 import { normalizeFeedbackItem, normalizeFeedbackItems } from './feedbackSchema'
+import {
+  CreateApiTokenResponseSchema, normalizeApiTokens,
+  type McpScope, type ReadReach,
+} from './mcpTokenSchema'
 
 // API response parser using Zod for runtime validation
 // This satisfies the no-type-assertions rule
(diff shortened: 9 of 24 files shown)


perrozzi:

This PR replaces the MCP `voc_ + 64 hex` credential with a self-describing `voc_{token_id}_{secret}` token resolved by a single keyed Query, moves token rows into a workspace-level `MCPTOKEN` partition, splits authorization into per-domain scopes plus a `read_reach` axis, puts the autoseed route behind the same gates, and adds a Zod normalizer + i18n for the token UI across all 8 locales. The design is coherent, the fail-closed dispatch (missing scope *or* missing reach kind → `-32603`) is the right shape, and the test coverage is unusually thorough. **The diff was truncated, so this review is partial** — notably `voc-datalake/lambda/shared/mcp_tokens.py` is cut off mid-`hash_secret`, and I cannot see `cdk/lib/api-stack.ts`, the shared-module tests, or the docs/plan updates.

### Blockers

- **`api_list_tokens` reads only the first page of the token partition.** `_query_all_tokens` (`voc-datalake/lambda/api/projects_handler.py`) does a single `table.query(KeyConditionExpression=Key('pk').eq(MCP_TOKEN_PK))` and returns `response.get('Items', [])` with no `LastEvaluatedKey` loop. All tokens now share one partition workspace-wide, so once the partition exceeds 1 MB the list silently truncates — and since the UI is the only revoke path, that directly breaks the invariant the PR claims (`"no credential can become unlistable and therefore unrevocable through the UI"`). `mcp_tokens.py` itself anticipates "tokens ever number in the thousands". Add pagination (and a regression test that returns a `LastEvaluatedKey`-bearing first page).
- **Not deployed / live MCP battery and `scripts/test-api.sh` not run**, as the PR states. Per this repo's convention that gates merge, especially for a change that alters the credential format and the DynamoDB key layout for every existing token.
- **Old token rows are left behind with live-looking material.** Rows at `PROJECT#{id} / TOKEN#{id}` still carry `token_hash` and will never authenticate, never be listed, and never be revocable through the new routes. Two consequences to resolve: (a) dead credential digests persist indefinitely with no cleanup path, and (b) any project-partition read that uses `begins_with`/no-`sk` filter will now surface stray `TOKEN#` items in project responses. Please include a delete script or at minimum document the manual cleanup, and confirm no project read enumerates the whole `PROJECT#{id}` partition.

### Should fix

- **Verify the API Gateway side in the truncated part of the diff.** The description says the `voc_` format was also spelled in "an inline Node authorizer in `api-stack.ts`". If that authorizer still validates `voc_` + 64 hex, every new credential is rejected before reaching the Lambda. Likewise confirm the `mcpRole` grant on the projects table has no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*` — `_authenticate` now `Query`s and `UpdateItem`s `pk=MCPTOKEN`, which such a condition would deny at runtime (the claim "no IAM change" only holds if the grant is resource-level, not key-conditioned).
- **Minting defaults to the widest credential.** `_validate_scopes(None)` returns every read scope and `_validate_read_reach(None)` returns `workspace`, so `POST /projects/{id}/api-tokens {"name":"x"}` mints a full-workspace, all-scope credential. That is the exact opposite of the (correct) enforcement-side stance that "a row with no usable `scopes` grants nothing". The UI always sends both fields, so requiring them explicitly at mint costs nothing and removes the fail-open path for direct API callers.
- **`mcpTokenSchema.test.ts` "never surfaces a secret hash as a token field" asserts the wrong thing.** `ApiTokenSchema` is `z.looseObject`, so `secret_hash: 'x'` *does* pass through into `parsed`; the test then asserts `not.toHaveProperty('scope')`, which is unrelated. Either assert `expect(parsed).not.toHaveProperty('secret_hash')` and strip it in the schema, or rename the test — as written it documents a guarantee that does not exist.
- **Confirm `get_caller_subject` is imported in `projects_handler.py`.** The visible import hunk only removes `secrets` and swaps `shared.tokens` for `shared.mcp_tokens`; the new `'created_by': get_caller_subject(...)` in `api_create_token` would be a `NameError` at mint time if it is not already imported in this module.
- **Unit tests for `shared/mcp_tokens.py` itself** (`parse_token` boundary cases, `reach_allows` matrix, `token_sk`) are not visible in the diff. They may be in the truncated portion; if not, the format module deserves direct tests rather than only handler-level coverage.

### Nits

- `_resolve_project_id` silently discards a wrong-typed `project_id` (`123`, `["p"]`) and falls back to the token's project — the test pins this as intended, but the tool then reads a *different* project than the client named. A `-32602` at the boundary would be more in keeping with "validation at system boundaries".
- The `-32602` "needs a project_id argument" branch in `_handle_tools_call` also fires when `read_reach` is `none` and the project is unresolved, telling the caller to add an argument that cannot help.
- `normalizeApiTokens` logs the whole payload (`console.warn(..., { raw })`); token names and project ids are not secrets but the shape of the log is against the no-sensitive-data-in-logs habit — the type/length is enough.
- New i18n keys are appended after `tokenPasteHint` rather than in alphabetical position in all 8 locale files, unlike the surrounding keys.
- `_token_response` returns `last_used_at: None` for never-used tokens; the frontend normalizer coerces it, but returning the key only when present would match the `expires_at` handling described elsewhere in the same module.

---

perrozzi:
Thanks — six accepted, six declined with evidence. Pushed as `1abd1e51`, redeployed, live battery now **42/42**.

## Accepted

**1. `api_list_tokens` read only the first page.** Correct, and it contradicted the design rather than just being untidy: the list is the only revoke path, so a truncated page makes credentials unrevocable — the exact invariant the mint route is written to guarantee. Now follows `LastEvaluatedKey`; the regression test feeds a first page carrying one and asserts both rows return *and* that the second Query resumes from the right key.

**2. Minting defaulted to the widest credential.** Correct, and the framing was the useful part: a fail-open mint boundary under a fail-closed enforcement path. `scopes` is now **required**.

`read_reach` deliberately still defaults, and I want to be explicit about the asymmetry rather than have it read as an oversight: the owner chose a *specific* reach default that the form warns about in amber, whereas there is no least-privilege scope set to fall back to — every candidate default is either everything or useless. `DEFAULT_SCOPES` had become a lie and is renamed `ALL_READ_SCOPES`.

**3. `mcpTokenSchema.test.ts` asserted the wrong thing.** Entirely right, and it was my own test claiming a guarantee that did not exist. The schema now strips `secret_hash` and `created_by`, and the test asserts *that*. Kept as a targeted strip rather than a strict object, since forward compatibility with new backend fields is why it is loose.

**4. `_resolve_project_id` silently discarding a wrong-typed `project_id`.** You were right and my test pinning the old behaviour was wrong. Falling back meant reading a *different* project than the client named and reporting success — worse than an error. Now `-32602`; absence still defaults, because absence and garbage are different intents.

**5. The `-32602` branch firing for a `none`-reach token.** Fixed by checking reach first, so a credential whose reach covers nothing is told about reach.

**6. `normalizeApiTokens` logging the whole payload.** Trimmed to the type.

**On the old rows** (your third blocker): confirmed **2** orphaned rows live at `PROJECT#{id} / TOKEN#{id}`. They can never authenticate, be listed, or be revoked, so they are dead data — agreed. I have **not** deleted them, since that is a destructive change to a live table and belongs to the maintainer's call; happy to add a one-off cleanup script or run it on request. Your sub-point (b) does not apply, though: those rows were *always* in the project partition, and both `get_project` and `projects.get_project` bucket by `sk` prefix (`META` / `PERSONA#` / `PRD#` / …), so `TOKEN#` items are ignored rather than surfaced. That is unchanged by this PR.

## Declined, with evidence

- **Not deployed** — stale. Deployed before your review; the PR body was updated with the evidence. `test-api.sh` 34/34, live battery 42/42.
- **The inline Node authorizer may still expect `voc_` + 64 hex** — fair thing to check, but it validates `startsWith('Bearer voc_')` and `length >= 20`. The new 89-char credential passes, and the live battery proves it end to end *through API Gateway*, which is the only way to be sure.
- **`mcpRole` may carry a `LeadingKeys` condition pinned to `PROJECT#*`** — `projectsTable.grant(mcpRole, 'dynamodb:Query', 'dynamodb:UpdateItem')` is resource-level with no condition, and the live run exercises both a `pk=MCPTOKEN` read and the `last_used_at` write. "No IAM change" holds.
- **Confirm `get_caller_subject` is imported** — it is, pre-existing in the `shared.api` import block.
- **Unit tests for `shared/mcp_tokens.py`** — 45 of them, in `lambda/shared/test/test_mcp_tokens.py`. Your diff was truncated.
- **`_token_response` returning `last_used_at: None`** — consistent with `expires_at` in the same projection. The absent-attribute rule you are recalling applies to the *stored item*, not the response.
- **i18n key ordering** — the surrounding `mcp` block is not alphabetical either (`expiry`, `expiryNever`, `expiresDate` sit after unrelated keys), so appending matches the file rather than diverging from it.

## Verification after the fixes

pytest **2713** · CDK **256** · stream **339** · frontend **2969** · ruff delta **−1** · eslint + typecheck clean.

**Mutation-proved 4/4** on the new invariants — each removed and the guarding test confirmed to fail: the pagination loop, required scopes, junk-`project_id` rejection, and the reach-first ordering. Review-driven tests get the same treatment as feature tests, since a test written to answer a comment is worth nothing if it cannot fail.

Live battery grew 37 → **42**, adding: omitted `scopes` refused 400, junk `project_id` refused `-32602` (three shapes), and that a sealed credential's refusal names *reach* rather than a missing argument.

perrozzi:

Follow-up review — note the full-context diff was **truncated**, so this remains a partial review; I cannot see `cdk/lib/api-stack.ts`, the docs/plan updates, or the tail of `test_projects_handler.py`.

Since the last round the author added `LastEvaluatedKey` pagination to `_query_all_tokens` (with a regression test), made `scopes` a required mint field (renaming `DEFAULT_SCOPES` → `ALL_READ_SCOPES` to remove the temptation of a default), made `ApiTokenSchema` strip `secret_hash`/`created_by`, turned a garbage `project_id` into `-32602` instead of a silent fallback, reordered the reach check ahead of the missing-argument branch, and narrowed the `normalizeApiTokens` warn log. The PR description now also reports a deploy plus a 34/34 `test-api.sh` and 37/37 live MCP battery. Most previous findings are addressed.

### Resolved
- Token list truncation: pagination loop added plus `test_list_follows_pagination_to_the_end`, which asserts `ExclusiveStartKey` resumption.
- Deploy/live verification: PR now records `VocApiStack` deployed, `scripts/test-api.sh` 34/34, live MCP battery 37/37 including the legacy-token 401 and revocation checks.
- Fail-open mint: `_validate_scopes(None)` now raises; `test_scopes_are_required_not_defaulted` pins it, and `client.ts` makes `scopes` non-optional.
- `mcpTokenSchema.test.ts` mis-named assertion: schema now genuinely strips `secret_hash`/`created_by`, and the test asserts that plus the forward-compat pass-through.
- `_resolve_project_id` silently reading a different project: present-but-unusable argument now raises `InvalidProjectArgument` → `-32602`.
- Misleading "needs a project_id" branch for `none`/`project-set` reach: `reach_covers_nothing` ordering fixes it, with two tests.
- `normalizeApiTokens` no longer logs the payload.
- `shared/test/test_mcp_tokens.py` exists (visible in the incremental diff).

### Still open
- Old `PROJECT#{id} / TOKEN#{id}` rows: no cleanup script or documented manual step, and no confirmation that project-partition reads don't now surface stray `TOKEN#` items.
- `cdk/lib/api-stack.ts`: still not visible — confirm the inline authorizer accepts `voc_{id}_{secret}` and that the `mcpRole` grant carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*` (the live battery passing is good evidence, but the code is unreviewed here).
- `get_caller_subject` import in `projects_handler.py` still not visible in any hunk.

### New findings
- **The mint form still pre-selects every scope** (`frontend/src/pages/ProjectDetail/McpAccessTab.tsx`: `useState<McpScope[]>([...DEFAULT_SCOPES])`, `DEFAULT_SCOPES = MCP_SCOPES` in `mcpTokenSchema.ts`). The backend now refuses to default `scopes` precisely so the laziest request doesn't mint the widest credential — but the default UI path does exactly that, so the change is enforcement-only. Consider starting with none checked (the Generate button is already disabled on an empty set) or an explicit narrower preselect.
- **Naming drift between the two mirrored vocabularies**: Python renamed `DEFAULT_SCOPES` → `ALL_READ_SCOPES` with a docstring saying it "is not a fallback", while `mcpTokenSchema.ts` still exports `DEFAULT_SCOPES`. These modules document each other as mirrors; keep the name (and the "not a default" comment) in step.
- Nit — `_resolve_project_id` in `lambda/api/mcp_handler.py` now uses `'project_id' in args`. For a non-dict `arguments` (e.g. `["project_id"]`) the membership test succeeds and `args['project_id']` raises `TypeError` inside dispatch rather than returning a `-32602`. An `isinstance(args, dict)` guard at the top of `_handle_tools_call` would close this at the boundary.
- Nit — the new `-32602` for a garbage `project_id` is returned *before* the reach check, so a `none`-reach token sending junk is told about the argument, which slightly contradicts the "reach is checked FIRST" rationale the same commit documents.

---

perrozzi:
Round 2 — three accepted, pushed as `a8300e35`, redeployed, live battery **45/45**.

## Accepted

**1. The "nit" about non-dict `arguments` is a 502, and it was mine.** You under-sold this one. `'project_id' in args` and `args['project_id']` both raise `TypeError` for a list/string/number, and that resolution runs *outside* the try/except around the handler — so it escaped as a 502 with no JSON-RPC envelope and no CORS headers.

Worth stating plainly: before round 1 this same input failed *inside* the try/except and became an ordinary `isError` result. Moving the resolution earlier is what turned it into a crash, so this is a regression from my own previous fix rather than a pre-existing edge. Now `-32602` at the boundary, tested across six input shapes through the full `lambda_handler` path (the envelope and status code being what broke), and live-verified for three shapes through API Gateway.

**2. The mint form still pre-selected every scope.** This is the sharpest finding of the two rounds, because it caught an inconsistency in my own argument: I made `scopes` required so the laziest request could not mint the widest credential, then shipped a form whose default path did exactly that. The requirement was enforcement-only.

The form now starts with nothing checked, and Generate stays disabled until a scope is ticked. One tick is enough — the friction should be a choice, not a wall.

**3. Naming drift between the mirrored vocabularies.** Right, and fixing (2) removed the export's only consumer, so `DEFAULT_SCOPES` is **deleted** from the TS module rather than renamed, with a comment on both sides recording that there is deliberately no default scope set.

**4. The ordering comment.** You're right that it contradicted itself. I kept the *code* order — a malformed argument is malformed whatever the token's reach, so syntax before authorization, as everywhere else — and corrected the comment: reach-first applies to the MISSING-argument branch, which is the one a `none`-reach token cannot fix by supplying an argument.

## On the three "still open" items

These are all *"not visible in the diff"*, and for two of them the reason is worth stating, because it is not truncation:

- **`api-stack.ts` is not in this PR at all.** It has zero changes — that is the claim, not an omission. The inline authorizer validates `startsWith('Bearer voc_')` and `length >= 20`; the new credential is 89 chars, so it passes unchanged. `projectsTable.grant(mcpRole, 'dynamodb:Query', 'dynamodb:UpdateItem')` is resource-level with no `LeadingKeys` condition. Both are exercised live: the battery reads `pk=MCPTOKEN` and writes `last_used_at` through API Gateway on every authenticated call.
- **`get_caller_subject`** is likewise unchanged — it was already imported in the `shared.api` block, so there is no hunk to show.

- **Old rows: confirmed 2 of them**, at `PROJECT#{id} / TOKEN#{id}`. They can never authenticate, be listed, or be revoked. The exact cleanup, for the maintainer to run or approve:

  ```bash
  aws dynamodb delete-item --table-name voc-projects-<acct>-<region> \
    --key '{"pk":{"S":"PROJECT#proj_20260717165917"},"sk":{"S":"TOKEN#tok_9b7cc1f172105d13"}}'
  aws dynamodb delete-item --table-name voc-projects-<acct>-<region> \
    --key '{"pk":{"S":"PROJECT#proj_20260603094346"},"sk":{"S":"TOKEN#tok_b6aad3855ff06b76"}}'
  ```

  I have not run it: deleting rows from a live table is the maintainer's call, and I am not adding a migration script to the repo for a two-row one-off in a single deployment.

  Your sub-point (b) does not apply, though, and I checked rather than assumed: those rows were *always* in the project partition — this PR stops adding them, it does not introduce them — and both `get_project` and `projects.get_project` bucket by `sk` prefix (`META` / `PERSONA#` / `PRD#` / `PRFAQ#` / …), so `TOKEN#` items fall through and are ignored. `list_projects` counts by the same prefixes. Nothing enumerates the partition and surfaces unknown `sk` kinds.

## Verification

pytest **2719** · frontend **2969** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean.

Mutation-proved the new guard: removing the `isinstance(arguments, dict)` check fails the six-shape test. Live battery 42 → **45**, adding the three non-object `arguments` shapes.

perrozzi:

Follow-up review — note the full-context diff was **truncated** again, so this is still a partial review (`cdk/lib/api-stack.ts`, `shared/mcp_tokens.py`, the docs/plan updates and the tail of `test_projects_handler.py` are not visible to me).

Since the last round the author removed the `DEFAULT_SCOPES` export, made the mint form start with **no** scope checked (with the test suite reworked around a `selectScopes` helper and a test that pins the empty-by-default state), added an `isinstance(arguments, dict)` boundary guard in `_handle_tools_call` with a parametrized full-`lambda_handler` regression test, and rewrote the ordering comment to explain why malformed-argument reporting precedes the reach check. All four of my previous new findings are addressed.

### Resolved
- Mint form no longer pre-selects every scope; `McpAccessTab.tsx` starts with `useState<McpScope[]>([])` and the reset callback matches.
- Naming drift fixed: `DEFAULT_SCOPES` is gone from `mcpTokenSchema.ts`, replaced by a comment mirroring `ALL_READ_SCOPES`' "not a fallback" note.
- Non-dict `arguments` no longer escapes as a 502: guarded at the boundary in `lambda/api/mcp_handler.py` with `test_non_object_arguments_are_a_protocol_error_not_a_500` asserting envelope, code and CORS headers.
- Ordering rationale for the `-32602` garbage-`project_id` path is now documented (syntax before authorization) rather than contradicting the reach-first comment.

### Still open
- Old `PROJECT#{id} / TOKEN#{id}` rows: no cleanup script or documented manual step, and no confirmation that project-partition reads don't surface stray `TOKEN#` items.
- `cdk/lib/api-stack.ts` remains unreviewable here — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not visible in any hunk.

### New findings
- Minor (interop) — `lambda/api/mcp_handler.py`: the new guard refuses `"arguments": null` with `-32602`. `get_project` and `list_personas` take no required arguments, and some JSON-RPC/MCP clients serialize an omitted optional object as explicit `null`, so those callers now get a protocol error where `{}` would have worked. The test docstring even calls `None` "the one benign case" and then asserts it is refused. Coercing `None` → `{}` before the `isinstance` check (and keeping the refusal for lists/strings/numbers) would close the crash without the compatibility edge.
- Nit — the `DEFAULT_SCOPES` export removal is only verifiable against the files in the incremental diff plus `McpAccessTab.tsx`/`mcpTokenSchema.test.ts`; since the context diff is truncated I can't confirm no other importer exists. Typecheck/eslint being clean per the PR body is adequate evidence if that run post-dates this push.

---

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains a partial review (`cdk/lib/api-stack.ts`, `shared/mcp_tokens.py`, docs/plan and the tail of `test_projects_handler.py` are still not visible to me).

The only change since my last review is the `arguments is None` → `{}` coercion in `lambda/api/mcp_handler.py`, with `null` pulled out of the malformed-arguments parametrize list into its own `test_null_arguments_means_no_arguments_not_a_bad_request`, both driven through a shared `_tools_call_response` helper. That was my one new finding last round and it is fixed; the comment now explains why `params.get('arguments', {})` cannot cover a present-but-null key.

### Resolved
- `"arguments": null` is now treated as no arguments instead of a `-32602`, with a full-`lambda_handler` regression test and a matching revert story; the malformed list keeps lists/strings/numbers/bools.

### Still open
- Old `PROJECT#{id} / TOKEN#{id}` rows: no cleanup script or documented manual step, and no confirmation that project-partition reads don't surface stray `TOKEN#` items.
- `cdk/lib/api-stack.ts` remains unreviewable here — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not visible in any hunk.

### New findings
- Nit — `lambda/api/test/test_mcp_security.py`: `_tools_call_response` now patches `TOOL_HANDLERS` for the tool under test, so the malformed-arguments cases no longer exercise the real `_tool_get_project`. Harmless (the guard returns before dispatch) but the parametrized test's claim that these inputs "reached the project resolution" is now only enforced by the guard's position, not by the handler being real.

Otherwise the delta is clean.

---

perrozzi:
Round 3 — accepted, pushed as `55a135bd`, redeployed, live battery **46/46**.

**`"arguments": null` should mean "no arguments".** Agreed, and the sharp part is that you caught my own test contradicting itself: its docstring called `None` "the one benign case" and then asserted it was refused. `None` is now coerced to `{}` before the type check; lists, strings and numbers are still refused because those are genuinely malformed. Null has its own test asserting it *succeeds*, and it is called out as deliberately absent from the malformed-input list so the two cannot drift back together.

**On the `DEFAULT_SCOPES` nit:** confirmed zero importers remain anywhere under `frontend/` (grep, not just typecheck), and the typecheck/eslint run does post-date the push — both were re-run on this tree before committing.

**The three "still open" items are now in the PR description** rather than only in a comment, so they survive as durable review context:

- The **two orphaned rows** are named with the exact `delete-item` commands. I have deliberately not run them: a two-row one-off in one deployment does not justify a migration script in the repo, and deleting from a live table is the maintainer's call. Leaving them is safe in the meantime, and I checked rather than assumed — those rows were always in the project partition (this PR stops *adding* them), and every project-partition read buckets by `sk` prefix, so `TOKEN#` items fall through ignored.
- **`api-stack.ts` has zero changes in this PR**, which is the claim rather than an omission — so there is nothing for a diff-based review to see. Authorizer accepts the new 89-char credential (`Bearer voc_` prefix + `length >= 20`), and the grant is resource-level with no `LeadingKeys` condition. Both exercised on every authenticated call in the live battery.
- **`get_caller_subject`** was already imported, hence no hunk.

## Cumulative verification across the three rounds

pytest **2719** · frontend **2969** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean.

**Mutation-proved 11/11** — every invariant added across the original change and all three review rounds was removed and its guarding test confirmed to fail. Review-driven tests get the same treatment as feature tests, since a test written to answer a comment is worth nothing if it cannot fail.

Live battery grew **37 → 46** over the rounds. Three of the nine additions exist only because of this review: omitted `scopes` refused, junk `project_id` refused across three shapes, and `"arguments": null` accepted.

Two of your findings were defects I introduced rather than pre-existing ones — the 502 from moving project resolution outside the try/except, and the mint form re-adding the very default the backend had just been changed to refuse. Both were worth the round.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains a partial review (`cdk/lib/api-stack.ts`, `shared/mcp_tokens.py`, the plan doc and the tail of `test_projects_handler.py` are still not visible to me).

The only change since my last review is in `voc-datalake/lambda/api/test/test_mcp_security.py`: `_tools_call_response` now takes an opt-in keyword-only `stub_handler`, used solely by the null-arguments test, so the malformed-arguments parametrize cases once again run against the real `_tool_get_project`. That was my sole new finding last round and it is fixed, with the docstring stating exactly why the stub is opt-in.

### Resolved
- Malformed-`arguments` cases dispatch against the real handler again, so their claim that such input "would reach the project resolution" is enforced by the wiring, not just the guard's position.

### Still open
- Old `PROJECT#{id} / TOKEN#{id}` rows: documented as a manual `delete-item` step in the PR description, but no script or test; noted as the maintainer's call rather than fixed in code.
- `cdk/lib/api-stack.ts` still not visible here — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not shown in any hunk.

Otherwise the delta is clean — a test-only change with no behavioural surface.

---

perrozzi:
Round 4 — accepted, pushed as `c94bcf64`. Test-only change, so no redeploy.

**The `TOOL_HANDLERS` stub was too broad.** You're right that it was harmless but weakened what the test claimed: with the handler mocked, "these inputs would otherwise reach the project resolution" rested on the guard's position rather than on the code being genuinely wired up.

Stubbing is now opt-in and used only by the null-arguments case, which needs to reach a *successful* dispatch without a real project read. The malformed cases run against the real handler, and I re-proved it by mutation: removing the `isinstance` guard still fails them with the real handler in place.

## The three recurring "still open" items

These are re-listed each round because a diff-based review cannot see them, so let me answer once in a way that closes them rather than repeating myself:

- **They are now in the PR description**, not just in comments — the two orphaned rows with their exact `delete-item` commands, and a section listing the files this PR deliberately does *not* touch.
- **`api-stack.ts` has zero changes in this PR.** That is the claim, not an omission: there is nothing for a diff to show. The authorizer checks `Bearer voc_` + `length >= 20`, which the 89-char credential satisfies unchanged, and the grant is resource-level with no `LeadingKeys` condition. Every authenticated call in the live battery exercises both a `pk=MCPTOKEN` read and the `last_used_at` write, which is stronger evidence than reading the file would be.
- **`get_caller_subject`** was already imported in the `shared.api` block, hence no hunk.

I don't think any of the three is actionable further without changing the PR's scope, so I'd treat them as answered.

## Final state

pytest **2719** · frontend **2969** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean. Live battery **46/46** against the deployed stack, `test-api.sh` **34/34**.

**Mutation-proved 12/12** across the original change and four review rounds.

Four rounds found two defects I had introduced — the 502 from moving project resolution outside the try/except, and the mint form re-adding the exact default the backend had just been changed to refuse — plus two places where my own comments or test docstrings claimed something the code did not do. Worth the rounds.

perrozzi:
Converged — your last review found nothing new ("the delta is clean"). Final state below.

**Round 4 fix** (`c94bcf64`): the `TOOL_HANDLERS` stub was too broad. You were right that it left the malformed-arguments tests' central claim resting on the guard's position rather than on the code being wired up. Stubbing is now opt-in and used only by the null-arguments case, which needs a *successful* dispatch without a real project read. Re-proved by mutation: removing the `isinstance` guard still fails the malformed cases with the real handler in place.

**Rebased onto `development`** (`b3dcacbf`) to pick up #349 (CHANGELOG + 0.2.0) before the final deploy. Those commits touch only docs and version fields, but deploying a branch that predates the tip is exactly how a previous PR of mine shipped Lambdas missing someone else's merge — so the merge happens first and the residual diff was checked to contain only my 23 files.

## Final verification, on the merged tree

| Gate | Result |
|---|---|
| pytest | **2719** passed |
| frontend vitest | **2969** passed |
| CDK vitest | **256** passed |
| stream vitest | **339** passed |
| ruff | delta **−1** |
| eslint / typecheck | clean (frontend, CDK, stream) |
| `scripts/test-api.sh` | **34/34** |
| Live MCP battery | **46/46** |
| Mutation proofs | **12/12** |

Deployed to `VocApiStack` us-east-1 with zero synth warnings.

## The three recurring items, answered rather than repeated

- **Two orphaned rows** — documented in the PR description with the exact `delete-item` commands. Not run, and no migration script added: it is a two-row one-off in a single deployment, and deleting from a live table is the maintainer's call. Safe to leave meanwhile, which I checked rather than assumed — every project-partition read buckets by `sk` prefix, so `TOKEN#` items fall through ignored.
- **`api-stack.ts`** — zero changes in this PR, so there is nothing for a diff-based review to see. That is the claim, not an omission. The authorizer accepts the 89-char credential unchanged (`Bearer voc_` + `length >= 20`) and the grant is resource-level with no `LeadingKeys` condition; every authenticated call in the live battery exercises both a `pk=MCPTOKEN` read and the `last_used_at` write.
- **`get_caller_subject`** — already imported, hence no hunk.

## What the four rounds were worth

Two of the findings were defects **I introduced**, not pre-existing ones: the 502 from moving project resolution outside the try/except, and the mint form re-adding the exact default the backend had just been changed to refuse. Two more caught my own comments and test docstrings claiming things the code did not do — a test that asserted `not.toHaveProperty('scope')` under the name "never surfaces a secret hash", and a docstring calling `null` "the one benign case" while asserting it was rejected.

The live battery grew **37 → 46** over the rounds; four of those nine additions exist only because of this review.

Ready for maintainer review.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains partial (`shared/mcp_tokens.py`, `cdk/lib/api-stack.ts`, the plan doc and the tail of `test_projects_handler.py` are still not visible to me).

Since the last round the only change is a real UI bug fix found in a browser: `McpAccessComponents.tsx` now passes `{ nsSeparator: false }` when resolving `mcp.scopeDesc.${scope}`, because the colon in `feedback:read` was being parsed as an i18next namespace separator and rendering the literal `"read"`; `McpAccessTab.test.tsx` gains a regression test asserting the three real description strings and that the bare fallback appears nowhere. That is a correct, minimal fix with the regression test the repo rules ask for.

### Resolved
- Scope descriptions now render the translated text instead of the `"read"` namespace-split fallback, covered by a behavioural regression test.

### Still open
- Old `PROJECT#{id} / TOKEN#{id}` rows remain a manual `delete-item` step in the PR description, with no script or test.
- `cdk/lib/api-stack.ts` still not visible to me — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not shown in any hunk.

### New findings
- Nit (`McpAccessComponents.tsx`): the fix is per-call, so the next colon-bearing key (e.g. a future `mcp.scopeDesc.feedback:write` badge, or a tooltip elsewhere) reintroduces the same silent fallback. Either set `nsSeparator: false` once in the i18n init config, or key the descriptions without colons (`scopeDesc.feedback_read`) and map from the scope constant — both make the class of bug unrepresentable rather than fixed at one site.
- Nit (`McpAccessTab.test.tsx`): `expect(screen.queryAllByText('read')).toHaveLength(0)` is an exact full-text match and will start failing for unrelated reasons if any element's entire text ever normalizes to `read`. Scoping it to the fieldset (`within(...)`) would keep the intent without the tripwire.

Otherwise clean — no behavioural surface beyond the string resolution.

---

perrozzi:
Live verification is now complete, including the two items that were standing open.

## Orphaned rows: deleted

Owner-authorised, so they are gone. Each delete carried a `ConditionExpression` requiring the row to still look legacy (`attribute_exists(token_hash) AND attribute_not_exists(secret_hash)`), which made it impossible to remove a new-format credential even from a stale key list, and the full items were dumped first.

Verified after: **40 → 38 rows, 0 `TOKEN#` rows remain**, all 3 project `META` rows and every artifact kind intact. Both API batteries re-run green afterwards.

## Browser verification — which found a real bug

The API battery cannot see render-time defects, so I drove the mint flow through the deployed site. It caught one:

**Scope descriptions were rendering the i18n fallback.** Scope names contain a colon, which is i18next's default `nsSeparator`, so `t('mcp.scopeDesc.feedback:read')` resolved as namespace `mcp.scopeDesc.feedback` + key `read` — and every checkbox showed the literal word **"read"** instead of its description. Fixed with `nsSeparator: false` in `38cf65e7`.

Worth being precise about why 2969 green tests missed it, because the lesson generalises: the suite asserted the checkbox *names* (which contain the scope, and were correct), nothing asserted the description text, and the broken lookup produced a *plausible word* rather than a visible key path — so it read as intentional. The tests *could* have caught it, since `src/test/setup.ts` loads the real `en` locale, so the gap was a missing assertion rather than the harness. Now asserted and mutation-proved by reverting the option. Those three are the only colon-bearing keys across all 8 locale files, and they all flow through that one call site.

Everything else in the flow checked out against the deployed site: nothing pre-checked with Generate disabled until a scope is ticked; workspace default carrying its warning and the hint switching to the sealed wording on `project-set`; an 89-char `voc_tok_<16hex>_<64hex>` credential on reveal; the correct 30-day deadline on the banner; the row badging "This project" in blue (not amber, since it is sealed) with only the two ticked scopes; the `mcp.json` snippet free of `X-Project-Id`; revoke emptying the list; and **zero console errors or warnings**.

## Final state

pytest **2719** · frontend **2970** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean · `test-api.sh` **34/34** · live MCP battery **46/46** · **mutation-proved 13/13** · browser E2E clean.

Deployed at `38cf65e7`. All three previously-open items are now closed rather than deferred.

perrozzi:
Both nits accepted — but the first one's suggested remedy would have broken two other features, so it is fixed the other way. Pushed as `12379650`, redeployed.

## Nit 1: the per-call fix was fragile — agreed, and it is now unrepresentable

Your diagnosis is right: `nsSeparator: false` at the call site fixes one lookup and leaves the next colon-bearing key to rediscover the same silent fallback.

**Your first suggested remedy is not available here, and the repo already says so.** Setting `nsSeparator: false` in the init config would break the app: `src/i18n/options.ts` documents that `SCORABLE_TYPE_META` resolves namespace-qualified keys like `prioritization:docType.prd`, and disabling the separator makes the Prioritization badge and the Feedback Forms document picker render raw key paths. `prioritizationUtils.test.ts` pins that deliberately.

So I took your **second** option — fix the key, not the config: `mcp.scopeDesc.feedback:read` → `mcp.scopeDesc.feedback_read` across all 8 locales, derived at the call site from the scope constant. No colon, no separator interaction, nothing else touched.

**And a guard so the class cannot return**, which is your actual goal rather than the specific mechanism: `src/i18n/i18nKeys.test.ts` walks every key path in all 8 locales across every namespace and fails on **any** key containing a colon, with a failure message that explains the mis-resolution *and* tells the next person explicitly not to reach for `nsSeparator: false`. That covers namespaces this change never touched. Mutation-proved by reinstating a colon key. It also carries a positive control asserting it found >20 files across 8 locales, so a glob matching nothing can't make it vacuously green.

I verified the rejected option really would have been a regression rather than taking the docs' word for it: after deploying, the Prioritization page renders resolved badges ("PRD", "PR/FAQ") with **zero** raw `prioritization:` or `docType.` key paths on the page.

## Nit 2: the unscoped assertion was a tripwire — agreed

`queryAllByText('read')` is an exact full-text match over the whole page, so any unrelated element that ever normalized to `read` would have failed that test for a reason having nothing to do with it. A false alarm is worse than no alarm. Now scoped to the scopes fieldset with `within()`, which keeps the intent without the surface.

## The three recurring items

All three are closed, and two of them are answered by the PR description rather than the diff:

- **Orphaned rows: deleted.** Not a manual step any more — done in this session with owner authorisation, each delete carrying `attribute_exists(token_hash) AND attribute_not_exists(secret_hash)` so a new-format credential could not be removed even from a stale key list. Verified 40 → 38 rows, 0 `TOKEN#` remaining, all 3 project METAs and every artifact kind intact.
- **`api-stack.ts` has zero changes in this PR**, so there is nothing for a diff-based review to see. Authorizer accepts the 89-char credential (`Bearer voc_` + `length >= 20`); the grant is resource-level with no `LeadingKeys` condition. Both exercised on every authenticated call in the live battery.
- **`get_caller_subject`** was already imported, hence no hunk.

## Verification

frontend **2972** (183 files) · pytest **2719** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean · `npm run i18n:check` no new findings · `test-api.sh` **34/34** · live MCP battery **46/46** · **mutation-proved 15/15** · browser E2E clean with zero console errors.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains partial (`shared/mcp_tokens.py`, `cdk/lib/api-stack.ts`, the plan doc and the tail of the Python token tests are still not visible to me).

Since the last round the i18n fix was reworked from a per-call `nsSeparator: false` to the key side: `mcp.scopeDesc` entries are renamed to underscore form in all 8 locale files, the component derives the key with `scope.replace(':', '_')`, and a new `src/i18n/i18nKeys.test.ts` fails on **any** colon-bearing locale key across every namespace. The `queryAllByText('read')` tripwire is now scoped to the scopes fieldset. Both nits from the previous round are addressed, and the chosen fix is the stronger one — the note that `prioritization:docType.prd` genuinely depends on `nsSeparator` makes the global-disable option correctly off the table.

### Resolved
- The class of bug is now unrepresentable at test time rather than fixed at one call site (`src/i18n/i18nKeys.test.ts`), and all 8 locales were renamed consistently.
- `McpAccessTab.test.tsx` negative assertion is scoped via `within(getByRole('group', …))`, removing the whole-page exact-text tripwire.
- Per the PR description the orphaned `PROJECT#{id} / TOKEN#{id}` rows have been deleted under a legacy-shape `ConditionExpression` and verified, so the manual-cleanup concern is closed operationally.

### Still open
- `cdk/lib/api-stack.ts` still not visible to me — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not shown in any hunk.

### New findings
- Nit (`McpAccessComponents.tsx` + locales): nothing asserts that every `MCP_SCOPES` entry has a matching `mcp.scopeDesc.<underscored>` key in each locale. Adding a scope without the key now renders a raw key path rather than a plausible word — visible, so not a repeat of the original bug — but a small loop in `i18nKeys.test.ts` (or `McpAccessTab.test.tsx`) over `MCP_SCOPES` would close the derivation gap the `replace(':', '_')` introduces.
- Nit (`src/i18n/i18nKeys.test.ts`): `expect(new Set(...).size).toBe(8)` hard-codes the locale count, so adding a 9th locale fails a test about colon keys. `toBeGreaterThanOrEqual(8)` keeps the positive control without the unrelated failure.

Otherwise clean — the delta is string-resolution plumbing plus a structural guard, with no behavioural surface.

---

perrozzi:
Both accepted, pushed as `0e69a402`. Test-only, no behavioural surface, so no redeploy.

**1. The derivation gap is now pinned.** Correct catch — `replace(':', '_')` made the link between `MCP_SCOPES` and the locale entries implicit, and Phase 3 adds a write scope, so this was going to bite on a schedule. `i18nKeys.test.ts` now loops every scope × every locale and fails on a missing or blank description.

I put it there rather than in the component test on purpose: a component test exercises only `en`, while this covers all 8. Mutation-proved twice — dropping one key from one locale fails it, and adding `projects:write` to `MCP_SCOPES` with no locale entry fails it, which is precisely the Phase 3 scenario.

Your severity read is right and worth keeping in the record: a visible `mcp.scopeDesc.x` is obviously broken, where the original literal `"read"` looked like real copy. That's why this one is a guard rather than a fix.

**2. The hard-coded locale count is relaxed.** Agreed — `toBe(8)` would have made adding a 9th language fail a test about colon keys, which teaches the next person nothing. Now `toBeGreaterThanOrEqual(8)`, with a note that the exact count belongs to `i18n-check`, not here.

**Also fixed something no gate covers.** While verifying, I found a type error of my own: `McpAccessTab.test.tsx` annotated `mockToken` as `ApiToken` without importing the type. `typecheck` excludes test files and `typecheck:tests` is not part of `npm run check`, so test-file types are effectively ungated in this repo — worth knowing independently of this PR.

## State

frontend **2973** (183 files) · pytest **2719** · CDK **256** · stream **339** · ruff delta **−1** · eslint + typecheck clean · `typecheck:tests` clean for every file I touched · `test-api.sh` **34/34** · live MCP battery **46/46** · **mutation-proved 17/17** · browser E2E clean.

## The two remaining items

Both are structural limits of a diff-based review rather than open work, so I'd treat them as answered:

- **`api-stack.ts` has zero changes in this PR.** There is no hunk because there is no edit. The authorizer validates `Bearer voc_` + `length >= 20`, which the 89-char credential satisfies; the grant is resource-level with no `LeadingKeys` condition. Every authenticated call in the live battery exercises the authorizer, a `pk=MCPTOKEN` read and the `last_used_at` write — stronger evidence than reading the file.
- **`get_caller_subject`** was already imported in the `shared.api` block, so likewise no hunk.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains partial (`shared/mcp_tokens.py`, `cdk/lib/api-stack.ts`, the plan doc and the tail of the Python token tests are still not visible).

The delta is two test-only changes: `i18nKeys.test.ts` gains a loop asserting every `MCP_SCOPES` entry has a non-empty `mcp.scopeDesc.<underscored>` entry in each locale (with its own positive controls), and the locale-count positive control is relaxed from `toBe(8)` to `toBeGreaterThanOrEqual(8)`. Both nits from the previous round are addressed; no behavioural surface changed.

### Resolved
- The derivation gap `scope.replace(':', '_')` opened is now covered across all locales, so a Phase 3 write scope without locale entries fails at test time (`src/i18n/i18nKeys.test.ts`).
- Locale-count assertion no longer hard-codes 8, so adding a 9th language cannot fail a test about colon keys.

### Still open
- `cdk/lib/api-stack.ts` not visible to me — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not shown in any hunk.

### New findings
- Nit (`src/i18n/i18nKeys.test.ts`): `(json as { mcp?: { scopeDesc?: Record<string, unknown> } })` is a type assertion, which the rest of this PR goes out of its way to avoid (see the copy-and-delete comment in `mcpTokenSchema.ts` citing the codebase's no-assertion rule). A small `unknown`-narrowing guard, or reusing the existing `keyPaths()` helper to look for `mcp.scopeDesc.<key>` in the flattened path list, would keep the test consistent with that rule.

Otherwise clean.

---

perrozzi:
Accepted — pushed as `5ce5293d`. Test-only, no behavioural surface, no redeploy.

**You caught an inconsistency in my own work.** This PR added a comment in `mcpTokenSchema.ts` explaining that I used copy-and-delete specifically to avoid a type assertion, then I used `(json as { mcp?: { scopeDesc?: ... } })` two commits later. Worth fixing on that basis alone.

Worth noting *why* nothing stopped it: `eslint.config.js` ignores test files outright, so `consistent-type-assertions` is advisory in `*.test.ts`. Combined with `typecheck` excluding tests and `typecheck:tests` not being in `npm run check`, test files in this repo are unlinted for assertions *and* untypechecked by the gate. That's a standing gap rather than anything about this change.

**I took your second suggestion, and it is less code than the cast.** The existing walker now returns `{ path, value }` leaves rather than paths only, so both tests share it: the colon check maps to `.path`, and the scope-coverage check looks the flattened `mcp.scopeDesc.<key>` path up in a `Map`. An `isRecord` guard narrows `unknown`, so the file now holds no assertion at all.

**One near-miss, recorded because the failure mode is quiet.** My first draft of that comment contained the glob `**` + `/*.test.ts` — whose `*/` terminated the JSDoc block early and turned the prose into code. **vitest reported "no tests" rather than a parse error**, which is easy to skim past; `typecheck:tests` is what surfaced it. Rephrased to avoid the sequence.

**Re-proved 4/4 after the refactor**, since reworking the walker could have made the assertions vacuous: a locale missing one key, a new scope with no entry (the Phase 3 case), a present-but-blank entry, and a reintroduced colon key each fail the suite.

## State

frontend **2973** (183 files) · pytest **2719** · CDK **256** · stream **339** · ruff delta **−1** · eslint clean · typecheck clean · `typecheck:tests` clean for every file I touched · `test-api.sh` **34/34** · live MCP battery **46/46** · **mutation-proved 19/19** · browser E2E clean.

The two `api-stack.ts` / `get_caller_subject` items remain answered-not-open: neither file is edited by this PR, so there is no hunk for a diff-based review to show, and both claims are exercised on every authenticated call in the live battery. Details are in the PR description.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains partial (`shared/mcp_tokens.py`, `cdk/lib/api-stack.ts`, the plan doc and the tail of the Python token tests are still not visible).

The delta is a single test-only refactor: `keyPaths` becomes `leafEntries` returning `{ path, value }` behind an `isRecord` type guard, and the scope-coverage check now looks the description up in the flattened path map instead of casting into `mcp.scopeDesc`. Behaviour of both assertions is unchanged (empty objects still yield no leaf, arrays still land as a non-string value and are reported as blank), and the last outstanding nit is gone.

### Resolved
- The `as { mcp?: { scopeDesc?: … } }` assertion in `src/i18n/i18nKeys.test.ts` is replaced by an `unknown`-narrowing guard, consistent with the no-assertion rule cited in `api/mcpTokenSchema.ts`.

### Still open
- `cdk/lib/api-stack.ts` not visible to me — confirm the inline authorizer accepts `voc_{id}_{secret}` and that `mcpRole` carries no `dynamodb:LeadingKeys` condition pinned to `PROJECT#*`.
- `get_caller_subject` import in `projects_handler.py` still not shown in any hunk.

---

PR #355, opened 2026-08-19
feat(mcp): delegate every tool to the domain function that owns its route

perrozzi:
Phase 2a of `analysis/PLAN-mcp-platform-server.md` (decision **D1**). The MCP Lambda stops reading the data lake and becomes a protocol adapter: validate the credential, resolve scope and reach, translate the tool's arguments into a route call, synthesize an API Gateway proxy event, invoke the owning domain function, reshape the answer.

Phase 2 in the plan is a large bundle. This is the **data-path half**; the protocol half (`server/discover`, version range, `resultType`, header validation, retiring `ANY /mcp/{proxy+}`, annotations, scope-filtered `tools/list`) is deliberately a separate PR — one changes what the tools read, the other changes the envelope around them, and they review on different terms.

## Why

A single Lambda holding the union of every domain's permissions inverts the domain-isolation pattern this repo mandates, and marches at the 20 KB role-policy ceiling with no synth-time warning. It is also a second implementation of every route it duplicates — and one had already drifted (see DUP2 below).

## The IAM result

This is the headline, and it is a narrowing rather than a trade:

- `mcpRole` **loses** its feedback-table and aggregates-table grants outright.
- Its projects-table grant — the last one left — **gains a `dynamodb:LeadingKeys` condition** naming only the `MCPTOKEN` partition. The one function in this stack reachable with a bearer token instead of a Cognito session can no longer read a persona, a PRD, a PR/FAQ or a prototype through its own credentials at all.
- It **gains** `lambda:InvokeFunction` on exactly two functions, by exact ARN. Not `grantInvoke`, which also grants `<fn>.Arn:*` — every version and alias — and would need a cdk-nag IAM5 suppression for a wildcard nothing uses.

Delegating autoseed is what made the condition possible: it was the last in-process reader of project artifact rows.

## It retires a filed defect by construction

`DUPLICATED-CODE`'s **DUP2**: the in-process `get_project` recognised **2 of 6** document sort-key prefixes, so an MCP client saw a third of a project's documents and was told nothing had been filtered. The route always knew all six. Visible in the live run below — `document`, `prototype` and `research` are kinds no MCP client could previously see.

## Breaking changes → server 2.0.0

`serverInfo.version` goes to `2.0.0`. Tool output shapes change:

- results carry `structuredContent` alongside the text block, and every tool declares an `outputSchema`;
- sad paths that used to arrive as prose inside a **successful** result are now tool errors (`isError: true`) — a model told `isError: false` treats "not found" as data and reports the item as empty instead of retrying;
- `get_metrics_summary` reports the route's shape, gaining `avg_sentiment` and `urgent_count`.

`get_metrics_breakdown` is added **in this PR rather than the next** because `/metrics/summary` does not carry the per-sentiment and per-category counts the old in-process tool computed. Delegating alone would have been a capability regression; with it, the pair reports strictly more than before.

## Two shapes where the obvious reading of the plan is wrong

- **`search_feedback` delegates to two routes.** `/feedback/search` refuses a query under two characters, so filter-only calls go to `/feedback`. Mapping the tool onto `/feedback/search` alone — what Appendix A implies — would make every filter-only call return nothing.
- **`_resolve_days` is deleted, not ported.** The window is bounded by the route's shared `validate_days`, which clamps rather than raises. One behaviour genuinely changes: `days: true` now means a 1-day window instead of the 7-day default, because `validate_int` coerces bools. Recorded next to the reasoning in the test file — having one endpoint treat that input differently from the other forty is a worse property than either answer.

## A bug found by verifying live

Both feedback tools projected `item.get('id')`. The processor never writes a plain `id` — the identifier is `feedback_id`, which is also what `GET /feedback/{id}` keys its GSI on. So `search_feedback` advertised an `id` and filled it with `""` for every item in the corpus, leaving `get_feedback_detail` unreachable: the only way to learn a feedback id is to search, and search reported none. Fixed at the shared projection (second commit). No unit test could have caught it — every fixture in the suite sets both fields.

## Tests

- **`TestClaimSynthesis`** — the highest-value new test. Synthesizing claims makes this function an authorization authority: it tells the domain Lambda who is calling and that Lambda believes it. Identity derives **only** from the stored credential, enforced *structurally* — `synthetic_claims` takes the token record and never sees the request, so there is nothing to filter and nothing to forget to filter. Pinned with a request that tries: arguments named `sub`, `cognito:groups`, a nested `requestContext`.
- **`test_mcp_delegation.py`** (38) — route selection, error mapping, projections, the synthesized event, structured output.
- The Python IAM lockstep got **stronger**: every tool now pays exactly one projects-table query (the credential lookup), and autoseed must not touch that table at all.
- **`api-stack.test.ts`** — the `LeadingKeys` condition (read from the Python constant, not re-typed), the exact invoke set with no `:*`, absence of the two dropped grants, and a `DOMAIN_ROUTES` lockstep proving every delegated route exists in its handler and is wired in API Gateway.
- **First policy-size measurement in the repo.** The 20 KB IAM limit that shapes this entire architecture was previously unmeasured anywhere in CI. Documented as an approximation (the template carries `Fn::GetAtt`, not resolved ARNs).
- `TestPartialResultReporting` (17) is **deleted, not relocated**: the behaviour is gone. There are no per-read failures to degrade now, and a failing route is reported as an error rather than a plausible answer with a flag on it. The reasoning is left in the file where the class was.

**Mutation-proved: 14 mutations, 14 killed, 0 survived.** One found a genuinely weak assertion of mine — a 5xx-detail check a leak walked straight through — now exact equality.

## Verification

| Gate | Result |
|---|---|
| backend pytest | **2911** (baseline 2881) |
| ruff | **−9**, zero new findings |
| CDK vitest | **266** (baseline 256) |
| stream / frontend | 339 / 2975, tsc + eslint clean |
| `cdk synth` | **zero warnings** |
| template baseline | regenerated; only the SHA moved, no physical name changed → no CloudFormation replacement. Still five stacks |

**Deployed to `VocApiStack` and verified live** (account 777200923596, us-east-1):

- MCP battery **75/75**, including: the conditioned `Query` *and* `UpdateItem` both permitted, `structuredContent` present, all four breakdown axes, a 404 arriving as `isError` inside a 200, both `search_feedback` branches, the search→detail chain, and autoseed answering the route's 404 instead of a 500.
- `test-api.sh`: **33/33** endpoints 200, zero non-2xx; `POST /chat/stream` SSE verified separately (metadata → text → done).
- No browser E2E: this branch touches **no** frontend file, so the deployed bundle is byte-identical to the one Phase 1 verified in-browser.

Every API Lambda's code hash moves, because `shared/mcp_delegate.py` is bundled into all of them by `createApiLambdaCode`. Behaviour for the other fourteen is unchanged — they gain an unused module — and `test-api.sh` covers them.

### DIFF
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -179,9 +179,13 @@ def _reserved_for(template: str, name: str) -> frozenset[str]:
     position = segments.index(f'{{{name}}}')
     prefix = '/' + '/'.join(segments[:position])
     if prefix not in _RESERVED_PATH_SEGMENTS:
-        raise InvalidToolArgument(
-            f'{name} cannot be validated: no reserved-segment set declared for {prefix}'
-        )
+        # DelegationUnavailable, not InvalidToolArgument: this is a SERVER
+        # misconfiguration, and -32602 "Invalid params" would tell the caller its
+        # arguments are wrong when nothing it could send would work. The state is
+        # unreachable in a deployed build — the prefix lockstep fails first — so
+        # this is about not lying if it ever is reached.
+        logger.error('No reserved-segment set declared', extra={'prefix': prefix})
+        raise DelegationUnavailable(f'no reserved-segment set declared for {prefix}')
     return _RESERVED_PATH_SEGMENTS[prefix]
 
 
@@ -196,8 +200,13 @@ def _validated_path_parameters(route_key: str, params: dict[str, str]) -> dict[s
         # Distinct messages per condition: the caller can act on each of these
         # differently, and "must be a single path segment" for a stray space sends
         # them looking for a slash they never sent.
-        if not isinstance(value, str) or not value or value.strip() != value:
+        if not isinstance(value, str) or not value:
             raise InvalidToolArgument(f'{name} must be a non-empty identifier')
+        if value.strip() != value:
+            # Split from the emptiness check on purpose: bundling them reported
+            # "must be a non-empty identifier" for a value that plainly was not
+            # empty, which is the same misdirection the messages below avoid.
+            raise InvalidToolArgument(f'{name} must not be surrounded by whitespace')
         offending = sorted(_FORBIDDEN_PATH_CHARS & set(value))
         if offending:
             raise InvalidToolArgument(
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -59,6 +59,20 @@ metrics = Metrics(namespace="VoC-MCP")
 projects_table = get_projects_table()
 
 # MCP Protocol version
+#
+# ⚠️ KNOWN SKEW, and it is temporary by plan rather than by oversight.
+# `structuredContent` and `outputSchema` (below) arrived in the 2025-06-18
+# revision, while this still advertises 2024-11-05 and `initialize` declares no
+# capability for them. The version range — negotiating 2026-07-28 / 2025-11-25 /
+# 2025-06-18 and answering `server/discover` — is the next phase of the plan, and
+# bumping the number here without that negotiation would claim conformance this
+# handler does not yet have (no `resultType`, no header validation).
+#
+# Shipping the structured fields early is safe in the meantime because both are
+# ADDITIVE: a 2024-11-05 client reads `content[0].text`, which still carries the
+# same payload serialized, and JSON-RPC clients ignore result fields they do not
+# know. The reverse order — negotiate first, add the fields later — would have
+# meant two breaking output changes instead of one.
 MCP_PROTOCOL_VERSION = "2024-11-05"
 
 # Semver on the SERVER, independent of the protocol revision above. 2.0.0
@@ -104,6 +118,52 @@ DOMAIN_ROUTES: dict[str, tuple[str, str, str]] = {
 }
 
 
+# What each path parameter must look like before it is allowed into a route path.
+#
+# 🔑 This is a ROUTE-CONFUSION guard, not input tidiness. The delegated path is
+# what the Powertools resolver matches on, so a parameter value is not data — it
+# is part of the routing key. Two ways an unvalidated value changes which route
+# answers:
+#
+#   • extra segments: project_id='p/api-tokens' builds '/projects/p/api-tokens'
+#     and lands on the token-list route instead of the project route;
+#   • a value that collides with a STATIC sibling: project_id='prioritization'
+#     builds '/projects/prioritization', and Powertools checks static routes
+#     BEFORE dynamic ones, so it lands on api_get_prioritization_scores — the
+#     surface deliberately excluded from MCP entirely. Segment counting does not
+#     catch this one, which is why these are format checks rather than a
+#     "no slashes" rule. The same trick reaches /feedback/search, /feedback/urgent
+#     and /feedback/entities through get_feedback_detail.
+#
+# Both formats are what the writers actually produce: `proj_` + a timestamp
+# (projects.py) and a 32-character hex digest (processor/handler.py). Anything
+# else fails CLOSED with a -32602, which is the right direction for a component
+# that decides both the route and the scope that applies to it. Pinned by
+# TestPathParameterConfinement, including a lockstep against the id generators.
+_PATH_PARAM_PATTERNS: dict[str, re.Pattern[str]] = {
+    'project_id': re.compile(r'^proj_[A-Za-z0-9_-]+$'),
+    'feedback_id': re.compile(r'^[0-9a-f]{32}$'),
+}
+
+
+def _validated_path_parameters(route_key: str, params: dict[str, str]) -> dict[str, str]:
+    """Refuse any path parameter that could change which route answers."""
+    for name, value in params.items():
+        pattern = _PATH_PARAM_PATTERNS.get(name)
+        if pattern is None:
+            # Fail closed on an UNDECLARED parameter rather than passing it
+            # through: a future route whose author forgets to add a pattern
+            # would otherwise reopen exactly this hole, silently.
+            raise InvalidToolArgument(
+                f"{route_key}: no validation declared for path parameter '{name}'"
+            )
+        if not isinstance(value, str) or not pattern.match(value):
+            # The expected SHAPE is not echoed back: it is a storage detail, and
+            # a caller who guessed wrong is better served by the parameter name.
+            raise InvalidToolArgument(f'{name} is not a valid identifier')
+    return params
+
+
 def _domain_call(
     route_key: str,
     *,
@@ -115,9 +175,12 @@ def _domain_call(
     The env var is read per call rather than captured at import so a test can
     set it, and so a missing one surfaces as this route being unconfigured
     rather than as the whole module failing to load.
+
+    Every delegated call is built here — tools and the autoseed route alike — so
+    this is the one place path parameters have to be confined.
     """
     domain, method, template = DOMAIN_ROUTES[route_key]
-    params = dict(path_parameters or {})
+    params = _validated_path_parameters(route_key, dict(path_parameters or {}))
     return DomainCall(
         function_name=os.environ.get(_DOMAIN_FUNCTION_ENV[domain], ''),
         method=method,
@@ -819,6 +882,9 @@ MCP_TOOLS = [
         "outputSchema": {
             "type": "object",
             "properties": _FEEDBACK_DETAIL_PROPERTIES,
+            # Every key is always emitted (the projection uses typed defaults),
+            # so declaring them costs nothing and lets a client rely on them.
+            "required": sorted(_FEEDBACK_DETAIL_PROPERTIES),
             "additionalProperties": False,
         },
     },
@@ -916,12 +982,16 @@ def _tool_search_feedback(args: dict, token_info: dict) -> ToolResult:
         call = _domain_call('feedback_list', query=shared_filters)
 
     body = _delegate(call, token_info).payload
-    items = body.get('items', []) if isinstance(body, dict) else []
+    raw_items = body.get('items', []) if isinstance(body, dict) else []
+    # Projected FIRST, then counted, so `count` describes what the caller
+    # received. Counting the route's list instead would let a non-dict entry make
+    # `count` exceed `len(items)` in the same payload.
+    items = [_project_feedback(item, summary=True) for item in raw_items
+             if isinstance(item, dict)]
     return ToolResult({
         "count": len(items),
         "query": query,
-        "items": [_project_feedback(item, summary=True) for item in items
-                  if isinstance(item, dict)],
+        "items": items,
     })
 
 
@@ -1502,10 +1572,17 @@ def _handle_autoseed(event: dict) -> dict:
     # through as the route's own query string rather than re-parsed here — the
     # comma-splitting used to be duplicated in both places.
     query_params = event.get('queryStringParameters') or {}
-    call = _domain_call('project_autoseed', path_parameters={'project_id': project_id}, query={
-        'persona_ids': query_params.get('persona_ids'),
-        'document_ids': query_params.get('document_ids'),
-    })
+    try:
+        call = _domain_call('project_autoseed', path_parameters={'project_id': project_id}, query={
+            'persona_ids': query_params.get('persona_ids'),
+            'document_ids': query_params.get('document_ids'),
+        })
+    except InvalidToolArgument as exc:
+        # This route takes its project id straight from the URL, so the
+        # route-confusion guard matters here as much as on a tool call — and it
+        # is a 400 rather than a 403, because the credential is fine and the
+        # path is not.
+        return _cors_response({'message': str(exc)}, status_code=400)
     try:
         result = call_domain(call, claims=synthetic_claims(token_info))
     except DelegationUnavailable:
@@ -1513,7 +1590,12 @@ def _handle_autoseed(event: dict) -> dict:
         return _cors_response({'message': 'Upstream service unavailable'}, status_code=502)
     # The route's own status travels with its body: a 404 for an unknown project
     # stays a 404 here instead of becoming the 500 this used to answer for it.
-    return _cors_response(result.payload, status_code=result.status_code)
+    #
+    # An empty upstream body becomes `{}` rather than `null`: this route's clients
+    # read fields off the response, and `null` makes that a TypeError where `{}`
+    # makes it a missing key.
+    return _cors_response(result.payload if result.payload is not None else {},
+                          status_code=result.status_code)
 
 
 # ============================================
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -118,49 +118,80 @@ DOMAIN_ROUTES: dict[str, tuple[str, str, str]] = {
 }
 
 
-# What each path parameter must look like before it is allowed into a route path.
+# ---------------------------------------------------------------------------
+# Path-parameter confinement
+# ---------------------------------------------------------------------------
 #
-# 🔑 This is a ROUTE-CONFUSION guard, not input tidiness. The delegated path is
-# what the Powertools resolver matches on, so a parameter value is not data — it
-# is part of the routing key. Two ways an unvalidated value changes which route
-# answers:
+# 🔑 A ROUTE-CONFUSION guard, not input tidiness. The delegated path is what the
+# Powertools resolver matches on — `pathParameters` is never consulted — so a
+# parameter value is not data, it is part of the routing key. Two ways an
+# unvalidated value changes which route answers:
 #
-#   • extra segments: project_id='p/api-tokens' builds '/projects/p/api-tokens'
-#     and lands on the token-list route instead of the project route;
-#   • a value that collides with a STATIC sibling: project_id='prioritization'
-#     builds '/projects/prioritization', and Powertools checks static routes
-#     BEFORE dynamic ones, so it lands on api_get_prioritization_scores — the
-#     surface deliberately excluded from MCP entirely. Segment counting does not
-#     catch this one, which is why these are format checks rather than a
-#     "no slashes" rule. The same trick reaches /feedback/search, /feedback/urgent
-#     and /feedback/entities through get_feedback_detail.
+#   • EXTRA SEGMENTS. `project_id='p/api-tokens'` builds `/projects/p/api-tokens`
+#     and lands on the token-list route instead of the project route.
+#   • A COLLISION WITH A STATIC SIBLING. `project_id='prioritization'` builds
+#     `/projects/prioritization`, and Powertools resolves static routes BEFORE
+#     dynamic ones, so it lands on `api_get_prioritization_scores` — the surface
+#     deliberately excluded from MCP altogether. Segment counting does not catch
+#     this: it is a well-formed single segment.
 #
-# Both formats are what the writers actually produce: `proj_` + a timestamp
-# (projects.py) and a 32-character hex digest (processor/handler.py). Anything
-# else fails CLOSED with a -32602, which is the right direction for a component
-# that decides both the route and the scope that applies to it. Pinned by
-# TestPathParameterConfinement, including a lockstep against the id generators.
-_PATH_PARAM_PATTERNS: dict[str, re.Pattern[str]] = {
-    'project_id': re.compile(r'^proj_[A-Za-z0-9_-]+$'),
-    'feedback_id': re.compile(r'^[0-9a-f]{32}$'),
+# The guard is a SHAPE rule plus a reserved-segment set, deliberately NOT a
+# format allowlist per id. An allowlist (`proj_…`, 32 hex) was the first
+# implementation and it bet on id PROVENANCE: any project seeded, imported or
+# minted by an older generator becomes permanently unreachable through MCP, and
+# reported as a malformed argument rather than as a missing project. That trades
+# a security property for an availability one, and it is avoidable — the shape
+# rule stops segment injection, and the reserved set stops sibling collisions,
+# without either caring where an id came from.
+#
+# The reserved sets are DERIVED from the owning handlers' static routes by
+# `test_reserved_segments_cover_every_static_sibling`, so a new static sibling
+# added to one of those handlers fails the suite instead of quietly becoming
+# reachable. That is the same lockstep shape as the route and throttle tests.
+_RESERVED_PATH_SEGMENTS: dict[str, frozenset[str]] = {
+    '/projects': frozenset({'config', 'prioritization'}),
+    '/feedback': frozenset({'search', 'urgent', 'entities'}),
 }
+
+# Characters that would let a value escape its segment or its path entirely.
+# `/` is the injection above; `?` and `#` would graft a query or fragment onto
+# the routing key; `%` would let a percent-encoded `/` arrive decoded at a
+# resolver that has already matched.
+_FORBIDDEN_PATH_CHARS = frozenset('/?#%\\')
+
+
+def _reserved_for(template: str, name: str) -> frozenset[str]:
+    """The static sibling segments a parameter must not impersonate.
+
+    Keyed on the path PREFIX above the parameter, so `/projects/{project_id}` and
+    `/projects/{project_id}/autoseed` share one set — both put the value in the
+    same position, which is what decides which siblings it can collide with.
+    """
+    segments = template.strip('/').split('/')
+    position = segments.index(f'{{{name}}}')
+    return _RESERVED_PATH_SEGMENTS.get('/' + '/'.join(segments[:position]), frozenset())
 
 
 def _validated_path_parameters(route_key: str, params: dict[str, str]) -> dict[str, str]:
     """Refuse any path parameter that could change which route answers."""
+    _domain, _method, template = DOMAIN_ROUTES[route_key]
     for name, value in params.items():
-        pattern = _PATH_PARAM_PATTERNS.get(name)
-        if pattern is None:
-            # Fail closed on an UNDECLARED parameter rather than passing it
-            # through: a future route whose author forgets to add a pattern
-            # would otherwise reopen exactly this hole, silently.
+        if f'{{{name}}}' not in template:
+            # Fail closed rather than ignoring it: a parameter the template does
+            # not interpolate means caller and route disagree about the call.
+            raise InvalidToolArgument(f"{route_key}: unexpected path parameter '{name}'")
+        if not isinstance(value, str) or not value or value.strip() != value:
+            raise InvalidToolArgument(f'{name} must be a non-empty identifier')
+        if _FORBIDDEN_PATH_CHARS & set(value):
+            raise InvalidToolArgument(f'{name} must be a single path segment')
+        if value in {'.', '..'} or any(c.isspace() or ord(c) < 0x20 for c in value):
+            raise InvalidToolArgument(f'{name} must be a single path segment')
+        if value in _reserved_for(template, name):
+            # Named explicitly, because "not found" would be a lie and the caller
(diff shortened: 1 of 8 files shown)


perrozzi:

This PR converts the MCP Lambda from a second implementation of five domain routes into a protocol adapter: it authenticates the bearer credential, resolves scope/reach, synthesizes an API Gateway proxy event with claims derived structurally from the token record, and invokes the owning domain function (`voc-datalake/lambda/shared/mcp_delegate.py`). The IAM outcome is a real narrowing — feedback/aggregates grants dropped, projects grant confined to the `MCPTOKEN` partition via `dynamodb:LeadingKeys`, plus two exact-ARN `lambda:InvokeFunction` grants — and it is pinned from both the Python and CDK sides. The claim-synthesis design (`synthetic_claims` takes only the token record, so the request is not in scope to leak from) is the right shape for making this function an authorization authority, and `TestClaimSynthesis` pins it well. Overall this is high-quality work; my findings are about the new coupling surface the delegation introduces, not the direction.

### Should fix

- **`structuredContent`/`outputSchema` ship ahead of the negotiated protocol revision.** `MCP_PROTOCOL_VERSION` stays `"2024-11-05"` (`lambda/api/mcp_handler.py`) while every tool now declares `outputSchema` and every success carries `structuredContent` — both introduced in the 2025-06-18 revision. A client that negotiated 2024-11-05 has no contract telling it to validate `structuredContent`, and `initialize` advertises no capability for it. Since the version-range work is explicitly deferred to the next PR, at minimum note in the handler why the two are temporarily out of step, or bump the advertised revision here.

- **Path parameters are interpolated into the routing key without validation or encoding.** `_domain_call` builds `path=template.format(**params)` and `build_proxy_event` puts that string in `event['path']`, which is exactly what the Powertools resolver matches on (`pathParameters` is never consulted). `feedback_id` is only checked for non-emptiness, so `get_feedback_detail({"feedback_id": "search"})` synthesizes `GET /feedback/search` and lands on the *search* route rather than the item route. No privilege gain today (method is fixed to GET and both routes are feedback-domain), but the adapter's route choice is currently user-influenceable, which is a bad property for the component that also decides which scope applies. Reject `/`, `?`, `#` and reserved segment names in path params, or assert after formatting that the concrete path still matches the declared template.

- **`_document_kind` depends on an internal storage key surviving the route's serialization.** It reads `item.get('sk', '')`; if `GET /projects/{project_id}` ever strips or renames `sk` in its response projection, every document silently reports `kind: ""` — and the `outputSchema` enum explicitly permits `""`, so nothing fails. This is precisely the fixture trap described for the `feedback_id` bug: `test_every_document_kind_is_reported` supplies `sk` itself, so it cannot detect the route dropping it. Either derive `kind` from a field the route contractually returns, or add a lockstep test that reads `projects_handler.py`'s document projection the way `api-stack.test.ts` reads `DOMAIN_ROUTES`.

- **The new policy-size guard measures the wrong resource set and possibly the wrong ceiling** (`lib/stacks/api-stack.test.ts`, `lambda role policies stay under the IAM size limit`). It only walks `AWS::IAM::Policy`, so inline `Policies` on `AWS::IAM::Role` and any `AWS::IAM::ManagedPolicy` are unmeasured — a role could exceed the quota with this test green. Also worth verifying the numbers: `AWS::IAM::Policy` is an *inline* role policy, whose documented quota is 10,240 characters (managed policies are 6,144), so a 20,480 hard limit and a 16,384 warning would both sit above the real ceiling and could never fire before a deploy rejection. Good idea to add this measurement; please confirm the threshold against the quota the deploy actually enforces.

- **No timeout budget for the nested synchronous invoke.** `call_domain` uses `get_lambda_client().invoke(InvocationType='RequestResponse')` with whatever botocore defaults that shared client carries — typically a 60s read timeout and retries. The MCP function's own timeout is 30s (`lib/stacks/api-stack.ts`), so a slow domain call ends as a Lambda timeout with *no* JSON-RPC response at all, rather than the `-32603` the error taxonomy promises; retries on `RequestResponse` also double-invoke (harmless for these GETs, not for Phase 3 writes). Give this call an explicit `read_timeout` comfortably under the adapter's timeout and pin retry behaviour.

### Nits

- `_tool_search_feedback` computes `"count": len(items)` from the raw route payload but then filters `if isinstance(item, dict)` when projecting, so `count` can exceed `len(items)` in `structuredContent`. Compute the projection first, then count it.
- `lib/stacks/api-stack.ts` hardcodes `'MCPTOKEN'` in the `LeadingKeys` condition. The test reading `MCP_TOKEN_PK` from `mcp_tokens.py` covers the drift, but a shared TS constant would make the lockstep structural rather than test-enforced.
- `get_feedback_detail`'s `outputSchema` has `additionalProperties: false` but no `required` list, unlike the other four tools; the projection always emits every key, so declaring them is free.
- `_handle_autoseed` passes `result.payload` straight to `_cors_response`; when the route answers with an empty body `DomainResult.payload` is `None`, which serializes as `null`. Worth normalizing to `{}`.
- `synthetic_claims` sets `email` to `mcp:<token_id>`, which is fine for display labels but will fail any route that validates email shape — flag it in the docstring as a Phase 3 hazard now that writes are on the roadmap.

---

perrozzi:

Follow-up review — note: the full-context diff was **truncated**, so this delta review is partial (I could not see the tail of `api-stack.ts`, including the final grant/env wiring).

Since my last pass the author added a path-parameter confinement guard (`_PATH_PARAM_PATTERNS` / `_validated_path_parameters`) applied in `_domain_call` for both tools and autoseed, a dedicated delegate Lambda client with an explicit timeout/retry budget, a documented protocol-version skew note, a `sk` lockstep test against the real `projects.get_project`, a rebuilt policy-size guard with the correct IAM quotas and all three policy shapes, and the small nits (count-after-projection, `required` on the detail schema, `{}` for an empty autoseed body, Phase-3 email hazard). Essentially every previous finding is addressed.

### Resolved
- Path parameters are now validated before interpolation, with fail-closed handling for undeclared params and a 400 on the autoseed URL path (`lambda/api/mcp_handler.py`).
- Nested invoke has an explicit 20 s read timeout, 3 s connect, no retries, on a dedicated client (`lambda/shared/mcp_delegate.py`).
- `_document_kind`'s dependency on `sk` is pinned against the real route (`TestDocumentKindLockstep`).
- Policy-size guard now measures `AWS::IAM::Policy`, `AWS::IAM::ManagedPolicy` and inline `Policies` on roles, against 10,240 / 6,144 (`lib/stacks/api-stack.test.ts`).
- Protocol-version/`structuredContent` skew is documented as deliberate and temporary.
- All five nits addressed except the shared TS constant one (below).

### Still open
- `'MCPTOKEN'` remains hardcoded in the `LeadingKeys` condition in `api-stack.ts` (nit; drift is test-covered only).

### New findings
- **`feedback_id`'s pattern is pinned to one writer, not to every writer** (`lambda/api/mcp_handler.py`, `^[0-9a-f]{32}$`). The lockstep test only asserts the pattern accepts `"0" * 32`; it does not prove that every path that writes a `feedback_id` produces 32 lowercase hex. Any ingestion route (feedback form, manual import, a source-supplied id) that stores a differently shaped id makes `get_feedback_detail` return `-32602` for a row `search_feedback` just handed the model — the same search→detail dead end the PR fixed, in a new form. Please derive the pattern from the shared id helper, or add a lockstep over all `feedback_id` writers the way the project-id test reads `projects.py`.
- **Same risk, lower probability, for `project_id`**: `^proj_[A-Za-z0-9_-]+$` assumes every reachable project was minted by the current generator. Legacy/seeded/imported projects with another shape become permanently unreachable via MCP, reported as a malformed argument. A guard that asserts the *formatted* path still matches the declared template shape (single segment, no reserved characters) plus an explicit static-sibling denylist would catch both attacks the comment describes without betting on id provenance.
- **Timeout budget is not in lockstep with the callees** (`lambda/shared/mcp_delegate.py`): 20 s is chosen against the *adapter's* 30 s, but nothing pins the metrics/projects function timeouts below 20 s. If either exceeds it, the adapter reports a 502/`-32603` while the domain function is still running — and with retries disabled there is no second attempt. Worth a CDK test asserting `timeout < _DELEGATE_READ_TIMEOUT_SECONDS` for the two invoke targets.
- Nit (`api-stack.test.ts`): `size()` strips *all* whitespace, including whitespace inside string values (Sids, condition values), which IAM does count; and `policies()` is recomputed three times in the vacuity test. Both cosmetic given the ~2 KB headroom.

---

perrozzi:
Thanks — this found a real one. All five "should fix" items are addressed in e29ea451, four of five nits taken, one declined with reasoning.

## The path-parameter finding is worse than "no privilege gain today"

You were right about the mechanism and generous about the impact. Powertools resolves **static** routes before dynamic ones, so a *single-segment* value — which segment counting would happily pass — lands on a sibling route:

```
get_project({"project_id": "prioritization"})
  -> GET /projects/prioritization
  -> api_get_prioritization_scores      # a STATIC route, checked first
```

Prioritization is deliberately excluded from MCP altogether (a ballot is a human judgment record, and #338 made team the sort key). So the tool that reads one project could reach the one surface the design says must not be reachable. `config` lands on `api_get_config` the same way, and on the feedback side `search` / `urgent` / `entities` are all static siblings of `/feedback/<feedback_id>`. Multi-segment values compound it rather than being the main event.

That is why the guard is a **format check per parameter** rather than the shape assertion you suggested: `proj_` + timestamp, and a 32-character hex digest — what `projects.py` and `processor/handler.py` actually mint. A segment-count or no-slash rule passes `prioritization`. Enforced in `_domain_call`, the single choke point every delegated call goes through (tools and the autoseed route alike), and an *undeclared* parameter fails closed so a future route cannot reopen it silently.

Verified live: `project_id='prioritization'` now returns `-32602` and no prioritization data comes back.

## The policy-size ceiling was above the real limit

Also correct, and worth being blunt about: the guard could not have caught what it was written to catch. The repo's docs say "20 KB"; the enforced quotas are **10,240** characters for an inline role policy (what every `grant*` here produces) and **6,144** for a managed policy, and IAM does not count whitespace. It now measures all three shapes — `AWS::IAM::Policy`, `AWS::IAM::ManagedPolicy`, and inline `Policies` on a role — against their own quotas, whitespace-stripped, warning at 70%.

## The rest

- **Timeout budget** — dedicated Lambda client for delegation, `read_timeout=20`, retries off. The shared client's 60s default could never fire before the adapter's own 30s timeout, so a slow call ended as a timeout with no JSON-RPC envelope instead of `-32603`. Retries also re-run the work, which is the Phase 3 hazard you named.
- **`_document_kind` / `sk`** — good catch on it being the same trap as the `feedback_id` bug. There is now a lockstep test that drives the real `projects.get_project` and asserts `sk` survives, rather than a fixture supplying it.
- **Protocol skew** — documented where `MCP_PROTOCOL_VERSION` lives, including why additive-fields-first is the right order: negotiating first and adding the fields later would mean two breaking output changes instead of one.
- **Nits** — `count` computed from the projection; autoseed normalizes an empty body to `{}`; `get_feedback_detail` declares `required`; the synthetic `email` carries a Phase 3 hazard note.
- **Declined:** a shared TS constant for `MCPTOKEN`. The test reads `MCP_TOKEN_PK` out of the Python source, so the two cannot drift silently, and a cross-language module for one literal used in one place is not worth its weight. Happy to revisit if a second consumer appears.

## One thing your review caused indirectly

The new guard **refused the suite's own fixtures** (`proj-1`, `f1`). Those are now real-shaped ids throughout — which is the same root cause as the `feedback_id` bug in 14624cc2: fixtures that were more convenient than real are how both defects survived a green suite.

Gates: pytest 2933 (was 2911), ruff unchanged with zero new findings, CDK 266, tsc clean, baseline regenerated. Mutations 19/19 killed — and one of the new mutations exposed a missing `count` test, now added. Redeployed and re-verified live: **82/82**, including the malformed-vs-absent split on both the tool and the autoseed route.

perrozzi:

Follow-up review — note: the full-context diff was **truncated**, so this delta review is partial (I could not see the tail of `api-stack.ts` or the rest of the policy-size block).

Since the last pass the path-parameter guard was rewritten: the per-id format allowlist is gone, replaced by a shape rule (`_FORBIDDEN_PATH_CHARS`, no whitespace/control chars, no `.`/`..`) plus `_RESERVED_PATH_SEGMENTS` derived from the owning handlers, with two new lockstep tests; the policy-size measurement dropped the whitespace strip and is now memoized in `beforeAll`; and a documented timeout-budget describe block was added. All four previous findings are addressed, one of them by explicitly rejecting my suggestion with a reason I accept.

### Resolved
- `feedback_id`/`project_id` provenance bet removed — legacy/imported ids are now delegated rather than refused (`lambda/api/mcp_handler.py`, `test_the_guard_does_not_bet_on_where_an_id_came_from`).
- Timeout relationship is now pinned in CDK (`the delegation timeout budget`); the asymmetry is asserted in the direction the callees can actually honour, with the 29 s gateway ceiling as the argument — fine.
- `size()` no longer strips whitespace inside string values; `policies()` measured once.

### Still open
- `'MCPTOKEN'` remains hardcoded in the `LeadingKeys` condition in `api-stack.ts` (nit).

### New findings
- **The reserved-segment lookup fails OPEN for an undeclared prefix** (`lambda/api/mcp_handler.py`, `_reserved_for` → `_RESERVED_PATH_SEGMENTS.get(prefix, frozenset())`). The old guard failed closed on an unrecognised parameter; the new one silently permits sibling collisions for any future templated route whose prefix has no entry. `test_every_reserved_prefix_is_one_a_declared_route_interpolates` only checks `declared ⊆ interpolated`; nothing asserts the reverse. Please require an explicit entry per interpolated prefix (an explicit empty `frozenset()` is fine) and assert `interpolated ⊆ set(_RESERVED_PATH_SEGMENTS)` — otherwise adding a route is again how the hole reopens, which is exactly the property the previous version had.
- **The derivation lockstep can go vacuous** (`test_reserved_segments_cover_every_static_sibling`): `siblings <= declared` passes trivially when the regex matches nothing. The regex covers only `@app.get|post|put|delete` on a literal `app` in a single file per prefix — a route registered with `@app.patch`, `@app.route`, or on a `Router` in another module included into the metrics/projects app would not be found, and the whole security argument now rests on this set being complete. Add a positive control (`assert siblings`, per prefix) and broaden the verb/receiver match.
- Nit: `_validated_path_parameters` raises the same generic "must be a single path segment" message for three distinct conditions; the reserved-segment case is named explicitly but `%`-encoding and whitespace are not distinguishable to a caller. Cosmetic.

---

perrozzi:
Round 2 addressed in fd91cc34. The provenance criticism was right and the guard is redesigned rather than patched.

## Path parameters: shape + reserved segments, no id-format bet

You were right that `^proj_[A-Za-z0-9_-]+$` and `^[0-9a-f]{32}$` bet on where an id came from, and that the cost lands on availability: a seeded, imported, or older-generator project becomes permanently unreachable through MCP, reported as a *malformed argument* rather than a missing project. Replaced with what you suggested:

- **shape rule** — exactly one segment, no `/?#%\`, no whitespace or control characters, not `.`/`..`, and not a parameter the template does not interpolate;
- **reserved-segment set per path prefix** — which is the part a shape rule cannot do, since `prioritization` is a well-formed single segment.

The sets are **derived from the owning handlers' static routes by a lockstep test**, not hand-maintained, so a new static sibling in `projects_handler` or `metrics_handler` fails the suite rather than quietly becoming reachable. A second test asserts the reverse direction — no reserved set for a prefix nothing interpolates — so a stale entry cannot make the first vacuous.

Net versus round 1: same two attacks refused, and an unexpectedly-shaped id is now delegated and 404s honestly. `test_the_guard_does_not_bet_on_where_an_id_came_from` pins that with `legacy-project-42`.

## On the `feedback_id` half specifically

Checked the writers rather than assuming either way, and it is more interesting than "one writer":

- the **stored** field has exactly one writer — the processor's `generate_deterministic_id`, which is `sha256(...)[:32]`, so the hex pattern was accurate;
- but `feedback_form_handler.py:562` mints a `uuid4()` that it returns to the submitter **as `feedback_id`**, while the record it queues carries it as `id` and the processor then computes its own. So there is already a differently-shaped thing called `feedback_id` in the codebase, just not in the table.

That is exactly enough ambiguity to justify not encoding either shape in the adapter, which is where your finding lands even though the specific pattern happened to be correct.

## Timeout lockstep — added in one direction, deliberately

A test now pins `_DELEGATE_READ_TIMEOUT_SECONDS < the adapter's own Lambda timeout`, which is the invariant that actually prevents the failure mode (a Lambda timeout with no JSON-RPC envelope).

I did **not** assert the callees' timeouts below 20 s, and a second test records why so the asymmetry reads as chosen: `metrics` and `projects` serve the browser too, where 30 s is the right budget, and API Gateway caps the outer request at 29 s regardless — so a delegated call that outran the adapter's patience was never going to be delivered to the client anyway. The callee still running when the adapter gives up is inherent to a timeout, and it is why retries are off. Lowering those two functions to satisfy a lockstep would degrade the browser path to make a test symmetric.

## Nits

- `size()` no longer strips whitespace at all — good catch that it could only ever have stripped whitespace *inside* string values, which IAM counts. `JSON.stringify` emits none structurally, so the strip was pure downside.
- The policy walk is memoized in `beforeAll` instead of running three times.
- Still declining the shared TS constant for `MCPTOKEN`, same reasoning as last round.

## Verification

Gates: pytest 2939, ruff zero new findings, CDK 268, tsc clean, baseline regenerated. Mutations **20/20 killed**, including one that empties the reserved set and one that drops the reserved check, so the redesign is held to the same standard the format check was.

Redeployed and re-verified live: **84/84**. The live battery now separates the two cases explicitly — `search`, `urgent`, `entities`, `f1/similar`, `..` and `prioritization` refused before any delegation with no prioritization data in the response; `definitely-not-a-real-id` and `0`×32 delegated and returned as the route's 404.

> the full-context diff was truncated, so this delta review is partial (I could not see the tail of `api-stack.ts`)

For the record, the part you could not see is unchanged since round 1: the `mcpRole` grants (invoke on two exact ARNs, `LeadingKeys`-conditioned projects Query/UpdateItem, KMS) and the `METRICS_FUNCTION` / `PROJECTS_FUNCTION` env wiring. Both are pinned by `mcp Lambda IAM grants` and `mcp delegation stays in step with the stack`.

perrozzi:

Follow-up review — the full-context diff was **truncated** again, so this remains a partial review (I could not see the tail of `api-stack.test.ts` or `api-stack.ts`). Since the last pass the incremental changes are confined to the path-parameter guard and its tests: `_reserved_for` now raises instead of falling back to an empty set, the sibling-derivation regex was broadened (`patch`/`head`/`options`/`route`, any receiver) and hoisted to a shared helper, a positive control was added, and the prefix lockstep now asserts both directions. Both of my previous new findings are addressed as asked.

### Resolved
- Undeclared prefix now fails closed (`_reserved_for` raises) and `test_an_undeclared_prefix_fails_closed` + `interpolated <= declared` pin it (`lambda/api/mcp_handler.py`, `test_mcp_delegation.py`).
- Derivation lockstep can no longer go vacuous: `test_the_sibling_derivation_actually_finds_routes` asserts non-empty per prefix plus two named segments, and the verb/receiver match is broadened.
- Distinct error messages per rejection condition.

### Still open
- `'MCPTOKEN'` remains hardcoded in the `LeadingKeys` condition in `api-stack.ts` (nit; the test reads it from Python, the stack does not).

### New findings
- Nit — a missing reserved-segment declaration is a **server** misconfiguration but is reported as `-32602 Invalid params` (and `400` on the autoseed path), telling the caller its arguments are wrong when nothing it can send will work. `-32603` would be more honest; low severity because `test_every_interpolated_prefix_declares_a_reserved_set` makes the state unreachable at build time.
- Nit — the new "distinct messages" intent is only partly met: a leading/trailing space still falls into the first branch and reports `must be a non-empty identifier`, which is the same misdirection the comment argues against (`_validated_path_parameters`).
- Residual (not introduced here, but the security argument now rests on it): `_static_siblings` still scans only the single owner file per prefix, so a route registered on a `Router` in another module and included into the metrics/projects app is still invisible to the lockstep. Fine if that pattern is not used in this repo; worth a sentence in the docstring if so.

---

perrozzi:
Round 3 addressed in e043e7ea. Both findings were **regressions introduced by round 2's redesign**, which is worth stating plainly rather than filing as new work: swapping the format allowlist for a shape rule lost a property the allowlist had.

## The reserved lookup failed open

`_reserved_for` used `.get(prefix, frozenset())`, so a future templated route whose prefix nobody declared silently permitted sibling collisions — the exact hole the redesign closed, reopened by omission. You identified it as "the property the previous version had", and that is precisely right.

It now **raises** for an unknown prefix, and the companion test asserts **both** directions: `interpolated ⊆ declared` is the security property (a route with no reserved set fails the build), `declared ⊆ interpolated` is the staleness check I already had. Opting out is an explicit `frozenset()` — a line in a diff rather than an absence.

## The derivation lockstep could go vacuous — and already had a gap

Also right, and worse than hypothetical: the regex covered `get|post|put|delete`, while `projects_handler` uses **`@app.patch` twice**. So the set the whole security argument rests on was being derived by a pattern with a live gap. (No hole today — the patch route sits at `/projects/prioritization`, already reserved via its GET — but that is luck, not design.)

Fixed both ways you suggested:

- **positive control** — `test_the_sibling_derivation_actually_finds_routes` asserts routes were found per prefix, plus two specific members (`prioritization`, `search`), so "found something" cannot pass on an unrelated match;
- **broadened match** — every Powertools verb including `patch`, `head`, `options` and `route`, and any decorator receiver rather than a literal `app`.

## Nit

The three shape conditions now report distinct messages. You were right that "must be a single path segment" for a stray space sends a caller looking for a slash they never sent; the forbidden-character message also names the offending character now.

## Verification

Mutations **22/22 killed**, including two new ones that restore each regression exactly — the `.get(prefix, frozenset())` fallback, and a narrowed derivation regex that empties the sibling set. That second mutation is the one I would have wanted on this PR from the start: it fails the positive control rather than passing a vacuous subset test.

Gates: pytest 2941, ruff zero new findings, CDK 268, tsc clean, baseline regenerated. Redeployed and re-verified live: **84/84**.

perrozzi:

Follow-up review — the full-context diff was **truncated** again (I could not see the tail of `api-stack.test.ts`/`api-stack.ts`), so this remains a partial review. The incremental change is small and targeted: `_reserved_for` now raises `DelegationUnavailable` (→ `-32603`) with a `logger.error`, the whitespace rejection was split out with its own message, `_static_siblings` gained a docstring about its single-file scope, and a new `test_no_route_is_registered_outside_the_scanned_file` plus an end-to-end `-32603` test were added. All three of my previous nits are addressed.

### Resolved
- Server misconfiguration is now reported as `-32603` rather than `-32602`, pinned both at the helper and end-to-end (`mcp_handler.py`, `test_mcp_delegation.py`).
- Leading/trailing whitespace now reports `must not be surrounded by whitespace` instead of the misleading non-empty message (`_validated_path_parameters`).
- The single-file `_static_siblings` scan is now documented *and* asserted (`Router` / `include_router` absent in both owner handlers).

### Still open
- `'MCPTOKEN'` is still hardcoded in the `LeadingKeys` condition in `api-stack.ts` (nit).

### New findings
- Nit — `_handle_autoseed` (`mcp_handler.py`) still only catches `InvalidToolArgument` around `_domain_call`; now that `_reserved_for` raises `DelegationUnavailable`, an undeclared prefix on that path escapes both `try` blocks and falls to whatever the outer `lambda_handler` catch-all answers, rather than the 502 the adjacent `DelegationUnavailable` branch establishes. Unreachable today (the `/projects` prefix is declared and the build lockstep enforces it), and untested — worth either widening the first `except` or noting why not.
- Nit — the new `_static_siblings` docstring says "That is not documented and left to trust" immediately before naming the test that asserts it; stale sentence from the previous iteration, plus the test name is line-wrapped mid-identifier.

Otherwise the delta is clean.

---

perrozzi:
Round 4 addressed in d5e40c6f. All three nits taken.

**The -32602 misclassification was the substantive one.** A missing reserved-segment declaration is a server misconfiguration, and reporting it as `Invalid params` told the caller its arguments were wrong when nothing it could send would work. It now raises `DelegationUnavailable` → `-32603` (and 502 on the autoseed path) and logs the prefix. Two tests pin it, including one asserting end-to-end that the caller sees `-32603` and never `-32602`. You are right that the state is unreachable in a deployed build; this is about not lying if it ever is.

**The "distinct messages" change was half done** — a leading/trailing space still fell into the emptiness branch and reported "must be a non-empty identifier" for a value that plainly was not empty, which is the exact misdirection the change was meant to remove. Split out.

**The Router blind spot is now asserted rather than documented.** You offered "fine if that pattern is not used, worth a sentence if so" — I took the third option: `_static_siblings`' docstring names the limitation, and a test fails if either owning handler starts referencing `Router` or `include_router`. "We don't use Routers" is exactly the kind of claim that stops being true without anyone revisiting the code that depends on it, and the reserved-segment argument depends on it.

On the remaining open nit — `'MCPTOKEN'` in the `LeadingKeys` condition — still declining, same reasoning: the test reads `MCP_TOKEN_PK` out of `mcp_tokens.py`, so the two cannot drift silently, and a cross-language constant module for one literal in one place is not worth its weight. Worth revisiting when a second consumer appears.

Gates: pytest 2943, ruff zero new findings, CDK 268, tsc clean, baseline regenerated, mutations 22/22 killed. Redeployed and re-verified live: **84/84**.

Four rounds in, the findings have moved from a real security bug to error-message ergonomics, which reads like convergence. I'll merge unless the next round turns up something substantive.

perrozzi:

Follow-up review — the full-context diff was **truncated** again (the tail of `api-stack.test.ts` is cut off), so this remains a partial review. The delta is two things: `_handle_autoseed` now wraps `_domain_call` and `call_domain` in a single `try` so a `DelegationUnavailable` from `_reserved_for` answers 502 like the adjacent delegation failure, with a new test pinning it; and the `_static_siblings` docstring's stale "left to trust" sentence and mid-identifier line wrap are fixed. Both of my previous nits are addressed.

### Resolved
- `_handle_autoseed` no longer lets a missing reserved-segment declaration escape to the outer catch-all; the 502 is asserted end-to-end in `test_autoseed_reports_a_missing_declaration_as_a_server_fault` (`mcp_handler.py`, `test_mcp_delegation.py`).
- The `_static_siblings` docstring now points at the asserting test instead of claiming the property is untested, and no longer wraps an identifier (`test_mcp_delegation.py`).

### Still open
- `'MCPTOKEN'` is still hardcoded in the `LeadingKeys` condition in `api-stack.ts` (nit).

No new findings — the merged `try` widens the caught set without changing any reachable behaviour (`call_domain` cannot raise `InvalidToolArgument`), and the new test exercises the real handler entry point rather than the helper. Clean delta.

---

perrozzi:
Round 5 addressed in 633fe52f. Both nits taken.

The autoseed one was a genuine inconsistency I introduced last round: making `_reserved_for` raise `DelegationUnavailable` gave `_domain_call` two failure kinds, and the autoseed path still caught only one. Both steps now sit in a single `try` with both handlers, and a test drives the 502 path so the tool path and the autoseed path cannot diverge again.

The stale docstring sentence was an artefact of adding the test after writing the comment — fixed, and thank you for reading them as carefully as the code.

One thing worth recording from this round: the mutation harness caught that restructuring the `try` block had **orphaned** the "autoseed reads project rows in-process" mutation — its anchor no longer existed, so it had been silently passing rather than testing anything. Retargeted, and 22/22 kill again. That is the second time this PR that a mutation exposed a hole in its own scaffolding, which is a decent argument for running them on every round rather than once.

Gates: pytest 2944, ruff zero new findings, CDK 268, tsc clean, baseline regenerated. Redeployed and re-verified live: **84/84**.

Merging this now — five rounds, and the last two produced only message ergonomics and an unreachable-path consistency fix. Phase 2b (protocol conformance: `server/discover`, the version range, `resultType`, header validation, retiring `ANY /mcp/{proxy+}`, annotations, scope-filtered `tools/list`) follows as its own PR.

PR #356, opened 2026-08-20
fix(personas): make list_personas callable and fix the import prompt

perrozzi:
## Summary

`list_personas` could not be called at all, and tracing why led to the writer that caused it.

The tool declared `pain_points` and `behaviors` as `array<string>` while every stored persona holds objects. Clients validate `structuredContent` against the schema a tool publishes — the official MCP SDK does so by default — so **every call failed with `-32602`** and the model received nothing. Raw JSON-RPC returned the data, which is why 2 944 pytest cases, 268 CDK cases and an 84/84 live HTTP battery all passed: the schema was published as a contract and never checked against data that exists.

Found by connecting a real MCP client to the deployed server for the first time. Full write-up of this and seven sibling findings is in `analysis/BUG-mcp-tool-truthfulness-GITHUB-ISSUE.md` (M1 here; M2–M8 not addressed).

## Why the shapes disagreed

`persona_importer` built its prompt inline with a schema string whose every section was the literal `{...}`:

```
'{"identity": {...}, "goals_motivations": {...}, "pain_points": {...}, "behaviors": {...}, …}'
```

The model was told the section **names** and nothing about their **contents**, and it complied — imported personas carry `primary_frustration`, `frustration`, `tooling`, `current_practices` where generated ones carry the canonical keys. So the imported rows are not malformed; they are the correct response to a prompt that asked for nothing in particular.

Meanwhile `api/prompts/persona-import.json` has carried the full canonical key set — with example values that pin types and enums — the whole time, and **nothing loaded it**. `shared/prompts.py` declares constants for five templates and had none for it.

## Changes

**Reader** — `lambda/api/mcp_handler.py`

- `_PERSONA_SECTIONS` + `_QUOTE_PROPERTIES` mirror `schemas/persona.schema.json`. All seven content sections are reported: identity, goals & motivations, pain points, behaviors, context & environment, quotes, scenario. Section 8, `research_notes`, is researcher annotation rather than persona content and is the one deliberate exclusion.
- Six declared keys that exist on no stored row and reported empty forever are gone: `goals`, `age_range`, `occupation`, `quote`, `journey_stage`, `type`.
- `get_project`'s persona summary reports `tagline` instead of `type`.
- **Every field is coerced to the type its own schema declares**, derived from the declarations (`_declared_types` → `_COERCIONS`) rather than written out field by field. `_project_persona` used to read each field with `item.get(key, default)`, which fires only on an absent key and cannot correct a value of the wrong type — the mechanism that laundered the mismatch. Deriving the coercion is the other half of that lesson: a hand-kept list of "which keys are arrays" left `confidence` unchecked beside a defensively `int()`-wrapped `feedback_count`.
- A scalar in a declared array slot is wrapped; a list in a declared string slot is joined; a dict is JSON rather than a Python `repr`. `workarounds` is a string on one live row and a list on another.
- `get_project` coerces the declared strings in its own projection too. It is the tool that **was** callable, so it is the live half of the same defect, and it reads `tagline` from the writer that pinned nothing.
- Sections keep `additionalProperties: true` and unrecognised keys are preserved. Declaring `false` would make the tool fail on its own product; dropping the keys would answer *"this persona has no pain points"* about a persona whose pain points sit under an unforeseen key. A section that arrives as something other than an object is logged rather than discarded in silence — without its content, which is customer-derived text.
- `MCP_SERVER_VERSION` → **3.0.0**. The constant's own rule is that a changed tool output shape is a major bump, and it is what a client reads from `serverInfo`. Nothing could have consumed the removed keys — they were always empty and the tool was uncallable — but a client coded against the names loses them.

**Writer** — `lambda/jobs/persona_importer/handler.py`, `lambda/shared/prompts.py`

- Loads the template via `load_prompt_file` behind a new `PERSONA_IMPORT_PROMPTS` constant, taking `system_prompt`, `output_schema` (JSON-dumped, so what the model sees cannot drift from what the file declares), `user_prompts[input_type]` and `max_tokens` from it.
- Records `llm_metadata.prompt_version`; imported personas carried none, so a row with odd keys could not be attributed to the prompt that produced it.
- The template's `pdf` prompt is deliberately **not** wired — `pdf` is in `DEFERRED_INPUT_TYPES` and refused upstream, so a branch would be unreachable code advertising a declined capability.

**No CDK or bundling change.** `createJobLambdaCode` already copies `api/prompts` to the bundle root and `get_prompts_dir()` resolves `/var/task/prompts` first, so the template was already deployed beside the code ignoring it. Verified before wiring, because `load_prompt_file` would otherwise have raised `FileNotFoundError` at runtime and no test would have caught it — tests resolve the repo layout.

## What this does not do

Fixing the prompt does **not** make `additionalProperties: true` removable, and that is design rather than a leftover. A prompt is a request, not schema enforcement, so the model can still deviate; and discarding a field genuinely extracted from a customer's own document (`industry`, `role`, `segment` off a persona card) is data loss rather than tidiness. What changes is that variance goes from **systematic** — guaranteed, because the prompt asked for `{...}` — to **incidental**.

Write-side validation of the parsed output was considered and left out: the template's example values pin the types, so type variance is addressed at source, and a second normalizer before measuring whether variance persists would be speculative. If it recurs, the enforcement point is `persona_importer` before `put_item`.

## Tests

- Replaced the fixture that let this through — flat string lists and a `quote` string, a shape no writer has ever produced.
- New fixtures copy the shapes of **live rows**: generated, imported, sparse.
- **A schema-conformance gate that did not exist anywhere in this repo**: it reads each tool's own `outputSchema` from `MCP_TOOLS` so it cannot drift, and it has a positive control, because a conformance test that cannot fail proves nothing.
- **The completeness claim is enforced, not asserted.** A docstring saying "the canonical 8-section shape" is what let two persisted sections go unreported. The check now reads `schemas/persona.schema.json` and keys off the `Section N:` marker the schema itself writes, so a ninth section fails CI until it is either reported or named in an exclusion set — and the exclusion set is checked for being a dumping ground.
- **Every declared type must have a coercion**, and every array declared inside a section must declare string items, since the array coercion would otherwise flatten object items. Both guards have their own mutation as a control.
- **The declared output shapes are pinned to `MCP_SERVER_VERSION`.** The previous guard (`major >= 2`) passes forever and would have allowed this PR's own shape change to ship under the version describing the previous one.
- Prompt-provenance guards: the prompt names the canonical inner keys of every section the reader reports, shows types rather than key names alone, interpolates `{content}`, leaves the `pdf` prompt unwired (compared against the template's own text, not a phrase copied out of it), and pins the persisted version to the template's. One further test asserts the template still carries every key the handler subscripts, so a dropped key fails CI rather than reaching a user as a `KeyError` through `job_handler`'s user-facing error text.

## Validation

| Gate | Result |
|---|---|
| pytest | **2967** passed (from 2952) |
| CDK vitest | **268** passed |
| ruff | **delta zero** on every file touched — same findings, same rule histogram |
| mutations | **15 run, 15 caught** on the reader; **9 run, 8 caught** on the writer |
| real data | the projection run against the **five live persona rows** conforms, empties no populated section, emits no phantom field |

The one writer mutation **not** caught is recorded rather than hidden: reverting `max_tokens` to the literal `4096` equals the template's current value, so it is behaviour-neutral and no test can distinguish it.

Worth noting for reviewers who mutation-test: reverting the *projection* alone does **not** trip the conformance gate, because with the schema corrected a dict is now valid. The mutation that proves that gate earns its keep is reverting the **schema**. Testing one side only would have claimed coverage that did not exist.

## Not deployed

The shared slot is contended and this is unmerged. `list_personas` still fails against the live server until this ships.

## Follow-ons, deliberately not bundled

- **`_project_feedback` has the same `.get(key, '')`-under-a-declared-string pattern** for `source`, `sentiment`, `category`, `urgency`, `persona_type` and the text fields. Same class, different writer, its own tests, shipped in #355 and untouched by this diff — and no live counter-example is known, where the persona case had five rows in four shapes. Folding it in would put a second contract into a PR about the persona one.
- **Three existing persona rows stay in their non-canonical shape by decision**, not by omission: the reader handles both shapes, and this PR fixes the import path rather than the data. No stored data is touched.
- **Seven prompt builders read phantom keys** (`p.get('goals')`, `p.get('frustrations')`, singular `quote`) in `jobs/document_generator`, `jobs/document_merger`, `research/research_step_handler`, `projects.py` autofill and its two legacy builders, plus `project-context.ts` and `persona-prompt.ts`. Every generated PRD, PR/FAQ, merged document, research report and roundtable prompt is therefore built with blank persona goals and frustrations. Arguably higher product impact than this PR, and it changes LLM inputs, so it is its own change — in flight on `fix/personas-reach-prompts`, which shares no file with this branch.

### DIFF
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -683,17 +683,104 @@ _DOCUMENT_KINDS: dict[str, str] = {
     'PROTOTYPE#': 'prototype',
 }
 
+_STRING_LIST: dict[str, Any] = {"type": "array", "items": {"type": "string"}}
+
+# The persona shape, mirroring `schemas/persona.schema.json` — which is the
+# repo's canonical declaration, what both writers persist, and what the frontend
+# `ProjectPersona` type already describes.
+#
+# 🔴 It did NOT mirror it before, and that is what made `list_personas`
+# uncallable: it declared a pre-8-section shape (`goals`, `age_range`,
+# `occupation`, `quote`, `journey_stage`, `type` at the top level, with
+# `pain_points`/`behaviors` as `array<string>`). Six of those ten keys exist on
+# no stored row, and the two that do are OBJECTS. A client validates
+# `structuredContent` against this schema, so every call failed `-32602` while
+# raw JSON-RPC returned the data — which is why a curl battery never saw it.
+#
+# `additionalProperties` is deliberately OPEN on each section. These values are
+# LLM-authored: the generation prompt pins the canonical keys, but the import
+# path does not, and live rows carry `primary_frustration`, `frustration`,
+# `tooling`, `current_practices`, `related_issues`. Declaring `false` would make
+# the tool fail on its own product; declaring the known keys and permitting the
+# rest is the honest contract. The variance belongs to the WRITER and is fixed
+# there, not laundered here.
+_PERSONA_SECTIONS: dict[str, dict[str, Any]] = {
+    # Section 1 — Identity & Demographics.
+    "identity": {
+        "age_range": {"type": "string"},
+        "location": {"type": "string"},
+        "occupation": {"type": "string"},
+        "income_bracket": {"type": "string"},
+        "education": {"type": "string"},
+        "family_status": {"type": "string"},
+        "bio": {"type": "string"},
+    },
+    # Section 2 — Goals & Motivations.
+    "goals_motivations": {
+        "primary_goal": {"type": "string"},
+        "secondary_goals": _STRING_LIST,
+        "success_definition": {"type": "string"},
+        "underlying_motivations": _STRING_LIST,
+    },
+    # Section 3 — Pain Points & Frustrations.
+    "pain_points": {
+        "current_challenges": _STRING_LIST,
+        "blockers": _STRING_LIST,
+        "workarounds": _STRING_LIST,
+        "emotional_impact": {"type": "string"},
+    },
+    # Section 4 — Behaviors & Habits.
+    "behaviors": {
+        "current_solutions": _STRING_LIST,
+        "tools_used": _STRING_LIST,
+        "activity_frequency": {"type": "string"},
+        "tech_savviness": {"type": "string"},
+        "decision_style": {"type": "string"},
+    },
+}
+
 _PERSONA_PROPERTIES: dict[str, Any] = {
     "persona_id": {"type": "string"},
     "name": {"type": "string"},
-    "type": {"type": "string"},
-    "age_range": {"type": "string"},
-    "occupation": {"type": "string"},
-    "goals": {"type": "array", "items": {"type": "string"}},
-    "pain_points": {"type": "array", "items": {"type": "string"}},
-    "behaviors": {"type": "array", "items": {"type": "string"}},
-    "quote": {"type": "string"},
-    "journey_stage": {"type": "string"},
+    # `tagline` is the persona's one-line characterisation and is REQUIRED by the
+    # canonical schema. It replaces the old `type`, which no row has ever carried.
+    "tagline": {"type": "string"},
+    "confidence": {"type": "string"},
+    "feedback_count": {"type": "integer"},
+    **{
+        section: {
+            "type": "object",
+            "properties": properties,
+            "additionalProperties": True,
+        }
+        for section, properties in _PERSONA_SECTIONS.items()
+    },
+    # Section 6 — Representative Quotes. Objects on the row, not strings, and the
+    # old single `quote` key never existed.
+    "quotes": {
+        "type": "array",
+        "items": {
+            "type": "object",
+            "properties": {
+                "text": {"type": "string"},
+                "context": {"type": "string"},
+            },
+            "additionalProperties": True,
+        },
+    },
+}
+
+# Which declared keys are arrays, so the projection can coerce a scalar the
+# writer left unwrapped. Derived from the declarations above by VALUE, not by
+# identity with `_STRING_LIST`: an author who inlines the literal instead of
+# reusing the constant must still get coercion, or this derivation would have
+# exactly the silent-omission hole it exists to close.
+_PERSONA_LIST_KEYS: dict[str, frozenset[str]] = {
+    section: frozenset(
+        key for key, declared in properties.items()
+        if declared.get("type") == "array"
+    )
+    for section, properties in _PERSONA_SECTIONS.items()
 }
 
 # A window argument, stated once. `maximum` is the route's real ceiling
@@ -869,7 +956,10 @@ MCP_TOOLS = [
                         "properties": {
                             "persona_id": {"type": "string"},
                             "name": {"type": "string"},
-                            "type": {"type": "string"},
+                            # `tagline`, not `type`: no stored persona has ever
+                            # carried a `type`, so this summary reported an empty
+                            # string for every persona in every project.
+                            "tagline": {"type": "string"},
                         },
                         "additionalProperties": False,
                     },
@@ -1124,26 +1214,89 @@ def _document_kind(item: dict) -> str:
     return ''
 
 
-def _project_persona(item: dict) -> dict:
-    """The persona fields the tools report, with typed defaults.
+def _as_string_list(value: Any) -> list[str]:
+    """Coerce a declared-array persona field to the list of strings it promises.
 
-    Shared by both project-shaped tools so a persona reads the same either way.
-    Everything else on the row — avatar keys, source feedback ids, notes,
-    timestamps — is dropped: a project's personas are the answer, not its
-    storage layout.
+    A writer that leaves a single value unwrapped (`workarounds` is a string on
+    some imported rows and a list on generated ones) must not make the payload
+    contradict its own schema, so the boundary coerces instead of passing the
+    scalar through. Non-string entries are stringified rather than dropped: the
+    value is a model's evidence, and silently losing it is worse than reporting
+    it in the declared type.
     """
+    if value is None or value == '':
+        return []
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, (list, tuple)):
+        return [v if isinstance(v, str) else str(v) for v in value if v not in (None, '')]
+    return [str(value)]
+
+
+def _persona_section(value: Any, list_keys: frozenset[str]) -> dict:
+    """One canonical persona section, coerced to its declared types.
+
+    Unrecognised keys are PRESERVED, not dropped. The section's declared keys
+    are what the generation prompt pins; the import path pins nothing, so real
+    rows also carry `primary_frustration`, `tooling`, `related_issues`. Dropping
+    those would make the tool answer "this persona has no pain points" about a
+    persona whose pain points are simply under a key this file did not predict —
+    the same class of silent under-report the surface is being fixed for.
+    """
+    if not isinstance(value, dict):
+        # Absent, or a shape no writer produces. Every writer persists an object
+        # (both use a `{}` default), and all five live rows are objects, so there
+        # is no flat-list case to salvage — and guessing a destination key would
+        # file content under a misleading heading (`sorted()` would pick
+        # `blockers` for a list of pain points).
+        return {}
     return {
+        key: _as_string_list(inner) if key in list_keys else inner
+        for key, inner in value.items()
+    }
+
+
+def _project_persona(item: dict) -> dict:
+    """One persona in the canonical shape of `schemas/persona.schema.json`.
+
+    Used ONLY by `list_personas`. `get_project` deliberately renders a two-field
+    summary of its own — an earlier version of this docstring claimed the two
+    shared a projection, which was never true and is why the schema mismatch
+    below went unnoticed for so long.
+
+    🔑 The bug this replaces: every field was read with `item.get(key, default)`,
+    which fires only when a key is ABSENT and cannot correct a value of the wrong
+    TYPE. `pain_points` and `behaviors` are objects on every stored row while the
+    schema declared `array<string>`, so the dicts travelled unchanged and a
+    schema-validating client rejected the whole result. Six other declared keys
+    (`goals`, `age_range`, `occupation`, `quote`, `journey_stage`, `type`) exist
+    on no row at all and were reported as empty forever.
+
+    Everything not declared — avatar keys, source feedback ids, research notes,
+    timestamps, llm metadata — is still dropped: a project's personas are the
+    answer, not its storage layout.
+    """
+    projected: dict[str, Any] = {
         "persona_id": item.get('persona_id', ''),
         "name": item.get('name', ''),
-        "type": item.get('type', ''),
-        "age_range": item.get('age_range', ''),
-        "occupation": item.get('occupation', ''),
-        "goals": item.get('goals', []),
-        "pain_points": item.get('pain_points', []),
-        "behaviors": item.get('behaviors', []),
-        "quote": item.get('quote', ''),
-        "journey_stage": item.get('journey_stage', ''),
+        "tagline": item.get('tagline', ''),
+        "confidence": item.get('confidence', ''),
     }
+
+    # `feedback_count` is declared an integer, so a missing or unparseable value
+    # must not become `''` — that would violate the schema exactly as the old
+    # string defaults did.
+    try:
+        projected["feedback_count"] = int(item.get('feedback_count') or 0)
+    except (TypeError, ValueError):
+        projected["feedback_count"] = 0
+
+    for section, list_keys in _PERSONA_LIST_KEYS.items():
+        projected[section] = _persona_section(item.get(section), list_keys)
+
+    quotes = item.get('quotes')
+    projected["quotes"] = [q for q in quotes if isinstance(q, dict)] if isinstance(quotes, list) else []
+    return projected
 
 
 def _get_project_payload(token_info: dict) -> tuple[dict, list[dict], list[dict]]:
@@ -1190,7 +1343,7 @@ def _tool_get_project(args: dict, token_info: dict) -> ToolResult:
         "document_count": len(documents),
         "personas": [
             {"persona_id": p.get('persona_id', ''), "name": p.get('name', ''),
-             "type": p.get('type', '')}
+             "tagline": p.get('tagline', '')}
             for p in personas
         ],
         "documents": [
--- a/voc-datalake/lambda/api/mcp_handler.py
+++ b/voc-datalake/lambda/api/mcp_handler.py
@@ -12,6 +12,7 @@ from the Authorization header against SHA-256 hashes in the projects table.
 import json
 import os
 import re
+from collections.abc import Callable
 from dataclasses import dataclass
 from datetime import datetime, timezone
 from typing import Any
@@ -75,12 +76,23 @@ projects_table = get_projects_table()
 # meant two breaking output changes instead of one.
 MCP_PROTOCOL_VERSION = "2024-11-05"
 
-# Semver on the SERVER, independent of the protocol revision above. 2.0.0
-# because tool output shapes changed: results now carry `structuredContent`
-# alongside the text block, sad paths that used to arrive as prose in a
-# successful result are now tool errors, and `get_project` reports the
-# documents it previously dropped.
-MCP_SERVER_VERSION = "2.0.0"
+# Semver on the SERVER, independent of the protocol revision above, and the only
+# signal a client gets that a tool's output shape moved — it is advertised in
+# `serverInfo` on every `initialize`.
+#
+# 2.0.0 carried `structuredContent` alongside the text block, turned sad paths
+# that used to arrive as prose in a successful result into tool errors, and made
+# `get_project` report the documents it had been dropping.
+#
+# 3.0.0 changes the persona shape both persona-reporting tools publish:
+# `list_personas` now mirrors `schemas/persona.schema.json` and no longer
+# declares `goals`, `age_range`, `occupation`, `quote`, `journey_stage` or
+# `type`, and `get_project`'s persona summary reports `tagline` in place of
+# `type`. Those six keys existed on no stored row and `list_personas` was
+# uncallable against real data, so nothing could have consumed them — but a
+# client coded against the NAMES loses them, which is a major bump by this
+# file's own rule rather than a judgement about how many clients exist.
+MCP_SERVER_VERSION = "3.0.0"
 
 
 # ============================================
@@ -685,25 +697,22 @@ _DOCUMENT_KINDS: dict[str, str] = {
 
 _STRING_LIST: dict[str, Any] = {"type": "array", "items": {"type": "string"}}
 
-# The persona shape, mirroring `schemas/persona.schema.json` — which is the
-# repo's canonical declaration, what both writers persist, and what the frontend
-# `ProjectPersona` type already describes.
-#
-# 🔴 It did NOT mirror it before, and that is what made `list_personas`
-# uncallable: it declared a pre-8-section shape (`goals`, `age_range`,
-# `occupation`, `quote`, `journey_stage`, `type` at the top level, with
-# `pain_points`/`behaviors` as `array<string>`). Six of those ten keys exist on
-# no stored row, and the two that do are OBJECTS. A client validates
-# `structuredContent` against this schema, so every call failed `-32602` while
-# raw JSON-RPC returned the data — which is why a curl battery never saw it.
+# The persona, mirroring `schemas/persona.schema.json` — the repo's canonical
+# declaration, what both writers persist, and what the frontend
+# `ProjectPersona` type already describes. Section numbers below are the
+# canonical schema's own: 1-5 and 7 are objects and are declared here, 6
+# (`quotes`) is an array declared under `_QUOTE_PROPERTIES`, and 8
+# (`research_notes`) is researcher annotation rather than persona content and is
+# deliberately not reported. That division is enforced against the schema file
+# by `test_every_canonical_persona_section_is_reported_or_excluded`, so a ninth
+# section cannot go missing here the way 5 and 7 first did.
 #
 # `additionalProperties` is deliberately OPEN on each section. These values are
-# LLM-authored: the generation prompt pins the canonical keys, but the import
-# path does not, and live rows carry `primary_frustration`, `frustration`,
-# `tooling`, `current_practices`, `related_issues`. Declaring `false` would make
-# the tool fail on its own product; declaring the known keys and permitting the
-# rest is the honest contract. The variance belongs to the WRITER and is fixed
-# there, not laundered here.
+# LLM-authored, and a prompt is a request rather than enforcement: live rows
+# carry `primary_frustration`, `frustration`, `tooling`, `current_practices`,
+# `related_issues`. Declaring `false` would make the tool fail on its own
+# product; declaring the known keys and permitting the rest is the honest
+# contract, and the variance belongs to the writer.
 _PERSONA_SECTIONS: dict[str, dict[str, Any]] = {
     # Section 1 — Identity & Demographics.
     "identity": {
@@ -737,6 +746,35 @@ _PERSONA_SECTIONS: dict[str, dict[str, Any]] = {
         "tech_savviness": {"type": "string"},
         "decision_style": {"type": "string"},
     },
+    # Section 5 — Context & Environment.
+    "context_environment": {
+        "usage_context": {"type": "string"},
+        "devices": _STRING_LIST,
+        "time_constraints": {"type": "string"},
+        "social_context": {"type": "string"},
+        "influencers": _STRING_LIST,
+    },
+    # Section 7 — Scenario / User Story.
+    "scenario": {
+        "title": {"type": "string"},
+        "narrative": {"type": "string"},
+        "trigger": {"type": "string"},
+        "outcome": {"type": "string"},
+    },
+}
+
+# Section 6 — Representative Quotes. Objects on the row, not strings, and the
+# old single `quote` key never existed. Named separately from the sections above
+# because it is an array, so it needs its own projection.
+#
+# `source_feedback_id` is the quote's own citation and is declared, unlike the
+# persona's top-level `source_feedback_ids`, which is dropped: one is where this
+# sentence came from and is the id `get_feedback_detail` takes, the other is the
+# row's provenance list.
+_QUOTE_PROPERTIES: dict[str, Any] = {
+    "text": {"type": "string"},
+    "context": {"type": "string"},
+    "source_feedback_id": {"type": "string"},
 }
 
 _PERSONA_PROPERTIES: dict[str, Any] = {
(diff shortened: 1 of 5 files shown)


perrozzi:

This PR fixes a real, well-diagnosed defect: `list_personas`' `outputSchema` in `mcp_handler.py` declared a pre-8-section persona shape (`pain_points`/`behaviors` as `array<string>`, plus six keys no row carries), so schema-validating MCP clients rejected every response. It rewrites the schema to mirror `schemas/persona.schema.json`, rewrites `_project_persona` to project structurally, and fixes the upstream writer by loading `api/prompts/persona-import.json` via the new `PERSONA_IMPORT_PROMPTS` constant instead of the inline `{...}` placeholder schema. The fix is in the right places, the schema-conformance gate with a positive control is a genuinely good addition, and the regression tests do fail if the schema or projection is reverted. My concerns are all about the completeness of the new contract rather than its correctness. Note the code comments are extremely verbose narrative — see nits.

### Should fix

- **Two stored canonical sections are silently unreachable.** `persona_importer/handler.py` persists `context_environment` and `scenario` (the latter is visible in the diff's persist block), and the generation path does too, but `_PERSONA_PROPERTIES` in `voc-datalake/lambda/api/mcp_handler.py` declares only `identity`, `goals_motivations`, `pain_points`, `behaviors`, `quotes`. Because the persona object is `additionalProperties: false` (proved by `test_the_schema_checker_itself_rejects_a_bad_payload`'s `journey_stage` case), a client can never see those two sections, while the test asserting this is named `test_personas_are_listed_in_full_by_the_persona_tool` and the docstring claims "the canonical 8-section shape". Either declare/project them or rename the tool's contract so it doesn't claim completeness it doesn't deliver.

- **`quotes` entries that are strings are dropped entirely.** `projected["quotes"] = [q for q in quotes if isinstance(q, dict)]` in `mcp_handler.py`. Since these values are LLM-authored on a path that (until this PR) pinned nothing, `quotes: ["I cannot prove what changed."]` is a plausible live shape, and this filter answers "this persona has no quotes". That is the exact silent under-report the rest of the change argues against; coerce a string entry to `{"text": q}` the way `_as_string_list` coerces scalars.

- **Only array-typed fields are type-coerced; declared string/integer fields still pass through unchecked.** `_persona_section` coerces `key in list_keys` and returns everything else verbatim, so a row with `emotional_impact: ["a", "b"]` or `primary_goal: {...}` reproduces M1 precisely — a payload contradicting its own `outputSchema`. Same for top-level `confidence`, which is taken as `item.get('confidence', '')` with no coercion while `feedback_count` is defensively `int()`-wrapped. Given the argument that the prompt is "a request, not schema enforcement", the boundary should coerce declared scalars too (or the sections should stop declaring inner scalar types).

### Nits

- `_as_string_list` on a dict produces `[str(value)]`, i.e. a Python `repr` (`"{'a': 'x'}"`) shipped to an LLM client. `json.dumps` would at least be machine-readable.
- `_persona_section` returns `{}` for a non-dict section. The justification ("all five live rows are objects") holds for today's table but silently discards any older flat-list row; a `logger.warning` (no persona content) would make that visible instead of invisible.
- `test_the_pdf_prompt_is_not_wired` asserts `'PDF document' not in prompt`, coupling the test to the template's exact wording. Asserting `user_prompts['pdf'] not in prompt_text` would survive a prompt reword.
- The production comments in `mcp_handler.py` and `persona_importer/handler.py` narrate PR history with emoji (`🔴 ROOT-CAUSE FIX`, `🔑 The bug this replaces`, mutation-testing notes). That belongs in the PR body and the linked write-up; in-file it will read as stale in six months. A two-line "why `additionalProperties` is open" note is the part worth keeping.
- Consider whether the two removed fields on `get_project`'s persona summary (`type` → `tagline`) plus the six dropped `list_personas` fields warrant a note in the MCP tool changelog/docs, since it is a breaking schema change for any client already coded against the old names — even though they were always empty.

---

perrozzi:
## Round 2 — disposition of the review on `3310f355`

All three `should_fix` findings are valid and all three are taken. They are one class — *the tool declares something the payload does not honour* — which is the class this PR exists to fix, so the review caught me applying the argument to the old code and not to the new. Two more instances of the same class turned up when I re-audited the fix for it rather than waiting to be told, and they are in here too.

### Should fix

**1. `context_environment` and `scenario` were unreachable while the contract claimed completeness — fixed, and the claim is now enforced rather than asserted.**

Both are persisted by both writers (`persona_importer` and the generation path) and `api/prompts/persona-import.json` asks the model for both with their full inner keys. They are now declared and projected, so `list_personas` reports seven of the schema's eight sections.

The docstring/test-name problem needed more than a rename: prose cannot fail. `test_every_canonical_persona_section_is_reported_or_excluded` reads `schemas/persona.schema.json` and keys off the `Section N:` marker the schema itself writes in each description, so a ninth section added there fails CI until it is either reported or named in an explicit exclusion set. The one exclusion is `research_notes` — researcher annotation added through the notes routes after the fact, a human's working notes *about* the persona rather than the persona, and already asserted absent by `test_storage_layout_is_not_exposed`. The test also checks the exclusion set is not a dumping ground: every name in it must be a real section and must really not be declared. `len(sections) == 8` guards the marker itself, so a changed description cannot make the check vacuously true.

The tool's `description` now names the sections it reports and says researcher notes are not included, since that string is what a model reads to decide whether to call it.

*(One correction to the finding's mechanism, which does not change its conclusion: `additionalProperties: false` on the persona object was not why those sections were unreachable — the projection simply never emitted them. The closed top level is why an undeclared key could not have leaked through either.)*

**2. `quotes` string entries — fixed, coerced rather than filtered.**

`{"text": q}`, exactly as suggested. The shape is not hypothetical: the import path's own test fixture models the model returning `quotes: ["I spend too much time in meetings"]`, and until this PR nothing pinned the schema it was asked for. An entry with no content still yields nothing, and that distinction is tested — reporting `{"text": ""}` would invent a quote.

**3. Declared scalars uncoerced — fixed at the mechanism, not the instances.**

Coercion is now derived from each field's **declared type** (`_declared_types` → `_COERCIONS`) instead of a hand-kept set of array keys, so `confidence` is coerced for the same reason `secondary_goals` is, and the asymmetry you found (`feedback_count` defensively `int()`-wrapped beside an unchecked `confidence`) cannot recur field by field. `emotional_impact: ["a", "b"]` becomes `"a; b"`; a dict in a string slot becomes JSON.

Two guards keep the derivation honest, each with a mutation as its positive control:
- every declared type must have a coercion, so adding a `boolean` or `number` field without teaching the boundary about it fails CI;
- every array declared inside a section must declare `items: string`, because the array coercion produces a list of strings and would otherwise flatten object items — the same contradiction one level down. `quotes` is the only array of objects and has its own projection.

### Found by re-auditing the fix, not by the review

**`get_project` had the live instance.** It reads `persona_id`/`name`/`tagline` with `.get(key, '')` under a schema declaring strings, on the same LLM-authored rows — and unlike `list_personas` it was callable all along, so this was the half that could actually have gone wrong in production. This PR is also what put `tagline` there. Coerced, along with the other declared strings in the same projection, since a split rule would have needed a paragraph to justify.

**A quote's own `source_feedback_id` is canonical and was undeclared.** It already travelled (the quote object is open), but a client could not rely on it, and it is the id `get_feedback_detail` takes. Declared, with a test derived from the canonical schema so a future quote field cannot be quietly omitted. Kept distinct from the persona's top-level `source_feedback_ids`, which stays dropped: one is where a sentence came from, the other is the row's provenance list.

### Nits

| Nit | Disposition |
|---|---|
| `_as_string_list` ships a Python `repr` for a dict | **Taken** — `json.dumps` via `DecimalEncoder`, asserted on a dict in a string slot |
| a `logger.warning` when a section is not an object | **Taken**, with the value never logged (customer-derived text) and a positive control that an *absent* section stays silent — most live rows omit sections, so warning on absence would make the log useless |
| `test_the_pdf_prompt_is_not_wired` couples to template wording | **Taken** — asserts the template's own `user_prompts['pdf']` is absent, so a reword cannot turn it green for an unrelated reason |
| production comments narrate PR history with emoji | **Taken** — the diff now adds zero emoji lines; what remains is the load-bearing why (`additionalProperties` open, unknown keys preserved, and the `.get` default that cannot correct a type). The PR-history narration lives here and in the write-up |
| a note about the breaking schema change | **Taken as a version bump, declined as a CHANGELOG entry.** `MCP_SERVER_VERSION` → **3.0.0**: the constant's own comment already records that a changed tool output shape is why 2.0.0 happened, and it is advertised in `serverInfo` on every `initialize`, so it is the signal a client actually receives. Not bumping while changing an output shape would contradict the rule sitting next to it. `CHANGELOG.md` is written at release time from claims verified against the code (the 0.2.0 precedent, #349) and has no `Unreleased` section; adding one here would invent a per-PR convention this repo does not have, and this change belongs in the next release's entry |

### Deliberately not bundled

`_project_feedback` has the identical `.get(key, '')`-under-a-declared-string pattern for `source`, `sentiment`, `category`, `urgency`, `persona_type` and the text fields. It is real and it is the same class — recorded here rather than fixed, because those rows come from a different writer with its own validation, the feedback tools have their own shape and tests, they shipped in #355 untouched by this diff, and no live counter-example is known (the persona case had five rows in four shapes). Folding it in would put a second contract into a PR about the persona one.

### Validation

| Gate | Result |
|---|---|
| pytest | **2966** passed (from 2958) |
| CDK vitest | **268** passed |
| ruff | **delta zero** on every file touched — same findings, same rule histogram |
| mutations | **14 run, 14 caught** |

The mutation set keeps the rule this PR learned the hard way — when a change touches both a schema and its producer, mutate **per side**: reverting the projection alone still passes the conformance gate, because with the schema corrected a dict is valid. Reverting the schema to `array<string>` is the mutation that proves the gate. Each new guard also gets its own control, since a guard that cannot fail proves nothing: a declared `boolean` with no coercion and an array of objects inside a section are both fed to their guards and both go red.

perrozzi:

Since the last review the author added `context_environment` and `scenario` to `_PERSONA_SECTIONS`, split `quotes` out into `_QUOTE_PROPERTIES` with a `_as_quote` projection that coerces bare strings instead of dropping them, generalised the coercion from arrays-only to per-declared-type (`_declared_types` + `_COERCIONS`), applied the same coercion to `get_project`'s summary fields, added a `logger.warning` for unreportable sections, decoupled the pdf-prompt test from template wording, trimmed the narrative/emoji comments, and bumped `MCP_SERVER_VERSION` to `3.0.0`. All three Should-fix items and every nit are addressed, and the new schema-vs-`persona.schema.json` completeness test is a good structural guard.

### Resolved
- `context_environment` and `scenario` are now declared, projected, asserted per-section, and enforced against the schema file by `test_every_canonical_persona_section_is_reported_or_excluded`.
- Bare-string `quotes` entries are coerced to `{"text": …}` via `_as_quote` instead of being filtered out.
- Declared strings/integers are now coerced too (`_as_string`/`_as_int` driven by `_declared_types`), including top-level `confidence` and `get_project`'s `tagline`, with `test_every_declared_persona_type_has_a_coercion` guarding future types.
- Nits: dict-in-string-slot is `json.dumps` not `repr`; non-object sections log a warning (with a no-log positive control and no customer text); pdf test compares against the template's own prompt; production comments de-narrated; breaking shape change signalled via the `3.0.0` server bump.

### New findings
- **Should check — `MCP_SERVER_VERSION` bump is not reflected anywhere else.** `voc-datalake/lambda/api/mcp_handler.py` moves `2.0.0` → `3.0.0`, but the incremental diff touches no test or doc that pins it. If any existing test asserts the `serverInfo.version` from `initialize`, or any `docs/`/README MCP section states `2.0.0`, they now disagree with the server; worth confirming and updating the MCP tool docs with the persona-shape change while you're there.
- **Nit — `assert len(sections) == 8` in `voc-datalake/lambda/api/test/test_mcp_delegation.py` undercuts the test's own stated purpose.** The docstring says a ninth canonical section "now fails here until it is either reported or excluded", but the count assertion fires first even when the new section *is* correctly reported, so the failure message misdirects. `assert len(sections) >= 8` plus the existing marker sanity check would keep the anti-vacuity guard without the false negative.
- **Nit — `_as_string` uses `json.dumps(value, cls=DecimalEncoder)`;** confirm `DecimalEncoder` is in scope in `mcp_handler.py` (it isn't visible in the diff's import block), since a dict in a declared string slot is exactly the untested-at-runtime path this helper exists for.

---

perrozzi:
## Round 3 — disposition of the follow-up review on `62df9e10`

Three findings, all taken. No production behaviour changed this round; two of the three were about whether a claim is *enforced*, which is the theme of the whole PR, so they land in tests.

**Should check — the `3.0.0` bump is reflected nowhere else.** Confirmed, and the reason is worse than a missed reference: nothing pins it. Checked every candidate —

- no test asserts an exact version. The only existing check is `test_the_server_version_is_semver_and_past_1`, i.e. `major >= 2`, which passes forever and **would have let this PR's own shape change ship under the version that describes the previous one**;
- no `docs/` page or README states an MCP server version. `docs/deployment.md`'s only MCP mention is a Lambda name, and the `2.0.0` in `docs/plugin-architecture.md` is a plugin manifest's `minPlatformVersion` example, unrelated;
- the remaining `2.0.0` strings in `mcp_handler.py` and `test_mcp_security.py` are prose about what changed *at* 2.0.0. Still true, so they stay;
- there are no MCP tool docs to update — the contract lives in `MCP_TOOLS` and the plan document, which is why the fix belongs in code.

So rather than update a reference, the declared `outputSchema`s are now **fingerprinted in lockstep with `MCP_SERVER_VERSION`** (`test_a_changed_tool_output_shape_moves_the_server_version`). Editing any published shape fails with a message saying which way to move the version — minor for an added field, major for a removal or retype, because a client validates `structuredContent` against these schemas. `sort_keys` means reordering a declaration is free; only content moves the fingerprint. This is the repo's existing lockstep pattern (14 such tests) applied to the one contract that leaves the account. A mutation proves it fires: adding a field to a tool's `outputSchema` goes red.

**Nit — `assert len(sections) == 8` misdirects.** Correct, and it contradicted the test's own docstring. Now `>= 8`: the count exists only so the `Section N:` marker check cannot pass vacuously, and a correctly reported ninth section must fail on the reporting check, which names it, rather than on a count that points at the wrong thing.

**Nit — confirm `DecimalEncoder` is in scope.** It is: `from shared.api import DecimalEncoder` at the top of `mcp_handler.py`, unchanged by this PR, which is why only `Callable` appears in the diff's import block. The path is also not untested at runtime — `test_a_list_in_a_DECLARED_STRING_slot_is_coerced` sends `scenario.narrative` a dict through the real handler and asserts the JSON that comes back, so a missing import fails there rather than in production.

Also folded in while touching the file: the canonical schema is read through one loader instead of being parsed separately in two tests.

The PR description has been rewritten to match the current diff — the numbers, the file list and the follow-on list in the original text described `3310f355`.

| Gate | Result |
|---|---|
| pytest | **2967** passed |
| CDK vitest | **268** passed |
| ruff | **delta zero** on every file touched |
| mutations | **15 run, 15 caught** |

perrozzi:

Since my last review the only changes are in `voc-datalake/lambda/api/test/test_mcp_delegation.py`: a `_canonical_persona_properties()` helper factored out of two schema reads, the anti-vacuity assertion relaxed to `>= 8` with an explanatory comment, and a new `test_a_changed_tool_output_shape_moves_the_server_version` that fingerprints every tool's `outputSchema` and pins it to `MCP_SERVER_VERSION`. No production code changed.

### Resolved
- The `== 8` count assertion is now `>= 8`, so a correctly-reported ninth section fails on the reporting check with a message that names it.
- `MCP_SERVER_VERSION` is now pinned by a test (`("3.0.0", "<fingerprint>")`), which covers the "nothing asserts the bump" half of my earlier finding and adds a real lockstep guard for future shape changes.

### Still open
- Confirm no `docs/`/README MCP section still states server version `2.0.0` — the new test pins the constant but not the documentation.
- Confirm `DecimalEncoder` is imported/defined in `mcp_handler.py` (used by `_as_string`); not visible in either diff.

### New findings
- **Nit — `voc-datalake/lambda/api/test/test_mcp_delegation.py`: `tool["outputSchema"]` will `KeyError` rather than fail meaningfully if any `MCP_TOOLS` entry omits `outputSchema`.** Either use `tool.get("outputSchema")` or assert presence first, so the failure reads as "tool X declares no output shape" instead of a raw `KeyError` inside the fingerprint helper.
- **Nit — the docstring's "reordering a declaration is free" is only true for object keys.** `sort_keys=True` does not normalise list order, so reordering `required` or `enum` entries moves the fingerprint despite being semantically identical; worth softening the claim so the next author isn't surprised by a "shape changed" failure on a no-op reorder.

Otherwise the incremental changes are clean and the fingerprint gate is a genuine improvement over `major >= 2`.

---

perrozzi:
## Round 4 — disposition of the follow-up review on `3faa6210`

Both nits taken; test-only, no production change.

**`tool["outputSchema"]` would `KeyError` instead of failing meaningfully.** Fixed with an explicit presence assertion that names the tool. Deliberately an `assert` rather than `.get()`: presence is `test_every_tool_declares_an_output_schema`'s claim, and swallowing an absent schema into a `None` would let the fingerprint quietly hash the absence instead of reporting it.

**"reordering a declaration is free" overstated `sort_keys`.** Correct — it normalises object keys, not list order, so reordering `required` or `enum` entries is semantically identical and still moves the fingerprint. The docstring now separates the two cases and says to restore the order rather than bump the version when that is what happened.

### The two "still open" confirmations

Both were answered with evidence in the disposition posted just before that push, and neither needs a change. Restating, since a diff-scoped review cannot see a comment:

- **No `docs/` page or README states an MCP server version.** `docs/deployment.md`'s only MCP mention is a Lambda name; the `2.0.0` in `docs/plugin-architecture.md` is a plugin manifest's `minPlatformVersion` example. The remaining `2.0.0` strings in `mcp_handler.py` and `test_mcp_security.py` are prose about what changed *at* 2.0.0 and are still true. There is no MCP tool documentation page to update — the published contract is `MCP_TOOLS` itself, which is why the new lockstep is the right place for the guard.
- **`DecimalEncoder` is imported** at `voc-datalake/lambda/api/mcp_handler.py:25` — `from shared.api import DecimalEncoder`. It predates this PR, which is why it is not in the diff's import block; only `Callable` was added there. The path is covered at runtime too: `test_a_list_in_a_DECLARED_STRING_slot_is_coerced` sends `scenario.narrative` a dict through the real handler and asserts the JSON that returns, so a missing import fails that test rather than production.

| Gate | Result |
|---|---|
| pytest | **2967** passed |
| ruff | **delta zero** on every file touched |
| mutations | **15 run, 15 caught** |

perrozzi:

Since the last review the only change is in `voc-datalake/lambda/api/test/test_mcp_delegation.py`: the fingerprint helper now asserts `"outputSchema" in tool` with a naming message before indexing, and the docstring distinguishes object-key order (free) from list order (`required`/`enum`, counts). Both of my previous nits are addressed; no production code moved.

### Resolved
- `tool["outputSchema"]` no longer raises a bare `KeyError` — failure now names the offending tool.
- The "reordering a declaration is free" claim is scoped to object keys, with the `required`/`enum` caveat spelled out.

### Still open
- Confirm no `docs/`/README MCP section still advertises server version `2.0.0`.
- Confirm `DecimalEncoder` is imported/defined in `mcp_handler.py` (used by `_as_string`); still not visible in either diff.

### New findings
None — the incremental change is clean.

---

perrozzi:
## Converged at `ce1274ac`

Round 4 returned **no new findings**. Four rounds, and the record for anyone picking this up:

| Round | Head | Findings | Outcome |
|---|---|---|---|
| 1 | `3310f355` | 3 `should_fix` + 5 nits | all taken; all three were the PR's own defect class turned on itself |
| 2 | `62df9e10` | 1 `should check` + 2 nits | version bump made enforceable rather than asserted |
| 3 | `3faa6210` | 2 nits | failure messages made to say what they mean |
| 4 | `ce1274ac` | none | — |

The two items the reviewer lists as "still open" are confirmation requests, not changes, and a diff-scoped review cannot see the answers. Both re-verified at this head across the **whole repo**, not just the diff:

- **`DecimalEncoder`** — imported at `voc-datalake/lambda/api/mcp_handler.py:25`. It predates this PR, which is why only `Callable` appears in the diff's import block, and the branch that uses it runs through the real handler in `test_a_list_in_a_DECLARED_STRING_slot_is_coerced`.
- **No MCP server version outside `mcp_handler.py`.** Checked `README.md`, all of `docs/`, and the frontend: `docs/deployment.md` mentions MCP only as a Lambda name, `docs/plugin-architecture.md`'s `2.0.0` is a plugin manifest's `minPlatformVersion` example, `McpAccessTab.tsx` shows no version, and the remaining `2.0.0` strings in `mcp_handler.py` / `test_mcp_security.py` are prose about what changed *at* 2.0.0 and are still true. There is no MCP tool documentation page — the published contract is `MCP_TOOLS`, which is why the new lockstep test is where that guard belongs.

Gates at this head: pytest **2967** · CDK vitest **268** · ruff **delta zero** on every file touched · **15 mutations run, 15 caught** · repo status check green · `MERGEABLE` / `CLEAN`.

Still true from the description, and neither is a review item: this is **not deployed** (the shared slot is contended, so `list_personas` keeps failing against the live server until it ships), and the seven phantom-key prompt builders remain a separate change.

perrozzi:
## Verified against the deployed server — 141 live checks

The reader half is now running on the unprefixed slot and **`list_personas` works against real data**. The before/after was captured with a real MCP client, which is the only instrument that ever saw this bug.

**Before** (this IDE's MCP client, pre-deploy):

```
MCP error -32602: Structured content does not match the tool's output schema:
data/personas/0/pain_points must be array, data/personas/0/behaviors must be array,
... (all five personas)
```

**After**, over four independent batteries:

| Battery | Result | What it proves |
|---|---|---|
| Targeted live checks | **33/33** | the published schema, the payload, and the live rows |
| Official MCP SDK client | **44/44** | a *validating* client accepts the payload — the instrument that found M1 |
| Raw transport | **30/30** | gateway, auth, throttle and protocol behaviour unchanged |
| REST sweep (`test-api.sh`) | **34/34** | the delegated routes behind the tools are healthy |

Highlights, all against the five real persona rows:

- `serverInfo.version` is **3.0.0**; the published `outputSchema` declares `pain_points`/`behaviors` as objects, carries all six object sections plus `quotes`, declares a quote's `source_feedback_id`, excludes `research_notes`, and **all six phantom keys are gone** (`goals`, `age_range`, `occupation`, `quote`, `journey_stage`, `type`).
- The **live payload conforms to the schema the server itself publishes** — the check that was impossible before, now run against production data rather than fixtures.
- Every declared field on every live row holds its declared type, and the imported rows' unforeseen keys survive the projection: `frustration`, `frustrations`, `primary_frustration`, `related_issues`. That is the `additionalProperties: true` decision paying off on real data.
- `get_project` reports a populated `tagline` for all five personas; `type` is gone.
- Four of five personas return all six sections; the fifth is genuinely sparse and returns four. `context_environment` and `scenario` — unreachable before this PR — carry content.

The harness's own M1 reproducer had to be **inverted rather than deleted**: it asserted "list_personas VIOLATES its own outputSchema" and now asserts the call succeeds and the shape is correct. Deleting it would have taken the evidence with it.

### One thing to know at merge time, found only by deploying

**An `outputSchema` change is invisible to an already-connected client.** Validation happens client-side against the tool list cached at connect time, so this IDE's session kept rejecting the *corrected* payload — with a newly mixed error naming both the old constraint and 8 `must NOT have additional properties`, the additional properties being the new fields. A fresh SDK process passes 44/44 against the same server.

The server honestly declares `capabilities: {"tools": {"listChanged": false}}`, so there is no notification path by design — this is a limitation, not a defect. The practical consequence: **after this merges and deploys, every existing MCP client needs one reconnect.** Worth a line in the release note.

### Deployment status

Deliberately **not** a `cdk deploy`. The app-wide diff moves 34 Lambdas across four stacks (every Lambda bundles `lambda/shared/`, which this PR touches), **adds** a `SubmitAnthropicUseCase` custom resource in `VocWebSearchStack`, and may replace the `DeployWebsite` custom resource — which would republish a `frontend/dist` built from a checkout shared with other sessions. None of that belongs to this PR.

So only `voc-mcp-api`'s code was updated, with the pre-change package captured for rollback. Lambda timestamps confirm the scope: `voc-mcp-api` at `2026-08-20T08:01:30Z`, every other function still `2026-08-19T17:47`.

Consequences, stated plainly: this is **CloudFormation drift** and any `cdk deploy` of `VocApiStack` overwrites it (harmlessly, once this is merged); the **writer half** (`persona_importer`) is **not** deployed, so imports still run the old prompt; and the description's "not deployed" line is temporarily true only of the writer.

perrozzi:
## Now properly deployed and tested end-to-end — 159 live checks

Supersedes the surgical single-Lambda note above. **Both halves are deployed** via a scoped `cdk deploy VocApiStack --exclusively`, and the **writer has now been exercised against a real model call for the first time**.

### The deploy, scoped rather than app-wide

`cdk deploy VocApiStack --exclusively --require-approval never` — exit 0 in 246s, 21 resources.

Pre-checked before running, which is what made `--require-approval never` defensible: the scoped diff had **no IAM and no security-group changes**, 0 added, 0 removed, 21 modified — 20 Lambdas (they all bundle `lambda/shared/`, which this PR touches) plus `DeployWebsite`. `--exclusively` kept the dependency stacks out, so `VocWebSearchStack`'s new `SubmitAnthropicUseCase` custom resource and the four-days-stale Core/Ingestion/Processing were **not** deployed.

`DeployWebsite` ships `Source.asset('frontend/dist')`, so the frontend was **rebuilt from `ce1274ac` in the deploy worktree** (3.5s, tracked tree clean) rather than publishing the copy I had taken from a checkout other sessions are editing. `config.json` is injected by CDK via `Source.data()`, so the build needed no env vars. The earlier CloudFormation drift is resolved — the MCP Lambda is CFN-managed again.

### The writer half, tested for real

`POST /projects/{id}/personas/import` with a pasted persona description → job `running → calling_ai → complete` → **18/18 checks**.

The persisted persona (`persona_20260820090319`, "Renata Bakalar") came back with **only canonical inner keys, in all six sections**:

| Section | Keys the model returned |
|---|---|
| `identity` | age_range, bio, education, family_status, income_bracket, location, occupation |
| `goals_motivations` | primary_goal, secondary_goals, success_definition, underlying_motivations |
| `pain_points` | current_challenges, blockers, workarounds, emotional_impact |
| `behaviors` | activity_frequency, current_solutions, decision_style, tech_savviness, tools_used |
| `context_environment` | usage_context, devices, time_constraints, social_context, influencers |
| `scenario` | title, narrative, trigger, outcome |

- **Zero keys from the old unconstrained prompt** — no `primary_frustration`, `frustration`, `tooling`, `current_practices`, `related_issues`. Those are exactly what the three pre-fix imported rows carry, and they are the reason this PR exists.
- **Types match the template's example values**: `current_challenges`/`blockers`/`workarounds` are lists, `primary_goal` is a string.
- **Quotes are objects**, not bare strings: `{"text": "I do not need another alert…", "context": "…"}`.
- **`llm_metadata: {"model": "global.anthropic.claude-sonnet-5", "prompt_version": "1.0.0"}`** — provenance an imported persona has never carried, so a row with odd keys can now be attributed to the prompt that produced it.

This is the check no unit test can perform. The tests assert what the prompt **says**; this asserts what the model **does** with it. Before deploying, the runtime `load_prompt_file` path was also verified against the deployed artifact — `prompts/persona-import.json` sits at the bundle root, so `/var/task/prompts` resolves, and the deployed template carries every key the handler subscripts.

### Reader and writer together

The newly imported persona is served correctly by the fixed reader: the SDK battery passes **44/44 at 6 personas**, including "every persona reports `pain_points` as an OBJECT" and "the two formerly-unreachable sections are present". So a canonically-shaped imported row now round-trips through `list_personas` to a validating client.

### Full tally against the deployed stack

| Battery | Result |
|---|---|
| Import end-to-end (fixed prompt, real Bedrock call) | **18/18** |
| Targeted MCP live checks | **33/33** |
| Official MCP SDK client (6 personas) | **44/44** |
| Raw transport | **30/30** |
| REST sweep post-deploy (`test-api.sh`) | **34/34** |

The REST sweep is a genuine regression test rather than a control this time: all 20 API Lambdas were redeployed, so it covers code that actually moved.

One operational note still stands from the earlier comment: an `outputSchema` change is invisible to an already-connected MCP client, and the server declares `tools.listChanged: false`, so **every existing client needs one reconnect** after this ships.

perrozzi:
## UI verification — and it found user-visible impact the MCP work never showed

Two things needed checking in the browser after the scoped deploy, and both were worth doing.

### 1. The published bundle works

The deploy republished `frontend/dist` from a bundle I built in a worktree, so that needed a smoke test rather than an assumption. On the live CloudFront site: Home and Project Detail render, the sidebar and menu config resolve, i18n strings resolve (no raw key paths), the Cognito session works, and **zero console errors or warnings**.

### 2. The import defect was visible in the product, not just through MCP

This is the part I did not expect to be so stark. Same Personas tab, same view, one expanded at a time — measured content length under each section heading:

| Section | New import (canonical prompt) | Pre-fix import (`Priya Raman`) |
|---|---|---|
| Identity & Demographics | 446 chars | 92 chars, and the fields shown are `company size`, `experience level`, `industry`, `role` |
| Goals & Motivations | 359 | 69 (primary goal only) |
| Pain Points & Frustrations | 455 | **0 — blank** |
| Behaviors & Habits | 222 | **0 — blank** |
| Context & Environment | 204 | **0 — blank** |
| Representative Quotes | 188 | **2 — renders as `""`** |
| Scenario | 610 | **0 — blank** |

The newly imported persona shows its real content throughout: occupation "Freight Dispatch Coordinator", location Rotterdam, scenario "Midnight Load Risk Check", Android device, the ten-minute handover constraint, and the quote *"…going to embarrass me at 6am."*

So the unconstrained prompt did not merely produce keys an MCP client rejected — **it produced personas that display with four empty sections and an empty quote block in the dashboard**, because the frontend reads the canonical key names and the old prompt invented its own. Every imported persona has looked like that since the feature shipped. That is a stronger argument for this PR than the schema story on its own.

### One consequence worth surfacing

The decision to leave the three existing non-canonical rows as they are is still reasonable, but it now has a measurable cost: those three personas will keep rendering with four blank sections and an empty quotes block until they are regenerated or re-imported. Fixing the writer stops new ones joining them; it does not repair them. Happy to re-import those three against the fixed prompt if that is wanted — it is the same one-call path this PR just exercised.

(Not a defect in this PR, noted while looking: the UI does not appear to render `context_environment.influencers`, which the canonical schema declares and the model now populates. Separate from the persona-shape contract.)

PR #357, opened 2026-08-20
fix(personas): stop feeding prompts fields no writer produces

perrozzi:
## Summary

Seven prompt builders read persona fields that **no writer has ever produced**: `goals`, `frustrations`, `needs`, and singular `quote`. Stored personas follow `schemas/persona.schema.json`, where goals live under `goals_motivations.primary_goal` / `.secondary_goals`, frustrations under `pain_points.current_challenges` / `.blockers`, and the voice line under `quotes[0].text`.

`.get(key, [])` returns the default for all of them, so every builder emitted its labels with empty values. Every generated PRD, PR/FAQ, merged document and research report was built from a persona block that looked like this:

```
**Priya Shah** — The Habitual Skimmer
- Goals:
- Frustrations:
```

That is worse than omitting the persona. An empty `Goals:` reads to a model as an assertion that the persona *has* no goals — and the document then records `used_persona_ids` in its derivation, claiming provenance from content that never arrived. Project chat rendered the same empty sections, and the roundtable prompt told the model to **be** a persona while handing it an empty identity.

Nothing errored, and the documents look complete, which is why this survived.

## Why every gate missed it

**No test at any of the seven sites asserted on the persona text.** They assert the persona's **name**, which is present whether or not the content arrives. The one test in the repo that asserted a persona *value* (`project-context.test.ts`) used a phantom-shaped fixture, so it asserted a value that only existed in the fixture.

## Changes

**Two shared readers, one per runtime**

- `lambda/shared/persona_context.py` — `persona_voice`, `persona_goals`, `persona_frustrations`, `persona_prompt_block`, `personas_prompt_context`. Framed like `shared/derivation.py`: one relation that had been spelled five different ways.
- `lambda/stream/src/context/persona-fields.ts` — the TypeScript twin for the two stream sites.

Both cap their lists deliberately: the research string crosses a **Step Functions state boundary** and the chat prompts compete with the conversation history budget. This is why neither reuses `projects.py::_persona_to_markdown`, which is whole-document, uncapped, and opens an `# H1` per persona — that would flatten the heading structure of the prompt it is embedded in. Its field paths are reused; its format is not.

A label is emitted **only when it has content**, so a persona with nothing recorded now says nothing about goals rather than asserting it has none. And `personas_prompt_context([])` returns `''`, so a caller's `or '(none)'` fallback can finally fire — it never could before, because the string contained the labels.

**Call sites**

| Site | Status |
|---|---|
| `jobs/document_generator/handler.py` | the **live** PRD/PR-FAQ path |
| `jobs/document_merger/handler.py` | live |
| `research/research_step_handler.py` | live; also fixes singular `quote` |
| `api/projects.py` — PR/FAQ wizard autofill | live |
| `api/projects.py` — `generate_prd` / `generate_prfaq` | documented `LEGACY — NOT REACHED`; migrated anyway, because leaving a phantom-key builder in the file is what invites the next copy-paste |
| `stream/src/context/project-context.ts` | project chat |
| `stream/src/context/persona-prompt.ts` | the roundtable identity block |

**The Zod boundary, which had to go first**

`projectItemSchema` declared `quote` / `goals` / `frustrations` / `needs` and omitted `goals_motivations`, `pain_points` and `quotes` **entirely** — so it stripped the real fields before either TypeScript reader could see them. It now declares them with `.passthrough()` on each object. Without this, a reader-only fix would have been inert.

**`needs`**

The key never existed on any row. Rather than drop the concept, it maps to the two canonical fields that answer the same question for a model: `goals_motivations.underlying_motivations` topped up with `pain_points.workarounds`.

## Tests

- **15 new guards** on the shared renderer, with fixtures copied from the shapes of live rows: generated, imported, and sparse.
- **Value assertions added where only the name was asserted.** Reverting the builder now fails a test at the document-generator and research sites.
- Four phantom-shaped fixtures converted to canonical.
- `project-context.test.ts`'s null-tolerance regression test now nulls the **canonical** fields. It nulled `goals`/`frustrations`/`needs`, which the schema no longer declares, so it would have kept passing while exercising nothing.

## Validation

| Gate | Result |
|---|---|
| pytest | **2959** passed — verified **+15** against a 2944 baseline on `development` (collect-only, both sides) |
| stream vitest | **339** passed |
| `typecheck:stream` | clean |
| `lint:stream` | back to its one pre-existing warning |
| ruff | **no new rule classes** on any changed file |
| mutations | **4 run, 4 caught** |

Reading phantom `goals`, reading phantom `frustrations`, reading singular `quote`, and emitting the label unconditionally each fail a test. The last reproduces the defect verbatim: `'**Sparse**\n- Goals: '`.

## For reviewers: this increases the prompt-injection surface

Persona text is LLM-authored from **customer-supplied documents**, and on these paths it was silently dropped. Making it arrive means an imported persona containing instruction-shaped text now reaches the PRD, research, merge and chat prompts.

I don't think it's a blocker — feeding persona content to models is the feature's purpose, and the content comes from the customer's own documents — but the exposure genuinely grows, so it belongs in the description rather than being discovered later.

## Scope

- **Independent of #356**: branched from `development`, disjoint file set, verified with `merge-base --is-ancestor`. Mergeable in either order.
- **Not deployed.**
- Not touched: `frontend/src/test/fixtures.ts::mockPersona` is also phantom-shaped and its docstring claims it matches the DynamoDB structure. Frontend-only, no production impact, left for a separate change.

### DIFF
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -1005,7 +1005,12 @@ def generate_prd(project_id: str, body: dict) -> dict:
     # Format personas for context. Migrated with the live sites even though this
     # function is unreached (see the LEGACY note above): leaving a phantom-key
     # builder in the file is what invites the next copy-paste.
-    personas_context = personas_prompt_context(personas)
+    #
+    # `or '(none)'` because the template hard-codes a `USER PERSONAS:` header
+    # before the placeholder. An empty value leaves that header with nothing under
+    # it, which is the same silent-empty-label defect this change removes, one
+    # layer up.
+    personas_context = personas_prompt_context(personas) or '(none)'
 
     feature_idea = body.get('feature_idea', 'Improve customer experience based on feedback')
 
@@ -1357,7 +1362,8 @@ def generate_prfaq(project_id: str, body: dict) -> dict:
     # reason. This one only ever lost the quote — `quote` is singular here while
     # rows store `quotes: [{text, context}]` — so it never had the empty
     # Goals/Frustrations lines the other builders did.
-    personas_context = personas_prompt_context(personas)
+    # `or '(none)'`: the template hard-codes a `USER PERSONAS:` header.
+    personas_context = personas_prompt_context(personas) or '(none)'
     
     feature_idea = body.get('feature_idea', 'New feature based on customer feedback')
 
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -14,6 +14,7 @@ from botocore.exceptions import ClientError
 from shared.logging import logger, tracer, metrics
 from shared.aws import get_dynamodb_resource, get_bedrock_client, BEDROCK_MODEL_ID
 from shared.api import validate_days, MAX_PERSONAS_PER_GENERATION
+from shared.persona_context import personas_prompt_context
 from shared.converse import converse_chain
 from shared.exceptions import (
     ConfigurationError,
@@ -1001,15 +1002,10 @@ def generate_prd(project_id: str, body: dict) -> dict:
     feedback_items = get_feedback_context(filters, limit=FEEDBACK_LIMIT_PRD)
     feedback_context = format_feedback_for_llm(feedback_items)
 
-    # Format personas for context
-    personas_context = ""
-    for p in personas:
-        personas_context += f"""
-**{p.get('name')}** - {p.get('tagline', '')}
-- Quote: "{p.get('quote', '')}"
-- Goals: {', '.join(p.get('goals', [])[:3])}
-- Frustrations: {', '.join(p.get('frustrations', [])[:3])}
-"""
+    # Format personas for context. Migrated with the live sites even though this
+    # function is unreached (see the LEGACY note above): leaving a phantom-key
+    # builder in the file is what invites the next copy-paste.
+    personas_context = personas_prompt_context(personas)
 
     feature_idea = body.get('feature_idea', 'Improve customer experience based on feedback')
 
@@ -1091,14 +1087,12 @@ def autofill_prfaq_questions(project_id: str, body: dict) -> dict:
     feedback_items = get_feedback_context(filters, limit=FEEDBACK_LIMIT_AUTOFILL)
     feedback_context = format_feedback_for_llm(feedback_items)
 
-    personas_context = ""
-    for p in personas:
-        personas_context += (
-            f"\n**{p.get('name')}** — {p.get('tagline', '')}\n"
-            f"Quote: \"{p.get('quote', '')}\"\n"
-            f"Goals: {', '.join(p.get('goals', [])[:3])}\n"
-            f"Frustrations: {', '.join(p.get('frustrations', [])[:3])}\n"
-        )
+    # `goals`, `frustrations` and singular `quote` are keys no writer produces, so
+    # this block used to reach the model with every label present and every value
+    # empty. Worse than omitting the persona: `Goals: ` with nothing after it reads
+    # as an assertion that the persona has none. Field paths now live in
+    # shared/persona_context.py.
+    personas_context = personas_prompt_context(personas)
 
     feature_idea = (body or {}).get('feature_idea', '').strip()
     title = (body or {}).get('title', '').strip()
@@ -1359,13 +1353,11 @@ def generate_prfaq(project_id: str, body: dict) -> dict:
     feedback_items = get_feedback_context(filters, limit=FEEDBACK_LIMIT_PRFAQ)
     feedback_context = format_feedback_for_llm(feedback_items)
 
-    # Format personas
-    personas_context = ""
-    for p in personas:
-        personas_context += f"""
-**{p.get('name')}**: {p.get('tagline', '')}
-Quote: "{p.get('quote', '')}"
-"""
+    # Format personas. Also unreached (LEGACY note above), migrated for the same
+    # reason. This one only ever lost the quote — `quote` is singular here while
+    # rows store `quotes: [{text, context}]` — so it never had the empty
+    # Goals/Frustrations lines the other builders did.
+    personas_context = personas_prompt_context(personas)
     
     feature_idea = body.get('feature_idea', 'New feature based on customer feedback')
 
--- a/voc-datalake/lambda/api/test/test_projects_ai_assists.py
+++ b/voc-datalake/lambda/api/test/test_projects_ai_assists.py
@@ -17,8 +17,12 @@ from unittest.mock import patch, MagicMock
 PROJECT_DATA = {
     'project': {'project_id': 'proj-1', 'name': 'Test', 'filters': {'days': 30}},
     'personas': [{
-        'name': 'Kim Jisu', 'tagline': 'Power user', 'quote': 'I love it',
-        'goals': ['speed'], 'frustrations': ['crashes'],
+        # Canonical shape; was flat `quote`/`goals`/`frustrations`, which no
+        # writer produces.
+        'name': 'Kim Jisu', 'tagline': 'Power user',
+        'quotes': [{'text': 'I love it'}],
+        'goals_motivations': {'primary_goal': 'speed'},
+        'pain_points': {'current_challenges': ['crashes']},
     }],
 }
 
--- a/voc-datalake/lambda/jobs/document_generator/handler.py
+++ b/voc-datalake/lambda/jobs/document_generator/handler.py
@@ -145,7 +145,11 @@ def _gather_context(
             # empty while `used_persona_ids` below recorded provenance from
             # content the model never received. Field paths now live in
             # shared/persona_context.py.
-            personas_context = personas_prompt_context(personas)
+            # `or '(none)'`: prd-generation.json and prfaq-generation.json both
+            # hard-code a `USER PERSONAS:` header before the placeholder, so an
+            # empty value leaves a bare header — the same empty-label defect being
+            # removed here, one layer up.
+            personas_context = personas_prompt_context(personas) or '(none)'
             for p in personas:
                 pid = p.get('persona_id')
                 if pid:
--- a/voc-datalake/lambda/jobs/document_generator/handler.py
+++ b/voc-datalake/lambda/jobs/document_generator/handler.py
@@ -47,6 +47,7 @@ from shared.jobs import job_handler, JobContext, update_job_status
 from shared.aws import get_dynamodb_resource
 from shared.converse import converse_chain
 from shared.feedback import query_feedback_by_date
+from shared.persona_context import personas_prompt_context
 from shared.api import validate_date_basis
 from shared.prompts import get_prd_generation_steps, get_prfaq_generation_steps
 from shared.prototypes import prototype_s3_key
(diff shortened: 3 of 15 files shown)


perrozzi:

This PR replaces seven ad-hoc persona prompt builders that read non-existent flat keys (`goals`, `frustrations`, `needs`, `quote`) with two shared readers — `lambda/shared/persona_context.py` and `lambda/stream/src/context/persona-fields.ts` — that read the canonical `goals_motivations` / `pain_points` / `quotes` paths, omit labels with no content, and cap list lengths. The diagnosis is correct and well-evidenced, the Python side is well tested, and widening `projectItemSchema` to declare the real fields is the right ordering. My main concerns are on the TypeScript side: the newly declared nested Zod shapes tighten a boundary that previously stripped these fields entirely, and the new TS reader has no direct test coverage.

### Blockers

- `voc-datalake/lambda/stream/src/context/project-context.ts`: declaring `quotes`, `goals_motivations` and `pain_points` in `projectItemSchema` means their *inner* values are now validated for the first time. Previously undeclared → stripped → never validated. Now a row with `pain_points: { blockers: null }`, `goals_motivations: { secondary_goals: null }`, or `quotes: [{ text: null }]` fails `z.array(z.string()).optional()`. `.passthrough()` only rescues *unknown* keys, not wrong-typed known ones, and the `nullsToUndefined` preprocess in front of the object will only help if it recurses (the diff doesn't show it, and the existing regression test in `project-context.test.ts` only nulls top-level fields). This is the exact failure mode the null-tolerance test was written for — "opaque Unknown error" taking down project chat — and these fields are LLM-authored, so nested nulls/odd shapes are likely. Please either confirm `nullsToUndefined` is deep, or make the nested leaves tolerant (`.nullish()`, or `z.array(z.unknown())` / `z.union([z.string(), z.record(z.unknown())])` for the LLM-written arrays), and add a test that nulls *nested* fields (`pain_points: { current_challenges: null }`) plus one where `secondary_goals` contains objects rather than strings. Note the asymmetry with the Python reader, which deliberately tolerates non-string list entries — the TS boundary currently rejects the whole item instead.

- Missing regression coverage on the TypeScript half of the fix. `persona-fields.ts` is new with no `persona-fields.test.ts`, and `persona-prompt.ts` — described in the PR as "the worst instance" — gets no new assertions at all. Per the repo rule, reverting `personaVoice`/`personaGoals`/`personaNeeds` or the label-omission logic in `personaIdentitySection` should fail a test. The `project-context.test.ts` fixture conversion may cover `buildPersonasContext` via pre-existing value assertions, but nothing covers the roundtable identity block or `personaNeeds` (whose whole mapping — `underlying_motivations` topped up with `workarounds` — is new behaviour with no test).

### Should fix

- `voc-datalake/lambda/shared/persona_context.py::_clean_items`: the `else: text = str(entry).strip()` branch will put Python reprs into a model prompt (`{'goal': 'x', 'weight': 2}`) for imported rows whose list entries are objects. Given the PR's own note that persona content is LLM-authored from customer documents, this is both noisy and an injection vector with extra structure. Prefer extracting a known text-ish key (`text`/`description`/`value`) and dropping otherwise, and add a fixture covering a list-of-dicts field.
- `personas_prompt_context` can now return `''`, which is new behaviour for `generate_prd`, `autofill_prfaq_questions`, `generate_prfaq` (`voc-datalake/lambda/api/projects.py`) and the document-generator path. The docstring asserts callers guard with `or '(none)'`, but none of those call sites in the diff show a fallback. Please confirm each prompt template doesn't now emit a bare `PERSONAS:` header with nothing under it — that is the same defect class this PR is fixing, one layer up. `document_merger/handler.py` gets this right with its `if persona_text:` guard.
- The two "twins" have drifted: Python caps at 3 and has no `needs` concept; TS caps at 4 and adds `personaNeeds`; labels differ (`- Voice:` vs `**Their voice:**`, `Frustrations` vs `What You Need`). The cap divergence is justified in the docstrings; the missing `needs` in Python and the label divergence are not. Either state explicitly that the Python renderer intentionally omits needs, or add it.
- `document_generator/handler.py` and `document_merger/handler.py` still append to `used_persona_ids` for personas that render to nothing but a name. The PR argues false provenance is the core harm; a persona contributing only its name still claims derivation. Not blocking, but worth a follow-up.

### Nits

- The module/class docstrings and inline comments are essay-length and use emoji (`🔴`, `🔑`) as headings, including inside test bodies. Two or three lines plus a pointer to the schema would age better; the "why every gate missed it" narrative belongs in the PR, not in `persona_context.py`.
- The `persona_context.py` module docstring is not a raw string, so the `\n` in the quoted example lines is interpreted as a real newline — the reproduction snippet renders mangled in `help()`. Use `r"""`.
- Type hints: `personas_prompt_context(personas: list, ...)` and `persona: dict` — prefer `Sequence[dict[str, Any]]` / `dict[str, Any]` to match the module's own `from typing import Any` usage.
- `_clean_items`'s `elif entry in (None, '', [], {})` is doing equality against a tuple of falsy containers; `elif not entry: continue` is equivalent here and clearer.
- `test_the_cap_bounds_the_block` asserts via `block.count('goal ')`, which would also match a goal string containing the substring elsewhere; asserting the rendered line equality is a tighter guard.

---

perrozzi:

This push adds one new file, `voc-datalake/lambda/shared/test/test_persona_context_lockstep.py`, a grep-based cross-runtime lockstep test pinning the field paths, caps, and exported names of the Python and TS readers. It is a useful addition, but none of my previous Blockers or Should-fix items were touched — no production code changed since the last review.

### Resolved
- Partially: reverting `personaVoice`/`personaGoals`/`personaFrustrations`/`personaNeeds` in `persona-fields.ts` to a phantom key (`.goals`, `.quote`, `.needs`) now fails `test_neither_runtime_reads_a_phantom_key`, and the Python/TS `needs` asymmetry is now explicitly pinned rather than undocumented drift.

### Still open
- Nested Zod leaves in `voc-datalake/lambda/stream/src/context/project-context.ts` (`pain_points: { current_challenges: null }`, `secondary_goals: [{...}]`) are now validated for the first time and will reject the whole item; no test nulls or mis-shapes a *nested* field.
- No behavioural test for the TS half: `persona-fields.test.ts` still absent, and `personaIdentitySection` label-omission in `persona-prompt.ts` has no assertion (the new lockstep test greps `persona-fields.ts` only).
- `_clean_items`'s `str(entry)` repr fallback in `persona_context.py`.
- Empty-string return from `personas_prompt_context` unguarded at the `projects.py` and `document_generator` call sites (bare `PERSONAS:` header risk).
- `used_persona_ids` still recorded for personas that render to a name only.
- Nits: emoji/essay docstrings (the new test file continues the pattern), `persona_context.py` docstring not raw, loose type hints, `block.count('goal ')` assertion.

### New findings
- `test_every_python_reader_has_a_typescript_counterpart` and `test_the_caps_differ_on_purpose_and_are_both_pinned` read `_ts_source()` (raw text) rather than `_ts_code()`, so a commented-out `export function personaVoice` or a `DEFAULT_PERSONA_ITEMS = 4` mentioned only in a doc comment would satisfy them. Use `_ts_code()` for consistency with the other two tests.
- The phantom-key guard scans only `persona_context.py` and `persona-fields.ts` — not the sites where the defect actually lived. Reintroducing `p.goals` in `persona-prompt.ts` or `project-context.ts::buildPersonasContext`, or `p.get('goals', [])` in `document_generator/handler.py`, passes this test. The Zod schema no longer declaring those keys is a partial TS-side backstop, but the Python call sites have none; consider extending the scanned file list to the seven call sites, which is where a copy-paste regression would land.
- `test_both_runtimes_read_the_same_canonical_field_paths` asserts bare substring presence anywhere in the file, so a field name surviving in a dead constant or unused helper satisfies it while the renderer no longer reads it. Acceptable for a grep lockstep, but the docstring's "KNOWN CEILING" note should say this too, not just the indirect-access case.
- `_repo_root()` hard-codes `parents[3]` plus a repo-relative TS path; if these Python tests are ever run against a packaged layer copy of `shared/` (without the sibling `stream/` tree), the `assert path.is_file()` turns into a spurious failure rather than a skip. Worth a `pytest.skip` when the TS file is absent, matching whatever the cited `test_avatar_image_model_lockstep.py` does.

---

perrozzi:

This push is production-code plus tests: `or '(none)'` guards added at the two `projects.py` legacy sites and `document_generator`, `str(entry)` replaced with a `_TEXTUAL_KEYS` reader in `persona_context.py` (raw docstring, tighter hints, caller-contract note), the nested Zod leaves in `project-context.ts` collapsed to opaque containers, `persona-fields.ts` reads routed through `isRecord`/`readKey`, a new `persona-fields.test.ts` with nested-malformed and roundtable label-omission cases, `_ts_code()` used consistently, a `pytest.skip` when repo files are absent, a per-call-site phantom-key scan, and the `count('goal ')` assertion replaced with whole-line equality. Essentially every previous finding is addressed.

### Resolved
- Nested Zod leaves no longer validated for the first time (`quotes`/`goals_motivations`/`pain_points` now opaque), and `persona-fields.test.ts` covers nulled/mis-shaped nested fields.
- Behavioural TS coverage exists, including `personaIdentitySection` label omission via `buildSinglePersonaPrompt`.
- `_clean_items`'s `str(entry)` repr fallback removed in favour of `_entry_text`.
- Empty-string return now guarded at `document_generator` and both legacy `projects.py` sites, with the caller contract documented on `personas_prompt_context`.
- `_ts_source()` → `_ts_code()` in the cap/counterpart tests; phantom-key scan extended to the call sites; `_repo_root()` missing-file case now skips; `block.count('goal ')` replaced; docstring not raw; loose hints tightened; KNOWN CEILING now states the substring limitation.

### Still open
- `used_persona_ids` is still recorded for personas that render to a name only.
- `autofill_prfaq_questions` remains bare `personas_prompt_context(personas)` — worth confirming that prompt does not hard-code a persona header the way `prd-generation.json` does.
- Emoji/essay-length docstrings throughout the new files.

### New findings
- **Should-fix — `voc-datalake/lambda/stream/src/context/project-context.ts`:** `z.record(z.unknown())` rejects non-object inputs and `z.array(z.unknown())` rejects non-array inputs, so a row with `goals_motivations: 'nope'` or `pain_points: []` or `quotes: {…}` now **throws** at `parse()` and takes down the whole chat-context build — the exact failure the new comment says is being avoided. Before this PR those keys were undeclared and silently stripped, so this is a new throw path. The tolerance tests in `persona-fields.test.ts` cast raw objects and bypass Zod, so they don't cover it. `z.unknown().optional()` for all three would match the stated "normalize at the reader, never reject the row" intent, since `isRecord`/`Array.isArray` already guard every read.
- **Runtime divergence not pinned:** `_entry_text` in `persona_context.py` extracts `text`/`description`/`value`/`name` from a dict entry, while TS `cleanStrings` drops non-string entries outright — so `secondary_goals: [{text: 'x'}]` reaches a PRD but not chat. The lockstep test pins field paths, caps and export names, not this. Also, the comment in `persona-fields.test.ts` ("The Python reader stringifies") is now stale.
- **Nit — `test_persona_context_lockstep.py`:** `CALL_SITES` lists six files while the comment and docstring say "seven files"; the seven *sites* live in six files.
- **Nit — `persona-fields.ts`:** two consecutive JSDoc blocks sit above `isRecord`; the first one describes `readKey`.

---

perrozzi:

This push is small and narrowly scoped: TypeScript `entryText`/`TEXTUAL_KEYS` added to `persona-fields.ts` so object list entries survive in both runtimes, the corresponding vitest cases rewritten (including the stale "the Python reader stringifies" comment), and a new lockstep test pinning the shared text keys. That resolves the runtime-divergence finding; the Zod throw-path finding from last round is untouched.

### Resolved
- Runtime divergence on object list entries: `entryText` now mirrors `_entry_text`, pinned by `test_both_runtimes_read_the_same_text_keys_off_an_object_entry`, and the stale test comment is gone.

### Still open
- **Should-fix — `stream/src/context/project-context.ts`:** `z.record(z.unknown())` / `z.array(z.unknown())` still throw on `goals_motivations: 'nope'`, `pain_points: []`, `quotes: {…}`; `z.unknown().optional()` matches the stated intent.
- `used_persona_ids` still recorded for name-only personas.
- `autofill_prfaq_questions` still bare `personas_prompt_context(personas)` — confirm no hard-coded persona header in that template.
- Emoji/essay-length docstrings; `CALL_SITES` says "seven files" for six; duplicated JSDoc block above `isRecord`.

### New findings
- **Nit — `shared/test/test_persona_context_lockstep.py`:** the new guard is `f"'{key}'" in ts`, a bare quoted-substring match. `'text'` already appears in `readKey(quote, 'text')` inside `personaVoice`, so removing `'text'` from `TEXTUAL_KEYS` would still pass; `'name'` is similarly at risk from any future quoted literal. Anchoring on the `TEXTUAL_KEYS = [...]` array body (as `test_the_caps_differ_on_purpose_and_are_both_pinned` does with a regex) would make it bite. It also reaches into the private `_TEXTUAL_KEYS`.
- **Nit — numeric entries still diverge:** DynamoDB returns numbers as `Decimal`, and `_entry_text`'s `isinstance(entry, (int, float))` rejects `Decimal`, while TS `typeof entry === 'number'` accepts it — so a numeric list entry renders in chat and vanishes from a PRD. Also `Number.isFinite` guards inf/NaN on the TS side with no Python counterpart. Both are marginal for LLM-authored text lists, but the new lockstep test doesn't cover the numeric branch it now shares.

---

PR #358, opened 2026-08-20
fix(mcp): make three declared contracts actually hold

perrozzi:
Three findings from the MCP client audit. All the same shape: something the server **declares** was not what it did, and the gap was invisible because the wrong answer came back with `isError: false`.

That last part is why these matter more through MCP than on the dashboard. A human sees `total_feedback: 6239` and a 99-item chart on one screen and notices. **A model calls one tool and reports the answer, with no second number to cross-check.**

## M1-class — `_project_feedback` now coerces by declared type

The follow-on named in #356's own PR body. Every field was read with `item.get(key, default)`, and a default fires only when a key is **absent** — it cannot correct a value of the wrong **type**. That is the defect that made `list_personas` uncallable, living a second time in the feedback projection.

Sharper here than in the persona case: `date` and `text` are **sliced** (`[:10]`, `[:_SUMMARY_TEXT_LIMIT]`), so a row storing either as a number or a dict raised `TypeError` inside the projection. One malformed row failed the **entire** `search_feedback` call, not just its own entry. Coercing first makes both slices safe by construction.

Driven by the declarations through the existing `_declared_types` / `_coerce_declared` rather than field by field — the second half of #356's lesson, where the hand-written version guarded `feedback_count` and left `confidence` unchecked. A new `_FEEDBACK_SOURCE_KEYS` states the renames once, since the projection renames and so the declared key is not the key to read.

Looked up with `.get(key, (key,))` rather than `[key]` on purpose: an unmapped declared property must not raise and kill the tool, which is the M1 failure mode itself. A test pins the map against the declarations, so the omission is still red in CI — the fallback only decides whether the symptom is a failing test or a dead tool in production.

## M2 — a query under the documented minimum was an empty *success*

The tool description has always said "at least 2 characters" while its `inputSchema` accepted `"a"` and the route answered `{'count': 0}` with HTTP 200. A claim about the **corpus** standing in for a fact about the **request**.

New `SEARCH_QUERY_MIN_LENGTH` in `shared/api.py`, beside `MAX_PERSONAS_PER_GENERATION` and for the reason recorded there: three places size themselves against this bound, and a prose promise cannot fail CI. The route now raises `ValidationError`; the schema **declares** `minLength` and builds its own description from the same constant.

The split is deliberate. A blank `q` is still a successful empty answer, because no search term means no search was asked for, and the filter-only answer belongs to `/feedback` — which is where the adapter already routes such a call. Only a term that is **present and too short** is a refusal.

The adapter needed no new validation: `_delegate` already maps a 4xx to `ToolRouteError` carrying the route's own message, so the model reads the real reason. Route owns the rule, adapter declares it.

## M3 + M5 — search read 30 days whatever you asked for, and never said so

`search_feedback` called `_scan_recent_items(min(days, 30), …)` while its `cutoff_date` came from the caller's full `days`. The two disagreed: at `days=365` the filter admitted a year of items and the candidate set only ever held thirty days of them, so anything older was unreachable by text search at **any** `days` value. On the corpus this was found on, a 5,240-item import sits 37 days back.

Both halves ship together, because the audit is explicit that the first alone moves the lie rather than fixing it — *"raising it without M5's truncation signal makes a wrong answer slower"*:

- the scan reads the **requested** window;
- `is_partial` is **kept** rather than discarded (`candidates, _ =`) and reported, so `count: 0` can be told apart from "the scan stopped early". The soft cap still bounds candidate collection, which is precisely why the signal is needed.

Named `is_partial_window` on the route as `/feedback` names it, and projected to `is_partial` to match its sibling tool `get_metrics_breakdown`. **Required** in the schema, because a flag that is sometimes missing reads as absence of truncation — the same mistake as asserting it false. Coerced with `bool()`, since a route answering `null` would otherwise reproduce M1 inside the very field added to fix M5.

### ⚠️ Existing MCP clients need one reconnect

`MCP_SERVER_VERSION` goes 3.0.0 → 3.1.0 — minor for an added field, by this repo's own rule — with the output fingerprint updated in the same commit as its test demands.

Minor is not invisible. These output shapes carry `additionalProperties: false`, and a client validates against the `tools/list` it cached at **connect**, so a session predating the deploy will reject the new field. The server honestly declares `tools.listChanged: false`, so there is no notification path to tell it. This is the deploy-only finding from #356, recurring exactly as recorded.

## Deliberately not in scope

- **The TypeScript twin** `Math.min(days, 30)` in the streaming chat tool. Same idea, third runtime, tracked by `BUG-chat-30-day-search-window`. Worth knowing: it is currently pinned by **no test at all**.
- **Parallelising the day fan-out.** The loop is forced by the schema — `gsi1pk = DATE#YYYY-MM-DD`, one partition per day, and no index supports a bounded multi-day Query with a usable projection. 365 sequential days measures ~4–5.5 s against a 20 s delegated-read timeout, and batching would change early-break semantics for six other `_scan_window_items` callers.
- **Pushing the text match into a `FilterExpression`.** DynamoDB `contains()` is case-**sensitive** while this route lowercases both sides, so a push-down would silently **drop** valid matches. Doing it properly needs a lowercased attribute written by the processor plus a backfill of the existing corpus — a separate change with a data migration.

## Verification

| Gate | Result |
|---|---|
| pytest (backend) | **3019**, up from 2995 — +24, matching the collected count exactly |
| ruff | delta **zero** on the changed files, identical rule distribution |
| `typecheck:all` | clean |
| CDK | 268 |
| stream | 365 |
| frontend | 3048 |

Mutation checks, **all caught**: restore the 30-day cap · assert `is_partial` false instead of computing it · drop the refusal · undeclare `minLength` · pass a null flag through.

One mutant **survived**, recorded rather than hidden: swapping `.get(key, (key,))` back to `[key]` is defensive only — the map is complete, so no test can distinguish it.

Two process notes, since both produced false confidence before being caught:

- `command -v ruff` finds nothing here (ruff is not on `PATH`), so a check invoked that way returns **empty with exit 1** — a clean result that actually means "did not run". The correct invocation is `voc-datalake/.venv/bin/python -m ruff`, as `npm run lint:python` does, and the harness now carries a positive control.
- The mutation anchor `'is_partial_window': window_truncated,` matched **twice**, because `list_feedback` publishes the same key from a same-named local. The harness reported anchor rot instead of a false "caught".

`npm run lint` still exits 1 on a pre-existing 514-error ruff baseline across `lambda/` and `plugins/`. Not from this change — six files are touched and their delta is zero.

## Also here

One unrelated four-line commit, kept separate: `.kiro/settings/mcp.json` stores its bearer token inline, the repo's own `.kiro/settings/` is tracked, and `git check-ignore` matched no rule for it — so sharing an MCP config with the team by moving it into the repo would have published a working credential to a public repo. Verified narrow: the target is now ignored, the tracked sibling is not, and the index is unchanged. No real credential is committed anywhere; a scan for token-shaped strings matches only synthetic fixtures.

## Follow-ups this leaves open

M4 (`is_partial: false` asserted unconditionally on the aggregate path) and M6 (`get_metrics_summary` and `search_feedback` contradicting each other, because `aggregator/handler.py` is INSERT-only so REMOVE never decrements) are **platform** defects that the adapter relays faithfully. M6 has a backfill question attached. Different size, different owner, not adapter work.

M7 (the persona dimension reading a field nothing populates) and M8 (a revoked credential completing the full handshake, which Phase 2b's version work touches anyway) remain open.

---

## Review rounds

**Round 1 — both blockers were real.** The first was one I documented and then failed to act on: `/feedback/search` trims `q` before applying its minimum while the frontend gate counted raw `.length`, so `"a "` was two characters on the client and one on the server, and a debounced search box could surface an HTTP 400 from ordinary typing. My own test docstring named that case and I left the client alone. Fixed, with the trim now at the API boundary every caller passes through.

The second blocker asked what happens to the new `is_partial_window` key at the frontend boundary. Checked rather than assumed: `searchFeedback` uses `fetchApi<T>`, a type assertion with no Zod there, so it cannot break at runtime — and it is not dropped either, because `pickActiveSource` → `toSimpleSource` → `extractTotals` already reads that key and `FeedbackPage` already declares it. What was missing was the declaration on the response type, now added with three tests covering true / false / omitted.

**Round 2** took the follow-up findings, and the reviewer's version of the main one is better than mine — I had patched one caller's gate instead of the shared boundary, and the body overclaimed that the frontend "sizes itself against" the Python constant when `SEARCH_MIN_CHARS` was an unpinned TS literal. Both remedies are in: the trim moved to `api.searchFeedback`, and a new `test_search_minimum_lockstep.py` parses the TS constant and asserts equality with `SEARCH_QUERY_MIN_LENGTH`, following `test_indexes.py` and carrying its positive control (a rename makes the parser return `None`, and without the control the equality check would pass while comparing against nothing — verified by mutation).

### Answers to the two items flagged as unconfirmed

These were answered in commit messages rather than here, which is why they read as open.

**Does the timeout cover `days=365` on the direct dashboard path?** Yes, with room. `api-stack.ts` gives `MetricsApi` a **30 s** timeout at 512 MB, against API Gateway's 29 s integration limit, and the measured worst case for a 365-day scan is **~4–5.5 s** — roughly 5× headroom. Result size stays bounded by `per_day_limit=300` and the 1,000-candidate soft cap; only the *query count* scales with `days`, and an empty day partition is ~0.5 RCU, so a 365-day scan of a sparse corpus is ~180 RCU on on-demand billing. The dashboard's own search default is `days=30`, so interactive latency is unchanged unless someone picks a custom year-long range.

**`query: ""` is schema-invalid for behaviour the route supports.** Confirmed the adapter performs no generic `inputSchema` argument validation, so nothing regresses server-side: a blank query still takes the filter-only branch. What remains is a strict schema over a lenient server, which is the safe direction — it produces no wrong answer, unlike the strict-claim/lenient-reality mismatches this PR fixes. The argument description already says "Omit to list by filters alone". Left as is deliberately; happy to relax it if you would rather the declaration mirror the tolerance exactly.

### Residual risk, stated rather than papered over

The lockstep pins the two **constants**, not the comparison. A future caller that gates on raw `.length` can still fire a request whose trimmed term is too short, because trimming at the boundary normalises what is sent without repairing a bad gate. An exported `isSearchable(text)` predicate would close it. I have deliberately not added one — there is exactly one caller today, so it would be an abstraction for a hypothetical second. Your call.

### Rounds 3–8

Converged. Findings decayed from blockers to formatting, and the substantive ones are all closed:

| Round | Substance |
|---|---|
| 3 | The `client.ts` comment overstated the boundary trim (it normalises the VALUE, not the DECISION); a guard pinned CHARACTERS not behaviour; a lockstep read raised `FileNotFoundError` instead of skipping |
| 4 | An assertion that could not fail for its stated reason — `metricsTimeout() <= 29 + 1` permitted exactly the 30 > 29 case its name claimed to prevent |
| 5 | `_DAYS_ARG.maximum` was still the literal `365` with the bound named only in prose — this PR's own defect class, in this PR's own file |
| 6–8 | Review narration in comments, a bodyless trailer, backticks, a blank line |

Two of those deserve calling out because they were mine and they were the same mistake this PR is about.

**The `+1` fudge.** I added a CDK guard named "keeps the budget inside what API Gateway will actually wait for" and then wrote `<= CEILING + 1` so it would pass against the current 30 s config. An assertion shaped around the value it is meant to constrain cannot fail for the reason it advertises. Resolved by dropping the pretence rather than widening it: the 30 s Lambda outliving the 29 s integration is a choice (`MetricsApi` serves every `/metrics/*` and `/feedback/*` route; a Lambda that dies sooner turns a slow-but-valid answer into a 502), so the test now asserts only what must hold and the reasoning sits in the `describe` header.

**`_DAYS_ARG` declared `365` as a literal** with the real ceiling named only in the comment above it — declaration-vs-enforcement drift, in the file whose purpose here is removing exactly that. Now imported from the new `MAX_FEEDBACK_WINDOW_DAYS`, which `validate_days` also takes its default from and which the CDK guard parses. Value-preserving, so the output fingerprint does not move.

### The two standing questions, answered with evidence

**"Grep to confirm the cited test name before merge."** Discharged by construction rather than assertion — the qualified name is extracted *from the comment* and handed to pytest as a node id:

```
CITED=$(grep -oE 'test_mcp_delegation\.py::TestRouteSelection::test_a_blank_query_is_not_a_search' \
        voc-datalake/lambda/api/mcp_handler.py)
.venv/bin/python -m pytest "lambda/api/test/$CITED"   # 1 passed
```

It resolves and passes, at `test_mcp_delegation.py:374`. It is pre-existing, which is why it appears in no diff.

**"The merged citation line is ~105 chars — confirm ruff is still clean."** It is 106, and it is not a violation: the project's `ruff.toml` does not select E501. Forcing `--select E501` reports **40** pre-existing long lines in that file alone, so line length is deliberately unenforced here. Consistent with the measured delta:

```
ruff on the changed files — HEAD=28  WORKING=28  delta=0   (identical rule distribution)
```

### Final gate state

pytest **3021** · frontend **3057** · CDK **269** · stream **365** · `typecheck:all` clean · eslint `--max-warnings 0` clean · ruff delta **zero**.

Mutation-verified rather than assumed, across the whole change: restore the 30-day cap · assert `is_partial` false instead of computing it · drop the query refusal · undeclare `minLength` · pass a null flag through · drop the type coercion · swap presence for truthiness · revert the client trim · diverge the two language constants · rename the TS constant (fails the positive control first) · raise the window ceiling to 3650 (fails the fan-out budget with `expected 30 to be greater than or equal to 109.5`). One survivor recorded rather than hidden: `.get(key, (key,))` → `[key]` is defensive only, since the map is complete.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -79,3 +79,7 @@ voc-datalake/Workshop/*
 .kiro/memory/
 .kiro/sessions/
 .kiro/NEW_SESSION_PROMPT.md
+# MCP client config — carries a live bearer token inline in headers.Authorization.
+# .kiro/settings/ is otherwise tracked (context-rules.json), so without this line
+# sharing an MCP config with the team publishes a working credential to a PUBLIC repo.
+.kiro/settings/mcp.json
--- a/voc-datalake/frontend/src/api/client.test.ts
+++ b/voc-datalake/frontend/src/api/client.test.ts
@@ -1790,10 +1790,9 @@ describe('searchFeedback trims the query at the boundary', () => {
   // present-but-too-short term with a 400. Trimming here means the string that
   // is SENT is the string the route measures, whatever a caller passed in.
   //
-  // Asserted on the REQUEST URL rather than on the source text of client.ts. A
-  // sibling guard used to assert the literal substring `q: params.q.trim()`,
-  // which a Prettier reflow or an extracted local would break with no behaviour
-  // change — pinning characters instead of behaviour.
+  // Asserted on the REQUEST URL, not on the source text of client.ts. Matching a
+  // literal like `q: params.q.trim()` would pin characters rather than behaviour,
+  // and break on a reformat or an extracted local with nothing having changed.
   beforeEach(() => {
     ;(useConfigStore.getState as ReturnType<typeof vi.fn>).mockReturnValue(DEFAULT_STORE_STATE)
     ;(authService.refreshSession as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
--- a/voc-datalake/frontend/src/api/client.test.ts
+++ b/voc-datalake/frontend/src/api/client.test.ts
@@ -1784,3 +1784,57 @@ describe('getDateRangeParams', () => {
     })
   })
 })
+
+describe('searchFeedback trims the query at the boundary', () => {
+  // `/feedback/search` trims `q` before applying its minimum and refuses a
+  // present-but-too-short term with a 400. Trimming here means the string that
+  // is SENT is the string the route measures, whatever a caller passed in.
+  //
+  // Asserted on the REQUEST URL rather than on the source text of client.ts. A
+  // sibling guard used to assert the literal substring `q: params.q.trim()`,
+  // which a Prettier reflow or an extracted local would break with no behaviour
+  // change — pinning characters instead of behaviour.
+  beforeEach(() => {
+    ;(useConfigStore.getState as ReturnType<typeof vi.fn>).mockReturnValue(DEFAULT_STORE_STATE)
+    ;(authService.refreshSession as ReturnType<typeof vi.fn>).mockResolvedValue(undefined)
+    global.fetch = vi.fn()
+  })
+
+  afterEach(() => {
+    vi.restoreAllMocks()
+  })
+
+  const okOnce = (body: unknown) =>
+    (global.fetch as ReturnType<typeof vi.fn>).mockResolvedValueOnce({
+      ok: true,
+      json: () => Promise.resolve(body),
+    })
+
+  const requestedUrl = () =>
+    String((global.fetch as ReturnType<typeof vi.fn>).mock.calls[0][0])
+
+  it('sends the trimmed term when the caller passes surrounding whitespace', async () => {
+    okOnce({ count: 0, items: [], entities: {}, query: 'delivery' })
+
+    await api.searchFeedback({ q: '  delivery  ' })
+
+    expect(requestedUrl()).toContain('q=delivery')
+  })
+
+  it('preserves interior spaces, which are part of the term', async () => {
+    okOnce({ count: 0, items: [], entities: {}, query: 'slow delivery' })
+
+    await api.searchFeedback({ q: '  slow delivery  ' })
+
+    // URLSearchParams encodes the space; what matters is that it survives.
+    expect(decodeURIComponent(requestedUrl().replace(/\+/g, ' '))).toContain('q=slow delivery')
+  })
+
+  it('surfaces the truncation flag the route reports', async () => {
+    okOnce({ count: 1, items: [{ feedback_id: '1' }], entities: {}, query: 'x', is_partial_window: true })
+
+    const result = await api.searchFeedback({ q: 'delivery' })
+
+    expect(result.is_partial_window).toBe(true)
+  })
+})
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -189,15 +189,25 @@ export const api = {
   },
   
   searchFeedback: async (params: { q: string; days?: number; date_basis?: DateBasis; limit?: number; source?: string; sentiment?: string; category?: string }) => {
-    // `q` trimmed HERE, at the single boundary every caller goes through, rather
-    // than in each caller's gate.
+    // `q` trimmed HERE, at the single boundary every caller goes through, so the
+    // string that is SENT is the string the route measures.
     //
     // `/feedback/search` trims before applying `SEARCH_QUERY_MIN_LENGTH` and
-    // REFUSES a present-but-too-short term with a 400, so any caller comparing
-    // raw `.length` could ship `"a "` — two characters here, one there — and
-    // surface a server error from ordinary typing. Trimming at the boundary makes
-    // every caller agree with the route by construction; a caller's own gate then
-    // only decides WHETHER to search, not what counts as long enough.
+    // refuses a present-but-too-short term with a 400, so a caller passing `"a "`
+    // through untrimmed would have the server measure something different from
+    // what the caller measured.
+    //
+    // ⚠️ Precisely what this does and does not buy: it normalises the VALUE, not
+    // the DECISION. A caller that gates on raw `.length` will still let `"a "`
+    // past its own gate, and this boundary will faithfully send `q=a` and get a
+    // 400. Only a caller's own TRIMMED gate prevents that — `useFeedbackListData`
+    // has one, and `test_search_minimum_lockstep.py` pins its constant to the
+    // route's.
+    //
+    // Deliberately NOT short-circuiting a too-short term into an empty result
+    // here, which was the reviewer's alternative: returning `count: 0` for a
+    // search that never ran is exactly the ambiguity the route was just fixed to
+    // stop producing. Better a loud 400 than a quiet zero.
     const searchParams = buildSearchParams({ ...params, q: params.q.trim() })
     // `is_partial_window` declared, not merely surviving the spread below: the
     // route sets it when the candidate scan stops on its soft cap, and
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -204,10 +204,10 @@ export const api = {
     // has one, and `test_search_minimum_lockstep.py` pins its constant to the
     // route's.
     //
-    // Deliberately NOT short-circuiting a too-short term into an empty result
-    // here, which was the reviewer's alternative: returning `count: 0` for a
-    // search that never ran is exactly the ambiguity the route was just fixed to
-    // stop producing. Better a loud 400 than a quiet zero.
+    // A too-short term is deliberately NOT short-circuited into an empty result
+    // here, because returning `count: 0` for a search that never ran is the same
+    // ambiguity the route was fixed to stop producing — moving it from the server
+    // to the client would not make it honest. A loud 400 beats a quiet zero.
     const searchParams = buildSearchParams({ ...params, q: params.q.trim() })
     // `is_partial_window` declared, not merely surviving the spread below: the
     // route sets it when the candidate scan stops on its soft cap, and
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -190,7 +190,12 @@ export const api = {
   
   searchFeedback: async (params: { q: string; days?: number; date_basis?: DateBasis; limit?: number; source?: string; sentiment?: string; category?: string }) => {
     const searchParams = buildSearchParams(params)
-    const res = await fetchApi<{ count: number; items: FeedbackItem[]; entities: EntitiesResponse['entities']; query: string }>(`/feedback/search?${searchParams}`)
+    // `is_partial_window` declared, not merely surviving the spread below: the
+    // route sets it when the candidate scan stops on its soft cap, and
+    // `extractTotals` already reads that key for the search branch, so the "N+"
+    // display works either way. Declaring it is what tells the next reader the
+    // field is real rather than incidental.
+    const res = await fetchApi<{ count: number; items: FeedbackItem[]; entities: EntitiesResponse['entities']; query: string; is_partial_window?: boolean }>(`/feedback/search?${searchParams}`)
     return { ...res, items: normalizeFeedbackItems(res.items) }
   },
   
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -189,13 +189,23 @@ export const api = {
   },
   
   searchFeedback: async (params: { q: string; days?: number; date_basis?: DateBasis; limit?: number; source?: string; sentiment?: string; category?: string }) => {
-    const searchParams = buildSearchParams(params)
+    // `q` trimmed HERE, at the single boundary every caller goes through, rather
+    // than in each caller's gate.
+    //
+    // `/feedback/search` trims before applying `SEARCH_QUERY_MIN_LENGTH` and
+    // REFUSES a present-but-too-short term with a 400, so any caller comparing
+    // raw `.length` could ship `"a "` — two characters here, one there — and
+    // surface a server error from ordinary typing. Trimming at the boundary makes
+    // every caller agree with the route by construction; a caller's own gate then
+    // only decides WHETHER to search, not what counts as long enough.
+    const searchParams = buildSearchParams({ ...params, q: params.q.trim() })
     // `is_partial_window` declared, not merely surviving the spread below: the
     // route sets it when the candidate scan stops on its soft cap, and
     // `extractTotals` already reads that key for the search branch, so the "N+"
     // display works either way. Declaring it is what tells the next reader the
     // field is real rather than incidental.
     const res = await fetchApi<{ count: number; items: FeedbackItem[]; entities: EntitiesResponse['entities']; query: string; is_partial_window?: boolean }>(`/feedback/search?${searchParams}`)
+
     return { ...res, items: normalizeFeedbackItems(res.items) }
   },
   
--- a/voc-datalake/frontend/src/pages/Categories/useFeedbackListData.test.tsx
+++ b/voc-datalake/frontend/src/pages/Categories/useFeedbackListData.test.tsx
@@ -262,3 +262,65 @@ describe('useFeedbackListData', () => {
     })
   })
 })
+
+describe('the search gate measures what the server measures', () => {
+  it('does not search when the term trims below the minimum', async () => {
+    // `/feedback/search` trims `q` and now REFUSES a present-but-too-short term
+    // with a 400 instead of answering an empty success. Gating on raw `.length`
+    // made `"a "` two characters here and one there, so an ordinary typing
+    // sequence produced a server error. Nothing should leave the client.
+    const { result } = renderData({ ...baseFilters, searchText: 'a ' })
+
+    await waitFor(() => expect(result.current.isSearching).toBe(false))
+    expect(mockSearchFeedback).not.toHaveBeenCalled()
+  })
+
+  it('sends the trimmed term so the client and the route agree on its length', async () => {
+    renderData({ ...baseFilters, searchText: '  delivery  ' })
+
+    await waitFor(() => expect(mockSearchFeedback).toHaveBeenCalled())
+    expect(mockSearchFeedback.mock.calls[0][0]).toMatchObject({ q: 'delivery' })
+  })
+
+  it('still searches when only the interior has spaces', async () => {
+    renderData({ ...baseFilters, searchText: 'slow delivery' })
+
+    await waitFor(() => expect(mockSearchFeedback).toHaveBeenCalled())
+    expect(mockSearchFeedback.mock.calls[0][0]).toMatchObject({ q: 'slow delivery' })
+  })
+})
+
+describe('a truncated search window reaches the display', () => {
+  it('reports isPartialWindow from the search response', async () => {
+    // The route sets `is_partial_window` when the candidate scan stops on its
+    // soft cap. Without this the dashboard would show a bare count for a
+    // truncated scan — the same ambiguity the MCP tool was fixed for.
+    mockSearchFeedback.mockResolvedValue({
+      count: 1, items: [makeItem()], is_partial_window: true,
+    })
+
+    const { result } = renderData({ ...baseFilters, searchText: 'delivery' })
+
+    await waitFor(() => expect(result.current.isPartialWindow).toBe(true))
+  })
+
+  it('reports a complete search window as complete', async () => {
+    mockSearchFeedback.mockResolvedValue({
+      count: 1, items: [makeItem()], is_partial_window: false,
+    })
+
+    const { result } = renderData({ ...baseFilters, searchText: 'delivery' })
+
+    await waitFor(() => expect(result.current.filteredFeedback).toHaveLength(1))
+    expect(result.current.isPartialWindow).toBe(false)
+  })
+
+  it('treats a route that omits the flag as complete', async () => {
+    mockSearchFeedback.mockResolvedValue({ count: 1, items: [makeItem()] })
+
+    const { result } = renderData({ ...baseFilters, searchText: 'delivery' })
+
+    await waitFor(() => expect(result.current.filteredFeedback).toHaveLength(1))
+    expect(result.current.isPartialWindow).toBe(false)
+  })
+})
--- a/voc-datalake/frontend/src/pages/Categories/useFeedbackListData.ts
+++ b/voc-datalake/frontend/src/pages/Categories/useFeedbackListData.ts
@@ -92,13 +92,21 @@ export function useFeedbackListData(
   filters: CategoryFiltersState,
   apiEndpoint: string
 ): FeedbackListData {
-  const isSearching = filters.searchText.length >= SEARCH_MIN_CHARS
+  // TRIMMED, and measured on the same string the server measures.
+  //
+  // `/feedback/search` trims `q` before applying its own minimum and now REFUSES
+  // a present-but-too-short term instead of answering an empty success. Gating on
+  // raw `.length` would send `"a "` — two characters here, one there — and the
+  // user would see an error from an ordinary typing sequence. Sending the trimmed
+  // value too keeps the cache key, the request and the server's view identical.
+  const searchText = filters.searchText.trim()
+  const isSearching = searchText.length >= SEARCH_MIN_CHARS
   const enabled = computeEnabledQueries(apiEndpoint, isSearching, filters.showUrgentOnly)
   const commonParams = buildCommonParams(dateParams, filters)
 
   const searchQuery = useQuery({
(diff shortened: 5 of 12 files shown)


perrozzi:

This PR closes three MCP contract gaps: `_project_feedback` now coerces every field to its declared type (driven by a new `_FEEDBACK_SOURCE_KEYS` map plus the existing `_declared_types`/`_coerce_declared`), `/feedback/search` refuses a present-but-too-short `q` with a `ValidationError` sourced from a new `shared.api.SEARCH_QUERY_MIN_LENGTH`, and the search candidate scan reads the requested window while surfacing `is_partial_window` → `is_partial`. The backend work is well-targeted, the constant is shared correctly rather than restated, and each fix ships with a regression test. Two things need attention: the client side of the new 400, and the latency/cost profile of an un-capped day fan-out on a dashboard-facing route.

### Blockers

- `voc-datalake/lambda/api/metrics_handler.py` — the new 400 is a breaking change for the existing frontend and nothing on that side was updated. The route trims before measuring, while (per your own test docstring in `test_whitespace_that_trims_below_the_minimum_is_refused`) the frontend's `SEARCH_MIN_CHARS` gate counts raw `.length`. So a debounced search box goes from "1 char → gated" to "`\"a \"` → 2 raw chars → request → HTTP 400", which is an easily reachable typing sequence, not an exotic edge case. Either trim in the frontend gate before comparing, or treat trim-below-minimum as the blank case server-side. Whatever error surface this lands on also needs i18n keys for all locales rather than a raw relayed message.
- `voc-datalake/lambda/api/metrics_handler.py` (search response shape) — `is_partial_window` is a new field on `/feedback/search`. If the frontend validates this response with a Zod object that isn't passthrough (repo pattern is Zod at the boundary), the added key breaks dashboard search at runtime; if it *is* passthrough, the flag is silently dropped and the dashboard keeps showing `count: 0` for a truncated scan — the exact ambiguity this PR argues is unacceptable for MCP. The schema/type and its consumer should be updated in this PR, not left implicit.

### Should fix

- `candidates, window_truncated = _scan_recent_items(days, per_day_limit=300, …)` removes the `min(days, 30)` budget on a route the dashboard hits interactively. The PR body measures 365 sequential days at ~4–5.5 s against a delegated-read timeout, but that budget is for the MCP path only — the direct API path is bounded by the metrics Lambda timeout and API Gateway's 29 s integration limit, and per-request RCU now scales with `days`. Please confirm the metrics Lambda timeout in the CDK stack covers the worst case, and consider whether `days=365` search should stay reachable without a stricter `per_day_limit`/soft cap.
- Test coverage of the source-key map is asymmetric: `test_source_key_map_covers_every_declared_field` pins `_FEEDBACK_DETAIL_TYPES` only. Add the summary side (`_FEEDBACK_SUMMARY_TYPES - mapped == set()`) so a summary-only declared property can't fall through to the `.get(key, (key,))` self-name read.
- The MCP `query` argument now declares `minLength: 2`, but the server still treats a present-and-blank `q` as a successful filter-only answer. A client that sends `query: ""` (rather than omitting it) is now schema-invalid for behaviour the route still supports. The description says "Omit", so this is mostly fine — but if the adapter validates `arguments` against `inputSchema` anywhere, that path should be checked so blank doesn't start erroring.

### Nits

- `_declared_types` (context line in the diff) filters out `"object"`/`"array"` declared types. Worth confirming `keywords` (the one declared array) still lands in `_FEEDBACK_DETAIL_TYPES`; if it doesn't, the detail payload silently loses the field and both `test_a_bare_string_in_the_declared_keywords_array_becomes_a_list` and the map-coverage test would be the ones to catch it — good, but the interaction deserves an explicit comment.
- `_FEEDBACK_SOURCE_KEYS` mixes double-quoted keys with single-quoted tuple values in the same literal; pick one.
- `_row_value(item: dict, …)` — annotate as `dict[str, Any]` to match the surrounding code.
- `test_a_scan_that_stops_on_the_soft_cap_says_so` relies on rows *omitting* `date` so `basis_date` returns `''` and everything filters out; that's load-bearing but only explained in the sibling test. A one-line comment would keep the next reader from "fixing" it.
- `.gitignore`: consider also committing a `.kiro/settings/mcp.json.example` with a placeholder token so the ignore rule doesn't just make the config undiscoverable for new contributors.

---

perrozzi:

Since the last review the author fixed the client side: `useFeedbackListData` now trims before gating and sends the trimmed term (so `"a "` can no longer produce a 400), `client.ts` declares `is_partial_window?: boolean` on the search response and the hook's `isPartialWindow` is pinned by four new tests, the source-key map test now covers the summary declaration set too, and the two annotation/comment nits are done. Both previous blockers are addressed; two Should-fix items remain.

### Resolved
- Frontend gate now measures the same string the route measures (`useFeedbackListData.ts`), so the new 400 is unreachable from ordinary typing — and no error surface means no new i18n keys needed.
- `is_partial_window` is declared on the search response type and reaches the display, with truncated/complete/omitted cases tested (`client.ts`, `useFeedbackListData.test.tsx`).
- `test_source_key_map_covers_every_declared_field` now asserts the summary set and the superset relation separately.
- `_row_value(item: dict[str, Any], …)` annotation; load-bearing "`date` omitted on purpose" comment added to `test_a_scan_that_stops_on_the_soft_cap_says_so`.

### Still open
- Un-capped day fan-out on `/feedback/search`: still no confirmation the metrics Lambda timeout / 29 s API Gateway integration limit covers `days=365` on the direct dashboard path.
- MCP `query` now declares `minLength: 2` while the route still supports a present-but-blank `q`; a client sending `query: ""` is schema-invalid for supported behaviour.

### New findings
- `voc-datalake/frontend/src/pages/Categories/useFeedbackListData.ts` — the trim was applied to this one hook only. The PR body claims the frontend "sizes itself against" `SEARCH_QUERY_MIN_LENGTH`, but `SEARCH_MIN_CHARS` is still an independent TS literal with nothing pinning the two languages together, and any other consumer of `SEARCH_MIN_CHARS` / `api.searchFeedback` that compares raw `.length` can still ship `"a "` and get a 400. Worth either centralising the trim in `api.searchFeedback` (trim `q` there, so every caller agrees with the route by construction) or adding a check that the TS constant matches the Python one.
- `voc-datalake/frontend/src/pages/Categories/useFeedbackListData.test.tsx` — the four `isPartialWindow` tests were added with no accompanying production change in the hook, so they pin behaviour that must already flow through `extractTotals`. Fine if green, but the "reports isPartialWindow from the search response" case is the only one that would catch a regression there; the other three pass on a `false`/`undefined` default. Consider also asserting the "N+" display consumes it, so a future refactor of `extractTotals` can't quietly drop the flag again.

---

perrozzi:

Since the last review the trim moved to the single API boundary (`api.searchFeedback` now sends `params.q.trim()`, with the hook's own trimmed gate retained), and a new cross-language guard test (`lambda/shared/test/test_search_minimum_lockstep.py`) parses `SEARCH_MIN_CHARS` out of the hook and asserts it equals `SEARCH_QUERY_MIN_LENGTH`, plus a positive control against parser drift. That addresses both of my previous new findings; the two Should-fix items from the earlier review are unchanged.

### Resolved
- Trim centralised in `api.searchFeedback`, so every caller sends the same string the route measures (`frontend/src/api/client.ts`).
- TS `SEARCH_MIN_CHARS` is now pinned to the Python `SEARCH_QUERY_MIN_LENGTH` in CI, with a positive control for the parser (`test_search_minimum_lockstep.py`).

### Still open
- Un-capped day fan-out on `/feedback/search` vs. the metrics Lambda timeout / 29 s API Gateway integration limit at `days=365` on the direct dashboard path.
- MCP `query` declares `minLength: 2` while the route still supports a present-but-blank `q`, so `query: ""` is schema-invalid for supported behaviour.

### New findings
- `voc-datalake/frontend/src/api/client.ts` — the comment overstates what the boundary trim buys: trimming changes *what* is sent, not *whether* it is sent. A caller that still gates on raw `filters.searchText.length` will pass `"a "` its own gate, and the boundary will faithfully send `q=a` → 400. Only `useFeedbackListData`'s trimmed gate prevents the error; the "every caller agrees with the route by construction" claim only holds for the value, not for the decision. Either soften the comment or have `searchFeedback` short-circuit a trimmed term below the minimum (returning the empty shape) so no caller can emit a refusal.
- `voc-datalake/lambda/shared/test/test_search_minimum_lockstep.py` — `test_the_client_trims_before_measuring_against_the_route` asserts the literal substring `'q: params.q.trim()'`. That is a formatting-sensitive assertion: a Prettier/ESLint reflow, or extracting `const q = params.q.trim()`, breaks a green test with no behaviour change. A regex tolerant of whitespace, or (better) a frontend test asserting the request payload is trimmed for a *non-hook* caller, would pin the behaviour rather than the characters.
- Same file, minor: `_frontend_minimum()` reads `_GATE_SOURCE` unguarded, so if the frontend tree is absent from the checkout the equality test raises `FileNotFoundError` rather than skipping. The precedent in `test_indexes.py` makes this acceptable, but a `pytest.mark.skipif(not _GATE_SOURCE.exists())` on the equality/trim tests (keeping the positive control as the loud one) would fail more legibly.

---

perrozzi:

Since the last review the author replaced the character-level substring assertion with real behavioural tests on the request URL (`frontend/src/api/client.test.ts`), softened/expanded the `searchFeedback` comment to state exactly what the boundary trim does and does not buy, added `skipif(not _GATE_SOURCE.exists())` to the lockstep comparison, and added a new CDK test coupling `validate_days`' ceiling to the metrics Lambda timeout and the API Gateway integration limit. All three of my previous new findings are addressed, and the day-fan-out concern is now pinned in CI.

### Resolved
- Trim behaviour is pinned on the request URL rather than on source text; both mutations (drop the trim) fail the new tests (`client.test.ts`).
- Formatting-sensitive `'q: params.q.trim()'` substring assertion removed from `test_search_minimum_lockstep.py`.
- Equality/mirror test now skips cleanly when the frontend tree is absent; positive control kept loud.
- The window-ceiling ↔ metrics-timeout coupling is now asserted (`lib/stacks/api-stack.test.ts`), with `validate_days`' max read from source rather than restated — though note it pins a *projection* (`15 ms/day`, hand-measured), not a measurement, so it guards config drift, not real latency.
- The `client.ts` comment no longer overclaims "every caller agrees by construction".

### Still open
- MCP `query` declares `minLength: 2` while the route still supports a present-but-blank `q`, so `query: ""` is schema-invalid for supported behaviour.

### New findings
- `voc-datalake/lib/stacks/api-stack.test.ts` — the second test's name and comment say the budget must stay "inside what API Gateway will actually wait for", but the assertion is `metricsTimeout() <= API_GATEWAY_INTEGRATION_CEILING_SECONDS + 1`, i.e. it *permits* 30 s > 29 s — exactly the "trade a 502 for a slower 504" case the comment says it exists to prevent. The `+1` is an unexplained fudge to accommodate the current 30 s config. Either assert `<= 29` (and adjust `MetricsApi`), or drop the pretence and assert only the worst-case projection, documenting why the Lambda deliberately outlives the integration.
- `voc-datalake/lib/stacks/api-stack.test.ts` (minor) — `maxWindowDays()`'s two-step regex (`def validate_days(...) -> int:` then `max_val: int = (\d+)`) will silently stop matching if the ceiling is ever expressed as a module constant or renamed; the `toBeDefined()` guards make that a loud failure rather than a false pass, so this is acceptable, but a reference to `SEARCH_QUERY_MIN_LENGTH`-style named constant in `shared/api.py` would be more durable.
- `voc-datalake/lambda/shared/test/test_search_minimum_lockstep.py` (nit) — the new class docstring says the control "must still fail in a normal checkout" and is "not skipped on the file's contents, only on its absence", but the control is not skipped at all; it asserts existence. Reword so the next reader isn't hunting for a skip marker that doesn't exist.
- `voc-datalake/frontend/src/api/client.ts` (nit) — the comment now narrates the review conversation ("which was the reviewer's alternative"). Review context doesn't age well in source; state the rationale impersonally ("a too-short term is deliberately not short-circuited into `count: 0`, because …").

---

perrozzi:

Since my last review the author renamed the API Gateway test to assert only the latency projection (dropping the `+1` fudge and explaining why the 30 s Lambda deliberately outlives the 29 s integration), replaced the two-step `validate_days` signature regex with a new named `MAX_FEEDBACK_WINDOW_DAYS` constant in `shared/api.py`, reworded the lockstep docstring and the `client.ts` comment, and documented the deliberate schema-stricter-than-server asymmetry in `mcp_handler.py`. All previous findings are addressed; the remaining items are nits.

### Resolved
- `lib/stacks/api-stack.test.ts` — the misleading `metricsTimeout() <= CEILING + 1` assertion is gone; the test now asserts only what must hold and records the chosen 30 s > 29 s asymmetry.
- `lib/stacks/api-stack.test.ts` — window ceiling now read from `MAX_FEEDBACK_WINDOW_DAYS`, a stable named declaration, instead of walking into a signature default.
- `test_search_minimum_lockstep.py` — docstring no longer implies the control carries a skip marker.
- `frontend/src/api/client.ts` — comment states the rationale impersonally; no review narration.
- MCP `minLength: 2` vs. supported blank `q`: not changed, but now explicitly documented as a deliberate strict-declaration/lenient-server asymmetry with the direction-of-harm argument. Acceptable as a documented decision.

### New findings
- `voc-datalake/lambda/api/mcp_handler.py` (minor) — the new comment says the blank-query tolerance is "pinned by `test_a_blank_query_is_not_a_search`", but no test of that name exists in this PR; the closest are `test_a_blank_query_is_still_a_successful_empty_answer` (metrics) and the unnamed filter-only path in `TestSearchReportsItsTruncation`. Cite the real test name, or the comment's claim that "the tolerance is tested rather than incidental" is unverifiable and will read as stale on the first grep.
- `voc-datalake/lambda/shared/api.py` (minor) — now that `MAX_FEEDBACK_WINDOW_DAYS` exists, `_DAYS_ARG`'s `maximum` in `mcp_handler.py` is still the literal `365` with the ceiling only mentioned in prose ("`validate_days` bounds to 365"). That is exactly the declaration/enforcement drift this PR fixes for `SEARCH_QUERY_MIN_LENGTH`; importing the constant there would close it in the same shape.
- `voc-datalake/lib/stacks/api-stack.test.ts` (nit) — with the config assertion removed, the second test reduces to `365 * 15 / 1000 < 29`, arithmetic over two constants in the same file plus one parsed value. It only fails if the window ceiling grows ~5×, and no longer touches infrastructure at all; consider folding it into the first test's message rather than presenting it as a separate guard.

---

perrozzi:

Since the last review the author cited the full test path for the blank-query tolerance, imported `MAX_FEEDBACK_WINDOW_DAYS` into `_DAYS_ARG.maximum`, and folded the API Gateway ceiling assertion back into the single latency test. All three previous findings are addressed; nothing functional changed.

### Resolved
- `mcp_handler.py` — blank-query comment now names `test_mcp_delegation.py::TestRouteSelection::test_a_blank_query_is_not_a_search` instead of an unqualified name.
- `mcp_handler.py` — `_DAYS_ARG.maximum` reads `MAX_FEEDBACK_WINDOW_DAYS` rather than the literal `365`, closing the declaration/enforcement drift I flagged (value unchanged, so the output fingerprint is unaffected).
- `lib/stacks/api-stack.test.ts` — the arithmetic-only second test is gone; both bounds now assert on the one projected number.

### New findings
- `voc-datalake/lib/stacks/api-stack.test.ts` (nit) — the surviving comment block is now a bodyless trailer inside the `describe`, and both it and the in-test comment narrate the review history ("A separate test … was split out at first", "An earlier version asserted `metricsTimeout() <= CEILING + 1`"). Same note I made on `client.ts`: state the rationale impersonally (30 s Lambda > 29 s integration is chosen, and why) without the changelog. The `mcp_handler.py` parenthetical "(a pre-existing test, which is why it appears in no diff that adds this comment)" is the same shape.
- `voc-datalake/lambda/api/mcp_handler.py` (nit) — the cited test is outside this diff, so I can't confirm it exists; if `TestRouteSelection::test_a_blank_query_is_not_a_search` is not the actual name in the repo today the comment is still unverifiable. Worth a grep before merge.

Otherwise the delta is clean.

---

perrozzi:

Comment-only delta: the three narrative-history comments I flagged (`client.test.ts`, `test_search_minimum_lockstep.py`, `api-stack.test.ts`) were rewritten impersonally, the bodyless trailing comment block was folded into the `describe` header as forward-looking rationale, and the `mcp_handler.py` parenthetical was dropped. No behaviour changed; both previous nits are addressed.

### Resolved
- `client.test.ts` / `test_search_minimum_lockstep.py` — "used to assert / A sibling guard used to…" changelog prose replaced with the impersonal rule.
- `api-stack.test.ts` — bodyless trailer removed; the 30 s > 29 s asymmetry is now stated as a choice in the `describe` header rather than as review history.
- `mcp_handler.py` — "(a pre-existing test, which is why it appears in no diff…)" parenthetical gone.

### Still open
- `mcp_handler.py` — the cited `TestRouteSelection::test_a_blank_query_is_not_a_search` is still outside this diff; worth a grep before merge to confirm the name.

### New findings
- `voc-datalake/lambda/api/mcp_handler.py` (nit) — the reworded citation now splits the qualified name across two separately-backticked fragments (`` `test_mcp_delegation.py::TestRouteSelection` `` / `` `::test_a_blank_query_is_not_a_search` ``), which reads as two identifiers rather than one path; close the backticks once around the whole thing.
- `voc-datalake/lib/stacks/api-stack.test.ts` (nit) — stray blank line left before the closing `});` of the `describe` where the removed comment block used to sit.

Otherwise the delta is clean.

---

perrozzi:

Comment-only delta again: the split-backtick citation in `mcp_handler.py` is now a single qualified name, and the stray blank line before the `describe` close in `api-stack.test.ts` is gone. Both nits from the last round are addressed; no behaviour changed.

### Resolved
- `voc-datalake/lambda/api/mcp_handler.py` — test citation is now one backticked path (`test_mcp_delegation.py::TestRouteSelection::test_a_blank_query_is_not_a_search`).
- `voc-datalake/lib/stacks/api-stack.test.ts` — trailing blank line inside the `describe` removed.

### Still open
- `voc-datalake/lambda/api/mcp_handler.py` — the cited `TestRouteSelection::test_a_blank_query_is_not_a_search` remains outside this PR's diff; grep to confirm the name before merge.

### New findings
- `voc-datalake/lambda/api/mcp_handler.py` (nit, optional) — the merged citation line is now ~105 chars; if the ruff config for this file enforces a line length below that, this would be a new violation. Worth confirming with `.venv/bin/python -m ruff` since the PR claims a zero delta.

Otherwise the delta is clean.

---

PR #363, opened 2026-08-21
fix(scripts): report every check leg instead of stopping at the first failure

perrozzi:
## Summary

The root aggregate scripts chained their legs with `&&`, so the earliest failing leg short-circuited the run and masked every later one. A contributor had to re-run `npm run check` once per problem to discover them all.

The three aggregate scripts now delegate to a small shell helper that runs each named leg in sequence, always runs all of them, and exits non-zero if any leg failed:

- `lint` → `lint:frontend lint:stream lint:python`
- `typecheck:all` → `typecheck typecheck:cdk typecheck:stream`
- `check` → `lint typecheck:all test test:cdk test:stream test:backend`

The runner is `voc-datalake/scripts/run-legs.sh`, a POSIX `sh` script with no new dependency (it started as an inline one-liner in `package.json`; review asked for a file, following the `build-layers.sh` precedent). Per leg it prints a `>>> npm run <leg>` banner before and a `<<< <leg>: PASS` / `<<< <leg>: FAIL (exit N)` marker after, so the interleaved output of six tools stays attributable. On failure it writes a `Failed legs: ...` summary to stderr; on success, `All legs passed: ...`.

⚠️ Invoked WITHOUT `cd voc-datalake`, unlike `build:layers`: the legs are root npm scripts, so `cd`-ing first would resolve `lint:frontend` against `voc-datalake/package.json`. The script header says so.

Semantics preserved:
- Still exits `0` only when every leg passes, so `deploy:all` (`npm run check && ...`) still gates deployment exactly as before. On failure it exits with the **first failing leg's own status** rather than a flat `1` (review point): `&&` propagated the status of the leg that stopped the chain, so anything distinguishing ruff's `2` from a `1` reads the same number as before.
- The nested case composes correctly: `check` calls `lint`/`typecheck:all`, which are themselves leg runners, and an inner failure propagates outward while inner siblings still run. A nested run suppresses its own summary (`VOC_LEGS_NESTED`), so exactly one `Failed legs:` line is authoritative.
- No arguments leak into the legs (verified a leg sees an empty `argv` tail), so `test` keeps its existing `-- --run` behaviour.

### Before vs after

Before, on this checkout, `npm run check` stopped at the first red leg and reported only that. After, the same tree reports all of them.

Real output, from `npm run lint` in a fresh worktree (no `node_modules`, no venv — hence the `127`s), rather than a sketch:

```
>>> npm run lint:frontend
<<< lint:frontend: FAIL (exit 127)

>>> npm run lint:stream
<<< lint:stream: FAIL (exit 127)

>>> npm run lint:python
<<< lint:python: FAIL (exit 127)

Failed legs: lint:frontend lint:stream lint:python
```

Legs after the first red one were previously invisible behind it — `test:cdk` and `test:backend` behind `lint:python`, for instance. Any such failures are pre-existing on `development` and untouched by this change; surfacing them is the point of the PR.

## Issue

No issue number was provided with the task. The original request was scoped to "Root package.json scripts only"; review correctly rejected that scope, so this now also carries the extracted script, its test, and the doc/steering updates the change made stale.

## Build and test results

Commands run from the repository root (after `npm install --prefix` for `voc-datalake`, `voc-datalake/frontend` and `voc-datalake/lambda/stream`, which the container did not have installed):

| Command | Result | Notes |
| --- | --- | --- |
| `mise run build` | **n/a** | `mise ERROR no tasks defined` — this repo has no mise tasks. |
| `npm run build` | **FAIL (environment)** | Vite dies with `EMFILE: too many open files`; the container's `ulimit -n` is capped at 1024 and cannot be raised. Unrelated to this change, which does not touch `build`. |
| `npm run typecheck:all` | **PASS**, exit 0 | All three legs pass — confirms the all-green path still exits 0. |
| `npm run lint` | **FAIL**, exit 1 | `lint:frontend` PASS, `lint:stream` PASS, `lint:python` FAIL (pre-existing ruff findings on `development`). All three legs reported. |
| `npm run check` | **FAIL**, exit 1 | Reports `lint` FAIL, `typecheck:all` PASS, `test` PASS, `test:cdk` FAIL, `test:stream` PASS, `test:backend` FAIL. |
| `node -e "JSON.parse(...)"` | **PASS** | `package.json` is valid JSON after the edit. |

Because the repo's own gates are red on `development` for unrelated reasons, I also validated the runner in isolation with a scratch `package.json` containing synthetic `ok`/`bad` legs:

- `ok1 bad ok2` → all three legs ran, `ok2` executed after `bad` failed, summary listed `bad`, exit **1**.
- `ok1 ok2` → both ran, "All legs passed", exit **0**.
- single leg → leg receives an empty argument list (no `--` leakage), exit **0**.

## Decisions made

- **Inline `sh` helper rather than a script file.** The task restricted the change to root `package.json` scripts, so no `scripts/` file and no `npm-run-all`/`concurrently` dependency was added. The trade-off is one dense line; the alternative would have been out of scope.
- **Sequential, not parallel.** Running legs concurrently would also report everything, but it interleaves output from six tools unreadably and changes resource usage. Sequential keeps the existing behaviour and cost profile, and the per-leg banners give the attribution.
- **`lint` and `typecheck:all` converted too, not just `check`.** They have the same masking problem, and leaving them chained would mean `check` still only showed the first lint error. This is what makes the `lint:python`-behind-`lint:frontend` case visible.
- **Kept the docs as-is.** `docs/deployment.md` and `.kiro/steering/deployment.md` describe `check` as "chained with `&&`", which is now stale, but both files are outside the stated scope of this change. Worth a follow-up.

## Agent notes

**What went well.** The change is tiny and self-contained, and the repo's pre-existing red legs turned out to be an ideal live fixture: `check` genuinely had three independent failures stacked behind one another, so the before/after difference is demonstrable rather than hypothetical.

**What was difficult.** Verifying an exit-code contract in a tree where the real gates are already red is awkward — a `1` proves little on its own. I resolved it by testing the runner against synthetic pass/fail legs so both the 0 and 1 paths are covered, then confirming the real `typecheck:all` still exits 0 when genuinely green. Quoting also needed care: the helper lives inside a JSON string inside an `npm` script, so `$@`/`$*` and the trailing `--` (which stops `sh` from eating the first leg name as `$0`) all had to be right. The container fought back a bit too — the npm cache sits on a full 1 GB NFS mount (worked around with `npm_config_cache=/tmp/npmcache`) and the fd limit is hard-capped at 1024, which is what breaks the Vite build.

**Repo conventions discovered.**
- Root `package.json` is a thin orchestration layer; every real leg delegates via `--prefix voc-datalake`, `--prefix voc-datalake/frontend`, or `--prefix voc-datalake/lambda/stream`.
- There are four independent test suites (frontend Vitest, CDK Vitest, stream Vitest, Python pytest) and Python tooling expects the checked-in venv at `voc-datalake/.venv/bin/python`.
- `npm run lint:python` and `npm run test:backend` / `test:cdk` are red on `development` independently of any change, so scoped per-file runs are the practical gate.
- Commits and PR titles use conventional commits with a required scope, often with a trailing `(#NNN)`.
- `mise run build` does not exist here; the build entry point is `npm run build`.

- Update the two `deployment.md` tables that still say `check` is "chained with `&&`" — deliberately left alone here to respect the stated scope.
- Fix or re-baseline the ~550 pre-existing ruff findings (mostly `I001` import sorting). Now that `check` reports every leg, a permanently-red `lint:python` is the main thing still degrading the signal.
- Consider `printf`-ing the summary through `npm --silent` or trimming npm's own `notice run` lines; they add noise around the per-leg banners.
- Per-leg timing would be a natural addition now that the runner is a real script.


perrozzi:

This PR converts the three root aggregate scripts (`lint`, `typecheck:all`, `check`) from `&&` chains to a new inline `run:legs` POSIX-`sh` runner so every leg runs and all failures are reported, with the aggregate still exiting non-zero if any leg failed. The mechanics look correct to me: `sh -c '…' --` makes `--` the `$0` placeholder so npm's appended leg names land in `$1..$n`, `for leg in "$@"` iterates them in dash, no single quotes appear inside the single-quoted script body, and the `test` leg keeps its `-- --run` tail because legs are invoked as bare `npm run "$leg"`. I found no blockers; the remaining points are maintainability/consistency and stale docs.

### Should fix

- `package.json` — the docs are now factually wrong. `docs/deployment.md` and `.kiro/steering/deployment.md` describe `check` as "chained with `&&`" and fail-fast. The steering file in particular is agent-facing guidance, so leaving it stale actively misleads future contributors/agents. This is a two-line edit; the self-imposed "root `package.json` only" scope isn't a reason to ship a diff that contradicts checked-in documentation.
- `package.json` — a ~500-character `sh` one-liner embedded in a JSON string is hard to read, hard to diff, and hard to extend. The repo already has a precedent for shell helpers (`voc-datalake/scripts/build-layers.sh`, invoked from `build:layers`), so `scripts/run-legs.sh` would be more consistent with existing conventions than an inline expression, without adding a dependency.
- `package.json` — exit-code fidelity is lost: `&&` propagated the failing tool's actual status, whereas the runner always exits `1`. `deploy:all` only checks non-zero so gating is preserved, but if anything (CI job, wrapper script) distinguishes e.g. ruff's `2` from `1`, that signal is gone. Worth either preserving the last non-zero status or noting the change explicitly.
- CI cost: `check` no longer short-circuits, so a PR that fails `lint:python` now also runs three Vitest suites and pytest before failing. That's the intent, but if CI invokes `npm run check` on every push it's a real wall-clock/runner-minutes increase — please confirm the CI workflow is happy with that, or keep the CI entrypoint fail-fast and use the leg runner locally.
- No test covers the new behaviour. The repo convention is that a bug fix ships a regression test that fails when reverted; here the verification is manual notes plus a scratch `package.json`. A small vitest case in the CDK suite that shells out to the runner with synthetic pass/fail legs and asserts (a) all legs ran and (b) exit code, would lock in both the `0` and `1` paths cheaply.

### Nits

- `sh -c` hard-codes a POSIX shell. On Windows with npm's default `script-shell` of `cmd.exe`, all three aggregate scripts break rather than just the already-POSIX-only legs (`.venv/bin/python`, `cd … && …`). Low impact given the repo is already POSIX-leaning, but it removes any chance of these three working on Windows.
- `run:legs` now shows up in `npm run` output as a first-class task even though it's an internal helper and silently exits `0` with "All legs passed:" when given no legs. Consider naming it `_run:legs` (or documenting it) so it isn't mistaken for a user-facing command.
- The "Before vs after" block in the PR description isn't the format the code actually prints (the implementation emits `>>> npm run lint:frontend` / `<<< lint:frontend: PASS`, unindented, with no combined `typecheck / typecheck:cdk / typecheck:stream` line). Illustrative output in a PR body is fine, but it should match reality or be labelled as a sketch.
- Nested runners double the banners for `check` (`>>> npm run lint` … inner `Failed legs: …` on stderr … `<<< lint: FAIL`), which slightly dilutes the summary the PR is trying to make prominent. Suppressing the inner summary when a `VOC_LEGS_NESTED`-style env var is set would keep one authoritative "Failed legs" line.
- Worth calling out in the PR body that `npm run check` is red on `development` for pre-existing reasons; the all-green path for `check` (as opposed to `typecheck:all`) was never exercised end-to-end here.

---

perrozzi:
## Disposition — review at `765bfee8`

All five should-fix items addressed; two of them turned out to be answerable with evidence rather than code. Three nits taken, one retired by construction, one answered.

### Should fix

**1. The docs now contradict the code. Taken — and it was worse than one place.** `docs/deployment.md` had it twice (the table row and the "first failure hides every later step" bullet) and `.kiro/steering/deployment.md` three times (the inline comment on the `check` line, its table row, and its own trap bullet). All five updated. You were right that the steering file matters most: it is loaded into agent context automatically, so a stale claim there is not passive documentation, it is an instruction. The scope argument was wrong and I have dropped it — a diff that leaves checked-in docs contradicting the code is not "narrow", it is incomplete.

**2. The ~500-character `sh` one-liner. Taken — extracted to `voc-datalake/scripts/run-legs.sh`**, following the `build-layers.sh` precedent you pointed at. Now readable, diffable and testable. One deliberate difference from `build:layers`: it is invoked as `sh voc-datalake/scripts/run-legs.sh …` **without** `cd voc-datalake`, because the legs are root npm scripts — `cd`-ing first would resolve `lint:frontend` against `voc-datalake/package.json`. That constraint is stated at the top of the script so the next person does not "fix" it into the build-layers shape.

**3. Exit-code fidelity. Taken.** The runner now exits with the **first** failing leg's own status rather than a flat `1`. First rather than last, deliberately: `&&` propagated the status of the leg that stopped the chain, so anything distinguishing ruff's `2` from a `1` reads the same number it read before. Later legs still run — that is the point — they just do not overwrite the answer. Pinned by `exits with the first failing leg's own status, not a flattened 1`, and demonstrated end to end below.

**4. CI cost. Answered with evidence: no CI job runs `check`, so there is no runner-minute increase.** This repo has exactly one workflow, `.github/workflows/no-event-logging-guard.yml`, and it runs a single test file with plain `python -m pytest lambda/api/test/test_no_event_logging.py` — no npm, no `check`, no `lint`. The only mention of `npm run check` in that file is a comment explaining that the real gate lives in `package.json` and is *not* wired to CI. So the extra wall-clock lands only on a developer who chose to run the full gate locally, which is where wanting all the results is the point.

**5. No test covered the new behaviour. Taken**, as the vitest-with-synthetic-legs case you described: `voc-datalake/lib/tooling/run-legs.test.ts`, 7 cases. It copies the script into a sandbox at the same relative path and drives it with fake legs (`ok`, `bad`, `bad:ruff` exiting `2`), so it exercises the runner and never the real suites. It lives under `lib/` because `vitest.config.ts` includes exactly `lib/**/*.test.ts` — that is the only path `test:cdk` collects — and the file says so, since a shell-script test under `lib/` otherwise looks misfiled.

Mutation-checked by running each revert, not predicting it:

| Revert | Fails |
|---|---|
| hardcode `exit 1` | `exits with the first failing leg's own status` |
| `break` after a failure (restore `&&`) | `runs every leg even after one fails` **+** `reports the failing legs together on stderr` |
| remove the `VOC_LEGS_NESTED` guard | `prints no summary when nested` |
| drop the empty-args guard | `refuses a run with no legs` |

### Nits

- **`run:legs` appearing as a first-class task, and its silent success on no legs. Retired by construction rather than renamed:** the npm script is **gone**. The three aggregates call the shell script directly, so there is no internal task to mistake for a user-facing one, and one less layer of `npm run` indirection. A run with **no legs is now an error** (`exit 2`, "no legs given") instead of reporting success.
- **Nested double-banners. Taken.** A nested run (`check` → `lint`, `typecheck:all`) sets `VOC_LEGS_NESTED=1` and prints no summary, so exactly one authoritative `Failed legs:` line survives. Per-leg PASS/FAIL lines are kept — those are the diagnosis, not the noise.
- **PR description output block didn't match reality. Fixed** — the body now shows what the code prints (`>>> npm run lint:frontend` / `<<< lint:frontend: PASS (exit …)` on failure), unindented, with no invented combined line.
- **POSIX `sh` on Windows.** Acknowledged and unchanged: the legs themselves are already POSIX-only (`.venv/bin/python`, `cd … && …`), so these aggregates never worked under `cmd.exe`. Now stated in the script header rather than implied.

### Verification

- `npx vitest run lib/tooling/run-legs.test.ts` — **7 passed**.
- Full CDK suite `npx vitest run` — **286 passed, 18 files**. ⚠️ It reports 47 failures unless `voc-datalake/frontend/dist/` exists (`CannotFindAsset`); that is a pre-existing property of the suite, not this change — creating the directory turns the same tree green.
- `npx tsc --noEmit` (CDK, includes the new test) — clean.
- **End-to-end on the real aggregate**, which is also the demonstration this PR is for: `npm run lint` in a fresh worktree ran **all three** legs and exited `127`.

```
>>> npm run lint:frontend
<<< lint:frontend: FAIL (exit 127)
>>> npm run lint:stream
<<< lint:stream: FAIL (exit 127)
>>> npm run lint:python
<<< lint:python: FAIL (exit 127)

Failed legs: lint:frontend lint:stream lint:python
```

Under `&&` that run would have stopped after the first line and said nothing about the other two. (The `127`s are honest: no `node_modules` or venv in a fresh worktree — and the exit code is the *first* leg's, which is fidelity holding on real input.)

### Still not addressed, deliberately

`check` has **no `build` leg** and does not include `typecheck:tests`, so it remains an incomplete gate in a second, unrelated way. Out of scope here — this PR changes how the legs are *run*, not which legs exist — but worth its own change, since "check passed" currently says nothing about whether the frontend builds.

perrozzi:

Since my last pass the author extracted the one-liner into `voc-datalake/scripts/run-legs.sh`, updated both `deployment.md` files, added a regression test (`voc-datalake/lib/tooling/run-legs.test.ts`), made the exit status the first failing leg's real status, added a `VOC_LEGS_NESTED` guard for nested summaries, and made a no-leg invocation an error. That's essentially every should-fix from the previous review except the CI-cost confirmation.

### Resolved
- Stale docs: both `docs/deployment.md` and `.kiro/steering/deployment.md` now describe the every-leg behaviour instead of the `&&` chain.
- Inline ~500-char JSON string replaced by `voc-datalake/scripts/run-legs.sh`, consistent with the `build-layers.sh` precedent.
- Exit-code fidelity restored (first failing leg's own status), with a test pinning ruff's `2`.
- Regression test added covering all-pass, continue-after-failure, non-zero exit, stderr summary, nesting, and empty args.
- Nested double-summary nit fixed via `VOC_LEGS_NESTED`; the user-visible `run:legs` script is gone.

### Still open
- No confirmation that the CI entrypoint is happy losing short-circuit behaviour on `check` (extra runner minutes when an early leg fails).
- POSIX-`sh`-only nit remains (now an explicit `sh …` invocation), documented in the script header.

### New findings

**Should fix — `voc-datalake/lib/tooling/run-legs.test.ts`: the sandbox `npm run` inherits the parent npm's config env, so the test may fail under the real gate.** The test is collected by `npm run test:cdk`, i.e. `npm run test --prefix voc-datalake`, which exports `npm_config_prefix` / `npm_config_local_prefix` (and `npm_config_*` generally) into the vitest process. `spawnSync` passes `{ ...process.env }` through, so the nested `npm run ok` inside the temp sandbox can resolve its `package.json` against the repo/`voc-datalake` prefix rather than `cwd`, yielding `Missing script: ok` and a non-zero status — the test would pass under bare `vitest` but fail via `npm run test:cdk`. Strip npm-injected config before spawning, e.g. build the env by deleting every `npm_config_*` / `npm_lifecycle_*` key (and keep `VOC_LEGS_NESTED` handling as-is), or invoke the leg runner with an explicit `--prefix "$sandbox"`. Worth verifying the test actually passes via `npm run test:cdk`, not just directly.

**Nit — `voc-datalake/scripts/run-legs.sh`:** if `VOC_LEGS_NESTED=1` happens to be exported in the caller's environment (outer harness, CI wrapper), the outermost run silently suppresses its own `Failed legs:` summary. Using a dedicated argument/flag, or unsetting the variable for the top-level invocation in `package.json`, avoids that.

**Nit —** the script must run with the repo root as cwd yet lives under `voc-datalake/scripts/`; a root-level `scripts/run-legs.sh` would match its actual contract better. The header comment mitigates it.

---

perrozzi:
### Background agent — FAILED

| Field | Value |
|-------|-------|
| Task  | `01M0K1SHM87BD81PTZA2XD057B` |
| Status | **FAILED** |
| Last event | `task_failed` @ 2026-08-21T20:56:59.464Z |

perrozzi:
## Convergence run — 🔴 Failed — did NOT converge

A leg of the loop failed. The pull request is left wherever the last push put it.

**Why it stopped:** review task FAILED

| | |
|---|---|
| Outcome | **FAILED** |
| Iteration | 0 of 5 |
| Loop spend | $0.00 of $100.00 |
| Goal | Close out the remaining review findings on this PR (the leg runner behind the root lint/typecheck:all/check aggregates). Round 1's five should-fix items are all addressed and confirmed resolved by round 2; what follows is round 2's list only. ## Should fix — and READ THIS FIRST, because the pr… |
| Run | [`01M0K1SEW27F2F0SMJ6XTBY8QC`](https://d1t2aabk0oo81s.cloudfront.net/runs/01M0K1SEW27F2F0SMJ6XTBY8QC) |

### Each iteration

|---|---|---|---|---|---|---|
| 0 | no verdict recorded | — | — | — | — | — |

PR #365, opened 2026-08-21
fix(metrics): report a partial aggregate window instead of asserting completeness

perrozzi:
## Summary

Every route in `metrics_handler.py` that publishes `is_partial` published a **hardcoded `False` on its aggregates path** — the default path, a plain `?days=N` with no `source` and the import date basis. The flag was initialised beside the scan branch that sets it, and the aggregates branch never touched it, so the value was an *assertion* of completeness rather than a measurement. Audit finding **M4**: 99 items of a 6,239-item corpus returned as a complete answer. The reviewer flagged `get_category_metrics`; the same lie was in five other response sites, so all six are wired here.

### The two independent causes, which is the part worth checking

They are not the same fault and neither implies the other, so each gets its own signal and its own tests:

**1. Paging truncation — already detected, then thrown away.** `_query_metric_window` follows the cursor under a bound of `days` pages and, on exhausting the bound with a cursor still open, logged a warning and returned a short list. Its docstring justified that with *"these endpoints have no `is_partial` signal … Nothing in the response shape can express that"* — false even when written, since six response sites in the same file publish the field. It now returns `(items, truncated)`, following the existing `_scan_recent_items` / `_scan_window_items` convention rather than inventing a second one, and the docstring says what the function does. The warning stays as the *operator's* copy of the fact; `truncated` is the caller's.

A third site of the same class came out of tracing it: `/metrics/sources`, `/metrics/personas` and `/feedback/entities` read `gsi1-by-metric-type` with **one unpaged query**, so DynamoDB's 1 MB page limit bounds how many aggregate rows they see. `_read_was_truncated` reports the open cursor. **Reported, not followed** — paging it would change which data the answer is computed from.

**2. The aggregate retention horizon — detected by nothing at all.** This is a property of the *request*, not of a read: it holds even when every query succeeds and returns every row that still exists. Aggregate rows carry a 90-day TTL while `validate_days` admits 365, so `?days=365` queries a sort-key range whose early part DynamoDB has already deleted — the read succeeds, the totals under-report, nothing notices. A window wider than the retention **cannot** be answered completely from aggregates. Declared as `AGGREGATE_RETENTION_DAYS` in `shared/api.py`, beside `MAX_FEEDBACK_WINDOW_DAYS`, which is the ceiling it is in tension with. Strictly `>`: a window *equal* to the retention is still fully covered, and crying partial over a complete answer teaches callers to ignore the flag.

Where an endpoint reads several partitions (one per category, one per sentiment label, three for the summary) truncation in **any** of them makes the answer partial, so the flag ORs across them. Where an endpoint can take either path, the aggregates truncation ORs into whatever the scan path already computed.

`aggregator/handler.py` is **untouched** as instructed. The link is a lockstep test that reads `ttl_days`' default via `inspect.signature` and fails if the aggregator's TTL and the shared constant drift — in either direction.

### Not done

No window is widened or narrowed, no read gained a fallback to scanning, and no stored aggregate is backfilled. `period_days` and the queried date ranges are byte-identical; `/metrics/summary` still costs exactly three queries at every window including 365. Finding M7 (persona dimension bucketing) is untouched.

### One deliberate step beyond the brief

The MCP layer forwards these route bodies **unprojected**, so the newly-honest flag reaches a client whether or not it is declared — and `additionalProperties` is *absent* (not `false`) from both metrics tools' output shapes, so an undeclared `is_partial` validates and is then invisible to a model reading the catalogue. Since the audit's whole argument is that this matters more through MCP than on the dashboard, leaving it undeclared would have shipped the fix and kept the symptom. So `get_metrics_summary` declares it, `get_metrics_breakdown`'s description is corrected (it said *"an aggregate read failed"*, naming one of the two causes and thereby teaching a reader the other could not happen), and `MCP_SERVER_VERSION` moves 3.6.0 → 3.7.0 with the catalogue fingerprint updated in the same commit, as that lockstep test demands. The frontend `MetricsSummary` doc comment and the Dashboard's "exceeded the scan budget" hint described only the scan cause too.

## Tests

New `voc-datalake/lambda/api/test/test_metrics_partial_window.py` (46 cases) and `test_aggregate_retention_lockstep.py` (3 cases). Each module docstring states which mutation makes its assertions fail, per repo convention.

The **route list is derived** from the handler source by AST walk (`@app.<verb>` decorator + a response dict carrying `is_partial`, scoped to each function's own body) rather than hand-copied, so a future endpoint that publishes the flag without wiring it fails instead of silently never being tested. `test_the_derivation_finds_the_routes_that_publish_the_flag` is its positive control — without it an empty derivation would make every parametrized case pass by never running.

Every truncation assertion has a **positive control** asserting `false` on the complementary input, so the suite cannot pass by the flag being unconditionally true — which is the same defect pointing the other way.

## Mutation checks — run, not predicted

Each revert was applied to the working tree, the full suite run, then reverted:

| Mutation | Result |
|---|---|
| `_query_metric_window`'s `return items, True` → `return items, False` | **7 failed**, incl. `TestAggregatePathReportsPagingTruncation::test_a_truncated_aggregate_read_is_reported_partial[/metrics/summary, /metrics/sentiment, /metrics/categories]`, `::test_the_helper_returns_the_flag_beside_the_items`, `::test_one_truncated_partition_out_of_many_makes_the_answer_partial` |
| `_window_exceeds_aggregate_retention` → `return False` | **8 failed**, incl. `TestWindowWiderThanAggregateRetention::test_a_window_beyond_the_retention_is_partial_by_construction` on all **six** routes, `::test_the_predicate_is_about_the_request_not_about_a_read` |
| aggregator `ttl_days: int = 90` → `30` | **2 failed**: `TestAggregateRetentionLockstep::test_the_shared_constant_equals_the_aggregators_ttl_default`, `::test_both_writers_stamp_the_same_horizon` |
| `AGGREGATE_RETENTION_DAYS = 90` → `180` (the other drift direction) | same 2 failed |
| added a `/metrics/unwired` route publishing `is_partial` without wiring it | **3 failed**: the derivation control plus both parametrized suites picking the new route up |
| deleted `is_partial` from `get_metrics_summary`'s `outputSchema` | **3 failed**: `TestMetricsToolsDeclareTheFlagTheyPassThrough` ×2 plus the fingerprint lockstep |

## Build and test results

| Command | Result |
|---|---|
| `mise run build` | **not available** — `mise ERROR no tasks defined in <repo>`. No mise tasks exist in this repo (a known gap; the setup report also flagged build gating as inert). Ran the project's real build instead. |
| `npm run build` (frontend, the repo's `build` script) | **pass**, built in 8.82s |
| `pytest voc-datalake/lambda` | **3169 passed, 0 failed** (base tip 3110; +59 = 49 new cases plus 10 pre-existing that only run once `pydantic`/`aws-xray-sdk` are installed — see below) |
| `ruff check --select F,E4,E7,E9` on all changed Python files | **All checks passed!** |
| `ruff check lambda plugins` (repo-wide gate) | **508** — one *below* the pre-existing 509, not above |
| `npm run typecheck` (frontend) | **pass** |
| `npm run typecheck:cdk`, `typecheck:stream` | **pass** |
| `npx vitest run` (frontend) | **185 files, 3059 passed** |
| `npm test` (CDK) | **279 passed** — see note |
| `npm test` (lambda/stream) | **23 files, 365 passed** |
| `npm run lint` (frontend eslint, `--max-warnings 0`) | **pass** |

**Two environment notes, neither caused by this change:**

- The base suite reported 3110 on a checkout whose venv was missing `pydantic`, `aws-xray-sdk` and the layer deps — with those installed the *unmodified* base collects and passes 3112. My 3169 is measured against that same complete environment, so the delta is +57 from my 49 new cases and the collection differences, not a hidden regression. `aws-xray-sdk` is absent from `requirements-dev.txt` while `aws-lambda-powertools`' tracer imports it; worth a follow-up.
- The CDK suite fails 47/256 on a fresh checkout with `CannotFindAsset: Cannot find asset at voc-datalake/frontend/dist`. That is the missing frontend build, not a code fault: after `npm run build` all 279 pass. Verified pre-existing by stashing my changes.

## Decisions made

- **`_read_was_truncated` for the unpaged index reads.** Not named in the brief, but it is the same fact ("rows exist that were not counted") on three of the six routes being fixed, discarded the same way. Wiring the paging flag while leaving these silent would have left `/metrics/sources` and `/metrics/personas` still able to report a 1 MB-truncated read as complete. Reported only — no paging added.
- **`>` not `>=`** for the retention boundary, with a test on the exact-90 case pinning it.
- **`/feedback/entities`' `issues` map excluded** from the flag. It is a deliberate sample on *both* branches (newest rows of at most 7 days, capped at `limit`); folding it in would make the flag true on nearly every call, which is how a flag stops meaning anything. `is_partial` there describes the counts, matching what the scan branch's flag describes.
- **`test_the_retention_is_narrower_than_the_widest_requestable_window`** asserts the premise, so if retention ever reaches 365 the horizon check becomes provably dead code and fails loudly rather than sitting there as decoration.
- The `TestQueryMetricWindow` docstrings that asserted "these endpoints have no `is_partial` signal" were updated, since that claim is now the opposite of true.

## Agent notes

**What went well.** The brief's insistence that the two causes are independent turned out to be load-bearing for the *code*, not just the review: they need different mechanisms (one per-read and ORed, one computed once from `days`), and a single flag variable seeded by the horizon check and then ORed with each read's truncation is what lets both share one field honestly. Deriving the route list by AST instead of listing it caught a real gap — I'd initially wired only the `/metrics/*` four and the derivation immediately failed on `/feedback/entities`, which publishes the flag from its scan branch and had the same silent aggregates branch.

**What was difficult.** Three things. (1) The MCP boundary is where the audit says this defect actually bites, but touching it means moving `MCP_SERVER_VERSION` and a catalogue fingerprint — I nearly stopped at the handler, which would have shipped a fix whose symptom survived at the surface that matters. (2) Stubbing per-partition truncation needed reading the `pk` back out of a boto3 `Condition`; a `repr()`-based match silently matched nothing (object addresses) and the test passed for the wrong reason until I checked. Reaching through `get_expression()` is the honest way. (3) Getting the test environment green at all: the venv was absent and `pytest` failed at import on `aws_xray_sdk` and `pydantic` before any test ran.

**Repo conventions discovered/confirmed.** Lockstep tests are the idiom for any value owned in one place and consumed in another, and they read the other side's source (`inspect.signature`, regex over TypeScript, JSON schema files) rather than restating it — the point being that a restated copy is the thing that goes stale. Comments carry the *argument*, not a description, and load-bearing ones say why an alternative is wrong. Tests are named as outcomes and module docstrings name the mutation each assertion catches. Positive controls are expected wherever a check could pass by never running. `MCP_TOOLS` is fingerprinted against `MCP_SERVER_VERSION`, so any published-shape edit must bump both in one commit, and the version comment block wants a paragraph arguing major-vs-minor.

**Suggestions for future tasks here.** (a) `mise run build` still does not exist — worth adding, since the harness gates on it and currently cannot. (b) `aws-xray-sdk` belongs in `requirements-dev.txt`; without it `pytest` fails at collection. (c) The CDK suite depends on `frontend/dist` existing, so `npm run build` must precede `npm run test:cdk` — a note in the README's test section would save the next agent the same detour. (d) `ruff check lambda plugins` sits at ~508 pre-existing findings; treating that as a ratchet (assert it never rises) would be a cheap improvement. (e) `/metrics/sources` and `/metrics/personas` still read their index without paging, so a large enough deployment silently answers from one page — now at least *reported*, but the underlying read is worth fixing on its own.

---

## Review convergence (commits 3–6)

The automated review's blocker and all six should-fix items are addressed, plus the nits worth taking. What changed since the first two commits:

**Blocker — the hint was hardcoded English.** `partialCountsHint` now exists in all eight locale files and `MetricsGrid` renders `t('partialCountsHint')`. It went in `common`, beside the `partialWindowHint` the Categories results line already uses, rather than in `dashboard`: the string is about counts being a lower bound, not about the dashboard, and it keeps this file to one namespace — `scripts/i18n-check.mjs` infers a single namespace per file from the first `useTranslation`, so a `dashboard` hook in `MetricsGrid` would have made it misreport this file's existing `exportPdf` keys as missing. `Dashboard.test.tsx` forwards `hint` through its `MetricCard` mock, which was swallowing it, and asserts the copy read from `en/common.json` with a positive control on the complete-window case.

**Every `_query_metric_window` call site is now checked structurally.** `TestEveryCallerUnpacksTheTruncationFlag` asserts by AST that every call unpacks two values, and scans `lambda/` to assert nothing outside `metrics_handler.py` calls the helper at all. This is the case the parametrized suites cannot reach: they only see routes that *publish* the flag, and the failure of a non-publishing caller is not a crash — `items = helper(...)` silently answers `len == 2`.

**`is_partial` is representable everywhere it is emitted.** `SentimentBreakdown`, `CategoryBreakdown`, `SourceBreakdown`, `EntitiesResponse` and a new named `PersonaBreakdown` (replacing an inline generic at the `getPersonas` call site — where a field goes to be invisible) all declare it, pointing at the one canonical statement on `MetricsSummary.is_partial` instead of repeating it. No runtime validator sits in front of these responses, so nothing strips the field: all six getters are plain `fetchApi<T>` with no Zod schema or normalizer, and `useSummaryQuery` adds none.

**MCP coverage, asked in the other direction.** `test_every_mcp_reachable_route_that_publishes_the_flag_has_a_tool_declaring_it` derives the publishing routes from the handler source and the reachable set from the live `DOMAIN_ROUTES`, so `/feedback/entities` being deliberately unexposed is a fact the suite knows rather than a fact a reader has to check. Adding any publishing route to `DOMAIN_ROUTES` now fails until whichever tool serves it declares the field. The declaration is deliberately **not** `required` on these two tools — they forward the body unprojected and fall back to `{}`, so promising a key they do not build would turn an honest empty answer into a schema violation — and that choice is argued at `_IS_PARTIAL_DESCRIPTION` and pinned by a test scoped to `is_partial` alone.

**The retention lockstep no longer imports the aggregator.** It parses `aggregator/handler.py` with `ast`: importing it from an api test executes another package's module scope (a DynamoDB resource, `AGGREGATES_TABLE`), so the test failed for an environmental reason in the same red as real drift. Parsing also made two new assertions possible — no call site overrides `ttl_days` (positional overrides included; a single `ttl_days=30` would leave the lockstep green while those rows expire two months early), and each writer actually reads the parameter rather than stamping a literal. Its scope (bare-name calls inside that one file) is stated so it is not overread.

**Nits taken:** `_read_was_truncated` → `_index_read_was_truncated`; the ~48h TTL deletion lag written into `_window_exceeds_aggregate_retention`'s docstring, with the note that the boundary is the guarantee and not the observed state; and the rationale collapsed from five copies to two canonical statements (`AGGREGATE_RETENTION_DAYS` for the backend, `MetricsSummary.is_partial` for the frontend) with pointers elsewhere. **Nit not taken:** the cost of the truncation cases. Measured rather than assumed — `test_metrics_partial_window.py` runs 49 tests in 1.3s, so shrinking `days` would trade a real invariant (the paging cases sit just inside the retention horizon, which is what makes them about paging and not about the horizon) for nothing measurable.

**Mutations, re-run after the narrowing commits.** Each applied to the working tree, full suite run, then reverted:

| Mutation | Result |
|---|---|
| a `_query_metric_window` call site stops unpacking the flag | caught |
| an aggregator call site passes `ttl_days=30` | caught |
| a writer takes `ttl_days` and stamps a literal instead | caught |
| a publishing route added to `DOMAIN_ROUTES` with no tool declaring the flag | caught |
| `required: ["is_partial"]` added to an unprojected pass-through tool | caught (plus the fingerprint lockstep) |
| the dashboard hint points at a key no locale defines | caught |
| the hint deleted from one locale | caught |

**Gate re-run on the final commit:** `pytest lambda` **3177 passed** · frontend `vitest run` **185 files, 3060 passed** · `npm run lint` (eslint `--max-warnings 0`) pass · `tsc --noEmit` frontend and CDK pass · `ruff check --select F,E4,E7,E9` on the changed Python clean · `ruff check lambda plugins` **508**, unchanged from base · locale parity 31 tests pass · `i18n:check` shows the same two pre-existing `projectDetail:prototypeLink.*` findings as the base and nothing new.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/common.json
+++ b/voc-datalake/frontend/public/locales/de/common.json
@@ -77,6 +77,7 @@
     },
     "settings": "Einstellungen"
   },
+  "partialCountsHint": "Näherungswert: Der gesamte Zeitraum konnte nicht gelesen werden, die Zahlen sind eine Untergrenze",
   "partialWindowHint": "Filter eingrenzen, um alle zu sehen",
   "problemAnalysisPage": {
     "emptyHint": "Erweitere den Zeitraum oder passe die Filter an",
--- a/voc-datalake/frontend/public/locales/en/common.json
+++ b/voc-datalake/frontend/public/locales/en/common.json
@@ -77,6 +77,7 @@
     },
     "settings": "Settings"
   },
+  "partialCountsHint": "Approximate: the full window could not be read, counts are a lower bound",
   "partialWindowHint": "narrow filters to see all",
   "problemAnalysisPage": {
     "emptyHint": "Try expanding the time range or adjusting filters",
--- a/voc-datalake/frontend/public/locales/es/common.json
+++ b/voc-datalake/frontend/public/locales/es/common.json
@@ -80,6 +80,7 @@
     },
     "settings": "Configuración"
   },
+  "partialCountsHint": "Aproximado: no se pudo leer el periodo completo, los recuentos son un límite inferior",
   "partialWindowHint": "acota los filtros para verlos todos",
   "problemAnalysisPage": {
     "emptyHint": "Prueba a ampliar el rango de tiempo o ajustar los filtros",
--- a/voc-datalake/frontend/public/locales/fr/common.json
+++ b/voc-datalake/frontend/public/locales/fr/common.json
@@ -80,6 +80,7 @@
     },
     "settings": "Paramètres"
   },
+  "partialCountsHint": "Approximatif : la période complète n'a pas pu être lue, les décomptes sont une borne inférieure",
   "partialWindowHint": "affinez les filtres pour tout voir",
   "problemAnalysisPage": {
     "emptyHint": "Essayez d'élargir la plage de temps ou d'ajuster les filtres",
--- a/voc-datalake/frontend/public/locales/ja/common.json
+++ b/voc-datalake/frontend/public/locales/ja/common.json
@@ -77,6 +77,7 @@
     },
     "settings": "設定"
   },
+  "partialCountsHint": "概算値: 期間全体を読み取れなかったため、件数は下限値です",
   "partialWindowHint": "すべて表示するにはフィルターを絞り込んでください",
   "problemAnalysisPage": {
     "emptyHint": "期間を広げるかフィルターを調整してください",
--- a/voc-datalake/frontend/public/locales/ko/common.json
+++ b/voc-datalake/frontend/public/locales/ko/common.json
@@ -77,6 +77,7 @@
     },
     "settings": "설정"
   },
+  "partialCountsHint": "근사치: 전체 기간을 읽을 수 없어 집계는 최소값입니다",
   "partialWindowHint": "모두 보려면 필터를 좁히세요",
   "problemAnalysisPage": {
     "emptyHint": "기간을 넓히거나 필터를 조정해 보세요",
--- a/voc-datalake/frontend/public/locales/pt/common.json
+++ b/voc-datalake/frontend/public/locales/pt/common.json
@@ -80,6 +80,7 @@
     },
     "settings": "Configurações"
   },
+  "partialCountsHint": "Aproximado: não foi possível ler o período completo, as contagens são um limite inferior",
   "partialWindowHint": "refine os filtros para ver tudo",
   "problemAnalysisPage": {
     "emptyHint": "Tente ampliar o intervalo de tempo ou ajustar os filtros",
--- a/voc-datalake/frontend/public/locales/zh/common.json
+++ b/voc-datalake/frontend/public/locales/zh/common.json
@@ -77,6 +77,7 @@
     },
     "settings": "设置"
   },
+  "partialCountsHint": "近似值：无法读取完整时间范围，计数为下限",
   "partialWindowHint": "缩小筛选范围以查看全部",
   "problemAnalysisPage": {
     "emptyHint": "请尝试扩大时间范围或调整筛选条件",
--- a/voc-datalake/frontend/src/api/client.ts
+++ b/voc-datalake/frontend/src/api/client.ts
@@ -10,6 +10,7 @@ import type {
   SentimentBreakdown,
   CategoryBreakdown,
   SourceBreakdown,
+  PersonaBreakdown,
   IntegrationStatus,
   ScraperConfig,
   ScraperTemplate,
@@ -42,6 +43,7 @@ export type {
   SentimentBreakdown,
   CategoryBreakdown,
   SourceBreakdown,
+  PersonaBreakdown,
   IntegrationStatus,
   ScraperConfig,
   ScraperTemplate,
@@ -250,7 +252,7 @@ export const api = {
   },
   getPersonas: (range: DateRangeParams, source?: string) => {
     const searchParams = buildSearchParams({ ...range, source })
-    return fetchApi<{ period_days: number; personas: Record<string, number> }>(`/metrics/personas?${searchParams}`)
+    return fetchApi<PersonaBreakdown>(`/metrics/personas?${searchParams}`)
   },
   
   // Chat
--- a/voc-datalake/frontend/src/api/types.ts
+++ b/voc-datalake/frontend/src/api/types.ts
@@ -107,9 +107,20 @@ export interface MetricsSummary {
   avg_sentiment: number
   urgent_count: number
   /**
-   * True when the metrics were computed from a truncated raw-item scan
-   * (review-date basis or source filter on a very large window) and the
-   * counts are a lower bound rather than exact aggregates.
+   * True when the window could not be read in full, so the counts are a lower
+   * bound rather than a total. Three independent reasons, and the route reports
+   * all of them under this one name:
+   *
+   * - the raw-item scan was truncated (review-date basis, or a source filter on
+   *   a very large window);
+   * - a metric partition read stopped before the end of its window;
+   * - the requested window is wider than aggregates are retained for (~90 days),
+   *   in which case the older rows have already been deleted and no complete
+   *   answer exists to give.
+   *
+   * Optional only for backward compatibility with a deployed API that omitted it
+   * on the aggregates path; treat an absent value as `false` (which is what
+   * `?? false` at the call site does) rather than as unknown.
    */
   is_partial?: boolean
   daily_totals: {
--- a/voc-datalake/frontend/src/api/types.ts
+++ b/voc-datalake/frontend/src/api/types.ts
@@ -108,15 +108,17 @@ export interface MetricsSummary {
   urgent_count: number
   /**
    * True when the window could not be read in full, so the counts are a lower
-   * bound rather than a total. Three independent reasons, and the route reports
-   * all of them under this one name:
+   * bound rather than a total. THE canonical statement of this flag on the
+   * frontend — the sibling breakdown types point here rather than repeat it.
+   *
+   * Three independent reasons, reported under one name because the caller's
+   * response to all three is the same:
    *
    * - the raw-item scan was truncated (review-date basis, or a source filter on
    *   a very large window);
    * - a metric partition read stopped before the end of its window;
    * - the requested window is wider than aggregates are retained for (~90 days),
-   *   in which case the older rows have already been deleted and no complete
-   *   answer exists to give.
+   *   so the older rows are already deleted and no complete answer exists.
    *
    * Optional only for backward compatibility with a deployed API that omitted it
    * on the aggregates path; treat an absent value as `false` (which is what
@@ -137,18 +139,38 @@ export interface MetricsSummary {
 export interface SentimentBreakdown {
   period_days: number
   total: number
+  /** See {@link MetricsSummary.is_partial} — same flag, same three reasons. */
+  is_partial?: boolean
   breakdown: Record<string, number>
   percentages: Record<string, number>
 }
 
 export interface CategoryBreakdown {
   period_days: number
+  /** See {@link MetricsSummary.is_partial} — same flag, same three reasons. */
+  is_partial?: boolean
   categories: Record<string, number>
 }
 
 export interface SourceBreakdown {
   period_days: number
+  /** See {@link MetricsSummary.is_partial} — same flag, same three reasons. */
+  is_partial?: boolean
   sources: Record<string, number>
+}
+
+/**
+ * Response envelope for `/metrics/personas`.
+ *
+ * Named rather than inlined at the call site so the route's `is_partial` has
+ * somewhere to be declared: an inline generic is where a field goes to be
+ * invisible, which is the same defect this flag exists to close one layer down.
+ */
+export interface PersonaBreakdown {
+  period_days: number
+  /** See {@link MetricsSummary.is_partial} — same flag, same three reasons. */
+  is_partial?: boolean
+  personas: Record<string, number>
 }
 
 export interface IntegrationStatus {
@@ -214,6 +236,16 @@ export interface ScraperTemplate {
 export interface EntitiesResponse {
   period_days: number
   feedback_count: number
+  /**
+   * See {@link MetricsSummary.is_partial} — same flag, same three reasons.
+   *
+   * Describes the COUNTS (`feedback_count` and the category/source/persona maps).
+   * `entities.issues` is a deliberate sample on both of the route's paths — the
+   * newest rows of at most seven days, capped at `limit` — so it is not what this
+   * flag is about, and folding it in would make the flag true on nearly every
+   * call and therefore worth nothing.
+   */
+  is_partial?: boolean
   entities: {
     keywords: Record<string, number>
     categories: Record<string, number>
--- a/voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx
+++ b/voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx
@@ -35,12 +35,19 @@ vi.mock('../../store/configStore', () => ({
   })),
 }))
 
-// Mock child components to simplify testing
+// Mock child components to simplify testing.
+//
+// `hint` is forwarded, not dropped: it is the only user-visible statement that a
+// total is a lower bound rather than exact, and a mock that swallows it lets the
+// hint be deleted — or left as an untranslated literal — with this suite green.
+// MetricCard's own test covers what it DOES with the value (title + aria-label);
+// here it just has to be reachable.
 vi.mock('../../components/MetricCard', () => ({
-  default: ({ title, value }: { title: string; value: string | number }) => (
+  default: ({ title, value, hint }: { title: string; value: string | number; hint?: string }) => (
     <div data-testid={`metric-${title.toLowerCase().replace(/\s/g, '-')}`}>
       <span>{title}</span>
       <span>{value}</span>
+      {hint && <span data-testid="metric-hint">{hint}</span>}
     </div>
   ),
 }))
@@ -151,6 +158,29 @@ describe('Dashboard', () => {
         expect(screen.getByText('~1,234')).toBeInTheDocument()
       })
       expect(screen.getByText('~5')).toBeInTheDocument()
+      // The `~` alone does not say WHY, so the hint carries the meaning. Asserted
+      // as the resolved English copy rather than the key, because a missing or
+      // misspelled key renders as the key name and would otherwise pass: the test
+      // setup loads the real public/locales/en/common.json.
+      // getAllByTestId: both counted cards carry the hint, and a getByTestId
+      // would fail on the second rather than on the thing under test.
+      const hints = screen.getAllByTestId('metric-hint')
+      expect(hints).toHaveLength(2)
+      expect(hints[0]).toHaveTextContent(
+        'Approximate: the full window could not be read, counts are a lower bound'
+      )
+    })
+
+    it('does not hint at approximation when the window was read in full', async () => {
+      // The positive control for the case above. Without it the hint could be
+      // rendered unconditionally — a permanent "approximate" is as uninformative
+      // as never showing it, and both read as "ignore this".
+      render(<Dashboard />, { wrapper: createWrapper() })
+
+      await waitFor(() => {
+        expect(screen.getByText('1,234')).toBeInTheDocument()
+      })
+      expect(screen.queryByTestId('metric-hint')).not.toBeInTheDocument()
     })
 
     it('displays average sentiment metric', async () => {
--- a/voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx
+++ b/voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx
@@ -6,6 +6,9 @@ import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { render, screen, waitFor } from '@testing-library/react'
 import { QueryClient, QueryClientProvider } from '@tanstack/react-query'
 import { TestRouter } from '../../test/test-utils'
+// The copy under test, read from the locale file the app itself renders from —
+// the same source src/test/setup.ts loads into i18next.
+import enCommon from '../../../public/locales/en/common.json'
 
 // Mock API before importing component
 const mockGetSummary = vi.fn()
@@ -158,17 +161,20 @@ describe('Dashboard', () => {
         expect(screen.getByText('~1,234')).toBeInTheDocument()
       })
       expect(screen.getByText('~5')).toBeInTheDocument()
-      // The `~` alone does not say WHY, so the hint carries the meaning. Asserted
-      // as the resolved English copy rather than the key, because a missing or
-      // misspelled key renders as the key name and would otherwise pass: the test
-      // setup loads the real public/locales/en/common.json.
-      // getAllByTestId: both counted cards carry the hint, and a getByTestId
-      // would fail on the second rather than on the thing under test.
+      // The `~` alone does not say WHY, so the hint carries the meaning. The
+      // expected copy is READ from the English locale file rather than pasted
+      // here — the repo's lockstep idiom — so editing the wording is one change
+      // instead of two. A missing or misspelled key still fails, because i18next
+      // renders the key name and that is not what the file says.
+      //
+      // getAllByTestId, and no assertion on how MANY: more than one card carries
+      // the hint, a getByTestId would fail on the second rather than on the thing
+      // under test, and a count would be a tripwire on unrelated card edits.
       const hints = screen.getAllByTestId('metric-hint')
-      expect(hints).toHaveLength(2)
-      expect(hints[0]).toHaveTextContent(
-        'Approximate: the full window could not be read, counts are a lower bound'
-      )
+      expect(hints.length).toBeGreaterThan(0)
+      for (const hint of hints) {
+        expect(hint).toHaveTextContent(enCommon.partialCountsHint)
+      }
     })
 
     it('does not hint at approximation when the window was read in full', async () => {
--- a/voc-datalake/frontend/src/pages/Dashboard/Dashboard.tsx
+++ b/voc-datalake/frontend/src/pages/Dashboard/Dashboard.tsx
@@ -81,11 +81,15 @@ function MetricsGrid({ summary, sourcesCount }: Readonly<MetricsGridProps>) {
   const avgSentiment = summary ? Number(summary.avg_sentiment) : 0
   const sentimentTrend = avgSentiment > 0 ? 'up' : 'down'
   const sentimentColor = avgSentiment > 0 ? 'green' : 'red'
-  // Review-basis metrics come from a budget-bounded scan; when truncated the
-  // counts are lower bounds and the cards must say so instead of look exact.
+  // The route reports this on BOTH of its paths now, and the hint no longer names
+  // only one of them: a review-basis or source-filtered answer comes from a
+  // budget-bounded scan, and an aggregates answer is short when a metric read
+  // stopped early or when the window reaches past the ~90 days aggregates are
+  // retained for. Either way the counts are lower bounds and the cards must say so
+  // instead of looking exact.
   const isPartial = summary?.is_partial ?? false
   const partialHint = isPartial
-    ? 'Approximate: the window exceeded the scan budget, counts are a lower bound'
+    ? 'Approximate: the full window could not be read, counts are a lower bound'
     : undefined
   const approx = (n: number | string) => (isPartial ? `~${n}` : n)
 
--- a/voc-datalake/frontend/src/pages/Dashboard/Dashboard.tsx
+++ b/voc-datalake/frontend/src/pages/Dashboard/Dashboard.tsx
@@ -78,19 +78,20 @@ interface MetricsGridProps {
 }
 
 function MetricsGrid({ summary, sourcesCount }: Readonly<MetricsGridProps>) {
+  // 'common', matching the namespace this file already reads, and where the
+  // sibling partial-window copy lives (`partialWindowHint`, used by the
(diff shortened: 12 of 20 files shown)


perrozzi:

This PR makes `is_partial` a measured value on the aggregates paths of six metrics/feedback routes instead of a hardcoded `False`, by (a) having `_query_metric_window` return `(items, truncated)`, (b) adding `_read_was_truncated` for the unpaged `gsi1-by-metric-type` reads, and (c) adding a request-level `AGGREGATE_RETENTION_DAYS` horizon check in `shared/api.py`, plus MCP schema/version updates and two new pytest modules. The core logic is sound, the `>` boundary and the OR-across-partitions choice are correct, and the AST-derived route list is a genuinely good idea. Main gaps are around completeness of the change at the edges (other `_query_metric_window` call sites, other consumers of the newly-emitted field) and the i18n rule on the touched Dashboard string.

### Blockers

- **`voc-datalake/frontend/src/pages/Dashboard/Dashboard.tsx`** — the user-visible `partialHint` string (`'Approximate: the full window could not be read, counts are a lower bound'`) is hardcoded English. The repo rule is that every user-visible string has i18n keys for all locales; since this line is being edited anyway, move it to a translation key (`t('dashboard.metrics.partialHint')` or equivalent) and add it to every locale file. No frontend test in the diff pins the new copy either, so the change is unverified at the surface where users read it.

### Should fix

- **Verify there are no remaining `_query_metric_window` call sites.** The helper's return type changed from `list[dict]` to `tuple[list[dict], bool]`. The diff updates the seven call sites in `get_entities`, `get_summary`, `get_sentiment_metrics`, `get_category_metrics`; any other caller (another route in `metrics_handler.py`, or an importer elsewhere) would now silently iterate the 2-tuple and blow up on `item.get(...)` or count the boolean as a row. The AST-derived parametrization only covers routes that *publish* `is_partial`, so a non-publishing caller is exactly the case the new tests cannot catch. Worth an additional structural test (AST/grep) asserting every call to `_query_metric_window` unpacks two values.
- **New response fields have no typed frontend surface.** `is_partial` is now emitted by `/metrics/sources`, `/metrics/personas` and `/feedback/entities`, but only `MetricsSummary`'s doc comment was touched in `voc-datalake/frontend/src/api/types.ts`. If those responses have interfaces (sources/personas/entities), add `is_partial?: boolean` (and Zod schema field if one exists) so the flag is at least representable; otherwise it is emitted, forwarded and dropped — the same "invisible field" argument the PR makes for MCP.
- **MCP: only two tools were updated.** The PR's own argument is that an undeclared field validates and then goes unread. If any MCP tool forwards `/feedback/entities` (or `/metrics/sources` / `/metrics/personas` bodies outside `get_metrics_breakdown`), its `outputSchema` needs the same `is_partial` declaration in this commit, since the fingerprint lockstep is already being bumped here.
- **`test_aggregate_retention_lockstep.py` only pins the *default* `ttl_days`.** If any aggregator call site passes `ttl_days=` explicitly, the default can match `AGGREGATE_RETENTION_DAYS` while the rows actually written use a different horizon — the test would stay green over exactly the drift it exists to catch. Add an assertion that no call site overrides `ttl_days` (AST/regex over `aggregator/handler.py`), or read the value from where it is applied.
- **Cross-package import in `lambda/api/test/test_aggregate_retention_lockstep.py`.** `from aggregator import handler` executes the aggregator module, which per repo convention creates boto3 clients/resources at module level; in a CI environment without a region or the expected table env vars this fails at import (the helper defers it to test time, so it fails the test rather than collection — but it still fails for an environmental reason, not drift). Consider `pytest.importorskip` with a clear skip reason, or reading the signature from the source via AST instead of importing.
- Consider adding `is_partial` to `required` in `get_metrics_summary`'s `outputSchema` in `voc-datalake/lambda/api/mcp_handler.py`, for consistency with the existing `test_the_flag_is_declared_required_so_absence_cannot_read_as_complete` convention — or state in the version comment why it is deliberately optional here (backward compatibility with a deployed API that omits it), since the comment block currently argues only for declaring it.

### Nits

- `_aggregate_page(truncated=True)` returns an open cursor on *every* page, so `_query_metric_window` loops the full `days` bound — 89 queries per partition × categories × 6 parametrized routes in `test_metrics_partial_window.py`. Using a small `days` for the truncation cases (with a separate test pinning the bound) keeps the suite cheap.
- The same three-paragraph rationale is now restated in `shared/api.py`, `metrics_handler.py`, `mcp_handler.py`, `types.ts` and `Dashboard.tsx`. One canonical statement (the `AGGREGATE_RETENTION_DAYS` block) with short pointers would age better; five copies is five things to update.
- `_read_was_truncated(response)` is a general name for a check that only applies to the unpaged index reads; `_index_read_was_truncated` would prevent it being reused on a paged response where an open cursor is not truncation.
- The `or truncated` chains in `get_entities`/`get_summary` are five sequential reassignments; collecting the flags and `any(...)`-ing once would read closer to the "ORs across partitions" intent the comments describe.
- TTL deletion in DynamoDB can lag up to ~48h, so the horizon check will occasionally report partial for a window whose rows still exist. Harmless in the safe direction, but worth one sentence in `_window_exceeds_aggregate_retention`'s docstring so a future reader doesn't treat the boundary as exact.

---

perrozzi:

Since my last review the author addressed every blocker/should-fix: the Dashboard hint moved to a `partialCountsHint` key in all eight locales (with a rendered-copy test plus a positive control), the breakdown/persona/entities response types now declare `is_partial` (and `getPersonas` gained a named `PersonaBreakdown` instead of an inline generic), `_read_was_truncated` was renamed `_index_read_was_truncated`, the retention lockstep switched from importing `aggregator.handler` to parsing it with `ast` and gained call-site/parameter-use checks, a new `TestEveryCallerUnpacksTheTruncationFlag` covers the changed return type inside and outside the handler, an MCP test asks the reverse "every reachable publishing route has a tool declaring the flag" question, and the optional-vs-required choice is now argued and pinned. Only small things left.

### Resolved
- Hardcoded English `partialHint` → `t('partialCountsHint')` with all 8 locale files and Dashboard tests (blocker).
- `_query_metric_window` call sites: AST check that every call unpacks two values, plus a repo-wide check that nothing outside `metrics_handler.py` calls it.
- `is_partial?: boolean` now declared on `SentimentBreakdown`, `CategoryBreakdown`, `SourceBreakdown`, new `PersonaBreakdown`, `EntitiesResponse`.
- MCP completeness: `test_every_mcp_reachable_route_that_publishes_the_flag_has_a_tool_declaring_it` covers the "other tools" gap.
- Retention lockstep no longer imports the aggregator (environmental red vs. drift red), and now catches `ttl_days` overrides and a writer that ignores the parameter.
- `required` decision documented at `_IS_PARTIAL_DESCRIPTION` and pinned by a test.
- TTL-lag caveat added to `_window_exceeds_aggregate_retention`'s docstring; `types.ts` is now the single canonical frontend statement.

### Still open
- Whether a runtime validator (Zod/type guard) on the metrics responses strips the new `is_partial` — only the TS interfaces were touched, so worth a one-line confirmation that no schema layer sits in front of them.

### New findings
- `voc-datalake/lambda/api/test/test_mcp_delegation.py` — `test_the_flag_is_declared_optional_because_the_body_is_not_projected` asserts `"required" not in schema`, which is broader than the claim: declaring a genuinely always-present key (`period_days`) as required later fails a correct change. Prefer `"is_partial" not in schema.get("required", [])`.
- `voc-datalake/lambda/api/test/test_mcp_delegation.py` — the anti-vacuous line `assert publishing & reachable == served_by_the_metrics_tools` is stronger than "non-empty": a future `_BREAKDOWN_DIMENSIONS` entry pointing at a route that legitimately does not publish the flag fails a green situation. `assert publishing & reachable` is the check you described.
- `voc-datalake/lambda/api/test/test_mcp_delegation.py` — `from test_metrics_partial_window import PUBLISHING_ROUTES` couples two test modules through sys.path; it breaks under `--import-mode=importlib` or a different rootdir. Moving the derivation into a conftest/helper module would keep "one derivation, not two" without the import-order dependency.
- `voc-datalake/lambda/api/test/test_aggregate_retention_lockstep.py` — `_ttl_parameter_position` only walks `function.args.args`, so making `ttl_days` keyword-only fails with the misleading "no longer takes ttl_days"; and `test_no_call_site_overrides_the_ttl` only sees bare-`Name` calls inside `aggregator/handler.py`, so `handler.update_counter(..., ttl_days=30)` from another module is out of scope. Both are safe directions — worth one sentence in the docstring so the scope isn't overread.
- `voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx` — `toHaveLength(2)` and the literal English copy are both hand-copied; reading the expected string from `public/locales/en/common.json` would match the repo's own lockstep idiom and keeps the card count from being a tripwire on unrelated UI edits.

---

perrozzi:

Since my last review the author moved the AST derivation into a shared, non-collected helper (`test/metrics_publishing_routes.py`), narrowed the `required` assertion to `is_partial`, weakened the anti-vacuous check to non-emptiness, documented the scope limits of `_ttl_parameter_position` / `test_no_call_site_overrides_the_ttl` (and turned the keyword-only case into a distinct message), and switched the Dashboard hint assertion to read the copy from `public/locales/en/common.json` while dropping the card-count tripwire. All five of my previous new findings are addressed; the changes look clean.

### Resolved
- `test_the_flag_is_declared_optional_because_the_body_is_not_projected` now scopes to `"is_partial" not in schema.get("required", [])`.
- Anti-vacuous line is `assert publishing & reachable` instead of equality with the served set.
- Cross-test-module import replaced by `metrics_publishing_routes`, imported by both suites — one derivation, no import-order coupling.
- `_ttl_parameter_position` distinguishes "gone" from "keyword-only"; call-site scope caveat stated in the docstring.
- Dashboard test reads `enCommon.partialCountsHint` and no longer asserts `toHaveLength(2)`.

### Still open
- One-line confirmation that no runtime validator (Zod/type guard) in front of the metrics responses strips the newly declared `is_partial`.

### New findings
None blocking. Three nits:
- `voc-datalake/frontend/src/pages/Dashboard/Dashboard.test.tsx` — the JSON import reaches outside `src`; worth confirming `resolveJsonModule` and the tsconfig `include`/`rootDir` cover it under `typecheck` as well as vitest (PR reports both green, so probably fine).
- `voc-datalake/lambda/api/test/metrics_publishing_routes.py` — the docstring's importability claim rests on `conftest.py` putting `test/` on `sys.path`; if that insertion is only implicit via pytest's prepend mode, the "does not depend on pytest's import mode" sentence overstates it.
- `voc-datalake/lambda/api/test/metrics_publishing_routes.py` — `routes_publishing_is_partial` keeps only `paths[0]`, so a handler carrying two route decorators silently contributes one path; a comment (or recording all paths) would remove the ambiguity.

---

perrozzi:

Only `voc-datalake/lambda/api/test/metrics_publishing_routes.py` changed since my last review: the docstring now cites the exact `conftest.py` line that puts `test/` on `sys.path`, and `routes_publishing_is_partial` records every route path of a multi-decorator handler instead of `decorator_list[0]`. Both of my remaining nits there are addressed and the change is behaviour-preserving for the current handler (no route function carries two decorators, so the derived set is still the six the positive control pins).

### Resolved
- `metrics_publishing_routes.py` — importability claim now names the explicit `sys.path.append` in `conftest.py` rather than implying pytest's import mode.
- `metrics_publishing_routes.py` — all route paths of a multi-decorator handler are recorded, with the reasoning in the docstring; the ambiguity I flagged is gone.

### Still open
- One-line confirmation that no runtime validator (Zod/type guard) in front of the metrics responses strips the newly declared `is_partial`.
- Nit: `Dashboard.test.tsx` imports `../../../public/locales/en/common.json` from outside `src`; worth confirming `resolveJsonModule` / tsconfig `include` cover it under `typecheck` as well as vitest.

### New findings
None. Two very minor observations on the new code, neither blocking: a handler with two decorators now maps both paths to the same name, so if that ever happens `test_the_derivation_finds_the_routes_that_publish_the_flag`'s hardcoded six-path set will need updating (correct failure, just worth knowing it's the tripwire); and if two decorators on different functions ever declared the same path, the dict would silently keep the last handler name — harmless for the current use (only keys are consumed).

---

perrozzi:
## Answers to the two open confirmations

**No runtime validator sits in front of the metrics responses, so nothing strips `is_partial`.** All six getters in `frontend/src/api/client.ts` — `getSummary`, `getSentiment`, `getCategories`, `getSources`, `getPersonas`, `getEntities` — are plain `fetchApi<T>` calls whose type parameter is a compile-time annotation only; there is no `.parse()`, no `safeParse`, and no `normalize*` on any of those paths, and `useSummaryQuery` (the hook the dashboard and the sidebar badge share) adds none either. The Zod boundaries in this codebase are `feedbackSchema.ts`, `scrapersSchema.ts`, `categoriesSchema.ts`, `jobGroundingSchema.ts`, `mcpTokenSchema.ts` and the api-token normalizers — feedback rows, scraper configs, form/project payloads. None of them names a metrics envelope. So the new field reaches the component untouched, which is also why declaring it on the TS interfaces was the whole of the frontend work: adding a schema layer here would be a new boundary, not a fix.

**`resolveJsonModule` and the tsconfig scope do cover the new JSON import**, and `src/test/setup.ts` is the precedent — it imports the same `public/locales/en/*.json` files across the same boundary, which is how the resolved English copy is available in tests at all. Specifics:

- `resolveJsonModule` is implied by `"moduleResolution": "bundler"` in `tsconfig.app.json`, which is why neither that file nor `setup.ts` sets it explicitly.
- `npm run typecheck` (`tsc -b`) resolves to `tsconfig.app.json`, which **excludes** `src/**/*.test.tsx` and `src/test/**/*` — so it never sees this import, in this file or in `setup.ts`.
- The tests-inclusive config is `tsconfig.test.json` (`npm run typecheck:tests`, not part of `npm run check`). Run on this branch it reports **no error for the JSON import**; its output is the same set of pre-existing failures as on the base tip (`configStore.test.ts` missing `identityPoolId`, `test-utils.tsx` `verbatimModuleSyntax` and a `MemoryRouter` `future` prop). Nothing in it is from this PR.
- Vitest resolves it through Vite, and `src/pages/Dashboard/Dashboard.test.tsx` passes (25 tests).

**Third nit, addressed in code** (5b2c997b): `routes_publishing_is_partial` now records every path a publishing handler answers instead of `decorator_list[0]`. A handler with two route decorators would otherwise have contributed one path while the other published the flag untested — the same silence the derivation exists to prevent.

### Re-verification after the narrowing commits

All seven mutations re-run against the loosened assertions, each applied to the working tree, suite run, then reverted:

| Mutation | Result |
|---|---|
| a `_query_metric_window` call site stops unpacking the flag | caught |
| an aggregator call site passes `ttl_days=30` | caught |
| a writer takes `ttl_days` and stamps a literal instead | caught |
| a publishing route added to `DOMAIN_ROUTES` with no tool declaring the flag | caught |
| `required: ["is_partial"]` added to an unprojected pass-through tool | caught (plus the fingerprint lockstep) |
| the dashboard hint points at a key no locale defines | caught |
| the hint deleted from one locale | caught |

`pytest lambda` 3177 passed · frontend `vitest run` 185 files / 3060 passed · `npm run lint` (eslint, `--max-warnings 0`) pass · `tsc --noEmit` frontend and CDK pass · `ruff check --select F,E4,E7,E9` on the changed Python clean · `ruff check lambda plugins` 508, unchanged from the base · locale parity 31 tests pass · `i18n:check` reports the same two pre-existing `projectDetail:prototypeLink.*` findings as the base and nothing new, which is why the key went in `common` rather than `dashboard` (that script infers one namespace per file from the first `useTranslation`, so a `dashboard` hook in `MetricsGrid` would have made it misreport this file's existing `exportPdf` keys).

perrozzi:
## Deployed and live-verified (pre-merge)

Deployed the PR head (`5b2c997b`) to the **`b-` verification slot**, not the shared unprefixed one — the unprefixed slot is contended (its `VocApiStack` moved this morning without me), and the PR's behaviour needs no production data to prove.

`cdk deploy b-VocApiStack --exclusively` · **exit 0, 171s, zero warnings**. The diff was the safe shape: **20 Lambda `Code.S3Key` swaps** (every API Lambda bundles `lambda/shared/`, which is where `AGGREGATE_RETENTION_DAYS` lives) plus the website `SourceObjectKeys` — **no IAM, no stateful resources, no replacements**, which is what earned `--require-approval never`. Only `b-VocApiStack`'s `LastUpdatedTime` moved; the other nine stacks in the account are byte-identical, including all five unprefixed ones.

### The behaviour, on all six routes

The retention horizon is a property of the *request*, so it is provable in an empty slot: no aggregate rows are needed to show that a window wider than the rows' 90-day TTL cannot be answered completely.

| window | `/metrics/summary` · `/sentiment` · `/categories` · `/sources` · `/personas` · `/feedback/entities` |
|---|---|
| `days=7` | `is_partial: false` ×6 |
| `days=90` | `is_partial: false` ×6 |
| `days=91` | `is_partial: true` ×6 |
| `days=365` | `is_partial: true` ×6 |

**12 false / 12 true across 24 probes**, and `period_days` equals the requested `days` in every one — no window was silently widened or narrowed. The 90-vs-91 pair is the control: a hardcoded `false` (the defect) cannot produce the `true`s, and an unconditionally-true flag — the same defect pointing the other way — cannot produce the `false`s. `>` rather than `>=` is what makes 90 the last complete window, and that is now observable in a deployment rather than only in a test.

### The pre-fix control, at artifact level

The unprefixed slot still runs `development`, so its deployed `metrics_handler.py` is the pre-fix code. Both bundles were downloaded and asserted **structurally with `ast`** (a text grep would pass on the comments that name the new helpers):

| | b (this PR) | unprefixed (pre-fix) |
|---|---|---|
| `_window_exceeds_aggregate_retention` defined | ✅ | ❌ |
| `_index_read_was_truncated` defined | ✅ | ❌ |
| `_query_metric_window` returns a 2-tuple | ✅ | ❌ |
| routes publishing `is_partial` | **6** | **5** |

That 5→6 accounts for the defect more precisely than the description did: **five routes already published the flag and simply never measured it on the aggregates path**, and the sixth — `/metrics/summary` — **omitted the key entirely** there, which a caller reads as "complete" just the same. (Its review-basis branch builds its dict in `_summary_from_items`, a separate function, which is why the derivation attributes nothing to `get_summary` pre-fix.)

### MCP catalogue, from the deployed bundle

`MCP_SERVER_VERSION 3.7.0`; `get_metrics_summary` **and** `get_metrics_breakdown` both declare `is_partial` through the shared `_IS_PARTIAL_DESCRIPTION`, whose deployed text is *"True when the window could not be answered in full — a metric read stopped short, or the window is wider than aggregates are retained for. The counts are then a lower bound, not a total."* — i.e. both causes, in the artifact a model would actually read.

### The i18n blocker, at the delivery layer

All eight locale catalogues served by the b-slot CDN carry `partialCountsHint`, and **no locale is left on the English string**: en 72 chars · de 97 · es 85 · fr 95 · pt 88 · ja 29 · ko 30 · zh 20. That is the check that matters for the blocker — the copy is not merely present in the repo, it is being served.

### Regression control

`scripts/test-api.sh` against the b slot: **30/30 passed**, `POST /chat/stream` 200. (30 rather than the unprefixed slot's 34 is expected — four cases are data-dependent and the b slot holds no feedback.)

### Not done, deliberately

- **Not deployed to the unprefixed slot.** That is the shared environment and this PR is unmerged; the b slot proves the behaviour without competing for it.
- **The paging-truncation half is not exercised live** — it needs >1 MB of aggregate rows in a single metric partition, which no slot has. It is covered by the suite (stubbed cursors) and by the mutation table above; the horizon half is what a real deployment can show.
- **The unprefixed slot's `deployment-test` password has been rotated away from the repo default**, and I left it that way rather than resetting it to run a probe — that rotation is the recommended remedy for that credential being published, so undoing it for a test would be the wrong trade. The pre-fix comparison moved to artifact level instead, which needs no app credentials.

PR #361, opened 2026-08-21
docs(readme): add a Running the Tests section

perrozzi:
## Summary

Adds a short `## 🧪 Running the Tests` section to the README, placed directly after the Quick Start (the existing development instructions) and before Initial Login.

The section names the two things a newcomer needs and previously had to dig out of `package.json`:

- `npm run install:all` — the install step (root + CDK + frontend)
- `npm test` — the test command (frontend Vitest)

It then points at the other three suites (`test:cdk`, `test:stream`, `test:backend`), at `npm run check` as the full pre-PR gate, and links to the existing [Quality Checks](https://github.com/aws-samples/sample-voice-of-customer-datalake/blob/development/docs/deployment.md#quality-checks) table rather than duplicating it. 13 lines added, README-only, no code touched.

## Build and test results

- `mise run build` — **not available**: `mise ERROR no tasks defined in /workspace/...`. This repo has no mise tasks; the task harness also reported the initial `mise run build` failing before any of my changes, so this is pre-existing and not caused by this PR.
- `npm run lint` — **fails, pre-existing and unrelated**: `sh: 1: eslint: not found` at the very first leg (`lint:frontend`). The frontend `node_modules` are not installed in this container, so ESLint cannot start. The harness likewise recorded lint failing before my changes. No JS/TS/Python file is touched by this PR, so lint coverage of this diff is vacuous either way.
- Targeted validation of the change instead, all passing:
  - Every script named in the new section verified to exist in the root `package.json` via `node -e` against `require('./package.json').scripts` — `install:all`, `test`, `test:cdk`, `test:stream`, `test:backend`, `check` all report `OK`.
  - The link target `docs/deployment.md` exists and contains a literal `## Quality Checks` heading (line 45), so the `#quality-checks` anchor resolves.
  - Rendered Markdown re-read after editing to confirm heading level, fence, and wrapping are consistent with the surrounding sections.

## Decisions made

- **Placement**: after Quick Start, since that is where the install command already appears and the task asked for "near the existing development instructions."
- **`npm test` over `npm run test`**: shorter and equivalent; the deployment doc uses `npm run test`, but the README is the welcoming surface.
- **Kept it to a few lines** as requested: the per-command breakdown already exists in `docs/deployment.md`, so I linked it rather than restating the table.
[...]

PR #360, opened 2026-08-20
feat(mcp): bring the protocol envelope up to the spec it claims to speak

perrozzi:
Phase 2b of the MCP plan. Phase 2a (#355) moved what the tools READ — every tool
delegates to the route that owns its data. This is the envelope around those
tools, deliberately a separate change because the two review on different terms:
one is about data correctness, this one is about protocol conformance.

The comment above `MCP_PROTOCOL_VERSION` already named every piece of this as the
owed next step, and said why bumping the version alone would have been dishonest.

## What changed

**Version negotiation, and `server/discover`.** `SUPPORTED_PROTOCOL_VERSIONS`
replaces the pinned string with the three revisions the spec covers, ordered
newest first. `initialize` answers with the client's version when this server
implements it and the newest one otherwise — a counter-offer rather than a refusal,
because `initialize` is where a client learns what it is talking to and an error
there leaves it nothing to fall back to. The `MCP-Protocol-Version` HEADER is
stricter and refuses an unsupported value: by then the handshake has happened, so
a version this server never offered can only be a client contradicting itself.

`server/discover` reports the revisions, the methods, the transport headers, the
`resultType` vocabulary and the cost classes — every list derived from the registry
that implements the thing, so a method that exists is listed and one that is only
listed cannot exist. It is unauthenticated (it names no project, no tool, no data)
and deliberately does not list tools, because that answer is credential-shaped.

**`resultType` on every result.** Built into `_jsonrpc_result` as a REQUIRED
positional argument, which is the design: a new result shape cannot ship without
naming itself. `{}` was both a `ping` answer and a notification ack; a tool error
differed from a tool result by one boolean and by whether `structuredContent`
happened to be present. An AST test fails the moment a seventh JSON-RPC envelope
is hand-rolled outside the two builders, so the property belongs to the module
rather than to the six call sites somebody remembered.

**Transport-header validation**, including the `=?base64?...?=` encoded-word
sentinel form. A malformed sentinel is refused with the ENCODING named — comparing
it literally would report `unsupported version =?base64?...?=` and send the caller
after a version number for a fault in the encoding. `MCP-Method` and `MCP-Name`
are routing echoes, so they can DISAGREE with the body; a disagreement is refused
rather than resolved in either direction, because one hop routes on the header
while the next acts on the body and there is no way to know which was meant.
Validation runs before authentication, so a malformed request buys no probe of the
token store — the same property the Origin guard has. (The Origin guard itself was
[...]

PR #366, opened 2026-08-21
feat(prioritization): show the row count beside the heading

perrozzi:
## Summary

The Prioritization page's heading gave no sense of how long the list is, so a reader could not tell at a glance whether everything had loaded. A small count now sits beside the `h1`.

Kept to the heading area — the table is untouched.

```
Prioritization  ( 2 proposals )
Score and prioritize PRDs and PR/FAQs across all projects
```

## What it does

- `PrioritizationHeader` takes a `rowCount: number` and renders it as a pill beside the heading, never inside it, so the heading's accessible name stays the page's name.
- The number is the **list's own length** (`sortedRows.length`), not a second count derived from the documents, so the badge and the rows beneath it cannot disagree.
- **Nothing renders at zero.** "0 proposals" beside the heading asserts an empty backlog, and both states that arrive with nothing to count have not established one: the loading pass, and a genuinely empty list whose own empty state already says so in words *and* says which emptiness it is. A scores read that failed or is still paging also lands here with `isLoading` already cleared.
- Counts **rows**, the same unit as the "Total Proposals" card. One row can hold a PRD and a PR/FAQ describing one idea — real data already contains such a row.
- `rowCount_one` / `rowCount_other` in all eight locales.

## Verification

> ⚠️ **The verification block originally generated for this PR was wrong on both counts and is corrected here.** It reported `mise run build`: PASS and `mise run lint`: FAIL, which together read as "one gate failing". In fact:
>
> - **The reported `lint` FAIL is pre-existing and unrelated.** It is 507 ruff violations in `lambda/` and `plugins/`. Identical count on the base commit, and this PR touches no Python:
>   ```
>   ruff check lambda plugins @ this branch  → Found 507 errors.
>   ruff check lambda plugins @ f31a5f84     → Found 507 errors.   # base
>   ```
> - **The gate that mattered was never run.** `tsc` is a separate script (`npm run typecheck`), in `npm run check` but not in `lint`. The first commit on this branch declared `rowCount` in the props type and never destructured it, so it was an undefined identifier — `tsc` failed with two `TS2304`s and the page threw a `ReferenceError` on every render, taking 113 of its own vitest cases with it. ESLint passed clean throughout, because `no-undef` is off for TypeScript by design.

Gates re-run at this head:

| gate | result |
| --- | --- |
| `npm run typecheck:all` | **PASS** |
| `npm run lint:frontend` (`eslint . --max-warnings 0`) | **PASS** |
| `npx vitest run` (frontend) | **PASS** — 185 files, 3068 tests |
| `ruff check lambda plugins` | 507, **unchanged from base** (no Python in this diff) |
| `node scripts/i18n-check.mjs` | 0 missing / 0 extra / 0 unused for `rowCount` |
| `npx tsc -p tsconfig.test.json` | 102, **unchanged from base** (pre-existing; in no aggregate) |

### Tests

Nine cases. Each was checked by mutation rather than assumed:

| mutation | fails |
| --- | --- |
| drop the zero-gate **and** add a legal `rowCount_zero` form | 4 |
| delete `rowCount_one` | 1 (renders the raw key) |
| move the badge inside the `h1` | 1 |
| inflate the "Total Proposals" card by 10 | 1 |
| delete the badge's `data-testid` | 1 |
| make `sortRows` narrow (drop unscored rows) | 1 |
| make `sortRows` duplicate a row instead of dropping one | 1 |

The last two pin the invariant the badge/card agreement rests on: `sortRows` reorders without narrowing. The duplication case is there because a length-only assertion would accept a duplicated row masking a dropped one.

### Verified live

Deployed frontend-only and driven in a signed-in browser. Full evidence in a comment below; in short: badge, list and card agree at three different counts; `1 proposal` / `2 proposals` and `1 Vorschlag` / `2 Vorschläge` resolve against the real HTTP-loaded catalogues; the badge is genuinely absent over the empty state; zero console errors; sorting does not change the count; contrast 6.87:1.

Live testing also found that the deployed page has **two** `h1` elements — the app shell renders one too — which corrected a comment that had claimed "the page's only h1". That claim was unfalsifiable in the page suite, which renders without the app shell.

## Known, deliberately not addressed

- `StatsCards` is ungated, so "Total Proposals" prints a literal `0` in the same empty state where the badge stays silent. Pre-existing; confirmed live. If the badge's reasoning is right the follow-up is gating the card, not changing the badge.
- `es`/`fr`/`pt` have no CLDR `many` form, so they would render the raw key at a count of exactly 1,000,000. Also structurally invisible to `i18n-check`, which compares each locale to English key-for-key and English legitimately has no `_many`.
- No `aria-live` on the badge. Sorting is length-preserving and the page has no filter, so nothing a user does moves the count without also changing the list; a live region would announce a bare number while the rows changed silently, and would fire on the 0→N transition at mount. Measured live rather than argued.

---

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of the [project license](https://github.com/aws-samples/sample-autonomous-cloud-coding-agents/blob/main/LICENSE).

### DIFF
--- a/voc-datalake/frontend/public/locales/de/prioritization.json
+++ b/voc-datalake/frontend/public/locales/de/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} Dokumente, ein Stimmzettel"
   },
+  "rowCount_one": "{{count}} Vorschlag",
+  "rowCount_other": "{{count}} Vorschläge",
   "scores": {
     "confidence": "Konfidenz",
     "confidenceDescription": "Sicherheit bei Auswirkungs- und MeZ-Schätzungen",
--- a/voc-datalake/frontend/public/locales/en/prioritization.json
+++ b/voc-datalake/frontend/public/locales/en/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} documents, one ballot"
   },
+  "rowCount_one": "{{count}} proposal",
+  "rowCount_other": "{{count}} proposals",
   "scores": {
     "confidence": "Confidence",
     "confidenceDescription": "Certainty in impact and TTM estimates",
--- a/voc-datalake/frontend/public/locales/es/prioritization.json
+++ b/voc-datalake/frontend/public/locales/es/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} documentos, una papeleta"
   },
+  "rowCount_one": "{{count}} propuesta",
+  "rowCount_other": "{{count}} propuestas",
   "scores": {
     "confidence": "Confianza",
     "confidenceDescription": "Certeza en las estimaciones de impacto y TdC",
--- a/voc-datalake/frontend/public/locales/fr/prioritization.json
+++ b/voc-datalake/frontend/public/locales/fr/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} documents, un seul bulletin"
   },
+  "rowCount_one": "{{count}} proposition",
+  "rowCount_other": "{{count}} propositions",
   "scores": {
     "confidence": "Confiance",
     "confidenceDescription": "Certitude dans les estimations d'impact et de DmM",
--- a/voc-datalake/frontend/public/locales/ja/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ja/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "ドキュメント{{documents}}件、投票は1票"
   },
+  "rowCount_one": "提案{{count}}件",
+  "rowCount_other": "提案{{count}}件",
   "scores": {
     "confidence": "確信度",
     "confidenceDescription": "インパクトとTTM見積もりの確実性",
--- a/voc-datalake/frontend/public/locales/ko/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ko/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "문서 {{documents}}건, 표 1장"
   },
+  "rowCount_one": "제안 {{count}}건",
+  "rowCount_other": "제안 {{count}}건",
   "scores": {
     "confidence": "신뢰도",
     "confidenceDescription": "영향도 및 출시 시간 추정의 확실성",
--- a/voc-datalake/frontend/public/locales/pt/prioritization.json
+++ b/voc-datalake/frontend/public/locales/pt/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} documentos, uma cédula"
   },
+  "rowCount_one": "{{count}} proposta",
+  "rowCount_other": "{{count}} propostas",
   "scores": {
     "confidence": "Confiança",
     "confidenceDescription": "Certeza nas estimativas de impacto e TTM",
--- a/voc-datalake/frontend/public/locales/zh/prioritization.json
+++ b/voc-datalake/frontend/public/locales/zh/prioritization.json
@@ -121,6 +121,8 @@
   "row": {
     "multiDocument": "{{documents}} 个文档，一张选票"
   },
+  "rowCount_one": "{{count}} 个提案",
+  "rowCount_other": "{{count}} 个提案",
   "scores": {
     "confidence": "置信度",
     "confidenceDescription": "对影响力和上市时间估计的确定性",
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1114,15 +1114,15 @@ describe('Prioritization', () => {
     // beneath it from disagreeing.
 
     /**
-     * Everything the heading area says, as one string.
+     * The count badge, or null when the heading is showing none.
      *
-     * The h1's own wrapper, which holds the heading and the count and nothing else. Read
-     * as a whole so a test can assert what the area does NOT say: a count withheld is
-     * this reading back exactly the title, which stays true however a future plural form
-     * spells zero — whereas querying for the literal "0 proposals" only rules out one
-     * spelling of the claim.
+     * By testid, which is the only handle that holds for the ABSENCE cases in both of
+     * the directions they can rot. Querying for the text "0 proposals" rules out one
+     * spelling of the claim, so a later `rowCount_zero` form would pass unnoticed;
+     * reading the heading wrapper's whole text is spelling-proof but goes vacuous the
+     * moment anything is nested around the h1, which is a test that fails open.
      */
-    const headingArea = () => screen.getByRole('heading', { level: 1 }).parentElement
+    const rowCountBadge = () => screen.queryByTestId('prioritization-row-count')
 
     it('counts the rows beside the heading', async () => {
       // Two projects in the shared fixture, so TWO rows — p1's PRD and PR/FAQ are one
@@ -1144,9 +1144,11 @@ describe('Prioritization', () => {
       renderPrioritization()
 
       expect(await screen.findByText('2 proposals')).toBeInTheDocument()
-      const grid = screen.getByText('Total Proposals').closest<HTMLElement>('div.grid')
-      expect(within(grid ?? document.body).getByText('Total Proposals').previousElementSibling)
-        .toHaveTextContent('2')
+      // `toBe` on the text, not `toHaveTextContent`: that matcher is a SUBSTRING test
+      // for a string argument, so a card reading "12" or "20" would satisfy the very
+      // assertion this case exists to make.
+      expect(screen.getByText('Total Proposals').previousElementSibling?.textContent)
+        .toBe('2')
     })
 
     it('uses the singular form for a list of one', async () => {
@@ -1173,27 +1175,29 @@ describe('Prioritization', () => {
       renderPrioritization()
 
       expect(await screen.findByText('Loading documents...')).toBeInTheDocument()
-      expect(headingArea()).toHaveTextContent(/^Prioritization$/)
+      expect(rowCountBadge()).toBeNull()
     })
 
-    // Both empty branches, because the count is withheld for a reason that names them:
-    // the list says WHICH emptiness this is, and a bare zero cannot. The badge takes the
-    // same path through either, so this is one case run twice rather than two cases.
+    // EVERY empty branch, because the count is withheld for a reason that names them:
+    // the list says which emptiness this is, and a bare zero cannot. The badge takes one
+    // path through all three, so this is one case run three times — and `projects` is
+    // its own column rather than inferred from `documents`, so "a project that has no
+    // documents yet" is expressible instead of silently collapsing into "no projects".
     it.each([
-      ['no projects at all', 'No Documents Found', []],
-      ['only non-scorable documents', 'No Scorable Documents', [
-        { document_id: 'r1', document_type: 'research', title: 'Research Only', content: '', created_at: '2025-01-01' },
-      ]],
-    ])('does not claim "0 proposals" over the empty state for %s', async (_case, emptyTitle, documents) => {
-      mockGetProjects.mockResolvedValue({
-        projects: documents.length === 0 ? [] : [mockProjects[0]],
-      })
+      { emptiness: 'no projects at all', title: 'No Documents Found', projects: [], documents: [] },
+      { emptiness: 'a project with no documents yet', title: 'No Documents Found', projects: [mockProjects[0]], documents: [] },
+      {
+        emptiness: 'only non-scorable documents', title: 'No Scorable Documents', projects: [mockProjects[0]],
+        documents: [{ document_id: 'r1', document_type: 'research', title: 'Research Only', content: '', created_at: '2025-01-01' }],
+      },
+    ])('shows no count over the empty state for $emptiness', async ({ title, projects, documents }) => {
+      mockGetProjects.mockResolvedValue({ projects })
       mockGetProject.mockResolvedValue({ project_id: 'p1', documents })
 
       renderPrioritization()
 
-      expect(await screen.findByText(emptyTitle)).toBeInTheDocument()
-      expect(headingArea()).toHaveTextContent(/^Prioritization$/)
+      expect(await screen.findByText(title)).toBeInTheDocument()
+      expect(rowCountBadge()).toBeNull()
     })
 
     it('counts the rows still on screen when the score read fails', async () => {
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1129,7 +1129,13 @@ describe('Prioritization', () => {
       // row, which is why this is not "three documents".
       renderPrioritization()
 
-      expect(await screen.findByText('2 proposals')).toBeInTheDocument()
+      // Read off the BADGE, not merely found somewhere on the page. The two cases below
+      // assert this node is absent, and if nothing ever asserts it is present, deleting
+      // the testid leaves those two passing for the wrong reason and the zero-gate
+      // unguarded. Keeping the string in the assertion still proves the plural resolved
+      // through the real locale JSON rather than falling back to the key.
+      expect((await screen.findByTestId('prioritization-row-count')).textContent)
+        .toBe('2 proposals')
       // Beside the h1, not inside it: the page's only level-1 heading has to keep
       // reading as the page's name to a screen reader and the document outline, and it
       // is what the breadcrumb names.
@@ -1185,7 +1191,10 @@ describe('Prioritization', () => {
     // documents yet" is expressible instead of silently collapsing into "no projects".
     it.each([
       { emptiness: 'no projects at all', title: 'No Documents Found', projects: [], documents: [] },
-      { emptiness: 'a project with no documents yet', title: 'No Documents Found', projects: [mockProjects[0]], documents: [] },
+      // Named for the state it actually reaches: the shared fixture's rows still name
+      // `d1`/`d2`, which this project no longer has, so `collectRows` drops every row.
+      // The one case where a detail IS loaded and holds no documents at all.
+      { emptiness: 'a project whose documents are gone', title: 'No Documents Found', projects: [mockProjects[0]], documents: [] },
       {
         emptiness: 'only non-scorable documents', title: 'No Scorable Documents', projects: [mockProjects[0]],
         documents: [{ document_id: 'r1', document_type: 'research', title: 'Research Only', content: '', created_at: '2025-01-01' }],
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1107,6 +1107,116 @@ describe('Prioritization', () => {
     })
   })
 
+  describe('the heading says how long the list is', () => {
+    // A reader cannot tell from the heading alone whether everything loaded, so the
+    // count goes beside it. The number is the LIST's own length rather than a second
+    // count computed from the documents, which is what keeps the badge and the rows
+    // beneath it from disagreeing.
+
+    /**
+     * Everything the heading area says, as one string.
+     *
+     * The h1's own wrapper, which holds the heading and the count and nothing else. Read
+     * as a whole so a test can assert what the area does NOT say: a count withheld is
+     * this reading back exactly the title, which stays true however a future plural form
+     * spells zero — whereas querying for the literal "0 proposals" only rules out one
+     * spelling of the claim.
+     */
+    const headingArea = () => screen.getByRole('heading', { level: 1 }).parentElement
+
+    it('counts the rows beside the heading', async () => {
+      // Two projects in the shared fixture, so TWO rows — p1's PRD and PR/FAQ are one
+      // row, which is why this is not "three documents".
+      renderPrioritization()
+
+      expect(await screen.findByText('2 proposals')).toBeInTheDocument()
+      // Beside the h1, not inside it: the page's only level-1 heading has to keep
+      // reading as the page's name to a screen reader and the document outline, and it
+      // is what the breadcrumb names.
+      expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent(/^Prioritization$/)
+    })
+
+    it('agrees with the Total Proposals card', async () => {
+      // The badge counts `sortedRows` and the card counts `allRows`, so they agree only
+      // while the sort preserves length — which nothing else on the page asserts. They
+      // are two statements of the same quantity in the same viewport, and a reader who
+      // finds them disagreeing has reason to trust neither.
+      renderPrioritization()
+
+      expect(await screen.findByText('2 proposals')).toBeInTheDocument()
+      const grid = screen.getByText('Total Proposals').closest<HTMLElement>('div.grid')
+      expect(within(grid ?? document.body).getByText('Total Proposals').previousElementSibling)
+        .toHaveTextContent('2')
+    })
+
+    it('uses the singular form for a list of one', async () => {
+      // Not cosmetic. A count-only key needs both `rowCount_one` and `rowCount_other`
+      // under i18next's JSON v4 plurals, and a missing form renders the raw key path —
+      // the failure this page's own comments warn about elsewhere. Tests run the real
+      // locale JSON through the real i18next, so this case is what proves the forms
+      // resolve rather than that they merely exist in the file.
+      useLayout(oneRowPerDocument([
+        { document_id: 'd1', document_type: 'prfaq', title: 'Feature A PR/FAQ', content: '', created_at: '2025-01-01' },
+      ]))
+
+      renderPrioritization()
+
+      expect(await screen.findByText('1 proposal')).toBeInTheDocument()
+      expect(screen.queryByText('rowCount')).toBeNull()
+    })
+
+    it('says nothing while the list is still loading', async () => {
+      // The badge must not appear over the spinner: a count is a claim about a list,
+      // and there is no list yet.
+      mockGetProjects.mockReturnValue(new Promise(() => {})) // Never resolves
+
+      renderPrioritization()
+
+      expect(await screen.findByText('Loading documents...')).toBeInTheDocument()
+      expect(headingArea()).toHaveTextContent(/^Prioritization$/)
+    })
+
+    // Both empty branches, because the count is withheld for a reason that names them:
+    // the list says WHICH emptiness this is, and a bare zero cannot. The badge takes the
+    // same path through either, so this is one case run twice rather than two cases.
+    it.each([
+      ['no projects at all', 'No Documents Found', []],
+      ['only non-scorable documents', 'No Scorable Documents', [
+        { document_id: 'r1', document_type: 'research', title: 'Research Only', content: '', created_at: '2025-01-01' },
+      ]],
+    ])('does not claim "0 proposals" over the empty state for %s', async (_case, emptyTitle, documents) => {
+      mockGetProjects.mockResolvedValue({
+        projects: documents.length === 0 ? [] : [mockProjects[0]],
+      })
+      mockGetProject.mockResolvedValue({ project_id: 'p1', documents })
+
+      renderPrioritization()
+
+      expect(await screen.findByText(emptyTitle)).toBeInTheDocument()
+      expect(headingArea()).toHaveTextContent(/^Prioritization$/)
+    })
+
+    it('counts the rows still on screen when the score read fails', async () => {
+      // The count follows the LIST, and the list deliberately survives a failed scores
+      // read on the rows the ensure confirmed (see the sibling suite). Withholding the
+      // badge on that failure would leave the heading silent above rows the reader can
+      // see and count by hand — and the count is not one of the numbers that read
+      // failed, so it has nothing to retract.
+      useLayout(oneRowPerDocument([
+        { document_id: 'd1', document_type: 'prfaq', title: 'Feature A PR/FAQ', content: '', created_at: '2025-01-01' },
+        { document_id: 'd3', document_type: 'prfaq', title: 'Feature B PR/FAQ', content: '', created_at: '2025-01-02' },
+      ]))
+      mockGetPrioritizationScores.mockRejectedValue(new Error('API Error: 500'))
+
+      renderPrioritization()
+
+      expect(await screen.findByText('2 proposals')).toBeInTheDocument()
+      // And the failure is still reported. Counting the rows is not claiming the read
+      // worked.
+      expect(screen.getByRole('alert', { name: 'Scores could not be loaded' })).toBeInTheDocument()
+    })
+  })
+
   describe('PR/FAQ list', () => {
     it('displays PR/FAQ items after loading', async () => {
       renderPrioritization()
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1187,8 +1187,8 @@ describe('Prioritization', () => {
     // EVERY empty branch, because the count is withheld for a reason that names them:
     // the list says which emptiness this is, and a bare zero cannot. The badge takes one
     // path through all three, so this is one case run three times — and `projects` is
-    // its own column rather than inferred from `documents`, so "a project that has no
-    // documents yet" is expressible instead of silently collapsing into "no projects".
+    // its own column rather than inferred from `documents`, which is what lets the middle
+    // case have a project at all instead of collapsing into "no projects".
     it.each([
       { emptiness: 'no projects at all', title: 'No Documents Found', projects: [], documents: [] },
       // Named for the state it actually reaches: the shared fixture's rows still name
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1136,9 +1136,15 @@ describe('Prioritization', () => {
       // through the real locale JSON rather than falling back to the key.
       expect((await screen.findByTestId('prioritization-row-count')).textContent)
         .toBe('2 proposals')
-      // Beside the h1, not inside it: the page's only level-1 heading has to keep
-      // reading as the page's name to a screen reader and the document outline, and it
-      // is what the breadcrumb names.
+      // Beside the h1, not inside it: this heading has to keep reading as the page's
+      // name to a screen reader and the document outline, and it is what the breadcrumb
+      // names.
+      //
+      // `level: 1` is unambiguous HERE ONLY because this suite renders the page without
+      // the app shell. The DEPLOYED page has two level-1 headings — Layout renders the
+      // brand as one — so this query would be ambiguous in a test that mounted Layout,
+      // and no assertion in this file can speak for the page's overall heading
+      // structure. Verified in a browser, not here.
       expect(screen.getByRole('heading', { level: 1 })).toHaveTextContent(/^Prioritization$/)
     })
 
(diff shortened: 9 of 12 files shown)


perrozzi:

This PR adds a row-count badge next to the Prioritization page's `h1`, plumbing a new `rowCount` prop into `PrioritizationHeader` and adding `rowCount_one`/`rowCount_other` keys to all eight locale files. The change is small, well-scoped, and correctly leaves the table untouched; the i18n coverage is complete and pluralization is handled via i18next `count`. Two things hold it back: the reported lint failure and the absence of a test for the new behavior.

### Blockers

- **`mise run lint` reports FAIL** in the PR's own verification section. Whatever the violation is (likely formatting/comment-length in `voc-datalake/frontend/src/pages/Prioritization/Prioritization.tsx`), it needs to be green before merge — the build gate is disabled for this repo, so lint is the only automated signal available here.

### Should fix

- **No vitest coverage for the new badge.** `data-testid="prioritization-row-count"` was clearly added for testability but no test uses it. Add a test in the Prioritization page suite asserting (a) no badge while loading, (b) `"N proposals"` with the correct count once rows resolve, and (c) singular form at `count === 1`. This is the only guard against the `null`-vs-`0` distinction the prop doc goes to length to explain.
- **Error state shows a false "0 proposals".** `rowCount={isLoading ? null : sortedRows.length}` (`Prioritization.tsx:~902`) only withholds the count during load. If the query errors, `isLoading` is false and `sortedRows` is empty, so the page asserts an empty backlog — exactly the failure mode the prop comment says it wants to avoid. Gate on the error/`isSuccess` state too, e.g. `isLoading || isError ? null : sortedRows.length`.
- **Verify the badge's gate matches the list's spinner condition.** The doc comment claims the count appears only "once there is a list to count," but the list's spinner appears to depend on more than `isLoading` (the header itself reads `ownBallots.inHand` for `saveBlocked`, and the comment mentions "projects *and their documents*"). If documents load on a separate query, the badge can render a partial count above a still-spinning list. Worth confirming the two conditions are the same expression.
- **Confirm the i18next plural-suffix format matches the installed major version.** `_one`/`_other` is JSON v4 (i18next ≥ 21 / `compatibilityJSON` unset). If the project pins an older version or sets `compatibilityJSON: 'v3'`, these keys resolve to the literal fallback instead of the plural form and the badge silently renders the raw key.

### Nits

- `rowCount_one` is dead weight in `ja`, `ko`, and `zh` — those locales have only the `other` plural category under `Intl.PluralRules`, and both variants are byte-identical anyway. Harmless, but it invites future translators to think the distinction is real.
- Consider `aria-live="polite"` on the badge span: the count changes when sliders/filters re-sort or re-filter the list, and a screen-reader user currently gets no notification.
- The 12-line JSDoc on the `rowCount` prop plus the 4-line inline comment in the JSX is a lot of prose for "number of rows, or null while loading" — the rationale for `null` is the only part that isn't self-evident from the type.
- Commit subject `chore(agent): save uncommitted work from session end` doesn't describe the change; a `feat(prioritization): show row count beside heading` would be more useful in `git log` and matches the conventional-commit scope used elsewhere.

---

perrozzi:

Since the last review the author added a seven-case test suite for the badge and replaced the `number | null` prop with a plain `number` plus a "never render zero" gate in `PrioritizationHeader`, dropping the `data-testid` in favour of text queries. That addresses the test-coverage gap, the false "0 proposals" on error, the spinner-condition question (now documented as an invariant and covered by a loading test), and the i18next plural-version doubt (the singular case exercises the real locale JSON).

### Resolved
- Vitest coverage for the badge now exists, including loading, singular, both empty branches, and the failed-scores case.
- Error/empty state no longer claims "0 proposals" — zero is never rendered, so the `isLoading`-only gate concern is moot.
- Badge/spinner gate reconciled: `isLoading ⊂ rowCount === 0` is documented and tested.
- Plural-suffix format validated end-to-end by the `1 proposal` assertion rather than by inspection.

### Still open
- `mise run lint` is still reported as FAIL in the PR body; needs to be green before merge (it is the only automated gate here).
- No `aria-live` on the badge, so re-sorts/filters that change the count are silent to screen readers (nit, carried over).

### New findings

**Should fix**

- `Prioritization.test.tsx` — the `it.each` table mixes types per row, so TypeScript infers the tuple elements as `string | { document_id: string; ... }[]`. `mockGetProject.mockResolvedValue({ project_id: 'p1', documents })` will likely fail type checking (and may be the lint/tsc failure above). Annotate the table rows as a tuple type, e.g. `it.each<[string, string, DocumentFixture[]]>([...])`.
- `Prioritization.test.tsx` — verify `within` is imported in this file and that `useLayout` / `oneRowPerDocument` are in scope; the incremental diff introduces the first use of `within` in the new block without any import change.
- `Prioritization.test.tsx` — `getByRole('alert', { name: 'Scores could not be loaded' })` depends on the alert having an accessible name (`aria-label`/`aria-labelledby`). If the existing alert only has text content, this assertion will not match; confirm against the sibling suite it references.

**Nits**

- `expect(screen.queryByText('rowCount')).toBeNull()` is redundant once `findByText('1 proposal')` has passed.
- `getByText('Total Proposals').previousElementSibling` couples the test to card markup; a `data-testid` or scoped role query would survive a layout tweak.
- The prop JSDoc grew from 12 to ~20 lines plus a 7-line comment on `isLoading` and a 4-line comment at the call site — three places now restating one rule ("no badge at zero"). Trimming to one authoritative sentence on the prop would age better.

---

perrozzi:
Pushed `c356ddbc`. Every finding below was checked against the code rather than taken on trust, so a few come back as "not reproducible here" with the reason.

## The badge could not render at all — a blocker neither of us named

Before anything from the review: `rowCount` was declared in `PrioritizationHeader`'s props type and read in its JSX, but never destructured out of the parameter object. It was an undefined identifier.

```
$ npm run typecheck    # at 369198ac
src/pages/Prioritization/Prioritization.tsx(468,12): error TS2304: Cannot find name 'rowCount'.
src/pages/Prioritization/Prioritization.tsx(473,39): error TS2304: Cannot find name 'rowCount'.

$ npx vitest run src/pages/Prioritization
ReferenceError: rowCount is not defined
Tests  113 failed | 250 passed (363)
```

The whole page was blank — a `ReferenceError` into the router's error boundary on every render, not a missing badge. Fixed by destructuring it.

Worth recording *why* both of us missed it: `lint` in this repo is ESLint plus ruff, and ESLint's `no-undef` is off for TypeScript because `tsc` is supposed to own that class. But `tsc` is a separate script (`npm run typecheck`), which the verification pipeline never ran. The one gate that would have caught this with a one-line error was the one not wired in.

## Blockers

**`mise run lint` FAIL — not this PR.** It is 507 pre-existing ruff violations in `lambda/` and `plugins/`. Identical count on the base commit, and this diff touches no Python:

```
ruff check lambda plugins @ pr366        → Found 507 errors.
ruff check lambda plugins @ f31a5f84     → Found 507 errors.
```

`lint:frontend` passes clean at `--max-warnings 0`. This one is a standing repo condition, not a regression to clear.

## Should fix

**No vitest coverage — fixed, seven cases.** The count itself; agreement with the "Total Proposals" card; the singular form; silence during loading; silence over *both* empty branches; and a count that survives a failed scores read alongside the alert reporting it.

I verified they constrain the behaviour rather than merely passing, by mutation:

| mutation | result |
| --- | --- |
| drop the zero-gate **and** add a legal `rowCount_zero` form | 3 fail |
| delete `rowCount_one` | 1 fail (renders raw key `rowCount`) |
| move the badge inside the `h1` | 1 fail |

That first row is your "Nits" point about absence tests, caught early: asserting `queryByText('0 proposals')` is null only rules out one *spelling* of the claim. The assertions read the whole heading area instead, so it has to say exactly `Prioritization` and nothing else.

**Error state shows a false "0 proposals" — real, fixed differently than suggested.** Your literal suggestion, `isLoading || isError ? null : sortedRows.length`, does not work here, for two reasons:

1. There is no `isError` for the queries that produce rows. Only the *scores* query exposes one (`scoresFailed`).
2. Gating on `scoresFailed` would be backwards. This page deliberately keeps the rows through a failed scores read — `allRows` merges `ensuredRows` under `savedScores.rows` precisely so a 500 on the scores shows rows with the numbers marked unavailable rather than an empty backlog. Hiding the count there would blank the heading above rows the reader can see and count by hand.

So the fix is to withhold at **zero** instead: `rowCount === 0` renders nothing. That covers the state you identified (nothing to count, `isLoading` already cleared) plus the plain empty list, without contradicting the list. The empty state below already says it in words *and* says which emptiness — no documents at all, versus none of a scorable type — which a bare zero cannot.

That also collapsed the prop from `number | null` to `number`. The `null` arm was dead: `collectRows` early-returns `[]` until both project reads land, and neither query carries `placeholderData`/`initialData`, so a loading page always has zero rows. The two-state prop was one rule spelled twice.

**Badge gate vs the list's spinner condition — confirmed identical, and now recorded.** Both resolve to the same thing today: the spinner is `isLoading` (`loadingProjects || loadingDetails`), and `isLoading ⟹ rowCount === 0` via that `collectRows` early return. Your suspicion about documents loading separately was half right — they *do* come from a second query (`getProject` fan-out), but `loadingDetails` is the second half of `isLoading`, so both are covered.

The equivalence is incidental rather than structural, though, so I noted the dependency at `isLoading`: widening it to `isFetching` to spin on refetch would leave cached rows counted above a spinner, and the loading test would not catch it (initial load has no cached data either way).

**i18next plural-suffix format — confirmed correct.** `i18next@^25.9.0`, and `compatibilityJSON` is not set anywhere in the frontend, so JSON v4 `_one`/`_other` is right. Better than confirmed by inspection: tests run the *real* locale JSON through the real i18next singleton, so `screen.findByText('1 proposal')` is an end-to-end check that the forms resolve. Deleting either one turns the assertion into the raw key.

## Nits

**`rowCount_one` in ja/ko/zh — keeping it, it is the house pattern.** `stats.unreadable_one`/`_other` in the same file are byte-identical in all three of those locales too, as are `showAllCategories_*` and `showTop_*` in `categories.json`. Dropping it just for this key would make it the odd one out, and `i18n-check` already accepts the pair.

**`aria-live` — deliberately not adding it.** The premise does not hold: sorting is length-preserving (`[...rows].sort(...)`) and this page has no filter or search, so nothing a user does moves the count without also changing the list. A live region would announce a bare number while the rows it describes changed silently, and would fire spuriously on the 0→N transition at mount. The house line is the same — counts are plain text; `role="status"` is reserved for content that changes without a list change.

**Comment volume — trimmed, and one claim corrected.** The bigger problem was accuracy, not length: the prop claimed the badge and the "Total Proposals" card "cannot disagree". They share a *unit*, not a value — the badge counts `sortedRows`, the card counts `allRows`, equal only while the sort preserves length. Narrowed, and the new test pins the equality so it cannot drift silently. Under a future row filter they *should* differ, and each would be right about what it is labelled.

**Commit subject — PR title is now `feat(prioritization): show the row count beside the heading`.** Left the original commit in place rather than force-pushing over an existing review.

## Verification

```
npm run typecheck:all        PASS
npm run lint:frontend        PASS  (eslint --max-warnings 0)
npx vitest run               PASS  185 files, 3066 tests
node scripts/i18n-check.mjs        0 missing / 0 extra for `rowCount`
```

The page's own suite went from **113 failing** to **83 passing**.

## Not fixed, deliberately

- **`StatsCards` prints a literal `0` for "Total Proposals" during the same loading pass the badge stays silent through.** Pre-existing, and two blocks below the heading. If the reasoning behind withholding the badge is right, the follow-up is gating the card, not changing the badge — out of scope here.
- **`es`/`fr`/`pt` have no CLDR `many` form,** so they would render the raw key at a count of exactly 1,000,000. Also structurally invisible to `i18n-check`, which compares each locale to English key-for-key and English legitimately has no `_many`. Not worth carrying at this page's scale.

perrozzi:

Since the last review the author swapped the heading-wrapper text query for a `data-testid="prioritization-row-count"` handle (re-adding the testid to `PrioritizationHeader`), converted the `it.each` tuple table to an object table with an explicit `projects` column (plus a third empty branch), replaced the `within(...)/closest('div.grid')` card assertion with `previousElementSibling?.textContent` + `toBe`, and added an invariant comment to `sortRows`. That clears two of the three Should-fix items from the previous review.

### Resolved
- `it.each` type inference: object-table form with named columns removes the mixed-tuple inference problem (and adds the "project with no documents" branch).
- `within` import concern is moot — the new card assertion no longer uses `within` or `closest`.
- Absence assertions no longer depend on nothing being nested around the `h1`, and `toBe` on `textContent` fixes the substring-match weakness in the "Total Proposals" comparison.

### Still open
- `mise run lint` is reported FAIL in the PR body; must be green before merge.
- `getByRole('alert', { name: 'Scores could not be loaded' })` still assumes the alert has an accessible name — confirm the alert element carries `aria-label`/`aria-labelledby`, otherwise this assertion silently can't match.
- No `aria-live` on the badge (nit, carried over).

### New findings

**Nits**

- `Prioritization.test.tsx` — verify the object-table rows still typecheck: `document_type: 'research'` in a bare object literal widens to `string`, so if `mockGetProject`'s payload type expects a literal union for `document_type` this row will error (the same class of failure as the tuple issue, just relocated). An `as const` or a typed fixture helper settles it.
- `Prioritization.tsx` — the "no badge at zero / testid rationale" argument is now stated in four places (prop JSDoc, the JSX comment, the `isLoading` comment, the call site) and mirrored again in the test's `rowCountBadge` JSDoc. One authoritative sentence on the prop would age better than five copies.
- `prioritizationUtils.ts` — the new `sortRows` comment documents an invariant ("reorders, never narrows") that nothing asserts; a one-line unit test on length would enforce it more cheaply than the paragraph.

---

perrozzi:

Since the last review the only changes are in `Prioritization.test.tsx`: the "counts the rows beside the heading" case now reads the badge via `findByTestId('prioritization-row-count')` with `toBe('2 proposals')` (closing the loop so the two absence cases can't pass for the wrong reason), and the second `it.each` row was renamed/commented to describe the state it actually reaches. That addresses the remaining "nothing asserts the testid is present" concern; the rest of the previously open items are untouched.

### Resolved
- Badge presence is now asserted through the same `data-testid` the absence cases query, so deleting the testid fails a test instead of silently passing.

### Still open
- `mise run lint` reported FAIL in the PR body; must be green before merge.
- `getByRole('alert', { name: 'Scores could not be loaded' })` still assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the object table (nit).
- The zero-gate rationale is still duplicated across five places (nit).
- `sortRows`'s "reorders, never narrows" invariant is documented but still unasserted (nit).

No new problems introduced — the incremental changes are clean.

---

perrozzi:

Since my last review the only change is a one-sentence rewording of the explanatory comment above the `it.each` empty-state block in `Prioritization.test.tsx` — no assertions, fixtures, or production code touched. Previously resolved item stands; the remaining open items are untouched.

### Still open
- `mise run lint` reported FAIL in the PR body; must be green before merge.
- `getByRole('alert', { name: 'Scores could not be loaded' })` still assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the `it.each` table (nit).
- Zero-gate rationale still duplicated across five places (nit).
- `sortRows`'s "reorders, never narrows" invariant documented but unasserted (nit).

No new findings — the comment edit is clean and accurate.

---

perrozzi:
## Converged

Four local review passes, four commits on top of the original. The last pass returned APPROVED with nothing to fix before merge.

| commit | what it did |
| --- | --- |
| `c356ddbc` | destructured `rowCount` (the page did not render at all); collapsed `number \| null` → `number`; 7 tests |
| `07f526c9` | fixed three tests that could pass while the behaviour was broken |
| `0870a426` | made the badge testid load-bearing in both directions |
| `9d42b8db` | comment left stale by a case relabel |

### What actually blocked this

`rowCount` was declared in `PrioritizationHeader`'s props and read in its JSX but never destructured, so the whole Prioritization page threw a `ReferenceError` on every render — `tsc` failed with two TS2304s and 113 of the page's own vitest cases were failing. Neither reported gate caught it: `lint` here is ESLint plus ruff, ESLint's `no-undef` is off for TypeScript because `tsc` owns that class, and `tsc` lives in a separate script the pipeline never ran. The reported `mise run lint` FAIL was 507 pre-existing ruff errors in `lambda/`/`plugins/` — identical on the base commit, and this diff touches no Python.

### Design changes from the review

The badge is withheld at **zero** rather than during loading. That covers the false "0 proposals" over a failed scores read that the first review flagged, without the gate the review suggested — `isLoading || isError` would have hidden the count above a list that deliberately keeps showing rows through a failed scores read, contradicting the list it describes. `null` turned out to be dead: `collectRows` returns `[]` until both project reads land, so a loading page always has zero rows.

Two claims were narrowed rather than defended. The prop no longer says the badge and the "Total Proposals" card "cannot disagree" — they share a unit, not a value, and a future row filter should make them differ. `sortRows` now records that it reorders without narrowing, which is what the agreement rests on.

### Tests

Eight cases. Every one was mutation-checked, and three rounds of review were spent on cases that passed while the thing they named was broken:

| mutation | fails |
| --- | --- |
| drop the zero-gate **and** add a legal `rowCount_zero` form | 4 |
| delete `rowCount_one` | 1 (renders the raw key) |
| move the badge inside the `h1` | 1 |
| inflate the "Total Proposals" card to `rows.length + 10` | 1 |
| delete the badge's `data-testid` | 1 |

The last two are the interesting ones. `toHaveTextContent('2')` is a **substring** match for a string argument, so the card-agreement assertion passed against a card reading `12` — the one assertion whose entire purpose was numeric agreement. And the testid was asserted absent twice and present nowhere, so deleting it left the suite green with the zero-gate unguarded.

### Verification

```
npm run typecheck:all      PASS
npm run lint:frontend      PASS   eslint --max-warnings 0
npx vitest run             PASS   185 files, 3067 tests
node scripts/i18n-check.mjs       0 missing / 0 extra / 0 unused for rowCount
```

The page's own suite: **113 failing → 84 passing**.

### Left alone, deliberately

- **`StatsCards` prints a literal `0` for "Total Proposals"** during the same loading pass the badge stays silent through. Pre-existing. If withholding the badge is right, the follow-up is gating the card.
- **`es`/`fr`/`pt` have no CLDR `many` form**, so they render the raw key at a count of exactly 1,000,000. Also invisible to `i18n-check`, which compares each locale to English key-for-key and English legitimately has no `_many`.
- **`aria-live` on the badge.** Sorting is length-preserving and the page has no filter, so nothing a user does moves the count without also changing the list. A live region would announce a bare number while the rows changed silently, and would fire on the 0→N mount transition.
- **Badge position is not pinned by any test.** Moving the count into the subtitle carrying its testid along would stay green. The alternatives were considered across two review passes; each traded this for a worse coupling.

perrozzi:

Since my last review the only change is an expanded comment above the count badge in `Prioritization.tsx`, clarifying that the app shell also renders an h1 so this page has two in deployment. No production logic, markup, i18n keys, or assertions changed, so all previously open items stand.

### Still open
- `mise run lint` reported FAIL in the PR body; must be green before merge.
- `getByRole('alert', { name: 'Scores could not be loaded' })` still assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the `it.each` table (nit).
- Zero-gate rationale still duplicated across five places (nit).
- `sortRows`'s "reorders, never narrows" invariant documented but unasserted (nit).

### New findings
- Nit — `voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx`: the new component comment now explicitly disclaims "the only h1", but the test comment in `counts the rows beside the heading` still reads "the page's only level-1 heading"; the two docs now contradict each other. Worth aligning the test comment (the `getByRole('heading', { level: 1 })` query only works because the shell isn't rendered under test — saying so there is more useful than the "only h1" claim).

Otherwise the comment edit is accurate and self-contained.

---

perrozzi:
## Deployed and verified end to end on the live site

Frontend-only deploy of `b36c7ed` to the primary deployment (`VocCoreStack` / `VocApiStack`), then driven in a real signed-in Chrome session. No CDK, no Lambda, no infrastructure change — the diff touches 8 locale files and 3 files under `frontend/src/`.

**Blast radius was checked, not assumed.** Live was previously built from `99f5e58`. Between that and this head, the only commits that reach the bundle are this PR's: `#360` was already deployed and `#361` is docs-only.

```
$ git diff --name-only 99f5e58..HEAD -- frontend/src frontend/public frontend/index.html frontend/package.json
→ only voc-datalake/frontend/{public/locales/*/prioritization.json, src/pages/Prioritization/*}
```

Artifact provenance, at the network layer: the locale request carries the build stamp, so the served bundle names its own commit.

```
GET /locales/en/prioritization.json?v=b36c7ed     ← == git rev-parse --short HEAD
```

### The three states, on the deployed bundle

| state | how it was produced | heading area | list | "Total Proposals" card |
| --- | --- | --- | --- | --- |
| populated | real data | **`2 proposals`** | 2 rows (`#1`, `#2`) | `2` |
| singular | projects response trimmed 3 → 1 | **`1 proposal`** | 1 row | `1` |
| empty | projects response trimmed 3 → 0 | *badge absent* | "No Documents Found" + its sentence | `0` |

The two non-default states were produced by wrapping `fetch` in a `navigate_page` init script, so the fault is injected before React mounts. Read-only — no server data was touched — and each run reports what it intercepted, so the result cannot be a coincidence:

```json
{"interceptionFired":[{"url":".../v1/projects","was":3,"now":1}],
 "badgeText":"1 proposal","isSingular":true,"rawKeyLeaked":false,
 "listRows":["#1"],"totalProposalsCard":"1"}
```

**Badge, list and card agree in every state.** That was the invariant with no test behind it until this PR; it now also has live evidence at three different counts.

One detail worth noting because it validates the unit choice: row `#2` reads *"PR/FAQ PRD … 2 documents, one ballot"*. Real data already contains a row holding both document types, which is exactly why the count is 2 rows and not 3 documents.

### Structure and the plural, measured rather than eyeballed

```json
{"badgeTag":"SPAN","badgeIsInsideH1":false,"badgeParentIsH1Parent":true,
 "badgePrevSibling":"H1:Prioritization","pageH1Text":"Prioritization",
 "gapPx":8,"onSameVisualLineAsH1":true,"overflowsViewport":false}
```

Empty state, showing the absence is real rather than a mis-selected element:

```json
{"badgePresent":false,"headingAreaText":"Prioritization",
 "nextAfterH1":"none (h1 is last child of its wrapper)",
 "says0Proposals":false,"rawKeyLeaked":false,"emptyStateShown":true}
```

**The i18next plural question is now settled against the real HTTP-loaded catalogue**, not the statically-imported copy the tests use. German was checked as a second language with a genuine one/other distinction (unlike ja/ko/zh, where both forms are identical):

| locale | count 2 | count 1 |
| --- | --- | --- |
| en | `2 proposals` | `1 proposal` |
| de | `2 Vorschläge` | `1 Vorschlag` |

`de/prioritization.json?v=b36c7ed` was fetched and `document.body.innerText` never contained the string `rowCount`, so no form fell back to a raw key path.

### Other live checks

- **Zero console errors or warnings**, across the whole session including preserved messages. Worth stating explicitly: the originally submitted commit threw a `ReferenceError` on every render.
- **Sorting is length-preserving on real data** — clicking Date Created then Impact left the badge at `2 proposals` and the list at 2 rows. That is the claim that made `aria-live` the wrong call, and it is now measured rather than argued.
- **Contrast 6.87:1** for the rendered pill (`rgb(74,85,101)` on `rgb(243,244,246)`, 14px/500). Tailwind v4 emits `oklch()`, so the colours were rasterised to sRGB and read back rather than string-parsed; the harness carries a black-on-white control that returns exactly 21:1. This independently reproduces the figure already recorded for this pair elsewhere in the file. Passes AA; below AAA.
- **A true 320 CSS px viewport** (via device emulation — window resize is clamped at ~625 here, and `devicePixelRatio` is 1.6, so "320" asked for is not 320 got): badge visible at 12px, still beside the heading, no overflow and no horizontal scroll.
- **Four routes smoke-tested** (`/dashboard`, `/projects`, `/categories`, `/prioritization`) — all render, no error boundary. The deploy re-stamps every chunk even though only this page changed, so the other consumers were checked rather than assumed.

### One thing live testing found, now fixed

`b36c7ed` corrects a comment. It justified placing the count beside the heading by appealing to *"the page's only h1"* — and the deployed page has **two**, because the app shell's brand heading is also level 1:

```json
{"h1Count":2,"h1Texts":["VoC Analytics","Prioritization"]}
```

The design decision is unchanged and its reason still holds (this heading's accessible name must stay the page's name), so only the claim is narrowed; the pre-existing two-h1 condition is named as separate rather than folded in. This is the kind of error a test cannot catch, because the page suite renders `Prioritization` without the app shell and therefore genuinely has one `h1`.

### Pre-existing, now with live evidence

The earlier review's finding about `StatsCards` is confirmed in situ. In the empty state the card prints a literal **`0`** in 2xl bold two blocks below the heading, while the badge correctly says nothing:

```json
{"badgePresent":false,"emptyStateShown":true,"statsCardTotalProposals":"0"}
```

So the page already makes the assertion the badge withholds to avoid making. Still out of scope here — the follow-up is gating the card, not changing the badge — but it is no longer a code-reading observation.

### Owed after merge

`APP_VERSION` is baked from the git SHA, so `b36c7ed` becomes a dead SHA once this squash-merges. **A frontend re-deploy from the merged tip is owed** — functionally a no-op, but it stops the live bundle naming a commit that resolves in no canonical repo. Rollback for the current deploy is captured; the `assets/` sync runs without `--delete`, so the previous chunks are still in the bucket.

perrozzi:

Since my last review the author aligned the `level: 1` test comment with the component comment (resolving my prior new finding), dropped the redundant `queryByText('rowCount')` assertion with a correct rationale, and added a `sortRows` length/identity-preservation test in `prioritizationUtils.test.ts` that pins the badge-vs-card invariant. All other previously open items are untouched.

### Resolved
- Test comment in `counts the rows beside the heading` no longer contradicts the component's "not the only h1" note.
- `sortRows`'s "reorders, never narrows" invariant is now asserted (across all four fields and both directions, by id set rather than length alone).
- Removed assertion was genuinely redundant — a missing plural form renders the key path *instead of* the text, so `findByText('1 proposal')` already covers it.

### Still open
- `mise run lint` reported FAIL in the PR body; must be green before merge.
- `getByRole('alert', { name: 'Scores could not be loaded' })` still assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the `it.each` table (nit).
- Zero-gate rationale still duplicated across five places (nit).

### New findings
- Nit — `voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.test.ts`: the comment claims "every arm of the grouping is exercised", but `b`'s aggregate only sets `impact`/`reviewer_count`, so for the `time_to_market` and `created_at` sorts the intended scored/unscored split may not be what the comment describes; the assertion itself is still valid, only the claim is broader than the fixture. Also `idsOf(sorted).slice()` is redundant — `idsOf` already returns a fresh array.

Otherwise the incremental changes are clean and self-contained.

---

perrozzi:
## Disposition — five review rounds that had gone unread

I declared this converged off a local reviewer while five automated rounds sat unanswered here, one of them posted **18 seconds before** my "converged" comment, on the exact head it reviewed. That claim was about my own reviewer, not about this thread. Working through all of them now.

Pushed `9a2d697a`, and **the PR body has been rewritten** (body edits don't show in the timeline, hence this note).

### `mise run lint` FAIL — the reason this returned five times is the PR body

This was re-listed in every single round, and answering it in a comment could never work: the review reads the **diff and the body**, not comments. The body still carried the original generated block reporting `lint`: FAIL, so each round correctly re-derived a blocker from it.

**Fixed at the source — the body now carries the measurement instead of the claim:**

```
ruff check lambda plugins @ this branch  → Found 507 errors.
ruff check lambda plugins @ f31a5f84     → Found 507 errors.   # base, identical
```

507 pre-existing violations in `lambda/` and `plugins/`; this diff contains no Python. The body also now records the gate that actually mattered and was never run — `tsc`, which caught the undestructured `rowCount` that made the page throw on every render.

Standing correction rather than a rebuttal: an unfixable-by-this-PR gate result in the body will be re-raised forever. Put the correction where the reviewer looks.

### Fixed in `9a2d697a`

**`sortRows`' invariant was documented and unasserted** — right, and the paragraph was the weaker instrument. Now pinned across all four fields × both directions. Two mutations, because length alone is not enough:

| mutation | result |
| --- | --- |
| drop rows with no team aggregate | **fails** — `priority_score/asc: expected … length 3 but got 2` |
| duplicate a row instead of dropping one | **fails** — identity check catches what length would accept |

My first attempt at this was itself a no-op: I mutated to `.filter(r => views.has(r.row_id))`, but `views` is populated for *every* row passed in, so it filtered nothing and reported a false "caught". Worth stating because the mutation that proves nothing looks exactly like the one that proves something.

**The two h1 comments contradicted each other** — correct, and it was my own half-fix. The component comment was corrected to disclaim "the only h1" while the test comment still asserted it. Taken your suggested framing, which is the more useful fact: the `getByRole('heading', { level: 1 })` query is unambiguous *because this suite renders without the app shell*, and therefore no assertion in that file can speak for the deployed page's heading structure.

**`queryByText('rowCount')` redundant** — agreed, dropped. A missing plural form renders the key path *instead of* the text, so finding `1 proposal` already proves resolution; the two assertions cannot disagree.

### Not reproducible

**`document_type: 'research'` widening to `string`** — measured, and it does not error:

```
$ npx tsc --noEmit -p tsconfig.test.json | grep -c "Prioritization.test.tsx"
0                       # project baseline unchanged at 102, all pre-existing elsewhere
```

The cause is that `mockGetProject` is a bare `vi.fn()` with no type parameter, so `mockResolvedValue` takes `any` — there is no literal union for the widened string to violate. Note this file is typechecked by no gate in `npm run check`, so the measurement is `tsconfig.test.json` run by hand against a stashed baseline.

**`getByRole('alert', { name: 'Scores could not be loaded' })` assumes an accessible name** — it has one. The panel carries `aria-labelledby="scores-unavailable-title"`, added deliberately because two same-role regions without accessible names are indistinguishable to a screen reader *and* to a test. The assertion is self-discharging: it is a `getByRole` with a name filter, so it throws if the name does not match, and it passes.

### Declined, with reasons

**`aria-live` on the badge.** The premise doesn't hold here. Sorting is length-preserving and this page has no filter or search, so nothing a user does moves the count without also changing the list — measured on the deployed page, not argued: clicking Date Created then Impact left the badge at `2 proposals` and the list at 2 rows. A live region would announce a bare number while the rows it describes changed silently, and would fire on the 0→N transition at mount.

**`getByText('Total Proposals').previousElementSibling` couples to card markup.** True, and it is this file's own idiom at roughly six existing sites (`cardValue`). Introducing a competing convention for one assertion costs more than it saves.

**Rationale stated in five places.** Fair, and partly acted on: the two test-side copies are now one, and the remaining prose is the prop JSDoc plus two short pointers to it. I'm leaving those, because this file's house style is load-bearing rationale at the point of use, and the zero-gate is the one part of this change a future reader is likely to "simplify" by rendering the zero.

perrozzi:

Since my last review the only change is in `prioritizationUtils.test.ts`: the fixture comment was rewritten to describe the three grouping regimes accurately, and the redundant `.slice()` before `.sort()` was dropped — both of my prior new findings are addressed. Nothing else moved, so the previously open items stand.

### Resolved
- Comment no longer over-claims "every arm of the grouping is exercised"; it now names which fields land in which regime.
- `idsOf(sorted).slice().sort()` → `idsOf(sorted).sort()`, with a note explaining why the copy is unnecessary.

### Still open
- `mise run lint` FAIL reported in the PR body (author now attributes it to 507 pre-existing ruff violations; needs the repo's gate green or an explicit maintainer waiver).
- `getByRole('alert', { name: 'Scores could not be loaded' })` assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the `it.each` table (nit).
- Zero-gate rationale duplicated across five places (nit).

### New findings
- Should fix — `voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.test.ts`: the edit collapsed the two loop headers onto one line: `for (const field of [...] as const) {      for (const direction of ['asc', 'desc'] as const) {`. Almost certainly an accidental join during the comment rewrite; restore the line break (and the closing braces' indentation if it drifted) so the nesting is readable.

Otherwise the delta is clean.

---

perrozzi:

The only change since my last review is the whitespace fix in `prioritizationUtils.test.ts` — the accidentally joined `for` loop headers are now split across two lines with correct nesting. That was my sole outstanding new finding, so the delta is clean.

### Resolved
- `voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.test.ts`: nested `for` loops restored to separate lines with proper indentation.

### Still open
- `mise run lint` FAIL reported in the PR body (attributed to 507 pre-existing ruff violations; needs the repo gate green or an explicit maintainer waiver).
- `getByRole('alert', { name: 'Scores could not be loaded' })` assumes the alert carries an accessible name.
- No `aria-live` on the badge (nit).
- `document_type: 'research'` literal widening in the `it.each` table (nit).
- Zero-gate rationale duplicated across five places (nit).

No new findings.

---

perrozzi:
## Converged — "No new findings" on a review of the current head

`0e5804b8`, the review of that exact SHA reports **no new findings**. Stating the head explicitly because my earlier convergence claim did not: it came from a local reviewer while five rounds here were unread, one of them posted 18 seconds earlier on the head it reviewed.

### Rounds since that mistake

| head | outcome |
| --- | --- |
| `9a2d697a` | 3 resolved · 2 new nits (both mine, both valid) |
| `fdc93a17` | 2 resolved · 1 should-fix — **a real defect I introduced** |
| `0e5804b8` | 1 resolved · **no new findings** |

The `fdc93a17` finding is worth naming. The regex I used to strip a temporary probe from the test ate the newline between two `for` headers and joined them on one line. Valid JS, so **eslint, `tsc` and 3,068 tests all stayed green** — no gate can see it, and a reviewer caught it. Fixed in `0e5804b8`, and I then scanned the whole file for the same damage (a brace followed by code on the same line) and found none.

### The five carried items

**`mise run lint`** — the reason this survived five rounds is that the review reads the **PR body**, not comments, and the body still carried the original generated `FAIL`. **The body is now corrected** and this round re-framed it as needing a maintainer waiver rather than a flat blocker, which is the right shape: 507 pre-existing ruff violations in `lambda/`/`plugins/`, identical on the base, in a diff with no Python. Clearing it is a separate PR, and your call.

**`getByRole('alert', { name: … })` assumes an accessible name** — it has one, and here is the proof rather than the assertion. The panel is:

```tsx
<div role="alert" aria-labelledby="scores-unavailable-title" …>
  <h3 id="scores-unavailable-title">{t('scoresUnavailable.title')}</h3>
```

Stripping that one attribute fails the test:

```
× counts the rows still on screen when the score read fails
  → Unable to find an accessible element with the role "alert" and name "Scores could not be loaded"
```

So the assertion is load-bearing in both directions: it passes only because the name exists, and it is the thing that would notice if the name were removed. (The labelling is deliberate — two same-role regions without accessible names are indistinguishable to a screen reader *and* to a test.)

**`document_type: 'research'` widening** — measured, does not error. `npx tsc -p tsconfig.test.json` attributes **0** errors to `Prioritization.test.tsx`, project baseline unchanged at 102 (all pre-existing `is_frozen` errors elsewhere). `mockGetProject` is a bare `vi.fn()`, so `mockResolvedValue` takes `any` and there is no literal union to violate.

**`aria-live`** — declined, measured on the deployed page: sorting is length-preserving and the page has no filter, so clicking Date Created then Impact left the badge at `2 proposals` and the list at 2 rows. Nothing a user does moves the count without also changing the list, so a live region would announce a bare number while the rows changed silently, and would fire on the 0→N transition at mount.

**Rationale in five places** — partly taken: the two test-side copies are now one. The remaining prose is the prop JSDoc plus two short pointers, kept because the zero-gate is the one part of this change a future reader is most likely to "simplify" by rendering the zero.

### Where it stands

Nine tests, every one mutation-checked, including the two that pin `sortRows`' reorder-never-narrow property — the invariant the badge/card agreement rests on. Gates at this head: `typecheck:all` PASS · `lint:frontend` PASS · 185 files / 3,068 tests PASS · ruff and `typecheck:tests` both unchanged from base.

Deployed and verified live at an earlier head; the three commits since are test and comment changes only, so the live evidence still describes the shipped behaviour. A frontend re-deploy from the merged tip is owed either way, because `APP_VERSION` bakes the SHA.

PR #367, opened 2026-08-21
feat(prioritization): show the listed row count beside the heading

perrozzi:
## Summary

The Prioritization heading said nothing about how long the list below it is, so a reader could not tell at a glance whether the backlog had finished loading. A small count badge now sits next to the `<h1>`:

- **Counts the rows the list actually renders** — `allRows.length`, the same array `PRFAQList` maps over and the same one the "Total Proposals" card counts. Two numbers for one quantity is how a page loses the trust this badge exists to earn, so it reads the one array rather than re-deriving a count (a test pins the badge and the card to each other).
- **Withheld while the read is in flight.** A `0` there would be a claim about a backlog nobody has read yet, and it is the one number a reader would take as "the list finished and it is empty" — precisely the wrong reading. The list below already shows a spinner for that state. Once the read lands, `0 proposals` is honest and is the state the badge is most use in.
- **Counted with its noun** (`3 proposals`, `1 proposal`) via `count` interpolation, not a bare numeral: beside a heading a lone number has no subject, and a screen reader announcing "Prioritization 3" says nothing. New key `headingCount_one` / `headingCount_other` added to all eight locales (de, en, es, fr, ja, ko, pt, zh), which keeps `localeParity.test.ts` and `i18n-check` happy — CJK locales get the `_other` form they need.
- `text-gray-600` on the badge, per the measured contrast table already recorded in `BAND_STYLE` (gray-500 fails AA below 18.5px on this page's gray-50 background).

**The table/list itself is untouched** — the diff is the header component, the wiring of one prop from the page, the eight locale files, and tests.

### Decisions made

- **Rows, not projects or documents.** A project whose PRD and PR/FAQ describe one idea is *one* row on this page, and the stats cards were already fixed once for counting documents instead. The badge counts what is on screen.
- **Hidden rather than dashed during loading.** The stats cards use an explained `—` because they are permanently-present cards with labels; a badge beside a heading has no label to hang an explanation on, and it can simply not be there for the second the read takes.
- **No `aria-live`.** It is not an update announcement, it is part of the heading area; it renders with the list it describes.

## Build and test results

| Command | Result |
|---|---|
| `npm run typecheck` (`tsc -b --noEmit`) | **pass** |
| `npm run lint` (`eslint . --max-warnings 0`) | **pass** |
| `vitest run src/pages/Prioritization/Prioritization.test.tsx` | **pass** — 81/81 (5 new) |
| `vitest run src/i18n src/pages/Prioritization` | **pass** — 363/363, 12 files |
| `node scripts/i18n-check.mjs` | **unchanged** — `Total missing: 0 / extra: 12 / empty: 0 / untranslated: 20`, byte-identical to the same run on a stashed tree |
| `mise run build` | **not runnable** — `mise ERROR no tasks defined`; there is no mise task file in this repo |
| `vite build` (the real build, run directly) | **fails in this container for an environment reason, not the change**: `EMFILE: too many open files` from `lucide-react`'s per-icon ESM files. `ulimit -n` is 1024 and hard-capped, so it cannot be raised here; each rerun dies a few hundred modules further along at a different import. `tsc -b` — the type-checking half of `npm run build` — passes cleanly, and the setup notes record the build already failing before any of my changes. |

New tests, all of which fail without the change:

- counts the rows the list below renders, with their noun (and asserts the badge is the heading's sibling, i.e. actually *next to* it)
- agrees with the Total Proposals card
- says nothing while the list is still being read
- reads zero once the read lands empty
- counts a single row in the singular (`1 proposal`, not `1 proposals`)

```abca-verify
selector h1
selector #prioritization-row-count
text #prioritization-row-count "proposal"
no-console-errors
```

The three assertions are chosen to hold on an empty preview database: with no projects the read lands empty and the badge renders `0 proposals`, so the `text` line matches on the noun rather than on any seeded count. The badge's `id` is one I added, so it is a stable selector.

## Agent notes

**What went well.** This page turned out to have already solved the hard version of the problem twice, so the work was mostly a matter of not re-solving it differently. `StatsCards` had a long comment explaining exactly why "Total Proposals" counts rows and not documents, and why the other three cards dash rather than print `0` when the team read is uncountable — that reasoning transferred directly to the loading state of the badge, and reusing `allRows.length` (rather than counting projects, or `sortedRows`) is what let me write a test asserting the badge and the card agree.

**What was difficult.** The build. `mise run build` does not exist in this repo, and the real `vite build` cannot complete in this container: `lucide-react` ships one file per icon, the fd limit is 1024 and hard-capped, and the build dies with `EMFILE` at a different module each run. I verified the type-checking half (`tsc -b`) separately, which is the half that could plausibly be broken by a source change, and leaned on the test suite for the rest. Also worth knowing: `npx <pkg>` fails in this container (`npm error Lock compromised`) — `npm install` in `voc-datalake/frontend` and then `./node_modules/.bin/vitest` works.

**Conventions discovered.**
- Locale JSON is `json.dumps(..., indent=2, sort_keys=True)` exactly — keys sorted, two-space, trailing newline. I confirmed this by round-tripping all eight files before editing, which is worth doing: an unsorted insert produces a large spurious diff.
- Three separate gates watch i18n and they check different things: `localeParity.test.ts` (same key set everywhere, plus every English-pluralized key must have `_other` or a bare form in every locale), `i18nKeys.test.ts` (no colons in key paths — `:` is i18next's `nsSeparator` and this app depends on it), and `scripts/i18n-check.mjs` (missing/extra/empty/untranslated *and* unused keys). A new key must satisfy all three; the useful trick is to run `i18n-check` on a stashed tree and diff the totals, since it reports pre-existing noise.
- Keys must be **literals** inside `t(...)`. A ternary choosing between two keys inside the call makes `i18n-check` report both as unused — the existing code puts the condition outside the call and there are comments saying so.
- Comments in this codebase explain *why not* as much as *why*. I matched that register; the alternatives I rejected (bare numeral, `0` while loading, dashing like the cards) are written down at the code rather than only here.
- No `data-testid` anywhere in `src` — tests select by role and text. I added an `id` instead, which the page already uses for `aria-labelledby` targets and which the verify block needs.

- Adding a `mise.toml` (or setting `pipeline.buildCommand` to `npm run build`) would turn on the build gate that is currently inert. Bundling `lucide-react` imports, or raising the container fd limit, would make that build actually completable.
- `Prioritization.tsx` is ~1030 lines with four components in it; `PrioritizationHeader`, `StatsCards` and `SortControls` are independent and would each be easier to touch as their own file.
- `Prioritization.test.tsx` is ~2260 lines. Its `oneRowPerDocument` / `useLayout` / `R` proxy harness is genuinely good and worth reading before writing any new test on this page — but the file is at the point where splitting by concern (scores, read failures, notes, chrome) would help.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/prioritization.json
+++ b/voc-datalake/frontend/public/locales/de/prioritization.json
@@ -77,6 +77,8 @@
     "description": "Bewerten Sie jedes Dokument nach: <strong>Auswirkung</strong>, <strong>Markteinführungszeit</strong>, <strong>Strategische Passung</strong> und <strong>Konfidenz</strong>.",
     "title": "Priorisierungsrahmen"
   },
+  "headingCount_one": "{{count}} Vorschlag",
+  "headingCount_other": "{{count}} Vorschläge",
   "loading": "Dokumente werden geladen...",
   "noteTooLong": {
     "description": "Notizen sind auf {{max}} Zeichen begrenzt, und eine längere Notiz wird nicht gekürzt, sondern abgelehnt. Kürzen Sie die Notiz in der bearbeiteten Zeile und speichern Sie dann erneut. Es wurde noch nichts gespeichert.",
--- a/voc-datalake/frontend/public/locales/en/prioritization.json
+++ b/voc-datalake/frontend/public/locales/en/prioritization.json
@@ -77,6 +77,8 @@
     "description": "Score each document on: <strong>Impact</strong>, <strong>Time to Market</strong>, <strong>Strategic Fit</strong>, and <strong>Confidence</strong>.",
     "title": "Prioritization Framework"
   },
+  "headingCount_one": "{{count}} proposal",
+  "headingCount_other": "{{count}} proposals",
   "loading": "Loading documents...",
   "noteTooLong": {
     "description": "Notes are limited to {{max}} characters, and a longer note is refused rather than shortened. Trim the note on the row you edited, then save again. Nothing has been saved yet.",
--- a/voc-datalake/frontend/public/locales/es/prioritization.json
+++ b/voc-datalake/frontend/public/locales/es/prioritization.json
@@ -77,6 +77,8 @@
     "description": "Puntúe cada documento en: <strong>Impacto</strong>, <strong>Tiempo de comercialización</strong>, <strong>Ajuste estratégico</strong> y <strong>Confianza</strong>.",
     "title": "Marco de priorización"
   },
+  "headingCount_one": "{{count}} propuesta",
+  "headingCount_other": "{{count}} propuestas",
   "loading": "Cargando documentos...",
   "noteTooLong": {
     "description": "Las notas están limitadas a {{max}} caracteres, y una nota más larga se rechaza en lugar de recortarse. Acorte la nota de la fila que editó y vuelva a guardar. Todavía no se ha guardado nada.",
--- a/voc-datalake/frontend/public/locales/fr/prioritization.json
+++ b/voc-datalake/frontend/public/locales/fr/prioritization.json
@@ -77,6 +77,8 @@
     "description": "Évaluez chaque document sur : <strong>Impact</strong>, <strong>Délai de mise sur le marché</strong>, <strong>Adéquation stratégique</strong> et <strong>Confiance</strong>.",
     "title": "Cadre de priorisation"
   },
+  "headingCount_one": "{{count}} proposition",
+  "headingCount_other": "{{count}} propositions",
   "loading": "Chargement des documents...",
   "noteTooLong": {
     "description": "Les notes sont limitées à {{max}} caractères, et une note plus longue est refusée plutôt que tronquée. Raccourcissez la note de la ligne modifiée, puis enregistrez de nouveau. Rien n'a encore été enregistré.",
--- a/voc-datalake/frontend/public/locales/ja/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ja/prioritization.json
@@ -77,6 +77,8 @@
     "description": "各ドキュメントを次の基準で評価: <strong>インパクト</strong>、<strong>市場投入時間</strong>、<strong>戦略的適合性</strong>、<strong>確信度</strong>。",
     "title": "優先順位付けフレームワーク"
   },
+  "headingCount_one": "提案 {{count}} 件",
+  "headingCount_other": "提案 {{count}} 件",
   "loading": "ドキュメントを読み込み中...",
   "noteTooLong": {
     "description": "メモは{{max}}文字までで、これを超えるメモは切り詰められずに拒否されます。編集した行のメモを短くしてから、もう一度保存してください。まだ何も保存されていません。",
--- a/voc-datalake/frontend/public/locales/ko/prioritization.json
+++ b/voc-datalake/frontend/public/locales/ko/prioritization.json
@@ -77,6 +77,8 @@
     "description": "각 문서를 다음 항목으로 평가하세요: <strong>영향도</strong>, <strong>출시 시간</strong>, <strong>전략적 부합도</strong>, <strong>신뢰도</strong>.",
     "title": "우선순위 지정 프레임워크"
   },
+  "headingCount_one": "제안 {{count}}건",
+  "headingCount_other": "제안 {{count}}건",
   "loading": "문서 로드 중...",
   "noteTooLong": {
     "description": "메모는 {{max}}자까지 입력할 수 있으며, 이를 넘는 메모는 잘리지 않고 거부됩니다. 수정한 행의 메모를 줄인 후 다시 저장하세요. 아직 저장된 내용은 없습니다.",
--- a/voc-datalake/frontend/public/locales/pt/prioritization.json
+++ b/voc-datalake/frontend/public/locales/pt/prioritization.json
@@ -77,6 +77,8 @@
     "description": "Pontue cada documento em: <strong>Impacto</strong>, <strong>Tempo para o Mercado</strong>, <strong>Alinhamento Estratégico</strong> e <strong>Confiança</strong>.",
     "title": "Framework de Priorização"
   },
+  "headingCount_one": "{{count}} proposta",
+  "headingCount_other": "{{count}} propostas",
   "loading": "Carregando documentos...",
   "noteTooLong": {
     "description": "As notas são limitadas a {{max}} caracteres, e uma nota mais longa é recusada em vez de encurtada. Reduza a nota da linha que você editou e salve novamente. Nada foi salvo ainda.",
--- a/voc-datalake/frontend/public/locales/zh/prioritization.json
+++ b/voc-datalake/frontend/public/locales/zh/prioritization.json
@@ -77,6 +77,8 @@
     "description": "对每个文档进行评分：<strong>影响力</strong>、<strong>上市时间</strong>、<strong>战略契合度</strong>和<strong>置信度</strong>。",
     "title": "优先级排序框架"
   },
+  "headingCount_one": "{{count}} 个提案",
+  "headingCount_other": "{{count}} 个提案",
   "loading": "加载文档中...",
   "noteTooLong": {
     "description": "备注最多 {{max}} 个字符，超出长度的备注会被拒绝，而不会被截断。请缩短所编辑行的备注，然后重新保存。目前尚未保存任何内容。",
--- a/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
+++ b/voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx
@@ -1067,6 +1067,88 @@ describe('Prioritization', () => {
     })
   })
 
+  describe('the heading says how long the list is', () => {
+    /** The badge beside the `<h1>`, or null while it is withheld. */
+    const countBadge = () => document.getElementById('prioritization-row-count')
+
+    it('counts the rows the list below renders, with their noun', async () => {
+      // Two projects in the default harness, so two rows — p1's PRD and PR/FAQ are
+      // ONE row. Counting documents would say three, which is the miscount the stats
+      // cards already fixed once; this badge reads the same array the list maps over.
+      renderPrioritization()
+
+      await waitFor(() => {
+        expect(countBadge()).toHaveTextContent('2 proposals')
+      })
+      // Adjacent to the heading, not floated somewhere else on the page: the point is
+      // that a reader looking at the title can see the size without scrolling.
+      const heading = screen.getByRole('heading', { name: 'Prioritization' })
+      expect(heading.parentElement).toContainElement(countBadge())
+    })
+
+    it('agrees with the Total Proposals card', async () => {
+      // Two numbers for one quantity is how a page loses a reader's trust, which is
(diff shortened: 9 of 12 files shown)


perrozzi:

This is a frontend-only change: `PrioritizationHeader` gains a count badge next to the `<h1>` driven by two new props (`isLoading`, `rowCount={allRows.length}`), plus `headingCount_one`/`headingCount_other` in all eight locales and five new tests. The approach is right — reusing `allRows.length` so the badge can't disagree with the "Total Proposals" card, hiding rather than printing `0` during the initial read, and pluralizing with `count` interpolation all match existing conventions on this page. Two things need attention: the badge makes a false claim when the read *fails* (not just while it's in flight), and one of the new tests can pass vacuously.

### Blockers

- `voc-datalake/frontend/src/pages/Prioritization/Prioritization.tsx:461` (badge) and `:910` (wiring) — the badge is only withheld for `isLoading`. If the projects/scores read errors, TanStack Query leaves `isLoading` false while `allRows` is empty, so the page renders `0 proposals` beside an error state. That is exactly the reading the PR argues against ("the list finished and it is empty"), just for the failure path instead of the in-flight path. Gate on the error/no-data condition too (e.g. `isLoading || isError`, or drive it off whether the read landed at all), and add the matching regression test — the suite already has a read-failure section to hang it on.

### Should fix

- `Prioritization.test.tsx` — "agrees with the Total Proposals card" can pass without asserting anything:
  - `getByText('Total Proposals').previousElementSibling?.textContent` is optional-chained, and `expect(...).toContain(total ?? '')` makes `toContain('')` trivially true if the card's value node ever moves. Assert `expect(total).toBeTruthy()` (or read the value by a stable selector) before the comparison.
  - `toContain` on a bare numeral is also loose — `"12 proposals"` contains `"2"`. Compare against the full expected string, or at least anchor the numeral.
  - `within(grid as HTMLElement)` uses an `as` assertion on a possibly-null `closest()` result, against the repo's preference for type guards; `expect(grid).not.toBeNull()` first would both narrow and fail usefully.
- Accessibility rationale vs. implementation: the JSX comment argues the count needs its noun because "a screen reader announcing 'Prioritization 3' says nothing" — but the badge sits *outside* the `<h1>`, so the heading is still announced as "Prioritization" alone and the count is only reached in normal reading order. If the intent is for the heading to carry the size, put the span inside the heading or reference it with `aria-describedby` on the `<h1>`.
- The description notes `vite build` could not be completed locally (`EMFILE`). `tsc -b` passing covers most of the risk for a change this shape, but please confirm the production build on CI before merge rather than treating the build gate as inert.

### Nits

- The new tests select via `document.getElementById('prioritization-row-count')`, which departs from the file's role/text-based querying. Given the `id` exists for `aria-*`/verify purposes anyway this is defensible, but a `getByText(/proposals?$/)`-style query would keep the harness consistent.
- The ~20-line JSX comment largely restates the two prop doc comments added at `PrioritizationHeader`'s signature; one of the two locations would do.
- `headingCount_one` in `ja`/`ko`/`zh` is unreachable (those locales only have the `other` category). Harmless and required by `localeParity.test.ts`, just noting it's dead weight.
- Agreed with the description's own suggestion: `PrioritizationHeader` is now large enough with props/docs that extracting it out of the ~1030-line `Prioritization.tsx` would be a good follow-up.

---

perrozzi:

Since the last review the `isLoading` gate has been replaced by a new exported predicate `rowCountPublishable` (projects / details / savedScores / ensuredRows), wired into `PrioritizationHeader` as `countPublishable`, with 10 unit tests plus five new integration tests covering each failure path; the heading now carries a conditional `aria-describedby` pointing at the badge id; and the "agrees with Total Proposals" test was tightened. Both the blocker and the two substantive should-fix items are addressed.

### Resolved
- Blocker: the badge is now withheld on projects-read failure, fan-out failure, and "no row source landed", with unit tests and three integration regressions (`prioritizationUtils.ts:1335+`, `Prioritization.tsx:953+`).
- Vacuous card-agreement assertion fixed: `expect(total).toBe('2')` plus a full-string `toBe(\`${total} proposals\`)`, and the `closest()`/`as HTMLElement` cast is gone.
- Heading/count association implemented via `aria-describedby` gated on the same boolean, with `toHaveAccessibleName`/`toHaveAccessibleDescription` assertions and no dangling IDREF.

### Still open
- Confirming the production `vite build` on CI (local `EMFILE`) before merge.

### New findings
- Nit — `voc-datalake/frontend/src/pages/Prioritization/Prioritization.test.tsx`: "does not contradict the Total Proposals card when a read fails" and "says nothing when the project fan-out fails" set the same mock and assert the same thing (`countBadge()` null); one of them is redundant.
- Nit — `prioritizationUtils.ts:1335`: `rowCountPublishable` restates `collectRows`' short-circuit conditions rather than deriving from them, so a future change to `collectRows` can silently desync the gate. A comment in `collectRows` pointing back here (or sharing one "row source landed" helper) would guard against drift.
- Nit — the ~40-line docstring on a 6-line predicate plus the ~20-line JSX comment plus paragraph-length comments on nearly every new test is far above the surrounding file's density; the "why not" register is house style, but this could be trimmed by half without losing any rationale.
- Nit — `prioritizationUtils.test.ts:397` `notAList` launders `JSON.parse`'s `any` through a declared return type, which is the same escape hatch as the `as` cast flagged last time; fine for a wire-lie fixture, just noting it.
- Check — `within` may now be unused in `Prioritization.test.tsx` after the grid lookup was removed; lint is reported green, so presumably still used elsewhere, but worth a glance.

---

perrozzi:

Since the last review the changes are comment-only trims plus two structural tidies: the redundant "does not contradict the Total Proposals card when a read fails" test was deleted (its rationale folded into the fan-out test, which keeps the same assertion), and `collectRows` now carries a "MOVES WITH `rowCountPublishable`" note pointing at the gate. All three nits from the previous review are addressed; no behaviour changed.

### Resolved
- Duplicate test removed — the fan-out case now carries the "card is not asserted" rationale, with no coverage lost.
- `collectRows` gained the back-pointer to `rowCountPublishable`, guarding against the two staying silently out of sync.
- Comment density in `Prioritization.tsx`, `prioritizationUtils.ts` and both test files cut roughly in half without dropping any of the "why not" rationale.

### Still open
- Confirming the production `vite build` on CI (local `EMFILE`) before merge.

Nothing new. The delta is clean — comments only, plus one test deletion that leaves the withholding paths (projects read, fan-out, no row source, in-flight) each still covered by exactly one integration test and one unit case.

---

perrozzi:
### Production build: confirmed green

The description says `vite build` could not be completed (`EMFILE: too many open files` from `lucide-react`'s per-icon ESM files, with `ulimit -n` hard-capped at 1024 in that container), and the follow-up review carried "confirm the production build" as the one still-open item. That claim is stale — it was an environment limit, not the change.

Run on a host where `ulimit -n` is 1048575, against `35e4387a` in a clean worktree at this PR's head:

```
$ ./node_modules/.bin/tsc -b && ./node_modules/.bin/vite build
dist/assets/prioritizationUtils-zUD35Orn.js       26.16 kB │ gzip:   9.58 kB
...
dist/assets/index-DFu4xMh-.js                    461.35 kB │ gzip: 137.38 kB
✓ built in 3.66s
```

No `EMFILE`, both halves of `npm run build` green. (`npm run build` itself stops earlier in a bare worktree — its `prebuild` shells out to the parent package's `ts-node` for manifest generation, which is unrelated to this diff; the two generated files it writes are already committed, so `vite build` runs against the real inputs.)

Full gate set on the same tree:

| Command | Result |
|---|---|
| `npm run typecheck` (`tsc -b --noEmit`) | pass |
| `npm run lint` (`eslint . --max-warnings 0`) | pass |
| `npm run test -- --run` (whole frontend suite) | pass — 3081 tests, 185 files |
| `npm run test -- --run src/pages/Prioritization src/i18n` | pass — 380 tests, 12 files |
| `tsc -b && vite build` | pass — built in 3.66s |
| `npm run i18n:check` | output **byte-identical** to the pre-change tree |
| `npm run typecheck:tests` | 102 errors before **and** after, error sets identical modulo line numbers |

Two of those gates are already red on `development` and are not this PR's doing:

- `npm run typecheck:tests` — 102 pre-existing errors repo-wide (32 in `DataSourceWizard.test.tsx`, 11 in `configStore.test.ts`, 4 in `src/test/test-utils.tsx`, and 15 that were already in `prioritizationUtils.test.ts`). Verified unchanged by diffing the error sets with line and column numbers stripped. Worth naming as a consequence rather than a finding: while this gate is red, **type errors in test files are caught by nothing**, so the new tests' type safety rests on review alone.
- `npm run i18n:check` — exits 1 on the clean tree over two `projectDetail:prototypeLink.*` keys used in source but absent from the English catalogue, from an unrelated feature. Proved unchanged by `diff` of the full 740-line output, not by comparing the totals line. `headingCount_one` / `headingCount_other` appear in neither the missing nor the unused list, confirming the literal-key constraint still holds.

The blocker fix was also checked by reverting it rather than only by adding tests: with the gate put back to `countPublishable={!isLoading}`, five tests in the badge block fail, each showing the badge rendering `0 proposals`.

perrozzi:
Closing as a duplicate of #366, which is merged (`b9ebe44f`) and now deployed and verified live.

These two PRs are the **same task run twice**, not two designs. Both branches carry the same slug (`on-the-prioritization-page-the-heading-gives-no-se`) and, in their first commits, the same `Prompt-Version: 1c9c10e027a2` under different task ids, opened twelve minutes apart. #366's first commit is `chore(agent): save uncommitted work from session end` — an unfinished tree with no test file that did not compile (`rowCount` declared and read but absent from the destructure). This PR is the run that finished. #366 was then completed by hand.

Both runs also made the same two mistakes, which is worth recording rather than losing:

- the badge was gated on `isLoading` alone, so a **failed** read printed `0 proposals` about a backlog nobody had read;
- the prop doc claimed the badge and the "Total Proposals" card "cannot disagree".

#366's later commits fixed both. Its final gate withholds at zero, which covers every failure state as a side effect, and its doc now says the honest thing: the badge counts the **list** (`sortedRows.length`) while the card counts the **backlog** (`allRows.length`), *"the same unit, not a promise of the same number… the first row filter or search box put on this page should make them differ."*

That last point is why #366 is the better base, and it is the correction this PR still needs: it switched the basis to `allRows.length` while keeping a claim about the list, and `PRFAQList` actually renders `rows={sortedRows}`. Merging both would also put two badges beside one heading and two near-duplicate plural pairs in eight catalogues.

**What this PR did better, kept as follow-up candidates:**

1. **`aria-describedby` on the `<h1>`** pointing at the badge id, conditional so the IDREF never dangles. #366 leaves the heading's accessible name as the title alone and never associates the count, so on a heading jump a screen reader does not reach it. This is the one clear win here and it is small and self-contained.
2. **`rowCountPublishable`** — a unit-tested predicate over every input the count derives from (both project reads *and* the merged rows map), so the badge withholds when no read has earned a number but still publishes the honest `0` once a read lands empty. #366's withhold-at-zero is simpler and safe, but it can never show the honest empty state, which its own comment argues is the right call. Worth revisiting only if that judgement changes.

Both are recorded so they are not lost with the branch. No work here is wasted: the convergence on this PR is what surfaced the shared defects and produced the comparison that settled which base to keep.

perrozzi:
Superseded by #366 (merged as b9ebe44f, deployed and live-verified). Rationale in the comment above.

Issue #347, opened 2026-08-18
Harden setNested in fix-i18n.mjs so prototype protection is structural, not a denylist


perrozzi:
## Summary

`setNested` in `voc-datalake/frontend/scripts/fix-i18n.mjs` already guards against prototype pollution, but it does so in a shape CodeQL does not recognise as a sanitizer, so `js/prototype-pollution-utility` (medium, alert 6, first raised 2026-03-25) keeps firing on it. The guard should be restated so the protection is structural rather than a denylist the scanner has to infer.

## Current state

```js
const UNSAFE_KEYS = new Set(['__proto__', 'constructor', 'prototype'])

function setNested(obj, key, val) {
  const parts = key.split('.')
  if (parts.some(p => UNSAFE_KEYS.has(p))) return
  let cur = obj
  for (let i = 0; i < parts.length - 1; i++) {
    if (!cur[parts[i]] || typeof cur[parts[i]] !== 'object') cur[parts[i]] = {}
    cur = cur[parts[i]]
  }
  cur[parts[parts.length - 1]] = val
}
```

The early return covers every path segment, terminal and intermediate, against the three dangerous keys, so **this is not currently exploitable**. Two reasons the risk is low regardless:

- It is a local build script. It never ships to a browser, and it processes locale JSON that is already committed to this repository, so reaching it requires a malicious key in a tracked file.
- The denylist covers the keys that actually matter for pollution.

## Why change it anyway

A denylist is a maintenance liability: it is correct only as long as the list is complete, and nothing enforces that. Making the containers non-prototypal removes the class of bug instead of enumerating it, and has the side benefit of satisfying the scanner so the alert stops needing manual triage on every PR that carries this file.

Suggested direction:

- Create intermediate objects with `Object.create(null)` so there is no prototype to pollute.
- Keep the existing `UNSAFE_KEYS` early return as belt and braces rather than replacing it.
- If the parsed locale objects need to stay plain (for `JSON.stringify` round-tripping and deep-equality assertions), verify that with the existing i18n gates before switching — the scripts and their tests are sensitive to object shape.

## Acceptance criteria

- Writing a key containing `__proto__`, `constructor` or `prototype`, at any position in the path, cannot modify `Object.prototype`, and there is a check that fails if that protection is removed.
- The i18n scripts and their existing gates still pass, and generated locale files are byte-identical to before the change.
[...]

PR #370, opened 2026-08-23
fix(i18n): build setNested containers with null prototypes

slegarraga:
Addresses #347.

## What

`setNested` in `frontend/scripts/fix-i18n.mjs` now creates intermediate containers with `Object.create(null)` instead of `{}` literals, so there is no prototype to pollute regardless of what key reaches them. The `UNSAFE_KEYS` early return stays as defence in depth, because the two layers are genuinely complementary: with a prototypal root object (the real case, since `targetData` comes from `JSON.parse`), a leading `__proto__` segment resolves through the accessor on `Object.prototype` and would reach it even when every created container is null-prototypal. The denylist blocks that path; the structural layer removes the class of bug for everything else.

To make the protection testable against the real helper, `setNested` moves into `frontend/scripts/set-nested.mjs`. `fix-i18n.mjs` executes its whole main loop at import time (Bedrock calls included), so until extracted it could not be imported by any test.

## Why this clears alert 6

`js/prototype-pollution-utility` kept firing because the denylist shape is not something CodeQL recognizes as a sanitizer. With this change every assignment target that can be created by the function is prototype-less and the only remaining writes into prototypal objects are denied segments, which return before touching anything. Final confirmation of the alert clearing will come from this repo's CodeQL run on this PR; if anything residual reports, criterion 3 of the issue allows dismissal with reasoning recorded on the alert.

## Tests

New `frontend/scripts/set-nested.test.mjs` (house style, `node scripts/set-nested.test.mjs`, 10 cases):

- **Structural layer pinned:** created containers have null prototypes; reverting `Object.create(null)` to `{}` fails this test (verified via mutation run).
- **Denylist layer pinned:** pollution matrix with dangerous keys at leading/middle/trailing positions (`__proto__.x`, `a.__proto__.b`, `constructor.prototype.x`, ...) against both fresh and pre-populated prototypal trees; removing the early return fails it with `polluted` appearing on `Object.prototype` (verified via mutation run).
- **Denied keys write nothing anywhere**, including no own `__proto__` property.
- **Behaviour preservation:** writes land where they did before, existing subtree nodes are reused in place, last write wins, and `JSON.stringify` output is byte-identical to plain-literal trees using `saveJSON`'s exact format (2-space indent plus trailing newline), including a round trip through a real committed locale file. Generated locale files cannot change shape.

## Verification

```
node scripts/set-nested.test.mjs   # 10 passed
node scripts/i18n-check.mjs        # exit 0
node scripts/i18n-check.test.mjs   # 64 passed
```

No locale files are modified by this PR. The full Bedrock translation path was not executed locally (requires credentials); its output shape is what the serialization-equality tests pin.

### DIFF
--- a/docs/deployment.md
+++ b/docs/deployment.md
@@ -62,7 +62,7 @@ npm run check        # lint + typecheck:all + test + test:cdk + test:stream + te
 |---------|--------|
 | `npm run lint` | `lint:frontend` + `lint:stream` (ESLint) + `lint:python` (ruff over `lambda/`, `plugins/`) |
 | `npm run typecheck` | Frontend only — use `typecheck:all` for frontend + CDK + stream |
-| `npm run test` | Frontend Vitest only |
+| `npm run test` | Frontend Vitest plus i18n script regression tests |
 | `npm run test:cdk` | CDK Vitest (`voc-datalake`) |
 | `npm run test:stream` | Streaming chat Lambda Vitest |
 | `npm run test:backend` | Python pytest via `.venv/bin/python` |
--- a/package.json
+++ b/package.json
@@ -17,7 +17,7 @@
     "typecheck:cdk": "cd voc-datalake && npx tsc --noEmit",
     "typecheck:stream": "cd voc-datalake/lambda/stream && npx tsc --noEmit",
     "typecheck:all": "npm run typecheck && npm run typecheck:cdk && npm run typecheck:stream",
-    "test": "npm run test --prefix voc-datalake/frontend -- --run",
+    "test": "npm run test --prefix voc-datalake/frontend -- --run && npm run test:i18n-scripts --prefix voc-datalake/frontend",
     "test:backend": "cd voc-datalake && .venv/bin/python -m pytest -q --no-header --tb=short",
     "test:cdk": "npm run test --prefix voc-datalake",
     "test:stream": "npm run test --prefix voc-datalake/lambda/stream",
--- a/voc-datalake/frontend/package.json
+++ b/voc-datalake/frontend/package.json
@@ -13,6 +13,7 @@
     "test": "vitest",
     "test:watch": "vitest --watch",
     "test:coverage": "vitest --coverage",
+    "test:i18n-scripts": "node scripts/i18n-check.test.mjs && node scripts/set-nested.test.mjs",
     "test:ui": "vitest --ui",
     "typecheck": "tsc -b --noEmit",
     "typecheck:tests": "tsc --noEmit -p tsconfig.test.json",
--- a/voc-datalake/frontend/scripts/fix-i18n.mjs
+++ b/voc-datalake/frontend/scripts/fix-i18n.mjs
@@ -10,6 +10,8 @@ import { resolve, join, extname } from 'node:path'
 import { fileURLToPath } from 'node:url'
 import { execFileSync } from 'node:child_process'
 
+import { setNested } from './set-nested.mjs'
+
 const __dirname = fileURLToPath(new URL('.', import.meta.url))
 const LOCALES_DIR = resolve(__dirname, '..', 'public', 'locales')
 // Must match the shipped catalogues in public/locales/<lang>/ — `src/i18n/options.test.ts`
@@ -164,19 +166,6 @@ function flattenEntries(obj, prefix = '') {
 
 function getNested(obj, key) {
   return key.split('.').reduce((o, k) => o?.[k], obj)
-}
-
-const UNSAFE_KEYS = new Set(['__proto__', 'constructor', 'prototype'])
-
-function setNested(obj, key, val) {
-  const parts = key.split('.')
-  if (parts.some(p => UNSAFE_KEYS.has(p))) return
-  let cur = obj
-  for (let i = 0; i < parts.length - 1; i++) {
-    if (!cur[parts[i]] || typeof cur[parts[i]] !== 'object') cur[parts[i]] = {}
-    cur = cur[parts[i]]
-  }
-  cur[parts[parts.length - 1]] = val
 }
 
 // Resolve self-referencing plural values like "sidebar.messagesCount"
--- a/voc-datalake/frontend/scripts/set-nested.mjs
+++ b/voc-datalake/frontend/scripts/set-nested.mjs
@@ -7,8 +7,9 @@
  *
  * Two complementary protections:
  *
- * 1. Structural: intermediate containers are created with Object.create(null),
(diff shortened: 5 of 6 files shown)


perrozzi:

This PR extracts `setNested` from `frontend/scripts/fix-i18n.mjs` into `frontend/scripts/set-nested.mjs`, switches intermediate container creation to `Object.create(null)`, keeps the `UNSAFE_KEYS` denylist as defence in depth, and adds a standalone `set-nested.test.mjs` covering both layers plus serialization equivalence. The change is small, behaviour-preserving (JSON serialization only walks own enumerable properties, so saved locale files are unchanged), and the reasoning in the description about why the denylist is still load-bearing for prototypal roots is correct. No blockers.

### Should fix
- `voc-datalake/frontend/scripts/set-nested.mjs`: the existence check still reads through the prototype chain (`if (!cur[parts[i]] || typeof cur[parts[i]] !== 'object')`). Using `Object.hasOwn(cur, parts[i])` would mean inherited properties are never traversed or reused, which closes the residual prototypal-root path structurally rather than relying on the denylist — and is more likely to read as a sanitizer to CodeQL than the denylist did. Today the only inherited object-valued property reachable is `__proto__` (denied), so this is hardening rather than a live bug, but it removes the last dependence on the denylist for intermediate segments.
- The new test is a standalone `node scripts/set-nested.test.mjs` runner and the diff does not wire it into `package.json` scripts or CI. `i18n-check.test.mjs` presumably has the same shape, but please confirm the new file is actually executed by whatever runs the script tests — an unrun regression test does not protect the fix. (If script tests are invoked individually in CI, the workflow needs the new entry.)
- Confirm no downstream consumer of the trees built by `setNested` depends on prototype-bearing objects: `assert.deepStrictEqual` against `{}`-literal fixtures, `instanceof Object`, `obj.hasOwnProperty(...)`, or string coercion of a container would now behave differently. The PR states `i18n-check.test.mjs` (64 tests) passes, but `fix-i18n.mjs`'s own merge/sort/save path was not exercised end-to-end, and it is not visible in the diff.

### Nits
- `assert.equal(({}).prototype?.polluted, undefined)` is vacuous — `({}).prototype` is always `undefined`. Either drop it or assert on `Object.prototype.prototype`.
- `PROTO_SNAPSHOT` compares only own property *names* of `Object.prototype`, so an assignment that overwrites an existing member (e.g. a key ending in `__proto__.toString`) would not be detected. Snapshotting a name→value map, or asserting `typeof Object.prototype.toString === 'function'`, would make the matrix stricter.
- The 30-line header comment in `set-nested.mjs` is largely duplicated in the test file header; keeping the rationale in one place avoids the two drifting apart.
- `set-nested.test.mjs` reads `public/locales/en/common.json` and asserts `brandNewKeyFromTest` is absent; since both sides of the comparison derive from the same parse, the file dependency adds little beyond a possible spurious failure if the fixture path moves.
- Consider re-exporting `setNested` from `fix-i18n.mjs` or noting in `set-nested.mjs` that it is the single owner of the helper, so a future contributor does not reintroduce a local copy in another script.

---

perrozzi:

Since the last review the author added the `Object.hasOwn` own-property guard to the traversal (plus a new test pinning that inherited nodes are neither traversed nor reused), wired the script tests into `voc-datalake/frontend` (`test:i18n-scripts`) and the root `npm run test`, updated `docs/deployment.md`, and tightened the pollution matrix (dropped the vacuous `({}).prototype` assert, added a `Object.prototype.toString` identity check). Both actionable previous findings are addressed and the delta looks clean.

### Resolved
- `set-nested.mjs`: traversal now uses `Object.hasOwn`, so inherited properties are never reused — the residual prototypal-root path is closed structurally, with a dedicated test.
- Test wiring: `test:i18n-scripts` in `voc-datalake/frontend/package.json` is invoked by the root `test` script (and therefore `npm run check`), so the regression test actually runs; docs table updated to match.
- Nits: vacuous `({}).prototype?.polluted` assertion removed; `PROTO_NAMES_SNAPSHOT` + `OBJECT_TO_STRING` now catch overwrites of existing `Object.prototype` members as well as added names.

### Still open
- Confirmation that no downstream consumer of `setNested`-built trees relies on prototype-bearing containers (`deepStrictEqual` against `{}` fixtures, `obj.hasOwnProperty(...)`, string coercion) — `fix-i18n.mjs`'s merge/sort/save path is still not exercised end-to-end. Low risk given the serialization-equality tests, but unverified.

### New findings
None blocking. Two minor notes: `test:i18n-scripts` is not part of the frontend package's own `test` script, so a developer running tests from `voc-datalake/frontend` won't execute it (only the root `test` chains it); and `Object.hasOwn` requires Node ≥16.9 — fine for this repo's toolchain, just worth being aware of if any script runner pins something older.

---

slegarraga:
Addressed the follow-up in `82b85d9`.

What changed:
- `setNested` now uses `Object.hasOwn(cur, part)` before traversing a segment, so inherited object-valued properties are never reused. The new test `inherited subtree nodes are not traversed or reused` pins that exact residual path.
- The script tests are now wired into `voc-datalake/frontend` as `test:i18n-scripts`, and the root `npm run test` invokes them after frontend Vitest. Since root `check` already runs `npm run test`, this regression now participates in the standard gate.
- Tightened the pollution matrix: removed the vacuous `({}).prototype` assertion and added an `Object.prototype.toString` identity check so overwrites of existing prototype members are caught too.
- Updated the command table in `docs/deployment.md` so the documented root test gate matches reality.

Verification run on the pushed head:
- `npm ci --prefix voc-datalake/frontend`
- `npm run test:i18n-scripts --prefix voc-datalake/frontend`: 64 i18n tests passed, 11 set-nested tests passed
- `npm test`: 181 Vitest files passed, 2949 tests passed, then the i18n script tests ran again and passed
- `git diff --check`: clean
- added-line em dash scan: clean

Downstream consumer check: `setNested` is imported only by `frontend/scripts/fix-i18n.mjs` plus its tests. I also searched the frontend for `hasOwnProperty`, `instanceof Object`, and deep strict equality patterns around this path; nothing depends on prototype-bearing containers from `setNested`. The script still writes via `JSON.stringify(..., null, 2) + '\n'`, and the serialization equality test pins byte-identical output against a plain literal tree.

One unrelated local note: `npm run i18n:check --prefix voc-datalake/frontend` currently exits nonzero because of pre-existing English keys `projectDetail:prototypeLink.downloadHtml` and `projectDetail:prototypeLink.openNewTab`; that is separate from this helper hardening. The wired regression gate above is green.

Issue #359, opened 2026-08-20
Five mutating /integrations and /sources routes accept any authenticated user


perrozzi:
## Summary

Five mutating routes in `voc-datalake/lambda/api/integrations_handler.py` are
reachable by **any authenticated user**. Three of them read and write the same
shared Secrets Manager blob that PR #307 just admin-gated for the credentials
routes, so that gate is incomplete: a caller who cannot `PUT` credentials can
still `POST` an app configuration into the same secret.

The sharpest one is unrelated to the secret: **`PUT /sources/{source}/disable`
lets any authenticated user silently stop all data ingestion.**

Found while verifying the admin gate added in #307, by probing every route in that
handler with an admin token and a non-admin token.

## Measured

Account `{account}`, {region}, against the deployed API. Admin =
`<admin-user>` (in `admins`), non-admin = `<non-admin-user>` (no groups).

| route | admin | non-admin | effect |
|---|---|---|---|
| `GET /integrations/status` | 200 | **403** | gated by #307 |
| `GET /integrations/{source}/credentials` | 200 | **403** | gated by #307 |
| `PUT /integrations/{source}/credentials` | 200 | **403** | gated by #307 |
| `GET /integrations/{source}/apps` | 200 | **200** | reads the shared secret |
| `POST /integrations/{source}/apps` | 400 | **400** | **writes** the shared secret |
| `DELETE /integrations/{source}/apps/{app_id}` | 200 | **200** | **writes** the shared secret |
| `POST /sources/{source}/run` | 200 | **200** | invokes an ingestor Lambda (ingests data, incurs Bedrock cost) |
| `PUT /sources/{source}/enable` | 200 | **200** | enables an EventBridge schedule |
| `PUT /sources/{source}/disable` | 200 | **200** | **disables an EventBridge schedule — stops ingestion** |

### Why 400 rather than 403 is the proof, not a counter-example

`require_admin()` runs as the first statement of a gated route, before any input
handling. So a **validation** error means the request got past authorization and
into the handler body. The `POST` probe deliberately sends `{}` — no `app` field —
so it is refused by input validation without writing anything. The shared secret
was byte-identical before and after the whole probe run (2153 bytes, 26 keys).

## Why this was not fixed in #307

Gating these routes decides **who may configure a data source**, which is a product
question rather than a bug fix, and it needs coordinated frontend work:

- The Scrapers page is **not** admin-only (`frontend/src/routes.tsx:105`, versus
[...]

PR #307, opened 2026-08-09
fix(integrations): namespace-scope, allowlist and admin-gate the credentials endpoint

perrozzi:
## Summary

Resolves issue #261. Three security improvements to `voc-datalake/lambda/api/integrations_handler.py`:

- **Remove unprefixed fallback on read path**: the GET handler fell back to bare unnamespaced key lookups when the prefixed form was absent, allowing callers to exfiltrate top-level secret keys belonging to other features (e.g. `webscraper_api_key` or `other_feature_blob`). The fallback is removed outright; only `<source>_<key>` is consulted.
- **Key-form validation on write path**: the PUT handler previously accepted any caller-supplied key. Keys are now validated against a conservative regex (`^[a-z0-9][a-z0-9_]{0,62}[a-z0-9]$|^[a-z0-9]$`) before any write, and requests are capped at `MAX_CREDENTIAL_KEYS_PER_REQUEST = 20`. Invalid keys return 400 before the secret is read, let alone written.
- **Admin gate on both routes**: both GET and PUT now call `require_admin()` (the existing shared helper in `shared/api.py`) as their first operation. Non-admin callers receive 403.

## The allowlist decision

Plugin manifests declare each plugin's configuration fields and are the natural source of truth, but they are consumed at CDK synth time — this Lambda cannot see them at runtime. Adding a CDK environment variable to bridge that gap is explicitly out of scope (another change is in flight against the API stack).

The approach taken is **key-form validation**: a conservative character class (lowercase letters, digits, underscores only), a length bound (1–64 chars), and a rejection of keys that start or end with an underscore (preventing namespace-prefix confusion). A cap of 20 keys per request bounds the write surface.

A manifest-derived allowlist is the stronger fix and is left as follow-up. The form-based approach is chosen because it requires no CDK changes, is fully runtime-checkable, and eliminates the main unbounded-write risk while the manifest integration is designed properly.

## Read response narrowing (not done here)

The issue notes that returning only configured/not-configured status (rather than actual values) would be safer. This is very likely correct for the long term, but changes the contract the frontend consumes. It is noted as recommended follow-up and should be done as a separate change with coordinated frontend work.

## Regression mapping — which test catches which revert

| Revert | Caught by |
|--------|-----------|
| Remove `require_admin` from GET | `TestAdminGateOnCredentialsRoutes::test_non_admin_read_rejected` |
| Remove `require_admin` from PUT | `TestAdminGateOnCredentialsRoutes::test_non_admin_write_rejected` |
[...]

PR #371, opened 2026-08-23
fix(integrations): admin-gate the mutating app-config and source-schedule routes

slegarraga:
Addresses phases 1 and 2 of #359. Phase 3 remains open for the product decision in the issue.

Backend:
- Added `require_admin(app.current_event.raw_event)` as the first statement in `POST /integrations/<source>/apps`, `DELETE /integrations/<source>/apps/<app_id>`, `PUT /sources/<source>/enable`, and `PUT /sources/<source>/disable`.
- Added negative tests for non-admin writes, positive admin controls that assert persistence, no-touch assertions for Secrets Manager and EventBridge, and a non-admin GET app-config read test.
- Redacts configured secret app-config fields before returning GET app-config payloads, plus a manifest lockstep test so future secret fields must be added to redaction coverage.
- Left `POST /sources/<source>/run` ungated with an inline phase 3 comment.

Frontend:
- Non-admin users keep read-only Scrapers access, but app-config add/edit/delete and schedule toggles are hidden or disabled.
- App-config control copy uses `pluginConfig.*` i18n keys across all eight locales.
- Added `PluginConfigModal`, `AppConfigCard`, and `AppConfigList` wiring coverage, with store reset in the new test.

Verification on final SHA `44f550e`:
- API suite: 2142 passed.
- Frontend suite: 187 files, 3068 tests passed.
- `npm run typecheck`: exit 0.
- `npm run lint -- --max-warnings 0`: exit 0.
- `ruff check` on touched backend files: all checks passed.
- Focused locale assertion: new `pluginConfig.*` app-control keys are present in all Scrapers locales.
- `git diff --check`: clean. No em dashes or truncation artifacts in diff/published text.

### DIFF
--- a/voc-datalake/frontend/public/locales/de/scrapers.json
+++ b/voc-datalake/frontend/public/locales/de/scrapers.json
@@ -161,12 +161,24 @@
     "titlePlaceholder": "Bewertungstitel (optional)"
   },
   "pluginConfig": {
+    "adminRequired": "Administratorzugriff erforderlich",
+    "addApp": "App hinzufügen",
+    "addFirstApp": "Erste App hinzufügen",
+    "addNewApp": "Neue App hinzufügen",
+    "cancel": "Abbrechen",
+    "configuredApps": "Konfigurierte Apps",
     "automaticSchedule": "Automatischer Zeitplan",
     "close": "Schließen",
     "configuration": "Konfiguration",
+    "deleteAppMessage": "Möchten Sie diese App-Konfiguration wirklich entfernen?",
+    "deleteAppTitle": "App löschen",
+    "delete": "Löschen",
     "disabled": "Deaktiviert",
+    "edit": "Bearbeiten",
+    "editApp": "App bearbeiten",
     "enabled": "Aktiviert",
     "hide": "Ausblenden",
+    "noAppsConfigured": "Noch keine Apps konfiguriert",
     "runNow": "Jetzt ausführen",
     "runTriggered": "Ausführung gestartet",
     "save": "Speichern",
@@ -174,7 +186,8 @@
     "scheduleDescription": "Diese Quelle nach konfiguriertem Zeitplan ausführen",
     "select": "Auswählen...",
     "show": "Anzeigen",
-    "test": "Testen"
+    "test": "Testen",
+    "unnamedApp": "Unbenannte App"
   },
   "refresh": "Aktualisieren",
   "status": {
--- a/voc-datalake/frontend/public/locales/en/scrapers.json
+++ b/voc-datalake/frontend/public/locales/en/scrapers.json
@@ -161,12 +161,24 @@
     "titlePlaceholder": "Review title (optional)"
   },
   "pluginConfig": {
+    "adminRequired": "Admin access required",
+    "addApp": "Add App",
+    "addFirstApp": "Add your first app",
+    "addNewApp": "Add New App",
+    "cancel": "Cancel",
+    "configuredApps": "Configured Apps",
     "automaticSchedule": "Automatic Schedule",
     "close": "Close",
     "configuration": "Configuration",
+    "deleteAppMessage": "Are you sure you want to remove this app configuration?",
+    "deleteAppTitle": "Delete App",
+    "delete": "Delete",
     "disabled": "Disabled",
+    "edit": "Edit",
+    "editApp": "Edit App",
     "enabled": "Enabled",
     "hide": "Hide",
+    "noAppsConfigured": "No apps configured yet",
     "runNow": "Run Now",
     "runTriggered": "Run triggered",
     "save": "Save",
@@ -174,7 +186,8 @@
     "scheduleDescription": "Run this source on its configured schedule",
     "select": "Select...",
     "show": "Show",
-    "test": "Test"
+    "test": "Test",
+    "unnamedApp": "Unnamed App"
   },
   "refresh": "Refresh",
   "status": {
--- a/voc-datalake/frontend/public/locales/es/scrapers.json
+++ b/voc-datalake/frontend/public/locales/es/scrapers.json
@@ -172,12 +172,24 @@
     "titlePlaceholder": "Título de la reseña (opcional)"
   },
   "pluginConfig": {
+    "adminRequired": "Se requiere acceso de administrador",
+    "addApp": "Agregar app",
+    "addFirstApp": "Agrega tu primera app",
+    "addNewApp": "Agregar nueva app",
+    "cancel": "Cancelar",
+    "configuredApps": "Apps configuradas",
     "automaticSchedule": "Programación automática",
     "close": "Cerrar",
     "configuration": "Configuración",
+    "deleteAppMessage": "¿Seguro que quieres eliminar esta configuración de app?",
+    "deleteAppTitle": "Eliminar app",
+    "delete": "Eliminar",
     "disabled": "Deshabilitado",
+    "edit": "Editar",
+    "editApp": "Editar app",
     "enabled": "Habilitado",
     "hide": "Ocultar",
+    "noAppsConfigured": "Aún no hay apps configuradas",
     "runNow": "Ejecutar ahora",
     "runTriggered": "Ejecución iniciada",
     "save": "Guardar",
@@ -185,7 +197,8 @@
     "scheduleDescription": "Ejecutar esta fuente según su programación configurada",
     "select": "Seleccionar...",
     "show": "Mostrar",
-    "test": "Probar"
+    "test": "Probar",
+    "unnamedApp": "App sin nombre"
   },
   "refresh": "Actualizar",
   "status": {
--- a/voc-datalake/frontend/public/locales/fr/scrapers.json
+++ b/voc-datalake/frontend/public/locales/fr/scrapers.json
@@ -172,12 +172,24 @@
     "titlePlaceholder": "Titre de l'avis (optionnel)"
   },
   "pluginConfig": {
+    "adminRequired": "Accès administrateur requis",
+    "addApp": "Ajouter une application",
+    "addFirstApp": "Ajoutez votre première application",
+    "addNewApp": "Ajouter une nouvelle application",
+    "cancel": "Annuler",
+    "configuredApps": "Applications configurées",
     "automaticSchedule": "Planification automatique",
     "close": "Fermer",
     "configuration": "Paramétrage",
+    "deleteAppMessage": "Voulez-vous vraiment supprimer cette configuration d'application ?",
+    "deleteAppTitle": "Supprimer l'application",
+    "delete": "Supprimer",
     "disabled": "Désactivé",
+    "edit": "Modifier",
+    "editApp": "Modifier l'application",
     "enabled": "Activé",
     "hide": "Masquer",
+    "noAppsConfigured": "Aucune application configurée",
     "runNow": "Exécuter maintenant",
     "runTriggered": "Exécution lancée",
     "save": "Enregistrer",
@@ -185,7 +197,8 @@
     "scheduleDescription": "Exécuter cette source selon sa planification configurée",
     "select": "Sélectionner...",
     "show": "Afficher",
-    "test": "Tester"
+    "test": "Tester",
+    "unnamedApp": "Application sans nom"
   },
   "refresh": "Actualiser",
   "status": {
--- a/voc-datalake/frontend/public/locales/ja/scrapers.json
+++ b/voc-datalake/frontend/public/locales/ja/scrapers.json
@@ -161,12 +161,24 @@
     "titlePlaceholder": "レビュータイトル（オプション）"
   },
   "pluginConfig": {
+    "adminRequired": "管理者アクセスが必要です",
+    "addApp": "アプリを追加",
+    "addFirstApp": "最初のアプリを追加",
+    "addNewApp": "新しいアプリを追加",
+    "cancel": "キャンセル",
+    "configuredApps": "設定済みアプリ",
     "automaticSchedule": "自動スケジュール",
     "close": "閉じる",
     "configuration": "設定",
+    "deleteAppMessage": "このアプリ設定を削除してもよろしいですか？",
+    "deleteAppTitle": "アプリを削除",
+    "delete": "削除",
     "disabled": "無効",
+    "edit": "編集",
+    "editApp": "アプリを編集",
     "enabled": "有効",
     "hide": "非表示",
+    "noAppsConfigured": "アプリはまだ設定されていません",
     "runNow": "今すぐ実行",
     "runTriggered": "実行が開始されました",
     "save": "保存",
@@ -174,7 +186,8 @@
     "scheduleDescription": "設定されたスケジュールでこのソースを実行",
     "select": "選択...",
     "show": "表示",
-    "test": "テスト"
+    "test": "テスト",
+    "unnamedApp": "名前のないアプリ"
   },
   "refresh": "更新",
   "status": {
--- a/voc-datalake/frontend/public/locales/ko/scrapers.json
+++ b/voc-datalake/frontend/public/locales/ko/scrapers.json
@@ -161,12 +161,24 @@
     "titlePlaceholder": "리뷰 제목 (선택사항)"
   },
   "pluginConfig": {
+    "adminRequired": "관리자 권한이 필요합니다",
+    "addApp": "앱 추가",
+    "addFirstApp": "첫 앱 추가",
+    "addNewApp": "새 앱 추가",
+    "cancel": "취소",
+    "configuredApps": "구성된 앱",
     "automaticSchedule": "자동 일정",
     "close": "닫기",
     "configuration": "구성",
+    "deleteAppMessage": "이 앱 구성을 제거하시겠습니까?",
+    "deleteAppTitle": "앱 삭제",
+    "delete": "삭제",
     "disabled": "비활성화됨",
+    "edit": "편집",
+    "editApp": "앱 편집",
     "enabled": "활성화됨",
     "hide": "숨기기",
+    "noAppsConfigured": "아직 구성된 앱이 없습니다",
     "runNow": "지금 실행",
     "runTriggered": "실행이 시작되었습니다",
     "save": "저장",
@@ -174,7 +186,8 @@
     "scheduleDescription": "구성된 일정에 따라 이 소스 실행",
     "select": "선택...",
     "show": "표시",
-    "test": "테스트"
+    "test": "테스트",
+    "unnamedApp": "이름 없는 앱"
   },
   "refresh": "새로고침",
   "status": {
--- a/voc-datalake/frontend/public/locales/pt/scrapers.json
+++ b/voc-datalake/frontend/public/locales/pt/scrapers.json
@@ -172,12 +172,24 @@
     "titlePlaceholder": "Título da avaliação (opcional)"
   },
   "pluginConfig": {
+    "adminRequired": "Acesso de administrador necessário",
+    "addApp": "Adicionar app",
+    "addFirstApp": "Adicione seu primeiro app",
+    "addNewApp": "Adicionar novo app",
+    "cancel": "Cancelar",
+    "configuredApps": "Apps configurados",
     "automaticSchedule": "Agendamento automático",
     "close": "Fechar",
     "configuration": "Configuração",
+    "deleteAppMessage": "Tem certeza de que deseja remover esta configuração de app?",
+    "deleteAppTitle": "Excluir app",
+    "delete": "Excluir",
     "disabled": "Desabilitado",
+    "edit": "Editar",
+    "editApp": "Editar app",
     "enabled": "Habilitado",
     "hide": "Ocultar",
+    "noAppsConfigured": "Nenhum app configurado ainda",
     "runNow": "Executar agora",
     "runTriggered": "Execução iniciada",
     "save": "Salvar",
@@ -185,7 +197,8 @@
     "scheduleDescription": "Executar esta fonte no agendamento configurado",
     "select": "Selecionar...",
     "show": "Mostrar",
-    "test": "Testar"
+    "test": "Testar",
+    "unnamedApp": "App sem nome"
   },
   "refresh": "Atualizar",
   "status": {
--- a/voc-datalake/frontend/public/locales/zh/scrapers.json
+++ b/voc-datalake/frontend/public/locales/zh/scrapers.json
@@ -161,12 +161,24 @@
     "titlePlaceholder": "评论标题（可选）"
   },
   "pluginConfig": {
+    "adminRequired": "需要管理员权限",
+    "addApp": "添加应用",
+    "addFirstApp": "添加第一个应用",
+    "addNewApp": "添加新应用",
+    "cancel": "取消",
+    "configuredApps": "已配置的应用",
     "automaticSchedule": "自动计划",
     "close": "关闭",
     "configuration": "配置",
+    "deleteAppMessage": "确定要移除此应用配置吗？",
+    "deleteAppTitle": "删除应用",
+    "delete": "删除",
     "disabled": "已禁用",
+    "edit": "编辑",
+    "editApp": "编辑应用",
     "enabled": "已启用",
     "hide": "隐藏",
+    "noAppsConfigured": "尚未配置应用",
     "runNow": "立即运行",
     "runTriggered": "运行已触发",
     "save": "保存",
@@ -174,7 +186,8 @@
     "scheduleDescription": "按配置的计划运行此源",
     "select": "选择...",
     "show": "显示",
-    "test": "测试"
+    "test": "测试",
+    "unnamedApp": "未命名应用"
   },
   "refresh": "刷新",
   "status": {
--- a/voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.test.tsx
+++ b/voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.test.tsx
@@ -1,0 +1,56 @@
+import { describe, it, expect, vi } from 'vitest'
+import { render, screen } from '@testing-library/react'
+import { AppConfigCard } from './AppConfigComponents'
+import type { PluginManifest } from '../../plugins/types'
+
+const plugin: PluginManifest = {
+  id: 'app_reviews_android',
+  name: 'Android App Reviews',
+  icon: 'Android',
+  description: 'Collect reviews from Google Play Store',
+  category: 'reviews',
+  config: [],
+  hasIngestor: true,
+  hasWebhook: false,
+  hasS3Trigger: false,
+  version: '1.0.0',
+  enabled: true,
+}
+
+const app = {
+  id: 'app-1',
+  app_name: 'Zara',
+  package_name: 'com.inditex.zara',
+  frequency_minutes: '1440',
+  max_reviews_per_run: '500',
+}
+
+function renderCard(canManage: boolean) {
+  return render(
+    <AppConfigCard
+      app={app}
+      plugin={plugin}
+      canManage={canManage}
+      onEdit={vi.fn()}
+      onDelete={vi.fn()}
+      onRun={vi.fn()}
+      isRunning={false}
+    />,
+  )
+}
+
+describe('AppConfigCard', () => {
+  it('shows edit and delete actions for admins', () => {
+    renderCard(true)
+    expect(screen.getByTestId('app-config-edit')).toBeInTheDocument()
+    expect(screen.getByTestId('app-config-delete')).toBeInTheDocument()
+    expect(screen.getByTestId('app-config-run').getAttribute('aria-label')).toBeTruthy()
+  })
+
+  it('hides edit and delete actions for non-admins but keeps Run Now', () => {
+    renderCard(false)
+    expect(screen.queryByTestId('app-config-edit')).not.toBeInTheDocument()
+    expect(screen.queryByTestId('app-config-delete')).not.toBeInTheDocument()
+    expect(screen.getByTestId('app-config-run')).toBeInTheDocument()
+  })
+})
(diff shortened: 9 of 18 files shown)


perrozzi:

This PR adds `require_admin` gates to the four mutating integrations routes (`POST`/`DELETE` app configs, `PUT` source enable/disable), hides/disables the corresponding frontend affordances for non-admins, and adds backend regression tests with both negative (403) and positive (admin) controls. The backend change is minimal, correctly placed as the first statement in each handler, and mirrors the existing credentials gate; overall this looks solid. My comments are about the new hardcoded UI string, a missing frontend regression test, and hermeticity of one test.

### Should fix

- `voc-datalake/frontend/src/pages/Scrapers/PluginConfigModal.tsx`: `title={!isAdmin ? 'Admin access required' : undefined}` is a new user-visible string hardcoded in English, while the same component already uses `useTranslation('scrapers')` for the toggle labels. Add a key (e.g. `pluginConfig.adminRequired`) to all locale files and use `t(...)`. The test asserts the literal, so it should assert the translated value instead (or the key via the test i18n setup). Same literal reportedly exists in `GeneratorConfigModal` — worth extracting one shared key rather than duplicating.
- No frontend regression test covers the `canManage` gating on the Scrapers page itself (`AppConfigComponents.tsx` / `Scrapers.tsx`). Reverting `canManage={isAdmin}` in `Scrapers.tsx` would leave the whole suite green, since only `PluginConfigModal.test.tsx` was extended. Add a vitest case asserting Edit/Delete are absent and Run is present for a non-admin `AppConfigCard`.
- `test_non_admin_write_rejected` (`test/test_integrations_security.py`) is the only new negative test that does not patch `integrations_handler.secretsmanager`. If the gate is reverted, that test exercises the real client instead of failing fast on an assertion; patch it like the DELETE case and add `put_secret_value.assert_not_called()` so the mutation signal is "no write happened", not just "status != 403".

### Nits

- Prop naming is inconsistent within the same file: `ScheduleToggle` takes `isAdmin` while `AppCard`/`AppListSection` take `canManage`. Pick one (`canManage` reads better as a capability flag decoupled from the auth source).
- The PR description says the admin POST control verifies "secret write verified through the mock", but `test_admin_write_succeeds` only asserts `success is True`; it never checks `put_secret_value`. Add the assertion or adjust the claim.
- `test_non_admin_delete_rejected_and_secret_untouched` docstring reasons about the unset mock raising; that reasoning is incidental (a `MagicMock` would fail later in `json.loads`) and the explicit `assert_not_called()` checks are what actually prove it. Consider trimming the docstring.
- The diff does not show an import for `useIsAdmin` in `Scrapers.tsx` (unlike `PluginConfigModal.tsx`, where it is added). Presumably it is already imported there — worth a double check that `tsc -b` was run after the last edit.
- `enable_source`/`disable_source` are now admin-gated while `POST /sources/<source>/run` stays open by design; a short comment in the handler noting that deliberate asymmetry (issue #359 phase 3) would help the next reader.

---

perrozzi:

This PR adds `require_admin(...)` gates to the four mutating integration routes (`POST`/`DELETE` app configs, `PUT` source enable/disable) and threads an `isAdmin`/`canManage` flag through the Scrapers UI so non-admins no longer see affordances that would now 403. The backend change matches the `#307` precedent exactly and the Python regression tests are well-targeted (negative + positive per route, with `assert_not_called` proof that the handler short-circuits before touching Secrets Manager/EventBridge). The main problem is on the frontend: a new user-visible string is hardcoded in English.

### Blockers

- `voc-datalake/frontend/src/pages/Scrapers/PluginConfigModal.tsx`: `title={!isAdmin ? 'Admin access required' : undefined}` is a hardcoded user-visible string in a component that already has `const { t } = useTranslation('scrapers')`. Per the repo i18n rule this needs a key (e.g. `pluginConfig.adminRequired`) added for **all** locales. The i18n check script passing only means no key is missing/unused — it can't see a literal. Note the corresponding assertion in `PluginConfigModal.test.tsx` (`toHaveAttribute('title', 'Admin access required')`) hardcodes the same literal and will need to assert the key/translated value instead.

### Should fix

- `voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.tsx`: `AppConfigCard` gained a required `canManage` prop that hides Edit/Delete, but there is no vitest coverage for it. The `PluginConfigModal` non-admin cases were added; add the equivalent for `AppConfigCard` (or the `AppConfigList` render in `Scrapers.tsx`) so the gate can't silently regress.
- Confirm no non-admin code path calls the newly gated schedule routes as a side effect. If any flow (e.g. first app save, or a plugin card toggle elsewhere) issues `PUT /sources/<source>/enable` outside an admin session, it will now fail with a 403 that the UI does not surface. A quick grep of the frontend API client callers for `enable`/`disable` in the PR description would settle this.
- `POST /sources/<source>/run` remains ungated. The deferral to the product decision in #359 is reasonable for read routes, but `run` is a mutating, cost-incurring action reachable by any authenticated user; worth calling out explicitly in the issue/docs so it isn't mistaken for an oversight.

### Nits

- `PluginConfigModal.tsx`: `readonly plugin: PluginManifest;` gained a trailing semicolon while the two sibling members have none — the file already mixes both styles, but this line was touched needlessly.
- `PluginConfigModal.test.tsx` adds a stray extra blank line after the `vi.mock('../../store/configStore', ...)` block.
- A `title` attribute on a `disabled` input is not reliably announced by screen readers and tooltips are suppressed on disabled controls in some browsers; consider wrapping in a labelled element or adding `aria-describedby` instead.
- `test_non_admin_write_rejected` does not patch `integrations_handler.secretsmanager`, so unlike the DELETE test it can't assert the secret was untouched. Adding the same `assert_not_called` pair would make the two symmetric.
- If there is an API reference or OpenAPI doc enumerating which routes require the `admins` group, it should be updated alongside these four routes.

---

perrozzi:

This PR adds `require_admin` gates to four mutating routes in `voc-datalake/lambda/api/integrations_handler.py` (app-config create/delete, source enable/disable), threads an explicit `isAdmin`/`canManage` prop through `PluginConfigModal`/`AppConfigCard` so the UI doesn't surface actions that will now 403, and adds an `adminRequired` i18n key to all eight locales. The backend change follows the existing `require_admin(app.current_event.raw_event)` pattern and is placed before any validation or Secrets Manager access, so it fails closed; the new pytest classes include both negative and positive-path controls plus "secret/rule untouched" assertions. Overall this looks clean and correctly scoped — no blockers.

### Should fix

- `voc-datalake/frontend/src/pages/Scrapers/PluginConfigModal.tsx` / `AppConfigComponents.tsx`: the touched components still carry hardcoded English strings (`"Configured Apps"`, `"Add App"`, `"Add your first app"`, `"No apps configured yet"`, `title="Edit"`, `title="Delete"`, `title="Run now"`). You correctly moved `GeneratorConfigModal`'s `'Admin access required'` to `t('pluginConfig.adminRequired')`; the newly-gated affordances right next to it should get keys too, otherwise the admin gate is localized while the controls it hides are not.
- The new tests key off those English literals (`screen.getByTitle('Edit')`, `getByText('Admin access required')` in `PluginConfigModal.test.tsx`, `AppConfigComponents.test.tsx`). Once the strings above are i18n'd these break. Prefer `getByRole('button', { name: ... })` against the resolved translation, or assert on a `data-testid`/`aria-label` that isn't the display copy.
- Please confirm the remaining mutating routes in `integrations_handler.py` (webhook/credential-adjacent writes, the run/trigger path behind the still-open "Run now" button) are either already gated or explicitly part of the deferred phase 3. The PR body says non-admins can still run sources, but a viewer being able to trigger paid ingestion runs is worth an explicit note in the issue so it isn't silently accepted.

### Nits

- `ScheduleToggle` uses a module-hardcoded DOM id (`'plugin-config-schedule-admin-required'`) for `aria-describedby`. `useId()` is collision-safe if the toggle is ever rendered more than once.
- The `<label>` wrapping the schedule checkbox keeps `cursor-pointer` even when the input is disabled; the `disabled:cursor-not-allowed` on the input won't win over the label's cursor.
- `test_admin_write_succeeds` only asserts `success is True`; the delete counterpart verifies the written `SecretString`. Mirroring that assertion would make the positive-path control stronger.
- `PluginConfigModal` still carries `// eslint-disable-next-line complexity`; the added prop plumbing is a reasonable moment to split the schedule/app sections out, though not required here.
- Consider a non-admin test asserting the empty-state `"Add your first app"` link is hidden — that branch is gated in the diff but untested.

---

perrozzi:

This PR adds `require_admin` gates to the four mutating integration routes (`POST`/`DELETE /integrations/<source>/apps`, `PUT /sources/<source>/{enable,disable}`) following the existing #307 pattern, threads an explicit `isAdmin` prop through `PluginConfigModal`/`AppConfigCard` to hide or disable mutating affordances, and moves several hardcoded English strings into `scrapers:pluginConfig.*` across all 8 locales. Backend regression tests include both negative (403) and positive-path controls plus assertions that the secret/EventBridge rule is untouched, which matches the repo's mutation-proof expectation. Overall this is clean and well-scoped; my findings are mostly i18n-completeness and consistency.

### Should fix

- **`card.*` i18n keys are used but not added.** `voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.tsx` now renders `t('card.runNow')`, `t('card.edit')`, `t('card.delete')`, and `PluginConfigModal.tsx` uses `t('card.edit')`, `t('card.delete')` plus `confirmLabel={t('card.delete')}`; `Scrapers.tsx` also uses `t('card.delete')`. The locale diffs only add keys under `pluginConfig.*`. If `card.edit` / `card.delete` / `card.runNow` don't already exist in every `public/locales/*/scrapers.json`, these become raw-key tooltips/aria-labels/confirm buttons. Please confirm they exist in all 8 locales (the PR's focused i18n check only covered `pluginConfig.adminRequired`), or add them. A `pluginConfig.delete`-style key would also be more consistent with the rest of this change than reaching into the `card` namespace.
- **The new frontend test doesn't cover translated output.** `AppConfigComponents.test.tsx` only asserts `data-testid` presence, so a missing `card.*` key wouldn't fail. One assertion on the rendered `aria-label`/`title` text (as `PluginConfigModal.test.tsx` does with `'Admin access required'`) would close the gap above.
- **`Run Now` remains reachable for non-admins in both surfaces while `POST /sources/<source>/run` is ungated.** This is documented as deferred to phase 3, but the net effect is that a viewer can still trigger ingestion runs and (via the ungated `GET /integrations/<source>/apps`) read app configs. Worth a `TODO(#359)` comment next to the `canManage` props / run handlers so the remaining gap is discoverable from the code, not just the issue.

### Nits

- `PluginConfigModal.tsx` now calls `clsx(...)` in `ScheduleToggle`; the diff doesn't show a `clsx` import being added. Please double-check it's already imported at the top of the file (typecheck reportedly passes, so probably fine).
- New locale keys are inserted out of alphabetical order in all 8 files (e.g. `adminRequired` before `addApp`; `cancel`, `configuredApps` before `automaticSchedule`, `close`, `configuration`). The surrounding blocks are sorted — worth normalizing to avoid future merge churn / any sorted-keys lint.
- `Scrapers.tsx` calls `useIsAdmin()` again inside `AppConfigList` while the parent `Scrapers` already has `isAdmin`. Passing it down as a prop would match the "explicit `isAdmin` threading" goal of this PR and keep the component testable without store mocks.
- `TestAdminGateOnAppConfigRoutes.test_non_admin_write_rejected` doesn't patch `integrations_handler.secretsmanager`, so unlike the DELETE case it can't assert the secret was untouched. Adding the same `assert_not_called()` pair would make the two negative tests symmetric.
- The new `TestAdminGateOnAppConfigRoutes.test_non_admin_write_rejected` shares its name with the pre-existing credentials test in the same module; a more specific name (e.g. `test_non_admin_save_app_config_rejected`) makes `-k` selection and failure output less ambiguous.

---

perrozzi:

This PR adds `require_admin` gates to the four mutating integrations routes (`POST`/`DELETE /integrations/<source>/apps`, `PUT /sources/<source>/enable|disable`), threads an `isAdmin`/`canManage` flag through `Scrapers` → `PluginConfigModal` → `AppConfigCard` so non-admins get a read-only view, and replaces a batch of hardcoded English strings with `pluginConfig.*` keys in all eight locale files. Backend gates are placed as the first statement in each handler, matching the #307 pattern, and each is pinned by a 403 test plus a positive-path control — that part looks clean. Remaining concerns are frontend i18n/test coupling and the deliberately deferred `POST /sources/<source>/run` gap.

### Should fix

- `voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.tsx` and `PluginConfigModal.tsx` now render `t('card.edit')`, `t('card.delete')`, `t('card.runNow')`, but no `card.*` keys are added in this diff. If any of the eight `scrapers.json` files is missing those keys, users see raw key strings instead of a label. Please confirm all eight locales already define them (the PR's "focused locale assertion" only covered the new `pluginConfig` keys), or add them.
- Mixing namespaces for the same concept is confusing: the app-config controls use `card.edit`/`card.delete`/`card.runNow` while everything else in the same components uses `pluginConfig.*` (`pluginConfig.runNow` already exists). Pick one grouping so future translators/reviewers don't have to guess.
- `AppConfigComponents.test.tsx` asserts the literal English string `'Run now'` on `aria-label`. That couples a permission-gating test to a translation value; a copy tweak in `en/scrapers.json` breaks it. Assert on the testid/presence and leave the string assertion to a dedicated i18n test. The same file also renders `AppConfigCard` bare, without the `createWrapper()` used in `PluginConfigModal.test.tsx` — if it relies on a global i18n init in setup, that's fine, but it's inconsistent with the sibling test.
- `ScrapersContent` still renders `ScraperCard` with unconditional `onEdit`/`onDelete`/`onRun` (`Scrapers.tsx` ~line 265) while the adjacent `AppConfigList` is now gated by `isAdmin`. Either confirm the scraper CRUD routes are already admin-gated server-side and note it, or gate the UI too — the mixed treatment on one page will read as an oversight.
- `POST /sources/<source>/run` remains open and the non-admin Run Now button stays enabled. I accept this as the phase-3 deferral, but as written a viewer can trigger ingestion runs on any source, which is the same class of action as enable/disable. The inline `// Phase 3 (#359)` comments help; please make sure the tracking issue is explicit that this is an open authz gap, not just a UX question.

### Nits

- New locale keys break the existing alphabetical ordering in every file: `adminRequired` before `addApp`, and `configuredApps`/`deleteApp*` inserted before `automaticSchedule`/`close`. Re-sort to keep diffs reviewable.
- `test_non_admin_delete_rejected_and_secret_untouched` docstring claims an unconfigured `get_secret_value` mock "would raise" — a `MagicMock` returns a mock, so the proof is really the `assert_not_called()` lines. Trim the misleading sentence.
- The `app.app_name === '' ? t('pluginConfig.unnamedApp') : app.app_name` fallback is now duplicated in `AppConfigComponents.tsx` and `PluginConfigModal.tsx`; a small helper in `scraper-helpers` would keep it single-sourced (and it silently ignores `undefined`/whitespace names).
- `ScheduleToggle` renders the `adminRequired` hint even while `scheduleLoading` is true, when the `aria-describedby` target checkbox isn't mounted; harmless but the dangling `id` reference is avoidable.
- Consider adding a `data-testid` to the plugin-level Run Now button too, so the non-admin test doesn't depend on the `/run now/i` accessible-name regex matching only one element as more controls appear.

---

perrozzi:

This PR extends the existing `require_admin` pattern from #307 to the four mutating integration routes (`POST`/`DELETE` `/integrations/<source>/apps`, `PUT /sources/<source>/{enable,disable}`), threads an explicit `isAdmin` prop through `Scrapers` → `PluginConfigModal` / `AppConfigList` / `AppConfigCard` to hide non-permitted controls, and i18n-izes the previously hardcoded app-config control strings across all eight locales. The backend gates are correctly placed as the first statement of each handler (before validation, so no info leak on the reject path) and each is pinned by both a negative test and a positive control, plus assertions that Secrets Manager / EventBridge are never touched. Overall this looks clean and matches repo conventions; my comments are mostly about scope-pinning tests and small polish.

### Should fix

- `voc-datalake/lambda/api/integrations_handler.py` / `test_integrations_security.py`: the PR's explicit scope decision is that `GET /integrations/<source>/apps` stays readable for non-admins, but nothing pins that. Add a positive test (non-admin GET returns 200) in `TestAdminGateOnAppConfigRoutes`; otherwise a future "gate everything" change silently regresses the non-admin Scrapers page, which is exactly the risk the phase split was meant to avoid.
- `POST /sources/<source>/run` remains ungated while the UI still shows Run Now to non-admins (`AppConfigCard`, `PluginConfigModal` footer). Deferring the product decision to phase 3 is reasonable, but this is an authenticated non-admin write that spends money and drives ingestion; the phase-3 comments are only in the frontend. Please add a matching `# Phase 3 (#359)` comment (or an issue-linked TODO) next to the `run_source` handler in `integrations_handler.py` so the backend gap is discoverable from the backend.
- `voc-datalake/frontend/src/pages/Scrapers/GeneratorConfigModal.tsx`: confirm the `t` in scope is bound to the `scrapers` namespace (`useTranslation('scrapers')`); the diff only shows the call site. If it is a default-namespace `t`, `t('pluginConfig.adminRequired')` will render the raw key instead of the translation, and there is no test covering that title.
- `voc-datalake/frontend/src/pages/Scrapers/AppConfigComponents.test.tsx` covers `AppConfigCard` directly but nothing asserts that `Scrapers`/`AppConfigList` actually forwards `isAdmin` → `canManage`. A single `AppConfigList`-level (or `Scrapers`-level) test would make the prop-threading regression-proof; today a wiring mistake in `Scrapers.tsx` would keep all tests green.

### Nits

- Locale files: the new keys break the existing alphabetical ordering in every file (`adminRequired` before `addApp`, `delete` after `deleteAppTitle`, `configuredApps` before `automaticSchedule`). Re-sort to keep future diffs minimal.
- `pluginConfig.edit`/`pluginConfig.delete` duplicate the existing `card.edit`/`card.delete` values noted in the description; consider reusing one set rather than maintaining two per locale.
- `test_non_admin_delete_rejected_and_rule_untouched`'s sibling docstring claims an unstubbed `get_secret_value` "would raise" — a `MagicMock` returns a mock, and the failure would actually surface later as a 500 from `json.loads`. The `assert_not_called()` is the real proof; the prose is slightly misleading.
- `ScheduleToggle`: the amber "admin access required" hint renders even while `scheduleLoading` is true (no toggle visible yet); gating it on `!scheduleLoading` would avoid a hint with nothing to describe (and a dangling `aria-describedby` target).
- `AppConfigComponents.test.tsx` asserts the literal `'Run Now'` for `aria-label`; fine given the global i18n test setup, but a locale wording change will break this test rather than the behaviour under test.

---

perrozzi:

This PR adds `require_admin` gates to the four mutating integration routes (`save_app_config`, `delete_app_config`, `enable_source`, `disable_source`) in `voc-datalake/lambda/api/integrations_handler.py`, hides/disables the corresponding admin-only affordances in the Scrapers UI, replaces several hardcoded English strings with `pluginConfig.*` i18n keys across all eight locales, and adds both pytest 403/positive-path tests and vitest coverage. The change is focused, follows the existing phase-1 auth pattern (gate first, before validation or any Secrets Manager access), and includes revert-detecting regression tests. I found no blockers.

### Should fix

- `voc-datalake/frontend/public/locales/*/scrapers.json`: the new keys break the existing alphabetical ordering inside `pluginConfig` (`adminRequired`, `addApp`, `addFirstApp`, `addNewApp`, `cancel`, `configuredApps` are inserted before `automaticSchedule`, and `delete` lands after `deleteAppTitle`). All eight locales are consistently wrong in the same way, so it will not trip a key-parity check, but it makes future diffs noisier and may break any sort-order lint. Re-sort the block.
- `voc-datalake/lambda/api/test/test_integrations_security.py::test_admin_write_succeeds`: the positive control only asserts a 200. Since the sibling negative tests assert `put_secret_value.assert_not_called()`, the admin path should assert `put_secret_value` *was* called (and ideally that the persisted blob contains the new app), otherwise a handler that silently stops persisting would still pass.
- `voc-datalake/frontend/src/pages/Scrapers/ScrapersAppConfigList.test.tsx`: the test primes `qc.setQueryData(['all-app-configs', pluginWithApp.id], [{ pluginId, apps }])` — a per-plugin key holding a multi-plugin array shape. The query key/shape isn't visible in this diff, so if it drifts the test will silently fall through to a network fetch and both assertions could pass/fail for the wrong reason. Consider importing the shared query-key helper (if one exists) instead of hand-writing the tuple.

### Nits

- `AppConfigList` in `voc-datalake/frontend/src/pages/Scrapers/Scrapers.tsx` is now exported purely for testability; extracting it into its own module would be cleaner than widening the page's public surface.
- The phase-3 markers are duplicated in three places (`integrations_handler.run_source`, `PluginConfigModal.tsx` `runMutation`, `Scrapers.tsx` `handleRun`). Two of them add no information beyond the backend one; consider keeping just the backend comment plus the issue link.
- `ScheduleToggle` computes `useId()` unconditionally even when the hint isn't rendered (loading state) — harmless, just noting.
- `test_non_admin_delete_rejected_and_secret_untouched`'s docstring explains that the unset mock return value would raise if reached; that reasoning is subtle and already covered by the explicit `assert_not_called()` — the docstring could be trimmed.
- The PR description refers to `post_app_config`, but the handler is `save_app_config`; worth aligning the description for future archaeology.

---

perrozzi:

This PR adds `require_admin` gates to the four mutating integrations routes (`POST`/`DELETE` app-configs, `PUT` source enable/disable), keeps the `run` route ungated with an explicit phase‑3 marker, and mirrors the gate in the UI (`PluginConfigModal`, `AppConfigCard`, `AppConfigList`) while replacing the remaining hardcoded strings with `pluginConfig.*` keys across all eight locales. The backend change is minimal and correctly placed (gate is the first statement, before any Secrets Manager / EventBridge call), and the negative tests include "no-touch" assertions on the mocked clients so they genuinely fail if the gate is reverted. Overall this looks clean; a couple of points below are worth resolving before merge.

### Should fix

- `voc-datalake/lambda/api/test/test_integrations_security.py` — `test_non_admin_list_app_configs_allowed` deliberately pins non-admin read access to `GET /integrations/<source>/apps`, whose payload is deserialized straight out of the shared Secrets Manager blob (`<source>_configs`). The editor form renders app fields through `PluginField ... showSecrets={false}` (`PluginConfigModal.tsx`), which implies some plugin config fields are secret-typed. Please confirm `list_app_configs` masks/redacts secret-typed fields before returning; otherwise this test locks in a viewer-readable path to credential material. If it already redacts, add an assertion to that test so the redaction can't regress.
- `voc-datalake/frontend/src/pages/Scrapers/ScrapersAppConfigList.test.tsx` — `useConfigStore.setState(...)` mutates the shared Zustand store at render time and is never restored, so this file can leak `apiEndpoint` into unrelated suites depending on run order. Reset it in an `afterEach` (or use a fresh store/wrapper helper as other tests do).
- `POST /sources/<source>/run` remains ungated (`integrations_handler.py`). The deferral to phase 3 is documented, but this route triggers billable ingestion for arbitrary sources; consider at minimum rate/scope limiting or an issue-linked TODO in the CDK-side docs, since the UI now advertises "Run Now" to non-admins (`AppConfigCard`, `PluginConfigModal` tests assert this).

### Nits

- The new keys in all eight `scrapers.json` files are inserted out of alphabetical order (`adminRequired`…`configuredApps` before `automaticSchedule`; `delete` after `deleteAppTitle`). Existing blocks appear sorted - keep them consistent to avoid noisy future diffs.
- `AppConfigList` is now exported from `Scrapers.tsx` solely for the new test. Moving it next to `AppConfigComponents.tsx` would keep the page module's public surface intentional.
- `TestAdminGateOnAppConfigRoutes.test_admin_write_succeeds` shares a name with the pre-existing credentials-route test of the same name; a suffix (`..._app_config`) would make failures easier to attribute.
- UX inconsistency: `GeneratorConfigModal` disables its action with an `adminRequired` tooltip, while app-config add/edit/delete are hidden entirely. Either is defensible, but pick one pattern for the admin gate.
- `appConfigsQueryKey(plugins: Pick<PluginManifest,'id'>[])` could take `readonly Pick<...>[]` so callers can pass readonly arrays without a cast.

---

perrozzi:

This PR adds `require_admin` gates to the four mutating app-config / source-schedule routes in `integrations_handler.py`, keeps read-only Scrapers access for non-admins, hides or disables the corresponding frontend affordances, moves hardcoded control copy to `pluginConfig.*` keys across all eight locales, and adds Python + vitest coverage for both the negative (403) and positive (persisted) paths. Overall this is a clean, well-tested change that follows the repo's existing admin-gate pattern; no blockers found. My comments are about the new secret-redaction helper, the deliberately ungated run route, and a few test/style details.

### Should fix

- `voc-datalake/lambda/api/integrations_handler.py`: `APP_CONFIG_SECRET_FIELDS` ships with empty sets for both plugins, so `_redact_app_config_secrets()` is a no-op in production and the only exercise of the redaction path is a test-injected fake field (`@patch.dict(... {'app_reviews_android': {'api_token'}})` in `test_integrations_security.py`). This is a denylist that must be manually kept in lockstep with each plugin's config schema - exactly the drift class the `base_ingestor._get_known_prefixes()` rule exists to prevent. Prefer deriving the returnable fields from the plugin manifest (allowlist / default-deny), or at minimum add a test that asserts every secret-typed field in the `app_reviews_ios`/`app_reviews_android` config schemas is present in `APP_CONFIG_SECRET_FIELDS`, so adding a credential field to a plugin fails loudly instead of leaking to non-admin `GET /integrations/<source>/apps`.
- `run_source` remains ungated (`integrations_handler.py`, and the mirrored comment in `Scrapers.tsx` / `PluginConfigModal.tsx`). Deferring the product decision is reasonable, but a non-admin can currently trigger unbounded scraper/ingestor invocations (cost and rate-limit exposure against third-party stores). Consider gating it now behind admin and relaxing later if phase 3 decides viewers should have it - loosening is cheaper than a leak. If it stays open, note it in the API docs so operators know viewers can trigger runs.

### Nits

- `voc-datalake/frontend/src/pages/Scrapers/Scrapers.tsx`: `AppConfigList` is now exported solely so `ScrapersAppConfigList.test.tsx` can import it, which pulls the whole page module (and its store/query imports) into the test. Moving it next to `AppConfigCard` in `AppConfigComponents.tsx` would keep the page's public surface unchanged.
- `test_integrations_security.py`, `test_non_admin_delete_rejected_and_secret_untouched` docstring: the claim that an unconfigured `get_secret_value` mock "would raise" is inaccurate - a bare `MagicMock` returns a mock and the failure would surface later as a 500, not a 403. The `assert_not_called()` assertions are what actually prove the gate; consider trimming the docstring.
- Locale files: the new keys break the existing alphabetical ordering inside `pluginConfig` (e.g. `adminRequired`/`addApp`/`configuredApps` inserted before `automaticSchedule`, `delete` after `deleteAppTitle`) in all eight `scrapers.json` files. Cheap to re-sort while the diff is fresh.
- `AppConfigComponents.test.tsx` asserts the literal English string `'Run Now'` via `aria-label`. Asserting the resolved copy couples the test to `en/scrapers.json`; matching on the test id plus a non-empty label would be more durable.
- `ScraperCard` edit/delete in `Scrapers.tsx` are still rendered for non-admins while the new app-config controls are hidden. Worth confirming the webscraper CRUD routes are already admin-gated elsewhere, otherwise the read-only UX is inconsistent within the same page.

---

perrozzi:

This PR adds `require_admin` gates to the mutating app-config (`POST`/`DELETE /integrations/<source>/apps`) and schedule (`PUT /sources/<source>/{enable,disable}`) routes, introduces a redaction hook for secret app-config fields on the read path, and makes the Scrapers UI read-only for non-admins with new `pluginConfig.*` i18n keys in all eight locales. The backend gating is minimal and correctly placed as the first statement in each handler, and the tests include both negative (403 + "no boto3 call") and positive persistence controls, which is exactly the regression shape this repo asks for. No blockers found; a few follow-ups below.

### Should fix

- `voc-datalake/lambda/api/integrations_handler.py`: `APP_CONFIG_SECRET_FIELDS` is empty for both plugins, so `_redact_app_config_secrets` is a no-op today. The latent problem is on the write path: `save_app_config` replaces the whole app entry by id, so as soon as a secret field is added to the denylist, a non-secret edit round-trip (GET redacted → POST) will silently wipe the stored secret. Either merge unspecified secret fields from the existing stored config in `save_app_config`, or add a test that pins the merge behaviour now, before the first secret field lands.
- `POST /sources/<source>/run` remains reachable by non-admin viewers (both backend and `AppConfigList.handleRun`). Deferring the decision to phase 3 is fine as a process choice, but this is an unauthenticated-by-role write that triggers Lambda/Bedrock/scraper cost and can be replayed; consider gating it by default and relaxing later, or at minimum add a test documenting current intended behaviour so the state is asserted rather than implied by a comment.
- `voc-datalake/lambda/api/test/test_integrations_security.py`: `test_app_config_secret_field_denylist_matches_manifest` reaches out of the Lambda test tree into `frontend/src/plugins/manifests.json` via `parents[3]`. That couples the Python suite to the frontend directory layout and will fail confusingly if either path moves. Prefer asserting against a backend-owned source of truth (or, if the manifest is genuinely canonical, add a comment plus a clear failure message when the file is missing).

### Nits

- Locale files: the new keys break the existing alphabetical ordering (`addApp`/`adminRequired` before `automaticSchedule`, `delete` after `deleteAppTitle`). Worth re-sorting so future diffs stay small.
- `AppConfigList` is now exported from `Scrapers.tsx` only so `ScrapersAppConfigList.test.tsx` can import it; moving it into its own module (next to `app-config-query.ts`) would keep the page component's public surface intentional.
- `ScrapersAppConfigList.test.tsx` pre-seeds the query cache with `schedule_enabled: true` (boolean) while `AppConfig` is `Record<string, string>`, and seeding means the query path itself is never exercised. Mocking `api.getAppConfigs` would test closer to reality.
- `AppConfigComponents.test.tsx`: `expect(...getAttribute('aria-label')).toBeTruthy()` is a very weak assertion; assert the resolved label text or use `getByRole('button', { name: ... })`.
- The delete-confirmation strings are now duplicated between `Scrapers.tsx` and `PluginConfigModal.tsx`; a small shared confirm-config helper would avoid the two call sites drifting.

---

PR #372, opened 2026-08-23
Fix BEDROCK_MODEL_ID bypassing model resolution

### LINKED ISSUE #273: [Bug]: BEDROCK_MODEL_ID used directly at two call sites, bypassing model resolution

mayuriphad:
Fixes #273 by replacing direct usages of \BEDROCK_MODEL_ID\ with \get_active_model_id('documents')\. This ensures that model resolution logic and per-surface overrides are correctly applied during persona generation and research steps.

### DIFF
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -12,7 +12,8 @@ from botocore.exceptions import ClientError
 
 # Shared module imports
 from shared.logging import logger, tracer, metrics
-from shared.aws import get_dynamodb_resource, get_bedrock_client, BEDROCK_MODEL_ID
+from shared.aws import get_dynamodb_resource, get_bedrock_client
+from shared.model_config import get_active_model_id
 from shared.api import validate_days, MAX_PERSONAS_PER_GENERATION
 from shared.converse import converse_chain
 from shared.exceptions import (
@@ -789,7 +790,7 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
                 'created_at': now,
                 'updated_at': now,
                 'llm_metadata': {
-                    'model': BEDROCK_MODEL_ID,
+                    'model': get_active_model_id('documents'),
                     'prompt_version': PERSONA_PROMPT_VERSION,
                     'generation_time_ms': llm_time
                 },
--- a/voc-datalake/lambda/api/projects.py
+++ b/voc-datalake/lambda/api/projects.py
@@ -656,8 +656,18 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
         logger.info("[PERSONA] Step 4/6: Executing LLM chain (this may take several minutes)...")
         update_progress(20, 'executing_llm_chain')
         
+        # Resolved once and passed to converse_chain so the recorded
+        # llm_metadata.model below can never drift from what actually ran,
+        # even if the surface's active model changes between this call and
+        # the read at persona-save time.
+        persona_model_id = get_active_model_id('documents')
         try:
-            results = converse_chain(chain_steps, progress_callback=lambda p, s: update_progress(p, s), surface='documents')
+            results = converse_chain(
+                chain_steps,
+                progress_callback=lambda p, s: update_progress(p, s),
+                surface='documents',
+                model_id=persona_model_id,
+            )
             logger.info(f"[PERSONA] LLM chain returned {len(results)} results")
         except Exception as e:
             logger.error(f"[PERSONA] LLM chain execution failed: {e}")
@@ -790,7 +800,7 @@ def generate_personas(project_id: str, filters: dict, progress_callback: callabl
                 'created_at': now,
                 'updated_at': now,
                 'llm_metadata': {
-                    'model': get_active_model_id('documents'),
+                    'model': persona_model_id,
                     'prompt_version': PERSONA_PROMPT_VERSION,
                     'generation_time_ms': llm_time
                 },
--- a/voc-datalake/lambda/api/test/test_persona_model_id_consistency.py
+++ b/voc-datalake/lambda/api/test/test_persona_model_id_consistency.py
@@ -1,0 +1,87 @@
+"""
+Regression test for the persona-generation model-id drift fix (PR #372).
+
+``generate_personas`` used to call ``get_active_model_id('documents')`` twice:
+once implicitly inside ``converse_chain`` (per step) and again, independently,
+when stamping ``llm_metadata.model`` on each saved persona. If the surface's
+active model changed between those two reads, the stored metadata would name
+a model that never actually produced the persona. The fix resolves the model
+once and threads it into both ``converse_chain`` and the saved metadata.
+
+This test forces `get_active_model_id` to return a different value on each
+call, which would make the drift observable if the single-resolution fix were
+reverted: the metadata would then record the second call's value instead of
+the one actually passed to ``converse_chain``.
+"""
+import json
+from unittest.mock import MagicMock, patch
+
+
+MINIMAL_PERSONA_JSON = json.dumps([{
+    "name": "TestUser",
+    "tagline": "a tester",
+    "confidence": "high",
+    "feedback_count": 10,
+    "identity": {},
+    "goals_motivations": {},
+    "pain_points": {},
+    "behaviors": {},
+    "context_environment": {},
+    "quotes": [],
+    "scenario": {},
+    "supporting_evidence": [],
+}])
+
+
+def _make_feedback_item(i):
+    return {
+        "feedback_id": f"fb-{i}",
+        "original_text": f"Feedback item {i}.",
+        "source_platform": "app_store",
+        "sentiment": "neutral",
+        "category": "general",
+    }
+
+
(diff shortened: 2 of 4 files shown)


perrozzi:

This PR swaps two direct `BEDROCK_MODEL_ID` usages for `get_active_model_id('documents')` so persona generation metadata and the research step honor model-resolution/per-surface overrides. The intent is right, but the research-handler change resolves the model at module import time, which defeats the purpose of dynamic resolution, and there are unverified consistency/IAM concerns plus no regression test.

### Blockers

- `voc-datalake/lambda/research/research_step_handler.py`: `MODEL_ID = get_active_model_id('documents')` is evaluated at module scope, i.e. once per cold-start container. Any config-driven override (DynamoDB/SSM/env-refresh) will be frozen for the life of the execution environment, so this only *appears* to fix #273 — behavior depends on which container serves the request. It also moves a potentially remote/IO call into import time: a transient failure or added latency there breaks/slows every invocation, and it bypasses the repo's lazy-init pattern used for `feedback_table`/`projects_table` right below. Resolve it inside the handler (or behind a `_get_model_id()` accessor with per-invocation resolution) and pass it down rather than relying on a module constant.
- Model id ↔ IAM lockstep: `get_active_model_id()` can now return a model other than `BEDROCK_MODEL_ID` for these two Lambdas. The diff contains no CDK change, so please confirm the `bedrock:InvokeModel` resource ARNs granted to the projects and research-step Lambdas cover every model the resolver can return (including inference-profile/cross-region ARNs). If they don't, this converts a wrong-model bug into an `AccessDeniedException` at runtime.
- No test accompanies the fix. Per repo policy a bug fix needs a pytest regression test that fails when reverted — e.g. patch the model config to a non-default value and assert the resolved id is used in `llm_metadata['model']` and in the research step's Converse call.

### Should fix

- `voc-datalake/lambda/api/projects.py:790`: this only changes the *recorded* metadata. If the actual invocation path (`converse_chain`) still resolves the model independently (e.g. from `BEDROCK_MODEL_ID` or its own default), `llm_metadata.model` will now be authoritative-looking but wrong — worse than before, since persona records are persisted. Make the resolved id the single source passed into `converse_chain`, and assert that in the test.
- Surface key `'documents'` is hardcoded in both call sites. Persona generation and research steps are not obviously the "documents" surface; if they have (or should have) their own surface keys, this silently couples them to document-model overrides. Please confirm the key is intentional, and consider a named constant rather than a bare string literal duplicated across modules.

### Nits

- `research_step_handler.py`: missing blank line between `MODEL_ID = ...` and `def _get_feedback_table():` (PEP 8 / consistent with the rest of the file).
- If `MODEL_ID` remains a module symbol for backwards compatibility, add a short comment stating it is only a fallback, to prevent new call sites from reintroducing the stale-value bug.

---

mayuriphad:
Pushed a fix: removed the module-level MODEL_ID (it was resolved once at cold-start and never even read; converse() already resolves the model per-invocation). Also threaded a single resolved model_id through converse_chain into persona generation so llm_metadata.model can't drift from the model that actually ran, with a regression test covering that. Left the IAM/ARN coverage question for a maintainer to confirm since I can't verify the deployed role's Bedrock ARNs from here.

perrozzi:

Since the last review the author dropped the research-handler change entirely (the `MODEL_ID` module constant and the `get_active_model_id` import are now just deleted), threaded a single resolved `persona_model_id` through a new `model_id` parameter on `converse_chain`, and added a regression test. That addresses the import-time-resolution blocker, the metadata-vs-actual-invocation drift, and the missing test; the IAM lockstep concern is now moot for persona generation since the resolved id equals what each step would have resolved from `surface='documents'` anyway.

### Resolved
- Import-time `get_active_model_id()` in `research_step_handler.py` removed — no cold-start-frozen model, no IO at import.
- `projects.py` now resolves once and passes `model_id` into `converse_chain`, so `llm_metadata.model` matches what actually ran.
- Regression test added (`test_persona_model_id_consistency.py`) that fails if the double-resolution is reintroduced.
- Model↔IAM lockstep: no new model can be invoked (same resolver, same surface), so no CDK change needed.
- PEP 8 blank-line nit in `research_step_handler.py` gone with the removed line.

### New findings

**Blocker — verify no dangling `MODEL_ID` reference in `voc-datalake/lambda/research/research_step_handler.py`.** The incremental diff deletes the `MODEL_ID = ...` assignment (and the `BEDROCK_MODEL_ID` import) without touching any other line in the file. If `MODEL_ID` is referenced further down (e.g. passed to `converse` or logged), this is a `NameError` on every invocation; the diff gives no evidence either way. If it was genuinely dead code, please say so in the PR description — because then this PR no longer fixes the research-step half of #273 at all, and the issue should stay open or the title/description should be narrowed to persona generation.

**Should fix — `voc-datalake/lambda/shared/converse.py:624`, precedence between chain-level `model_id` and per-step `surface`.** `model_id=step.get('model_id', model_id)` means an explicit chain-level `model_id` now silently overrides a step's own `surface` override, which the docstring still advertises as a supported per-step knob. Either let a step-level `surface` win over the chain-level `model_id`, or state the precedence order explicitly in the `Args`/step-key docs.

**Nit — `converse.py` step-key docstring** lists `system`, `user`, `max_tokens`, `thinking_budget`, `step_name`, `surface` but not `model_id`, which the code now reads from each step. Add it.

**Nit — `research_step_handler.py` has no accompanying test.** With the module constant gone, a one-line test asserting the handler imports and that `converse` is called with `surface=` (no explicit model id) would lock in the deletion.

---

PR #368, opened 2026-08-21
test(mcp): prove every declared outputSchema against a real payload shape

perrozzi:
## Summary

A parametrised guard that, for **every tool in the published registry**, validates a payload the tool really produces against that tool's own declared `outputSchema`.

Phase 3 (plan §10) adds roughly thirty declarations, each with `additionalProperties: false`, and nothing today compares a declaration against the shape its delegated route actually returns. That gap is already on record as finding **M1** of `analysis/BUG-mcp-tool-truthfulness-GITHUB-ISSUE.md`: `list_personas` declared `pain_points` / `behaviors` as `array` while live rows held nested objects, so a validating client rejected the whole response with `-32602` — the tool was uncallable in production while every test passed. #356 fixed that tool; this closes the class before thirty more chances at it land.

One logical change: a test plus a pinned test-only dependency.

- **`voc-datalake/lambda/api/test/test_mcp_output_schema_conformance.py`** (new, 720 lines)
- **`voc-datalake/requirements-dev.txt`** — `jsonschema>=4.23.0,<5`, test-only, with a comment saying why it is not in any Lambda layer (nothing at runtime validates against these schemas; the server publishes them and the client checks them).

### How it is a guard, not a test of six tools

1. **The tool list is derived, never restated.** Parametrisation walks `mcp_handler.MCP_TOOLS`, so a Phase 3 tool is covered the moment it is published.
2. **A tool with no registered sample FAILS.** Twice, deliberately: once in `test_every_published_tool_has_a_registered_sample_payload` (naming it), and once as a parametrised case with id `<tool>-NO-SAMPLE`, which is what a Phase 3 author reads first. There is no skip path. The completeness check is a *function* (`_tools_missing_samples`) so it has its own positive control — called with the real registry minus one entry, it must report exactly that entry, so "the check found nothing" cannot be because the check finds nothing ever. The reverse direction is checked too: a sample for a tool that is no longer published is as invisible as a missing one.
3. **A real JSON Schema validator**, `Draft202012Validator`. Not hand-rolled: a partial implementation that ignores `additionalProperties` or never descends into `items` passes exactly the declarations that are wrong (M1 lived inside `items`). Each declaration is also run through `check_schema`, so a malformed declaration is not a client's problem to discover.
[...]

PR #375, opened 2026-08-24
fix(core): give the website bucket server access logging (CKV_AWS_18)

perrozzi:
## Summary

`WebsiteBucket` was the only one of the three buckets in `core-stack.ts` without server access logging. It now carries the same pair as its two siblings:

```ts
serverAccessLogsBucket: this.accessLogsBucket,
serverAccessLogsPrefix: 'website-bucket/',
```

`AccessLogsBucket` is a fourth producer on an existing destination — it already receives the raw-data bucket's logs, the S3-import bucket's logs and the CloudFront distribution logs — so no new bucket, no new lifecycle rule, no new IAM principal.

**The `AwsSolutions-S1` suppression is retired in the same commit.** That rule is cdk-nag's check for exactly this missing property, so with the property present the suppression describes a condition that has gone. All three pieces are removed:

- the one-entry `websiteBucketSuppressions` export in `lib/utils/nag-suppressions.ts`
- the `NagSuppressions.addResourceSuppressions(this.websiteBucket, …)` call
- the now-unused name from `core-stack.ts`'s import list

**cdk-nag reports nothing new on that bucket in its place.** I checked directly rather than inferring it from a green suite: a full-app synth reports `0` diagnostics (`aws:cdk:warning` + `aws:cdk:error`) across all five stacks, with `readsRealAnnotations: true` so the empty list means "clean" rather than "the collector parsed nothing". No fresh suppression was needed or added.

### Severity is the scanner's rating, not ours

Worth stating plainly for a reviewer weighing this against the other findings in this file: **this bucket is reachable only through CloudFront with origin access control, and all four public-access blocks are on.** The practical value here is uniformity and a quiet scanner rather than new visibility. It is not a live exposure.

The one substantive thing the new logs add over the existing CloudFront logs is coverage of requests that reach the bucket *without* going through the distribution — CloudFront logs viewer requests, S3 server access logs record what actually arrived at the bucket. That is a real difference, but a small one given the access controls above.

### `AccessLogsBucket` deliberately still logs nowhere

Left exactly as it was, as instructed. The scanner raises the same rule against it and that one is a false positive: S3 does not support a bucket delivering its own access logs to itself. That is now an *assertion* rather than an unstated assumption, so a later well-meaning "fix all the S1 findings" pass fails a test instead of shipping a self-referential destination.

### Test added

Five cases in `lib/stacks/core-stack.test.ts` (not `api-stack.test.ts` — that file is owned by another in-flight change):

[...]

PR #374, opened 2026-08-24
chore(frontend): delete two dead public assets and their lint ignores

perrozzi:
## Summary

Deletes three files that were published to the website bucket — and therefore served from the CloudFront distribution — while being reachable from nothing in the codebase, plus the two ESLint ignores that kept two of them out of every gate. **755 lines are deleted against 235 added, across 9 files.**

The scope grew through review, and the shape is worth stating up front: the deletion is the core, but the docs it touched turned out to name a route that does not exist, and nothing in the repo would have caught either the dead assets or the wrong URL. So the PR now also corrects those docs and adds the two gates that make both classes of defect fail a test rather than survive unnoticed. Review history: [round 1](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/374#issuecomment-5394853058), [round 2](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/374#issuecomment-5395296738).

### `voc-datalake/frontend/public/feedback-widget.js` (373 lines)

A stale second copy of the embeddable feedback widget. The live one is `voc-datalake/lambda/api/static/feedback-widget.js`, read by `feedback_form_handler.py:339` and served from `/feedback-forms/{form_id}/widget.js`. The deleted copy called the singular `/feedback-form/*` routes retired by #277, so anyone following an old integration snippet got a widget that failed against routes that no longer exist. Publishing a broken widget from the CDN is worse than publishing none.

### `voc-datalake/frontend/public/mockServiceWorker.js` (349 lines)

An `msw`-generated service worker. `msw` appears in no `package.json` manifest in the repo (confirmed: `frontend`, `voc-datalake`, `lambda/stream` and root — it survives only as a transitive mention inside `package-lock.json` trees). Nothing imports it and nothing calls `navigator.serviceWorker.register`; the frontend tests mock with Vitest and the hand-written `mock-server.js`. A checked-in worker from an uninstalled version can only ever be registered against a runtime it no longer matches, so if browser-side API mocking is wanted later, `msw` should be installed deliberately and its worker regenerated rather than resurrected from git.

### `voc-datalake/frontend/public/vite.svg` (1 line)

The stock Vite scaffolding logo, added by `create-vite` and unmodified since. Published to the CDN, referenced by nothing: `index.html` names only `/favicon.ico`, and a repo-wide grep for `vite.svg` matched nothing but the inventory guard's own entry — which claimed it *was* referenced by `index.html`.

[...]

PR #373, opened 2026-08-24
fix(stream): widen the chat tool's lookback to 90 days and declare truncation

perrozzi:
## Summary

The streaming chat tool answered from a third of the window the REST route uses, and said nothing about it. Two halves of one defect, fixed together.

**Part 1 — the window.** `fetchCandidatesByDate` clamped its day loop at a bare `30` while the REST feedback routes spend `shared/feedback.py`'s `MAX_LOOKBACK_DAYS = 90` (lines 553, 589). A user asking the chat about a quarter of feedback got a month of it. The clamp is now `Math.min(days, MAX_LOOKBACK_DAYS)` against an exported constant, pinned to Python by the new lockstep test rather than retyped as a loose literal. The constant is declared in `feedback-scan.ts` and spent in exactly one place — `resolveSearchParams` in `search-feedback.ts` — so the scan bound, the cutoff filter and the notice cannot disagree about which window an answer covers. Clamping in two places is the defect `metrics_handler.py` carries a warning about from having had it.

**Part 2 — saying when the answer is capped.** The tool had three silent stopping points (`MAX_CANDIDATES`, the `fetchDayPages` recursion guard, the day-loop `break`) and reported none. Widening the window makes that silence more dangerous, not less. Review found three more, so truncation ended up modelled as a **set of reasons rather than a boolean**: `fetchCandidatesByDate` returns `{ candidates, reasons, unmeasured }`, where `reasons` is drawn from `windowClamped` · `dayPartiallyRead` · `daysUnread` · `dayReadFailed` · `rowsDropped`. Each gets its own clause in the model-facing notice, because one cause named for all of them misattributes the rest — "the candidate cap" is a wrong explanation for a throttled partition. `metrics_handler._scan_recent_items`'s `(items, is_partial)` is still the reference; this runtime needs more because it has failure modes Python's does not (`_query_partition` propagates a failed read, this scan survives one — so a survived failure has to be reported rather than left looking like an empty day).
**Two distinctions inside that set carry the weight.** `scanWasIncomplete()` separates "did the scan read what it scanned" from "does the answer cover less than the question": a clamped-but-fully-read window keeps its completeness claim and gets a paragraph naming the window it did read, instead of being called a truncated scan it never was. And a day reports `reached` explicitly, so a partition whose *later* page fails keeps the rows its first page returned — deriving that from "did this day end with an error" discarded every row of a substantially-read window and told the model the store was unreachable.

[...]

PR #378, opened 2026-08-24
feat(mcp): add five read tools over routes that already exist

perrozzi:
## Summary

Five new READ tools on the MCP server, each adapting a route that already exists — declaration and projection work, no route handler, aggregator or frontend touched. `MCP_SERVER_VERSION` moves `3.7.0` → `3.8.0` with the published-shape fingerprint in the same commit.

| Tool | Route | Scope / reach | Cost class |
|---|---|---|---|
| `list_feedback` | `GET /feedback` | `feedback:read` / workspace | expensive |
| `get_similar_feedback` | `GET /feedback/{feedback_id}/similar` | `feedback:read` / workspace | moderate |
| `list_urgent_feedback` | `GET /feedback/urgent` | `feedback:read` / workspace | expensive |
| `list_feedback_facets` | `GET /feedback/entities` | `feedback:read` / workspace | expensive |
| `list_jobs` | `GET /projects/{project_id}/jobs` | `projects:read` / project | moderate |

All eight registration sites are updated (`_TOOL_DECLARATIONS`, `TOOL_COST_CLASSES`, `TOOL_TITLES`, `TOOL_HANDLERS`, `TOOL_SCOPE_REQUIREMENTS`, `TOOL_REACH_KINDS`, `DOMAIN_ROUTES` × 4 new routes, `_TOOL_SAMPLES`), and the tables' structure is left as it is — collapsing the eight into one record is a separate change, as asked.

### What each tool reports that nothing did before

- **`list_feedback`** carries `total`, `offset`, `limit` and `is_partial_window` through beside the items. `search_feedback` delegates to the same route for its filter-only branch and drops all four, so a model saw a page and could not tell a last page from a full one, nor ask for the next. `total` is described as the size of the **filtered candidate window** and as a **lower bound** when `is_partial_window` is true — both of which are what the route means by it. Items reuse `_project_feedback(item, summary=True)` unchanged.
- **`get_similar_feedback`** declares `maximum: 50` on `limit`, matching the route's own `validate_limit(..., max_val=50)` exactly rather than inventing a tighter bound. An unknown id is the route's 404 arriving as a tool error, which `_delegate` already does; a successful empty answer would tell a model a complaint is unique when the id was simply wrong.
- **`list_urgent_feedback`**'s `count` description states that the number is **this page's length**, because the scan stops at `limit` and the route carries no truncation flag, and points a reader at `get_metrics_summary`'s `urgent_count` for a window total. That misreading is a defect on record in the route's own comment: the sidebar urgent badge read `count` as a total with `limit=10` and could never display more than 10. The route's real defaults are declared — `days: 30`, `limit: 50` (cap 100).
- **`list_feedback_facets`** reports `categories` and `issues`. `keywords` is **left out of both the projection and the declaration**, with a comment saying the route has no keyword data to report yet: it is the hardcoded literal `{}` in both branches. `is_partial` is coerced with `bool(body.get('is_partial'))` the way `_tool_search_feedback` does, because only one of the two branches sends it. `personas`/`sources` are dropped for a weaker reason, also commented: `get_metrics_breakdown` already reports each per axis with the legacy-bucket caveat attached.
- **`list_jobs`** leaves `result` out of the projection — arbitrary payload, and the same argument `_tool_get_project` makes for listing documents by title (the 6 MB synchronous-invoke ceiling and the model's context budget, both reachable by one completed prototype). The description states the route's hardcoded `Limit=50` and the absent continuation token, so a 51st job is silently unreachable. `inputSchema` is `project_id` alone, because the route reads no query parameters; the id comes from the existing resolution via `_PROJECT_ID_ARG` and `token_info['project_id']`, as `get_project` does.

`additionalProperties: false` on all five, since all five project their answers. Existing property blocks are reused (`_FEEDBACK_SUMMARY_PROPERTIES`, `_DAYS_ARG`, `_PROJECT_ID_ARG`); the four `/feedback` filter arguments were hoisted out of `search_feedback` into `_CATEGORY_ARG`/`_SENTIMENT_ARG`/`_SOURCE_ARG`/`_DATE_BASIS_ARG` because four tools now forward them — the published values are byte-identical, so the fingerprint moves for the five additions alone.

## Gates

| Gate | Baseline on `development` | This branch |
|---|---|---|
| `pytest lambda -q --no-header --no-cov` | **3406 passed** | **3479 passed, 0 failed** |
| `ruff check lambda plugins` | **509 findings** (all pre-existing) | **509** — unchanged, new code clean |
| `npm run test:cdk` | 279 passed | **279 passed** (17 files) |

Commands run exactly as specified, from `voc-datalake/`:

```
env -u AWS_PROFILE AWS_ACCESS_KEY_ID=testing AWS_SECRET_ACCESS_KEY=testing \
    AWS_SESSION_TOKEN=testing AWS_DEFAULT_REGION=us-east-1 \
    .venv/bin/python -m pytest lambda -q --no-header --no-cov
ruff check lambda plugins
npm run test:cdk        # after `cd frontend && npm run build`, which a fresh
                        # checkout needs for BucketDeployment
```

The +73 tests are 37 new parametrised conformance cases (three or more sample cases per new tool, times the file's per-tool checks) and 36 new delegation tests. `api-stack.test.ts` passes the four new `DOMAIN_ROUTES` entries unchanged: each is registered by its handler's decorators, wired in API Gateway (`/feedback/{id}/similar`, `/feedback/urgent`, `/feedback/entities` explicitly; `/projects/{proxy+}` covers the jobs route), and owned by a domain whose function env var the stack sets. No IAM change — all five stay inside the metrics and projects Lambdas `mcpRole` already invokes.

### Lockstep tables this touched, and why each had to move

Four fail-closed tables reported the change rather than silently absorbing it, which is the pattern working:

1. `test_a_changed_published_tool_shape_moves_the_server_version` — fingerprint updated to `("3.8.0", "8d4a76e902615ab7")` in the same commit as the version.
2. `test_every_mcp_reachable_route_that_publishes_the_flag_has_a_tool_declaring_it` — `/feedback/entities` publishes `is_partial` and is now reachable, so it needed a serving tool that declares the flag, which `list_feedback_facets` does. **The docstring's claim that the route "cannot be reached" because it is in `_RESERVED_PATH_SEGMENTS` was conflating two things** and is corrected: that entry stops a *feedback id* from impersonating the route (it is a static sibling of `/feedback/{feedback_id}`) and never bore on whether a tool may address the route by its own key. The accounting is now a route → serving-tool table and each named tool's declaration is checked, since "some tool serves it" says nothing if that tool does not declare the flag.
3. Two loops over every registered tool called each with `{}` and carried an inline `if name == ...` chain for the two tools with a required argument. `get_similar_feedback` is the third, so it was refused `-32602` and the loop failed on a missing `result` key. Replaced with one `_MINIMAL_ARGUMENTS` table, so the next required argument is added in one place.

### Test-integrity evidence (mutation)

The new delegation tests were checked against mutations of the production code rather than by inspection:

| Mutation | Failures |
|---|---|
| drop `total` from `list_feedback`'s projection and `required` | **3**, across 3 files — the pagination test, PR #368's substance check, and the fingerprint |
| declare `keywords` on `list_feedback_facets` | **3** — including `test_every_declared_property_is_demonstrated_by_some_sample[list_feedback_facets]`, exactly the guard the task predicted |
| inline `result` in `_project_job` | **8** — 4 delegation, 4 conformance (including the closed-subschema control at `jobs/[]`) |

Two of the new assertions read `metrics_handler.py`'s **source** rather than a fixture: the hardcoded `'keywords': {}` in both branches, and the urgent route's flagless `return`. So the day either route changes, the tool's caveat is reported as stale instead of quietly outliving its reason — and the fix in each case is to declare the field, which is the outcome to want.

## Conformance samples

Each new tool registers three cases (two for `get_similar_feedback`), and every route body is **provenance level 2** of `_TOOL_SAMPLES`' own ranked convention: the route handler's `return` transcribed field for field, with the function named in a comment. Level 1 was unavailable — no fixture exists for these four bodies — but the **rows inside them are level 1**, reused from `_feedback_row()`, `_HOSTILE_FEEDBACK_ROW` and `_LEGACY_FEEDBACK_ROW`, which is where the shape that broke M1 actually lives.

The samples cover the branches a single schema has to describe: both branches of `get_entities` (aggregates and scan), a truncated candidate window, an isolated item with no neighbours, a running job whose `completed_at`/`error` the route sends as `None` (M1's exact shape, and the ordinary case for any unfinished job), and a failed job — the only shape that fills `error`, without which that field would be declared and demonstrated by nothing. `_ENTITIES_BODY` keeps `keywords: {}` deliberately: it is the evidence for the omission, so a future reader can see the route sending nothing rather than wondering whether the field was forgotten.

## Decisions made

- **Cost classes.** `list_feedback`, `list_urgent_feedback` and `list_feedback_facets` are `expensive`; `get_similar_feedback` and `list_jobs` are `moderate`. The axis is the shape of the read, per the vocabulary's own comment: the first three are window walks or per-item `get_item` fan-outs bounded by a soft cap, the last two are one keyed read plus one partition Query. Note `test_the_expensive_class_is_the_tool_that_can_truncate` only constrains tools requiring `is_partial`, and `list_feedback` requires `is_partial_window` — a different key, so it does not trip that test either way.
- **`list_feedback_facets` requires `feedback:read`, not `metrics:read`.** Its answer is counted rather than verbatim, but it aggregates the feedback corpus and its `issues` map holds real problem summaries, so a credential that may not read feedback should not read counts of it. `metrics:read` covers the pre-aggregated dashboards, which are a different set of rows.
- **`offset` declares no `maximum`.** The route's ceiling is `metrics_handler`'s own `MAX_FEEDBACK_OFFSET`, which lives in a different Lambda bundle this one does not import; writing `5000` here would be exactly the un-CI-able copy `_DAYS_ARG`'s comment argues against. The route clamps rather than refusing, so the bound is named in prose and not declared.
- **`get_similar_feedback` reports the caller's own `feedback_id`** as `source_feedback_id` rather than the route's echo of it. Same value; this keeps the required field present even if the route ever stopped echoing.

## No route change was needed

Every one of the five is truthful against its route as it stands. Two facts are un-improvable from the adapter and are stated in the declarations instead, both flagged here as the task asked rather than fixed:

- `list_urgent_feedback` cannot detect truncation — the route publishes no flag, and `count == limit` is indistinguishable from a page that happened to be exactly `limit` long. A route-side `has_more` or a companion `total` would close it.
- `list_jobs` cannot detect its own truncation either, for the same reason: `Limit=50` with no `LastEvaluatedKey` makes a full page and a truncated one the same answer. Adding a continuation token (or `limit`/`status`/`job_type` parameters) is left for a later slice, as instructed.

## Out of scope, as asked

The registration tables' structure; `list_projects` and `list_documents`; the route handlers, aggregator and frontend; and any `limit`/`status`/`job_type` parameter on the jobs route.

## Agent notes

**What went well.** The repo's own machinery did most of the reviewing. Every table this change had to touch is fail-closed or fingerprinted, so the first full-suite run named all four things I had missed — the fingerprint, the `is_partial` route accounting, and two loops whose inline argument chains predated a third tool with a required argument. That is a much better experience than a suite that passes and leaves the gaps for production. PR #368's substance check earned its keep immediately: the `keywords` mutation is reported by name, so the task's claim about it is verifiable rather than a matter of trust.

**What was difficult.** Three things.

*Which `is_partial` is which.* `/feedback` publishes `is_partial_window`, the metrics and entities routes publish `is_partial`, `search_feedback` renames the former to the latter, and the two flags mean genuinely different things (a soft-capped text scan versus a candidate window that also bounds `total`). I kept `list_feedback` on the route's own name and asserted that `is_partial` is *absent* from its payload, because one name for two facts would be worse than two names — but this is the kind of distinction that will be re-litigated by whoever adds the next feedback tool. It is worth a paragraph somewhere central.

*The reserved-segment/reachability conflation.* `test_every_mcp_reachable_route...`'s docstring asserted that `/feedback/entities` could not be reached *because* it is in `_RESERVED_PATH_SEGMENTS`. That was wrong on its own terms and it is the sort of wrong that only surfaces when someone acts on it — I had to be confident the reserved entry was about path-parameter impersonation, not about tool addressing, before adding the route. The correction is in this PR, and the same shape of claim may exist elsewhere.

*Where a bound may be written down.* `_DAYS_ARG` imports `MAX_FEEDBACK_WINDOW_DAYS` from `shared.api`, which the MCP bundle does include; `MAX_FEEDBACK_OFFSET` lives in `metrics_handler`, which it does not. So two bounds a page apart in the same declaration get opposite treatments for a reason that is invisible from the declaration. Promoting the offset cap to `shared/api.py` beside the window ceiling would let it be declared rather than narrated, and would be a small, self-contained follow-up.

**Conventions discovered.**
- The comment-to-code ratio in `mcp_handler.py` is very high and it is load-bearing, not decoration: comments record *which defect a line prevents* and are cited by test docstrings. A change here that adds code without adding that reasoning reads as unfinished.
- The recurring rule is "a declaration looser than reality is a wrong answer; stricter is merely another spelling". Every schema decision in this PR was made against it.
- Test docstrings carry a REVERT STORY — which mutation makes which assertion fail — and the mutation is expected to have been *run*, with the resulting count recorded. I followed that and put the numbers in this PR body.
- `python -m pytest lambda` from `voc-datalake/` needs the layer requirements too (`lambda/layers/ingestion-deps`, `lambda/layers/processing-deps`), not just `requirements-dev.txt`; without them 13 modules fail to collect on `aws_xray_sdk` and the run is *interrupted*, which reads as a broken checkout rather than a missing dependency. `docs/deployment.md`'s venv recipe covers it, but the failure mode is unhelpful enough to be worth a line in the test docs.
- `npm run test:cdk` needs `frontend/dist` to exist. Confirmed: `cd frontend && npm run build` first, as the task said.

1. **Collapse the eight registration tables into one record per tool.** Five tools of bookkeeping was acceptable, as the task says; the plan's remaining ~25 will not be. A single `@dataclass` per tool (name, declaration, handler, scope, reach, cost, title) with the six tables *derived* from it turns six chances to forget into zero, and the existing "the tables agree" tests become the derivation's own trivial identities. This is the highest-value refactor available here.
2. **Replace `test_mcp_delegation.py`'s `_schema_errors` with `jsonschema`.** It is a deliberate hand-rolled subset and its docstring says `jsonschema` "is not installed in this suite" — which stopped being true when PR #368 pinned it in `requirements-dev.txt`. Two validators of different strengths over the same declarations is exactly the drift risk the rest of the file argues against, and the weaker one is the one most tests call.
3. **Give the urgent and jobs routes honest truncation signals**, in their own PRs. Both tools currently carry a prose caveat where a field belongs, and my tests pin the route shapes so those PRs will be told to update the tools.
4. **A `list_projects` tool needs a new reach kind first** — the task set it aside for that reason, and that reach work is a prerequisite worth doing on its own rather than inside a tool PR.

```abca-verify
no-ui
```


perrozzi:

This PR adds five read-only MCP tools (`list_feedback`, `get_similar_feedback`, `list_urgent_feedback`, `list_feedback_facets`, `list_jobs`) that adapt existing routes, bumps `MCP_SERVER_VERSION` to `3.8.0` with the published-shape fingerprint in the same commit, hoists the four shared `/feedback` filter blocks into module constants, and replaces two inline `if name == ...` chains with a `_MINIMAL_ARGUMENTS` table. No route handlers, IAM, aggregator or frontend are touched, and all four `DOMAIN_ROUTES` additions stay inside the two functions `mcpRole` already invokes, so the 20KB policy and least-privilege rules are unaffected. Overall this is clean and unusually well guarded by the repo's own fail-closed tables; my remaining concerns are one unverified truthfulness claim in the declarations and two brittle test assertions.

### Should fix

- **`list_urgent_feedback` declares four filters the diff never proves the route honours.** `voc-datalake/lambda/api/mcp_handler.py` declares `category`/`sentiment`/`source`/`date_basis` and `_tool_list_urgent_feedback` forwards them via `_feedback_filters`, but `test_the_filters_reach_the_route_under_its_own_parameter_names` in `test/test_mcp_delegation.py` only asserts they arrive as query params against a stub — it cannot show `get_urgent_feedback` reads them. This is exactly the "declaration looser than reality" shape the file exists to prevent (and the reason `list_feedback_facets` correctly omits `category`). Please confirm against `metrics_handler.get_urgent_feedback` and drop any parameter it ignores; same check for `list_feedback_facets`'s `limit` ceiling of 200 and `get_similar_feedback`'s `default: 8`.
- **The two source-text assertions are coupled to formatting, not behaviour.** `test_the_route_still_has_no_truncation_flag_to_report` requires the literal `"return {'count': len(items), 'items': items[:limit]}"` and `test_the_route_really_does_hardcode_an_empty_keyword_map` requires `text.count("'keywords': {}") == 2` (`test/test_mcp_delegation.py`). A line wrap, a quote-style change or a third branch that also returns `{}` flips these with a message that misdescribes what happened. Prefer invoking `metrics_handler.get_entities` / `get_urgent_feedback` with a stubbed table and asserting on the returned body, or at least a whitespace/quote-tolerant regex.
- **Zero-filled pagination when the route omits a field.** `_tool_list_feedback` reports `_as_int(pagination.get('limit'))`, so a body lacking `limit` yields `limit: 0` (asserted in `test_a_route_that_omits_the_flag_reports_false_rather_than_null`, where all three come back 0). A page size of 0 is a value the route can never mean, and a client walking pages by `offset + limit` will loop forever. Falling back to the effective request value (`args.get('limit', 50)` / `args.get('offset', 0)`) is truer than 0.
- **Inconsistent degradation across the five handlers.** `_tool_list_feedback_facets` raises `DelegationUnavailable` on a non-dict body while `_tool_list_feedback`, `_tool_list_urgent_feedback` and `_tool_list_jobs` coerce the same condition to an empty answer. Either treatment is defensible; five tools over the same route family should pick one and say why in the usual comment.

### Nits

- `test/test_mcp_delegation.py` now uses `Path(__file__)` but the diff adds no `from pathlib import Path`; confirm it is already imported at module top.
- `get_similar_feedback`'s `feedback_id` could declare `"minLength": 1` so the schema matches the `InvalidToolArgument` guard in `_tool_get_similar_feedback`.
- `list_jobs`' `jobs.items` schema has no `required` list even though `_project_job` writes all nine keys unconditionally — the same argument the top-level `required` makes applies one level down.
- The `is_partial` vs `is_partial_window` distinction is now spread across three tools; the PR body's own suggestion to write it down once (near `_IS_PARTIAL_DESCRIPTION`) would be cheap here rather than in a follow-up.
- Collapsing the eight registration tables into one per-tool record is acknowledged as out of scope, but five new tools × six tables is already the largest remaining foot-gun in this file.

---

perrozzi:
## Round 1 disposition — three taken, one refuted with evidence, all five nits done

One commit, `8b683b5`. Every guard added or changed below was verified by **breaking the thing it
guards**, confirming the failure names the real cause, then restoring — 19 mutations in all, and I
re-ran the three that would change the merge decision through an independent driver.

### 1. The declared filters — **refuted, and the declarations are accurate**

You asked me to confirm against `metrics_handler.get_urgent_feedback` and drop any parameter it
ignores. It ignores **none**. All four are read and applied:

| Declared filter | Read at | Applied at |
|---|---|---|
| `date_basis` | `metrics_handler.py:460` | `:471` (widens the fetch) and `:493` (the window skip) |
| `source` | `:461` | `:495-496` |
| `sentiment` | `:462` | `:497-498` |
| `category` | `:463` | `:499-500` |

The other two claims in that finding are also accurate: `get_entities` really does cap the sample at
200 (`validate_limit(..., default=100, max_val=200)`, `:524`) and `get_similar_feedback` really does
default to 8 (`validate_limit(..., default=8, max_val=50)`, `:795`). So nothing was dropped.

**But your underlying point was right and is fixed.** The old test proved the parameters reach a
*stub*, which could never show the route honouring them — so the declaration was true and
*unguarded*. It is now proven behaviourally, invoking the real route against a patched table the way
`test_metrics_handler.py` already does:

| Mutation to the route | Result |
|---|---|
| delete the `category` branch | fails: *"GET /feedback/urgent no longer filters on `category`; list_urgent_feedback declares it, so the declaration now promises a filter the route ignores"* |
| delete the `sentiment` branch | fails, same shape |
| delete the `source_platform` branch | fails, same shape |
| delete the `basis_date(...) < cutoff_date` skip | fails: *"…a caller asking for what customers WROTE this month gets what was IMPORTED"* |
| urgent `max_val` 100 → 200 | fails: *"clamps a page to something other than 100"* |
| entities `max_val` 200 → 400 | fails, naming the sample size |
| similar `default` 8 → 20 | fails: *"no longer defaults to 8 neighbours"* |

`test_the_filters_reach_the_route_under_its_own_parameter_names` is **kept**: it proves the *adapter*
sends the route's own parameter names, which the route-driven tests bypass entirely. The two are
halves of one chain and both docstrings now say so.

### 2. The two source-text assertions — taken

Both replaced with behavioural assertions against the real routes, keeping each one's *intent* (they
exist so that the day the route grows the capability, the test fails and the tool can declare the
field — that is still the message they print).

- urgent truncation flag: now asserts the returned body grows no such field. Adding `'is_partial': False`
  to the route fails with *"the urgent route's answer grew a field (['is_partial']) — if it now reports
  truncation or a window total, replace the caveat"*.
- entities keywords: now asserts `entities['keywords'] == {}` in the returned body, **parametrised over
  both branches** (`aggregates` and `scan`) rather than counting a literal twice. Filling either fails
  naming that branch.

### 3. Zero-filled pagination — taken

`offset` and `limit` now fall back to the **effective request value**, not 0. A page size of zero is a
value the route can never mean and it is not inert: a client walking `offset + limit` never advances.

`total` deliberately keeps its 0, and that is argued in the code rather than left implicit: every
candidate fallback (`len(items)`, `offset + count`) would assert that this page is the whole filtered
window — the last-page claim the route did not make, and the exact invention `total` exists to prevent.
Nothing in the MCP process knows the window's size. A 0 beside a non-empty `items` is visibly
self-contradictory, which is the honest signal available.

The M1 property is preserved rather than traded away — `_as_int`'s `default` is **itself coerced** and
fires on absent *or uncoercible*, so a caller's own `"20"` cannot travel under an `integer`
declaration. Three mutations pin that: reverting to `_as_int(...)` alone, reading a legitimate
`offset: 0` as absence, and returning the fallback raw all fail with distinct messages.
`test_a_route_that_omits_the_flag_reports_false_rather_than_null` encoded the old behaviour and was
updated with the fix.

### 4. Inconsistent degradation — taken; **raising**, for all five

A non-object body from a route that returns objects is a malfunction, not a thin answer. Coercing it to
`{}` makes the tool say *the window is empty* or *this project has run nothing* — a claim about the
corpus the call never established, and one a model cannot retry because it was told the call succeeded.
"The call failed" is recoverable; "nobody said that" is not. That is the class the truthfulness register
is about.

Implemented as one helper, `_delegated_object`, which also absorbs the two inline copies `get_project`
and `get_feedback_detail` already carried, so the rule has a single implementation. `_projected_rows`
lost its own `isinstance(body, dict)` branch — it was the coercing half of the same inconsistency.

Two things pinned so this cannot drift back: a non-object body is a server fault **for all five tools**,
and an **empty object is still a real answer** for all five (that second one matters — making
`_delegated_object` also raise on `{}` turns a working call into a broken one, which a model retries
forever).

⚠️ **Three shipped tools deliberately still coerce**, documented in the helper: `get_metrics_summary`
and `get_metrics_breakdown` forward the body unprojected and declare nothing `required`, so `{}` is a
shape their own declaration permits — the same argument that makes their `is_partial` optional.
`search_feedback` predates the rule. Changing shipped tools' failure mode is a behaviour change with
its own version implication, not something to add beside five new tools.

### Nits — all five

- `from pathlib import Path` was already imported; the rewrite in item 2 removed the last use, so it is gone and ruff is unchanged.
- `get_similar_feedback`'s `feedback_id` now declares `minLength: 1`, matching its handler's guard. Removing it fails: *"declares None as its minimum while the handler refuses a blank id, so the schema is looser than the server"*.
- `list_jobs`' `jobs.items` now carries `required` for all nine keys. Removing it fails naming the fields *"declared without being promised, though `_project_job` writes every one of them unconditionally"*.
- The `is_partial` vs `is_partial_window` distinction is written down **once** near `_IS_PARTIAL_DESCRIPTION`; the three tools reference it.
- Collapsing the six registration tables is left alone as agreed out of scope.

### Version and fingerprint

`MCP_SERVER_VERSION` stays **3.8.0**. Every declaration touched belongs to a tool this same unreleased
version introduces, so no client has cached any of them. The fingerprint was **recomputed with the
pin's own algorithm** rather than copied out of the failure message — 11 tools, 24,522 bytes,
`1694a93761ac16a8`. A one-character description change still fires the lockstep, which is verified.

### Gates

| Gate | Result |
|---|---|
| `pytest lambda -q --no-cov` | **3511 passed** (was 3479; +32) |
| `ruff check lambda` | 383 — rule histogram **byte-identical** to `origin/development`, zero new |
| Files touched | `mcp_handler.py` + the four MCP test files. No routes, no IAM, no frontend |

### One follow-up, named rather than smuggled in

`get_feedback_detail`'s `feedback_id` carries the identical non-empty handler guard without
`minLength: 1`, so the argument in the nit above applies to it too. It is a declaration shipped in an
earlier version, so it is a one-line change plus a fingerprint recompute rather than part of this
slice. Say the word and it goes in.

perrozzi:

Since my last pass the author reworked all four should-fix items: the four `/feedback/urgent` filters, the facets `limit` ceiling and the similar-route `default` are now proven by driving the real `metrics_handler` routes over stubbed tables; both source-text assertions were replaced with route-driven equivalents; `_as_int` grew a coerced `default` so `limit`/`offset` fall back to the effective request rather than 0; and the four projecting handlers were unified behind a new `_delegated_object` with the raising treatment (plus `minLength`, the job-item `required`, and the `is_partial` vs `is_partial_window` note the nits asked for). The fingerprint moved with it and `MCP_SERVER_VERSION` correctly stays at the still-unreleased `3.8.0`. This is a clean delta.

### Resolved
- Filter/ceiling/default truthfulness for `list_urgent_feedback`, `list_feedback_facets` and `get_similar_feedback` is now asserted against the real routes (`TestTheDeclaredFiltersAreHonouredByTheRoute`).
- Both formatting-coupled source-text assertions replaced by route-driven ones (`test_the_route_really_does_answer_an_empty_keyword_map`, `test_the_route_still_has_no_truncation_flag_to_report`).
- `limit`/`offset` no longer report an impossible 0; `total` keeps its 0 with the reasoning recorded, and `_as_int` preserves a legitimate `0`.
- Degradation unified via `_delegated_object`; existing error strings are byte-identical, so `_get_project_payload` / `_tool_get_feedback_detail` behaviour is unchanged.
- Nits addressed: `minLength: 1`, job-item `required`, the single `is_partial` vs `is_partial_window` note.

### New findings
- **`_entities_body(aggregates, feedback, ...)` never uses `feedback`** (`test/test_mcp_delegation.py`). Harmless but misleading at the call sites, which pass the mock that the caller itself wired — drop the parameter or use it.
- **`limit`'s fallback is the *requested* value, not a clamped one.** In `_tool_list_feedback`, `_as_int(body.get('limit'), requested_limit)` would report `limit: 500` for a caller who sent 500 (the input schema is declared, not enforced) under a field described as "the page size the route applied". The real route always echoes, so this is hypothetical; clamping the fallback to the declared `maximum` would close it in one expression.
- **`list_feedback_facets`' `date_basis` is still stub-only.** `source` at least selects the scan branch in the new tests, but nothing shows `get_entities` applies `date_basis` to its window — the one remaining declaration in this PR with no route-side proof. `list_feedback`'s four filters are fine to leave, since `search_feedback` already declared them against the same route.

### Nits
- `assert set(body) == {"count", "items"}` is deliberately fail-closed, which is right, but it will also fire on a benign additive field; the message covers that, so no change needed.
- Confirm `api_gateway_event` / `lambda_context` come from the shared `test/conftest.py` — they are used but not defined in the diff.

---

perrozzi:
## Round 2 disposition — all three taken, in `c3afb29`

Both nits needed no change and I confirmed why rather than assuming: the fail-closed
`set(body)` assertion is right as it stands and its message already covers the additive case,
and `api_gateway_event` / `lambda_context` are provided by the shared `test/conftest.py`
(`:122` and `:170`).

### 1. `_entities_body`'s unused `feedback` parameter — dropped

Gone, with its three call sites. Misleading exactly where you said: the callers were passing the
mock they had themselves wired.

### 2. The fallback is now clamped to what the tool declares

You were right that this is hypothetical against the real route and right that it is one
expression. `inputSchema` is advertised and enforced on nothing — `tools/call` reaches a handler
with whatever `arguments` arrived — so a caller's `limit: 500` travelled unchallenged and would
have been reported under a field described as *the page the route applied*, for a page the route
would have capped.

New `_within_declared_bounds` **reads the bounds off the declaration** rather than restating
them, which is the rule `_DAYS_ARG` already follows by importing `MAX_FEEDBACK_WINDOW_DAYS`: a
literal 100 here would be a second source of truth that cannot fail CI when the schema moves, it
just quietly disagrees. It is placed beside `_as_int` because it is the same kind of thing — a
coercion that refuses to let an unchecked value travel under a declaration — and because it needs
both `_as_int` and `_TOOL_DECLARATIONS`, so it belongs after each. `_as_int`'s contract is
preserved, so the M1 property (a coerced default firing on absent **or** uncoercible) survives.

Both bounds are pinned, since a ceiling alone would not catch the floor:

| Caller sends | Reported | Guard |
|---|---|---|
| `limit: 500` | the declared `maximum` (100) | reverting to `_as_int(args.get('limit'), 50)` fails: *"reported back as 500; list_feedback declares 100 as its `maximum`"* |
| `limit: 0` | the declared `minimum` (1) | *"a reported page size of 0 makes `offset + limit` loop on one page"* |

Nothing published moved, so `MCP_SERVER_VERSION` and the fingerprint are unchanged —
`_within_declared_bounds` reads the declaration, it does not alter it.

### 3. `list_feedback_facets`' `date_basis` — now proven against the route

It is honoured: `get_entities` reads it at `metrics_handler.py:526`, uses it at `:530` to select
the scan branch, and passes it into `_scan_window_items` at `:531`. Previously only the first of
those was observable. Now a row is seeded inside the window by its import date and 400 days
outside by its review date, and only the basis changes between the two calls.

🔑 **`source` is passed both times deliberately.** `date_basis=review` is itself one of the two
conditions that selects the scan branch, so varying the basis alone would also change the branch
and compare two different code paths — the row could vanish because the basis worked *or* because
aggregates answered. Pinning the branch with `source` leaves the basis as the only variable.

Passing a literal `'imported'` to `_scan_window_items` fails it: *"a row written 400 days ago
still counts under the 'review' basis, so list_feedback_facets declares a basis the route ignores
and a caller asking what customers WROTE this week gets what was IMPORTED this week."*

Asserted as in-window versus not, rather than as an exact count: the scan queries one partition
per day and the stub answers every one with the same row, so the absolute figure is an artefact of
the mock (it is `days`, not 1). My first version asserted `== 1` and its own fixture-drift guard
caught it, which is the reason that guard was written that way.

### One methodological note, because it nearly cost a false pass

My first attempt at the `date_basis` mutation substituted a constant `metrics_handler` does not
import. The suite failed with a `NameError` and my driver reported it as **killed** — a guard that
checks nothing would read identically. A non-zero exit is not evidence the guard fired. The driver
now requires an `AssertionError` in the output before crediting a kill, and both mutations above
were re-run under that rule.

### Gates

| Gate | Result |
|---|---|
| `pytest lambda -q --no-cov` | **3514 passed** (was 3511; +3) |
| `ruff check lambda` | **383**, rule histogram byte-identical to `origin/development` |
| Files touched | `mcp_handler.py` and `test/test_mcp_delegation.py` only |

perrozzi:

Since my last pass the author closed all three of my findings: `_entities_body` lost its unused `feedback` parameter, the `list_feedback` page fallbacks are now clamped through a new `_within_declared_bounds` helper that reads `minimum`/`maximum` off the live declaration (with both a ceiling and a floor test), and `list_feedback_facets`' `date_basis` finally has route-side proof that pins the scan branch with `source` so the basis is the only variable. Published shapes are unchanged, so leaving the fingerprint and `MCP_SERVER_VERSION` alone is correct. Clean delta.

### Resolved
- `_entities_body(aggregates, ...)` no longer takes the unused `feedback` mock; all three call sites updated.
- `limit`'s fallback (and the value forwarded to the route) is clamped to the declared `maximum`/`minimum`, with bounds read from `_TOOL_DECLARATIONS` rather than restated, plus two new tests.
- `date_basis` on `list_feedback_facets` is now proven against the real `get_entities` window (`test_the_facets_date_basis_selects_which_date_the_window_uses`).

### New findings
- Nit: `offset` gets `_within_declared_bounds` too but declares no `maximum`, so an out-of-range `offset` fallback can still report a page the route clamped to `MAX_FEEDBACK_OFFSET`. Same asymmetry you documented in the declaration; only worth a line in the helper's docstring so the next reader doesn't assume both bounds are enforced.
- Nit: in the new `date_basis` test, the second `mock_feedback.query.return_value = {"Items": [straddling]}` is redundant (`return_value` persists without a `side_effect`) — harmless, but reads as if a reset were needed.

---

perrozzi:

Both nits from my last pass are closed and nothing else moved: the incremental diff is a docstring paragraph on `_within_declared_bounds` and the removal of the redundant `return_value` reset in the `date_basis` test. No declaration, projection or published shape changed, so `MCP_SERVER_VERSION` / the fingerprint correctly stay where they are.

### Resolved
- `_within_declared_bounds`' docstring now states that only the bounds the declaration carries are clamped, and names `offset`/`MAX_FEEDBACK_OFFSET` as the deliberate asymmetry (`lambda/api/mcp_handler.py`).
- The duplicate `mock_feedback.query.return_value` assignment in `test_the_facets_date_basis_selects_which_date_the_window_uses` is gone, with a comment explaining why one assignment covers both invocations (`lambda/api/test/test_mcp_delegation.py`).

Clean delta — no new findings.

---

perrozzi:
Both round-3 nits taken in `9a87871`, comment-only.

`_within_declared_bounds`' docstring now says it enforces **only what the declaration states**, and names the `offset` exception explicitly: `offset` declares a `minimum` and deliberately no `maximum`, because the route's ceiling is `MAX_FEEDBACK_OFFSET` in a Lambda bundle this one must not import — so the declaration names it in prose instead. You were right that a reader arriving at the clamp rather than at the declaration would assume symmetry. The asymmetry belongs to the declaration, and it is the honest option: the alternative is an un-CI-able copy of another bundle's constant, which is what that comment and `_DAYS_ARG`'s already argue against.

The redundant second `return_value` assignment in the `date_basis` test is gone, replaced by one assignment plus a note that it persists without a `side_effect` — which is what is actually true.

No behaviour changed and no published declaration moved, so the fingerprint and `MCP_SERVER_VERSION` stay put. Gates unchanged: **3514 passed**, ruff **383** (histogram identical to `origin/development`). Both round-2 guards were re-run under mutation after these edits and still fail on their own assertion.

Treating this as converged from my side: round 3 was nit-only and these were its last two. Happy to pick anything else up if you see it.

perrozzi:
## Merged `development`, deployed, and verified live — five new tools exercised end to end

`development` merged in (`1055934`), then deployed to the dev environment and verified over real
HTTPS. **19 live checks pass.**

### The merge

Brought in #373 (chat lookback), #374 (dead published assets) and #375 (bucket access logging).
**Zero file overlap** — `development` touched frontend, stream tools, core-stack and docs; this branch
touches only `mcp_handler.py` and its four MCP test files — so the merge was clean and taken to keep
the branch mergeable and to run every gate against the combined tree.

| Gate on the merged tree | Result |
|---|---|
| `pytest lambda -q --no-cov` | **3521 passed** (3514 + development's lockstep test) |
| `ruff check lambda` | 383, unchanged |
| CDK `vitest run` | **297 passed** — includes the `DOMAIN_ROUTES` lockstep, so the four new routes are proven handler-registered, gateway-wired and MCP-role-invokable |
| frontend `vitest` / `lint` / `tsc` | 186 files, exit 0, exit 0 |

### The deploy

`cdk diff` first, and it was the cheap thing that made this safe: **zero IAM changes**, 25 × `Code`
plus the frontend bundle key, only two resource types, nothing destroyed. That is what made
`--require-approval never` defensible. Deployed `VocApiStack --exclusively`; only that stack moved.

Confirmed from the **deployed artifact**, not from the exit code — downloaded the live Lambda bundle:

- `MCP_SERVER_VERSION = "3.8.0"`
- all five tools present, each with both a declaration and a handler
- both round-2/3 helpers shipped (`_delegated_object`, `_within_declared_bounds`)

### Live E2E over real HTTPS

| Check | Result |
|---|---|
| `initialize` | 200, `serverInfo.version` = **3.8.0** |
| `tools/list` | **11 tools** advertised; all five new ones listed |
| `list_feedback` | 200 — **4 real items**, `total=4`, projected rows carrying live values |
| `get_similar_feedback` | 200 against a **real** id — `count=3`, `source_feedback_id` echoed |
| `list_urgent_feedback` | 200 — `count=1` |
| `list_feedback_facets` | 200 — `feedback_count=8`, 2 categories, `is_partial=false`, `keywords` **absent** as declared |
| `list_jobs` | 200 — `count` + `jobs` |
| credential-less `tools/list` | **401** — the gateway authorizer refuses before the Lambda |

Two of those are the round-2 fixes, confirmed live rather than argued:

- **The clamp works.** A requested `limit: 500` came back as **100** — the declared `maximum`. Nothing
  rejects an out-of-range argument before the handler, so without the clamp the tool would have
  reported a page no route would serve.
- **No zero page.** `limit: 2, offset: 0` echoed exactly, and `count=2` beside `total=4` is the whole
  reason `list_feedback` exists over `search_feedback` — a model can tell a page from the window.

I deliberately pushed past a weak result here. The first run used a 7-day window, got `count=0`
everywhere, and passed. But all-zero proves the route was *reached*, not that it can *return data* —
a soft-failing read looks identical. Widening to 30 days produced real rows, which is the assertion
worth having.

### One finding from the E2E, and it is not a defect

`get_similar_feedback` first failed with `-32602 "feedback_id must be a non-empty string"`. That was
**my harness**, not the product: the projection renames `feedback_id` → **`id`**, so the key
identifying a row in `list_feedback`'s *output* is not the key `get_similar_feedback` takes as
*input*. I read the wrong one and passed nothing, and the handler's `InvalidToolArgument` guard
refused it correctly — which is the guard `minLength: 1` was added to match in round 1, firing live.

Worth flagging as ergonomics rather than a bug: a model chaining these two tools has to know that
`id` from one is `feedback_id` for the other. The rename lives in the shared `_project_feedback` and
predates this PR (`search_feedback` has the same shape), so it is out of scope here — but a sentence
in `get_similar_feedback`'s `feedback_id` description saying "the `id` field of a `list_feedback` or
`search_feedback` item" would remove the guesswork for one line and no behaviour change. Happy to add
it if you want it in this PR.

### Notes

- `scripts/test-api.sh` still dies at `Authenticating…` — the committed `deployment-test` password is
  rotated. Pre-existing and already tracked; unrelated to this deploy. The MCP path carries its own
  credential, which is why the battery above did not need it.
- The credential used was a short-lived read-only MCP token written directly to the projects table
  (one additive row, 60-minute TTL), **deleted afterwards and verified absent**.
- `cdk deploy` hung once after publishing assets — the recorded symptom, except `--method=direct` did
  **not** prevent it this time. Killing and re-running cleared it in 204 s with the published assets
  reused; the stack never entered `UPDATE_IN_PROGRESS` on the first attempt, so nothing was
  half-applied.
- The deployed frontend bundle is stamped with this branch's head rather than a `development` commit,
  which is expected for a PR-branch deploy and will correct itself on the post-merge redeploy.

PR #384, opened 2026-08-25
feat(prioritization): enlarge a row's prototype to fill the viewport

perrozzi:
## Summary

The remaining slice of issue #314. The QR halves (`FormQrCode/`, `SessionQrCode`) and the signed-URL
refresh scheduler have already shipped and are untouched here; this adds the one affordance that was
left: a way to enlarge the prototype a prioritization row is showing, and get back.

A pitch session looks at the prototype in a 384px pane inside half a table row. "Open in new tab"
already serves the reader who wants the artifact out of the app; it does not serve the *meeting*,
because the team's numbers, the reader's own sliders and the room-vote panel are all on this page and
a new tab abandons them. So the overlay is in-page.

**Three files, plus eight catalogues:**

- **`pages/Prioritization/PrototypeEnlargeButton.tsx`** (new) — the trigger and the full-viewport
  dialog it opens. It owns the *size* and nothing else: the prototype arrives as `children`, and the
  caller passes the very element it renders in the row.
- **`pages/Prioritization/PRFAQRow.tsx`** — `PrototypePanel` now describes its pane once and renders
  that description in two boxes (the row's `h-96` one and the overlay's full-height one). The control
  sits beside the existing open-in-a-new-tab anchor. A `PrototypePanelHeader` was split out; see
  *Decisions* for why.
- **`Prioritization.prototypeEnlarge.test.tsx`** (new, 13 tests) — one per behaviour, each shown red
  against the unfixed code (table below).
- `preview.enlarge` added to `prioritization.json` in all eight locales (en/es/fr/de/pt/ja/zh/ko).
  The dialog's close control reuses the existing `common:actions.close`, which is already present in
  all eight, so exactly one new string ships.

### Why it reuses `HtmlPrototypeFrame` rather than adding a second frame

`HtmlPrototypeFrame` carries `useLoadedUrl` — the handling that lets an hourly re-signing pass
*without* reloading the prototype under a reviewer who is several screens in, and that lets a URL
which was already dead on arrival be replaced. A frame written for the overlay would be free to drift
from that, and the drift would be invisible until somebody parked a pitch on screen for an hour. The
sizing difference is entirely the *container*, so `className` and a wrapper are the whole of it.

Dialog behaviour is `ModalShell`'s for the same reason: the audit behind #283 found 21 of 23 overlays
in this app missing dialog semantics precisely because each re-implemented them.

### Constraints honoured

- The open (and download) affordances remain **plain anchors**, per the reasoning in
[...]

PR #387, opened 2026-08-26
refactor: extract ModalShell frame focus helpers

### LINKED ISSUE #385: [Refactor]: extract ModalShell's pure cross-document focus arithmetic into frameFocus.ts
## Summary

`components/ModalShell/ModalShell.tsx` now carries a group of pure, module-level functions that do cross-document focus arithmetic and have nothing to do with React, the dialog's lifecycle, or the shell's props. They should live in their own module, `components/ModalShell/frameFocus.ts`, and be unit-tested directly.

Raised in review of #384 by both an automated reviewer and the author, who agreed it was out of scope for that PR: it would have mixed a large no-behaviour-change move into a diff whose reviewable content was a behavioural fix.

## Current state

At `c1e3ca4c`, the directory holds `ModalShell.tsx`, `ModalShell.test.tsx` and `index.tsx`. The pure group inside the component file is:

`FOCUSABLE_SELECTOR` · `focusCandidates` · `reachable` · `focusable` · `hasFocusable` · `frameDocument` · `asFrame` · `asElement` · `isFrameNode` · `holdsFrame` · `hostFrame` · `nestedDocuments` · `tabWouldLeave` · `tabOutOfFrame` · `tabAcrossFrame` · `tabWouldEnterFrame` · `tabWithinPanel`

Every one is a function of its arguments. None reads component state, props or a ref.

## Why it is worth doing

- **The subtle bugs in this area were all in the pure half.** The regression #384's last round fixed was `tabWouldLeave` asking the raw candidate count instead of the reachable one; the round before fixed a cross-realm `instanceof` in `asFrame`; the one before that fixed `tabOutOfFrame` resuming in the wrong document. Each was reached through `render` + a dispatched `KeyboardEvent`, which is a wide funnel for a narrow arithmetic question.
- **Direct unit tests would express those cases far more cheaply.** "A document whose candidates are all hidden reads as leaving" is one assertion over a `Document`, versus a rendered dialog, a frame, a written body and a synthetic keydown.
- **The component file would shrink to the part that is actually about React** — the effect, the refs, the capture invariant and the shell stack.

## Proposed shape

- Move the group to `components/ModalShell/frameFocus.ts` with no behaviour change, keeping the existing doc comments verbatim: they record which defect each line prevents and are the most valuable thing in the file.
- Add `frameFocus.test.ts` exercising the arithmetic directly, and keep the existing `ModalShell.test.tsx` cases that cover the *wiring* (which triggers re-scan, what Escape does, what the trap does at the panel's edges).
- Leave `index.tsx`'s public surface untouched — this is internal to the folder.

## Acceptance criteria

- [ ] The pure functions live in `frameFocus.ts`, and `ModalShell.tsx` imports them.
- [ ] No behaviour change: the existing `ModalShell.test.tsx` suite passes unmodified, except for imports if it reaches for a moved symbol.
- [ ] `frameFocus.test.ts` covers, as direct unit assertions, at least the three cases that were regressions in #384: all-candidates-hidden reads as leaving; `asFrame` resolves a frame belonging to another realm; `tabOutOfFrame` resumes in the frame's own document rather than the panel's.
- [ ] Each new test fails when the behaviour it covers is reverted, stated per test.
- [ ] `npx eslint . --max-warnings 0` and `npx tsc -b --noEmit` clean from `voc-datalake/frontend`; full frontend suite green with no net loss of cases.

## Out of scope

- Any change to focus behaviour. This is a move plus tests; if a defect turns up while moving, it wants its own issue and its own fix.
- Extracting anything from the effect itself (`listen`/`forget`/`reap` close over `listening`, `watching` and `panel`).

slegarraga:
Closes #385

## What
- Moves the cross-document focus helpers from `ModalShell.tsx` into `components/ModalShell/frameFocus.ts`.
- Keeps the existing wiring and public folder surface unchanged.
- Removes the one module-global `document` read from `tabOutOfFrame`; the panel document is now derived from the supplied panel items.
- Adds direct unit coverage for the three #384 regression cases:
  - all candidates hidden reads as leaving;
  - `asFrame` resolves a frame constructed in another realm;
  - `tabOutOfFrame` resumes in a nested frame's own document before walking outward.
- Each test documents the guarded behavior it pins.

## Verification
- `npm run typecheck`
- `npm run lint`
- `npm test -- --run` — 188 files / 3,121 tests passed.
- Targeted `ModalShell` + new `frameFocus` suite — 43 tests passed.


perrozzi:

This PR is a pure-move refactor: the cross-document focus helpers are lifted out of `ModalShell.tsx` into a new `components/ModalShell/frameFocus.ts` (all helpers exported verbatim, with doc comments intact) plus a small direct unit suite. The move itself looks clean and mechanical, but it is not behaviour-preserving: `tabOutOfFrame` swapped its terminal condition from `owner === document` to `owner === items[0]?.ownerDocument`, and that change is neither fail-closed nor covered by the new tests.

### Blockers

- `voc-datalake/frontend/src/components/ModalShell/frameFocus.ts` (`tabOutOfFrame`): deriving the panel document from `items[0]?.ownerDocument` fails open when `items` is empty. `undefined === owner` is false, so a frame that lives in the **top** document falls through to the "intermediate document" branch: `siblings = focusable(document.body)` walks the *whole page*, `siblings.indexOf(frame)` succeeds, and the function returns the focusable that follows the frame in the page — i.e. focus is moved **outside the dialog**. The old `owner === document` check could not reach that branch for a top-document frame. Every other edge in this module is deliberately fail-closed (`at === -1 → items[0]`, `hostFrame === null → items[0]`, the `tabAcrossFrame` detached-frame fallback); this one now leaks out of the trap. Prefer passing the panel document (or the panel element) in explicitly, e.g. `tabOutOfFrame(frame, panelDoc, items, back)`, and keep the identity check independent of `items` being non-empty and non-stale.

- The one behavioural change in the PR has no regression test. `frameFocus.test.ts`'s third case ("resumes in the frame's own document") passes identically under the old `owner === document` predicate — `nestedDocument !== document` either way — so it pins the pre-existing outward walk, not the new panel-document derivation. Per the repo rule, add a test that fails when the change is reverted (frame owned by the panel document → wrap through `items`) and one for the empty-`items` case above.

### Should fix

- `frameFocus.test.ts`: `new ForeignFrame() as unknown as HTMLIFrameElement` is exactly the double assertion the module's own comments say the repo bans, and the surrounding rule is "prefer type guards over `as`". Consider constructing the foreign realm with a real nested `<iframe>`/`contentWindow` document (jsdom gives you a second realm for free) rather than a hand-rolled object with a faked `ownerDocument`.
- Export surface: `reachable`, `focusCandidates`, `asElement`, `frameDocument`, `hostFrame`, `tabWouldEnterFrame` and `FOCUSABLE_SELECTOR` are now public but consumed by neither `ModalShell.tsx` nor the tests. Since the PR states the folder's public surface is unchanged, keep the internal ones module-private (or tag `@internal`) so the shell's invariants can't be bypassed from elsewhere.

### Nits

- `frameFocus.ts`: stray double blank line after the `@module` block.
- `frameFocus.test.ts`: non-null assertion on `nestedDocument.querySelector<HTMLIFrameElement>('iframe')!` — a small `expect(...).not.toBeNull()` (or a narrowing helper) keeps the file assertion-free.
- The `tabOutOfFrame` comment now claims the change makes the arithmetic "injectable"; nothing is actually injected — worth rewording once the panel document becomes a real parameter.
- Only the removals in `ModalShell.tsx` are visible in the diff, so I could not verify that the remaining component body needs no more than the six imported helpers (e.g. the `raisedIn`/`reap` call sites the comments reference); `npm run typecheck` presumably covers this.

---

slegarraga:
Great catches — thank you. I addressed both blockers and the cleanup notes in 46b6e8c:

- `tabOutOfFrame` now takes the panel `Document` explicitly, so its terminal identity no longer depends on a non-empty or current `items[0]`.
- The empty-panel path is fail-closed (`null`) instead of walking the host page.
- Added regression coverage for both: a panel-document frame wraps through the supplied item order, and an empty panel returns `null`.
- Replaced the hand-rolled cross-realm object with a real jsdom iframe realm and removed the test non-null assertion.

Validation: typecheck, full lint, and all 188 files / 3,123 frontend tests pass locally.

perrozzi:

Since the last review the panel document is now an explicit parameter (`tabOutOfFrame(frame, panelDocument, items, back)` / `tabAcrossFrame(doc, panelDocument, …)`, wired from `ModalShell.tsx` with `document`), the index reads gained `?? null`, the double assertion in the test was replaced with a real jsdom iframe realm plus a `required()` helper, and two tests were added. That addresses both blockers in substance, but one of the new regression tests does not actually discriminate the change.

### Resolved
- `tabOutOfFrame` no longer derives the panel document from `items[0]?.ownerDocument`, so the empty/stale-`items` fail-open path is gone.
- `?? null` on both index reads makes the empty-`items` return honour the declared `HTMLElement | null`.
- `frameFocus.test.ts` drops `as unknown as HTMLIFrameElement` in favour of a genuine second realm, and the non-null assertion is replaced by `required()`.
- The misleading "injectable" comment was reworded; the "stays fail-closed when the panel has no focusable items" test pins the new `?? null`.

### Still open
- `frameFocus.ts` exports `reachable`, `focusCandidates`, `asElement`, `frameDocument`, `hostFrame`, `tabWouldEnterFrame`, `FOCUSABLE_SELECTOR` although nothing outside the module consumes them.
- Stray double blank line after the `@module` block in `frameFocus.ts`.

### New findings

- **Blocker — the new panel-document test passes under a revert.** `frameFocus.test.ts`, "wraps a panel-document frame through the supplied panel order": `items` is exactly the panel document's focusable order (`[first, iframe, last]`), so the old global-`document` predicate takes the *intermediate* branch and computes `siblings.indexOf(frame) + 1` → the same `last` button. Both predicates return the identical element, so this pins nothing. To discriminate, make `items` diverge from the document's DOM order — e.g. keep a focusable control in the body that is absent from `items`, and put the frame at the end of `items` so the panel branch wraps to `items[0]` while the intermediate branch would step to the extra control.

- **Should fix — `tabAcrossFrame` was not made consistent with the new `?? null`.** `frameFocus.ts` still ends with `return frame ? tabOutOfFrame(…) : items[0]`, which yields `undefined` (typed `HTMLElement | null`) for empty `items`; the caller's `if (next === null) return` then falls through to `next.focus()`. `ModalShell.tsx` guards `items.length === 0` today, but the module is now independently importable and directly unit-tested, so use `items[0] ?? null`.

- **Nit — `panelDocument` is the second of two adjacent `Document` parameters** in both `tabOutOfFrame` and `tabAcrossFrame`, so transposing `doc`/`panelDocument` at a call site is type-silent; neither JSDoc block documents the new parameter. A `@param panelDocument` line (or an options object) would make the ordering self-checking.

---

slegarraga:
Thanks for the sharp follow-up — all addressed in d3498e8:

- The panel-order regression now makes the two branches disagree: `items` excludes a body control placed after the frame, so the panel branch wraps to `items[0]` while reverting to intermediate-DOM traversal would select that excluded control.
- `tabAcrossFrame` now uses `items[0] ?? null`, matching `tabOutOfFrame` on an empty panel.
- Added `@param panelDocument` to both functions and removed the stray double blank line.

Typecheck, full lint, and 188 files / 3,123 tests pass.

perrozzi:

Since the last review the panel-document regression test was reshaped so `items` diverges from DOM order, `tabAcrossFrame`'s fallback gained `?? null`, and the stray blank line was removed. Both prior blockers are addressed; the parameter-documentation nit was attempted but landed as an orphaned comment block.

### Resolved
- `frameFocus.test.ts` "wraps a panel-document frame…": `items = [iframe, #first]` now yields `#first` on the panel branch while the reverted intermediate branch would step to `#outside-items`, so the test discriminates.
- `tabAcrossFrame` returns `items[0] ?? null`, consistent with `tabOutOfFrame` and the declared `HTMLElement | null`.
- Double blank line after the `@module` block is gone.

### Still open
- `reachable`, `focusCandidates`, `asElement`, `frameDocument`, `hostFrame`, `tabWouldEnterFrame`, `FOCUSABLE_SELECTOR` are exported with no external consumer.

### New findings
- **Should fix — orphaned JSDoc block.** `frameFocus.ts` lines ~305-307: the new `/** @param panelDocument … */` block sits between `tabOutOfFrame` and `tabAcrossFrame`'s own JSDoc, so it is attached to nothing and documents neither function; tooling will drop it. Move a `@param panelDocument` line *inside* the existing JSDoc of both `tabOutOfFrame` and `tabAcrossFrame` (the two-adjacent-`Document`-parameters ordering hazard is still undocumented at either call site).

---

perrozzi:

The only change since my last review moves the orphaned `/** @param panelDocument … */` block into the existing JSDoc of both `tabOutOfFrame` and `tabAcrossFrame`. That addresses the last should-fix; the change is clean and carries no behavioural risk.

### Resolved
- `frameFocus.ts`: the dangling JSDoc block between `tabOutOfFrame` and `tabAcrossFrame` is gone, with `@param panelDocument` now inside each function's own comment so tooling picks it up at both call sites.

### Still open
- `reachable`, `focusCandidates`, `asElement`, `frameDocument`, `hostFrame`, `tabWouldEnterFrame`, `FOCUSABLE_SELECTOR` are exported with no consumer outside the module or its test.

---

slegarraga:
Fixed the orphaned JSDoc in 9d587c1: `@param panelDocument` now lives inside both existing doc blocks, so the adjacent `Document` parameter ordering is documented at each signature. Targeted typecheck, lint, and ModalShell/frameFocus tests pass.

PR #395, opened 2026-08-30
fix(projects): validate merge body before job creation

### LINKED ISSUE #380: [Security]: POST /documents/merge creates a billed job from a non-object body (same class as #377)
## Summary

`POST /projects/{project_id}/documents/merge` creates a billed job row from a request body that is not a JSON object. It is the same defect PR #377 closed on `POST /projects/{project_id}/document`, one route away, and it is the only one of the three adjacent generation routes that produces a job rather than a 500.

## Measured

Driving `lambda_handler` directly (`create_job` mocked to observe the call, everything else real):

| route | `[1,2]` | `"hi"` | `{not json` |
|---|---|---|---|
| `POST .../documents/merge` | **200, job CREATED** | **200, job CREATED** | 500, no job |
| `POST .../build-prototype` | 500, no job | 500, no job | 500, no job |
| `POST .../product-report` | 500, no job | 500, no job | 500, no job |
| `POST .../document` (fixed by #377) | 400, no job | 400, no job | 400, no job |

## Cause

`api_merge_documents` (`voc-datalake/lambda/api/projects_handler.py`) reads the body and creates the job before anything inspects the body's shape:

```python
body = app.current_event.json_body or {}
job_id, _ = create_job(project_id, 'merge_documents', 'merge_config', body, status='pending')
invoke_lambda_async(DOCUMENT_MERGER_FUNCTION, {..., 'merge_config': body})
```

Two problems, both already reasoned about in `_json_object_body`'s docstring:

1. A JSON array or string is truthy, so it passes `or {}` and is stored whole as `merge_config`. The async merger's first read is `merge_config.get('output_type', 'custom')` (`lambda/jobs/document_merger/handler.py:53`), which raises — **after** the job row exists. The user sees a job that fails; the account is billed for the invocation.
2. `or` collapses every falsy value, so `[]`, `false`, `0` and `""` arrive disguised as "no body" and are accepted, rather than refused.

## Fix

Swap the two lines for the existing helper, which already carries the reasoning and the tests:

```python
body = _json_object_body()
```

That answers 400 `the request body must be a JSON object` for a non-object and 400 `the request body must be JSON` for unparseable input, both before `create_job`. `_json_object_body` is in the same module and is already used by the prioritization routes and `api_generate_document`.

## Not the exposure, checked

`output_type` does reach a DynamoDB sort key in the merger, but only via `doc_type_prefix`, which is clamped to `prd`/`prfaq`/`doc` first (`document_merger/handler.py:144`, key built at :149). So this is **not** a second instance of the sort-key trust boundary that motivated `_validated_doc_type` — noting it because "same sort-key exposure" was an early reading and would send a fixer looking for something that is not there. The unguarded call there is `output_type.upper()` at :134, which fails inside the job.

## Scope note

`_json_object_body`'s docstring records that ~20 routes in this module share the latent `json_body or {}` shape and that sweeping them is its own diff. This issue is deliberately narrower: `api_merge_documents` is the one that creates a billed artefact from an unvalidated body, so it is worth fixing ahead of the sweep.

Raised out of PR #377 review, where the fix was judged out of scope for a change to one route.

kluacisco888:
## Summary

`POST /projects/{project_id}/documents/merge` now validates that a present request body is a JSON object **before** creating the job and invoking the asynchronous merger.

Fixes #380.

## The defect

The route currently does this:

```python
body = app.current_event.json_body or {}
job_id, _ = create_job(project_id, 'merge_documents', 'merge_config', body, status='pending')
```

That produces two failure modes:

1. Truthy non-objects such as `[1, 2]`, `"hi"`, `7` and `true` become `merge_config`, create a job, invoke the merger, and fail only later when the merger calls `.get(...)` on the value.
2. Falsy non-objects such as `[]`, `false`, `0` and `""` are collapsed by `or {}` and silently accepted as an empty configuration.

The first path leaves a failed job row and bills an asynchronous invocation for a request shape the route never supports.

## Change

Use the existing `_json_object_body()` helper in the same module:

```diff
-    body = app.current_event.json_body or {}
+    body = _json_object_body()
```

The helper is already used by adjacent prioritization and document-generation routes and already owns the distinction this route needs:

- absent body, zero-length body, and literal JSON `null` → `{}` (existing behavior preserved)
- JSON object → accepted unchanged
- JSON non-object → 400 `the request body must be a JSON object`
- unparseable JSON → 400 `the request body must be JSON`

Its scope docstring is updated to name `api_merge_documents` now that the route actually uses it.

## Automated-triage clarifications

- **Deployed API stage:** I reproduced through the real `lambda_handler`, not a deployed stage. There is no route-specific API Gateway request model or validator in this repository for `documents/merge`; the Lambda proxy handler is the request-shape boundary, so nothing upstream changes the body contract in deployment.
- **Legitimate empty callers:** absent, zero-length and JSON `null` bodies remain accepted as `{}`. Only a *present non-object* changes from accepted/500 to 400.
- **The wider sweep:** deliberately out of scope. The helper docstring records the remaining `json_body or {}` routes as a separate reviewable diff; this PR fixes the one route that creates a billed artifact before inspecting the shape.

## Tests

A new `TestMergeDocumentsBody` class covers 13 cases:

- valid object still creates and invokes a merge job;
- absent and JSON-null bodies still create a job with `{}`;
- eight non-object JSON values return 400 before `create_job`;
- two unparseable bodies return 400 before `create_job`.

Results:

```
$ pytest lambda/api/test/test_projects_handler.py -q --no-cov
118 passed
```

### Revert check

With the tests kept and only the handler restored to the old implementation:

```
10 failed, 3 passed
```

Restoring the one-line fix returns all 13 new tests to green. This proves the tests fail on the behavior being corrected rather than merely exercising the route.

### Ruff

`ruff check` on the two changed Python files reports the same three pre-existing findings in `projects_handler.py` (the long-standing import block and two broad exception catches at lines 1633 and 4257). No finding is in the new test class or changed route.

## Changelog

Adds a Security entry and an Upgrade note stating the new 400 behavior and the unchanged absent/null default.

## Scope

Exactly three files:

- `CHANGELOG.md`
- `voc-datalake/lambda/api/projects_handler.py`
- `voc-datalake/lambda/api/test/test_projects_handler.py`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -61,6 +61,9 @@ displays: the UI's build identifier is the short git commit SHA, injected at bui
   `prfaq` before creating the job. The field steered the job type, the execution path and the
   generated document's DynamoDB sort key straight from the request body, and each attempt billed a
   model call.
+- `POST /projects/{project_id}/documents/merge` now requires a JSON object before creating its job.
+  Arrays, strings, numbers and booleans previously became `merge_config` and failed only inside the
+  asynchronously invoked merger, after the job row existed and the invocation had been billed.
 
 ### Upgrade notes
 
@@ -75,6 +78,10 @@ displays: the UI's build identifier is the short git commit SHA, injected at bui
   previously started a default `prd` generation. Unparseable JSON is a 400 too, where it was
   previously a 500. A body that is absent altogether, a literal JSON `null`, or zero-length
   (`Content-Length: 0`), still means "generate a PRD with the defaults" and is unchanged.
+- **`POST /projects/{project_id}/documents/merge` now applies the same body-shape contract before
+  creating a job.** A present array, string, number or boolean, and unparseable JSON, answer 400.
+  An absent body, JSON `null` or zero-length body still means an empty merge configuration and is
+  unchanged.
 
 ## [0.2.0] - 2026-08-19
 
--- a/voc-datalake/lambda/api/projects_handler.py
+++ b/voc-datalake/lambda/api/projects_handler.py
@@ -438,7 +438,7 @@ def api_create_document(project_id: str):
 @tracer.capture_method
 def api_merge_documents(project_id: str):
     """Merge multiple documents."""
-    body = app.current_event.json_body or {}
+    body = _json_object_body()
     job_id, _ = create_job(project_id, 'merge_documents', 'merge_config', body, status='pending')
     invoke_lambda_async(DOCUMENT_MERGER_FUNCTION, {
         'project_id': project_id,
@@ -839,18 +839,10 @@ def _json_object_body() -> dict:
     Not extracted into `shared/` while it has two copies; the third one should do
     that rather than a second refactor of the first two.
 
-    SCOPE: the prioritization routes and `api_generate_document`. The other bodies
-    in this module have the same latent shape and predate this change, and
-    sweeping ~20 pre-existing routes is its own reviewable diff rather than a
-    rider on a change to one route.
-
-    One of those routes is not merely unswept but actively defective, and it is the
-    one a reader here would most likely assume is covered: `api_merge_documents`
-    still reads `json_body or {}` and calls `create_job` BEFORE anything inspects
-    the body, so a JSON array or a bare string produces a billed job row whose
-    `merge_config` is not an object. Tracked with the measurements in issue #380;
-    the fix is to call this helper. Adopting it here rather than there was the
-    scope line, not a judgement that the other route is fine.
+    SCOPE: the prioritization routes, `api_generate_document`, and
+    `api_merge_documents`. The other bodies in this module have the same latent
+    shape and predate this change; sweeping ~20 pre-existing routes is its own
+    reviewable diff rather than a rider on a change to one route.
     """
     try:
         body = app.current_event.json_body
--- a/voc-datalake/lambda/api/test/test_projects_handler.py
+++ b/voc-datalake/lambda/api/test/test_projects_handler.py
@@ -866,6 +866,105 @@ class TestGenerateDocumentDocType:
         )
 
 
+
+
+class TestMergeDocumentsBody:
+    """POST /projects/<id>/documents/merge validates the body before billing."""
+
+    @staticmethod
+    def _post(api_gateway_event, lambda_context, raw_body=None):
+        from projects_handler import lambda_handler
+
+        event = api_gateway_event(
+            method='POST',
+            path='/projects/proj-123/documents/merge',
+            path_params={'project_id': 'proj-123'},
+        )
+        if raw_body is not None:
+            event['body'] = raw_body
+
+        with patch(
+            'projects_handler.create_job', return_value=('job-1', {})
+        ) as mock_create_job, patch(
+            'projects_handler.invoke_lambda_async'
+        ) as mock_invoke:
+            response = lambda_handler(event, lambda_context)
+        return response, mock_create_job, mock_invoke
+
+    def test_an_object_body_still_starts_a_merge_job(
+        self, api_gateway_event, lambda_context
+    ):
+        response, mock_create_job, mock_invoke = self._post(
+            api_gateway_event,
+            lambda_context,
+            '{"output_type": "prd"}',
+        )
+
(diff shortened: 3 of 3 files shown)


perrozzi:

`api_merge_documents` now routes its request body through the existing `_json_object_body()` helper so a present non-object body returns 400 instead of creating a job row and billing an async merger invocation. The change is one line, matches the pattern already used by the prioritization and `api_generate_document` routes, is documented in `CHANGELOG.md` (including the falsy-non-object behavior change), and comes with a regression test class that the description confirms fails when the fix is reverted. This looks clean; only formatting nits below.

### Should fix

- `voc-datalake/lambda/api/test/test_projects_handler.py`: the new class runs straight into the next one with no separation — `mock_invoke.assert_not_called()` is immediately followed by `class TestDocumentCRUDEndpoints:`. Add the two blank lines (PEP 8 E302). Ruff's default rule selection doesn't include E3, which is likely why the reported ruff output was clean, but it still reads as if the class boundary was lost in a merge.

### Nits

- Same file: four blank lines are now present before `class TestMergeDocumentsBody` (two pre-existing plus two added). Trim to two.
- `test_an_absent_or_null_body_keeps_the_empty_config_default`'s `absent` case relies on `api_gateway_event` not supplying a default `body`. Setting `event['body'] = None` explicitly (or asserting `'body' not in event or event['body'] is None`) would make the case self-describing and immune to a fixture default changing later.
- The docstring rewrite drops the only in-repo pointer to the tracking issue for the remaining `json_body or {}` routes. Since the SCOPE note still defers "~20 pre-existing routes," consider citing a follow-up issue number there so the sweep stays discoverable.

---

perrozzi:

Only the test file changed since my last review: the stray blank lines around `TestMergeDocumentsBody` were normalized to two, and `_post` now always assigns `event['body'] = raw_body` so the "absent" case is explicit rather than relying on the fixture default. All three formatting/clarity nits are addressed; the handler and CHANGELOG are untouched.

### Resolved
- PEP 8 E302 separation restored between `TestMergeDocumentsBody` and `TestDocumentCRUDEndpoints`.
- Extra blank lines before `TestMergeDocumentsBody` trimmed to two.
- `absent` case now sets `event['body'] = None` explicitly instead of depending on `api_gateway_event` not supplying a default body.

### Still open
- Nit only: the `_json_object_body()` SCOPE note still defers "~20 pre-existing routes" with no issue number now that the #380 pointer is gone.

### New findings
- Nit, `voc-datalake/lambda/api/test/test_projects_handler.py`: with the unconditional assignment, the `absent` and `json_null` parametrize cases both reach the handler as `json_body is None`, so they are now the same path under two ids. Harmless (API Gateway proxy does represent a missing body as `body: None`), but the `absent` label no longer covers a genuinely missing `body` key.

Otherwise clean — no functional concerns with the incremental changes.

---

kluacisco888:
Addressed the automated review feedback in `07cd024`:

- normalized the blank lines on both sides of `TestMergeDocumentsBody`;
- made the no-body case explicit by always setting `event['body'] = raw_body` (including `None`);
- re-ran the 13 merge-body tests and the full `test_projects_handler.py` file (118 passed);
- ran default Ruff plus Ruff preview `E302`; both passed.

For the tracking-reference nit, I searched the repository issues and did not find a follow-up issue for the remaining `json_body or {}` routes. Issue #380 specifically tracked the merge route fixed by this PR, so I left the remaining-scope note without attributing that broader sweep to #380.

kluacisco888:
Thanks for the follow-up. No additional code change is needed for the new nit:

- `api_gateway_event()` already always emits a `body` key; when no body argument is supplied, the fixture sets `body: None`, matching the API Gateway proxy representation.
- The `absent` case therefore still sends raw `None`, while `json_null` sends the raw JSON text `'null'`.
- Those are distinct request inputs that intentionally converge to `app.current_event.json_body is None`, because the route preserves the same empty-config behavior for both no payload and an explicit JSON null.

For the remaining SCOPE-note nit, there is still no repository issue tracking the broader ~20-route sweep, and #380 specifically tracks the merge route completed by this PR. I am leaving the note unlinked rather than attributing broader scope to the wrong issue.

PR #402, opened 2026-09-02
fix(frontend): label Automatic from the deployment-wide model pin (#275)

### LINKED ISSUE #275: [Bug]: AI Models card shows wrong Automatic model when a global pin is deployed
## Summary

When a deployment sets the global model pin (`-c defaultModelId=<id>`, which seeds `settings.model_id`), the Settings → **AI Models** card still labels every surface with its built-in `SURFACE_DEFAULTS` model. The global pin outranks `SURFACE_DEFAULTS` in the resolver, so all five "Automatic — X" labels name a model that is not the one being used.

In a Workshop Studio event account the named models are also **not invocable**, so the card advertises Sonnet 5 / Opus 5 as the Automatic choice in an account where both return `AccessDeniedException`.

## Observed

Event account deployed with `-c defaultModelId=global.anthropic.claude-sonnet-4-6`.

`GET /settings/model` response (abridged):

```json
{
  "surfaces": [
    {"key":"chat",       "default_id":"global.anthropic.claude-sonnet-5",  "selected":null},
    {"key":"documents",  "default_id":"global.anthropic.claude-sonnet-5",  "selected":null},
    {"key":"prototype",  "default_id":"global.anthropic.claude-opus-5",    "selected":null},
    {"key":"enrichment", "default_id":"global.anthropic.claude-haiku-4-5-20251001-v1:0", "selected":null},
    {"key":"utility",    "default_id":"global.anthropic.claude-sonnet-5",  "selected":null}
  ],
  "model_id": "global.anthropic.claude-sonnet-4-6"
}
```

What the card renders:

| Surface | Card says | Actually resolves to |
|---|---|---|
| AI Chat | Automatic — Claude Sonnet 5 | Sonnet 4.6 |
| Document Generation | Automatic — Claude Sonnet 5 | Sonnet 4.6 |
| Prototype Builder | Automatic — Claude Opus 5 | Sonnet 4.6 |
| Feedback Enrichment | Automatic — Claude Haiku 4.5 | Sonnet 4.6 |
| Utilities | Automatic — Claude Sonnet 5 | Sonnet 4.6 |

Confirmed against a real job — document/persona generation logged:

```
```

...while the card claimed Automatic for that surface was Sonnet 5.

## Root cause

The API already returns everything needed — it sends both `default_id` per surface **and** the active `model_id` global override:

```python
# lambda/api/settings_handler.py
surfaces = [
    {
        'key': key,
        'default_id': SURFACE_DEFAULTS[key],
        'selected': stored.get(key) if stored.get(key) in ALLOWED_MODEL_IDS else None,
    }
    for key in PICKER_SURFACES
]
return {
    'available_models': ALLOWED_MODELS,
    'surfaces': surfaces,
    'model_id': legacy_global if legacy_global in ALLOWED_MODEL_IDS else None,
}
```

The frontend builds the "Automatic — X" label from `default_id` and ignores `model_id`, even though `model_id` wins in `shared/model_config.py`:

```
per-surface override > legacy global override (settings.model_id) > SURFACE_DEFAULTS > BEDROCK_MODEL_ID
```

## Impact

- The card misstates which model powers every AI feature whenever a global pin is deployed — which is the documented workshop configuration.
- Worse for the workshop: a participant reading "Automatic — Claude Sonnet 5" may conclude the pin did not apply and explicitly select Sonnet 5 (it is offered in the dropdown). That replaces a working Sonnet 4.6 with a model the account cannot invoke, breaking every AI surface with `AccessDeniedException` — a self-inflicted outage the UI actively invites.

## Suggested fix

1. When `model_id` is present, render the Automatic label from it, e.g. `Automatic — Claude Sonnet 4.6 (pinned for this deployment)`, so the label reflects actual resolution. The frontend already has `available_models` to map id → label.
2. Consider marking or disabling options the account cannot invoke. Bedrock exposes this via `ListFoundationModels` / `GetFoundationModelAvailability`; even a one-time check cached in settings would stop the picker offering models that fail. At minimum, an inline warning when a selected model returns `AccessDeniedException` would be much better than a silent break across all surfaces.

perrozzi:
## Summary

Resolves #275.

`GET /settings/model` already returns everything the card needs: each surface's built-in `default_id` **and** the deployment-wide `model_id` override. `AiModelSectionBody` built every `Automatic — X` label from `default_id` alone, but `shared/model_config.py` resolves in the order *per-surface override > global pin (`settings.model_id`) > `SURFACE_DEFAULTS`*. A deployment with `-c defaultModelId=<id>` therefore advertised a model it never invokes on all five surfaces.

The change is one derived value in the one place the label is computed:

- `pinnedId` — the returned `model_id` when truthy, else `null`.
- `effectiveId` — `pinnedId ?? surface.default_id`, mapped through the **existing** `labelForId` map from `available_models`, with the existing `?? effectiveId` id-string fallback.
- `SurfaceRow`'s `defaultLabel` prop is renamed `automaticLabel`, because it is no longer necessarily the surface default. No other prop, no new translation key, no change to the `aiModel.automaticOption` shape.

`selected` is deliberately untouched — the pin only re-labels the Automatic option, so a stored per-surface choice remains the selected option, matching the resolver.

### Deliberately out of scope

Suggestion 2 in the issue (marking or disabling options the account cannot invoke, via `ListFoundationModels` / `GetFoundationModelAvailability`) is **not** in this PR. It needs a new Bedrock permission on the settings Lambda, a cache, and a response-shape change; the task asks to keep availability discovery and invocability checks separate. The label fix alone removes the trap the issue describes as worst — an admin reading a wrong Automatic label and "correcting" it to an uninvocable model.

I also did not add `Automatic — X (pinned for this deployment)` as the issue sketches: that suffix is new user-facing English and would need eight locale files plus a new key. The label now names the right model, which is the acceptance criterion; the parenthetical is a separate copy decision for a maintainer.

## Acceptance criteria

| Criterion | Test |
|---|---|
| Non-null `model_id`, null per-surface selections → every Automatic names the pin | `names the deployment-wide pin in every Automatic option` |
| `model_id` absent or null → Automatic keeps naming `default_id` | `keeps each surface default in its Automatic option when the pin field is absent`, plus the pre-existing `shows each surface default inside its Automatic option` (fixture has `model_id: null`) |
| A non-null per-surface `selected` stays the selected option | `keeps an explicit per-surface selection selected under a global pin` |
| Unknown effective id degrades to the id string | `falls back to the pinned id when no friendly label is available` |

## Build and test results

Run from the repo root unless noted. `node_modules` was absent in the container, so `npm ci` was run in `voc-datalake/` and `voc-datalake/frontend/` first.

| Command | Result |
|---|---|
| `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` (focused) | **PASS** — 13/13 |
| `npm run test` (full frontend vitest) | **PASS** — 192 files, 3285 tests, exit 0 |
| `npm run typecheck` (frontend src) | **PASS** — clean |
| `npm run lint:frontend` (ESLint `--max-warnings 0`) | **PASS** — clean |
| `npm run build` (`tsc -b && vite build`) | **PASS** — built in 9.36s |
| `npm run typecheck:tests` | 87 errors — **pre-existing baseline**, byte-identical count with my change stashed; none in the files I touched (`configStore.test.ts`, `test/test-utils.tsx` etc.) |
| `npm run i18n:check` | exit 1 — **pre-existing baseline**, same exit with my change stashed; findings are hardcoded strings in `DataExplorer/`, `FeedbackDetail/`, `ChatSidebar/` and others, untouched here |

`mise run build` was attempted as instructed and reports `no tasks defined in /workspace/...` — this repo has no mise task file, so the equivalent `npm run build` above is the real build gate.

### Test integrity (red-then-green + mutation)

- Against unfixed `AiModelSection.tsx`, the new assertions were **3 failed / 10 passed**. The two that passed before the fix are intentional negative controls (absent-pin and empty-string-pin must keep the surface default, which the old code did by accident).
- After the fix: **13 passed**.
- Mutation check on the truthiness guard: replacing `data.model_id ? data.model_id : null` with `data.model_id ?? null` → **caught**, `ignores an empty-string pin and keeps the surface default` fails. Working tree restored and re-verified green.

## Agent notes

**What went well.** The issue's root-cause analysis was accurate and the API already returned `model_id`, so the fix needed no backend change, no new prop plumbing beyond a rename, and no new dependency — one derived value at the single place the label is computed. Writing the tests first paid off immediately: it surfaced that two of the five new cases pass against the *unfixed* code, which correctly identifies them as controls rather than regression coverage, and that distinction is now recorded in the REVERT MAP instead of being implied.

**What was difficult.** Separating my results from the container's baseline. Two gates (`typecheck:tests`, `i18n:check`) fail on `development` already, and the setup notes said both the initial build and lint "FAILED" — that turned out to be missing `node_modules` plus the absent mise task, not a broken tree. I stashed and re-ran each failing gate to pin the baseline rather than guess; that is the only way to report these honestly. Also worth flagging that the repo's `npm run check` chains everything with `&&`, so on this tree it can never reach the later legs — triaging by individual script is the only workable approach.

**Conventions discovered.** Comments here carry the *argument*, not the mechanics — the load-bearing one on `pinnedId` records what the previous implementation was, why it was wrong, and which named test fails if it returns. Tests are named as outcomes (`names the deployment-wide pin ...`), and the module docstring carries a REVERT MAP mapping each mutation to the assertion that catches it. Model product names are deliberately absent from locale files because the i18n gate rejects same-as-English values, which is exactly why reusing the server-provided `label` mapping was the right lever rather than adding a key.

**Suggestions for future tasks.** (1) The uninvocable-model half of #275 deserves its own issue — it is a backend/IAM change, not a label change, and bundling it would have hidden this one-value fix. (2) `typecheck:tests` at 87 errors and `i18n:check` red on `development` mean neither can gate a PR today; a "no new findings" baseline check (the pattern `ruff.toml` already uses) would make them useful again. (3) A `mise` task file, or a blueprint `pipeline.buildCommand` of `npm run build`, would turn on the build-regression gate that the harness reported as inert.

## Verification

```abca-verify
selector #ai-model-chat
selector #ai-model-prototype
text #ai-model-chat "Automatic"
no-console-errors
```

The AI Models card is admin-only and renders one `<select id="ai-model-{surface}">` per surface, with the Automatic option present on load regardless of data — the assertions target that chrome. Which *model name* follows "Automatic" depends on whether the preview deployment sets a global pin, so I assert the stable prefix rather than a specific model; the pin-versus-default branch itself is covered by the Vitest cases above.

### DIFF
--- a/voc-datalake/frontend/src/pages/Settings/AiModelSection.test.tsx
+++ b/voc-datalake/frontend/src/pages/Settings/AiModelSection.test.tsx
@@ -1,5 +1,16 @@
 /**
  * @fileoverview Tests for the per-surface AI model picker (issue #96).
+ *
+ * REVERT MAP for the Automatic-label precedence (issue #275) — each assertion
+ * names the mutation it catches:
+ *   - labelling from `surface.default_id` while a global pin is deployed →
+ *     'names the deployment-wide pin in every Automatic option'.
+ *   - labelling from `data.model_id` unconditionally (so a null/absent pin
+ *     erases the tuned per-surface defaults) → 'shows each surface default
+ *     inside its Automatic option' and 'ignores an empty-string pin'.
+ *   - letting the pin drive the *selected* value, not just the Automatic label →
+ *     'keeps an explicit per-surface selection selected under a global pin'.
+ *   - dropping the id-string fallback → 'falls back to the pinned id'.
  */
 import { describe, it, expect, vi, beforeEach } from 'vitest'
 import { render, screen, waitFor, within } from '@testing-library/react'
@@ -19,13 +30,16 @@ vi.mock('../../api/client', () => ({
 }))
 
 const SONNET5 = 'global.anthropic.claude-sonnet-5'
+const SONNET46 = 'global.anthropic.claude-sonnet-4-6'
 const OPUS5 = 'global.anthropic.claude-opus-5'
 const HAIKU45 = 'global.anthropic.claude-haiku-4-5-20251001-v1:0'
+
+const SURFACE_LABELS = ['AI Chat', 'Document Generation', 'Prototype Builder', 'Feedback Enrichment', 'Utilities']
 
 const modelSettingsFixture = {
   available_models: [
     { key: 'sonnet5', id: SONNET5, label: 'Claude Sonnet 5', description: 'Latest Sonnet' },
-    { key: 'sonnet46', id: 'global.anthropic.claude-sonnet-4-6', label: 'Claude Sonnet 4.6', description: 'Previous Sonnet' },
+    { key: 'sonnet46', id: SONNET46, label: 'Claude Sonnet 4.6', description: 'Previous Sonnet' },
     { key: 'opus5', id: OPUS5, label: 'Claude Opus 5', description: 'Deepest reasoning' },
     { key: 'haiku45', id: HAIKU45, label: 'Claude Haiku 4.5', description: 'Fastest' },
   ],
@@ -79,8 +93,7 @@ describe('AiModelSection', () => {
     render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
 
     await waitFor(() => expect(screen.getByLabelText('AI Chat')).toBeInTheDocument())
-    const surfaceLabels = ['AI Chat', 'Document Generation', 'Prototype Builder', 'Feedback Enrichment', 'Utilities']
-    for (const label of surfaceLabels) {
+    for (const label of SURFACE_LABELS) {
       const select = screen.getByLabelText(label)
       expect(select).toHaveValue('')
     }
@@ -94,6 +107,70 @@ describe('AiModelSection', () => {
     expect(within(prototypeSelect).getByText('Automatic — Claude Opus 5')).toBeInTheDocument()
     const enrichmentSelect = screen.getByLabelText('Feedback Enrichment')
     expect(within(enrichmentSelect).getByText('Automatic — Claude Haiku 4.5')).toBeInTheDocument()
+  })
+
+  it('keeps each surface default in its Automatic option when the pin field is absent', async () => {
+    // Positive control for the null case above: `model_id` missing from the
+    // payload (not merely null) must not be read as a pin either.
+    const { model_id: _absent, ...withoutPinField } = modelSettingsFixture
+    mockGetModelSettings.mockResolvedValue(withoutPinField)
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('Prototype Builder')).toBeInTheDocument())
+    expect(within(screen.getByLabelText('Prototype Builder')).getByText('Automatic — Claude Opus 5')).toBeInTheDocument()
+    expect(within(screen.getByLabelText('AI Chat')).getByText('Automatic — Claude Sonnet 5')).toBeInTheDocument()
+  })
+
+  it('names the deployment-wide pin in every Automatic option', async () => {
+    // The resolver ranks settings.model_id above SURFACE_DEFAULTS, so a pinned
+    // deployment runs Sonnet 4.6 everywhere — the label has to say so (#275).
+    mockGetModelSettings.mockResolvedValue({ ...modelSettingsFixture, model_id: SONNET46 })
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('AI Chat')).toBeInTheDocument())
+    for (const label of SURFACE_LABELS) {
+      const select = screen.getByLabelText(label)
+      expect(within(select).getByText('Automatic — Claude Sonnet 4.6')).toBeInTheDocument()
+      expect(select).toHaveValue('')
+    }
+    // No surface keeps advertising its built-in default under a pin.
+    expect(screen.queryByText('Automatic — Claude Opus 5')).not.toBeInTheDocument()
+  })
+
+  it('keeps an explicit per-surface selection selected under a global pin', async () => {
+    // The pin only re-labels Automatic; a stored per-surface choice still wins
+    // in the resolver and must stay the selected option.
+    mockGetModelSettings.mockResolvedValue({
+      ...modelSettingsFixture,
+      model_id: SONNET46,
+      surfaces: modelSettingsFixture.surfaces.map((s) =>
+        s.key === 'chat' ? { ...s, selected: OPUS5 } : s,
+      ),
+    })
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('AI Chat')).toHaveValue(OPUS5))
+    expect(within(screen.getByLabelText('AI Chat')).getByText('Automatic — Claude Sonnet 4.6')).toBeInTheDocument()
+  })
+
+  it('falls back to the pinned id when no friendly label is available', async () => {
+    mockGetModelSettings.mockResolvedValue({ ...modelSettingsFixture, model_id: 'global.anthropic.claude-unknown-9' })
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('AI Chat')).toBeInTheDocument())
+    expect(
+      within(screen.getByLabelText('AI Chat')).getByText('Automatic — global.anthropic.claude-unknown-9'),
+    ).toBeInTheDocument()
+  })
+
+  it('ignores an empty-string pin and keeps the surface default', async () => {
+    // '' is not an allowlisted id; treating it as a pin would label Automatic
+    // with the empty string and lose the default entirely.
+    mockGetModelSettings.mockResolvedValue({ ...modelSettingsFixture, model_id: '' })
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('Prototype Builder')).toBeInTheDocument())
+    expect(within(screen.getByLabelText('Prototype Builder')).getByText('Automatic — Claude Opus 5')).toBeInTheDocument()
   })
 
   it('saves a per-surface selection and shows the Saved badge', async () => {
--- a/voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
+++ b/voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
@@ -10,6 +10,9 @@
  * Free-form model IDs are rejected server-side, and the PUT is admin-gated
  * server-side too (not just hidden in the UI).
  *
+ * The Automatic label follows the resolver's precedence, not just the surface
+ * default — see the comment on `pinnedId` below (issue #275).
+ *
  * Model NAMES are product names (identical in every language), so they render
  * from the server-provided `label` and are intentionally NOT in locale files —
  * the i18n gate rejects same-as-English values. Surface labels/descriptions and
@@ -113,20 +116,37 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface, 
   // Model id → display label (product name, served by the backend).
   const labelForId = new Map(data.available_models.map((model) => [model.id, model.label]))
 
+  // 🔑 The deployment-wide pin (`settings.model_id`, seeded by
+  // `-c defaultModelId=<id>`) outranks every surface's built-in default in
+  // shared/model_config.py: per-surface override > global pin > SURFACE_DEFAULTS.
+  // Labelling Automatic from `surface.default_id` alone — which is what this did
+  // before #275 — made a pinned deployment advertise a model it never invokes,
+  // and invited an admin to "correct" it by explicitly selecting the advertised
+  // model, which in a workshop account cannot be invoked at all. Truthiness, not
+  // `?? null`: the field can arrive absent or as '' from an older payload, and
+  // neither is a pin. Tests: 'names the deployment-wide pin in every Automatic
+  // option' and 'ignores an empty-string pin and keeps the surface default'.
+  const pinnedId = data.model_id ? data.model_id : null
+
   return (
     <fieldset className="space-y-3" disabled={isSaving}>
       <legend className="sr-only">{t('aiModel.title')}</legend>
-      {data.surfaces.map((surface) => (
-        <SurfaceRow
-          key={surface.key}
-          surfaceKey={surface.key}
-          selected={surface.selected}
-          defaultLabel={labelForId.get(surface.default_id) ?? surface.default_id}
-          models={data.available_models}
-          justSaved={savedSurface === surface.key}
-          onSelect={(modelId) => onSelect(surface.key, modelId)}
-        />
-      ))}
+      {data.surfaces.map((surface) => {
+        // What Automatic actually resolves to for this surface. The pin does not
+        // touch `selected`: a stored per-surface choice still outranks it.
+        const effectiveId = pinnedId ?? surface.default_id
+        return (
+          <SurfaceRow
+            key={surface.key}
+            surfaceKey={surface.key}
+            selected={surface.selected}
+            automaticLabel={labelForId.get(effectiveId) ?? effectiveId}
+            models={data.available_models}
+            justSaved={savedSurface === surface.key}
+            onSelect={(modelId) => onSelect(surface.key, modelId)}
+          />
+        )
+      })}
     </fieldset>
   )
 }
@@ -134,13 +154,14 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface, 
 interface SurfaceRowProps {
   readonly surfaceKey: string
   readonly selected: string | null
-  readonly defaultLabel: string
+  /** Label for the model Automatic resolves to — the global pin when one is deployed. */
+  readonly automaticLabel: string
   readonly models: ModelSettings['available_models']
   readonly justSaved: boolean
   readonly onSelect: (modelId: string | null) => void
 }
 
-function SurfaceRow({ surfaceKey, selected, defaultLabel, models, justSaved, onSelect }: SurfaceRowProps) {
+function SurfaceRow({ surfaceKey, selected, automaticLabel, models, justSaved, onSelect }: SurfaceRowProps) {
   const { t } = useTranslation('settings')
   const selectId = `ai-model-${surfaceKey}`
   return (
@@ -163,7 +184,7 @@ function SurfaceRow({ surfaceKey, selected, defaultLabel, models, justSaved, onS
           value={selected ?? ''}
           onChange={(event) => onSelect(event.target.value === '' ? null : event.target.value)}
         >
-          <option value="">{t('aiModel.automaticOption', { model: defaultLabel })}</option>
+          <option value="">{t('aiModel.automaticOption', { model: automaticLabel })}</option>
           {models.map((model) => (
             <option key={model.id} value={model.id}>{model.label}</option>
           ))}


perrozzi:

`AiModelSection` was labelling the "Automatic" option from each surface's `default_id`, ignoring the deployment-wide `model_id` pin that outranks it in the backend resolver; this PR derives `pinnedId` from `data.model_id` (truthiness, so `null`/`''`/absent are not pins), computes `effectiveId = pinnedId ?? surface.default_id`, renames `SurfaceRow`'s `defaultLabel` prop to `automaticLabel`, and adds five vitest cases including negative controls. The change is minimal, correctly scoped (no new i18n keys, `selected` untouched), and well covered — I see no blockers, only a coupling concern and a schema question.

### Should fix

- **The resolver precedence is now duplicated in the frontend** (`voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx`, `pinnedId`/`effectiveId`). `shared/model_config.py` owns the order *per-surface override > global pin > `SURFACE_DEFAULTS`*, and the UI now re-implements the top two tiers. If that order ever changes server-side (or a third tier appears), the label silently lies again and no test in this PR catches it, because the fixture is hand-written. Preferably have `GET /settings/model` return an `effective_id` per surface so precedence lives in exactly one place; failing that, add a pytest in the settings/`shared/model_config` suite that pins the documented order, and reference it from the comment here so the two sides fail together.
- **Confirm the type/Zod schema for `model_id` matches what the component and tests now assume.** The diff doesn't touch the `ModelSettings` type or its runtime validator, yet the new tests exercise `model_id` absent, `null`, and `''`. Since `mockGetModelSettings` is mocked at the `api/client` boundary, these tests bypass validation entirely — if the schema declares `model_id` as required (or non-nullable), the absent-field case is unreachable in production and the test is misleading; if it's `.optional().nullable()`, worth a one-line comment saying so. Per repo convention (Zod at system boundaries), the tolerance should be visible in the schema, not just in the component's truthiness check.

### Nits

- The `pinnedId` comment (`voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx`) is ~10 lines with a 🔑 emoji and names two tests by title. The *why* is valuable; the test titles will rot on the first rename — the REVERT MAP in the test file's docstring already carries that mapping, so consider trimming the duplication and dropping the emoji.
- `expect(screen.queryByText('Automatic — Claude Opus 5')).not.toBeInTheDocument()` in `AiModelSection.test.tsx` is document-global; scoping it with `within(...)` per surface (or asserting the count of `Automatic — Claude Sonnet 4.6` options is 5) states the intent more precisely.
- `SURFACE_LABELS` hardcodes the five English labels next to a fixture that already contains them; deriving it from `modelSettingsFixture.surfaces` would keep the two in sync.
- A case where the pin equals a surface's `default_id` (label unchanged, no double-render) would cheaply document the no-op branch.
- The out-of-scope notes on `typecheck:tests` (87 errors) and `i18n:check` belong in a follow-up issue rather than the PR body — agreed they shouldn't block this change.

---

voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
```diff
@@ -113,34 +116,52 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface,
...
-          defaultLabel={labelForId.get(surface.default_id) ?? surface.default_id}
-          models={data.available_models}
-          justSaved={savedSurface === surface.key}
-          onSelect={(modelId) => onSelect(surface.key, modelId)}
-        />
-      ))}
+      {data.surfaces.map((surface) => {
+        // What Automatic actually resolves to for this surface. The pin does not
+        // touch `selected`: a stored per-surface choice still outranks it.
+        const effectiveId = pinnedId ?? surface.default_id
```
perrozzi, line 137: **Type**: issue
**Severity**: minor
**Title**: The resolver's precedence is now encoded in two places with nothing pinning them together

**Description**: `shared/model_config.py::get_active_model_id` is the sole owner of *per-surface override > legacy global (`settings.model_id`) > `SURFACE_DEFAULTS` > `BEDROCK_MODEL_ID`*, and `lambda/stream/src/bedrock/model-override.ts::resolveModelOverride` already mirrors the top two tiers for streaming chat. `effectiveId = pinnedId ?? surface.default_id` makes this the **third** independent copy of that ordering. Nothing in this PR fails if the backend order changes: the vitest fixture is hand-written, so it asserts the frontend agrees with itself, not with the resolver.

(Informed by repository memory: this repo already treats cross-language duplication of model configuration as a build gate — `model_config.py` carries an explicit "MUST stay in lockstep with" header naming `model-override.ts` and `api-stack.ts::allowlistedModelArns`, and there are Python lockstep tests that read the TS mirror and the CDK helper and fail the build when the three disagree. The precedence order now has the same duplication shape but none of that protection.)

Note the resolution is genuinely uniform today — `SURFACE_DEFAULTS` covers all five `PICKER_SURFACES`, so `default_id` is never empty and the fourth tier (`BEDROCK_MODEL_ID`) is unreachable through this endpoint. So this is a drift-prevention concern, not a present-day miscomputation, hence minor.

**Proposed fix**: Preferred: have `GET /settings/model` return an `effective_id` per surface (computed by calling into `model_config`'s resolution, or by replicating it once server-side next to the code that owns it), so the frontend renders a value instead of deriving one and precedence lives in exactly one place. Failing that, add a lockstep-style test in `lambda/shared/test/test_model_config.py` (or `lambda/api/test/test_settings_handler.py`) that pins the documented order as data, and reference that test by name from the comment on `pinnedId` so the two sides fail together — the same pattern the allowlist already uses.

**AI prompt**: In aws-samples/sample-voice-of-customer-datalake, the Automatic-label precedence (global `settings.model_id` pin over per-surface `SURFACE_DEFAULTS`) is currently implemented independently in three places: `voc-datalake/lambda/shared/model_config.py::get_active_model_id`, `voc-datalake/lambda/stream/src/bedrock/model-override.ts::resolveModelOverride`, and now `voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx` (`pinnedId` / `effectiveId`). Remove the frontend copy: add an `effective_id` field to each entry of the `surfaces` array returned by `GET /settings/model` in `voc-datalake/lambda/api/settings_handler.py::get_model_settings`, computed from the same precedence the resolver documents (per-surface stored override > allowlisted `model_id` > `SURFACE_DEFAULTS[key]`). Update the `getModelSettings` response type in `voc-datalake/frontend/src/api/client.ts` and make `AiModelSection.tsx` render `labelForId.get(surface.effective_id) ?? surface.effective_id` directly, deleting `pinnedId`/`effectiveId`. Add pytest cases in `voc-datalake/lambda/api/test/test_settings_handler.py` covering: no pin (effective_id == default_id), a pin (effective_id == pin for every surface), a pin plus a per-surface override (effective_id == the override), and a non-allowlisted stored `model_id` (ignored). Keep the existing vitest cases in `AiModelSection.test.tsx` green by updating the fixture to include `effective_id`.

voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
```diff
@@ -163,7 +184,7 @@ function SurfaceRow({ surfaceKey, selected, defaultLabel, models, justSaved, onS
           value={selected ?? ''}
           onChange={(event) => onSelect(event.target.value === '' ? null : event.target.value)}
         >
-          <option value="">{t('aiModel.automaticOption', { model: defaultLabel })}</option>
+          <option value="">{t('aiModel.automaticOption', { model: automaticLabel })}</option>
```
perrozzi, line 187: **Type**: issue
**Severity**: minor
**Title**: Under a pin the Automatic option is visually indistinguishable from the explicit option for the same model

**Description**: With a pin deployed, every surface renders `Automatic — Claude Sonnet 4.6` as the first option while `Claude Sonnet 4.6` also appears as an explicit option in the same `<select>` (it is in `available_models`). The two are one word apart but semantically different: `Automatic` writes nothing and keeps following the deployment pin, while the explicit option writes `settings.surfaces[<surface>]` and *stops* following the pin — so if the deployment is later re-pinned, or the pin is cleared, the surface stays behind on a model the admin never meant to freeze. That is a milder version of the same footgun issue #275 describes (an admin "correcting" the label into an explicit selection).

A second, related effect: the surface's tuned built-in default (`default_id`) is now nowhere in the UI when a pin is present. Previously that value was at least visible; now `Prototype Builder` and `Feedback Enrichment` give no hint that their untuned defaults are Opus 5 and Haiku 4.5, so an admin cannot tell whether `Automatic — Claude Sonnet 4.6` reflects a per-surface default or a deployment-wide pin overriding five different ones. The PR body acknowledges deferring the issue's `(pinned for this deployment)` suffix as a copy decision, which is reasonable — this comment is to record that both effects trace to that deferral, and that `docs/deployment.md` already has the vocabulary ("It is a floor, not a lock") if a maintainer wants to land the copy.

I would not block on this: the label now names the right model, which is the acceptance criterion.

**Proposed fix**: Add one new `settings` locale key used only when `pinnedId` is non-null — e.g. `aiModel.automaticPinnedOption` rendering `Automatic — {{model}} (pinned for this deployment)` — and select between it and the existing `aiModel.automaticOption` based on a new boolean `SurfaceRow` prop. Because the parenthetical is real user-facing prose it must be translated in all eight `public/locales/*/settings.json` files (the model name itself stays interpolated, so the existing i18n gate's same-as-English rejection is not triggered). Optionally also render the surface's own `default_id` label as muted helper text next to the select when a pin is masking it.

**AI prompt**: In `voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx`, distinguish the Automatic option when a deployment-wide pin is active. Pass a new `readonly isPinned: boolean` prop to `SurfaceRow` (true when `pinnedId !== null`) and have it render `t('aiModel.automaticPinnedOption', { model: automaticLabel })` instead of `t('aiModel.automaticOption', { model: automaticLabel })` when true. Add the `automaticPinnedOption` key under `aiModel` to all eight locale files in `voc-datalake/frontend/public/locales/*/settings.json`, keeping each file's existing key ordering, with a properly translated equivalent of `Automatic — {{model}} (pinned for this deployment)` in de/es/fr/ja/ko/pt/zh (do not copy the English string — the repo's `npm run i18n:check` gate rejects same-as-English values). Add a vitest case in `AiModelSection.test.tsx` asserting the pinned wording appears for every surface under `model_id: SONNET46`, and that the unpinned wording is still used when `model_id` is null. Run `npm run lint:frontend`, `npm run typecheck` and `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` from `voc-datalake/frontend`.

voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
```diff
@@ -113,34 +116,52 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface,
...
+  // `-c defaultModelId=<id>`) outranks every surface's built-in default in
+  // shared/model_config.py: per-surface override > global pin > SURFACE_DEFAULTS.
+  // Labelling Automatic from `surface.default_id` alone — which is what this did
+  // before #275 — made a pinned deployment advertise a model it never invokes,
+  // and invited an admin to "correct" it by explicitly selecting the advertised
+  // model, which in a workshop account cannot be invoked at all. Truthiness, not
+  // `?? null`: the field can arrive absent or as '' from an older payload, and
+  // neither is a pin. Tests: 'names the deployment-wide pin in every Automatic
+  // option' and 'ignores an empty-string pin and keeps the surface default'.
+  const pinnedId = data.model_id ? data.model_id : null
```
perrozzi, line 129: **Type**: issue
**Severity**: minor
**Title**: `pinnedId` and its ten-line comment can both shrink; the referenced test titles will rot

**Description**: Two small things on this block.

1. `data.model_id ? data.model_id : null` followed by `pinnedId ?? surface.default_id` is a two-step spelling of one expression. `data.model_id || surface.default_id` has identical behaviour for every input the tests exercise (`null`, absent, `''`, a real id) and removes the intermediate binding. If keeping `pinnedId` as a named concept is preferred for the comment to hang off, `const pinnedId = data.model_id || null` is still one operator shorter than the ternary.

2. The comment names two vitest titles verbatim (`'names the deployment-wide pin in every Automatic option'`, `'ignores an empty-string pin and keeps the surface default'`). Those strings will silently go stale on the first test rename, and the REVERT MAP added at the top of `AiModelSection.test.tsx` already carries exactly this mapping — so it is duplicated in the place least likely to be updated.

For the record, the `🔑` marker itself is fine and I would keep it: it is an established convention in this codebase for load-bearing "why" comments (`src/api/types.ts`, `src/api/projectsApi.ts`, `src/components/ProtectedRoute/ProtectedRoute.tsx`, `src/pages/ProjectDetail/usePrototypeBuild.ts`). (Informed by repository memory: comments in this repo deliberately carry the argument rather than the mechanics, so the *content* here is on-convention — only the test-title duplication is a maintenance liability.)

**Proposed fix**: Collapse to `const pinnedId = data.model_id || null` (or inline it entirely as `const effectiveId = data.model_id || surface.default_id`), and trim the comment to the argument — what the precedence is, that truthiness is deliberate because `null`/absent/`''` are all non-pins — dropping the two quoted test titles in favour of a pointer to the REVERT MAP in the sibling test file.

**AI prompt**: In `voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx`, replace `const pinnedId = data.model_id ? data.model_id : null` with `const pinnedId = data.model_id || null` and shorten the preceding comment: keep the precedence statement (per-surface override > global pin > SURFACE_DEFAULTS, per `shared/model_config.py`), keep the note that truthiness is deliberate because `null`, an absent field and `''` are all non-pins, drop the two verbatim quoted test titles, and instead point at the REVERT MAP in `AiModelSection.test.tsx`. Keep the `🔑` prefix — it is the repo's marker for load-bearing comments. Verify with `npm run lint:frontend`, `npm run typecheck` and `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` from `voc-datalake/frontend`.

voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
```diff
@@ -113,34 +116,52 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface,
...
+      {data.surfaces.map((surface) => {
+        // What Automatic actually resolves to for this surface. The pin does not
+        // touch `selected`: a stored per-surface choice still outranks it.
+        const effectiveId = pinnedId ?? surface.default_id
+        return (
+          <SurfaceRow
+            key={surface.key}
+            surfaceKey={surface.key}
+            selected={surface.selected}
+            automaticLabel={labelForId.get(effectiveId) ?? effectiveId}
```
perrozzi, line 143: **Type**: issue
**Severity**: minor
**Title**: The declared response type says `model_id: string | null`, but the component and tests rely on tolerating `undefined` and `''`

**Description**: `api.getModelSettings` declares `model_id: string | null` as a **required, non-optional** field (`src/api/client.ts`), and `ModelSettings` is derived from that return type. Two of the new tests deliberately feed shapes the type says cannot occur: `keeps each surface default in its Automatic option when the pin field is absent` deletes the key, and `ignores an empty-string pin and keeps the surface default` passes `''`. Both pass only because `mockGetModelSettings` is mocked at the `api/client` boundary, which bypasses the declared type and any validation.

That is not a runtime bug — the truthiness guard makes all four shapes safe, and `get_model_settings` in `settings_handler.py` provably emits `null` or an allowlisted id (`legacy_global if legacy_global in ALLOWED_MODEL_IDS else None`), so neither `undefined` nor `''` is reachable from the current server. But it leaves the tolerance asserted in tests and implemented in the component while the type contract denies it, so a future reader cannot tell whether the absent-field case is defensive depth or a real wire shape. Worth noting this endpoint also has no runtime validator, unlike the newer boundaries in this repo (`feedbackSchema.ts`, `prioritizationRowsApi.ts`, `wireRecord.ts`, `jobGroundingSchema.ts` all parse with Zod), and `model-override.ts` uses Zod on the same DynamoDB item server-side.

**Proposed fix**: Make the tolerance visible where the shape is declared: widen to `model_id?: string | null` in the `getModelSettings` return type in `src/api/client.ts` with a one-line comment saying an older/absent payload is treated as "no pin", or — if the field is genuinely guaranteed — drop the absent-field test and keep only the `null`/`''` cases so the tests stop asserting an unreachable shape.

**AI prompt**: In `voc-datalake/frontend/src/api/client.ts`, change the `getModelSettings` response type field `model_id: string | null` to `model_id?: string | null` and add a one-line comment explaining that an absent field (older payload) and `''` are both treated as "no deployment-wide pin", which is why `AiModelSection.tsx` guards on truthiness rather than `?? null`. Confirm `voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx` still typechecks with the now-optional field, then run `npm run typecheck`, `npm run lint:frontend` and `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` from `voc-datalake/frontend`.

voc-datalake/frontend/src/pages/Settings/AiModelSection.test.tsx
```diff
@@ -96,6 +109,70 @@ describe('AiModelSection', () => {
...
+    render(<AiModelSection apiEndpoint="https://api.example.com" isAdmin />, { wrapper: createWrapper() })
+
+    await waitFor(() => expect(screen.getByLabelText('AI Chat')).toBeInTheDocument())
+    for (const label of SURFACE_LABELS) {
+      const select = screen.getByLabelText(label)
+      expect(within(select).getByText('Automatic — Claude Sonnet 4.6')).toBeInTheDocument()
+      expect(select).toHaveValue('')
+    }
+    // No surface keeps advertising its built-in default under a pin.
+    expect(screen.queryByText('Automatic — Claude Opus 5')).not.toBeInTheDocument()
```
perrozzi, line 137: **Type**: issue
**Severity**: minor
**Title**: The strongest negative control in the new suite is document-global and would not localise a per-surface regression

**Description**: `screen.queryByText('Automatic — Claude Opus 5')` searches the whole document, so this assertion only proves that *no* surface still advertises Opus 5. A regression where, say, `prototype` correctly picks up the pin but `enrichment` silently falls back to its default would still satisfy it, because Haiku 4.5 is not the string being queried. The `getByText` inside the loop above does carry per-surface coverage, so the suite as a whole is sound — this is about the control being weaker than it reads.

Related, cheap gap: there is no case where the pin **equals** a surface's own `default_id`. That is the no-op branch (`pinnedId ?? surface.default_id` returning the same value either way) and the shape most likely to mask a regression, since a broken implementation still renders the right label for that one surface.

**Proposed fix**: Inside the existing `for (const label of SURFACE_LABELS)` loop, add a scoped negative assertion — `expect(within(select).queryByText(/^Automatic — (?!Claude Sonnet 4\.6)/)).not.toBeInTheDocument()`, or more simply assert `within(select).queryByText('Automatic — Claude Opus 5')` and `within(select).queryByText('Automatic — Claude Haiku 4.5')` are both absent — so a single drifting surface names itself in the failure. Then add a short case pinning `model_id: OPUS5` and asserting `Prototype Builder` (whose `default_id` is already Opus 5) renders exactly one `Automatic — Claude Opus 5` option while `AI Chat` also shows `Automatic — Claude Opus 5` rather than Sonnet 5.

**AI prompt**: In `voc-datalake/frontend/src/pages/Settings/AiModelSection.test.tsx`, strengthen the test `names the deployment-wide pin in every Automatic option`: move the document-global `expect(screen.queryByText('Automatic — Claude Opus 5')).not.toBeInTheDocument()` assertion into the existing `for (const label of SURFACE_LABELS)` loop and scope it with `within(select)`, also asserting `Automatic — Claude Haiku 4.5` and `Automatic — Claude Sonnet 5` are absent from that surface, so a failure names the offending surface. Then add a new test `labels Automatic with the pin even when it equals the surface default`, mocking `{ ...modelSettingsFixture, model_id: OPUS5 }` and asserting `Prototype Builder` renders exactly one option with text `Automatic — Claude Opus 5` (use `within(select).getAllByText` and check length 1) and that `AI Chat` also renders `Automatic — Claude Opus 5`. Run `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` from `voc-datalake/frontend` and confirm all cases pass.

voc-datalake/frontend/src/pages/Settings/AiModelSection.tsx
```diff
@@ -113,34 +116,52 @@ function AiModelSectionBody({ data, isLoading, isError, isSaving, savedSurface,
...
-          selected={surface.selected}
-          defaultLabel={labelForId.get(surface.default_id) ?? surface.default_id}
-          models={data.available_models}
-          justSaved={savedSurface === surface.key}
-          onSelect={(modelId) => onSelect(surface.key, modelId)}
-        />
-      ))}
+      {data.surfaces.map((surface) => {
+        // What Automatic actually resolves to for this surface. The pin does not
+        // touch `selected`: a stored per-surface choice still outranks it.
```
perrozzi, line 136: **Type**: question
**Severity**: minor
**Title**: Now that the deployment pin is visible in this card, should an admin be able to clear it here?

**Description**: `PUT /settings/model` already accepts a bodiless-surface form — `{'model_id': <allowlisted id>|null}` with no `surface` — which sets or clears the global pin (`_save_global_model`), and `docs/deployment.md` describes the pin as "a floor, not a lock" because the `Custom::ModelPin` custom resource is write-once. But this card only ever sends the per-surface form, so there is no UI path to clear the pin.

Before this PR that asymmetry was invisible. After it, an admin sees the pin's effect on all five surfaces with no way to remove it — the only escape is pinning each of the five surfaces explicitly, which is exactly the "explicit selection that stops following the pin" behaviour the issue warns about. Was leaving the pin read-only a deliberate call (i.e. deployment-owned config that Settings should not touch), or is a "clear deployment pin" control intended as a follow-up? Either answer is fine — I would just like it recorded, since the card now surfaces state it cannot edit.

perrozzi (commented):
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

## Summary

Reviewed `fda9144` — the head of the branch, 2 files, +115/−17, both under `voc-datalake/frontend/src/pages/Settings/`.

**What the PR does, verified against the code it claims to follow.** `GET /settings/model` (`voc-datalake/lambda/api/settings_handler.py::get_model_settings`) returns each surface's `default_id` *and* `model_id`, filtered to `ALLOWED_MODEL_IDS`. `shared/model_config.py::get_active_model_id` resolves *per-surface override > legacy global `settings.model_id` > `SURFACE_DEFAULTS` > `BEDROCK_MODEL_ID`*, and `lambda/stream/src/bedrock/model-override.ts::resolveModelOverride` mirrors the top two tiers for streaming chat. `core-stack.ts` seeds `settings.model_id` from `-c defaultModelId=`, so the bug in #275 is real on the documented workshop configuration: the old `labelForId.get(surface.default_id)` named a model no surface would invoke. The fix — `effectiveId = pinnedId ?? surface.default_id` with `pinnedId` guarded on truthiness — matches the resolver for every state the endpoint can produce, and leaving `selected` untouched is correct, since a stored per-surface value outranks the pin. Scope discipline is good: no new locale key, no prop plumbing beyond the `defaultLabel` → `automaticLabel` rename, no backend change, and the availability/invocability half of #275 correctly deferred (it needs a Bedrock permission on the settings Lambda and a response-shape change).

One thing genuinely worth calling out: the test suite distinguishes **negative controls from regression coverage** and records which mutation each assertion catches in a REVERT MAP, including the honest note that two of the five new cases pass against the unfixed code. That is the difference between a suite that proves something and one that merely runs, and it is rare to see it written down.

**Findings: 5 issues (all minor), 1 question, 0 comments.** Nothing here blocks merge — the acceptance criterion (the label names the model actually invoked) is met and proven. The findings are drift-prevention and clarity:

- The resolver's precedence now exists in three independent implementations (Python, the stream TS mirror, and this component) with nothing pinning them together — notable because this repo *already* gates the model **allowlist** across those same three places with lockstep tests, and the precedence order now has the same duplication shape without that protection. Highest-value follow-up: return `effective_id` per surface from the API.
- Under a pin, the `Automatic — X` option and the explicit `X` option are one word apart but write different state, and the surface's tuned `default_id` disappears from the UI entirely. Both trace to deferring the issue's `(pinned for this deployment)` suffix, which is a defensible call — recorded rather than demanded.
- `model_id` is declared required (`string | null`) while two new tests exercise `undefined` and `''`; the mock at the `api/client` boundary bypasses the type, so the tolerance is asserted in tests but denied by the contract. Not reachable from the current server, but this endpoint is also the odd one out in having no Zod validator, unlike the newer boundaries in `src/api/`.
- The suite's strongest negative control is document-global and would not localise a single-surface regression; the pin-equals-default no-op branch is untested.
- `pinnedId`'s ternary collapses to `||`, and its comment quotes two test titles that the REVERT MAP already maps. (The `🔑` marker is on-convention here — `src/api/types.ts`, `projectsApi.ts`, `ProtectedRoute.tsx` all use it — and I would keep it.)

## Build and test results

`mise run build` → **`no tasks defined in /workspace/...`**, confirming the harness's build gate is inert on this repo; the PR body's claim is accurate. I ran the real gates instead. `node_modules` was absent, so `npm ci` was needed in `voc-datalake/frontend/` **and** `voc-datalake/` (the frontend `prebuild` script runs `ts-node scripts/generate-manifests.ts` from the CDK package — without it `npm run build` dies at `sh: 1: ts-node: not found`, exit 127, which is a container artefact, not a PR defect).

| Command | Result |
|---|---|
| `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` | **PASS** — 13/13 |
| `npm run test` (full frontend vitest) | **PASS** — 192 files, 3285 tests, exit 0 |
| `npm run typecheck` | **PASS** |
| `npm run lint:frontend` (`--max-warnings 0`) | **PASS** |
| `npm run build` (`tsc -b && vite build`) | **PASS** — built in 9.29s |
| `npm run typecheck:tests` | 87 `error TS` — **pre-existing baseline**; zero mention `AiModelSection` |
| `npm run i18n:check` | exit 1 — **pre-existing baseline**; zero findings mention `AiModelSection` |

Every number in the PR body reproduced, including the two red baselines and their independence from the change. Backend suites were not run — this PR touches no Python or CDK code, and I confirmed by reading `settings_handler.py`, `model_config.py`, `model-override.ts` and `core-stack.ts` that the contract the component now relies on is unchanged.

## Not blocking, but worth a follow-up issue

Two observations that belong outside this PR and are already flagged in its body — recorded here so they are not lost. (1) The uninvocable-model half of #275 (suggestion 2 in the issue) needs its own issue: it is an IAM/backend change, not a label change. Note #273 is already open on adjacent ground (`BEDROCK_MODEL_ID` used directly at two call sites, bypassing resolution) — the same root cause family. (2) `typecheck:tests` at 87 errors and a red `i18n:check` on `development` mean neither can gate a PR today, and `npm run check` chains with `&&`, so on this tree it can never reach its later legs; a "no new findings vs. baseline" check would make them useful again.

perrozzi:
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

## Review summary

Posted [review #5087968174](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/402#pullrequestreview-5087968174) on `fda9144`.

**Findings: 6 total — 5 issues (all `minor`), 1 question, 0 comments.** No blockers.

| # | Type | Severity | Finding |
|---|---|---|---|
| 1 | issue | minor | Resolver precedence now duplicated in a third place with no lockstep pinning it |
| 2 | issue | minor | Under a pin, `Automatic — X` and explicit `X` are visually near-identical but write different state; `default_id` vanishes from the UI |
| 3 | issue | minor | `model_id` declared required while two new tests exercise `undefined` / `''` |
| 4 | issue | minor | Strongest negative control is document-global; pin-equals-default no-op branch untested |
| 5 | issue | minor | `pinnedId` ternary collapses to `\|\|`; comment quotes test titles the REVERT MAP already maps |
| 6 | question | — | Should an admin be able to clear the deployment pin from this card, now that its effect is visible? |

### Overall assessment

Good quality, correctly scoped, and the fix is right. I verified the claim rather than taking it on trust: `settings_handler.py::get_model_settings` does return both `default_id` and an allowlist-filtered `model_id`; `model_config.py::get_active_model_id` does rank `settings.model_id` above `SURFACE_DEFAULTS`; `core-stack.ts` does seed that attribute from `-c defaultModelId=`. So the bug in #275 is real on the documented workshop path, and `effectiveId = pinnedId ?? surface.default_id` matches the resolver for every state the endpoint can emit. Leaving `selected` untouched is correct — a stored per-surface value still outranks the pin. Deferring the availability/invocability half of #275 was the right call: it needs a Bedrock permission on the settings Lambda and a response-shape change.

One thing worth naming: the tests separate **negative controls from regression coverage** and record in a REVERT MAP which mutation each assertion catches — including the honest note that two of the five new cases pass against the *unfixed* code. That distinction is what makes the suite proof rather than decoration, and it is rare to see it written down.

### Areas that need attention before merge

Nothing must change to merge. In priority order if the author wants to act:

1. **Precedence duplication (finding 1)** — the highest-value item. The ordering now lives in `model_config.py`, `model-override.ts` and this component. This repo already gates the model **allowlist** across those same three files with Python lockstep tests that read the TS mirror and the CDK helper; the precedence order has the same duplication shape with none of that protection. Returning `effective_id` per surface from the API would delete the frontend copy entirely.
2. **Findings 4 and 5** are one-line each and improve the suite that is already this PR's strongest asset.
3. **Findings 2, 3 and 6** are judgement calls better answered than changed.

### Build / test results

`mise run build` → **`no tasks defined`** — the harness build gate is inert on this repo, exactly as the PR body states. I ran the real gates. `npm ci` was needed in both `voc-datalake/frontend/` and `voc-datalake/` (the frontend `prebuild` runs `ts-node` from the CDK package; without it `npm run build` exits 127 — a container artefact, not a PR defect).

- `npx vitest --run src/pages/Settings/AiModelSection.test.tsx` — **PASS**, 13/13
- `npm run test` (full frontend) — **PASS**, 192 files / 3285 tests, exit 0
- `npm run typecheck` — **PASS**
- `npm run lint:frontend` (`--max-warnings 0`) — **PASS**
- `npm run build` (`tsc -b && vite build`) — **PASS**, 9.29s
- `npm run typecheck:tests` — 87 `error TS`, **pre-existing baseline**, none naming `AiModelSection`
- `npm run i18n:check` — exit 1, **pre-existing baseline**, no finding naming `AiModelSection`

Every figure in the PR body reproduced, including both red baselines and their independence from this change. Backend suites were not run: the PR touches no Python or CDK code, and I confirmed by reading `settings_handler.py`, `model_config.py`, `model-override.ts` and `core-stack.ts` that the contract the component relies on is unchanged.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GP3QZ3T2654N9DDQRF3KRC` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T09:36:42.613Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/402) |
| Duration | 1563.6s |
| Cost | $2.4195 |

perrozzi:

**Why it stopped:** 0 blocking, 0 should_fix, build not regressed — decision=commented; 5 nit, 1 needs_author_input; build unknown (tolerated); mergeable [iteration 0, 5 nit, 1 needs_author_input]

| | |
|---|---|
| Outcome | **CONVERGED** |
| Iteration | 0 of 5 |
| Loop spend | $3.55 of $100.00 |
| Goal | Implement GitHub issue #275 on a new branch from development. Settings returns each AI surface's built-in default_id and the deployment-wide model_id override. Runtime resolution gives the global override higher precedence, but AiModelSectionBody currently labels every Automatic option from s… |
| Run | [`01M1GN30RPDS3DH0JYJ777NFSN`](https://d1t2aabk0oo81s.cloudfront.net/runs/01M1GN30RPDS3DH0JYJ777NFSN) |

### Each iteration

|---|---|---|---|---|---|---|
| 0 | 5 nit, 1 needs_author_input | — | ❌ | ✅ | — | — |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GR4G272797WATCWWCC1RC1` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T09:47:26.582Z |
| Duration | 106.6s |
| Cost | $0.3745 |

Issue #399, opened 2026-09-01
Follow-ups from #394: copy assertions a reword walks past, and resolveDerivation's per-call index


perrozzi:
Follow-ups from the human review of #394 (lineage-consistency and stale-frozen-row signals, #339 Phase 3). That PR was **approved and merged**: the classifier is correct, the temporal core reads `created_at` through an explicit allow-list grammar rather than `Date.parse`, the failure direction is toward silence, all eight locale catalogues carry all ten new keys with one placeholder in one key, and eighteen mutants were each caught by a named case. Nothing below is user-facing today. Three of the four items are test-sensitivity gaps of the same class the PR spent its last four commits closing for a different key, and the fourth is a pre-existing performance characteristic that #394 is the first caller to expose.

## A. Copy assertions that a reword walks past

- [ ] **`staleReason`'s scope qualifier is pinned in English only.** Restoring the pre-fix unqualified wording in any of `de`, `es`, `fr`, `ja`, `ko`, `pt`, `zh` leaves the suite green — confirmed, 7 of 8 locales. `coherentReason` already has the eight-locale positive-plus-negative pair that closes this; give `staleReason` the same. This is exactly the gap the PR closed for `coherentReason` in `2e48394`, one key over.

- [ ] **The `DIRECTIONAL` table catches only the exact pre-fix term.** A fluent directional reword that keeps the `GENERATION` control word goes green in **all eight** locales, not just the three (`zh: '较早'`, `ja: '旧バージョン'`, `ko: '이전 버전'`) named in review. Either widen each entry to an alternation of that language's directional forms, or narrow the test name and the file docstring's claim that it "catches the copy regressing" to what the table actually detects. A copy assertion keyed to one pre-fix string detects a `git revert`, not a regression.

- [ ] **Document the `staleReason` / `supersededSourceReason` directional asymmetry beside `DIRECTIONAL`.** The asymmetry is *correct* — the staleness rule compares instants while the crossing rule compares types only — but nothing next to the table says so, which is why it keeps being read as an oversight. One sentence.

## B. `resolveDerivation` rebuilds its project index on every call

[...]

PR #394, opened 2026-08-27
feat(prioritization): show lineage consistency and stale frozen rows

perrozzi:
Completes the two remaining Phase 3 criteria of #339: make a prioritization row's
evidence quality visible, and mark a frozen row stale when a genuinely fresher
coherent combination exists.

Frontend only. No backend route, ballot key, row persistence, API-client type,
scoring arithmetic, room-voting or feedback-form-evidence change. `api/derivation.ts`
is **read-only reuse** — not one line of it is touched — which keeps this disjoint
from #381's `api/client.ts` ownership.

## Summary

**`pages/Prioritization/rowLineage.ts` (new, pure, no React).** Small helpers over
`resolveDerivation`, which already normalises the declared `derivation` map and the
three legacy shapes (`source_prd_id`, `source_prfaq_id`, `source_documents`) to one
answer — so this module never asks which shape a lineage came from.

- `classifySelectionLineage` → `coherent` | `crossGeneration` | `absent`, each with
  its own `reason` code, from four rules in a load-bearing order: two documents of
  one **type** (decided with no derivation read at all, so no record can talk a row
  out of it) → a document built from another **generation** of a type the row also
  holds → nothing in the row records anything (`origin === 'none'`) → coherent.
- **Coherent claims "nothing contradicts one chain, and something can speak"**, not
  "a chain provably links them". Requiring a proven edge was tried and rejected: a
  PRD and a PR/FAQ generated from one sitting name feedback and personas, never each
  other, so the rule greyed the ordinary row — and a signal grey for everybody is not
  a signal.
- The rule is **role-blind** over the whole closed vocabulary, so a fifth
  `DERIVATION_ROLE` cannot leave it quietly stale.
- `fresherCoherentSelection` — the staleness arithmetic, under the same
  document-role expectations: the project's newest document of **each type the row
  holds, and nothing else**. Four conditions can each withhold staleness, because the
  sentence it drives asks a reviewer to create a row: one readable type per document;
  every type still resolving; **strictly** newer (nothing older, so a sideways move is
  not an improvement); and the candidate **itself coherent** — which is what makes the
  criterion "a real fresher coherent combination" rather than "something newer exists".
- `rowLineageOf` asks staleness of a **frozen** row only. An un-frozen row's answer is
  "edit this one", which is already on screen.

**Structured state, never colour.** `LineageState`, `LineageReason`, `stale` and
`fresherDocumentIds` are on `PrioritizationRowView`, resolved once in `collectRows`
[...]

PR #404, opened 2026-09-02
perf(derivation): build one source index per project collection pass (#399)

perrozzi:
Closes section **B** of #399 (`resolveDerivation` rebuilds its project index on every call). Sections A (locale copy) and C (data shapes) are deliberately untouched — the issue asks for them to be decided separately, and nothing in this diff touches a catalogue or a read-side dedupe.

## Summary

`resolveDerivation` built a `document_id → {title, document_type}` index over the whole project read on **every call**. That cost nothing while its only caller (`DocumentsTab`) held one document at a time. #394 made the prioritization page the first caller to invoke it in a loop: `collectRows` calls `rowLineageOf` per row, which calls `classifySelectionLineage` and `fresherCoherentSelection`, each of which calls the resolver per document on that row — so one unchanging project read of D documents was walked once per (row × document × rule).

The index is now built once and passed:

* **`api/derivation.ts`** — the private `sourceFieldIndex` becomes the exported `derivationSourceIndex`, and `resolveDerivationAgainst(document, index)` is the resolver against an index the caller already holds. `resolveDerivation` is now literally `resolveDerivationAgainst(document, derivationSourceIndex(projectDocuments))`, so there is no second resolution rule that can drift and its public behaviour is unchanged for every existing caller.
* **`pages/Prioritization/rowLineage.ts`** — `projectLineageSources(documents)` bundles the documents with one index over exactly those documents. All three exported rules (`classifySelectionLineage`, `fresherCoherentSelection`, `rowLineageOf`) take **either** a plain document list **or** a prepared read, so hoisting is the caller's optimisation and no existing call site had to change. The nested `classifySelectionLineage` call inside `fresherCoherentSelection` passes the prepared read through, so a frozen row's candidate check does not re-index either.
* **`pages/Prioritization/prioritizationUtils.ts`** — `collectRows` prepares each project read once, in the details loop, and passes it for every row of that project. `byId` (which resolves a row's stored ids and its prototype) was rebuilt per row over the same list and is hoisted alongside it.

**Explicit plumbing, not global memoization**, as the task asked: nothing is cached anywhere. The index's lifetime is the collection pass's own, so a landing project read, a deleted document or a recomposed row needs no invalidation — the next `collectRows` builds a new index from whatever the reads then say. That property is asserted, not just claimed (`consults the index once per source and holds no state between calls`).

`DocumentsTab` keeps calling `resolveDerivation` with a list: it holds one document, so a prebuilt index would be plumbing with nothing on the other end.

## Measurement (issue #399's acceptance criterion for B)

Median of five `collectRows` calls after one warm-up, over a synthetic single-project fixture of N frozen rows and D documents each declaring a source, run under the repo's own vitest/jsdom in this container. Wall clock here is ~3× the reviewer's figure for the same fixture (their 562 ms is this container's 1644 ms), so read the ratio rather than the absolute:

| fixture | before | after | |
|---|---|---|---|
| 30 rows / 100 documents | 27.3 ms | **11.9 ms** | 2.3× |
| 60 rows / 400 documents | 203.5 ms | **78.5 ms** | 2.6× |
| **200 rows / 1000 documents** | **1644.3 ms** | **615.3 ms** | **2.7×** |

**What was removed:** at 200/1000, the same 1000-document list was indexed roughly 2,600 times (200 rows × two derivation-reading rules × two documents per row, plus one candidate classification per frozen row) instead of once — every one of those passes producing a byte-identical map. The remaining 615 ms is *not* index construction: it is `fresherCoherentSelection`'s `newestOfType`, which filters the whole project list once per type per frozen row and is O(rows × documents) independent of any index. Measured by flipping the fixture's rows to un-frozen, which skips the staleness arithmetic entirely: 7.0 ms at 200/1000. That residual is a separate concern from the one #399 B names and is left for a follow-up rather than folded in here.

The measurement harness was scratch (`.tmp/`) and is not part of the diff, deliberately: a wall-clock assertion is exactly what the acceptance criteria ask *not* to gate on.

## Tests

Coverage is counted, never timed — the acceptance criterion is explicit about that, and a threshold would pass on a fast machine with the defect restored.

**`src/api/derivation.test.ts`** (+4 cases, semantic equivalence across all three origins):
* `answers exactly what resolveDerivation answers` — whole-object equality for a **declared**, a **legacy**, a legacy-with-**unresolved-source**, and a **nothing-recoverable** document, plus a control asserting the four really do produce `declared`/`legacy`/`legacy`/`none` so the equality is over answers that differ.
* `reads what a source is from the given index and never from the wire again` — an index disagreeing with every document list, so an answer carrying its titles could not have come from re-indexing anything.
* `consults the index once per source and holds no state between calls` — a `Map` subclass recording every `get`, asserting `['prd_1','prfaq_1','prd_1','prfaq_1']` over two calls. Its positive control (the titles actually resolved) is asserted first, so a count is never the silence of a resolver that looks nothing up.
* `indexes a document list once for every field a source displays` — the builder's own contract, including that an id-less record contributes no entry.

**`src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts`** (new file, the deterministic construction count). `derivationSourceIndex` is wrapped via `vi.mock` + `importOriginal` (the `ScraperCard.test.tsx` precedent) so the *real* builder runs and only its call count is observed:
* `indexes a project read once however many rows name it` — three rows in three different lineage states off **one** construction. The state assertion comes first as the positive control: a `collectRows` that stopped classifying lineage would also report one construction.
* `indexes each project read separately` — two projects, **two** constructions, so "once" cannot be satisfied by one page-wide index that would resolve one project's sources against another's documents. Controlled by a source that only project 1 holds staying unresolved for project 2's row.
* `builds nothing for a page whose project reads have not landed` — the floor.

**Mutation-tested in both directions** (each restored afterwards; the working tree is clean):

| mutation | result |
|---|---|
| `collectRows` passes `project.documents` instead of `project.lineage` | **caught** — 4 constructions, expected 1 |
| `fresherCoherentSelection`'s nested classify passes the raw list | **caught** — 3 constructions, expected 1 |
| one page-wide index shared across projects | **caught** — `indexes each project read separately` only, as its docstring claims |
| a memo cache added behind `resolveDerivationAgainst` | **caught** — 3 cases, including the lookup count |
| the prebuilt-index path drops the legacy fallback | **caught** — 8 cases including the equivalence case |
| `resolveDerivation` one-sidedly diverging (`projectDocuments.slice(1)`) | **caught** — 7 cases |

Each revert is recorded in the two files' REVERT MAP docstrings, per repo convention.

## Build and gate results

Run from `voc-datalake/frontend` unless noted.

| gate | result |
|---|---|
| `npx vitest run src/api/derivation.test.ts src/pages/Prioritization src/pages/ProjectDetail` | ✅ 58 files, **1054 passed** |
| `npm run test` (full frontend suite, repo root) | ✅ 193 files, **3287 passed** |
| `npm run typecheck` (`tsc -b --noEmit`) | ✅ clean |
| `npm run lint:frontend` (`eslint . --max-warnings 0`) | ✅ clean |
| `npm run typecheck:cdk` | ✅ clean |
| `npm run typecheck:tests` | ⚠️ **87 errors, all pre-existing** (baseline on a stashed tree: 88 — one of the 88 is this branch's own new test file failing to resolve the not-yet-added export, so the count goes *down* by one). No error in any of the four files this diff touches. |
| `npm run build` / `npx vite build` | ❌ **fails identically on the unmodified tree** — `Rollup failed to resolve import "react-markdown"`. This container blocks npm install scripts (`esbuild` postinstall), so `react-markdown` is installed without its `dist`. Verified by stashing the whole diff and re-running: same failure, exit 1. Not caused by this change. (`mise run build` reports `no tasks defined` — the repo defines no mise tasks.) |
| `node scripts/i18n-check.mjs` | ⚠️ exits 1 both with and without this diff (pre-existing hardcoded-string backlog in DataExplorer, FeedbackDetail, ChatSidebar and others). No catalogue is touched here. |

## Decisions made

* **Both shapes accepted by the exported rules** (`readonly unknown[] | ProjectLineageSources`) rather than a breaking signature change. A caller with one selection pays for one index exactly as before; only the loop prepares a read. This kept all 93 existing `rowLineage.test.ts` call sites and `RowLineagePanels`' eventual one untouched.
* **The index travels with the documents** in one `ProjectLineageSources` object, not as a second parameter. A separate prebuilt-index argument can be handed an index built from a *different* read, which would classify a row against one project and advise it against another; one object built by `projectLineageSources` cannot.
* **`lineageSourcesOf` narrows on `'sourceIndex' in project`, not `Array.isArray`** — `tsc` will not narrow a `readonly unknown[]` out of a union via `Array.isArray` (its signature answers `any[]`), which leaves the list in the prepared branch's type. Noted at the function.
* **`DerivationSourceIndex` is a `ReadonlyMap` of already-normalised fields**, not an opaque handle. Every value has been through `displayString`, so an unreadable type is `''` and never `null` — `hasSupersededSource` turns on exactly that distinction, and the type's docstring says to build it only through `derivationSourceIndex`.
* **The remaining `newestOfType` cost is called out, not fixed.** Folding an O(1) per-type newest-document lookup into the same change would have mixed two optimisations under one issue item; it is documented above with its measurement instead.

## Notes for reviewers

The `1644 ms`/`615 ms` figures now appear in three docstrings (`DerivationSourceIndex`, the `rowLineage` module header, `PrioritizationRowView.lineage`) alongside the reviewer's 562 ms, labelled as this container's jsdom rather than as the reviewing machine's. If the house convention is to quote only the reviewed figure, they are one search-and-replace apart.

```abca-verify
no-ui
```

## Agent notes

**What went well.** The defect had one choke point. Making `resolveDerivation` delegate to `resolveDerivationAgainst` meant the equivalence claim needed no argument — there is one resolution rule, called two ways — which left the tests free to pin the *counting* property instead of re-testing the contract. The accept-either-shape union kept the diff to four files with no call-site churn.

**What was difficult.**
1. **The build cannot be run here.** `npx vite build` fails on the unmodified tree because this container blocks npm postinstall scripts, so `react-markdown` unpacks without its `dist`. I confirmed the baseline failure by stashing before concluding it was pre-existing; `mise run build` is also a no-op (`no tasks defined`) — the setup notes' warning that build gating is inert is accurate.
2. **Getting a trustworthy measurement.** vitest has no benchmark harness wired up here and the 10 s `testTimeout` kills a 200/1000 fixture at the pre-fix cost, so the scratch harness needed `--testTimeout`. `console.log` inside a vitest run is swallowed by the reporter — I ended up appending to a scratch file. All of it lived in `.tmp/` and is gone.
3. **`Array.isArray` will not narrow `readonly unknown[]`.** Cost one typecheck cycle to diagnose; the `in`-operator narrowing carries a comment so the next reader does not "simplify" it back.

**Conventions discovered / confirmed.**
* Comments here carry the *argument*, not the mechanics — what the previous implementation was, why it was wrong, and which test goes red if it returns. I matched that, including REVERT MAP entries with measured mutation outcomes rather than assumed ones.
* Every non-vacuity concern gets its own positive control, asserted **first** so it is visibly green. A construction count of 1 is meaningless without proof the rows were classified at all.
* `vi.mock` + `importOriginal` (rather than a stub) is the established way to observe a module boundary without changing its answers.
* `poolOptions.forks.singleFork` in `vitest.config.ts` is **inert** on vitest 4 and prints a deprecation warning on every run. Out of scope here, but worth a one-line cleanup — any comment still justifying single-fork behaviour is stale.
* `npm run typecheck:tests` and `scripts/i18n-check.mjs` both carry substantial pre-existing failure backlogs and cannot be used as absolute gates; compare against a stashed baseline instead.

* **`newestOfType` is the next hot spot** and section B's fix exposes it the way #394 exposed this one: 615 ms of the remaining 200/1000 cost is `fresherCoherentSelection` filtering the whole project list once per type per frozen row (7.0 ms with the same rows un-frozen). A newest-per-type map on `ProjectLineageSources` — the object now exists to hold exactly this kind of derived-once value — would close it.
* Sections **A** and **C** of #399 are untouched and independent; A is purely locale catalogues plus `rowLineage.test.ts`, C is a read-site dedupe in `_row_document_ids` plus a UX decision.
* `poolOptions` → top-level options is a mechanical vitest 4 migration that would silence a warning on every single test run.

### DIFF
--- a/voc-datalake/frontend/src/api/derivation.test.ts
+++ b/voc-datalake/frontend/src/api/derivation.test.ts
@@ -12,13 +12,18 @@
  *    three origins → "answers exactly what resolveDerivation answers", which pins
  *    a declared, a legacy and an unresolved-source document as whole objects
  *    rather than field by field, because the promise is the whole answer;
- *  * `resolveDerivation` indexing per SOURCE rather than per call, or a memo
- *    creeping in behind it → "builds one index per call and none per source" and
- *    "builds no index of its own when given one", which count the builder's
- *    invocations through a `Map` subclass instead of timing anything. Their
- *    positive control is the first assertion of each: an index that was never
- *    consulted would report zero lookups, so a count of zero constructions could
- *    otherwise be a resolver that reads no index at all.
+ *  * a memo creeping in behind `resolveDerivationAgainst`, or the resolver reading
+ *    the wire a second time instead of the index it was handed → "consults the index
+ *    once per source and holds no state between calls" and "reads what a source is
+ *    from the given index and never from the wire again", which count the index's
+ *    lookups through a `Map` subclass instead of timing anything. Their positive
+ *    control is the first assertion of each — the titles the index actually resolved
+ *    — because a count is otherwise indistinguishable from the silence of a resolver
+ *    that looks nothing up;
+ *  * `derivationSourceIndex` dropping a field a source displays, or keeping an
+ *    entry for a record naming no document → "indexes a document list once for
+ *    every field a source displays", which is what lets `resolved: false` mean "not
+ *    among the documents supplied" rather than "supplied without a title".
  */
 import { describe, it, expect } from 'vitest'
 import {
--- a/voc-datalake/frontend/src/api/derivation.test.ts
+++ b/voc-datalake/frontend/src/api/derivation.test.ts
@@ -5,14 +5,31 @@
  *
  * Every expectation is a literal — nothing here is derived from the code under
  * test.
+ *
+ * REVERT MAP for the prebuilt-index seam (issue #399 B):
+ *
+ *  * `resolveDerivationAgainst` diverging from `resolveDerivation` in ANY of the
+ *    three origins → "answers exactly what resolveDerivation answers", which pins
+ *    a declared, a legacy and an unresolved-source document as whole objects
+ *    rather than field by field, because the promise is the whole answer;
+ *  * `resolveDerivation` indexing per SOURCE rather than per call, or a memo
+ *    creeping in behind it → "builds one index per call and none per source" and
+ *    "builds no index of its own when given one", which count the builder's
+ *    invocations through a `Map` subclass instead of timing anything. Their
+ *    positive control is the first assertion of each: an index that was never
+ *    consulted would report zero lookups, so a count of zero constructions could
+ *    otherwise be a resolver that reads no index at all.
  */
 import { describe, it, expect } from 'vitest'
 import {
   DERIVATION_ROLES,
+  derivationSourceIndex,
   emptyDerivation,
   normalizeDerivation,
   resolveDerivation,
+  resolveDerivationAgainst,
 } from './derivation'
+import type { DerivationSourceFields } from './derivation'
 
 const PRD = { document_id: 'prd_1', document_type: 'prd', title: 'Onboarding PRD' }
 const PRFAQ = { document_id: 'prfaq_1', document_type: 'prfaq', title: 'Onboarding PR/FAQ' }
@@ -508,5 +525,104 @@ describe('a cyclic reference chain', () => {
     expect(resolveDerivation(self, [self]).sources).toEqual([
       { document_id: 'self', role: 'reference', title: 'Self', document_type: 'prd', resolved: true },
     ])
+  })
+})
+
+describe('resolving against an index the caller already built', () => {
+  const DOCUMENTS = [PRD, PRFAQ]
+
+  /**
+   * One document per origin the contract distinguishes, plus the shape whose
+   * sources do NOT resolve — which is the case an index can get wrong in a way the
+   * other two cannot, because a missing entry and an empty one are what `resolved`
+   * tells apart.
+   */
+  const BY_ORIGIN = {
+    declared: {
+      document_id: 'prototype_1',
+      derivation: {
+        sources: [
+          { document_id: 'prd_1', role: 'prototype_prd' },
+          { document_id: 'prfaq_1', role: 'prototype_prfaq' },
+        ],
+        selected_document_count: 3,
+        feedback_count: 7,
+        persona_ids: ['persona_1'],
+        visual_document_ids: ['vis_1'],
+        product_context_included: true,
+      },
+    },
+    legacy: { document_id: 'merge_1', source_documents: ['prd_1', 'prfaq_1'] },
+    unresolved: { document_id: 'merge_2', source_documents: ['deleted_1'] },
+    none: { document_id: 'handwritten_1', title: 'Typed by hand' },
+  }
+
+  it('answers exactly what resolveDerivation answers', () => {
+    // WHOLE OBJECTS, not fields: the promise of the seam is the whole answer, and a
+    // per-field comparison is the one that lets `origin` or `resolved` drift. Asked
+    // of all four origins — declared, legacy, a source the project no longer holds,
+    // and nothing recoverable — because each takes a different path to the index.
+    const index = derivationSourceIndex(DOCUMENTS)
+    for (const [origin, document] of Object.entries(BY_ORIGIN)) {
+      expect(resolveDerivationAgainst(document, index), origin)
+        .toEqual(resolveDerivation(document, DOCUMENTS))
+    }
+    // The control: the four documents really do exercise the three origins, so the
+    // equality above is over answers that DIFFER rather than four empty ones.
+    expect(Object.values(BY_ORIGIN).map((d) => resolveDerivation(d, DOCUMENTS).origin))
+      .toEqual(['declared', 'legacy', 'legacy', 'none'])
+  })
+
+  it('reads what a source is from the given index and never from the wire again', () => {
+    // The decisive case for "the index is the caller's": an index disagreeing with
+    // any document list, so an answer carrying its titles could not have come from
+    // re-indexing anything. A resolver that rebuilt internally has no list to
+    // rebuild from here and would report both sources unresolved.
+    const index = new Map<string, DerivationSourceFields>([
+      ['prd_1', { title: 'Indexed PRD', document_type: 'prd' }],
+      ['prfaq_1', { title: 'Indexed PR/FAQ', document_type: 'prfaq' }],
+    ])
+
+    expect(resolveDerivationAgainst(BY_ORIGIN.legacy, index).sources).toEqual([
+      { document_id: 'prd_1', role: 'merge_input', title: 'Indexed PRD', document_type: 'prd', resolved: true },
+      { document_id: 'prfaq_1', role: 'merge_input', title: 'Indexed PR/FAQ', document_type: 'prfaq', resolved: true },
+    ])
+  })
+
+  it('consults the index once per source and holds no state between calls', () => {
+    // Counted rather than timed, so the assertion is about the work done and not
+    // about this machine: a Map that records every lookup. Two calls over one index
+    // is the shape the prioritization page's loop takes, and the second must cost
+    // the same as the first — a resolver caching its answers would show fewer
+    // lookups on the second pass and start reporting a document deleted between two
+    // reads as still present.
+    const lookups: string[] = []
+    class CountingIndex extends Map<string, DerivationSourceFields> {
+      override get(id: string): DerivationSourceFields | undefined {
+        lookups.push(id)
+        return super.get(id)
+      }
+    }
+    const index = new CountingIndex(derivationSourceIndex(DOCUMENTS))
+
+    const first = resolveDerivationAgainst(BY_ORIGIN.declared, index)
+    const second = resolveDerivationAgainst(BY_ORIGIN.declared, index)
+
+    // The positive control: the index WAS the source of the answer, so the count
+    // below is a real measurement rather than the silence of a resolver that never
+    // looked anything up.
+    expect(first.sources.map((s) => s.title)).toEqual(['Onboarding PRD', 'Onboarding PR/FAQ'])
+    expect(second).toEqual(first)
+    expect(lookups).toEqual(['prd_1', 'prfaq_1', 'prd_1', 'prfaq_1'])
+  })
+
+  it('indexes a document list once for every field a source displays', () => {
+    // One entry per readable document, carrying both resolved fields, and nothing
+    // for a record naming no document — which is what lets `resolved: false` mean
+    // "not among the documents supplied" rather than "supplied without a title".
+    const index = derivationSourceIndex([PRD, { title: 'No id' }, null, 'garbage', { document_id: '' }])
+
+    expect([...index.keys()]).toEqual(['prd_1'])
+    expect(index.get('prd_1')).toEqual({ title: 'Onboarding PRD', document_type: 'prd' })
   })
 })
--- a/voc-datalake/frontend/src/api/derivation.ts
+++ b/voc-datalake/frontend/src/api/derivation.ts
@@ -263,24 +263,20 @@ export type DerivationSourceFields = Pick<DerivationSource, 'title' | 'document_
  * A project's documents reduced to what resolving a source needs: document_id →
  * the fields a source displays.
  *
- * BUILT ONCE PER PROJECT COLLECTION PASS AND HANDED BACK IN, which is what this
- * type exists to make possible. `resolveDerivation` used to build one of these on
- * every call, which cost nothing while its only callers resolved one document at a
- * time — and became the whole cost the moment a caller asked about a LIST. The
- * prioritization page calls the lineage classifiers per row and each classifier
- * calls the resolver per document on that row, so one project read of D documents
- * was walked once per (row × document) instead of once: measured inside the page's
- * `useMemo` at 200 rows / 1000 documents, 1644 ms in this container's jsdom
- * (562 ms on the reviewing machine — issue #399 B).
+ * ONE INDEX PER PROJECT READ, NOT PER CALL, and it is the CALLER'S to build and
+ * hold: `resolveDerivation` used to build one on every call, which cost nothing
+ * while its callers resolved one document at a time and became the whole cost the
+ * moment a caller asked about a LIST (issue #399 B — the measurement and the loop
+ * that exposed it are recorded at `ProjectLineageSources`, which is the caller-side
+ * type the fix introduced).
  *
  * A LOOKUP TABLE RATHER THAN AN OPAQUE HANDLE, deliberately: everything a source
- * needs is already normalised into it (`displayString`, so an unreadable title or
- * type is '' and never null — `hasSupersededSource` in
- * pages/Prioritization/rowLineage.ts turns on exactly that), and nothing is
- * memoised behind it, so the index's LIFETIME is its holder's and there is no
- * cache anywhere to invalidate. Build it from `derivationSourceIndex` rather than
- * by hand: a map whose values did not come through that builder can carry a null
- * where every reader expects '', which is the one way to make a resolved source
+ * needs is already normalised into it by `displayString`, so an unreadable title or
+ * type is '' and never null — a distinction consumers turn on. Nothing is memoised
+ * behind it either, so the index's LIFETIME is its holder's and there is no cache
+ * anywhere to invalidate. Build it from `derivationSourceIndex` rather than by
+ * hand: a map whose values did not come through that builder can carry a null where
+ * every reader expects '', which is the one way to make a resolved source
  * indistinguishable from an unresolved one.
  */
 export type DerivationSourceIndex = ReadonlyMap<string, DerivationSourceFields>
--- a/voc-datalake/frontend/src/api/derivation.ts
+++ b/voc-datalake/frontend/src/api/derivation.ts
@@ -257,15 +257,41 @@ function isEmpty(derivation: DocumentDerivation): boolean {
 }
 
 /** What a source carries once its document was found. */
-type ResolvedSourceFields = Pick<DerivationSource, 'title' | 'document_type'>
+export type DerivationSourceFields = Pick<DerivationSource, 'title' | 'document_type'>
 
 /**
- * Index of document_id → the fields a source displays, over whatever the caller
- * supplied. One index for every resolved field, so a consumer needs one pass
- * over the document list rather than one per field it wants to render.
+ * A project's documents reduced to what resolving a source needs: document_id →
+ * the fields a source displays.
+ *
+ * BUILT ONCE PER PROJECT COLLECTION PASS AND HANDED BACK IN, which is what this
+ * type exists to make possible. `resolveDerivation` used to build one of these on
+ * every call, which cost nothing while its only callers resolved one document at a
+ * time — and became the whole cost the moment a caller asked about a LIST. The
+ * prioritization page calls the lineage classifiers per row and each classifier
+ * calls the resolver per document on that row, so one project read of D documents
+ * was walked once per (row × document) instead of once: measured inside the page's
+ * `useMemo` at 200 rows / 1000 documents, 1644 ms in this container's jsdom
+ * (562 ms on the reviewing machine — issue #399 B).
+ *
+ * A LOOKUP TABLE RATHER THAN AN OPAQUE HANDLE, deliberately: everything a source
+ * needs is already normalised into it (`displayString`, so an unreadable title or
+ * type is '' and never null — `hasSupersededSource` in
+ * pages/Prioritization/rowLineage.ts turns on exactly that), and nothing is
+ * memoised behind it, so the index's LIFETIME is its holder's and there is no
+ * cache anywhere to invalidate. Build it from `derivationSourceIndex` rather than
+ * by hand: a map whose values did not come through that builder can carry a null
+ * where every reader expects '', which is the one way to make a resolved source
+ * indistinguishable from an unresolved one.
  */
-function sourceFieldIndex(documents: readonly unknown[]): Map<string, ResolvedSourceFields> {
-  const index = new Map<string, ResolvedSourceFields>()
+export type DerivationSourceIndex = ReadonlyMap<string, DerivationSourceFields>
+
+/**
+ * Index the documents a source may resolve against — one pass over the list, for
+ * every resolved field at once, so a consumer needs one pass rather than one per
+ * field it wants to render.
+ */
+export function derivationSourceIndex(documents: readonly unknown[]): DerivationSourceIndex {
+  const index = new Map<string, DerivationSourceFields>()
   for (const raw of documents) {
     const record = asRecord(raw)
     if (!record) continue
@@ -299,10 +325,39 @@ function sourceFieldIndex(documents: readonly unknown[]): Map<string, ResolvedSo
  * never a source's sources. A cyclic chain (A built from B, B built from A) is
  * therefore inert — each call returns the other document, once, and there is no
  * traversal to loop.
+ *
+ * ONE DOCUMENT AT A TIME, and a caller asking about a LIST should reach for
+ * `resolveDerivationAgainst` instead: this overload indexes `projectDocuments`
+ * afresh on every call, so asking it about N documents of one project walks that
+ * project's list N times. Kept exactly as it was because that is the right
+ * signature for the callers that hold one document (DocumentsTab), and because
+ * every existing caller's behaviour is this function's contract.
  */
 export function resolveDerivation(
   document: unknown,
   projectDocuments: readonly unknown[] = [],
+): ResolvedDerivation {
+  return resolveDerivationAgainst(document, derivationSourceIndex(projectDocuments))
+}
+
+/**
+ * `resolveDerivation` against an index the caller already built — the same answer,
+ * without the per-call pass over the project's documents.
+ *
+ * THE SAME FUNCTION, not a faster approximation of it: `resolveDerivation` is now
+ * this function plus one `derivationSourceIndex` call, so there is no second
+ * resolution rule that could drift from the first. `derivation.test.ts` pins the
+ * equivalence on a declared, a legacy and an unresolved-source document anyway,
+ * because "one function delegates to the other" is a fact about today's source and
+ * the promise is about the answers.
+ *
+ * @param index Built by `derivationSourceIndex` from the project's documents, and
+ *   owned by whoever built it — see `DerivationSourceIndex` for why this is
+ *   explicit plumbing rather than a memo inside the resolver.
+ */
+export function resolveDerivationAgainst(
+  document: unknown,
+  index: DerivationSourceIndex,
 ): ResolvedDerivation {
   const record = asRecord(document)
   if (!record) return { ...emptyDerivation(), sources: [], origin: 'none' }
@@ -311,7 +366,6 @@ export function resolveDerivation(
   const useDeclared = !isEmpty(declared)
   const derivation = useDeclared ? declared : derivationFromLegacyFields(record)
 
-  const fields = sourceFieldIndex(projectDocuments)
   return {
     ...derivation,
     origin: originOf(derivation, useDeclared),
@@ -319,7 +373,7 @@ export function resolveDerivation(
       // One lookup decides all three: an unresolved source degrades every
       // resolved field to null together, so a consumer cannot render a type
       // without a title or vice versa.
-      const found = fields.get(source.document_id)
+      const found = index.get(source.document_id)
       return {
         document_id: source.document_id,
         role: source.role,
--- a/voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
+++ b/voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
@@ -1,0 +1,189 @@
+/**
+ * One project read is indexed ONCE for every row and every lineage rule asked
+ * about it (issue #399 B).
+ *
+ * COUNTED, NEVER TIMED, and that is the whole reason this file exists rather than
+ * an assertion inside `prioritizationUtils.test.ts`'s `collectRows` block: the
+ * defect was quadratic work, and a wall-clock threshold for it would either pass on
+ * a fast machine with the defect restored or fail on a loaded one without it. The
+ * countable thing is the number of times `derivationSourceIndex` is CALLED, which is
+ * a property of the code and identical on every machine — so the mock below wraps
+ * the real builder and changes nothing about the answers.
+ *
+ * REVERT MAP:
+ *
+ *  * `collectRows` passing `project.documents` instead of the prepared
+ *    `project.lineage` → "indexes a project read once however many rows name it"
+ *    (measured: the count goes from 1 to 13 at three rows, one per row per rule);
+ *  * `fresherCoherentSelection` passing `projectDocuments` to its nested
+ *    `classifySelectionLineage` → the same case (measured: 1 → 2, since one frozen
+ *    row's candidate is re-indexed);
+ *  * `projectLineageSources` hoisted out of the details loop, or shared between
+ *    projects → "indexes each project read separately", which is the assertion that
+ *    stops "once" being satisfied by an index built once for the whole page and
+ *    handed to every project — a row would then resolve its sources against another
+ *    project's documents;
+ *  * `lineageSourcesOf` building a fresh index even for a prepared read → both
+ *    cases.
+ *
+ * THE POSITIVE CONTROL each case carries is its lineage assertion, and it is not a
+ * formality: a `collectRows` that stopped classifying lineage altogether would also
+ * report one construction, or none. So each case first pins the row states the
+ * fixture is built to produce — a crossing, a coherent chain and an absent one — and
+ * only then the count.
+ */
+import { describe, it, expect, vi } from 'vitest'
+import type { PrioritizationRow, Project, ProjectDocument } from '../../api/types'
+
+/**
+ * The real builder, wrapped in a counter.
+ *
+ * `importOriginal` rather than a stub, following `ScraperCard.test.tsx`: the answers
+ * must be the production ones, because the control assertions below read the lineage
+ * states this fixture is built to produce. Only the call count is observed.
+ */
+const indexBuilds = vi.hoisted(() => ({ count: 0 }))
+vi.mock('../../api/derivation', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../api/derivation')>()
+  return {
+    ...actual,
+    derivationSourceIndex: (documents: readonly unknown[]) => {
+      indexBuilds.count += 1
+      return actual.derivationSourceIndex(documents)
+    },
+  }
+})
+
+const { collectRows } = await import('./prioritizationUtils')
+
+const project = (projectId: string, name: string): Project => ({
+  project_id: projectId,
+  name,
+  description: '',
+  status: 'active',
+  created_at: '',
+  updated_at: '',
+  persona_count: 0,
+  document_count: 0,
+})
+
+const doc = (
+  documentId: string,
+  documentType: ProjectDocument['document_type'],
+  createdAt: string,
+  extra: Record<string, unknown> = {},
+): ProjectDocument => ({
+  document_id: documentId,
+  document_type: documentType,
+  title: `${documentType} ${documentId}`,
+  content: '',
+  created_at: createdAt,
+  ...extra,
+})
+
+/** A declared derivation naming one source, as `lambda/shared/derivation.py` writes it. */
+const builtFrom = (...ids: readonly string[]) => ({
+  derivation: {
+    sources: ids.map((id) => ({ document_id: id, role: 'reference' })),
+    selected_document_count: ids.length,
+    feedback_count: 0,
+    persona_ids: [],
+    visual_document_ids: [],
+    product_context_included: false,
+  },
+})
+
+const storedRow = (
+  rowId: string,
+  projectId: string,
+  documentIds: readonly string[],
+  isFrozen = false,
+): PrioritizationRow => ({
+  row_id: rowId,
+  project_id: projectId,
+  document_ids: [...documentIds],
+  prototype_id: '',
+  is_default: true,
+  created_at: '2026-01-01',
+  is_frozen: isFrozen,
+})
+
+/**
+ * Documents of ONE project, shaped so the three rows below land in three different
+ * lineage states — which is what makes the control assertions distinguishing.
+ *
+ * `prfaq_1` is built from `prd_1`, and `prd_2` supersedes `prd_1`, so a row holding
+ * {prd_2, prfaq_1} crosses generations while {prd_2, prfaq_2} is one chain and
+ * {prd_3} (a hand-authored PRD recording nothing) is lineage-absent.
+ */
+const DOCUMENTS: ProjectDocument[] = [
+  doc('prd_1', 'prd', '2025-01-01T09:00:00Z'),
+  doc('prfaq_1', 'prfaq', '2025-01-01T09:00:00Z', builtFrom('prd_1')),
+  doc('prd_2', 'prd', '2025-02-01T09:00:00Z', builtFrom('prd_1')),
+  doc('prfaq_2', 'prfaq', '2025-02-01T09:00:00Z', builtFrom('prd_2')),
+  doc('prd_3', 'prd', '2025-03-01T09:00:00Z'),
+]
+
+const ROWS: Record<string, PrioritizationRow> = {
+  crossing: storedRow('crossing', 'p1', ['prd_2', 'prfaq_1'], true),
+  coherent: storedRow('coherent', 'p1', ['prd_2', 'prfaq_2'], true),
+  absent: storedRow('absent', 'p1', ['prd_3'], true),
+}
+
+/** State and reason per row id, so a control assertion reads as one object. */
+const lineageByRow = (rows: ReturnType<typeof collectRows>) => Object.fromEntries(
+  rows.map((row) => [row.row_id, `${row.lineage.state}/${row.lineage.reason}`]),
+)
+
+describe('collectRows indexes each project read once', () => {
+  it('indexes a project read once however many rows name it', () => {
+    indexBuilds.count = 0
+
+    const rows = collectRows(ROWS, [{ documents: DOCUMENTS }], [project('p1', 'P1')])
+
+    // The control, asserted FIRST: three rows really were classified, in three
+    // different states, off the shared index. A `collectRows` that skipped the
+    // lineage would report one construction too — or none — so the count below only
+    // means something once this holds.
+    expect(lineageByRow(rows)).toEqual({
+      crossing: 'crossGeneration/supersededSource',
+      coherent: 'coherent/oneChain',
+      absent: 'absent/noneRecorded',
+    })
+    // ONE, not "three" and not "one per rule": the three rows, the classifier's two
+    // derivation-reading rules, and the frozen rows' candidate classification all
+    // read the same index. Before this change the resolver built one per call, so the
+    // same fixture built 13.
+    expect(indexBuilds.count).toBe(1)
+  })
+
+  it('indexes each project read separately', () => {
+    // The other direction, and the reason "once" is not enough on its own: an index
+    // built once for the WHOLE page would satisfy the case above while resolving one
+    // project's sources against another project's documents. Two projects, so the
+    // count tracks project reads rather than being pinned at 1.
+    indexBuilds.count = 0
+    const second = [doc('prd_9', 'prd', '2025-01-01T09:00:00Z', builtFrom('prd_1'))]
+
+    const rows = collectRows(
+      { ...ROWS, other: storedRow('other', 'p2', ['prd_9'], true) },
+      [{ documents: DOCUMENTS }, { documents: second }],
+      [project('p1', 'P1'), project('p2', 'P2')],
+    )
+
+    // The control: `prd_9` names `prd_1` as its source, which project 2 does NOT
+    // hold — so it resolves against project 2's own index and stays unresolved,
+    // reading `coherent` rather than crossing with project 1's PRD.
+    expect(lineageByRow(rows).other).toBe('coherent/oneChain')
+    expect(indexBuilds.count).toBe(2)
+  })
+
+  it('builds nothing for a page whose project reads have not landed', () => {
+    // The floor: no project read, no index. Pins that the builder moved into the
+    // details loop rather than being called unconditionally per page render.
+    indexBuilds.count = 0
+
+    expect(collectRows(ROWS, undefined, undefined)).toStrictEqual([])
+    expect(indexBuilds.count).toBe(0)
+  })
+})
--- a/voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
+++ b/voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
@@ -10,6 +10,19 @@
  * a property of the code and identical on every machine — so the mock below wraps
  * the real builder and changes nothing about the answers.
  *
+ * COUNTING THE INDEX BUILDER ALONE WAS NOT ENOUGH, which is the reason the second
+ * counter below exists. `vi.mock` replaces the binding OTHER modules import;
+ * `resolveDerivation` calls `derivationSourceIndex` through the module-local binding
+ * inside `derivation.ts`, which no mock of this module can reach. So the single most
+ * likely regression — `hasSupersededSource`/`recordsNoLineage` going back to
+ * `resolveDerivation(raw, projectDocuments)`, which is exactly the code #399 B
+ * replaces — rebuilt one index per resolver call and still reported
+ * `indexBuilds.count === 1` (measured: the whole suite stayed green under that
+ * revert). What closes the hole is counting the OTHER side of the seam: every
+ * resolution on this page's path must go through `resolveDerivationAgainst`, so
+ * `resolveDerivation` must be called ZERO times during `collectRows`, and any route
+ * back through it is visible whether or not it happens to rebuild an index.
+ *
  * REVERT MAP:
  *
  *  * `collectRows` passing `project.documents` instead of the prepared
@@ -18,6 +31,13 @@
  *  * `fresherCoherentSelection` passing `projectDocuments` to its nested
  *    `classifySelectionLineage` → the same case (measured: 1 → 2, since one frozen
  *    row's candidate is re-indexed);
+ *  * `hasSupersededSource`/`recordsNoLineage` reverted to their pre-#399 B
+ *    `resolveDerivation(raw, projectDocuments)` — the faithful `git revert` of the
+ *    hot path, with `collectRows`' `projectLineageSources` left in place → "indexes a
+ *    project read once however many rows name it", and ONLY via its
+ *    `resolveDerivationCalls.count` assertion (measured: 11 calls, expected 0; the
+ *    `indexBuilds.count === 1` assertion beside it stays green, which is why that
+ *    assertion is not the one this entry names);
  *  * `projectLineageSources` hoisted out of the details loop, or shared between
  *    projects → "indexes each project read separately", which is the assertion that
  *    stops "once" being satisfied by an index built once for the whole page and
@@ -36,13 +56,23 @@ import { describe, it, expect, vi } from 'vitest'
 import type { PrioritizationRow, Project, ProjectDocument } from '../../api/types'
 
 /**
- * The real builder, wrapped in a counter.
+ * The real builder and the per-call resolver, each wrapped in a counter.
  *
  * `importOriginal` rather than a stub, following `ScraperCard.test.tsx`: the answers
  * must be the production ones, because the control assertions below read the lineage
- * states this fixture is built to produce. Only the call count is observed.
+ * states this fixture is built to produce. Only the call counts are observed.
+ *
+ * TWO COUNTERS BECAUSE ONE OF THEM CANNOT SEE INSIDE `derivation.ts`. This mock
+ * replaces the bindings other modules import, so `indexBuilds` counts the index
+ * builds `rowLineage.ts`/`prioritizationUtils.ts` ask for and NOT the one
+ * `resolveDerivation` makes for itself through its module-local binding.
+ * `resolveDerivationCalls` covers exactly that blind spot: it is the per-call
+ * resolver this optimisation replaces, so on the page's path it must never be
+ * reached at all. See the file docstring for the revert that proved one counter
+ * insufficient.
  */
 const indexBuilds = vi.hoisted(() => ({ count: 0 }))
+const resolveDerivationCalls = vi.hoisted(() => ({ count: 0 }))
 vi.mock('../../api/derivation', async (importOriginal) => {
   const actual = await importOriginal<typeof import('../../api/derivation')>()
   return {
@@ -50,6 +80,10 @@ vi.mock('../../api/derivation', async (importOriginal) => {
     derivationSourceIndex: (documents: readonly unknown[]) => {
       indexBuilds.count += 1
       return actual.derivationSourceIndex(documents)
+    },
+    resolveDerivation: (document: unknown, projectDocuments?: readonly unknown[]) => {
+      resolveDerivationCalls.count += 1
+      return actual.resolveDerivation(document, projectDocuments)
     },
   }
 })
@@ -138,6 +172,7 @@ const lineageByRow = (rows: ReturnType<typeof collectRows>) => Object.fromEntrie
 describe('collectRows indexes each project read once', () => {
   it('indexes a project read once however many rows name it', () => {
     indexBuilds.count = 0
+    resolveDerivationCalls.count = 0
 
     const rows = collectRows(ROWS, [{ documents: DOCUMENTS }], [project('p1', 'P1')])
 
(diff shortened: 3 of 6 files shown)


perrozzi:

This PR hoists the `document_id → {title, document_type}` index out of `resolveDerivation` so a whole `collectRows` pass builds one index per project read instead of one per (row × document × rule). The design is sound: `resolveDerivation` delegates to `resolveDerivationAgainst`, so there is exactly one resolution rule; the index travels bundled with the documents it was built from (`ProjectLineageSources`), which rules out the "index from a different read" failure mode; nothing is memoized, so there is no invalidation surface. No blockers found — the remarks below are about test reach and comment volume.

### Should fix

- **The construction-count test has a blind spot.** `prioritizationUtils.indexReuse.test.ts` mocks `../../api/derivation` and only observes *cross-module* calls to `derivationSourceIndex`. `resolveDerivation`'s internal call goes through the module-local binding inside `derivation.ts` and is never counted. So a regression where `rowLineage.ts` goes back to `resolveDerivation(raw, projectDocuments)` (arguably the most likely reversion, since that is the code being replaced in `hasSupersededSource` / `recordsNoLineage`) would still report `indexBuilds.count === 1` and pass. The listed mutations that were caught all happen to route through `projectLineageSources`. Consider also counting `resolveDerivation` invocations in the same mock (expected: 0 during `collectRows`), which closes the gap directly.
- **No direct test that the two accepted shapes give identical answers.** `rowLineage.ts` now takes `readonly unknown[] | ProjectLineageSources` on three exported rules, but `rowLineage.test.ts` only exercises the list arm (93 existing call sites) and the prepared arm is only covered indirectly through `collectRows`. One case per rule asserting `rule(sel, docs)` deep-equals `rule(sel, projectLineageSources(docs))` would pin the union's contract where it is declared.
- **`lineageSourcesOf` is no longer tolerant of a missing project read.** `'sourceIndex' in project` throws `TypeError` on `null`/`undefined`, whereas the old path bottomed out in `resolveDerivation(..., projectDocuments = [])` and simply resolved nothing. Types cover the in-repo call sites, but this is a boundary helper reachable from component code; an `Array.isArray(project) || !project` short-circuit (or `project != null && 'sourceIndex' in project`) keeps the previous failure mode.

### Nits

- The REVERT MAP added to `src/api/derivation.test.ts` cites case names that do not exist in the file ("builds one index per call and none per source", "builds no index of its own when given one"); the actual cases are "consults the index once per source…" and "reads what a source is from the given index…". A revert map that names missing tests is worse than none.
- The `1644 ms` / `615 ms` / `562 ms` figures are now duplicated in four places (`DerivationSourceIndex`, `rowLineage.ts` module header, `ProjectLineageSources`, `collectRows`, `PrioritizationRowView.lineage`) and are container-specific. Keep them in the PR/issue, or in one docstring, not five.
- `collectRows` now builds `byId` **and** the source index for every project read, including projects with no rows on the page; previously both were built lazily per row. Cheap in absolute terms, but it is a behaviour change the "builds nothing for a page whose project reads have not landed" test does not distinguish.
- The new prose-to-code ratio in `rowLineage.ts` and `prioritizationUtils.ts` is high even for this repo (~40 lines of logic, ~90 lines of rationale, partly restating each other). Trimming the repeated "why the index is the caller's" argument to a single canonical location and cross-referencing would age better.
- Out of scope but noted in the description: the `poolOptions.forks.singleFork` deprecation and the `newestOfType` O(rows × documents) residual are worth their own issues so they don't get lost.

---

voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
```diff
@@ -0,0 +1,189 @@
...
+ * `importOriginal` rather than a stub, following `ScraperCard.test.tsx`: the answers
+ * must be the production ones, because the control assertions below read the lineage
+ * states this fixture is built to produce. Only the call count is observed.
+ */
+const indexBuilds = vi.hoisted(() => ({ count: 0 }))
+vi.mock('../../api/derivation', async (importOriginal) => {
+  const actual = await importOriginal<typeof import('../../api/derivation')>()
+  return {
+    ...actual,
+    derivationSourceIndex: (documents: readonly unknown[]) => {
```
perrozzi, line 80: **Type**: issue
**Severity**: medium
**Title**: The construction-count mock cannot see `resolveDerivation`'s own index build, so the most likely regression passes

**Description**: This `vi.mock` replaces the `derivationSourceIndex` binding that *other* modules import. `resolveDerivation` (inside `derivation.ts`) calls `derivationSourceIndex` through the module-local binding, which the mock never touches — so any regression that routes back through `resolveDerivation` is invisible to `indexBuilds.count`.

I verified this empirically rather than by reading: I copied the tree to ignored scratch and reverted `hasSupersededSource`/`recordsNoLineage` in `rowLineage.ts` to their pre-PR signatures calling `resolveDerivation(raw, projectDocuments)`, leaving `collectRows`' `projectLineageSources` in place — i.e. exactly the `git revert` of the hot path. The full suite ran **195 files / 3330 tests, all green**, and all three cases in this file passed, `indexes a project read once however many rows name it` included. I also confirmed the reach directly: a plain `resolveDerivation(doc, docs)` call registers **0** on the counter while a direct `derivationSourceIndex(docs)` registers 1.

So the property this file exists to pin — "one project read, one index" — is not actually enforced against the reversion that is most likely to happen, since `resolveDerivation` is the code being replaced. The mutations listed in the REVERT MAP all happen to keep routing through `projectLineageSources`, which is why they were caught.

**Proposed fix**: Extend the same `vi.mock` factory to also wrap `resolveDerivation` with its own counter and assert it is **0** during `collectRows` (every resolution in the page's path should go through `resolveDerivationAgainst`). That closes the gap without any new machinery, and it stays a count rather than a timing. Add the corresponding REVERT MAP entry, ideally with the measured outcome of the `resolveDerivation(raw, projectDocuments)` revert.

**AI prompt**: In `voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts`, the `vi.mock('../../api/derivation', …)` factory currently only wraps `derivationSourceIndex`, which counts cross-module calls only — `resolveDerivation`'s internal call to `derivationSourceIndex` uses the module-local binding and is not counted. Extend the factory to also wrap `resolveDerivation` in a hoisted counter (e.g. `resolveDerivationCalls`), keeping `importOriginal` so real answers are returned. In `indexes a project read once however many rows name it`, after the existing lineage-state control assertion and the `indexBuilds.count === 1` assertion, add an assertion that `resolveDerivation` was called 0 times during `collectRows` (all resolution in this path must go through `resolveDerivationAgainst`). Update the file's REVERT MAP docstring with an entry for "`hasSupersededSource`/`recordsNoLineage` reverted to `resolveDerivation(raw, projectDocuments)`" naming the case that now catches it. Verify by making that revert in a scratch copy, confirming the new assertion fails, then restoring.

voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts
```diff
@@ -134,6 +146,68 @@ export interface RowLineage extends SelectionLineage {
...
+/**
+ * The prepared read, building one only for the caller that passed a plain list.
+ *
+ * Narrowed on `'sourceIndex' in project` rather than on `Array.isArray`, which
+ * `tsc` refuses to narrow a `readonly unknown[]` arm out of (its signature answers
+ * `any[]`, and a readonly array is not one) — leaving the list in the prepared
+ * branch's type.
+ */
+function lineageSourcesOf(project: ProjectLineageRead): ProjectLineageSources {
+  return 'sourceIndex' in project ? project : projectLineageSources(project)
```
perrozzi, line 224: **Type**: issue
**Severity**: minor
**Title**: `'sourceIndex' in project` throws on a null/undefined project read where the old path resolved nothing

**Description**: The `in` operator raises `TypeError: Cannot use 'in' operator to search for 'sourceIndex' in undefined`. Before this PR, `classifySelectionLineage(sel, undefined)` bottomed out in `resolveDerivation(raw, projectDocuments)` whose `projectDocuments = []` default absorbed it, and the rule simply resolved nothing. I verified both directions against the two revisions: on `development` `undefined` returns an answer and `null` throws; on this branch **both** throw.

Every in-repo call site is typed, so this is not reachable today and does not block. It is worth noting because (a) this is a boundary helper reachable from component code and the module docstring promises "Pure and total … no throwing", and (b) the failure mode this module chose everywhere else is silence, not an exception — a project detail that has not landed is exactly the shape a caller might pass through.

**Proposed fix**: Short-circuit the nullish case in `lineageSourcesOf`, e.g. `return project != null && 'sourceIndex' in project ? project : projectLineageSources(project ?? [])`, restoring the previous "resolves nothing" behaviour. One line, and it keeps the `in`-narrowing comment above intact.

**AI prompt**: In `voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts`, `lineageSourcesOf` uses `'sourceIndex' in project`, which throws a `TypeError` when a caller passes `null`/`undefined` for the project read — whereas the pre-#404 path resolved nothing (via `resolveDerivation`'s `projectDocuments = []` default). Make it total again: treat a nullish project read as an empty document list, e.g. `return project != null && 'sourceIndex' in project ? project : projectLineageSources(project ?? [])`. Keep the existing comment explaining why `Array.isArray` cannot narrow `readonly unknown[]`, and add one sentence noting the nullish guard restores the module's "no throwing" contract. Add a small case in `rowLineage.test.ts` asserting `classifySelectionLineage(selection, undefined as never)` returns an answer rather than throwing.

voc-datalake/frontend/src/api/derivation.test.ts
```diff
@@ -5,14 +5,31 @@
...
  * test.
+ *
+ * REVERT MAP for the prebuilt-index seam (issue #399 B):
+ *
+ *  * `resolveDerivationAgainst` diverging from `resolveDerivation` in ANY of the
+ *    three origins → "answers exactly what resolveDerivation answers", which pins
+ *    a declared, a legacy and an unresolved-source document as whole objects
+ *    rather than field by field, because the promise is the whole answer;
+ *  * `resolveDerivation` indexing per SOURCE rather than per call, or a memo
+ *    creeping in behind it → "builds one index per call and none per source" and
```
perrozzi: **Type**: issue
**Severity**: minor
**Title**: REVERT MAP names two test cases that do not exist in this file

**Description**: The entry cites `"builds one index per call and none per source"` and `"builds no index of its own when given one"`. Neither name appears in the file; the actual cases added are `consults the index once per source and holds no state between calls` and `reads what a source is from the given index and never from the wire again`. Per repository convention a REVERT MAP is a navigational contract — a reader hitting a red test greps for the named case — so a map naming missing tests actively misleads. (Informed by repository memory: REVERT MAP docstrings citing named cases with measured mutation outcomes are an established convention here, from earlier PRs in this series.)

**Proposed fix**: Replace the two cited names with the real case titles.

**AI prompt**: In `voc-datalake/frontend/src/api/derivation.test.ts`, the REVERT MAP in the file docstring cites two test names that do not exist: "builds one index per call and none per source" and "builds no index of its own when given one". Replace them with the actual case titles in this file — "consults the index once per source and holds no state between calls" and "reads what a source is from the given index and never from the wire again" — keeping the surrounding rationale about counting rather than timing and the positive-control note. Then grep the file to confirm every case name cited in the REVERT MAP appears in an `it(...)` title.

voc-datalake/frontend/src/api/derivation.ts
```diff
@@ -257,15 +257,41 @@ function isEmpty(derivation: DocumentDerivation): boolean {
...
+ * the fields a source displays.
+ *
+ * BUILT ONCE PER PROJECT COLLECTION PASS AND HANDED BACK IN, which is what this
+ * type exists to make possible. `resolveDerivation` used to build one of these on
+ * every call, which cost nothing while its only callers resolved one document at a
+ * time — and became the whole cost the moment a caller asked about a LIST. The
+ * prioritization page calls the lineage classifiers per row and each classifier
+ * calls the resolver per document on that row, so one project read of D documents
+ * was walked once per (row × document) instead of once: measured inside the page's
+ * `useMemo` at 200 rows / 1000 documents, 1644 ms in this container's jsdom
```
perrozzi: **Type**: issue
**Severity**: minor
**Title**: The `1644 ms` / `615 ms` / `562 ms` numbers are duplicated across five docstrings

**Description**: These container-specific measurements now appear at `DerivationSourceIndex` (this line), the `rowLineage.ts` module header, `ProjectLineageSources`, `collectRows`, and `PrioritizationRowView.lineage`. Two of them are already qualified differently ("in this container's jsdom" vs "as reviewed"), which is how five copies of one number start to disagree. The measurement matters once — the invariant ("one index per project read, not per call") is what every site needs to state.

Also worth noting: `derivation.ts` is a shared API-layer module, and this docstring reaches down into `pages/Prioritization/rowLineage.ts` for both its motivating example and its `hasSupersededSource` reference. The dependency direction of the prose is the opposite of the import direction.

**Proposed fix**: Keep the full figures in one canonical place — the `ProjectLineageSources` docstring reads like the right owner, since it is the type the fix introduced — and have the other sites state the invariant and cross-reference it (e.g. "see `ProjectLineageSources`, issue #399 B"). Prefer the reviewed figure (562 ms) where a number is quoted at all, since container wall clock is not reproducible for a future reader.

**AI prompt**: In `voc-datalake/frontend/src/api/derivation.ts`, `src/pages/Prioritization/rowLineage.ts` and `src/pages/Prioritization/prioritizationUtils.ts`, the measurements "1644 ms", "615 ms" and "562 ms" for the 200-row / 1000-document fixture are repeated in five docstrings. Consolidate: keep the full before/after figures only in the `ProjectLineageSources` docstring in `rowLineage.ts` (labelled with the fixture and issue #399 B), and reduce the other four sites to the invariant plus a cross-reference ("one source index per project read, not per call — see `ProjectLineageSources`, issue #399 B"). While there, trim the `DerivationSourceIndex` docstring in `api/derivation.ts` so the shared API module no longer explains itself through `pages/Prioritization` specifics; keep only the normalisation guarantee (every value has been through `displayString`, so an unreadable field is `''` and never `null`) and the instruction to build the index via `derivationSourceIndex`. Re-run `npx tsc -b --noEmit` and the derivation/Prioritization test files afterwards.

perrozzi (commented):
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

Reviewed the full contents of all five changed files, the base revisions of `rowLineage.ts`/`derivation.ts`, issue #399, and the existing bot review. The design is the right one for section B: `resolveDerivation` delegating to `resolveDerivationAgainst` means there is exactly one resolution rule, so the equivalence claim needs no argument; bundling the index with the documents it was built from (`ProjectLineageSources`) structurally rules out "index from a different read"; and nothing is memoized, so there is no invalidation surface. No blocking defect found in the production code.

**Positive note worth keeping:** `projectLineageSources` returning documents + index as *one* object (rather than a second prebuilt-index parameter) is the choice that makes the fast path unabusable, and the `indexes each project read separately` case pins it. That is the strongest part of this PR.

The findings below are about test reach and a few residuals. I verified two of them empirically with throwaway mutants in ignored scratch (`.tmp/`, removed; working tree clean):

1. **The construction-count test does not catch the most likely regression.** I reverted `hasSupersededSource`/`recordsNoLineage` in `rowLineage.ts` to their pre-PR signatures calling `resolveDerivation(raw, projectDocuments)` — i.e. a faithful `git revert` of the hot path, keeping `projectLineageSources` in `collectRows`. Result: **195 test files / 3330 tests, all green**, including all three cases of `prioritizationUtils.indexReuse.test.ts`. The `vi.mock` only intercepts *cross-module* imports; `resolveDerivation`'s own call to `derivationSourceIndex` uses the module-local binding and is invisible to the counter. I confirmed that reach separately (a direct `resolveDerivation` call registers 0 constructions on the counter). Detail inline.
2. **A missing project read now throws instead of resolving nothing.** On `development`, `classifySelectionLineage(sel, undefined)` returned an answer (`resolveDerivation`'s `projectDocuments = []` default absorbed it); with this diff `'sourceIndex' in undefined` raises `TypeError`. Verified against both revisions. Types cover every in-repo call site, so this is minor — but it is a behaviour change on a boundary helper reachable from component code, and it is not what the module's "pure and total … an unreadable document contributes nothing rather than failing a row" docstring promises.

Non-file-specific notes:

- **`newestOfType` residual, confirmed independently.** My own probe of the post-fix tree at 200 frozen rows / 1000 documents gives a ~517 ms median `collectRows` in this container — consistent with the PR's 615 ms figure and with the diagnosis that the remainder is `fresherCoherentSelection`, not index construction. The PR is right to leave it out; please open the follow-up issue so it is not lost, together with the `poolOptions` vitest 4 migration (that deprecation does print on every run here).
- **Sections A and C of #399 are correctly untouched** — nothing in this diff reaches a locale catalogue or `_row_document_ids`.

**Build / test results (step 2 of my workflow):**

| gate | result |
|---|---|
| `mise run build` | ❌ `no tasks defined` — this repo defines no mise tasks (matches the PR description) |
| `npm run build` (root → frontend) | ❌ fails in `prebuild`: `ts-node: not found`; root/`voc-datalake` workspaces have no `node_modules` in this container. Environment, not this PR |
| `npx tsc -b --noEmit` (frontend) | ✅ clean |
| `npx vitest run` (full frontend suite) | ✅ **193 files / 3287 tests passed** |
| `npx vitest run src/api/derivation.test.ts src/pages/Prioritization` | ✅ 13 files / 501 tests passed |
| `npx eslint` on the three source files | ✅ clean (the two `*.test.ts` files are matched by an ESLint ignore pattern, so `--max-warnings 0` reports them as ignored-file warnings — pre-existing repo behaviour) |

**Needs attention before merge:** finding 1 (the counting test's blind spot) — it is the one assertion this PR adds to protect the fix, and today it does not.

voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.ts
```diff
@@ -1547,23 +1550,42 @@ export function collectRows(
...
   }>()
   for (const [index, detail] of allProjectDetails.entries()) {
     const project = projects[index]
     if (!project || !detail) continue
+    const documents = detail.documents ?? []
     byProject.set(project.project_id, {
       name: project.name,
-      documents: detail.documents ?? [],
+      documents,
+      byId: new Map(documents.map((doc) => [doc.document_id, doc])),
```
perrozzi, line 1586: **Type**: comment
**Severity**: minor
**Title**: `byId` and the index are now built for projects with no rows on the page

**Description**: Both `byId` and `lineage` are now built for **every** project read, whereas before `byId` was built lazily per row (and thus only for projects that actually have rows). For a page whose visible rows cover a subset of loaded projects, that is strictly more work than before for the unused projects — one `Map` and one index over each of their document lists.

Almost certainly the right trade: the whole point is to move this out of the per-row loop, and `Prioritization.tsx` loads exactly the projects it shows, so the untouched-project case is largely hypothetical. Recording it because `builds nothing for a page whose project reads have not landed` covers only the *no reads at all* floor, not the *reads landed, no rows* case, so the behaviour change is not visible in the tests. No change requested.

voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts
```diff
@@ -134,6 +146,68 @@ export interface RowLineage extends SelectionLineage {
...
+ * `ProjectLineageSources` already prepared from them.
+ *
+ * BOTH, so hoisting the index is the CALLER'S optimisation and not a new contract
+ * every call site has to satisfy. A caller holding one selection and one document
+ * list — every case at this seam, and `RowLineagePanels`' eventual one — passes the
+ * list and pays for one index, exactly as before; the page's per-row loop prepares
+ * the read once and passes that. The two paths are the same code one line down:
+ * `lineageSourcesOf` builds from a list and returns a prepared read untouched.
+ */
+export type ProjectLineageRead = readonly unknown[] | ProjectLineageSources
```
perrozzi, line 205: **Type**: question
**Severity**: minor
**Title**: Is the prepared arm of the union covered anywhere the union is declared?

**Description**: All three exported rules now accept `readonly unknown[] | ProjectLineageSources`, but `rowLineage.test.ts` exercises only the list arm (its 93 existing call sites pass lists), and the prepared arm is reached only indirectly through `collectRows`. Was a per-rule equivalence case — `rule(sel, docs)` deep-equalling `rule(sel, projectLineageSources(docs))` for each of `classifySelectionLineage`, `fresherCoherentSelection` and `rowLineageOf` — considered and rejected, or just not needed given the `collectRows` coverage?

Asking because the equivalence argument that carries `resolveDerivationAgainst` ("one function delegates to the other") does not carry here in the same way: `fresherCoherentSelection` reads `sources.documents` for `newestOfType` and `sources.sourceIndex` for the nested classify, so the two arms diverge at three separate reads rather than one. `derivation.test.ts` pins the equivalence for the resolver seam by explicit choice; the same reasoning seems to apply one layer up.

perrozzi:
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

**Review summary for #404** (head `92f3594`) — [review](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/404#pullrequestreview-5088508857)

**Findings: 6 total** — 4 issues (1 medium, 3 minor), 1 comment, 1 question.

| # | type | severity | file | finding |
|---|---|---|---|---|
| 1 | issue | medium | `prioritizationUtils.indexReuse.test.ts` | the `vi.mock` counter cannot see `resolveDerivation`'s own index build, so a faithful revert of the hot path passes green |
| 2 | issue | minor | `rowLineage.ts` | `'sourceIndex' in project` throws on a nullish project read where the old path resolved nothing |
| 3 | issue | minor | `derivation.test.ts` | REVERT MAP cites two case names that do not exist in the file |
| 4 | issue | minor | `derivation.ts` (+4 sites) | the container-specific ms figures are duplicated across five docstrings |
| 5 | comment | — | `prioritizationUtils.ts` | `byId` + index are now built for project reads with no rows on the page; untested but almost certainly the right trade |
| 6 | question | — | `rowLineage.ts` | is the prepared arm of the accept-either union covered where the union is declared? |

**Positive note:** bundling the documents and their index into one `ProjectLineageSources` object — rather than taking a second prebuilt-index parameter — makes "index built from a different read" unrepresentable rather than merely discouraged, and `indexes each project read separately` pins it. That is the load-bearing design decision here and it is the right one.

**Overall quality:** high. The production change is small, correct, and structurally sound: one resolution rule (`resolveDerivation` = `resolveDerivationAgainst` + one build), no memoization and therefore no invalidation surface, no call-site churn. Sections A and C of #399 are correctly untouched. I found no defect in the shipped behaviour — all four issues are about test reach, a boundary-tolerance regression that types currently hide, and comment hygiene.

**Needs attention before merge:** finding 1. I verified it empirically in ignored scratch: reverting `hasSupersededSource`/`recordsNoLineage` to `resolveDerivation(raw, projectDocuments)` (the pre-PR code, so the most likely reversion) leaves **195 files / 3330 tests green**, this file's three cases included, because `vi.mock` only intercepts cross-module imports. Also confirmed separately that a direct `resolveDerivation` call registers 0 on the counter. Additionally counting `resolveDerivation` invocations (expected 0 during `collectRows`) closes it without new machinery. Finding 2 is a one-line nullish guard worth taking while nearby.

**Independent confirmation of the perf claim:** my own probe of the post-fix tree at 200 frozen rows / 1000 documents gives a ~517 ms median `collectRows` in this container, consistent with the PR's 615 ms and with the diagnosis that the remainder is `fresherCoherentSelection`/`newestOfType`, not index construction. Leaving that out of this PR is the right call — please open the follow-up (plus the `poolOptions` vitest 4 migration; that deprecation does print on every run here).

**Build / test results:**

| gate | result |
|---|---|
| `mise run build` | ❌ `no tasks defined` — the repo defines no mise tasks (matches the PR description) |
| `npm run build` | ❌ `prebuild` fails with `ts-node: not found`; the root and `voc-datalake` workspaces have no `node_modules` in this container. Environment, not this PR |
| `npx tsc -b --noEmit` (frontend) | ✅ clean |
| `npx vitest run` (full frontend suite) | ✅ **193 files / 3287 tests passed** |
| `npx vitest run src/api/derivation.test.ts src/pages/Prioritization` | ✅ 13 files / 501 tests passed |
| `npx eslint` on the three changed source files | ✅ clean (the two `*.test.ts` files match an ESLint ignore pattern — pre-existing) |

All verification mutants lived in ignored scratch and were removed; the working tree is clean.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GQXBX0GHVMRVHYQHS7AKJJ` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T10:34:15.232Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/404) |
| Duration | 3118.6s |
| Cost | $3.0993 |

voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
perrozzi: Fixed in dac0551 — and thank you for verifying it by actually performing the revert rather than by reading, because the gap was exactly as you describe and I had no way to see it from the mutations I ran.

The `vi.mock` factory now wraps `resolveDerivation` in a second hoisted counter (`resolveDerivationCalls`) alongside `derivationSourceIndex`, keeping `importOriginal` so both return production answers. `indexes a project read once however many rows name it` asserts `resolveDerivationCalls.count === 0` after the existing lineage-state control and the `indexBuilds.count === 1` assertion: every resolution on this page's path must go through `resolveDerivationAgainst`, which is countable whether or not the route back happens to re-index.

Verified by making your revert — `hasSupersededSource`/`recordsNoLineage` back to `resolveDerivation(raw, projectDocuments)`, `collectRows`' `projectLineageSources` left in place — in a scratch copy: the new assertion fails with **11 calls, expected 0**, while `indexBuilds.count === 1` on the line above stays green, reproducing your finding exactly. Restored afterwards; working tree clean.

The REVERT MAP has an entry for it naming that measured outcome, and explicitly naming `resolveDerivationCalls.count` as the catching assertion rather than the construction count — since as you showed, that one is not sensitive to this revert. The file docstring also now explains *why* two counters are needed (`vi.mock` cannot reach a module-local binding), so the next reader does not delete one as redundant.

voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts
perrozzi: Fixed in dac0551, using your proposed shape. `lineageSourcesOf` now short-circuits the nullish case:

```ts
if (project == null) return projectLineageSources([])
return 'sourceIndex' in project ? project : projectLineageSources(project)
```

Written as an early return rather than one expression so the `Array.isArray` narrowing comment above stays about narrowing and the nullish guard carries its own sentence about the totality contract — the two are unrelated reasons and reading as one would invite a "simplification" that drops one of them.

You are right that this is a real regression and not taste: the module docstring promises "no throwing", and the pre-#399 B path absorbed it in `resolveDerivation`'s `projectDocuments = []` default. Added `resolves nothing rather than throwing when the project read is missing entirely` to `rowLineage.test.ts`, over **both** spellings (`undefined` and `null`, since `== null` is what covers them together) and all three exported rules, with a positive control asserting the same selection DOES cross generations against a read that carries its source — so the case pins "resolved against nothing" rather than a classifier that answers `coherent` regardless. Measured: deleting the guard fails it with `TypeError: Cannot use 'in' operator to search for 'sourceIndex' in undefined`. REVERT MAP entry added.

voc-datalake/frontend/src/api/derivation.test.ts
perrozzi: Fixed in dac0551. Both invented names are gone, replaced by the real case titles — `consults the index once per source and holds no state between calls` and `reads what a source is from the given index and never from the wire again` — with each paired to the revert it actually catches (a memo behind `resolveDerivationAgainst`, and the resolver reading the wire instead of the index it was handed). Kept the counting-not-timing rationale and the positive-control note, and corrected the latter to name what the control asserts: the titles the index resolved, not "zero lookups".

Also added the missing entry for `indexes a document list once for every field a source displays`, which was in the file but in no map entry — the same navigational gap in the other direction.

Grepped to confirm the contract holds now: every case name cited in the REVERT MAP appears verbatim in an `it(...)` title in this file.

perrozzi:

Since the last review the author added a second counter (`resolveDerivation` invocations must be 0) to `prioritizationUtils.indexReuse.test.ts`, added both-shapes equivalence cases plus a nullish-read case to `rowLineage.test.ts`, added the `project == null` guard to `lineageSourcesOf`, corrected the `derivation.test.ts` REVERT MAP case names, and de-duplicated the wall-clock figures down to `ProjectLineageSources`. All three Should-fix items and two of the nits are addressed.

### Resolved
- Construction-count blind spot closed: `resolveDerivationCalls.count === 0` catches a revert to `resolveDerivation(raw, projectDocuments)` inside `hosSupersededSource`/`recordsNoLineage` (`prioritizationUtils.indexReuse.test.ts`), with the revert outcome recorded in the REVERT MAP.
- Union contract now pinned where it is declared: `rowLineage.test.ts` compares all three rules list-vs-prepared over a non-trivial fixture and over the empty read, with literal controls first.
- Missing project read no longer throws: `lineageSourcesOf`'s `project == null` short-circuit plus a case covering both `undefined` and `null` for all three rules.
- REVERT MAP in `src/api/derivation.test.ts` now names cases that exist.
- Wall-clock figures consolidated to one docstring; `collectRows`' "one `byId`/index more than before for a project with no visible rows" behaviour change is now stated explicitly.

### Still open
- Prose-to-code ratio: the incremental diff adds roughly 60 more lines of rationale (two-counter explanation stated three times — file docstring, mock docstring, inline at the assertion) than it removes; still a nit, not a blocker.

### New findings
- Nit, `src/pages/Prioritization/rowLineage.ts`: the restored "no throwing" contract covers nullish only. A truthy non-array, non-prepared value (e.g. `{}` from untyped component code) falls into `projectLineageSources(project)` and throws on the `for...of` in `derivationSourceIndex`. Same failure mode as pre-#399 B, so no regression — but the docstring's claim is broader than the guard.
- Nit, `src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts`: only the first case resets `resolveDerivationCalls.count`, so the counter is dead weight in the other two; either reset it in a `beforeEach` or assert it where it matters.

Otherwise the delta is clean — no blockers.

---

voc-datalake/frontend/src/api/derivation.ts
perrozzi: Fixed in dac0551, consolidated as you suggested. `ProjectLineageSources` is now the single owner of the figures, and the other four sites state the invariant with a cross-reference ("one source index per project read, not per call — see `ProjectLineageSources`, issue #399 B").

On which number to quote: I kept both rather than only 562 ms, but labelled the relationship instead of the provenance, since the disagreement you predicted comes from two sites qualifying one number differently — one site saying "1644→615 ms is jsdom under this container; the reviewing machine reported 562 ms before, so wall clock here is ~3× a reviewer's and reproduces nowhere" is what makes the ratio the evidence rather than the absolute. If the house convention is the reviewed figure alone, that is now a one-line edit in one place, which was the point.

Your second remark is the more useful one and I have acted on it: the `DerivationSourceIndex` docstring no longer explains a shared API-layer type through `pages/Prioritization` specifics. It now carries only the normalisation guarantee (every value has been through `displayString`, so an unreadable field is `''` and never `null` — stated as "a distinction consumers turn on" rather than by naming `hasSupersededSource`), the instruction to build via `derivationSourceIndex`, and a pointer to `ProjectLineageSources` as the caller-side type. The prose dependency now runs the same way as the import.

Re-ran `npx tsc -b --noEmit` (clean) and the derivation + Prioritization test files (all green).

voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts
perrozzi: Not considered and rejected — just not thought through, and your reasoning is right. I had carried the resolver seam's argument ("one function delegates to the other") one layer up where it does not hold: `fresherCoherentSelection` reads `sources.documents` for `newestOfType` and `sources.sourceIndex` for the nested classify, so the two arms diverge at three separate reads rather than one, and `collectRows` coverage exercises only the prepared arm without ever comparing it to the list arm.

Added in dac0551, as a new `describe` in `rowLineage.test.ts` right where the union is declared, per your suggested shape — `rule(sel, docs)` deep-equalling `rule(sel, projectLineageSources(docs))` for each of `classifySelectionLineage`, `fresherCoherentSelection` and `rowLineageOf`, over a fixture whose row both crosses generations and is stale so all three rules have something to say. The literal answers are asserted **first** as the positive control, since two arms agreeing on the empty/default answer would satisfy the equalities vacuously. A second case does the same for an empty document list, which is what would catch a prepared arm reading its documents from anywhere other than the read it was given.

Measured in both directions rather than assumed:
* the prepared arm reading `newestOfType` off an empty list while its index stays correct → **caught**;
* `classifySelectionLineage` resolving a prepared read against an empty index → **caught**, 8 failures including the new equivalence case.

Both restored; working tree clean. REVERT MAP entry added naming the divergence class and why it is pinned here rather than in `prioritizationUtils.indexReuse.test.ts`.

perrozzi:
## Iteration 2 — all five action-bearing threads addressed (dac0551)

**Changes**

1. **`prioritizationUtils.indexReuse.test.ts` — the construction-count blind spot (blocking).** The `vi.mock` factory now wraps `resolveDerivation` in a second hoisted counter alongside `derivationSourceIndex`, and `indexes a project read once however many rows name it` asserts it is called **0** times during `collectRows`. Verified against the reviewer's own revert (`hasSupersededSource`/`recordsNoLineage` → `resolveDerivation(raw, projectDocuments)`, `projectLineageSources` left in place): the new assertion fails with **11 calls, expected 0** while `indexBuilds.count === 1` stays green — reproducing the reported blind spot and closing it. REVERT MAP entry names the catching assertion explicitly.
2. **`rowLineage.ts` — nullish project read no longer throws.** `lineageSourcesOf` short-circuits `project == null` to an empty prepared read, restoring the pre-#399 B "resolves nothing" behaviour and the module's documented totality. New case covers both `undefined` and `null` across all three exported rules, with a control proving the answers are "resolved against nothing" rather than a classifier stuck on `coherent`. Deleting the guard fails it with the expected `TypeError`.
3. **`rowLineage.test.ts` — the `ProjectLineageRead` union is now pinned where it is declared.** Two cases assert `rule(sel, docs)` deep-equals `rule(sel, projectLineageSources(docs))` for each rule, over a rich fixture and an empty one, with the literal answers asserted first as the positive control. Both arms mutation-tested: prepared arm reading `newestOfType` off a different list → caught; prepared arm classifying against an empty index → caught (8 failures).
4. **`derivation.test.ts` — REVERT MAP corrected.** The two invented case names are replaced with the real titles, each paired to the revert it catches; the missing entry for `indexes a document list once for every field a source displays` was added. Every name cited now appears verbatim in an `it(...)` title.
5. **Measurements consolidated.** Full figures live only in `ProjectLineageSources`; the other four sites state the invariant and cross-reference it. `DerivationSourceIndex` in the shared API module no longer explains itself through `pages/Prioritization` specifics — the prose dependency now runs the same direction as the import. The per-project (vs. per-row) `byId`/index construction is recorded at `collectRows`.

**Threads:** #3913182850 fixed · #3913182858 fixed · #3913182867 fixed · #3913182873 fixed · #3913182879 answered + covered. No clarification needed. #3913182876 was context-only ("No change requested") — its trade-off is now recorded in the `collectRows` docstring.

**Gates** (run from `voc-datalake/frontend` unless noted)

| gate | result |
|---|---|
| `npm run test` (full frontend suite, repo root) | ✅ **193 files, 3290 passed** (was 3287 — +3 new cases) |
| `npx vitest run src/api/derivation.test.ts src/pages/Prioritization/rowLineage.test.ts src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts` | ✅ 96 passed |
| `npm run typecheck` (`tsc -b --noEmit`) | ✅ clean |
| `npm run lint:frontend` (`eslint . --max-warnings 0`) | ✅ clean |
| `npm run typecheck:tests` | ⚠️ 87 errors, **identical count on the stashed baseline**; none in the four files touched |
| `mise run build` | ⚠️ `no tasks defined` — the repo defines no mise tasks |
| `npx vite build` | ❌ fails **identically on the unmodified tree** (Rollup cannot resolve a dependency whose `dist` is missing; this container blocks npm install scripts). Confirmed by stashing the diff and re-running. Not caused by this change. |

All six mutations were restored; working tree and scratch (`.tmp/`) are clean.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GTYT5QRXRGMX600YDABYC4` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T10:52:19.330Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/404) |
| Duration | 1020.4s |
| Cost | $3.2900 |

voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts
```diff
@@ -134,6 +145,85 @@ export interface RowLineage extends SelectionLineage {
...
+
+/**
+ * The prepared read, building one only for the caller that passed a plain list.
+ *
+ * Narrowed on `'sourceIndex' in project` rather than on `Array.isArray`, which
+ * `tsc` refuses to narrow a `readonly unknown[]` arm out of (its signature answers
+ * `any[]`, and a readonly array is not one) — leaving the list in the prepared
+ * branch's type.
+ *
+ * NULLISH IS AN EMPTY READ, not a throw, which restores this module's "no throwing"
```
perrozzi, line 215: **Type**: issue
**Severity**: minor
**Title**: The restored "no throwing" claim is broader than the guard, which covers nullish only

**Description**: The nullish guard added here is the right fix and I verified it catches the regression I reported (deleting `project == null` fails `resolves nothing rather than throwing when the project read is missing entirely` with the `in`-operator TypeError). But the docstring states the fix as restoring "this module's 'no throwing' contract", and the guard is narrower than that sentence.

A truthy non-array, non-prepared value falls through `project == null`, fails `'sourceIndex' in project`, and reaches `projectLineageSources(project)`, which throws inside `derivationSourceIndex`'s `for...of`. I confirmed it: `classifySelectionLineage([{document_id:'a'}], {} as never)` throws `documents is not iterable`.

That is **not a regression** — the pre-#399 B path threw the same way for the same input, since `resolveDerivation`'s `projectDocuments = []` default only absorbs `undefined`, not `{}`. So this is purely about the claim's scope, not the behaviour, and it does not block. It matters only because a future reader auditing the module docstring's "Pure and total ... no throwing" promise against this guard will find one input where the two still disagree, and the comment currently reads as though the gap is closed.

**Proposed fix**: Either narrow the sentence to what the guard actually covers (nullish, which is the shape a component that has not landed its project detail passes), or widen the guard to a full non-iterable check. Narrowing the prose is the smaller and more honest change: add a clause noting a non-iterable truthy read still throws, unchanged from before this PR, so the module docstring's totality promise is scoped to unreadable *documents* rather than to an unreadable *project read*.

**AI prompt**: In `voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts`, the `lineageSourcesOf` docstring paragraph beginning "NULLISH IS AN EMPTY READ" claims the guard restores the module's "no throwing" contract, but the guard covers only `null`/`undefined`: a truthy non-array, non-prepared value (e.g. `{}`) reaches `projectLineageSources(project)` and throws `documents is not iterable` inside `derivationSourceIndex`'s `for...of`. Do not widen the guard. Instead add one sentence to that paragraph scoping the claim: note that a truthy non-iterable project read still throws, that this is unchanged from the pre-#399 B path (whose `projectDocuments = []` default absorbed only nullish), and that the module docstring's totality promise is about unreadable documents rather than an unreadable project read. Keep the existing `Array.isArray` narrowing comment and the nullish rationale intact.

voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts
```diff
@@ -0,0 +1,232 @@
...
+
+/** State and reason per row id, so a control assertion reads as one object. */
+const lineageByRow = (rows: ReturnType<typeof collectRows>) => Object.fromEntries(
+  rows.map((row) => [row.row_id, `${row.lineage.state}/${row.lineage.reason}`]),
+)
+
+describe('collectRows indexes each project read once', () => {
+  it('indexes a project read once however many rows name it', () => {
+    indexBuilds.count = 0
+    resolveDerivationCalls.count = 0
```
perrozzi, line 175: **Type**: issue
**Severity**: minor
**Title**: `resolveDerivationCalls` is reset and asserted in only the first case, so it is dead weight in the other two

**Description**: `resolveDerivationCalls.count` is zeroed here and asserted at line 200, but the other two cases (`indexes each project read separately`, `builds nothing for a page whose project reads have not landed`) reset only `indexBuilds.count`. The counter therefore keeps accumulating across those cases and is never read again.

No current case is wrong because of it — vitest runs these in declaration order and only the first case asserts the value it reset. But it makes the file order-dependent in a way its docstring does not claim: moving `indexes a project read once however many rows name it` below either sibling, or adding a case above it, silently changes what that assertion sees. The `indexBuilds.count` resets are per-case for exactly this reason, so the two counters are now maintained to different standards.

Worth closing because the second counter is the one that carries the blocking finding from the previous review round, and an order-sensitive assertion is a weaker guarantee than the one the file's docstring describes.

**Proposed fix**: Reset both counters in a `beforeEach` and drop the per-case resets, which removes the ordering dependency for all three cases at once. Optionally assert `resolveDerivationCalls.count === 0` in `indexes each project read separately` too — the "every resolution goes through `resolveDerivationAgainst`" property holds there as well, and a second project read is where a revert routing back through `resolveDerivation` would also show up.

**AI prompt**: In `voc-datalake/frontend/src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts`, `indexBuilds.count` is reset at the top of all three cases but `resolveDerivationCalls.count` is reset only in `indexes a project read once however many rows name it`, leaving the second counter accumulating and making that case's `expect(resolveDerivationCalls.count).toBe(0)` order-dependent. Replace the per-case resets with a single `beforeEach` that zeroes both counters (import `beforeEach` from vitest), removing the now-redundant in-case reset lines. Then add `expect(resolveDerivationCalls.count).toBe(0)` to `indexes each project read separately`, after its existing lineage control and `expect(indexBuilds.count).toBe(2)`, with a short comment noting the "every resolution goes through resolveDerivationAgainst" property holds per project read too. Re-run `npx vitest run src/pages/Prioritization/prioritizationUtils.indexReuse.test.ts` and confirm all three cases still pass.

perrozzi (commented):
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

Follow-up review of the delta `92f3594..dac0551` (6 files, +199/-42). The full PR is read-only context; only delta-touched files were re-read.

## Previous findings — all four resolved

I verified each by performing the revert in an ignored scratch copy rather than by reading the diff.

| finding | status |
|---|---|
| **[blocking]** construction-count mock cannot see `resolveDerivation`'s own index build | **Resolved** in `dac0551`. Reverting `hasSupersededSource`/`recordsNoLineage` to `resolveDerivation(raw, projectDocuments)` with `collectRows`' `projectLineageSources` left in place now fails with `expected 11 to be +0` on the new `resolveDerivationCalls.count` assertion, while `indexBuilds.count === 1` beside it stays green — reproducing my finding and the author's measurement exactly. |
| **[should_fix]** `'sourceIndex' in project` throws on a nullish project read | **Resolved** in `dac0551`. Deleting the `project == null` guard fails `resolves nothing rather than throwing when the project read is missing entirely` with `TypeError: Cannot use 'in' operator to search for 'sourceIndex' in undefined`. The case covers both spellings and all three rules, with a working positive control. |
| **[should_fix]** REVERT MAP names two test cases that do not exist | **Resolved** in `dac0551`. Both invented names are gone. I scripted the check across all three touched test files: every REVERT-MAP-cited case title now appears verbatim in an `it(...)` title. The author also added the previously unmapped `indexes a document list once for every field a source displays`. |
| **[needs_author_input]** is the prepared arm covered where the union is declared? | **Answered and closed** in `dac0551`. The new `the two shapes of project read a rule accepts answer identically` describe block compares all three rules list-vs-prepared, with literal controls asserted first. Verified it bites in both directions I could think of: `fresherCoherentSelection` reading `newestOfType` off an empty list while its index stays correct → caught (11 failures); `lineageSourcesOf` discarding a prepared read's `sourceIndex` → caught (5 failures). |

The wall-clock consolidation also landed: `1644`/`615`/`562` now appear in exactly one file (`rowLineage.ts`, at `ProjectLineageSources`), down from five docstrings, and `DerivationSourceIndex` no longer explains a shared API-layer type through `pages/Prioritization` specifics — the prose dependency now runs with the import direction. The `collectRows` per-project-read behaviour change I recorded as context-only is now stated explicitly in the docstring.

One positive note: fixing the blind spot by counting the *other* side of the seam — `resolveDerivation` must be called zero times on this path — is a better answer than the one I proposed, because it is insensitive to whether a regression happens to re-index. It catches the route back rather than one of its symptoms.

## New findings from the delta

Three, all minor; none blocks merge.

1. **The restored "no throwing" claim is broader than the guard it cites** (`rowLineage.ts:215`) — line comment below.
2. **`resolveDerivationCalls` is reset in only one of three cases** (`prioritizationUtils.indexReuse.test.ts:175`) — line comment below.
3. **Two docstrings reference `sourceFieldIndex`, a symbol this PR renamed away.** `rowLineage.ts:337` and `rowLineage.test.ts:725` both say "`sourceFieldIndex` runs it through `displayString`". That was the real private name in `derivation.ts` before this PR; `92f3594` renamed it to the exported `derivationSourceIndex`, so both references now dangle — `grep -rn sourceFieldIndex src/api/` returns nothing. Raising it here rather than as a line comment because neither line is in this PR's diff, and flagging it because it is the same class of defect the delta just fixed in `derivation.test.ts`'s REVERT MAP: a name that no longer resolves sends the next reader looking for code that is not there.

   **Type**: issue
   **Severity**: minor
   **Title**: Two docstrings still name `sourceFieldIndex`, renamed to `derivationSourceIndex` by this PR

   **Description**: See above. Both are pure prose; no behaviour is affected. (Informed by repository memory: this repo treats docstring symbol references as a navigational contract, which is the same reasoning that made the REVERT MAP name correction worth doing.)

   **Proposed fix**: Replace `sourceFieldIndex` with `derivationSourceIndex` at `rowLineage.ts:337` and `rowLineage.test.ts:725`.

   **AI prompt**: In `voc-datalake/frontend/src/pages/Prioritization/rowLineage.ts` line 337 and `voc-datalake/frontend/src/pages/Prioritization/rowLineage.test.ts` line 725, the docstrings reference `sourceFieldIndex`, which this PR renamed to the exported `derivationSourceIndex` in `src/api/derivation.ts`. Replace both references with `derivationSourceIndex`, keeping the surrounding sentences about `displayString` normalisation intact. Then run `grep -rn sourceFieldIndex voc-datalake/frontend/src` and confirm it returns nothing.

## Build and gate results

Dependencies were not present in this container, so I ran `npm install --ignore-scripts` first.

| gate | result |
|---|---|
| `mise run build` | ⚠️ `no tasks defined` — the repo defines no mise tasks, as the author reported. |
| `npm run build` | ❌ fails at the `prebuild` step: `ts-node: not found` (blocked install scripts). Environmental, not caused by this PR. |
| `npx vitest run` (full frontend suite) | ✅ **193 files, 3290 passed** |
| `npx vitest run src/api/derivation.test.ts src/pages/Prioritization` | ✅ 13 files, 504 passed |
| `npm run typecheck` (`tsc -b --noEmit`) | ✅ clean |
| `npm run lint` (`eslint . --max-warnings 0`) | ✅ clean |
| `npm run typecheck:tests` | ⚠️ **87 errors, all pre-existing.** I confirmed the baseline independently via a detached worktree at `63f25dc`: also exactly **87**, and the one error in a derivation-adjacent file (`DocumentsTab.derivation.test.tsx:40`) is byte-identical in both. No error in any file this PR touches. |

The `test.poolOptions` vitest 4 deprecation the author noted prints on every run; agreed it is out of scope here and worth its own issue, as are `newestOfType`'s O(rows × documents) residual and sections A/C of #399.

**Assessment:** the delta is clean and does exactly what the previous round asked, with each fix measured rather than asserted. Nothing here needs to change before merge.

perrozzi:
> 🤖 **ABCA automated review** — posted by the background agent, not a human.

**Round 2 follow-up** — delta `92f3594..dac0551` only (6 files, +199/-42). Full review: #pullrequestreview-5088829255

## Findings

| type | count |
|---|---|
| issues | 3 (all **minor**) |
| comments | 0 |
| questions | 0 |

No blockers. All three new findings are minor and none makes the merged PR incorrect.

- `rowLineage.ts:215` — the restored "no throwing" claim is broader than the nullish guard (a truthy non-iterable read still throws; **not a regression**, same as pre-PR).
- `prioritizationUtils.indexReuse.test.ts:175` — `resolveDerivationCalls` is reset in only one of three cases, making that assertion order-dependent.
- `rowLineage.ts:337` + `rowLineage.test.ts:725` — two docstrings still name `sourceFieldIndex`, which this PR renamed to `derivationSourceIndex`.

## Previous round: all four findings resolved in `dac0551`

Verified by performing each revert in an ignored scratch copy, not by reading the diff:

1. **[blocking]** construction-count blind spot → the new `resolveDerivationCalls.count === 0` assertion fails with `expected 11 to be +0` under the `resolveDerivation(raw, projectDocuments)` revert, while `indexBuilds.count === 1` stays green. Reproduced exactly.
2. **[should_fix]** nullish project read threw → deleting the `project == null` guard fails the new case with the `in`-operator `TypeError`.
3. **[should_fix]** REVERT MAP cited non-existent cases → scripted across all three touched test files: every cited title now appears verbatim in an `it(...)`.
4. **[needs_author_input]** prepared arm uncovered → answered with a new equivalence block; I confirmed it catches both divergences I could construct (empty-list `newestOfType` → 11 failures; discarded `sourceIndex` → 5 failures).

Wall-clock figures also consolidated from five docstrings to one, and `DerivationSourceIndex` no longer explains a shared API-layer type through `pages/Prioritization` specifics.

**Positive note:** closing the blind spot by counting the *other* side of the seam — `resolveDerivation` must be called zero times on this path — is better than the fix I proposed, because it catches the route back regardless of whether that route happens to re-index.

## Build / test results

Dependencies were absent in this container; installed with `npm install --ignore-scripts` first.

| gate | result |
|---|---|
| `mise run build` | ⚠️ `no tasks defined` (repo defines no mise tasks) |
| `npm run build` | ❌ `prebuild` fails: `ts-node: not found` — environmental, not this PR |
| `npx vitest run` (full suite) | ✅ **193 files / 3290 passed** |
| `npx vitest run` (derivation + Prioritization) | ✅ 13 files / 504 passed |
| `npm run typecheck` | ✅ clean |
| `npm run lint` | ✅ clean |
| `npm run typecheck:tests` | ⚠️ 87 errors — **all pre-existing**; independently confirmed 87 at baseline `63f25dc` via a detached worktree, none in any file this PR touches |

## Assessment

High quality. Each previous-round fix is measured rather than asserted, mutation outcomes are recorded with real numbers, and the REVERT MAP navigational contract now holds. **Nothing needs to change before merge** — the three minor items are polish. The `test.poolOptions` vitest 4 deprecation, the `newestOfType` O(rows × documents) residual, and sections A/C of #399 are all correctly scoped out and worth their own issues.

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GVYBGF2F8S5P1PCE7ZETV3` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T11:12:03.528Z |
| Pull request | [link](https://github.com/aws-samples/sample-voice-of-customer-datalake/pull/404) |
| Duration | 1188.1s |
| Cost | $2.2753 |

perrozzi:

**Why it stopped:** 0 blocking, 0 should_fix, build not regressed — decision=commented; 3 nit; build unknown (tolerated); mergeable [iteration 1, 3 nit]

| | |
|---|---|
| Outcome | **CONVERGED** |
| Iteration | 1 of 5 |
| Loop spend | $14.87 of $100.00 |
| Goal | Implement section B of GitHub issue #399 on a new branch from development. sourceFieldIndex(projectDocuments) is rebuilt inside every resolveDerivation call. Prioritization now calls lineage classifiers in a loop over rows, repeatedly rebuilding the same project-document index; the reviewed 200-… |
| Run | [`01M1GN7JZ02QE90D32Q0B93N7E`](https://d1t2aabk0oo81s.cloudfront.net/runs/01M1GN7JZ02QE90D32Q0B93N7E) |

### Each iteration

|---|---|---|---|---|---|---|
| 0 | 1 blocking, 2 should_fix, 1 nit, 1 needs_author_input, 1 info | — | ❌ | ✅ | 3 | `dac0551` |
| 1 | 3 nit | — | ❌ | ✅ | — | — |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M1GXMC150HPXHCKW5V1VMEMD` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-02T11:22:59.708Z |
| Duration | 84.3s |
| Cost | $0.2943 |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2DXVQYVGCM5YACF263PN5KW` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T17:57:35.837Z |
| Duration | 818.3s |
| Cost | $1.7080 |

PR #405, opened 2026-09-03
Add persistent PRD and PRFAQ versioning

perrozzi:
## Summary
- add persistent automatic PRD/PRFAQ versioning with canonical `Base (vN)` titles across REST, MCP, UI, exports, prioritization, project chat, and roundtable
- scope ChatStream project reads to the canonical Projects Lambda and document the generated wildcard suppression
- refresh only the intentional `VocApiStack` synthesized-template fingerprint

## Validation
- backend: 2,647 passed
- frontend: 2,951 passed; typecheck, lint, and production build passed
- stream: 356 passed; typecheck passed; lint has no errors (one pre-existing warning)
- CDK: TypeScript passed; targeted delegation test passed; full suite 257 passed
- `git diff --check` passed
- deployed API/SSE sweep: 34/34 passed
- live same-title PRD workflow verified stable v1/v2 assignment and deletion behavior
- authenticated browser smoke passed with clean console/network

The runtime feature and infrastructure are already deployed and live-tested from the source commits. This PR integrates only that feature onto current `main` plus its one-line synth-baseline ratchet; it does not merge unrelated development history.


perrozzi:

This PR adds persisted `(vN)` versioning for PRD/PR-FAQ documents: a new `shared/document_versions.py` owns version allocation (transactional counter + document + META write) and an on-read legacy backfill; `projects.py` gains paginated partition reads, a new `POST /projects/{id}/chat-context` route, and stricter `create_document`/`update_document` semantics; the ChatStream Lambda stops reading DynamoDB directly and instead invokes the Projects Lambda; the frontend locks managed titles and suppresses the ordinal fallback. Test coverage is genuinely thorough. However, the review below is **partial** — the diff was truncated (CDK stack changes, MCP tools, exports/prioritization, i18n files, and the rest of `schema.ts` are not visible), and several of my concerns can only be settled by the missing files. The main issues are the blocking write-on-read migration and the forged authorizer context used for the cross-Lambda call.

### Blockers

- **Legacy migration runs (and blocks) on the GET read path.** `get_project` now calls `persist_legacy_document_versions` (`voc-datalake/lambda/api/projects.py`), which per series acquires a lease that polls for up to `LEGACY_MIGRATION_WAIT_SECONDS = 15` with `time.sleep` (`voc-datalake/lambda/shared/document_versions.py`). Series are migrated sequentially in a `for` loop, so a project with several legacy PRD/PRFAQ series can exceed API Gateway's 29s limit, and every concurrent first read pays the wait. `_acquire_legacy_migration` also raises `ServiceError('Document versions are being initialized. Please retry.')` on timeout — i.e. a read of an existing project can now 5xx. Please move the backfill to a one-shot job/Step Function (or bound the *total* migration budget per request and fall back to the in-memory `normalize_document_versions` result without persisting) so reads stay read-only and latency-bounded.
- **The stream forges an API Gateway authorizer context.** `voc-datalake/lambda/stream/src/context/projects-client.ts` builds `requestContext: { authorizer: { claims: { sub: 'chat-stream' } } }` and invokes the Projects Lambda directly. Any per-user authorization or audit attribution the Projects handler derives from `claims` is bypassed for this path, and `/projects/{project_id}/chat-context` returns full document `content` for the selected IDs. Two things need to be true and neither is visible in the diff: (a) the stream handler must verify the authenticated caller actually owns/has access to `projectId` before invoking, and (b) `api_project_chat_context` in `voc-datalake/lambda/api/projects_handler.py` must not be reachable as a public route without the same ownership check (it is registered on the same resolver as the user-facing routes). Please make the caller identity explicit (pass the real `sub`) rather than substituting a synthetic principal.
- **`create_document` breaks non-PRD document types.** `voc-datalake/lambda/api/projects.py` now raises for *any* `document_type != 'custom'`, and the new test asserts `'research'` is rejected too. The stated scope is PRD/PR-FAQ only; `research` and `product_report` creation via the REST API and (likely) the MCP `create_document` tool now fail. Restrict the guard to `VERSIONED_DOCUMENT_TYPES` unless the removal of the other types is intentional and all callers (MCP tools, importer paths — not in the visible diff) were updated.

### Should fix

- **Lint regression:** `voc-datalake/lambda/jobs/document_merger/test/conftest.py` drops `# noqa: F401, F403` from `from jobs.conftest import *`, which ruff will flag.
- **Test shim hides the real write shape:** `voc-datalake/lambda/jobs/conftest.py` implements `transact_write_items` by re-dispatching to `mock_table.put_item`/`update_item` so old assertions keep passing. That means the job tests no longer verify the transaction (condition expressions, counter update, ordering) at all — a malformed `TransactItems` list would still pass. Prefer `moto` (as `shared/test/test_document_versions.py` already does) or assert on the transaction items directly.
- **`get_project_chat_context` only minimizes documents.** It projects documents down to `_CHAT_CONTEXT_DOCUMENT_FIELDS` but forwards the entire `project` item and all `personas` items verbatim across the Lambda boundary (`voc-datalake/lambda/api/projects.py`). Apply the same allow-list to project/persona fields; the test only proves `large_internal_field` is dropped from documents.
- **IAM/CDK verification needed (not in visible diff):** the new code requires `dynamodb:PutItem`, `UpdateItem`, `ConditionCheckItem` (transactions) on the projects table plus access to the new `DOCUMENT_VERSIONS#PROJECT#*` partition, and the stream role needs `lambda:InvokeFunction` scoped to exactly the Projects function. Please confirm these are least-privilege grants and that the added policy statements keep the Projects/ChatStream roles under the 20KB policy limit.
- **Dependency pinning inconsistency:** `voc-datalake/lambda/stream/package.json` adds `@aws-sdk/client-lambda` at an exact `3.1014.0` while every sibling AWS SDK dep uses `^3.750.0` (lockfile resolves others to `3.1101.0`). That pulls a second set of `@smithy/*` versions into the bundle; use the same range as the other clients.
- **`persist_versioned_document` retry budget:** the `observed_version is None` branch calls `persist_legacy_document_versions(...)` then `continue`, consuming one of only 4 attempts without backoff; four such iterations fall through to `raise RuntimeError('Document version allocation loop ended unexpectedly')`, surfacing as a 500. Give the bootstrap path its own (non-consuming) step or convert the terminal case into a `ServiceError`.
- **`document_id` format changed** from `prd_<timestamp>` to `prd_<sha256[:20]}` (`_document_id`). Any consumer that sorts or parses document ids chronologically (exports, prioritization, Kiro steering files — not visible here) will silently change behavior; worth a grep and, if found, a regression test.

### Nits

1. The managed-type predicate is duplicated inline in `ProjectDetail.tsx` (`handleSaveDocument`) and `ProjectModals.tsx`; extract a single `isVersionManaged(doc)` helper (and prefer it over the `d.version === undefined` check in `DocumentsTab.tsx`, which keys off a different signal).
2. `Wizards.tsx` — the merge title placeholder is still a hardcoded English literal on a line you're already touching; move it to an i18n key for all locales.
3. `voc-datalake/lambda/api/test/test_chat_context_bounds_lockstep.py` regex-scrapes `stream/src/schema.ts`; a shared generated constants file would be less brittle than a source-text assertion.
4. `delete_document` passes `document.get('sk')` (possibly `None`) straight into the delete key; the projected read guarantees `sk` in practice, but a `str(...) or raise` keeps the boundary honest.
5. `base_title`/`version` were added to the `ProjectDocument` interface in `frontend/src/api/types.ts` but not to any Zod schema, so the new server fields cross the boundary unvalidated; backend-authored error strings (e.g. the 413 / "cannot be renamed" messages) also reach users un-localized.

---

perrozzi:
Superseded by #406, which is based directly on development and contains the complete managed-document and prototype versioning implementation without the conflicting main-only history. Continuing automated review iteration on #406.

PR #406, opened 2026-09-03
Add persistent managed document versioning

perrozzi:
## Summary
- add durable automatic `(vN)` identity for PRD, PR/FAQ, and prototype documents
- make allocation, replay, legacy migration, deletion tombstones, project cleanup, and prototype S3 winner binding race-safe
- carry canonical titles through REST, MCP, stream chat, prioritization, UI labels, and exports while excluding executable prototype HTML from generic edit/export/grounding paths
- address the complete prior automated review, including read-only GET normalization, real Cognito subject forwarding, payload allowlists, Zod wire normalization, exact SDK pinning, Data Explorer denial, title validation, and direct-job crash recovery

## Validation
- backend: 4,276 passed
- frontend: 3,309 passed; production build and typecheck passed
- stream: 427 passed; typecheck passed; lint has only the existing non-literal fixture warning
- CDK: 332 passed; typecheck passed
- Ruff delta: 113 to 111, no new findings
- `git diff --check` passed
- final semantic and security reviews approved

## History
This supersedes #405. That PR was retargeted from `main` to `development`, but its old head retained unrelated main-only history and became conflicting. This branch is based directly on `development` and contains only the managed-document versioning commits.


perrozzi:
## Prior automated review disposition

The previous review comment on #405 was deleted when that PR was retargeted. Every reported item was re-evaluated and resolved or measured on this development-based replacement:

### Blockers
1. **Write-on-read legacy migration:** fixed. `GET /projects/{id}` normalizes copies only; durable migration occurs on managed mutations with bounded leases and retry-safe high-water updates.
2. **Synthetic stream identity:** fixed. Stream-to-project delegation requires and forwards the real Cognito `sub`; missing/blank subjects fail closed.
3. **Over-broad project chat payloads:** fixed. Project, persona, and document fields use explicit allowlists; prototype HTML, signed URLs, and internal fields never cross the boundary.

### Should-fix findings
4. **Prototype versioning:** implemented through the same allocator as PRD/PRFAQ with deterministic job-derived IDs, durable allocation tombstones, canonical `Title (vN)`, legacy migration, and replay-before-Bedrock/S3.
5. **Prototype S3/DynamoDB split-brain:** fixed. Conditional S3 create reads the conflict winner and persists its ETag/version identity in the document transaction.
6. **Data Explorer prototype mutation bypass:** fixed at both layers: handler rejection and an explicit IAM deny for `prototypes/*` writes/deletes.
7. **Replay after deletion:** fixed. Allocation history survives document deletion; delayed same-job replay is rejected rather than resurrecting a later version.
8. **Allocation retry budget:** fixed. Legacy bootstrap/migration has its own bounded phase and does not consume the four allocation transaction attempts.
9. **Frontend runtime trust boundary:** fixed with lenient Zod project-detail normalization plus render guards.
10. **Managed-document detection drift:** fixed with shared frontend lineage predicates and canonical backend metadata.
11. **Merge wizard localization/manual version hints:** fixed across all eight locale catalogs; managed titles are normalized before job creation.
12. **Bundled SDK drift:** `@aws-sdk/client-lambda` remains exactly pinned to `3.1014.0`.
13. **Grounding reference overclaim:** fixed. `referenced_documents` contains only textual documents actually supplied to the model; prototypes remain metadata-only.

### Nits / measured claims
14. **Generic document creation:** intentionally remains custom-only. PRD/PRFAQ/prototype creation uses dedicated workflows because managed identity and prototype S3 storage have different contracts; route/tool schemas now state this explicitly.
15. **IAM policy size:** measured by the existing CDK guard against the real 10,240-character inline-policy limit. New permissions remain domain-scoped; the synth baseline inventories the four exact self-invoke ARNs.
16. **Transaction shim, document-ID consumers, regex lockstep, and delete sort-key claims:** inspected and covered by focused tests; no production defect was found. The transaction fake mirrors all write actions, deterministic IDs are treated as opaque throughout consumers, Python/TypeScript bounds remain lockstep-tested, and deletion resolves the stored sort key before its conditional transaction.
17. **Merger conftest `noqa`:** not added because Ruff reports it as unused `RUF100`; the file is valid without it.
18. **Python lint debt:** compared as a base/head delta rather than sweeping unrelated repository debt. Result: 113 → 111, with no new findings.

Additional review rounds found and fixed late legacy arrivals after counter creation, allocation-vs-migration lease arbitration, atomic document/count deletion, retained project-deletion tombstones across Python and stream writers, exact merge output validation, committed replay routing in Step Functions, direct-job overlap exclusion, and hard-crash recovery through a budget-preserving self-redelivery baton.

Final local semantic and security reviews are approved.

perrozzi:

This PR adds durable `(vN)` versioning for PRD/PR-FAQ/prototype documents, tightens every projects-table write with conditional expressions and transactions, introduces a project tombstone + fenced deletion flow, adds a bounded `POST /projects/{id}/chat-context` route, makes prototype S3 writes conditional/replay-safe, and adds Zod normalization plus canonical-title plumbing on the frontend. **This review is partial: the diff was truncated.** The two central new modules (`shared/document_versions.py`, `shared/project_writes.py`), the stream package, and all CDK/IAM changes are not visible, so the version allocator itself, the response-payload contracts on the stream side, and IAM grants for the new transactional writes are unreviewed. The visible parts are generally careful and well-tested, but a few behavioral regressions stand out.

### Blockers

- `voc-datalake/lambda/api/projects.py` — `delete_document` can make a document permanently undeletable. The transaction requires `document_count >= :one` and decrements with a bare `document_count = document_count - :one`. Counts can already be drifted *low* on existing data (the previous `put_item` → unconditional `update_item` sequence created the document even when the increment failed, e.g. missing attribute), and in that state every delete attempt now cancels and surfaces as `ServiceError('… document count is inconsistent')` with no recovery path for the user. Prefer a self-healing form (`if_not_exists(document_count, :zero)` and clamp at 0, or drop the row and repair the counter separately) so a bad counter never blocks deleting a real document.

### Should fix

- `voc-datalake/frontend/src/components/DocumentExportMenu/DocumentExportMenu.tsx` + `pages/ProjectDetail/DocumentsTab.tsx` — legacy *inline* prototypes (JSON spec or pre-migration HTML, no `prototype_url`) now lose the export menu *and* the edit button, leaving only Delete. The comment says "prototype artifacts own their open/download controls in PrototypeView", but that only holds for URL-backed prototypes; for the inline legacy rows there is no replacement affordance. Either keep the text export for `document_type === 'prototype' && !prototype_url`, or confirm PrototypeView renders a download path for inline content.
- `voc-datalake/frontend/src/api/projectDetailSchema.ts` — `document_type` is a closed `z.enum`, and `normalizeProjectDetail` *drops* any row that fails to parse. A backend that adds a seventh document type (as it has repeatedly: `product_report`, `prototype`) will make those documents vanish from the UI until the bundle is redeployed. The old code deliberately tolerated this (`styles[type] ?? 'bg-amber-…'`). Suggest a `.catch('custom')`-style fallback or a passthrough "unknown" branch rather than discarding the row.
- `voc-datalake/lambda/jobs/document_generator/handler.py` — `put_object(..., IfNoneMatch='*')` requires a recent botocore; on an older runtime-provided SDK this raises `ParamValidationError` and breaks *all* prototype generation. The requirements pin isn't in the visible diff — please confirm the Lambda bundles a boto3/botocore that supports conditional `PutObject`, ideally with a test asserting the parameter is accepted.
- `voc-datalake/lambda/api/projects.py` — `delete_project` retains the META tombstone "forever" with no TTL attribute, so deleted projects accumulate indefinitely in the projects table (and `list_projects` now pays a per-item `is_project_tombstone` filter). Consider a TTL on the tombstone (long enough to outlive delayed events) rather than unbounded retention.
- `voc-datalake/lambda/api/projects.py` — `get_project` now does a fully paginated `ConsistentRead=True` scan of the partition on every GET, and `get_project_chat_context` calls it (reading all document bodies) before filtering down to the allowlist. That doubles RCUs and raises tail latency on the hottest read path; the "legacy assignment can collide" justification only needs consistency for the managed-mutation paths, not for the read-only GET.
- `voc-datalake/frontend/src/pages/ProjectDetail/Wizards.tsx` — the new `wizards.remixDocuments`, `wizards.remixInstructions`, `wizards.remixInstructionsPlaceholder`, `wizards.selectAtLeast2`, and `wizards.submitRemixDocuments` keys are not in the visible locale diffs (only `newDocTitlePlaceholder` changed). Please confirm all 8 locales gained every new key; note the ⚠️ prefix was also dropped from the warning and should be reflected in the strings.
- `voc-datalake/lambda/api/projects.py` — `delete_project` repair path: if the fence `update_item` fails (META absent) *and* the recovery `put_item` also fails because META reappeared concurrently without the marker, the final `update_item` (condition `attribute_exists(#deleting)`) throws and the caller gets a 500 after all child rows were already deleted. Worth handling that branch explicitly.

### Nits

- `delete_document` uses `getattr(projects_table, 'name', None)` while `delete_persona` uses the new `projects_table_name()` helper — pick one.
- `voc-datalake/lambda/jobs/conftest.py` contains a `# ponytail:` marker; use a conventional `TODO:`/`NOTE:` so it's greppable.
- `_bounded_chat_context_response` keys off `event.get('path')` (REST payload only — silently no-ops on payload v2 `rawPath`) and re-serializes the entire response just to measure it.
- `test_chat_context_bounds_lockstep.py` pins `MAX_CHAT_CONTEXT_SELECTED_DOCUMENTS` to the stream's `MAX_PERSONAS_DOCS_ARRAY`; the names describe different concepts, so a future intentional divergence will look like a bug.
- `pages/ProjectDetail/DocumentLabels.tsx` uses the namespace-prefixed key `projectDetail:documents.ordinal` even though the injected `t` is already bound to that namespace — inconsistent with the surrounding call sites and makes the component harder to reuse.

---

perrozzi:
## Automated review round 1 disposition

Addressed in `2a19b6d4` (pushing immediately after this disposition):

- **Blocker — stale-low `document_count`: fixed.** Document deletion now derives the post-delete count from the complete partition snapshot, atomically sets that value, CAS-checks the observed META count, and retries boundedly on concurrent count changes. Existing zero/missing/high drift no longer makes a real document undeletable.
- **Unknown document types disappearing: fixed.** The runtime schema accepts any non-empty type string; `KnownProjectDocumentType` remains the explicit product/export inventory, so future rows stay visible without becoming exportable/scorable implicitly.
- **Chat-context full-body read: fixed.** The route now uses a paginated metadata projection and fetches bodies only for at most 20 selected non-prototype documents. It no longer calls `get_project`, signs URLs, or materializes unselected/prototype content.
- **Project deletion recovery race: fixed.** Fence acquisition retries update-existing vs insert-tombstone races and aborts before any destructive sweep unless the winning META row is atomically marked.
- **Payload-v2 `rawPath`: fixed** as defense in depth.
- **Bound lockstep quality: fixed.** Tests now prove both values are applied in the Python validator and TypeScript Zod schema, not merely equal.
- **Table-name helper consistency: fixed** while preserving the API-facing configuration error.

Measured/design dispositions:

- **Inline prototypes:** no generic edit/export was restored. URL-backed HTML, inline HTML, and inline JSON already have dedicated sandboxed open/download controls. Managed prototype content is intentionally immutable through generic CRUD; exposing executable HTML/JSON through Markdown/PDF/Kiro export or in-place editing would violate that contract.
- **`IfNoneMatch`:** no package change. The Lambda uses Python 3.14 with the runtime-provided matched boto3/botocore pair; packaging intentionally strips both together rather than risking a mismatched partial override. Focused tests pin the conditional request and conflict-winner read.
- **Tombstone TTL:** deliberately rejected. A finite TTL would permit project-ID reuse and allow delayed work from a deleted incarnation to attach to a later project. The retained row is unindexed and contains only fence metadata.
- **Remix locale keys:** false positive from truncated diff. All required keys exist in all eight locales and the locale-parity suite passed.
- **`ponytail:` comment:** retained because workspace policy requires that exact marker for deliberate test-fake ceilings.
- **Ordinal namespace prefix:** retained because it is load-bearing for the repository i18n extraction gate; rendering and all locale tests pass.

Validation: backend 4,284; frontend 3,310; stream 427; CDK 332; build/lint/typechecks passed; Ruff has no new findings; `git diff --check` passed.

perrozzi:

Since the last review the author reworked `delete_document`'s counter handling (observed-count condition + bounded retry), extracted `_start_project_deletion` with its own retry loop, rewrote `get_project_chat_context` to use a projected consistent query plus per-selected-document `get_item` instead of `get_project`, opened `document_type` to a plain string on the wire (with a new `KnownProjectDocumentType` inventory and re-anchored Kiro lockstep regex), added the `rawPath` fallback for the response bound, and added regression tests for each. **This follow-up is partial: the full-context diff was truncated**, so `shared/document_versions.py`, `shared/project_writes.py`, the stream package, locale JSONs beyond the shown hunks, and CDK/IAM remain unreviewed.

### Resolved
- `delete_document` no longer wedges on a drifted counter: it writes `document_count = :remaining` under an observed-count/`attribute_not_exists` condition, with a regression test for the stale-zero case.
- `delete_project`'s fence/repair race now retries instead of 500-ing after child rows were deleted, with a test that exercises the lost-`put_item` branch.
- `projectDetailSchema.ts` no longer drops rows with an unknown `document_type`.
- `_bounded_chat_context_response` handles payload-v2 `rawPath`.
- `get_project_chat_context` no longer reads every document body before filtering to the allowlist.

### Still open
- Inline legacy prototypes (no `prototype_url`) still lose both edit and export with no confirmed replacement affordance in `PrototypeView`.
- `put_object(..., IfNoneMatch='*')` botocore pin still not visible in the diff.
- Project META tombstone still retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- New `wizards.*` keys still not visible in any of the 8 locale files.

### New findings
- **`voc-datalake/lambda/api/projects.py`** — two public entry points lost tracing: `@tracer.capture_method` was moved onto the new private helpers (`_query_project_chat_items`, `_start_project_deletion`) and `get_project_chat_context` / `delete_project` are now undecorated. Restore the decorator on the public functions (repo powertools pattern).
- **`voc-datalake/frontend/src/api/types.ts`** — `KnownProjectDocumentType` is exported but referenced nowhere in the app; with `document_type: string` there is no longer any tsc pressure to extend it, so `test_kiro_exportable_types_lockstep.py` now parses an inventory nothing enforces. Wire it into the gates that need it (e.g. `Record<KnownProjectDocumentType, string>` for the badge styles, or the export allowlist) so a seventh backend type still breaks something.
- **`voc-datalake/frontend/src/pages/ProjectDetail/MergeWizard.test.tsx`** — the test mocks `react-i18next` with `t: (key) => \`${namespace}:${key}\``, so it proves the labels go through keys but cannot detect that the five new `wizards.*` keys are absent from the locale JSONs. Add a check against `public/locales/*/projectDetail.json` (or extend the existing locale-parity test) rather than a stubbed `t`.
- **`voc-datalake/frontend/src/api/projectDetailSchema.test.ts`** — the new case uses bare `test(...)` while the file imports `{ describe, expect, it, vi }`, and it sits outside the `describe` block after two blank lines. Use `it` inside `describe('normalizeProjectDetail')` so the file does not depend on vitest globals being enabled.
- **`voc-datalake/lambda/api/projects.py`** — the `delete_document` retry re-runs `_query_partition_items` (consistent, full partition) *and* `persist_legacy_document_versions` / `preserve_versioned_document_allocation` on every attempt, up to 4×. Please confirm both version-module writes are idempotent under repeat invocation (that module is not in the visible diff) and consider capping the re-read rather than repeating the whole path.
- Nit: `_query_project_chat_items` projects `#status` and `created_at`, which no `_CHAT_CONTEXT_*` allowlist consumes — drop them to keep the projection minimal.

---

perrozzi:
## Automated review round 2 disposition

Addressed in `8c6fcc3a` (pushing immediately after this comment):

- Restored `@tracer.capture_method` on public `get_project_chat_context` and `delete_project`; helper spans remain traced too.
- Made `KnownProjectDocumentType` compiler-enforced through an exhaustive `Record<KnownProjectDocumentType, string>` badge-style map plus a runtime `Object.hasOwn` type guard. Unknown wire types still use the fallback.
- Moved the unknown-type schema regression inside `describe('normalizeProjectDetail')` and changed it to imported `it`, removing reliance on Vitest globals.
- Dropped the unused projected project status field. Retained `created_at` intentionally because `normalize_document_versions` uses it for deterministic ordering of legacy rows before chat summaries.
- Documented the bounded delete retry invariant: `persist_legacy_document_versions` is assignment/lease-idempotent and `preserve_versioned_document_allocation` is conditional winner-checked; both are covered by focused replay/concurrency tests.

Still-open items are measured design dispositions, not unresolved defects:

- Inline HTML/JSON prototypes already have dedicated sandboxed open/download controls. Generic edit/Markdown/PDF/Kiro export remains intentionally disabled for managed executable artifacts.
- Python 3.14 uses the runtime-provided matched boto3/botocore pair; packaging strips both together. `IfNoneMatch` and conflict-winner behavior are pinned by handler tests.
- The retained tombstone intentionally has no TTL: expiry would allow project-ID reuse and cross-incarnation delayed-write corruption.
- Full `get_project` remains a complete read because the detail UI needs every body and deterministic legacy normalization. The hot chat-context path is now projected and selected-body-only.
- All remix keys exist in all eight locale files. `localeParity.test.ts` recursively compares each namespace key set against English; the full frontend suite passed.
- The namespace-qualified ordinal key remains required by the repository extraction gate.

Validation: backend focused 226; frontend focused 49; frontend typecheck/lint; Ruff delta has no new findings; `git diff --check` passed.

perrozzi:

**What changed:** small follow-up commit — `projectDetailSchema.test.ts`'s stray `test(...)` was moved back inside `describe` as `it`, `KnownProjectDocumentType` is now wired into `DOCUMENT_TYPE_STYLES` in the new `DocumentLabels.tsx`, `@tracer.capture_method` was restored on `get_project_chat_context` / `delete_project`, `#status` was dropped from the chat-context projection, and comments were added explaining the `created_at` projection and the intentional replay in `delete_document`'s retry. All four actionable new findings from the last round are addressed; the older Still-open list is untouched. **This review is partial: the full-context diff was truncated**, so `shared/document_versions.py`, `shared/project_writes.py`, the stream package, most locale JSON content, and CDK/IAM remain unreviewed.

### Resolved
- Public tracing restored: `get_project_chat_context` and `delete_project` are decorated again; the private helpers keep their own spans.
- `KnownProjectDocumentType` now has real tsc pressure — `Record<KnownProjectDocumentType, string>` in `DocumentLabels.tsx` breaks if a member is added, so the Kiro lockstep inventory is enforced.
- `projectDetailSchema.test.ts`'s new case is inside `describe` and uses `it`, no longer relying on vitest globals.
- `_query_project_chat_items` no longer projects the unused `#status`; the `created_at` retention is now justified in a comment.
- `delete_document` retry idempotency is documented with a pointer to the replay tests (accepted as answered).

### Still open
- Inline legacy prototypes (no `prototype_url`) lose both edit and export with no confirmed replacement affordance in `PrototypeView`.
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- `wizards.remixDocuments` / `remixInstructions` / `remixInstructionsPlaceholder` / `selectAtLeast2` / `submitRemixDocuments` still not confirmed present in the 8 locale files (`MergeWizard.test.tsx` stubs `t`, so it cannot catch a missing key).

### New findings
- **`voc-datalake/lambda/api/projects.py`** — please confirm `is_project_tombstone` (in `shared/project_writes.py`, not visible) keys only off `PROJECT_DELETION_ATTRIBUTE`. Now that `#status` is no longer projected, a tombstone detected via `status == 'deleted'` would be invisible to `get_project_chat_context`, which would then serve a deleted project's summary.
- Nit: **`voc-datalake/frontend/src/api/projectDetailSchema.test.ts`** — new case body is under-indented and the file ends with no trailing newline (`\ No newline at end of file`); likely an `eol-last` lint hit.
- Nit: **`voc-datalake/frontend/src/pages/ProjectDetail/DocumentLabels.tsx`** — `Object.hasOwn` needs an ES2022 lib/target; a plain `type in DOCUMENT_TYPE_STYLES` guard avoids the dependency if the tsconfig target is older.

---

perrozzi:
## Automated review round 3 disposition

Addressed in `6cec8bad` (pushing immediately after this comment):

- Restored `status` to the chat metadata projection because `is_project_tombstone` intentionally recognizes either `deletion_started_at` or historical `status in {deleting, deleted}` rows. The projection comment now documents both `status` and `created_at` as load-bearing.
- Normalized the unknown-type schema test indentation, kept it inside the imported-`it` suite, and restored the trailing newline.
- Replaced `Object.hasOwn` with `Object.prototype.hasOwnProperty.call`, avoiding an ES2022 dependency and prototype-chain false positives.
- Added an explicit real-catalog test that reads all eight `public/locales/*/projectDetail.json` files and asserts every remix wizard key is present.
- Added explicit inline prototype affordance coverage: legacy inline HTML exposes Open in new tab + Download .html; inline JSON exposes Download .json. Both continue to exclude generic edit and prose export by design.

The repeated design items remain intentionally unchanged for the reasons already recorded: managed executable prototype immutability, Python 3.14 matched runtime SDK, permanent unindexed deletion fence, complete detail-route reads, and namespace-qualified extraction key.

Validation: backend focused 195; frontend focused 54; frontend typecheck/lint; `git diff --check` passed.

perrozzi:

**What changed:** a small polish commit — `projectDetailSchema.test.ts` re-indented and given a trailing newline, `Object.hasOwn` swapped for `Object.prototype.hasOwnProperty.call`, a new `localeParity` case pinning the five `wizards.remix*` keys across every catalog, new `DocumentsTab` assertions for inline prototype open/download affordances (HTML and JSON), and `#status` restored to the chat-context projection with a comment. All three nits and two of the older still-open items from the last round are addressed. **This review is partial: the full-context diff was truncated**, so `shared/document_versions.py`, `shared/project_writes.py`, the stream package, and CDK/IAM remain unreviewed.

### Resolved
- `wizards.remix*` locale keys are now enforced by a real parity test over all catalogs, not just a stubbed-`t` render test.
- Inline legacy prototypes (no `prototype_url`) are confirmed to keep `open in new tab` / `download .html` and, for JSON specs, `download .json` — the lost-export concern is closed by tests.
- `Object.hasOwn` ES2022 dependency removed.
- `projectDetailSchema.test.ts` indentation and missing trailing newline fixed.
- `#status` back in `_query_project_chat_items`, so a status-only historical tombstone stays visible to `is_project_tombstone`.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.

### New findings
- Nit: **`voc-datalake/lambda/api/test/test_projects.py`** — the `#status` restoration fixes a real leak path (a `status == 'deleted'` META with no `PROJECT_DELETION_ATTRIBUTE` would otherwise be served by `get_project_chat_context`), but there's no regression test for it; `test_retained_project_tombstone_is_not_read_as_a_project` only covers `get_project`. A one-line `TestProjectChatContext` case with a status-only tombstone would lock the projection in.
- Nit: **`voc-datalake/frontend/src/i18n/localeParity.test.ts`** — `[REFERENCE_LOCALE, ...locales]` likely double-checks `en` if `locales` already includes it; harmless but the "eight real catalogs" wording then undercounts.

---

perrozzi:
## Automated review round 4 disposition

Addressed in `6fd9f8b0` (pushing immediately after this comment):

- Added the requested chat-context regression for a historical tombstone that has `status: deleted` but no `deletion_started_at`; the route returns not found and performs no selected-body reads.
- The locale-loop nit is a false positive: `locales` is built with `.filter(entry => entry.name !== REFERENCE_LOCALE)`, so `[REFERENCE_LOCALE, ...locales]` covers each of the eight catalogs exactly once.

The three repeated still-open items remain explicit maintainer decisions with regression evidence: matched Python 3.14 runtime SDK for `IfNoneMatch`, permanent unindexed tombstone to prevent cross-incarnation resurrection, and complete detail-route reads while chat uses projected selected-only reads.

Validation: backend focused 195 plus status-only tombstone test; frontend focused 54; frontend typecheck/lint; `git diff --check` passed.

perrozzi:

**What changed:** a single new test, `test_chat_context_rejects_a_status_only_historical_tombstone` in `voc-datalake/lambda/api/test/test_projects.py`, which is exactly the regression case I asked for last round; it asserts `NotFoundError` and that no `get_item` content fetch happens for a status-only META tombstone, so the `#status` projection in `_query_project_chat_items` is now pinned. **This review is partial: the full-context diff was truncated**, so `shared/document_versions.py`, `shared/project_writes.py`, the stream package, and CDK/IAM remain unreviewed.

### Resolved
- Status-only historical tombstone leak through `get_project_chat_context` now has a regression test.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- Nit: `localeParity.test.ts` still iterates `[REFERENCE_LOCALE, ...locales]`, likely double-checking `en`.

No new problems introduced — the incremental change is clean.

---

perrozzi:
## Final automated-review disposition at `6fd9f8b0`

The latest bot review reports no new problems and the incremental change is clean. Maintainer disposition of the remaining repeated items:

- **Runtime SDK / `IfNoneMatch`: accepted design.** Python 3.14 supplies a matched runtime boto3/botocore pair; the layer strips both together. Vendoring only one half would be less safe. Conditional put and conflict-winner behavior are tested.
- **Permanent tombstone: accepted durability design.** The row is unindexed and content-free. A TTL would permit project-ID reuse and let delayed work from a deleted incarnation attach to a later project.
- **Complete `get_project` read: accepted detail-route design.** The detail UI needs all bodies and deterministic legacy normalization. The hot chat route has already been split to projected metadata plus selected-only body reads.
- **Locale loop: false positive.** `locales` explicitly filters `entry.name !== REFERENCE_LOCALE`; `[REFERENCE_LOCALE, ...locales]` visits each of `de,en,es,fr,ja,ko,pt,zh` exactly once, and the test asserts that exact set.

There are no unresolved blocker/should-fix code defects on the current head.

perrozzi:
## Post-completion tombstone review fix at `f9350ba6`

A final completion-hook review found one real lifecycle mismatch after the prior review had converged: project reads recognized both `deletion_started_at` and historical `status=deleting|deleted` tombstones, while write fences rejected only the marker, and the public update path could assign reserved lifecycle statuses.

Fixed at this head by:
- centralizing a status-aware DynamoDB writable predicate and applying it to shared/versioned/custom Python write transactions;
- using the equivalent condition for stream custom-document creation;
- limiting public project status updates to `active|archived`;
- adding real moto regressions for marker-only and both status-only tombstones, plus direct transaction-shape coverage.

Validation at this exact local commit before push: backend 4,305/4,305; frontend 3,312/3,312; stream 427/427; CDK 332/332; all typechecks and ESLint passed; Ruff delta is zero; fresh semantic and security review are clean.

Please evaluate automated review and CI against exact head `f9350ba67c0440d5a7bb4d9aa2f9da99322c64a5`.

perrozzi:

**What changed:** the deletion fence was widened from `attribute_not_exists(deletion_started_at)` to a shared `PROJECT_WRITABLE_CONDITION` that also rejects `status in {deleting, deleted}`, centralized in `shared/project_writes.py` and threaded through `projects.py`, `shared/document_versions.py` and the stream `executeCreateDocument` transaction; `update_project` now validates `status` against `{active, archived}`; new parametrized tests cover all three tombstone shapes plus the persona-delete transaction. None of the previously listed open items were touched. **This review is partial: the full-context diff was truncated.**

### Resolved
- Status-only tombstones (no `deletion_started_at`) can no longer slip past the write fence on any of the audited write paths, and `_project_accepts_writes` now uses `is_project_tombstone` consistently.
- `update_project` can no longer write a reserved `deleting`/`deleted` status through the public route.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- Nit: `localeParity.test.ts` still iterates `[REFERENCE_LOCALE, ...locales]`.

### New findings
- `voc-datalake/lambda/stream/src/tools/update-document.ts` now restates the fence by hand (`:deletingStatus`/`:deletedStatus`) with no lockstep test tying it to `PROJECT_WRITABLE_CONDITION` in `shared/project_writes.py`. The repo already uses this pattern (`test_chat_context_bounds_lockstep.py`); without it the two fences will drift the next time a terminal status is added. The new `update-document.test.ts` assertion pins the TS side only.
- `voc-datalake/lambda/shared/project_writes.py` defines `PROJECT_TERMINAL_STATUSES` but `PROJECT_WRITABLE_ATTRIBUTE_VALUES` re-hardcodes `'deleting'`/`'deleted'` as separate literals, so adding a terminal status updates the read-side guard without updating the condition values. Derive the values from the frozenset (or assert equality in a test).
- `voc-datalake/lambda/shared/document_versions.py` drops the `PROJECT_DELETION_ATTRIBUTE` import; please confirm no other reference to that name survives elsewhere in the file (a missed reference would be a runtime `NameError` on an untested branch rather than an import failure). I could not verify the rest of the file — it is outside the incremental diff and the full diff was truncated.

Otherwise the incremental change is clean: no unused `ExpressionAttributeNames`/`Values` at any of the call sites, and the OR group in the condition is parenthesized so the `AND {count_condition}` / `AND persona_count >= :one` suffixes bind correctly.

---

perrozzi:
## Automated review round 5 disposition

Addressed the two new drift findings in `ab693862`:
- `PROJECT_DELETING_STATUS` and `PROJECT_DELETED_STATUS` now feed both `PROJECT_TERMINAL_STATUSES` and the DynamoDB placeholder values, removing duplicate literals.
- Added `test_project_writes_lockstep.py`, which parses the stream create transaction and requires its marker, status attribute, terminal values, and predicate references to match the Python contract.

The requested import verification is also complete: `PROJECT_DELETION_ATTRIBUTE` has zero references in `shared/document_versions.py`, and the full backend suite imported/exercised the module successfully.

Focused follow-up validation: 45/45 Python tests, Ruff clean on the two changed files, and `git diff --check` clean. The previously dispositioned botocore/runtime pairing, permanent tombstone, bounded detail-read tradeoff, and locale test nit remain unchanged accepted designs/non-blocking items.

Please review exact head `ab693862`.

perrozzi:

**What changed:** `shared/project_writes.py` now names `PROJECT_DELETING_STATUS`/`PROJECT_DELETED_STATUS` and builds both `PROJECT_TERMINAL_STATUSES` and `PROJECT_WRITABLE_ATTRIBUTE_VALUES` from them, and a new `shared/test/test_project_writes_lockstep.py` pins the stream `executeCreateDocument` fence to the Python tombstone contract. Two of my three new findings are addressed; earlier open items are untouched. **This review is partial: the full-context diff was truncated.**

### Resolved
- The stream fence in `lambda/stream/src/tools/update-document.ts` is now lockstep-tested against `PROJECT_TERMINAL_STATUSES` and `PROJECT_DELETION_ATTRIBUTE`, matching the `test_chat_context_bounds_lockstep.py` pattern.
- `PROJECT_WRITABLE_ATTRIBUTE_VALUES` no longer re-hardcodes the status literals.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- Unverified (outside both diffs): no surviving `PROJECT_DELETION_ATTRIBUTE` reference in `shared/document_versions.py` after the import was dropped.
- Nit: `localeParity.test.ts` still iterates `[REFERENCE_LOCALE, ...locales]`.

### New findings
- Should-fix — `voc-datalake/lambda/shared/project_writes.py`: the frozenset and the values dict share constants, but a third terminal status still requires three coordinated edits (frozenset, values dict, and the hand-written `PROJECT_WRITABLE_CONDITION` placeholder list). The new test only detects Python↔stream drift, not Python-internal drift: add `set(PROJECT_WRITABLE_ATTRIBUTE_VALUES.values()) == set(PROJECT_TERMINAL_STATUSES)` plus a check that every placeholder appears in `PROJECT_WRITABLE_CONDITION`.
- Nit — `voc-datalake/lambda/shared/test/test_project_writes_lockstep.py`: `source.split('export async function executeCreateDocument', 1)[1]` scans to end-of-file, so `:*Status` placeholders in any function declared after `executeCreateDocument` are folded into the equality assertion. Bound the slice to the next `export ` (or `^}` ) to keep the assertion about this function only.

---

perrozzi:
## Automated review round 6 disposition

Addressed both follow-ups in `50ddc53c`:
- the Python-side lockstep test now requires terminal-status values to equal `PROJECT_TERMINAL_STATUSES` and every configured placeholder to appear in `PROJECT_WRITABLE_CONDITION`;
- the stream parser is bounded to the `executeCreateDocument` export rather than scanning the remainder of the file.

Validation: 2/2 focused lockstep tests, Ruff clean, and `git diff --check` clean. The removed import remains verified by zero-match source search plus the previously completed 4,305-test backend run.

Please review exact head `50ddc53c`.

perrozzi:

**What changed:** Only `voc-datalake/lambda/shared/test/test_project_writes_lockstep.py` changed: a new `test_python_write_fence_matches_python_tombstone_contract` pins `PROJECT_WRITABLE_ATTRIBUTE_VALUES` against `PROJECT_TERMINAL_STATUSES` and asserts each placeholder appears in `PROJECT_WRITABLE_CONDITION`, and the stream-source slice is now bounded by a regex lookahead to the next `export `/EOF. Both of my previous new findings are addressed; the older open items are untouched. **This review is partial: the full-context diff was truncated.**

### Resolved
- Python-internal drift is now detected: a third terminal status can no longer be added to the frozenset without the values dict and `PROJECT_WRITABLE_CONDITION` following.
- The stream fence assertion no longer folds in `:*Status` placeholders from functions declared after `executeCreateDocument`.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- Unverified (outside both diffs): no surviving `PROJECT_DELETION_ATTRIBUTE` reference in `shared/document_versions.py` after the import was dropped.
- Nit: `localeParity.test.ts` still iterates `[REFERENCE_LOCALE, ...locales]`.

### New findings
- Nit — `voc-datalake/lambda/shared/test/test_project_writes_lockstep.py`: `placeholder in PROJECT_WRITABLE_CONDITION` is a bare substring check, so a placeholder that is a prefix of another (`:deleted_status` inside a hypothetical `:deleted_status_v2`) would pass while unused, and nothing asserts each placeholder is compared with `<>`. Matching `rf'#status <> {re.escape(placeholder)}\b'`-style per-placeholder patterns would close both gaps; low value given the current two-status set.

---

perrozzi:
## Automated review round 7 disposition

Accepted the final low-value test nit in `126cfc3a`: Python placeholder checks now require an exact word-bounded `#status <> <placeholder>` comparison rather than a bare substring.

Validation: 2/2 lockstep tests, Ruff clean, and `git diff --check` clean. Please review exact head `126cfc3a`.

perrozzi:

**What changed:** The only new change is in `voc-datalake/lambda/shared/test/test_project_writes_lockstep.py`, where the bare `placeholder in PROJECT_WRITABLE_CONDITION` substring check is now a per-placeholder `re.search(rf'#status <> {re.escape(placeholder)}\b', ...)`. That closes the one nit from my previous review; the older open items are untouched. **This review is partial: the full-context diff was truncated.**

### Resolved
- Each writable-fence placeholder is now asserted to appear as an actual `#status <> :placeholder` comparison, so a prefix-colliding or unused placeholder can no longer pass.

### Still open
- `put_object(..., IfNoneMatch='*')` botocore pin not visible.
- Project META tombstone retained with no TTL.
- `get_project` still does a fully paginated `ConsistentRead=True` partition read on the hottest read path.
- Unverified (outside both diffs): no surviving `PROJECT_DELETION_ATTRIBUTE` reference in `shared/document_versions.py` after the import was dropped.
- Nit: `localeParity.test.ts` still iterates `[REFERENCE_LOCALE, ...locales]`.

No new findings — the incremental change is a clean, strictly tighter test assertion.

---

perrozzi:
## Final post-completion review disposition at `126cfc3a`

The latest automated review reports **no new findings** and confirms the incremental assertion is clean.

Maintainer disposition of repeated older items remains unchanged:
- runtime-provided matched boto3/botocore is deliberate; pinning only botocore would create an unsupported pair;
- permanent project tombstones are required to prevent project-ID reuse and delayed-write corruption;
- the full project detail read is retained for detail UI while chat uses bounded projected reads;
- the locale parity duplication is a non-blocking test nit;
- zero surviving `PROJECT_DELETION_ATTRIBUTE` references in `shared/document_versions.py` was independently verified by source search and the full 4,305-test backend run.

Review is converged on exact head `126cfc3a7fd4f8c835edab0173e47e1439a51d1c`.

PR #410, opened 2026-09-08
fix(manual-import): key CSV rows by content, not the raw CSV id

perrozzi:
## Problem

Uploading two different CSV files through **Scrapers → CSV upload**, each with 400 rows, produced only 400 records in the feedback table. Both uploads succeeded, both archived to S3, and both responses reported 400 rows queued — so nothing surfaced the loss.

The cause is that a CSV row's identity was the raw `id` column value:

- `_parse_csv_to_items` passes an explicit `id` / `review_id` through unchanged as the SQS message `id`.
- Every CSV upload is enqueued with the constant `source_platform='manual_import'`.
- The processor derives **both** its Powertools idempotency key and its DynamoDB item key from `source_platform + source_id`.

So row `1` of *any* CSV resolves to the same feedback item as row `1` of *every other* CSV. Filename, upload UUID, S3 key, `source_channel`, date and text take no part in identity. Any two files numbered `1..N` overwrite or deduplicate against each other, and when the uploads overlap in time the surviving set is an arbitrary mixture of both.

This is easy to hit, because numbering rows from 1 is the obvious thing to do when exporting a CSV.

## Fix

Identify a row by a versioned fingerprint over its normalized content:

```
sha256(['csv-row-v1', id, text, rating, date, author, title, url, source])[:32]
```

Behaviour:

| Case | Before | After |
|---|---|---|
| Different rows, same `id`, different files | collapse to one | both kept |
| Different rows, same `id`, one file | second skipped | both kept |
| Exact same row re-uploaded | deduplicated | deduplicated (unchanged) |
| Same file, rows reordered | re-keyed by position when `id` absent | stable |
| Byte-identical rows | skipped as `duplicate id` | skipped as `duplicate row` |

Two details that matter:

- The fingerprint uses the **raw** date column, computed before the missing-date `now()` default is applied. Including the generated timestamp would give a re-uploaded row a fresh identity every time and break idempotency.
- Row index is **not** included, so reordering a file does not re-key it.

## Scope

Deliberately confined to the CSV boundary:

- `generate_deterministic_id` in the processor is **untouched** — re-keying there would re-key every plugin source.
- `source_platform`, `source_channel` and `ingestion_method` are unchanged, so provenance and dashboards are unaffected.
- No new SQS message fields, so the shared message schema is unaffected.
- No frontend change. The existing success wording already says "rows queued for processing", which remains accurate.

`csv-row-v1` is a persisted compatibility contract: field order and normalization are frozen once released, and changing the recipe re-keys prior uploads.

## Verification

Against the two 400-row corpora that triggered this:

| Check | Result |
|---|---|
| Each file parsed alone | 400 rows → 400 unique ids, 0 warnings |
| Both files combined | 800 ids → **800 unique, 0 collisions** |
| Re-parsing either file | reproduces its ids exactly |

Tests:

- `test_manual_import_csv_upload.py` — 22/22 pass, with new cases for cross-file id reuse, row reordering, and exact-duplicate skipping.
- Full `lambda/api/test/` suite — 2601/2601 pass.
- **Mutation-checked**: reverting identity to the raw CSV id fails exactly 4 tests, including the cross-file collision case. The new tests discriminate rather than merely pass.
- Ruff findings on the touched file go 23 → 22; no new lint debt.

## Migration note

Because identity changes, rows imported by the old code carry different keys than the same rows would today. An environment already holding collided records should have that set removed before re-importing, otherwise the old and new keys coexist.


perrozzi:

This PR replaces the CSV row identity (previously the raw `id`/`review_id` column, falling back to `sha1(text|row_index)`) with a versioned content fingerprint `sha256(['csv-row-v1', id, text, rating, date, author, title, url, source])[:32]`, fixing the real cross-file collision where row `1` of every CSV mapped to the same feedback item under `source_platform='manual_import'`. The diagnosis is correct, the change is well scoped to the CSV boundary, and the versioned prefix + pinned golden hash in tests is the right way to freeze a persisted contract. One regression is introduced, however: removing the row index from the fallback path makes the handler drop legitimate repeated rows in id-less CSVs.

### Blockers

- `voc-datalake/lambda/api/manual_import_handler.py` — id-less CSVs with repeated content now lose rows. Previously the fallback id was `sha1(f'{text[:200]}|{idx}')`, so `text\ngood\ngood\n` produced two items; now both rows fingerprint identically (no `id`, no `date`, no `author`) and the second is dropped with `duplicate row — skipped`. Short repeated free-text answers ("Good", "OK", "n/a") in a survey export are exactly the common case, so this re-introduces the silent record-loss class the PR is fixing. Either keep `idx` in the fingerprint when no distinguishing column is present, or only dedupe when the row carries an explicit `id`. Either way, add a regression test (e.g. `test_keeps_repeated_identical_rows_without_id`) — the current suite has no case for duplicate content in an id-less file, so this behavior change is untested.

### Should fix

- The original CSV `id` is now discarded entirely: it feeds the hash but is not emitted in the item, so nothing in DynamoDB carries the customer's own row identifier (only the archived raw CSV at `s3_raw_uri` retains it). Support workflows like "find the record for review_id 4711" become impossible. Consider carrying it in an additive, optional field rather than dropping it — the "no new SQS message fields" constraint is a self-imposed one and worth trading here.
- Identity now includes mutable content (`text`, `rating`, `title`, `author`, `url`). Re-uploading a corrected file — fixing a typo in one row, back-filling a rating — creates a *second* record instead of updating the existing one, since the processor keys DynamoDB off `source_platform + source_id`. That's the inverse failure mode of the bug being fixed and isn't covered by the PR's before/after table; it should at least be documented as a deliberate trade-off, ideally with a test asserting the intended outcome.
- `source` falls back to `default_source`, which comes from the request, so the same file uploaded twice with a different source selection in the UI now yields two full sets of records. Worth calling out or excluding `source` from the fingerprint.
- The migration note ("an environment already holding collided records should have that set removed before re-importing") exists only in the PR description. Since old-keyed rows persist after deploy, this belongs in the repo (CHANGELOG / manual-import docs) where an operator will actually find it.

### Nits

- `import csv as _csv`, `io`, `hashlib` remain function-local; while pre-existing, this is a good moment to hoist them to module scope per the repo's module-level-client convention.
- `voc-datalake/lambda/api/test/test_manual_import_csv_upload.py`: annotate `'c29620c3acab1a13e1b27dd4a18aaabe'` with a comment stating it pins the frozen `csv-row-v1` recipe, so a future reader doesn't "fix" it by recomputing.
- Consider lifting `'csv-row-v1'` and the field order into a module-level constant with the contract documented next to it, rather than inline in the `row_id` closure.
- `test_keeps_different_rows_that_reuse_an_explicit_id` varies text, date *and* url simultaneously; varying only one field per case would pin down which inputs are load-bearing.
- Warning strings (`duplicate row — skipped`, `empty text — skipped`) are surfaced to users but not i18n'd — pre-existing, but the frontend i18n rule arguably applies if these are rendered verbatim.

---

perrozzi:
Thanks — the blocker was real and is fixed in `65f7b4fe`, along with all four "should fix" items and four of the five nits. Dispositions below.

## Blocker — fixed

**id-less CSVs with repeated content lose rows.** Correct, and the survey-export framing is the right one: `text\ngood\ngood\n` collapsed to a single row, which is the same silent-loss class the PR exists to remove.

The recipe now carries the row's 1-based position, but **only for rows that have no `id` of their own**:

| Row has | Identified by | Reorder the file | Repeated identical rows |
|---|---|---|---|
| an `id` column | id + content | stable | exact duplicate → skipped |
| no `id` column | content + position | re-keyed | **kept separate** |

A row with an id has a stable anchor, so position is deliberately excluded and reordering does not re-key it. A row without one has nothing else to distinguish it, so position is the only honest answer — and position is stable for an unchanged file, so re-upload remains idempotent. `source_id` and `row_index` occupy separate slots in the field list, so an explicit id of `"3"` cannot collide with the third row of an id-less file.

Tests added: `test_keeps_repeated_identical_rows_without_id` (3 rows in, 3 distinct ids out, no warnings) and `test_repeated_rows_without_id_keep_their_ids_on_re_upload`.

## Should fix — all four addressed

**Original CSV `id` discarded.** Agreed, and worse than the review states: before this PR `source_id` on the item held that value, so the change was an outright provenance regression rather than a missing nicety. The row's id now travels as `csv_row_id` on the message and is persisted on the item. You were also right that "no new SQS fields" was a self-imposed constraint not worth keeping — though while acting on it I found the constraint was already broken: `IngestMessage` sets `extra="forbid"` and does **not** declare `ingestion_method`, which manual-import has been sending for some time. It goes unnoticed only because the processor's bundle copies `processor/` and `shared/` but not `plugins/_shared/`, so `safe_validate_message` is never importable and validation is permanently disabled in the deployed function (`Could not import validation schemas - validation disabled` appears in its live logs). I added `csv_row_id` to the model so this field is correct whichever way that gets resolved, but the latent mismatch is wider than this PR and I have not touched it here.

**Mutable content in identity → corrected re-upload creates a second record.** Accepted as a deliberate trade-off and now documented in the CHANGELOG rather than left implicit, with `test_edited_row_gets_a_new_id_rather_than_updating_in_place` asserting the intended outcome. Worth naming the underlying limit: an update path needs an identity that is stable while content changes, which means a key the uploader controls and which is unique across files — a dataset/file namespace plus the row's id. The modal does not currently send anything of the sort (`default_source` is user-editable and reused), so that is a `csv-row-v2` design with a frontend change attached, not something to slip in here.

**`source` falling back to `default_source`.** Good catch, and I took the exclusion. Identity now uses only a `source` column present in the file; the request-level default is no longer an input, so re-uploading one file under a different label no longer imports a second copy. Both halves are pinned: `test_request_default_source_does_not_change_row_ids` and `test_row_source_column_does_change_row_ids`.

**Migration note only in the PR description.** Moved into `CHANGELOG.md` under Fixed, phrased for someone deploying over an existing stack, and covering both the delete-before-re-import step and the "don't upload a corrected file over an old one" consequence.

## Nits — four taken, one declined

- `csv`, `io`, `hashlib` hoisted to module scope. Taken.
- Golden hash now carries a comment stating it pins the frozen `csv-row-v1` recipe and must not be recomputed. Taken, and reinforced with `test_row_id_recipe_field_order_is_frozen`.
- `csv-row-v1` and the field order lifted to module-level `CSV_ROW_ID_VERSION` / `CSV_ROW_ID_FIELDS` with the contract documented beside them. Taken. `_csv_row_id` is now module-level and raises `KeyError` on a partial call, so a future field addition cannot silently key some rows differently.
- Multi-field collision test split into one case per varied field (text, date, rating). Taken.
- **i18n on the warning strings — declined here.** These are API-layer strings assembled server-side with row numbers and column values, and no server response text in this repo is localized today; doing it properly means deciding where server messages get translated and threading a locale through the request, which is a broader change than this fix should carry.

## Verification

- 33 CSV upload tests, 84 processor + schema tests, 4321/4322 backend. The single failure is `test_competing_allocations_receive_unique_versions`, a pre-existing concurrency flake unrelated to this change — it passes 3/3 in isolation on both this branch and the base.
- Ruff over `lambda plugins scripts`: 510 → 511. The one addition is `UP045` on the `Optional[str]` annotation I added, chosen to match the 15 sibling fields of that model rather than be the lone `str | None` in the block.
- Against the two 400-row files that triggered the bug: 400 unique ids each, 800 unique across both, 0 collisions, and re-parsing either reproduces its ids exactly.
- Not verified locally: CDK and frontend suites. This commit touches no TypeScript, no routes and no IAM, and the worktree has no `node_modules`. The deployment will exercise a real synth.

Identity of a given row changes again in this commit. Nothing is deployed yet, so `csv-row-v1` is still unclaimed and the tag is unchanged.

perrozzi:

Since my last review, the fallback path now folds a 1-based `row_index` into the fingerprint only when the row has no `id` column (with two regression tests for repeated id-less rows and re-upload stability), the customer's row identifier is carried through as an optional `csv_row_id` on the item → SQS message → processor item → `IngestMessage` schema, `default_source` was removed from the fingerprint in favour of the row's own `source` column, the imports were hoisted to module scope, the recipe was lifted into documented `CSV_ROW_ID_VERSION`/`CSV_ROW_ID_FIELDS` constants with a pinned-hash comment, and the migration note plus the edited-row trade-off are now in the CHANGELOG. All previous findings are addressed.

### Resolved
- Blocker: id-less CSVs with repeated content keep one record per row (`row_index` in the recipe + `test_keeps_repeated_identical_rows_without_id`, `test_repeated_rows_without_id_keep_their_ids_on_re_upload`).
- Original CSV `id` is no longer discarded — persisted as `csv_row_id` end-to-end, with tests in both the API and processor suites and an optional, sanitized, length-bounded schema field.
- Mutable-content identity trade-off documented in `test_edited_row_gets_a_new_id_rather_than_updating_in_place` and the CHANGELOG.
- `default_source` excluded from the fingerprint; both directions pinned (`test_request_default_source_does_not_change_row_ids`, `test_row_source_column_does_change_row_ids`).
- Migration note now lives in `CHANGELOG.md` rather than only the PR description.
- Nits: module-level `csv`/`io`/`hashlib`, pinned-hash comment, recipe constants, one-field-per-case tests.

### New findings
- `voc-datalake/lambda/api/manual_import_handler.py` (should fix): `csv_row_id` copies the raw column value with no length bound, while `IngestMessage.csv_row_id` caps it at `MAX_ID_LENGTH`. A CSV whose identifiers exceed that cap now fails validation on a field that is purely informational — the derived `id` is unaffected. Truncate (or drop with a warning) at the CSV boundary so an over-long customer identifier can't reject an otherwise importable row.
- `voc-datalake/lambda/processor/handler.py` (nit): the comment asserts `csv_row_id` is "omitted via None-stripping when absent"; that stripping isn't visible in this diff, so the guarantee rests entirely on `test_omits_csv_row_id_when_not_sent`. Worth pointing the comment at the helper that actually strips, so a refactor of that helper doesn't silently start writing null attributes.
- Mixed files (an `id` column populated for some rows, blank for others) key the blank rows by position, so reordering such a file re-keys just those rows and re-imports them. Narrower than the trade-offs already documented, but the CHANGELOG's "reordering is safe" framing doesn't cover it — one clause would.

---

perrozzi:
All three follow-up findings addressed in `afc56887`.

**`csv_row_id` unbounded at the CSV boundary while the schema caps it.** Fixed, and the asymmetry you identified is the reason it mattered: validation refuses a message *whole*, so an over-long identifier would have cost the entire row for the sake of a field that only exists to make that row findable. An over-long value is now dropped with a warning and the row imports.

Dropped rather than truncated, on the grounds that a truncated identifier is worse than none: an operator searching for the full value finds nothing, and has no way to tell that what is stored is a prefix. Two supporting points are pinned by tests:

- the fingerprint still consumes the id at **full** length, so two rows differing only past the bound remain distinct records (`test_over_long_identifier_still_keys_the_row_at_full_length`) — dropping the informational copy must not weaken identity;
- the boundary is exact, not approximate (`test_keeps_an_identifier_exactly_at_the_length_bound` at 256, dropped at 257), matching what the model accepts and refuses.

Since this makes the number a second copy of one the API Lambda cannot import, it comes with `test_csv_row_id_bound_lockstep.py`, following the existing `test_visual_selection_bound_lockstep.py` pattern — literals read as source text, so the assertion cannot be satisfied by whatever either module resolves at import time, and it does not need the plugin package's own imports. The assertion is `<=` rather than `==` deliberately: dropping *earlier* than the schema would refuse costs traceability but never rows, so it stays allowed, while the direction that costs rows is pinned. Raising the handler's copy to 512 fails it; I ran that mutation rather than assuming it.

**Processor comment asserting invisible None-stripping.** Fair — the guarantee was resting on a test rather than on anything a reader of that block could see. The comment now names the step that performs the stripping and states what starts happening if that step is refactored away.

**Mixed files not covered by the CHANGELOG's reordering framing.** Correct, and the one clause is in. `test_mixed_file_keys_only_the_id_less_rows_by_position` pins both halves: in a file where only some rows carry an `id`, reordering re-keys the rows that don't and leaves the identified ones untouched.

## Verification

- 4329/4329 backend. The concurrency test that flaked in the previous round passed here too, consistent with it being load-dependent rather than related to this change.
- Ruff over `lambda plugins scripts` unchanged from the last round at 510 → 511, still the single `UP045` on the `Optional[str]` annotation matching its 15 sibling fields.
- Lockstep mutation-checked in both directions, as above.

perrozzi:

Since my last review the only functional change is at the CSV boundary: an `id` column value longer than the new `MAX_CSV_ROW_ID_LENGTH = 256` is now dropped (with a warning) instead of being carried into a message the schema would reject, backed by three unit tests plus a source-text lockstep test against `plugins/_shared/schemas.py`'s `MAX_ID_LENGTH`; the processor comment now names the None-stripping comprehension, and the CHANGELOG gained the reorder clause for id-less rows plus a `test_mixed_file_keys_only_the_id_less_rows_by_position` regression test. All three previous findings are addressed.

### Resolved
- Over-long `csv_row_id` no longer rejects an importable row — dropped with a warning at the CSV boundary, identity still computed at full length (`test_imports_the_row_but_drops_an_over_long_identifier`, `test_over_long_identifier_still_keys_the_row_at_full_length`).
- Processor comment now points at the "Remove None values" comprehension that actually strips the field, so a refactor there is flagged rather than silent.
- Mixed id/id-less files documented in the CHANGELOG and pinned by a test, so the "reordering is safe" framing is no longer overstated.

### New findings
- `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py` (nit): the lockstep is source-text-only, so it holds the two numbers together but not the two *semantics* — it can't catch `IngestMessage.csv_row_id` being pointed at a different constant (e.g. `MAX_URL_LENGTH`) while `MAX_ID_LENGTH` stays 256. A one-line assertion that `csv_row_id` is declared with `max_length=MAX_ID_LENGTH` in the schema source would close that gap.
- `voc-datalake/lambda/api/manual_import_handler.py` (nit): the drop warning is emitted per row, so a file whose whole `id` column is over-long produces one warning per row in the API response. Pre-existing pattern for other warnings, so only worth a cap if response size is ever a concern.

---

perrozzi:
First nit taken in `012195c4`; second declined, because the cap it asks about already exists.

**Lockstep held the numbers but not the semantics.** Good catch — the assertion was about a constant, not about the field it is supposed to describe, so repointing `csv_row_id` at `MAX_URL_LENGTH` would have left it passing about a number nothing reads any more. The lockstep now also asserts the field is declared as `Optional[str] = Field(None, max_length=MAX_ID_LENGTH)`. I ran that exact mutation: repointing at `MAX_URL_LENGTH` while leaving every number untouched fails the new assertion and nothing else.

**Per-row drop warning.** Declining, because the response is already bounded: `csv_upload` serialises `warnings[:20]`, so a file whose entire `id` column is over-long returns 20 warnings, not one per row. The list is unbounded in memory during the parse, but it is one short string per skipped row against a body already capped at 10 MB and 50,000 rows, so it is not the dominant term. Capping earlier would also change which warnings a user sees for mixed files — the first 20 problems is a more useful selection than a per-category quota, and it is the behaviour the other warnings already have. Happy to revisit if the warning list ever becomes a response-size concern in its own right.

That leaves the PR with no open findings from any round. Summary of what changed across the four commits:

| Round | Prompted by | Change |
|---|---|---|
| `a269bacb` | — | Content fingerprint replaces the raw CSV `id` as row identity |
| `65f7b4fe` | blocker + 4 should-fix + 4 nits | `row_index` for id-less rows; `csv_row_id` provenance; `default_source` excluded; recipe constants; CHANGELOG migration note |
| `afc56887` | 3 findings | Bound the carried identifier, with a lockstep against the schema |
| `012195c4` | 1 nit | Lockstep also pins which constant the schema field uses |

Verification at this head: 4329/4329 backend, ruff 510 → 511 (the single `Optional[str]`), and against the two 400-row files that triggered the original report — 400 unique ids each, 800 across both, 0 collisions, re-parse reproduces ids exactly.

CDK and frontend suites remain unverified locally: this branch touches no TypeScript, no routes and no IAM, and the worktree has no `node_modules`. I am deploying this head to a demo environment next, which exercises a real synth, and will report the outcome along with the live import result.

perrozzi:

Since my last review the only change is the addition of `test_the_schema_field_is_still_bounded_by_the_constant_compared_above` to `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py`, which pins `csv_row_id`'s declaration to `max_length=MAX_ID_LENGTH` in the schema source. This closes the one substantive nit from the previous review; I verified the regex matches the current declaration in `plugins/_shared/schemas.py`.

### Resolved
- The lockstep now checks semantics as well as the two numbers: re-pointing `csv_row_id` at another bound (or a literal) fails the new assertion instead of passing silently.

### New findings
- `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py` (nit): `SCHEMA_FIELD_PATTERN` requires the closing paren immediately after `MAX_ID_LENGTH`, so adding a harmless kwarg (`description=...`) or a trailing comma to the field breaks the test even though the bound is unchanged. Dropping the trailing `\s*\)` (or allowing `[^)]*\)`) would keep the guarantee without the false positive — and it fails loudly rather than silently, so this is optional.

The remaining previous nit (one drop-warning per row) is unchanged and still acceptable as a pre-existing pattern. Otherwise clean.

---

perrozzi:
Taken in `5a60bc18` — the pattern now ends at a word boundary after `MAX_ID_LENGTH` instead of requiring the closing paren, with the reason recorded next to it so it isn't tightened again later.

Checked both directions rather than assuming: adding `description='x',` to the field keeps all four assertions green, and repointing the bound at `MAX_URL_LENGTH` still fails the semantic one. So the false positive is gone and the guarantee is intact.

No open findings remain. Moving on to deployment and the live import verification; I'll report both here.

perrozzi:

Since my last review the only change is relaxing `SCHEMA_FIELD_PATTERN` in `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py` to end at `MAX_ID_LENGTH\b` instead of requiring the closing paren, with a comment explaining why. That was the one nit outstanding; it's addressed.

### Resolved
- The lockstep no longer false-fails when a harmless `description=`/trailing comma is added to the `csv_row_id` field, while still pinning the bound to `MAX_ID_LENGTH`.

### New findings
- `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py` (nit, optional): `\b` still admits an arithmetic wrapper (`max_length=MAX_ID_LENGTH // 2`), which would change the effective bound while the pattern passes. Narrow residual gap, well below the cost of tightening it further.

Otherwise clean — no changes to handler, schema, or processor logic in this increment.

---

perrozzi:
Closed in `03f9f5fd` after all — the cost turned out to be one character class rather than a rewrite, so it was cheaper than leaving it open.

`[,)]` is the complete set of things that may legally follow a keyword argument's value, so it accepts the `description=`/trailing-comma case the previous round deliberately made room for, and rejects an expression over the constant. Verified all three mutations: `description='x',` passes, `max_length=MAX_ID_LENGTH // 2` fails, repointing at `MAX_URL_LENGTH` fails.

Worth stating why that one mattered despite being rated optional: `// 2` is the one variant that defeats **both** existing assertions at once — the effective bound becomes 128 while every literal in both files still reads 256, so nothing else in the suite would have noticed the handler carrying identifiers the schema refuses.

No open findings. Deployment and live verification next.

perrozzi:

Only change since my last review is tightening `SCHEMA_FIELD_PATTERN` to `MAX_ID_LENGTH\s*[,)]` with an explanatory comment — exactly the residual nit I flagged. Verified it still matches the current declaration `Field(None, max_length=MAX_ID_LENGTH)` while rejecting expression wrappers.

### Resolved
- `voc-datalake/lambda/api/test/test_csv_row_id_bound_lockstep.py`: `[,)]` now rejects `max_length=MAX_ID_LENGTH // 2`, closing the last gap where the bound could move while both files' numbers stayed unchanged.

No new findings — the increment is test-only and clean; handler, schema, and processor logic are untouched.

---

perrozzi:
## Deployed and verified against the data that triggered this

This head is deployed to an internal demo environment and the collided import has been repaired there. Reporting it because the interesting evidence is not what the tests say — it is what the pipeline did with the two real files.

### Deployment

`cdk diff` came back code-only: `Code.S3Key` on the Lambdas, no IAM, resource or configuration change. Three stacks moved rather than one, which is worth noting for anyone reviewing the blast radius:

- **Api** — every API Lambda re-keys, because the api asset stages the whole `api/` directory, so touching one handler changes the hash for all of them.
- **Processing** — the processor, for the persisted `csv_row_id`.
- **Ingestion** — the plugin bundles, because they carry `plugins/_shared/`, which is where the schema field was added.

Synth ran with real container bundling, not assembly-only. After deploying, I pulled the published Lambda zips back down and confirmed the code actually running is this code rather than trusting `UPDATE_COMPLETE`: the manual-import function contains `CSV_ROW_ID_VERSION = 'csv-row-v1'` and `MAX_CSV_ROW_ID_LENGTH = 256`, and the processor contains the `csv_row_id` assignment.

### The repair

The environment held 400 records from the original collided import. I snapshotted all 400 to disk first (with a restore script, so the delete was not a one-way door), then deleted exactly the snapshotted keys — not "whatever is in the partition" — so anything written in between would have been left alone rather than destroyed unrecorded.

The snapshot is also the clearest statement of the bug. Of the 400 surviving records:

| Source file | Records retained |
|---|---|
| `consumer_reviews_cloud_individual.csv` | 366 |
| `consumer_reviews_agentic_tools.csv` | 34 |

Not 400 and 0. An arbitrary mixture, because the two uploads raced per colliding id. Every record carried a `source_id` in `1..400` and none carried `csv_row_id`.

### Re-import on the fixed code

Both files uploaded sequentially, each waiting for the queue to fully drain first — which makes the second upload the strictest case, since the first file's records are already stored when it arrives.

| File | HTTP | `imported_count` | Warnings | Drain |
|---|---|---|---|---|
| cloud_individual | 200 | 400 | 0 | 91s |
| agentic_tools | 200 | 400 | 0 | 20s |

### Result

| Check | Result |
|---|---|
| Records stored | **800** |
| Distinct item keys | **800** |
| Per file | **400 + 400** |
| Records carrying `csv_row_id` | 800 |
| Distinct `csv_row_id` values | **400** |
| Processing queue / DLQ | empty / empty |
| `GET /feedback?source=manual_import` | `total: 800`, `is_partial_window: false` |
| Dashboard "Total Feedback" | **800** |

The two middle rows are the fix stated precisely: **800 distinct records holding only 400 distinct customer identifiers.** The `id` columns still collide two-to-one, exactly as before — they are simply no longer what identity is made of.

The dashboard number is worth having independently of the store count, because a correct table and a windowed query that under-reports it are different failures, and only one of them is visible to a user.

Environment specifics are deliberately omitted here. Cleaned up after: the throwaway Cognito user used for the authenticated uploads is deleted, and the snapshot remains on the operator's machine only.

perrozzi:
## Pre-merge gates — every suite now run at `03f9f5fd`

Earlier rounds only had the backend suite behind them; the TypeScript side was unverified because the worktree had no `node_modules`. All of it has now been run individually rather than through `npm run check`, since that chains with `&&` and the first failure would hide everything after it.

| Gate | Result |
|---|---|
| `lint:frontend` (ESLint, `--max-warnings 0`) | **clean** |
| `lint:stream` (ESLint) | 0 errors, 1 warning — pre-existing, in `indexes.test.ts` |
| `lint:python` (ruff over `lambda plugins scripts`) | **511** vs 510 on base |
| `typecheck` (frontend) | **pass** |
| `typecheck:cdk` | **pass** |
| `typecheck:stream` | **pass** |
| Frontend vitest | **3312 passed** (196 files) |
| CDK vitest | **336 passed** (18 files) |
| Stream vitest | **427 passed** (26 files) |
| Backend pytest | **4330 passed** |
| `i18n:check` | **pass** |

The single ruff addition is the `Optional[str]` annotation on the new schema field, written that way to match the 15 sibling fields of the model it joins rather than be the lone `str | None` in the block.

### Two gates that need a caveat rather than a tick

**`typecheck:tests` fails — but it fails identically on the base.** 56 errors, in `configStore.test.ts` and `test/test-utils.tsx`. This branch touches no TypeScript at all (the diff is six Python files plus the CHANGELOG), so it cannot be the cause; I confirmed it by running the same gate on a clean `origin/development` worktree and getting the same 56. Consistent with `typecheck:tests` not being part of `npm run check`.

**`i18n:validate` should not be used as a check — it mutates the repo.** It is `npx i18next-parser && node scripts/i18n-check.mjs`, and the parser step *writes*: running it modified 53 locale JSON files and created 8 new `ns.json` files. The additions are pre-existing drift on `development` — keys referenced in source but absent from the locale files, added as placeholder values like `"lineage.absent": "lineage.absent"` — plus an `ns.json` produced from a dynamic namespace the parser cannot resolve. None of it is related to this PR.

I reverted all of it; nothing was committed and the head is unchanged. The read-only half, `node scripts/i18n-check.mjs`, passes and is what I recorded above. Worth knowing generally: anyone running `i18n:validate` to *check* a branch will silently dirty their tree with unrelated placeholder keys, and could easily commit them by accident.

### Follow-ups filed

The two things #410 deliberately left open are now tracked rather than resting in a PR conversation:

- **#411** — re-uploading a corrected row adds a record instead of updating it. Needs a dataset namespace (`csv-row-v2`), which needs a frontend change and a decision about what a namespace *is*, since `default_source` is user-editable and the filename is never sent.
- **#412** — `IngestMessage` sets `extra="forbid"` yet omits `ingestion_method` (and `source_origin`, `manual_import_job_id`). Harmless only because the processor's bundle excludes `plugins/_shared`, so the validator never imports and logs `validation disabled` on every cold start. The trap is that fixing that packaging — a clearly correct change — breaks manual import for a reason that looks unrelated.

### State

Deployed and verified on the data that triggered the report (400 → 800, detail in the comment above). Every gate available to me is green or explained. `MERGEABLE` / `CLEAN`.

What I cannot supply is the part that matters most for a public sample repo: **a human review.** There are none on this PR, and it is my own, so I have not merged it. Over to a maintainer.

Issue #253, opened 2026-08-05
[Infra]: Async Lambda paths have no failure destinations or DLQs


perrozzi:
## Summary

There is one DLQ in the entire platform. Async invocations elsewhere — including the long-running
generation jobs — have no failure destination, so a message that exhausts its retries is gone.

## Current state

`grep -rn 'deadLetterQueue\|onFailure\|DestinationConfig' lib/stacks/*.ts` returns a single hit:
`lib/stacks/ingestion-stack.ts:244` (`maxReceiveCount: 3`).

Not covered:

- async self-invocations on the API side
- the job Lambdas under `lambda/jobs/*` (document generation, merging, persona import) — the
  longest-running and most expensive work in the system
- Step Functions task failures, beyond whatever the state machine defines
- DynamoDB Streams → aggregator: retries are bounded (`processing-stack-consolidated.ts:198`,
  `retryAttempts: 3`) but there is **no `onFailure` destination and no `bisectBatchOnFunctionError`**

The stream case is the sharpest, though not in the way it first looks. Because retries *are* bounded, a
poison record does not stall the shard — it is **discarded** after three attempts, along with the rest of
its batch, since batch bisection is off. So a single malformed record silently drops the good records
batched with it, and the aggregate counters are permanently short. With no alarms (see the alarms issue),
nothing reports it.

## Proposed solution

- Failure destination (SQS) on every async Lambda invocation.
- DLQ on the job Lambdas, with enough context in the message to retry the job.
- On the streams event source: add an `onFailure` destination so discarded batches are recoverable, and
  `bisectBatchOnFunctionError` so one bad record stops taking its whole batch down with it.
- Alarm on every DLQ depth (tracked in the alarms issue; the two are worth doing together — a DLQ with
  no alarm just moves the silence).

## Acceptance criteria

- [ ] Every async Lambda has a failure destination or DLQ.
- [ ] The streams event source has `onFailure` and `bisectBatchOnFunctionError` configured.
- [ ] A deliberately-poisoned record lands in the failure destination, and the valid records batched with
      it are still processed.
- [ ] Each DLQ has a depth alarm.

## Area

Infrastructure (CDK) — `lib/stacks/processing-stack-consolidated.ts`, `lib/stacks/api-stack.ts`,
`lib/stacks/ingestion-stack.ts`

perrozzi:
> 🤖 **Automated triage.** The labels and duplicate candidates below are suggestions, not decisions.

[...]

PR #413, opened 2026-09-09
fix(bedrock): stop a long generation retrying itself into the Lambda ceiling

perrozzi:
A generation needing more than five minutes could not succeed on any Bedrock surface, and a failed prototype build cost roughly 45 minutes. Two independent retry layers were multiplying on top of one 15-minute function, and neither was visible in the logs.

Measured on a live deployment: one `build_prototype` job watched from start to kill, with the immediately preceding successful build of the same project as the control.

## The client budget

`shared/aws.py::get_bedrock_client` built the **one cached** Bedrock client with `read_timeout=300` and `max_attempts=3`. **3 × 300 = 900**, exactly the timeout of the job Lambdas that build prototypes, documents, personas and research. So the first read timed out at 5 minutes, botocore retried twice more on its own, and Lambda killed the function part-way through the third with nothing returned. Each abandoned attempt re-submitted the prompt and re-paid for a full generation.

| | |
|---|---|
| Application retry counter, entire lifetime | `Attempt 1/5` — never advanced |
| Invocation outcome | `Duration: 900000.00 ms` |
| Control: same step, same project, minutes earlier | **154,884 ms**, `stop_reason=end_turn` |

Those retries happen **below** `shared/converse.py`'s own loop, which is why the log read "Attempt 1/5" for the whole 900 s: the run looked like one slow call rather than three abandoned generations. The control proves it is size-dependent, not an outage.

`converse` is non-streaming, so the socket is legitimately idle until the generation completes — a read timeout measures "big", not "broken", and the larger the requested output the more certain it fires. `max_tokens=32000` on the prototype surface therefore invited output that a 300 s read timeout forbade: the configuration contradicted itself.

**Now `BEDROCK_READ_TIMEOUT_SECONDS = 840` with `BEDROCK_MAX_ATTEMPTS = 1`.** `mode: 'standard'` is stated explicitly because that mode counts *total* attempts, where legacy mode's reading of the same key is ambiguous — `shared/mcp_delegate.py` already uses that pair.

**`ReadTimeoutError` is now non-retryable in `_converse_with_retry`.** This is load-bearing rather than tidiness: the loop's `except Exception` would otherwise inherit botocore's job and make it 5 × 840 instead of 3 × 300 — the same defect one layer up with a bigger multiplier. `ConnectTimeoutError` is a sibling class, not a subclass, and stays retryable: it costs 10 s and no generation.

## The async re-drive, which is where the 45 minutes came from

The four job Lambdas are invoked with `InvocationType='Event'`, where AWS re-runs a failed invocation **twice more by default** — invisibly, because each re-run wrote progress to the same job row and looked like the first attempt making none. The log shows a second `START` with the same request id after the first died.

They now carry `retryAttempts: 0`. An LLM generation is neither cheap nor idempotent, and `shared/jobs.py` already records the job `failed` for the UI to render, so a hidden retry buys nothing. Step Functions is unaffected: an `EventInvokeConfig` governs async invocations only, so the PRD/PR-FAQ path's explicit, visible `llmRetry` still applies.

## The same defect in a second place

`product_doc_extractor` builds its own Bedrock client and had the same collision at a smaller scale: botocore's default 60 s read timeout with default retries against a 120 s function, so its last attempt was always killed in flight. Now 80 s × 1, which also lets a slow image description finish where it previously gave up. That handler is stdlib+boto3 only and cannot import `shared/`, so its constants are duplicated by design, like `_is_conditional_check_failure`.

Found by grepping every `bedrock-runtime` client rather than only the reported one. `shared/avatar.py`'s image client was checked and left alone: 120 × 3 = 360 s fits inside its caller, and throttling is its expected failure, so its retries are wanted.

## Tests

The collision spans a Python constant and a CDK timeout, and neither file can see both — so it is pinned from both sides.

- **`lambda/shared/test/test_aws.py`** — what actually reaches botocore, the explicit standard mode, one attempt, and the budget against the 900 s Lambda maximum with a reserve for recording the failure. Nothing covered `get_bedrock_client` before (that file's docstring records dropping client-factory tests), which is how values that read as reasonable line by line survived.
- **`lambda/shared/test/test_converse.py`** — the read timeout raising after exactly one attempt with no backoff, plus a connect-timeout control so the branch is narrow rather than "nothing retries".
- **`lib/stacks/api-stack.test.ts`** — reads the Python constants and compares them to the synthesized job-Lambda timeouts, and asserts the `EventInvokeConfig` **exists** with `MaximumRetryAttempts: 0`, since a missing one is indistinguishable from the default at a glance. Same shape as the delegation-timeout suite above it.
- **`lib/stacks/core-stack.test.ts`** — the same for the extractor.
- **`lambda/product_doc_extractor/test/test_handler.py`** — that the client actually *uses* those constants. The rest of that suite injects fakes into `_clients`, so `_bedrock()` never ran and a dropped `config=` would have left the CDK guard passing while the deployed client went back to botocore's defaults.

Four mutations were run to confirm each case fails for its stated reason: restoring 300 × 3, raising the read timeout to 900, dropping one `retryAttempts`, and disabling the read-timeout branch.

## Baseline

`lib/test-support/baseline.json` is regenerated because `retryAttempts` adds four `AWS::Lambda::EventInvokeConfig` resources. The template was dumped before and after and diffed: the delta is exactly those four plus Lambda asset hashes, with the readable name inventory byte-identical — no table, bucket or user pool is renamed. (`canonicalTemplate` normalises asset hashes, which is why only VocApiStack's digest moves.)

## Two things this deliberately does NOT claim

1. **The budget is not below every caller's timeout, and cannot be.** One cached client serves 30 s API handlers and 15-minute jobs alike; a per-caller budget needs a keyed client cache plus the caller's remaining time plumbed through `converse()`, which nothing passes today. What is asserted instead is the property that makes one shared budget safe for the short callers: **one attempt**, so nobody is handed a multiple of the read timeout. A caller shorter than 840 s still dies at its own ceiling — that residue is the separately documented job-timeout defect, whose fix covers OOM and every other kill too, so a read-timeout-shaped version would only be a partial one.
2. **No live prototype build over 300 s has been run.** Unit and synth level only. The arithmetic now permits it (the control emitted 15,833 output tokens in 155 s, so 32000 tokens sits far inside 840 s) but permitted is not observed.

Also unchanged: routing an exhausted async invocation somewhere durable (a failure destination / DLQ) is #253. `retryAttempts: 0` stops the multiplier; it does not add a destination.

## Verification

Run on this base (`f7e07593`):

| Leg | Result |
|---|---|
| `test:backend` | **4,339 passed** |
| `test:cdk` | **339 / 340** — the one failure is pre-existing and unrelated: `api-stack.test.ts`'s repo-markdown scan walks a local untracked `.tmp/*.md` scratch file. Moving that file aside gives 81/81 in the file |
| `typecheck:cdk` (`tsc --noEmit`) | clean |
| `lint:python` (ruff) | no new findings — the touched files report the same 18 pre-existing ones before and after |

No frontend or stream code is touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,27 @@ displays: the UI's build identifier is the short git commit SHA, injected at bui
 
 ### Fixed
 
+- A generation that needs more than five minutes of model time now succeeds, instead of failing three
+  times over and reporting nothing. The shared Bedrock client waited five minutes for a response and
+  then retried twice by itself, so its real budget was fifteen minutes — exactly the ceiling of the
+  job functions that build prototypes, documents, personas and research. A long generation therefore
+  had no path to success: it was abandoned and restarted twice, each attempt re-paying for a full
+  generation, and the function was killed part-way through the third with nothing to show while the
+  job still read as running. Those retries also happened below the platform's own retry loop, so the
+  logs said "attempt 1 of 5" for the entire fifteen minutes and the run looked like one slow call.
+  The client now makes a single attempt and waits fourteen minutes for it, and a read timeout is
+  reported and recorded as a failed job rather than retried into the ceiling. Prototype builds are
+  where this was measured, because they ask for the largest output, but every Bedrock surface shared
+  the same client and the same ceiling.
+- A failed prototype build no longer costs up to three quarters of an hour. The job functions are
+  invoked asynchronously, where AWS re-runs a failed invocation twice more by default, so one click
+  became roughly forty-five minutes of model work — invisible, because each re-run wrote progress to
+  the same job row and looked like the first attempt making none. A failure is now reported once, and
+  retrying is the reader's decision.
+- The product-document extractor, which builds its own Bedrock client, had the same collision at a
+  smaller scale: it waited sixty seconds for an image description with retries behind it, against a
+  two-minute function, so its final attempt was always cut off. It now waits ninety seconds once,
+  which also lets a slow description finish where it previously gave up.
 - A CSV upload no longer overwrites an earlier one whose `id` column covered the same numbers.
   A row's identity was the `id` column value as written, and every CSV upload enters the pipeline
   under the single `source_platform` of `manual_import`, from which the processor derives both its
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,11 @@ displays: the UI's build identifier is the short git commit SHA, injected at bui
   reported and recorded as a failed job rather than retried into the ceiling. Prototype builds are
   where this was measured, because they ask for the largest output, but every Bedrock surface shared
   the same client and the same ceiling.
+  Two of the shorter jobs — merging documents at ten minutes, importing a persona at five — are
+  bounded by their own limit rather than by that fourteen minutes, so a generation longer than their
+  own budget still ends without an explanation attached to it, exactly as it did before. Retrying a
+  throttled request is unaffected everywhere: that is worth a second attempt and is still made, with
+  backoff, by the platform rather than by the AWS client underneath it.
 - A failed prototype build no longer costs up to three quarters of an hour. The job functions are
   invoked asynchronously, where AWS re-runs a failed invocation twice more by default, so one click
   became roughly forty-five minutes of model work — invisible, because each re-run wrote progress to
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -46,7 +46,7 @@ displays: the UI's build identifier is the short git commit SHA, injected at bui
   retrying is the reader's decision.
 - The product-document extractor, which builds its own Bedrock client, had the same collision at a
   smaller scale: it waited sixty seconds for an image description with retries behind it, against a
-  two-minute function, so its final attempt was always cut off. It now waits ninety seconds once,
+  two-minute function, so its final attempt was always cut off. It now waits eighty seconds once,
   which also lets a slow description finish where it previously gave up.
 - A CSV upload no longer overwrites an earlier one whose `id` column covered the same numbers.
   A row's identity was the `id` column value as written, and every CSV upload enters the pipeline
--- a/voc-datalake/lambda/api/manual_import_processor.py
+++ b/voc-datalake/lambda/api/manual_import_processor.py
@@ -15,6 +15,7 @@ from shared.logging import logger, tracer, metrics
 from shared.aws import get_dynamodb_resource, get_bedrock_client
 from shared.exceptions import ValidationError
 from shared.model_config import get_active_model_id, uses_adaptive_thinking
+from shared.converse import bedrock_call_with_retry
 
 dynamodb = get_dynamodb_resource()
 bedrock = get_bedrock_client()
@@ -151,11 +152,19 @@ def process_job(job_id: str) -> None:
         
         logger.info(f"Invoking Bedrock for job {job_id} with model {model_id}")
         
-        bedrock_response = bedrock.invoke_model(
-            modelId=model_id,
-            body=json.dumps(request_body),
-            contentType="application/json",
-            accept="application/json"
+        # Through the shared retry policy: this raw invoke_model path does not go
+        # through converse(), and the shared client makes one botocore attempt by
+        # design (see BEDROCK_READ_TIMEOUT_SECONDS), so without this a single
+        # throttle would fail the whole import job.
+        bedrock_response = bedrock_call_with_retry(
+            lambda: bedrock.invoke_model(
+                modelId=model_id,
+                body=json.dumps(request_body),
+                contentType="application/json",
+                accept="application/json"
+            ),
+            step_name=f'manual_import_parse_{job_id}',
+            call_label='client.invoke_model()',
         )
         
         response_body = json.loads(bedrock_response['body'].read())
--- a/voc-datalake/lambda/api/product_context.py
+++ b/voc-datalake/lambda/api/product_context.py
@@ -31,6 +31,7 @@ from shared.logging import logger, tracer
 from shared.aws import get_dynamodb_resource, get_bedrock_client
 from shared.image_limits import IMAGE_CONTENT_TYPE_EXTENSIONS, MAX_IMAGE_BYTES
 from shared.model_config import get_active_model_id, omits_temperature
+from shared.converse import bedrock_call_with_retry
 from shared.project_writes import (
     put_project_item,
     put_project_item_and_increment,
@@ -482,15 +483,24 @@ def interview_turn(project_id: str, body: dict) -> dict:
     if not omits_temperature(model):
         inference_config['temperature'] = 0.3
     try:
-        resp = client.converse(
-            modelId=model,
-            messages=messages,
-            system=[{'text': system_prompt}],
-            inferenceConfig=inference_config,
-            toolConfig={'tools': [_build_interview_tool()]},
+        # Through the shared retry policy, not bare: this raw call does not go
+        # through converse(), so without it the FIRST throttle would become a
+        # user-visible "AI interview unavailable" — the shared client makes one
+        # botocore attempt by design (see BEDROCK_READ_TIMEOUT_SECONDS).
+        resp = bedrock_call_with_retry(
+            lambda: client.converse(
+                modelId=model,
+                messages=messages,
+                system=[{'text': system_prompt}],
+                inferenceConfig=inference_config,
+                toolConfig={'tools': [_build_interview_tool()]},
+            ),
+            step_name='interview_turn',
         )
     except Exception as e:
         logger.exception(f'Interview Bedrock call failed: {e}')
+        raise ServiceError('AI interview unavailable. Please try again.')
+    if not isinstance(resp, dict):  # throttled out with retries disabled
         raise ServiceError('AI interview unavailable. Please try again.')
 
     output_blocks = resp.get('output', {}).get('message', {}).get('content', [])
--- a/voc-datalake/lambda/api/product_context.py
+++ b/voc-datalake/lambda/api/product_context.py
@@ -498,9 +498,10 @@ def interview_turn(project_id: str, body: dict) -> dict:
             step_name='interview_turn',
         )
     except Exception as e:
+        # Covers sustained throttling too: bedrock_call_with_retry raises rather
+        # than returning None while raise_on_throttle is left at its default, so
+        # there is no empty-result case to check for below.
         logger.exception(f'Interview Bedrock call failed: {e}')
-        raise ServiceError('AI interview unavailable. Please try again.')
-    if not isinstance(resp, dict):  # throttled out with retries disabled
         raise ServiceError('AI interview unavailable. Please try again.')
 
     output_blocks = resp.get('output', {}).get('message', {}).get('content', [])
--- a/voc-datalake/lambda/jobs/persona_importer/handler.py
+++ b/voc-datalake/lambda/jobs/persona_importer/handler.py
@@ -138,7 +138,7 @@ def handle_job(ctx: JobContext, project_id: str, job_id: str, import_config: dic
             inferenceConfig={'maxTokens': prompt_config['max_tokens']}
         ),
         step_name='import_persona',
-    ) or {}
+    )
     
     response_text = response.get('output', {}).get('message', {}).get('content', [{}])[0].get('text', '')
     
--- a/voc-datalake/lambda/jobs/persona_importer/handler.py
+++ b/voc-datalake/lambda/jobs/persona_importer/handler.py
@@ -24,6 +24,7 @@ from shared.persona_import import validate_import_config
 from shared.prompts import PERSONA_IMPORT_PROMPTS, format_prompt, load_prompt_file
 from shared.image_limits import converse_image_format
 from shared.aws import get_dynamodb_resource, get_bedrock_client
+from shared.converse import bedrock_call_with_retry
 from shared.model_config import get_active_model_id
 from shared.project_writes import put_project_item_and_increment
 from api.projects import generate_persona_avatar
@@ -122,15 +123,22 @@ def handle_job(ctx: JobContext, project_id: str, job_id: str, import_config: dic
     # nothing to omit for temperature-restricted models.
     model_id = get_active_model_id('documents')
     logger.info(f"[IMPORT_PERSONA_JOB] Invoking Bedrock with model {model_id}")
-    response = bedrock.converse(
-        modelId=model_id,
-        system=[{'text': system_prompt}],
-        messages=[{'role': 'user', 'content': converse_content}],
-        # From the template too, so the budget lives beside the schema it has to
-        # produce rather than as a literal here that nobody updates when the
-        # schema grows.
-        inferenceConfig={'maxTokens': prompt_config['max_tokens']}
-    )
+    # Through the shared retry policy, because this raw call does not go through
+    # converse() and the shared client makes one botocore attempt by design (see
+    # BEDROCK_READ_TIMEOUT_SECONDS): without it a single throttle would fail the
+    # import job outright, and this surface is reached straight after an upload.
+    response = bedrock_call_with_retry(
+        lambda: bedrock.converse(
+            modelId=model_id,
+            system=[{'text': system_prompt}],
+            messages=[{'role': 'user', 'content': converse_content}],
+            # From the template too, so the budget lives beside the schema it has
+            # to produce rather than as a literal here that nobody updates when
+            # the schema grows.
+            inferenceConfig={'maxTokens': prompt_config['max_tokens']}
+        ),
+        step_name='import_persona',
+    ) or {}
     
     response_text = response.get('output', {}).get('message', {}).get('content', [{}])[0].get('text', '')
     
--- a/voc-datalake/lambda/product_doc_extractor/handler.py
+++ b/voc-datalake/lambda/product_doc_extractor/handler.py
@@ -282,22 +282,18 @@ JPEG_SCAN_BYTES = 256 * 1024
 
 MAX_DESCRIPTION_TOKENS = 4096
 
-# This Lambda's Bedrock budget. Deliberately NOT imported from shared/aws.py —
-# this handler is stdlib+boto3 only so CoreStack stays container-free (see the
-# module docstring), the same reason _is_conditional_check_failure is duplicated
-# below. The reasoning is one place, there; only the numbers differ, because they
-# are sized against THIS function's 120 s timeout rather than the 15-minute jobs.
+# This Lambda's Bedrock budget. Same invariant as shared/aws.py's — the two numbers
+# multiply and the product must stay under this function's own timeout — but sized
+# against 120 s rather than the 15-minute jobs, and duplicated rather than imported
+# because this handler is stdlib+boto3 only (see the module docstring), like
+# _is_conditional_check_failure below. The reasoning lives there.
 #
-# botocore's DEFAULTS are wrong here in exactly the way the shared client's old
-# ones were: a 60 s read timeout with the default retry budget can outlast this
-# function's own 120 s ceiling, so the final attempt is always killed in flight —
-# a guaranteed-doomed retry instead of a diagnosis. One attempt, and a read
-# timeout that still leaves time to record the failure. 90 s is also a real
-# widening for a slow description: the default gave up at 60 s.
-#
-# Raising these means re-checking the 120 s timeout in core-stack.ts, which is in
-# turn bounded by product_context.py's EXTRACTION_STALL_SECONDS (300).
-BEDROCK_READ_TIMEOUT_SECONDS = 90
+# botocore's DEFAULTS break that invariant here: 60 s with the default retry budget
+# outlasts 120 s, so the final attempt is always killed in flight. 80 s x 1 leaves
+# time to record the failure, and widens the window for a slow description either
+# way. Raising it means re-checking core-stack.ts's timeout, itself bounded by
+# product_context.py's EXTRACTION_STALL_SECONDS.
+BEDROCK_READ_TIMEOUT_SECONDS = 80
 BEDROCK_MAX_ATTEMPTS = 1
 BEDROCK_CONNECT_TIMEOUT_SECONDS = 10
 
--- a/voc-datalake/lambda/product_doc_extractor/handler.py
+++ b/voc-datalake/lambda/product_doc_extractor/handler.py
@@ -282,6 +282,25 @@ JPEG_SCAN_BYTES = 256 * 1024
 
 MAX_DESCRIPTION_TOKENS = 4096
 
+# This Lambda's Bedrock budget. Deliberately NOT imported from shared/aws.py —
+# this handler is stdlib+boto3 only so CoreStack stays container-free (see the
+# module docstring), the same reason _is_conditional_check_failure is duplicated
+# below. The reasoning is one place, there; only the numbers differ, because they
+# are sized against THIS function's 120 s timeout rather than the 15-minute jobs.
+#
+# botocore's DEFAULTS are wrong here in exactly the way the shared client's old
+# ones were: a 60 s read timeout with the default retry budget can outlast this
+# function's own 120 s ceiling, so the final attempt is always killed in flight —
+# a guaranteed-doomed retry instead of a diagnosis. One attempt, and a read
+# timeout that still leaves time to record the failure. 90 s is also a real
+# widening for a slow description: the default gave up at 60 s.
+#
+# Raising these means re-checking the 120 s timeout in core-stack.ts, which is in
+# turn bounded by product_context.py's EXTRACTION_STALL_SECONDS (300).
+BEDROCK_READ_TIMEOUT_SECONDS = 90
+BEDROCK_MAX_ATTEMPTS = 1
+BEDROCK_CONNECT_TIMEOUT_SECONDS = 10
+
 # The prototype generator's neutral default palette is indigo #4F46E5 (see
 # PROTOTYPE_HTML_SYSTEM_PROMPT in lambda/jobs/document_generator/handler.py). A
 # description that comes back with adjectives instead of values leaves the
@@ -354,7 +373,12 @@ def _s3():
 
 def _bedrock():
     if 'bedrock' not in _clients:
-        _clients['bedrock'] = boto3.client('bedrock-runtime')
+        from botocore.config import Config
+        _clients['bedrock'] = boto3.client('bedrock-runtime', config=Config(
+            read_timeout=BEDROCK_READ_TIMEOUT_SECONDS,
+            connect_timeout=BEDROCK_CONNECT_TIMEOUT_SECONDS,
+            retries={'max_attempts': BEDROCK_MAX_ATTEMPTS, 'mode': 'standard'},
+        ))
     return _clients['bedrock']
 
 
--- a/voc-datalake/lambda/product_doc_extractor/test/test_handler.py
+++ b/voc-datalake/lambda/product_doc_extractor/test/test_handler.py
@@ -17,6 +17,7 @@ passes for the wrong reason:
     build_product_context_block depends on the second one.
 """
 import pytest
+from unittest.mock import patch
 
 from .conftest import (
     DOCUMENTS_DEFAULT_MODEL,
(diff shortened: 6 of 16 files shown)


perrozzi:

This PR narrows the Bedrock client budgets so read timeout × retries can no longer exceed the enclosing Lambda ceiling (`shared/aws.py`: 300×3 → 840×1 with explicit `mode: 'standard'`; `product_doc_extractor/handler.py`: botocore defaults → 90×1), makes `ReadTimeoutError` non-retryable in `shared/converse.py`, and sets `retryAttempts: 0` on the four async-invoked job Lambdas, with Python + synth-level tests pinning the cross-language arithmetic from both sides. The diagnosis is sound and the test coverage is unusually well targeted (client factory, retry branch, and CDK↔Python arithmetic). I found no blocking defects, but two consequences of the new numbers deserve attention before merge.

### Should fix

- **The 840 s budget is now *above* two of the callers it serves, which regresses their failure diagnosis.** `lib/stacks/api-stack.ts` keeps `DocumentMergerJob` at 10 min and `PersonaImporterJob` at 5 min while `shared/aws.py`'s single cached client waits 840 s. Previously a stalled generation in the merger read-timed out at 300 s, inside its 600 s ceiling, so `shared/converse.py`/`shared/jobs.py` could record `failed`; now the read timeout can never fire first and the invocation is silently killed with the job row stuck on `running`. The PR text calls this "residue", but for the merger it is a behavioural change introduced *by* this diff, not pre-existing. At minimum, note it in the CHANGELOG entry (which currently reads as an unqualified improvement for all four job Lambdas) and reference the tracking issue; the `never multiplies inside a caller too short for the read timeout to bind` test in `lib/stacks/api-stack.test.ts` documents the asymmetry but does not cover the merger regression.

- **Audit `get_bedrock_client()` call sites that do not go through `shared/converse.py`.** Dropping `max_attempts` from 3 to 1 also removes botocore's *throttling* retries, not just the read-timeout multiplier. `converse.py` compensates for its own callers, but any direct `invoke_model` / `converse` / embeddings use of the shared client (chat, category generation, scrapers are all mentioned as sharing it) now surfaces `ThrottlingException` on the first 429 with no backoff anywhere. If such a call site exists, it needs its own retry loop or a keyed client. The diff shows no evidence this was checked, and no test covers it.

- **Audit the remaining `invoke_lambda_async` targets.** `retryAttempts: 0` is applied to exactly four functions in `lib/stacks/api-stack.ts`. If any other Lambda is invoked with `InvocationType='Event'` (ingestion, CSV processing, other job handlers in other stacks), the same invisible 2× re-drive still applies there. Either extend the change or say why those are idempotent — the `does not retry the whole generation behind the caller` test only enumerates the four.

- **`core-stack.test.ts` should assert a failure-recording reserve, like its `api-stack.test.ts` sibling.** The extractor guard is `budget < fn.Timeout` (90 < 120), while the shared-client guards require ≥ 30 s of headroom. As written, raising `BEDROCK_READ_TIMEOUT_SECONDS` to 119 would pass while leaving no time to record anything. Use the same `FAILURE_RECORDING_RESERVE_SECONDS` shape.

### Nits

- The comment blocks in `lambda/shared/aws.py` (≈40 lines for three constants) and `lambda/product_doc_extractor/handler.py` (≈18 lines) restate the PR description and CHANGELOG almost verbatim. Keep the invariant ("these multiply; product must stay under the caller's timeout; pinned in `api-stack.test.ts`") and drop the incident narrative — three copies of it will drift.
- `BEDROCK_READ_TIMEOUT_SECONDS: Final = 840` — prefer `Final[int]` for explicitness, consistent with the rest of the typed shared modules.
- `lambda/product_doc_extractor/test/test_handler.py` imports `patch` inside `_build`; the module-level import would be simpler and the explanatory docstring could then go away.
- Reading Python constants from TypeScript via regex (`api-stack.test.ts`, `core-stack.test.ts`) is guarded by `toBeDefined()`, which is the right call, but three near-identical regex pairs now exist — a small shared helper in `lib/test-support` would keep them honest.
- The `ReadTimeoutError` log message in `converse.py` advises "Reduce `max_tokens` for this step or split it"; worth confirming that raising here does not bypass any model/region fallback path a caller of `_converse_with_retry` relies on, since the old behaviour let botocore retry underneath.

---

perrozzi:
Thanks — one finding was a real defect I introduced, and it's fixed. Dispositions below, in the review's order. Pushed as `49ab4b2c`.

## Should fix

**1. The 840 s budget is above the merger (600 s) and importer (300 s).** Correct that they are, but the mechanism as described doesn't hold, so I've kept the numbers and made the asymmetry testable instead of prose.

At the old `300 × 3` the merger's first read timeout did **not** surface either: botocore retried internally, so attempt 2 was already running when the function was killed at 600 s, and the exception never reached `converse.py` or `jobs.py`. Same for the importer, where 300 s *was* its ceiling. So neither could record `failed` from a read timeout before this PR — the band-2 outcome is unchanged, not introduced. In one direction it strictly improves: a merger generation between 300 s and 600 s previously always failed (timeout, restart, killed) and now succeeds.

What band 2 does cost is the **diagnosis** — the row stays `running` instead of going `failed` — and that is the separately tracked job-timeout defect, whose fix (a `get_remaining_time_in_millis()` guard) covers OOM and every other kill too. A read-timeout-shaped fix here would be a partial version of it.

Acted on anyway, because you're right that the asymmetry deserved more than a comment:
- new case `classifies every job Lambda as one the read timeout binds inside, or one it does not`, which enumerates both bands. A new job Lambda, or a timeout change that moves an existing one across the line, now fails there and has to be classified in review.
- the CHANGELOG entry names the two jobs bounded by their own limit rather than reading as an unqualified win, and states that throttle retries are unaffected.

**2. Audit `get_bedrock_client()` call sites that bypass `converse.py`.** ✅ **You found a real regression. Fixed.**

There are five call sites; four are raw. Dropping to one attempt would have meant:
- `product_context.interview_turn` — first 429 becomes a user-visible "AI interview unavailable"
- `manual_import_processor.process_job` — first 429 fails the import job
- `persona_importer.handle_job` — first 429 fails the import job
- `avatar.generate_avatar_prompt_with_llm` — degrades to a template prompt

The policy is now one function: `shared/converse.py::bedrock_call_with_retry` holds the decision botocore cannot express — retry a throttle (a round trip, no generation), never retry a read timeout (the generation outran the client). `_converse_with_retry` delegates to it and keeps only the Converse-specific response logging, so its **72 existing tests pass unchanged** over the shared implementation. The three job- and user-facing raw callers now route through it.

`avatar.py` is deliberately exempt and recorded as such: it already degrades gracefully, its **image** client is a separate one that keeps botocore's retries on purpose (a 10-wide avatar fan-out makes throttling the expected failure), and it maintains a narrow import graph — `test_avatar` asserts importing it pulls in no `cryptography` — that importing `converse` would widen.

And a guard, because nothing about the client's signature makes a missing retry visible at the call site: `test_bedrock_retry_policy.py` walks the AST of every module that touches Bedrock and fails on a `.converse()`/`.invoke_model()` call that no `bedrock_call_with_retry(...)` encloses. Exemptions carry their reason, are asserted to still exist, and the detector is tested against a bare call so the scan can't pass vacuously. Verified by reverting one call site — it names the file and line.

**3. Audit the remaining `invoke_lambda_async` targets.** Enumerated, and the reasoning now lives on `JOB_ASYNC_RETRY_ATTEMPTS` rather than only in the PR. The others keep the AWS default because for them a re-drive is a benefit rather than a repeated bill: manual import re-does bounded, content-keyed work its idempotency records de-duplicate, and the scraper/integration invocations are watermark-driven, so a repeat resumes where it left off. What sets the four apart is that one invocation is one large generation, which a re-drive re-pays in full and cannot succeed at for the reason the first attempt failed.

**4. `core-stack.test.ts` needs the failure-recording reserve.** ✅ Right, and it was exactly as weak as you describe — 119 s against a 120 s function would have passed. Both suites now take the reserve from one shared constant, and the extractor's read timeout is 80 s to fit it (still a widening over botocore's 60 s default).

## Nits

- **Comment blocks too long.** Agreed and trimmed: 44 → 28 lines in `shared/aws.py`, 15 → 11 in the extractor. Kept the invariant, why one attempt rather than only a longer timeout, and where it's pinned; dropped the incident narrative the CHANGELOG carries.
- **`Final[int]`.** Declining on convention: there is no `Final[int]` or `Final[str]` anywhere in first-party code. 23 scalars use bare `Final`; the 4 uses of `Final[T]` are all collections where the parameter adds information (`frozenset[str]`, `tuple[str, ...]`). The direct precedent is `mcp_delegate.py`'s own botocore timeout constants, same kind of value in the same package.
- **`patch` imported inside `_build`.** Hoisted to module level; the docstring explaining the local import is gone with it.
- **Three near-identical regex pairs.** Now one helper, `lib/test-support/cross-language-invariants.ts`. It **throws** rather than returning undefined: a regex that quietly stops matching turns every assertion built on it into a vacuous pass, which is worse than no guard because it still reads like one.
- **Does raising sooner bypass a fallback path?** Checked, and no. Nothing in the repo keys a model, region or `max_tokens` change off an exception from these paths — every handler either fails the job, returns a canned response, or re-raises, identically whether the error arrives after 1 attempt or 5. Two near-misses that look like fallbacks and aren't: `_raised_empty_budget` moves `max_tokens` **up** and is triggered by a `ValidationException`, never a timeout; and the Step Functions retries are keyed on the throttling alias, so a read timeout already fell to `addCatch` before this change. For the job Lambdas the earlier raise is strictly better — `5 × 840 s` would be killed by the 15-minute ceiling long before the loop finished, so the old path produced no `failed` row at all.

## Verification at this head

backend **4,347** passed · CDK **340/341** · `tsc --noEmit` clean · ruff unchanged at 22 findings across the touched files, clean on both new ones · template unchanged, so `baseline.json` needed no regeneration.

The one CDK failure is pre-existing and unrelated: `finds no THIRD document stating these numbers unpinned` walks repo markdown and picks up an untracked local `.tmp/*.md` scratch file from July. Moving that file aside gives 81/81 in the file.

Still not claimed, unchanged from the description: **no live prototype build over 300 s has been run.** The arithmetic permits it; permitted is not observed.

perrozzi:

Since the last review the author added the throttling-retry layer I asked for (`bedrock_call_with_retry` extracted from `_converse_with_retry` and wrapped around the three raw-client call sites), an AST-based guard test that enforces it repo-wide, a shared `pythonIntConstant` / `BEDROCK_FAILURE_RECORDING_RESERVE_SECONDS` helper now used by both CDK suites, a CHANGELOG paragraph covering the two short-timeout jobs, a comment in `api-stack.ts` justifying scoping `retryAttempts: 0` to the four generators, and trimmed comment blocks. All four previous should-fix items are addressed; the extractor budget was dropped 90 → 80 s to satisfy the new 30 s reserve.

### Resolved
- Merger/importer asymmetry now documented in the CHANGELOG and *enumerated* as bands in `api-stack.test.ts` rather than left as prose.
- Throttling retries restored for the three raw `get_bedrock_client()` callers, with `test_bedrock_retry_policy.py` guarding against a fourth appearing unguarded.
- Other `invoke_lambda_async` targets audited and the scoping rationale recorded in `lib/stacks/api-stack.ts`.
- `core-stack.test.ts` now uses the same failure-recording reserve as its api-stack sibling; the regex duplication is gone via `lib/test-support/cross-language-invariants.ts`.
- Nits on the oversized comment blocks and the function-local `patch` import are addressed.

### New findings

- **`CHANGELOG.md` is now wrong about the extractor: it says "It now waits ninety seconds once" while `lambda/product_doc_extractor/handler.py` sets `BEDROCK_READ_TIMEOUT_SECONDS = 80`.** The PR description also still says 90 s. Update the changelog line (the only user-visible artefact of the two).
- **`lambda/api/manual_import_processor.py` doesn't handle the new `None` return.** `bedrock_call_with_retry` can return `None`, and the very next line is `bedrock_response['body'].read()` → `TypeError` instead of a diagnosis. Today it's unreachable (`raise_on_throttle` defaults to `True`), but the other two new call sites both guard (`or {}` in `persona_importer`, `isinstance` in `product_context`), so the inconsistency is the hazard — either guard all three or drop the guards and state the invariant once.
- **The dead branch in `lambda/api/product_context.py`** (`if not isinstance(resp, dict)`) is unreachable with `raise_on_throttle=True` and untested; the comment "throttled out with retries disabled" doesn't match the actual default. Either pass `raise_on_throttle=False` deliberately or remove the branch.

### Nits
- `_first_party_modules()` in `test_bedrock_retry_policy.py` skips any path containing `/test`, which also excludes production modules like `shared/testing_*.py`; and the guard only inspects files whose source mentions `get_bedrock_client`/`bedrock-runtime`, so a module handed a client as a parameter escapes it. Worth a line in the docstring so the coverage claim isn't overstated.
- `bedrock_call_with_retry` has no return annotation (`-> object | None`), while the rest of `shared/converse.py` is annotated.
- `shared.converse` is now imported by `api/product_context.py` and `api/manual_import_processor.py`; if any cold-start/import-graph assertion exists for the API handlers (the kind `test_avatar` makes), confirm it still holds.

---

perrozzi:
All three new findings fixed in `d6ca8c75`. The first two shared a root cause, so they're fixed at the helper rather than at the call sites.

**1. CHANGELOG said ninety seconds, code said eighty.** ✅ Corrected, and the PR description with it. It was accurate when written and went stale when the new 30 s reserve moved the number — the changelog being the user-visible artefact of the two is exactly why it's the one that had to be right.

**2 + 3. `manual_import_processor` didn't handle `None`; the `product_context` branch was dead and its comment wrong.** ✅ You identified the real hazard as the *inconsistency*, and that's where I fixed it: three call sites answering the same question three ways is how one of them ends up answering it wrong. Rather than guard all three, the helper's contract is now total.

`bedrock_call_with_retry` returns `None` **only** when `raise_on_throttle=False`; with the default it either returns a result or raises. That needed one real change, not just a docstring: dropping `and last_exception` from the final raise. `max_retries=0` is the single path where the loop finishes having recorded no exception, and it's a caller mistake — it now raises `BedrockThrottlingError` naming that ("no attempt was made") instead of returning an empty value that reads like a Bedrock answer. Pinned by `test_a_zero_attempt_budget_raises_rather_than_returning_nothing`.

Both guards are then deleted as provably unreachable. Worth noting the dead branch wasn't only dead — its comment *described a default the call site didn't use*, which is the more expensive half of the problem, since the next reader would have trusted it.

## Nits

- **Return annotation** — added, `-> object | None`.
- **`_first_party_modules()` overstates coverage** — fair, and now stated in its docstring: a production module named `testing_*.py` would be skipped by the path filter (none exists today, and if one appears that line is why), and a module handed an already-built client as a **parameter** escapes the scan entirely. That second shape is precisely `shared/avatar.py`'s, which is exempt on its own merits anyway.
- **Import graph for the API handlers** — checked, nothing to hold. No cold-start or import assertion exists for them; the only guard of that kind is `test_avatar`'s `cryptography` one, and `avatar.py` is untouched. And `shared/converse.py` brings in nothing new: `random`, `time`, `typing`, `botocore.exceptions`, `shared.logging`, `shared.aws`, `shared.model_config` — all already in both bundles. `api/product_context.py` was importing it before this PR, just from inside a function.

## Verification at this head

backend **4,348** passed · CDK **340/341** · `tsc --noEmit` clean · ruff unchanged at 14 findings across the touched files.

Same single pre-existing CDK failure as before (`finds no THIRD document…` walking an untracked local `.tmp/*.md` from July), and the same open caveat: **no live prototype build over 300 s has been run.**

perrozzi:

Since the last review the author fixed the CHANGELOG's extractor figure (ninety → eighty seconds), removed the dead `isinstance` branch in `product_context.py` and the `or {}` in `persona_importer`, and made `bedrock_call_with_retry` a *total* contract instead: with the default `raise_on_throttle=True` it now always returns a result or raises (including the degenerate `max_retries=0`), documented in the docstring, annotated `-> object | None`, and covered by a new test. The docstring limits nit on `_first_party_modules()` is also addressed. All three previous findings are resolved.

### Resolved
- `CHANGELOG.md` now says "eighty seconds", matching `BEDROCK_READ_TIMEOUT_SECONDS = 80` in `lambda/product_doc_extractor/handler.py`.
- `manual_import_processor.py`'s unguarded `None` is no longer a hazard: the helper's contract is now total under the default, stated once and tested, and the two inconsistent ad-hoc guards are gone.
- Dead `if not isinstance(resp, dict)` branch removed from `lambda/api/product_context.py`, replaced by a comment that matches the actual default.
- Nits on the missing return annotation and the overstated AST-guard coverage claim.

### Still open
- The import-graph nit: `api/product_context.py` and `api/manual_import_processor.py` now import `shared.converse`; no evidence that any cold-start/import-graph assertion for the API handlers was re-checked.

### New findings

Nothing blocking. Two small points on the incremental change:

- **`lambda/shared/converse.py`: `max_retries=0` now raises where it previously returned `{}`/`None`.** That is the right call, but it is a behaviour change for any caller that derives `max_retries` from config rather than the literal default — worth a quick grep to confirm none does (e.g. a value read from a prompt template or env var that could be `0`).
- **`-> object | None` makes the two new call sites unverifiable by type.** `persona_importer/handler.py` and `product_context.py` both do `response.get(...)` directly on it now that the guards are gone; `Any` (or a `TypeVar` bound to `call`'s return) would express the contract the docstring describes without inviting the guards back.

---

perrozzi:
Both points taken, `bbf6b6f4`.

**`-> object | None` makes the call sites unverifiable by type.** ✅ Fixed as suggested — the helper is now generic in its call's return type, `Callable[[], _CallResult] -> _CallResult | None`. Your reasoning is the part worth keeping: two sites now read the response directly (`.get('output')`, `['body'].read()`) precisely because the contract says they need not guard, so a return type that hid the shape would have invited the guards straight back. The constant carries that note.

**`max_retries=0` is a behaviour change for a config-derived caller.** Checked rather than assumed, and there is no such caller. All three research call sites take the default; every other site passes a literal (5, 3, 2). The research prompt templates drive `max_tokens` but carry no `max_retries` — `inference['max_tokens']` is the only config-sourced inference value in that path. So the raise is reachable only from a caller mistake, which is what it is for.

**On the import-graph item still listed as open** — it was checked, and the answer is that there is nothing to re-check. There is no cold-start or import-graph assertion for the API handlers at all: the only guard of that kind in the repo is `test_avatar.py`'s `test_importing_the_module_does_not_pull_in_cryptography`, and `shared/avatar.py` is untouched by this PR. And `shared/converse.py` adds no new dependency to either bundle — its imports are `random`, `time`, `typing`, `botocore.exceptions`, `shared.logging`, `shared.aws`, `shared.model_config`, all already present. `api/product_context.py` was importing it before this PR, from inside `generate_report`; the only change there is that the import moved to module scope. If a cold-start assertion for the API handlers would be worth having, that is a reasonable ask on its own, but it would be new coverage rather than something this PR put at risk.

## Verification at this head

lambda suite **3,988** passed · ruff unchanged at 4 findings in the touched file · no TypeScript changed in this round.

Standing caveat, unchanged: **no live prototype build over 300 s has been run.** Everything here is unit- and synth-level, and that is the one acceptance criterion from the original diagnosis that only a deployed run can close.

perrozzi:

The only change since my last review is in `voc-datalake/lambda/shared/converse.py`: `bedrock_call_with_retry` now takes `Callable[[], _CallResult]` and returns `_CallResult | None` via a new module-level `TypeVar`, with a comment explaining why the result must not be flattened to `object`. That was the second of my two non-blocking points; the first (grep for a config-derived `max_retries=0`) is not visibly addressed.

### Resolved
- `-> object | None` replaced by `_CallResult | None`, so `persona_importer/handler.py` and `api/product_context.py` reading `response.get(...)` directly is now type-checkable rather than opaque.

### Still open
- No evidence that `max_retries` is never derived from config/env (the value `0` now raises where it previously returned `{}`/`None`) — non-blocking, a grep is enough.
- Import-graph nit: `api/product_context.py` and `api/manual_import_processor.py` now pull in `shared.converse`; no confirmation any cold-start/import-graph assertion for the API handlers was re-checked.

### New findings

Nothing blocking; the change is clean. One optional refinement: the docstring's "None ONLY when `raise_on_throttle=False`" is still invisible to a type checker, so strict callers must narrow `| None` anyway. A pair of `@overload`s keyed on `Literal[True]`/`Literal[False]` for `raise_on_throttle` would express the total contract exactly and keep the guards out for good — worth it only if you run a strict checker over `lambda/`.

---

perrozzi:
Both remaining items are now evidence in the code rather than claims in a comment thread — `65c6d134`. That's the right call on your part: a disposition you can't see from the diff is one the next reader can't either.

**Import graph.** Now a test: `test_converse_pulls_in_nothing_that_the_client_module_does_not`. Asserted as a *relative* property — `shared.converse` pulls in no third-party package that `shared.aws` does not — because every module that reaches Bedrock already imports `shared.aws` for the client, which makes that exactly the question, and unlike a curated allowlist it cannot go stale. It uses the same subprocess technique as `lambda/conftest.py`'s `imports_cryptography` fixture, for the same reason that fixture gives: half this suite has already imported these modules, so an in-process `sys.modules` check would pass no matter what the graph looks like. Verified by hoisting `requests` into `converse` — the guard fails and names `['charset_normalizer', 'idna', 'requests', …]`.

That also converts the concern into something durable: the two API handlers answer browser requests, so the real risk was never this PR's imports but the *next* one hoisted into `converse`.

**`max_retries`.** The grep result is now in the parameter's own docstring: every caller passes a literal or takes the default, and nothing derives it from configuration or the environment — the research prompt templates carry `max_tokens`, not this. So the `0` case is a programming error rather than something a settings row can produce.

**`@overload` on `Literal[True]`/`Literal[False]`** — declining, on your own condition. There is no strict type checker over `lambda/`: no `mypy.ini`, no `pyrightconfig.json`, nothing in `requirements-dev.txt` or the `check` script. `typecheck` and `typecheck:tests` are TypeScript only. So the overloads would be read by nothing, and the pair would be two more signatures to keep in step with the one that runs. Worth revisiting the moment a checker is added to the gate — at which point it becomes the right way to say what the docstring currently says.

## Verification at this head

backend **4,349** passed · ruff unchanged at 4 findings in `converse.py`, clean on the new test file.

Standing caveat, unchanged and the only thing I'd hold merge on: **no live prototype build over 300 s has been run.** Everything here is unit- and synth-level, and it's the one acceptance criterion from the original diagnosis that only a deployed run can close.

perrozzi:

Since my last review the only changes are a docstring expansion on `bedrock_call_with_retry`'s `max_retries` (documenting that no caller derives it from config/env) and a new `TestThePolicyCostsTheApiHandlersNothingAtColdStart` class in `lambda/shared/test/test_bedrock_retry_policy.py` that compares the third-party import set of `shared.converse` against `shared.aws` in a subprocess. Both of my previous open items are now addressed.

### Resolved
- `max_retries` docstring states the `0` case is a programming error and cannot come from a settings row — the grep I asked for, recorded in the file.
- Import-graph concern covered mechanically by the new relative-subset test, rather than left as an unverified claim about the two API handlers.

### New findings
Nothing blocking. Three small notes on `test_bedrock_retry_policy.py`:

- `{m.split('.')[0] for m in ...}` collapses every first-party module to `shared`, so the guard only sees new **top-level third-party packages**. A future `import boto3.dynamodb.transform`-style submodule cost inside `converse` (or inside a `shared.*` module it newly imports) is invisible. Worth one line in the docstring so the guard isn't read as stronger than it is.
- `sys.stdlib_module_names` is 3.10+; fine for current runtimes, but the subprocess uses `sys.executable`, so on an older local interpreter this fails with an opaque `AttributeError` in `result.stderr` rather than a skip.
- `_third_party_imports(...) -> set` — bare `set`; `set[str]` matches the typing conventions used elsewhere in this PR (e.g. `guarded: set[int]` in the same file).

---

perrozzi:

Since my last review the only change is documentation and typing polish in `lambda/shared/test/test_bedrock_retry_policy.py`: the new class docstring now states the top-level-package collapsing limitation and the Python 3.10+ requirement for `sys.stdlib_module_names`, and `_third_party_imports` is annotated `-> set[str]`. All three of my previous notes are addressed.

### Resolved
- Docstring now names what the import guard does *not* see (submodule-level additions inside an already-present package), so it can't be read as stronger than it is.
- 3.10+ `sys.stdlib_module_names` dependency documented, with the note that failure surfaces as the subprocess's stderr rather than a wrong answer.
- `_third_party_imports(...) -> set[str]` matches the typing conventions used elsewhere in the file.

Nothing new — these changes are comment/annotation only, with no behavioural or assertion impact. Looks clean.

---

perrozzi:
All three taken, `d6e4cbd7`.

- **The guard only sees top-level packages.** Right, and now stated in its docstring. The collapsing is deliberate — the comparison is about what has to be installed and loaded at all, and first-party `shared.*` names would otherwise dominate the diff with modules that are in both bundles by construction — but you're correct that it makes a `boto3.dynamodb.transform`-shaped submodule cost invisible, so the limit is recorded rather than the guard being read as stronger than it is.
- **`sys.stdlib_module_names` is 3.10+.** Noted in the docstring, with the runtimes it actually runs on (3.14 Lambda, 3.13 venv). Not converted to a skip: on an older interpreter the existing assertion surfaces the subprocess's own stderr, so the failure is legible and, more importantly, it fails rather than silently reporting a wrong answer — which is the direction I'd want an environment problem to break in a guard like this.
- **`-> set[str]`.** Annotated, matching `guarded: set[int]` and `offenders: dict[str, list[str]]` in the same file.

## Where this leaves the PR

Two consecutive reviews with nothing blocking, and every should-fix item from the first review closed — one of them (`bedrock_call_with_retry` and the AST guard) a real defect this review caught that I had shipped, and worth saying plainly: without it this PR would have traded a 45-minute retry storm for a first-throttle failure on three surfaces.

Guards added along the way, all mutation-verified: the client budget and its arithmetic in both languages, the read-timeout branch in both directions, the async retry cap, the retry policy's presence at every raw Bedrock call site, the helper's total contract, and the cold-start cost of making `converse` a dependency of two API handlers.

**One thing still open, and it is the only reason I would not merge this myself:** no live prototype build over 300 s has been run. Everything here is unit- and synth-level. The arithmetic permits the case that failed — the measured control emitted 15,833 output tokens in 155 s, so 32000 tokens sits far inside 840 s — but permitted is not observed, and that is the one acceptance criterion from the original diagnosis only a deployed run can close.

perrozzi:
## Deployed and verified end to end — the last open acceptance criterion is closed

The description said the one thing unproven was AC2: *"a generation longer than 300 s of server time completes successfully on the prototype surface."* It has now been observed on a live deployment.

Deployed this PR's head `d6e4cbd7` to `us-east-1`, four stacks `UPDATE_COMPLETE`. `VocWebSearchStack` was left untouched via `--exclusively` (its pending diff would create `SubmitAnthropicUseCase` — an account-level submission unrelated to this change).

### The proof run

A prototype **revision** — `base_prototype_id` pointing at a prototype built minutes earlier, plus `feedback` asking for substantially more screens — which folds the prior prototype's 48 k chars back into the prompt. A fresh build was not enough: it reached only 243 s, which the old configuration would also have survived.

| | control (fresh build) | **proof (revision)** |
|---|---|---|
| prompt | 25,254 + 2,914 system | **50,063** + 2,914 — the original failure was 52,293 |
| attempts logged | `Attempt 1/5` once | `Attempt 1/5` once |
| Bedrock time | 243.71 s | 🏁 **579.05 s** (`max_tokens` → auto-continuation → `end_turn`) |
| invocation | `Duration: 246234 ms` | **`Duration: 579651.66 ms`, completed** |
| output | 48,772 chars | 127,222 chars |
| `START` per request id | 1 | **1** — no async re-drive |
| job row | `running/40` → `completed/100` | `running/40` → `completed/100` |

The original failure, for comparison, was `Duration: 900000.00 ms` with the application log stuck on `Attempt 1/5` and the job row left on `running`. Under `300 × 3` that first call would have read-timed out at 300 s, been retried twice inside botocore, and been killed at the ceiling with nothing returned.

Note the `Attempt 1/5` line now comes from `bedrock_call_with_retry`, so the retry policy added in this PR is what ran.

### Configuration, measured on the deployment rather than inferred

- `shared/aws.py`, `shared/converse.py` and `product_doc_extractor/handler.py` pulled out of the deployed zips are **byte-identical** to this branch; live values `840 / 1 / 10` and `80 / 1`.
- All four job Lambdas report `MaximumRetryAttempts: 0`. The control — `voc-manual-import-processor`, deliberately excluded — has **no** `EventInvokeConfig` at all, i.e. still the AWS default of 2. That is the async half proven in both directions.
- `scripts/test-api.sh` → **34/34 passed**.
- CloudFront serving the bundle built for this deploy.

Both probe prototypes were deleted afterwards and their absence confirmed by re-reading the project (9 documents → 8).

### What merging ships

`origin/development` is still `f7e07593`, the base this branch was cut from, so a squash merge produces exactly this branch's tree `d5eecb6f` — **the tree that was deployed and exercised above**. If anything else lands on `development` first, that identity no longer holds and the gates are worth re-running on the rebased tree.

Issue #249, opened 2026-08-05
[Bug]: SQS message validation is dead in production, and would reject real messages if switched on


perrozzi:
## Summary

The processor's message-validation layer never runs in production, because the module it imports is not
bundled. If the bundling were fixed, validation would reject messages that real producers send — so the
naive fix silently drops manual imports and form submissions.

## Problem

Two defects that mask each other.

**1. The import always fails, so validation is off.** `lambda/processor/handler.py:34-40`:

```python
try:
    from _shared.schemas import safe_validate_message
    VALIDATION_ENABLED = True
except ImportError:
    logger.warning("Could not import validation schemas - validation disabled")
    VALIDATION_ENABLED = False
```

The processor asset is built in `lib/stacks/processing-stack-consolidated.ts:117`:

```
cp -r /asset-input/processor/* /asset-output/ && cp -r /asset-input/shared /asset-output/
```

`plugins/_shared` is **not** copied, so `_shared.schemas` is never importable and `VALIDATION_ENABLED`
is always `False`. The guard at line 145 then skips validation entirely. A documented security boundary
does nothing, and only a warning-level log says so.

**2. The schema would reject valid messages.** `plugins/_shared/schemas.py:104` declares `IngestMessage`
with `model_config = {"extra": "forbid"}` (line 106). Producers send fields the model does not declare —
`ingestion_method` (added for import provenance) and `preset_category` among them. So enabling
validation would fail messages from the manual-import and feedback-form paths.

**3. A validation failure loses the message.** On failure the record is treated as handled and deleted
rather than routed to a DLQ, so the combination of fixing (1) without (2) means silent data loss.

## Proposed solution

Fix in this order — the order matters:

1. **Reconcile the schema with reality first.** Enumerate what every producer actually sends and either
   declare those fields or relax `extra` to `ignore`. `forbid` on a message contract with several
   independent producers is a standing liability.
2. **Route failures to a DLQ** before enabling anything, so a mismatch is visible instead of lossy.
3. **Then fix the bundling** to include `plugins/_shared`.
4. **Make the disabled state loud.** A warning log for a dead security control is not enough — fail
   closed, or emit a metric an alarm can catch (see the alarms issue).

## Acceptance criteria

- [ ] `IngestMessage` accepts every field real producers send; a test enumerates producers and asserts
      round-trip validation for each.
- [ ] Validation failures land in a DLQ; none are silently deleted.
[...]

PR #420, opened 2026-09-11
fix: align ingest schema with manual imports

### LINKED ISSUE #412: IngestMessage forbids extras but omits fields manual import sends; validation is silently disabled in the processor
## Summary

`IngestMessage` rejects unknown fields, but the manual-import path has been sending a field the model does not declare. It causes no harm today only because the validator is never importable in the function that would run it. Anyone who fixes that packaging — a reasonable, unrelated cleanup — turns a latent mismatch into rejected messages.

Found while working on #410; deliberately left alone there because it is wider than that fix.

## The three pieces

**1. The model forbids extras.** `voc-datalake/plugins/_shared/schemas.py`:

```python
class IngestMessage(BaseModel):
    model_config = {"extra": "forbid"}  # Reject unknown fields
```

**2. Manual import sends fields it does not declare.** `ingestion_method` is sent by both the CSV and JSON upload paths and by `confirm_import`, and is persisted by the processor — but it is not a field on `IngestMessage`. `confirm_import` additionally sends `source_origin` and `manual_import_job_id`.

**3. The validator never loads.** `lambda/processor/handler.py`:

```python
try:
    from _shared.schemas import safe_validate_message
    VALIDATION_ENABLED = True
except ImportError:
    logger.warning("Could not import validation schemas - validation disabled")
    VALIDATION_ENABLED = False
```

The processor's asset copies `processor/` and `shared/`, not `plugins/_shared/`, so that import always fails. The deployed function logs the warning on every cold start — it is observable in CloudWatch right now.

So the security boundary described in that module's own docstring ("acts as a security boundary between untrusted plugin output and trusted processing") is not in force in the processor.

## Why it matters

The failure is ordered badly. Someone fixing the bundling to enable validation would be doing something clearly correct, and the immediate consequence is that manual-import messages fail validation whole — `extra="forbid"` rejects the message, not the offending field — so imports stop working for a reason that looks unrelated to the change they made.

## Suggested resolution

Reconcile the model with what the pipeline actually sends **before** anything enables validation:

- add `ingestion_method`, `source_origin` and `manual_import_job_id` to `IngestMessage`, bounded and sanitized like their neighbours;
- audit the other producers for fields absent from the model (the plugin ingestors are the other senders);
- then decide separately whether the processor should bundle `plugins/_shared` so validation is real, and make that a deliberate change with its own testing.

A test asserting that every field the manual-import paths emit is accepted by `IngestMessage` would keep the two from drifting again. #410 added `csv_row_id` to the model for exactly this reason, so that field is already correct either way.

## Note

Not a live defect — nothing is failing today. It is a trap for the next person, and cheap to defuse now.

proton0210:
## Summary

- declare all manual-import provenance fields on `IngestMessage`, including `source_url`, which manual confirm emits alongside `url`
- apply explicit length bounds and the existing URL/string sanitization to the newly modeled fields
- add regression coverage for the manual-confirm, CSV-upload, and JSON-upload message shapes
- derive rejection-boundary tests from the schema constants and verify every manual provenance field survives validation

The producer audit found that plugin ingestors route through `BaseIngestor`/`BaseWebhook` normalization and already emit fields represented by the schema. The broader feedback-form mismatch and enabling processor-side validation remain part of #249 and are intentionally outside this focused change.

## Testing

- `python -m pytest -q --no-header --tb=short` — 4,354 passed
- `python -m ruff check plugins/_shared/schemas.py plugins/_shared/test/test_schemas.py` — no new findings compared with the repository baseline
- `git diff --check`

Closes #412


perrozzi:

This PR declares three manual-import provenance fields (`ingestion_method`, `source_origin`, `manual_import_job_id`) on the strict `IngestMessage` schema with length bounds plus the existing string sanitizer, drops the duplicated `source_url` field from the confirm-import SQS payload, and adds parametrized schema tests for the manual/CSV/JSON upload shapes. The change is small, focused, and consistent with the schema's existing style of bounded optional strings; my main concern is that the `source_url` removal is a producer-side behavior change with no test or evidence of consumer coverage.

### Blockers

- `voc-datalake/lambda/api/manual_import_handler.py`: removing `source_url` from the queue message is the only behavioral change in the PR, and it has no regression test. Per repo policy a bug fix needs a test that fails when the fix is reverted — the added tests in `plugins/_shared/test/test_schemas.py` only exercise the schema, so reverting the handler line would leave all tests green. Please add a test asserting the message emitted by `confirm_import()` validates against `IngestMessage` (which both pins the field set and proves `url` still carries the value). Additionally, confirm nothing downstream (processor, enrichment, Glue/Athena projections, any dead-letter replay tooling) reads `source_url`; the PR description asserts this but the diff carries no evidence.

### Should fix

- `voc-datalake/plugins/_shared/schemas.py`: `manual_import_job_id` is bounded by `MAX_ID_LENGTH` and merely sanitized. If this value is ever interpolated into an S3 key/prefix or DynamoDB key downstream, sanitizing control characters is not sufficient validation at the boundary — it is generated as a UUID, so a `pattern=r"^[0-9a-fA-F-]{36}$"` (or similar) would be a tighter, cheap constraint.
- `voc-datalake/plugins/_shared/test/test_schemas.py`: the bounds test hardcodes `257` / `65` rather than deriving from `MAX_ID_LENGTH` and the literal `64`. If the constants change, the test silently stops testing the boundary. Use `'s' * (MAX_ID_LENGTH + 1)` and export/reference the channel-length constant.

### Nits

- `schemas.py`: the three new fields use `str | None` while every surrounding field uses `Optional[str]`. Pick one style within the model (the file's existing convention is `Optional`).
- Consider also asserting the sanitized/bounded fields on the *positive* manual-import test (e.g. `result.manual_import_job_id`), not just `ingestion_method`; today the CSV/JSON cases don't assert the provenance fields survive validation.
- The sanitizer silently strips `\x00`/`\x1f`/`\x7f` (so `csv\x00_upload` becomes the valid `csv_upload`). That matches existing behavior, but for enum-like fields such as `ingestion_method` a `Literal[...]` would be stricter than sanitize-then-accept.
- No IAM, Bedrock model id, or frontend/i18n surface is touched, so those repo rules don't apply here.

---

perrozzi:

This PR widens the strict `IngestMessage` schema so manual-import producers (manual confirm, CSV upload, JSON upload) validate cleanly, adding `source_url`, `ingestion_method`, `source_origin`, `manual_import_job_id` plus a new length constant, and wires the new string fields into the existing sanitization/URL validators. Tests cover all three producer shapes, boundary rejection derived from the constants, and sanitization. The change is small, focused, and matches the existing schema patterns — I see no blockers.

### Should fix

- `voc-datalake/plugins/_shared/schemas.py`: `source_url` is documented as a legacy alias for `url` but is declared as an independent field with no reconciliation. A producer that emits only `source_url` will validate fine while every consumer reading `.url` sees `None` (and vice versa). Prefer a real alias (`Field(None, validation_alias=AliasChoices("url", "source_url"))`) or a `model_validator(mode="after")` that backfills one from the other, so downstream code has a single source of truth. The pre-existing `channel`/`source_channel` pair has the same latent problem — worth fixing both while you're here.
- `ingestion_method` is a closed set at this boundary (`manual`, `csv_upload`, `json_upload`). A `Literal[...]` (or enum) would validate the actual invariant instead of an arbitrary 64-char cap, and would catch typos in producers rather than propagating them into the lake. If you keep the free-form string, that's a conscious tradeoff worth a comment.

### Nits

- Style drift: new fields use `str | None` while surrounding fields use `Optional[str]`. Pick one within the file (the file's convention is `Optional`).
- `source_channel`/`channel` still use the literal `64`; now that `MAX_INGESTION_METHOD_LENGTH` exists, consider a `MAX_CHANNEL_LENGTH` constant so the rejection tests can be derived the same way.
- `test_rejects_oversized_manual_import_provenance` loops with `pytest.raises` inside; a `@pytest.mark.parametrize` over `(field, limit)` gives per-field failures instead of stopping at the first one (consistent with the parametrized test above it).
- `test_sanitizes_manual_import_provenance` asserts control characters are *removed* (`'csv\x00_upload' -> 'csv_upload'`); since `sanitize_string`'s body isn't in the diff, double-check it strips rather than replaces so the assertion is testing behavior, not coincidence.
- `source_url` goes through `validate_url` but not `sanitize_string`; confirm `validate_url` rejects/strips embedded control characters, otherwise a URL with `\x00` could pass.

---

proton0210:
Thanks for the automated review. The blocker from the first revision is resolved by preserving the producer payload and modeling `source_url`; there is no longer a producer-side behavior change. The boundary tests now derive from schema constants, and the positive case asserts every confirm-import provenance field.

I kept URL/channel canonicalization out of this PR because the live manual-confirm producer emits both `source_url` and `url`, while the processor intentionally consumes `url`. Changing alias semantics would affect the broader multi-producer contract tracked in #249 rather than the focused precondition in #412.

proton0210:
Superseded by #421. GitHub closed this cross-repository PR when the fork branch was renamed from `codex/issue-412-ingest-message-schema` to the neutral `fix/issue-412-ingest-message-schema`; the commit and proposed changes are unchanged.

PR #421, opened 2026-09-11
fix: align ingest schema with manual imports

### LINKED ISSUE #412: IngestMessage forbids extras but omits fields manual import sends; validation is silently disabled in the processor

proton0210:
## Summary

- declare all manual-import provenance fields on `IngestMessage`, including `source_url`, which manual confirm emits alongside `url`
- apply explicit length bounds and the existing URL/string sanitization to the newly modeled fields
- add regression coverage for the manual-confirm, CSV-upload, and JSON-upload message shapes
- derive rejection-boundary tests from the schema constants and verify every manual provenance field survives validation

The producer audit found that plugin ingestors route through `BaseIngestor`/`BaseWebhook` normalization and already emit fields represented by the schema. The broader feedback-form mismatch and enabling processor-side validation remain part of #249 and are intentionally outside this focused change.

## Testing

- `python -m pytest -q --no-header --tb=short` — 4,354 passed
- `python -m ruff check plugins/_shared/schemas.py plugins/_shared/test/test_schemas.py` — no new findings compared with the repository baseline
- `git diff --check`

Replaces #420 after its fork branch was renamed.

Closes #412


perrozzi:

This PR extends `IngestMessage` with the manual-import provenance fields (`source_url`, `ingestion_method`, `source_origin`, `manual_import_job_id`), wires them into the existing string-sanitization and URL validators, and adds pytest coverage for the manual-confirm / CSV-upload / JSON-upload message shapes. It's a small, focused, low-risk schema alignment and looks essentially clean; the notes below are polish and test-robustness items.

### Should fix

- `voc-datalake/plugins/_shared/schemas.py`: `ingestion_method` is a closed set in practice (`manual`, `csv_upload`, `json_upload`), but it's modeled as a free 64-char string. Per the "validate at system boundaries" rule, a `Literal[...]` (or enum) would reject typos at ingest instead of letting an unknown method flow into the lake. If the openness is deliberate (e.g. plugins may add methods), a comment saying so would help.
- Duplicated URL semantics: `url` and `source_url` are now both accepted with no relationship enforced. Downstream consumers that read only one of the two will silently lose data for producers that emit only the other. Consider a `model_validator` that backfills whichever is missing, so the canonical field is always populated.
- `test_rejects_oversized_manual_import_provenance` loops over three fields inside one test; the first failure masks the remaining cases. Convert to `@pytest.mark.parametrize` (as done in the accept test) so each bound is reported independently.
- Regression-test sanity check: please confirm `IngestMessage`'s model config is `extra="forbid"`/`"ignore"` rather than `"allow"`. With `extra="allow"`, `getattr(result, 'ingestion_method')` in `test_accepts_messages_from_each_manual_import_path` would still pass with the schema change reverted, so the new tests wouldn't actually pin the fix.

### Nits

- Type-annotation style is now mixed in the same block: new fields use `str | None` while the surrounding fields use `Optional[str]`. Pick one (the file's prevailing `Optional[...]`) for consistency.
- `source_url` is not included in the `sanitize_string` validator list — consistent with `url`, but worth confirming `validate_url` also strips surrounding whitespace/control chars for that field.
- Tests hardcode a future `created_at` (`2026-01-01T12:00:00Z`); if a future-timestamp guard is ever added these will fail for unrelated reasons. A fixed past date or a relative timestamp is safer.
- `MAX_INGESTION_METHOD_LENGTH` is only used once; fine, but if the field becomes a `Literal` the constant can be dropped.

---

PR #422, opened 2026-09-12
feat(verification): target-owned ABCA fixture provider

perrozzi:
## Summary

Adds a private, target-owned fixture provider so the ABCA verification lane can seed and tear down a deterministic prioritization baseline against a deployed slot, plus the `.abca/fixture-manifest.json` contract that declares its shape.

The design keeps declaration and authority separate:

- **`.abca/fixture-manifest.json` is declaration-only.** It states the capability alias, document/row roles, expected counts and the provider protocol version. It carries no ARN, IAM, secret, table name, command, or write recipe, so a commit to this repository cannot grant itself privileged access.
- **The provider owns all VoC semantics.** Every key, item shape, relationship, expiration and cleanup rule lives in `verification_fixture_provider.py`. Nothing about this repository's storage model moves into the platform.
- **Invocation authority is operator-owned.** The provider is reachable only by direct Lambda invoke from an explicitly registered caller — no API Gateway route, Cognito authorizer, Function URL, frontend config, or model access.

## Changes

| file | |
|---|---|
| `.abca/fixture-manifest.json` | new — capability, roles, counts, protocol |
| `voc-datalake/lambda/api/verification_fixture_provider.py` | new — setup / probe / teardown |
| `voc-datalake/lambda/api/test/test_verification_fixture_provider.py` | new — 10 tests |
| `voc-datalake/lib/stacks/api-stack.ts` | provider function, role, scoped DynamoDB + KMS grants, ARN output |
| `voc-datalake/lib/utils/lambda-asset-excludes.ts` | keep provider source out of unrelated Lambda bundles |
| plus the two matching test files | |

The manifest sits at the **repository root**, not under `voc-datalake/`: the platform resolves that fixed path against the Git worktree root, and a manifest anywhere else is silently invisible to it.

## Testing

- `verification_fixture_provider` — 10/10 (`pytest`), covering deterministic setup, idempotent reuse, collision fencing, expired-TTL repair, exact probe relationships, error mapping and ownership-fenced teardown with zero-residue proof.
- `lambda-asset-excludes` — 15/15. The provider source is excluded from the shared `lambda/` bundles so its edits do not churn the other API and job function fingerprints; its own bundle deliberately omits that exclusion.
- `api-stack` — 86 passing. The one failure is the pre-existing docs-lockstep case, unrelated to this branch.
- Provider lifecycle was also exercised end to end against a deployed non-production slot: created, reused with identical ids, probed, torn down twice, with exact-key absence and zero ownership markers remaining afterwards.

## Review note

The provider is currently created **unconditionally** in `VocApiStack`, so merging would add it to every deployment of this sample, production included. It is IAM-gated with no external surface, but if you would rather it exist only in non-production deployments, say so and I will gate it on the deployment prefix before merge.


perrozzi:

This PR adds a private, direct-invoke-only Lambda ("fixture provider") plus a repo-root `.abca/fixture-manifest.json` contract so an external verification harness can seed/probe/tear down a deterministic prioritization baseline in a deployed slot, along with CDK wiring (role, scoped DynamoDB + KMS grants, ARN output), asset-exclude plumbing, and unit tests. The Python module itself is carefully written — closed request schema, ownership-fenced conditional writes, atomic transactions, idempotent teardown with residue proof — but the deployment posture is the problem: a verification-only component that can write and delete items in the live `Projects`/`Aggregates` tables is created unconditionally in `VocApiStack`, with no principal-scoped invoke authority and no evidence that the TTL-based cleanup it relies on actually exists.

### Blockers

- **Unconditional deployment into every environment, production included** (`voc-datalake/lib/stacks/api-stack.ts`, ~L706-790). The PR description already flags this. A function whose whole purpose is to mutate the production data tables for CI must not exist in a normal deployment of a public AWS sample. Gate it on deployment prefix/CDK context (default off), or move it out of `VocApiStack` into a separate opt-in stack. As written, every downstream consumer of this sample inherits it.
- **No resource policy restricting who may invoke it.** The description claims invocation is "reachable only by direct Lambda invoke from an explicitly registered caller", but nothing in the diff narrows the principal — no `grantInvoke`, no `addPermission` with a specific role/account, no `SourceAccount`/`SourceArn` condition. The only control is "whoever happens to hold `lambda:InvokeFunction`", and the function ARN is published as a CFN output (`VerificationFixtureProviderArn`). Add an explicit resource-based policy for the registered caller, and drop the output unless the gate above is in place.
- **Fixture data is indistinguishable from real data in user-facing queries, and cleanup is unproven.** `verification_fixture_provider.py::_records` writes the project META item with `gsi1pk: 'TYPE#PROJECT'` and `gsi1sk: '2026-01-01T00:00:00+00:00'`, so a fixture project will be returned by the normal project-list GSI query (and, being future-dated, sorted to the top). Non-TTL cleanup is only guaranteed if teardown runs; the module comments and the `ttl` attribute assume DynamoDB TTL is enabled on `ttl` for both `projectsTable` and `aggregatesTable`, and the diff neither shows nor asserts that. If TTL isn't enabled on those tables, abandoned fixtures persist indefinitely in the real project list. Please confirm/enable TTL and add a stack assertion, and consider a filterable marker so app queries can exclude fixture rows.

### Should fix

- **`lambda_handler` has zero test coverage.** `test_verification_fixture_provider.py` imports `parse_provider_request`, `setup_fixture`, `probe_fixture`, `teardown_fixture`, `_transact` — never `lambda_handler`. That leaves the entire boundary untested: `_ERROR_CODES` mapping, the `_failure` envelope, the `operation if operation in _OPERATIONS else 'unknown'` sanitization, and the bare-`Exception` path. Add handler-level tests for each error class → `error_code` pair.
- **Missing Metrics per repo convention.** The module uses `logger` and `tracer` from `shared.logging` but no Metrics decorator, unlike other Python Lambdas in this repo. Either add it or note explicitly why an internal-only provider opts out.
- **`Table.grant(...)` widens the resource set.** `projectsTable.grant(...)` / `aggregatesTable.grant(...)` (api-stack.ts) expand to the table ARN *plus* `…/index/*` when the table has GSIs, which is both unnecessary for `GetItem`/`PutItem`/`DeleteItem` and likely to trip cdk-nag `AwsSolutions-IAM5`. Prefer an explicit `PolicyStatement` scoped to the two table ARNs, and extend the new api-stack test to assert `Resource` (it currently only asserts `Action`).
- **Reuse never extends the TTL.** `_existing_expiry` returns the stored `expires_at`/`ttl` unchanged, and the test asserts "same-owner repair is byte-stable". A verification job that reuses a fixture near the end of the 6h window can have DynamoDB TTL delete its records mid-run. Consider renewing when remaining lifetime drops below a threshold (and adjust the byte-stability assertion accordingly).
- **Source-text assertions are brittle.** `lambda-asset-excludes.test.ts` asserts on the literal text of `api-stack.ts` (`expect(source).toContain("handlerFileName === 'verification_fixture_provider.py'")`, plus a regex match count). Any reformatting or refactor breaks these without a behavior change. The behavioral assertions right below (`unrelatedIgnores(...)`) already cover the intent — prefer synthesizing the template and comparing asset hashes/keys.
- **`.abca/` in a public AWS sample.** A vendor-specific verification contract at the repository root, plus a Lambda dedicated to it, is product surface that consumers of this sample will inherit and have to explain. At minimum document it (README / CONTRIBUTING) so the file isn't unexplained; ideally keep it out of the shipped app stack entirely.

### Nits

- `_transact`: the trailing `raise ServiceError(failure_message)` after the `for` loop is unreachable — the final attempt already raises inside the loop.
- `_transact`: `isinstance(reasons, list)` is checked both in the comprehension guard and the trailing conditional; drop the inner one.
- `setup_fixture` uses `assert expires_at is not None and ttl is not None` for an invariant in Lambda code; raise `ServiceError`/`ConfigurationError` instead so the behavior doesn't depend on `-O`.
- `_ids(request)` is computed in `setup_fixture` and again inside `_records`; pass the already-computed `ids` through.
- The new api-stack case is appended at file top level rather than inside a `describe` block, unlike the rest of `api-stack.test.ts`.

---

perrozzi:

## Follow-up review

Since the last pass the provider (role, KMS/DynamoDB grants, function, log group, ARN output) is now wrapped in `if (this.deploymentPrefix)` with a matching "default deploy omits it" template test, `lambda_handler` gained a dedicated test class (error-code mapping, unknown-operation sanitization, non-dict event, per-operation dispatch), the duplicated `isinstance(reasons, list)` guard was dropped, and the `assert` on resolved expiry became a `ServiceError`. The first Blocker and two of the Should-fix/Nit items are addressed; the invoke-authority and fixture-data-visibility Blockers are not.

### Resolved
- Unconditional deployment — provider and its `VerificationFixtureProviderArn` output now exist only in a prefixed deployment (`voc-datalake/lib/stacks/api-stack.ts`), with a regression test asserting the default template contains no `voc-fixture-provider` reference.
- `lambda_handler` boundary coverage — `test_verification_fixture_provider.py::TestLambdaHandler` covers each `_ERROR_CODES` pair, the unmapped-`ApiError` → `service` degradation, the bare-`Exception` → `internal` path, operation sanitization, and dispatch.
- Nits: redundant `isinstance(reasons, list)` guard removed; `assert expires_at is not None and ttl is not None` replaced with `raise ServiceError`.

### Still open
- No resource-based policy narrowing who may invoke the function (no `grantInvoke`/`addPermission`, no `SourceAccount`/`SourceArn`); the ARN is still published as an output.
- Fixture project META still carries `gsi1pk: 'TYPE#PROJECT'` with a future-dated `gsi1sk`, so it surfaces (and sorts first) in the normal project list; TTL-on-`ttl` for `Projects`/`Aggregates` is still neither shown nor asserted.
- No Metrics decorator / no stated reason for opting out.
- `Table.grant(...)` still widens to `…/index/*`; the api-stack test asserts actions but never that `Resource` excludes indexes.
- Reuse still returns the stored expiry unchanged, so a long job can be TTL-deleted mid-run.
- `lambda-asset-excludes.test.ts` still asserts on literal `api-stack.ts` source text.
- `.abca/` contract still undocumented in README/CONTRIBUTING.
- Nits: unreachable trailing `raise ServiceError` in `_transact`; `_ids(request)` computed twice; the two new api-stack cases are appended at file top level rather than inside a `describe`.

### New findings
- **`deploymentPrefix` is not a non-production signal** (`voc-datalake/lib/stacks/api-stack.ts`, gate at ~L710). It only means "side-by-side deployment"; a prefixed production slot silently gets a Lambda that can write and delete rows in the live `Projects`/`Aggregates` tables. Prefer an explicit, default-off CDK context flag (e.g. `vocEnableVerificationFixtureProvider`) — or keep the prefix gate but AND it with such a flag — and document it, so the capability is opt-in by intent rather than by deployment topology.
- **The provider's IAM now only exists in the prefixed template**, so cdk-nag findings from `Table.grant`'s `index/*` expansion (`AwsSolutions-IAM5`) will only appear in a synth shape CI may not check. Please confirm the nag/aspect suite covers the prefixed synth, otherwise the still-open grant-scoping issue becomes invisible in CI.
- **New tests depend on a `lambda_context` fixture** (`voc-datalake/lambda/api/test/test_verification_fixture_provider.py`) that isn't introduced in this PR — worth confirming it's provided by the shared `conftest.py` for this test directory, since the whole `TestLambdaHandler` class errors out at collection if it isn't.

---

perrozzi:

## Follow-up review

The only change since the last pass is the gate: the provider now requires `deploymentPrefix` **and** an explicit `enableVerificationFixtureProvider` context value (accepting only `true`/`'true'`), with the api-stack tests reorganised into a `describe` covering the three single-condition shapes plus non-canonical truthy spellings. That addresses the first of my previous "New findings"; the rest of the still-open list is untouched.

### Resolved
- Topology-only gating — `voc-datalake/lib/stacks/api-stack.ts` now requires an explicit opt-in flag in addition to the prefix, with negative-shape and spelling tests in `api-stack.test.ts`.
- The two new api-stack cases are now grouped in a `describe` (previous nit).

### Still open
- No resource-based invoke policy (`grantInvoke`/`addPermission`, `SourceAccount`/`SourceArn`) while the ARN is exported.
- Fixture project META still carries `gsi1pk: 'TYPE#PROJECT'` so it appears in the normal project list; TTL-on-`ttl` for `Projects`/`Aggregates` still unasserted.
- No Metrics decorator and no stated reason for opting out.
- `Table.grant(...)` still widens to `…/index/*`; the api-stack assertion checks actions, not that `Resource` excludes indexes.
- Reuse returns the stored expiry unchanged, so a long job can be TTL-deleted mid-run.
- `lambda-asset-excludes.test.ts` still asserts on literal `api-stack.ts` source text.
- `.abca/` contract still undocumented in README/CONTRIBUTING.
- cdk-nag coverage: provider IAM exists only in the prefixed synth — now *prefixed **and** flagged*, so the shape is even less likely to be exercised by the nag/aspect suite; please confirm CI synths it.
- Nits: unreachable trailing `raise ServiceError` in `_transact`; `_ids(request)` computed twice in `setup_fixture`.

### New findings
- **The new operator-facing flag is undocumented** (`voc-datalake/lib/stacks/api-stack.ts`). `enableVerificationFixtureProvider` exists only as a code comment — nothing in README/CONTRIBUTING or `cdk.json` tells an operator it exists, what it creates, or that it must never be set on a production slot. This is the one thing standing between a live table and a delete-capable Lambda, so it needs an explicit doc entry alongside the still-missing `.abca/` contract docs.
- **Minor, `api-stack.test.ts`**: the "truthy-looking spellings" case only asserts `fixtureProviderEntry(...)` is undefined; the other negative case also asserts no `voc-fixture-provider` string and no output. Reusing the same three assertions there would catch a partial leak (role/log group created but function skipped) for the mis-spelled-flag path too.

---

perrozzi:

## Follow-up review

Since the last pass the author added the `docs/deployment.md` section for `.abca/` + `enableVerificationFixtureProvider`, a `verification_fixture_id` marker helper in `shared/project_writes.py` with a filter in `list_projects`, a TTL renewal margin (`RENEWAL_MARGIN_SECONDS`) with two boundary tests, a documented Metrics opt-out, a hand-written table-scoped DynamoDB policy statement (replacing `Table.grant`) plus GSI-bearing test tables and `Resource` assertions, and a cdk-nag case over the prefixed+flagged shape. That clears most of the previous list.

### Resolved
- Operator-facing flag and `.abca/` contract now documented in `docs/deployment.md`, including the "never on production" statement.
- Fixture projects no longer surface in the human project list — `projects.py` skips them via `is_verification_fixture`, with the marker asserted in lockstep from the provider test.
- Reuse of a nearly-expired fixture now renews instead of handing back a mid-run-deletable expiry (`RENEWAL_MARGIN_SECONDS`, plus inside/outside-margin tests).
- DynamoDB grant is now scoped to the two table ARNs with no `index/*`, and the test asserts `Resource` (with GSIs present so the assertion isn't vacuous).
- cdk-nag now runs over the gated shape in `api-stack.test.ts`; Metrics opt-out has a stated reason; `_ids(request)` no longer computed twice; the mis-spelled-flag case asserts all three shapes.

### Still open
- No resource-based invoke policy (`grantInvoke`/`addPermission` with `SourceAccount`/`SourceArn`) while the ARN is exported.
- TTL enablement on `Projects`/`Aggregates` (`timeToLiveAttribute: 'ttl'`) still unasserted — the docs now promise "records also carry a TTL", so this is worth a test.
- `lambda-asset-excludes.test.ts` still asserts on literal `api-stack.ts` source text.
- Unreachable trailing `raise ServiceError` in `_transact`.

### New findings
- **Should fix — the prioritization row is unfiltered** (`voc-datalake/lambda/api/verification_fixture_provider.py`, aggregates item under `pk: 'PRIORITIZATION'`). The new marker/filter pair covers only `list_projects`; the fixture's `ROW#…` item is written into the same partition the real prioritization read paths (and aggregation jobs) enumerate, so a verification run still shows up as somebody's prioritization row. Either apply `is_verification_fixture` on that read path too, or explain in the module docstring/docs why it is acceptable there but not for projects.
- **Should fix — no regression test where the fix lives** (`voc-datalake/lambda/api/test/test_projects.py`). `test_every_record_carries_the_marker_the_project_list_filters_on` proves the provider writes the attribute, but nothing asserts `list_projects` actually skips an item carrying it; a revert of the two-line filter in `projects.py` would keep the whole suite green.
- Minor, `voc-datalake/lambda/api/test/test_verification_fixture_provider.py`: the new `from shared.project_writes import …` is inserted above `from shared.exceptions import …`, breaking the alphabetical import block — check it passes the repo's lint config.
- Minor, `voc-datalake/lambda/shared/project_writes.py`: `is_verification_fixture` keys on mere attribute presence, so any project item that ever acquires a `verification_fixture_id` field silently disappears from the list. Writes are attribute-restricted today, so this is a note rather than a bug — a one-line comment saying so would keep it that way.

---

perrozzi:

## Follow-up review

Since the last pass the author added an optional `verificationFixtureInvokerArn` context that attaches a resource-based `addPermission` (with synth-time ARN validation and four negative cases), filtered fixture rows out of `api_get_prioritization_scores`, added the two missing regression tests (mocked `list_projects` + a moto test that puts a marked `ROW#` item into the real partition), asserted the Aggregates/Projects TTL asymmetry in `core-stack.test.ts`, corrected the docs' TTL claim, fixed the import order and documented the presence-keying caveat. That closes almost everything.

### Resolved
- Invoke can now be pinned to one principal via `verificationFixtureInvokerArn`, validated at synth and covered by tests.
- Prioritization read path now skips fixture rows (`projects_handler.py`), with a moto regression test that fails if the filter is reverted.
- `list_projects` filtering is now asserted directly in `test_projects.py`, not only via the provider-side marker lockstep.
- TTL is asserted per table in `core-stack.test.ts`, and the docs no longer overclaim TTL as the cleanup mechanism.
- Import block ordering and the `is_verification_fixture` presence-keying note.

### Still open
- `lambda-asset-excludes.test.ts` asserts on literal `api-stack.ts` source text (and the new case adds two more such assertions).
- Unreachable trailing `raise ServiceError` in `_transact`.

### New findings
- **Should fix — the ARN regex accepts a path wildcard** (`voc-datalake/lib/stacks/api-stack.ts`, `verificationFixtureInvokerArn` validation). `/^arn:[a-z0-9-]+:iam::\d{12}:(role|user)\/.+/` has no end anchor and no `*` rejection, so `arn:aws:iam::111122223333:role/*` passes synth; IAM rejects a wildcard-path principal at *deploy* time, which contradicts the docs promise that "anything else fails at synth rather than deploying a policy that grants nobody" — and a reader could reasonably read it as a broad-but-valid grant. Anchor the pattern and reject `*`, with a case in the existing `it.each`.
- Minor — only one prioritization read is filtered. `is_verification_fixture` is now applied in `api_get_prioritization_scores`, but the docs say fixture rows are excluded "from the prioritization read" generally; please confirm nothing else enumerates `pk = 'PRIORITIZATION'` (aggregator/jobs, ballot summaries) or add the filter there too, since the marker is the only thing separating fixture rows from real ones.
- Minor — `voc-datalake/lib/stacks/core-stack.test.ts`: `tableNamed()` picks the first table whose `TableName` JSON merely *contains* the fragment, so it silently binds to whichever table sorts first if another name ever contains `projects`; matching the construct id or the exact name would be less fragile. Also note the second case pins `TimeToLiveSpecification` as *undefined*, so enabling TTL on Projects later reads as a test violation rather than an improvement — the comment covers the intent, but flag it if that changes.

---

perrozzi:

## Follow-up review

Since the last pass the author anchored the invoker-ARN regex and added four negative cases for it, rebound the core-stack TTL assertions to construct ids with an explicit failure message, narrowed the docs claim about the prioritization partition to "the one read that enumerates it" (with the by-exact-key rationale for the rest), and documented why the `lambda-asset-excludes` wiring check asserts on source text. Everything looks clean; only nits remain.

### Resolved
- `verificationFixtureInvokerArn` now uses `/^…:(role|user)\/[^*\s]+$/`, so `role/*`, `role/team-*`, trailing junk and an empty role name all fail at synth, each with a case in the `it.each`.
- `core-stack.test.ts` binds by construct id with a "expected exactly one" guard instead of a `TableName` substring, and the Projects case carries a message telling a future author what to update if TTL is added.
- Docs no longer overclaim "the prioritization read" generally; the exact-key rationale for the remaining accessors is stated.

### Still open
- Unreachable trailing `raise ServiceError` in `_transact` (`verification_fixture_provider.py`).
- `lambda-asset-excludes.test.ts` still asserts on `api-stack.ts` source text — now with an explicit rationale and a real behavioral test of the exclude list next to it, so I'm fine leaving it as an accepted trade-off.

### New findings
- Nit — `voc-datalake/lib/stacks/api-stack.ts`: the pattern rejects `*` but not `?`, so `arn:aws:iam::111122223333:role/te?m` still passes synth and fails at deploy. Same class of gap as the one just closed, though far less likely to be written deliberately; add `?` to the negated class if you want the "fails at synth" promise to be airtight.

---

perrozzi:

## Follow-up review

The only change since the last pass is the one-character regex fix — `[^*\s]` became `[^*?\s]` in the `verificationFixtureInvokerArn` validation — plus a matching `it.each` case for `role/te?m` and an updated comment. That closes the nit from my previous review.

### Resolved
- `voc-datalake/lib/stacks/api-stack.ts`: both IAM wildcards (`*` and `?`) are now rejected in the principal path, with a covering negative case in `api-stack.test.ts`.

### Still open
- Unreachable trailing `raise ServiceError` in `_transact` (`voc-datalake/lambda/api/verification_fixture_provider.py`) — nit.
- `lambda-asset-excludes.test.ts` asserting on `api-stack.ts` source text — previously accepted as a documented trade-off.

No new findings; the incremental change is clean and correctly scoped.

---

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2CY906ACTAP3MSA8TRRVV66` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T08:34:03.542Z |
| Duration | 131.2s |
| Cost | $0.4009 |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2D1Z46TJMA7GN99Z9ZB116B` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T09:40:04.428Z |
| Duration | 219.1s |
| Cost | $0.5613 |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2D2CH6CHF7WJHN264EV72S4` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T09:47:50.470Z |
| Duration | 258.3s |
| Cost | $0.6388 |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2D3WEF2KA23VHFJA4J3F1RW` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T10:13:33.473Z |
| Duration | 229.8s |
| Cost | $0.6440 |

perrozzi:

| Field | Value |
|-------|-------|
| Task  | `01M2D6ZA926B3PASF4MNRYWQN7` |
| Status | **COMPLETED** |
| Last event | `task_completed` @ 2026-09-13T11:05:02.999Z |
| Duration | 84.9s |
| Cost | $0.2939 |

PR #423, opened 2026-09-22
AIDLC Discovery 1.5.0: one package for Quick, Kiro, and Claude Code

perrozzi:
*Issue #, if available:* —

## Description of changes

**AIDLC: Discovery 1.5.0** makes the repository root directly installable in **Amazon Quick**, **Kiro** (IDE and CLI), and **Claude Code** — one package, no build step. It replaces the previous flow where `build-plugin.sh` had to assemble a Quick-only import folder.

### How each harness installs it

| Harness | Install |
|---|---|
| Amazon Quick | GitHub branch ZIP (or clone) → Plugins → **Import folder** on the repo root |
| Kiro | Powers panel → GitHub URL / folder; or skill-only from `skills/aidlc-discovery/` |
| Claude Code | `claude plugin marketplace add aws-samples/sample-voice-of-customer-datalake@v2` → `/plugin install aidlc-discovery@voc-datalake`; or `--plugin-dir` / `--plugin-url` |

### What changed

- **One manifest, two vocabularies.** Root `plugin.json` carries the [Agent Plugins 1.0.0](https://agent-plugins.org/) fields Kiro reads and Quick's v0.1 fields (`id`, `format_version`, `created_at`) in the same file; the Agent Plugins spec requires clients to ignore fields they don't define. `mcps.json`/`tasks.json` sit beside it for Quick; `.claude-plugin/plugin.json` + `marketplace.json` make the branch its own Claude Code marketplace.
- **Skill layout.** `SKILL.md`, the 15 reference prompts, sample data, and `config.default.md` moved into `skills/aidlc-discovery/` (skill id renamed from `quick-aidlc-discovery` — one identifier across all three harnesses). All references inside `SKILL.md` were already skill-root-relative and are unchanged.
- **Harness-neutral instructions.** Workflow steps name capabilities; a new *Working Across Harnesses* table maps each capability to the Quick/Kiro/Claude tool. Config and artifacts are written to the user's working folder, never into the installed skill (Kiro/Claude install into caches).
- **Dedicated agents that actually run.** `.kiro/agents/` (11 JSON agents, `prompt` → the phase/specialist `AGENT.md`, sub-agent wiring on the orchestrators) and `agents/` (the same 11 as Claude Code plugin subagents) replace the Quick agent YAML experiments that Quick's import ignored.
- **Removed:** `build-plugin.sh`, `build-workshop-zip.sh`, `plugin/`, `.qplugin` (Quick hangs on file import; folder import is the supported path).

### Verified (headless, on this branch)

- **Claude Code 2.1.261** — `--plugin-dir <root>` loads 1 skill + 11 agents; `plugin validate . --strict` passes; a live session invoked the skill and resolved its bundled prompts and sample data. Found and fixed two loader contracts: plugin agents load **only** from the root `agents/` directory (a manifest `agents` list validates but loads nothing), and the skills validator passes malformed frontmatter, so `plugin details` is the loading truth.
- **Kiro CLI 2.23** — all 11 agents load from the repo root; `skill://skills/aidlc-discovery/SKILL.md` resolves (negative control confirmed); found and fixed: Kiro reads frontmatter literally, so a folded `description: >-` reaches the model as `>-` — `description`/`compatibility` are now single-line (test-guarded).
- **Test suite** — `python3 -m unittest discover tests`: 16 tests green (manifest contracts for all three loaders, Agent Skills frontmatter limits, harness-neutral prose, prompt/agent resolution, doc links; runs `kiro-cli agent validate` and `claude plugin validate` when present).

### Still open (needs the Quick app / Kiro IDE)

- Quick folder import of the new root (extra entries `.claude-plugin/`, `.kiro/`, `agents/`, docs, tests; extra manifest keys; kebab-case `name`).
- Kiro IDE power/skill import as distinct from the CLI.
- Existing installs under the old ids (`quick-ai-plc`, `quick-aidlc-discovery`) must be removed manually; see INSTALL.md.

By submitting this pull request, I confirm that you can use, modify, copy, and redistribute this contribution, under the terms of your choice.


perrozzi:

This PR restructures the v2 (prompt-only) branch into a single repo-root package installable by Amazon Quick, Kiro, and Claude Code: it moves `SKILL.md`/`agents/`/`knowledge-base/`/`config.default.md` under `skills/aidlc-discovery/`, hoists `plugin.json`/`mcps.json`/`tasks.json` to the root, adds `.claude-plugin/` manifests plus 11 Kiro JSON agents and 11 Claude Code subagent stubs, deletes the two build scripts, and rewrites the docs and the package-shape tests. Note that the platform-specific rules for this repo (Lambda/IAM sizing, ingestor prefix registration, Bedrock ARN lockstep, frontend i18n/Zod) don't apply here — nothing in the diff touches CDK, Python Lambdas, or the React app. Overall the restructure is coherent and the new tests pin a real contract, but a few harness details look wrong and some of the new assertions may not hold against files that weren't updated.

### Blockers

- **Claude Code orchestrators declare a non-existent `Agent` tool.** `agents/aidlc-signal-analyzer.md` and `agents/aidlc-ideation.md` set `tools: Read, Grep, Glob, Bash, Write, Edit, Agent` and their bodies say "delegate … with the Agent tool". Claude Code's subagent-invocation tool is `Task`; there is no `Agent` tool, so these two agents get no delegation capability at all and the documented parallel specialist flow (README, INSTALL, WORKSHOP-SETUP tables) silently degrades. Either switch to `Task` or drop the `tools` key so the agents inherit the full tool set. The CHANGELOG's "Verified" section only claims agent *loading* was checked, not delegation, and `tests/test_documentation_and_packages.py::test_claude_agents_resolve_prompt` only validates `name`/`description`/prompt path — add an assertion on the delegation tool name once fixed.

### Should fix

- **The documented Kiro global-agent install breaks every `prompt` path.** `.kiro/agents/*.json` use `"prompt": "file://../../skills/aidlc-discovery/…/AGENT.md"`, which only resolves when the file sits at `<repo>/.kiro/agents/`. INSTALL.md tells users to "copy the folder's files into `~/.kiro/agents/`", where `../../` resolves to `~/` while the skill lives at `~/.kiro/skills/aidlc-discovery/`. Either document the required relative depth per destination, or ship a variant/absolute-path instruction. `test_kiro_agents_resolve_prompt_and_subagents` resolves the path relative to `agent_dir`, so it can't catch this.
- **`test_skill_prose_names_capabilities_not_quick_tools` asserts over prompt files this PR did not touch.** Only `03-prototype/skills/html-prototype.md`, `04-validation/AGENT.md`, `04-validation/skills/results-dashboard.md`, and `survey-generator.md` were de-Quick-ified in the diff; files like `agents/01-signal-analyzer/sub-voc-analyst/skills/signal-analysis.md`, `02-ideation/skills/prd-generation.md`, and `03-prototype/skills/ide-exporter.md` moved with 100% similarity. Any lingering `run_python` / `file_rag_search` / `file_write` mention in them without a `Quick:` qualifier now fails the suite — please confirm `python3 -m unittest discover tests` is green on this branch.
- **The primary Quick install path is documented but explicitly unverified.** The CHANGELOG's "Known open checks" admits Quick's folder importer hasn't been run against the new root (kebab-case `name` now doubles as Quick's display name, plus new top-level entries `.claude-plugin/`, `.kiro/`, `agents/`, `tests/`). INSTALL.md/README present it as the recommended step with no caveat; either verify it or mark the Quick path as provisional in the user-facing docs.
- **`.gitignore` no longer ignores `dist/`.** The build scripts are gone, but `tests/…SKIP_DIRS` still skips `dist`, so the intent is inconsistent and any locally produced `dist/` becomes committable. Drop it from `SKIP_DIRS` or keep the ignore entry.

### Nits

- `WORKSHOP-SETUP.md` Level 1/Level 2 tables show labels like `` `agents/01-signal-analyzer/AGENT.md` `` while linking to `skills/aidlc-discovery/agents/...`; update the visible text to match the real path.
- `plugin.json` `created_at: "2026-09-22T00:00:00Z"` and the CHANGELOG heading `## 1.5.0 — 2026-09-22` are future-dated (same issue as 1.4.0).
- `plugin.json` `"repository"` points at `…/tree/v2`; Agent Plugins consumers generally expect a plain repository URL — consider moving the branch pointer to `homepage`.
- The new `skill_frontmatter` helper is a hand-rolled YAML parser with a bare `assert` and a special case for `>-`; if the toolchain allows it, `yaml.safe_load` on the frontmatter block would be less brittle (the block-scalar guard can stay as a separate regex check).
- Repo convention for Python tests is pytest; these remain `unittest` (they still run under pytest, but README instructs `python3 -m unittest discover tests`) — worth aligning or noting why.

---

perrozzi:

Since the last review the author added a caveat about the unverified Quick folder import, replaced the "copy into `~/.kiro/agents/`" instruction with a path-rewriting recipe for global installs, fixed the `WORKSHOP-SETUP.md` table labels, split `homepage`/`repository` in `plugin.json`, and dropped `dist` from the test `SKIP_DIRS`. The `Agent`-tool blocker is untouched.

### Resolved
- `plugin.json` now has a plain `repository` URL with the branch pointer in `homepage`.
- `WORKSHOP-SETUP.md` Level 1 labels now match the real `skills/aidlc-discovery/...` paths.
- `dist` removed from `SKIP_DIRS`, consistent with the dropped `.gitignore` entry.
- Quick folder-import path now carries an explicit "not yet re-verified" note in `INSTALL.md` (README's harness table still presents it without a caveat — minor).

### Still open
- **Blocker:** `agents/aidlc-signal-analyzer.md` and `agents/aidlc-ideation.md` still declare the non-existent `Agent` tool, and `INSTALL.md` still says the orchestrators "delegate to their specialists with the Agent tool" — should be `Task` (or drop `tools`), plus an assertion in `test_claude_agents_resolve_prompt`.
- `test_skill_prose_names_capabilities_not_quick_tools` still scans prompt files this PR didn't modify; please confirm `python3 -m unittest discover tests` is green on this branch.
- `plugin.json` `created_at` / CHANGELOG heading remain future-dated (2026-09-22).
- `skill_frontmatter` is still a hand-rolled parser with a bare `assert`; tests remain `unittest` rather than the repo's pytest convention.

### New findings
- **`INSTALL.md` global-install recipe will fail on a fresh machine.** `mkdir -p ~/.kiro/agents` creates only the agents dir, then `cp -R skills/aidlc-discovery ~/.kiro/skills/` (trailing slash, directory semantics) errors when `~/.kiro/skills` does not exist. Use `mkdir -p ~/.kiro/agents ~/.kiro/skills`.
- **`INSTALL.md` rewrite injects a literal `~` into a `skill://` URI.** `sed … 's|skill://skills/|skill://~/.kiro/skills/|'` is single-quoted, so the tilde is not expanded by the shell, and a URI-style resource path is unlikely to be tilde-expanded by Kiro either — per the CHANGELOG only the repo-root layout was verified. Prefer a double-quoted `"s|skill://skills/|skill://$HOME/.kiro/skills/|"` (or an absolute `file://`) and mark the global path as untested until exercised.
- Neither of the two commands above is covered by a test, while `test_kiro_agents_resolve_prompt_and_subagents` still only validates the in-repo depth — so the documented global variant can silently rot; a small check that the documented `sed` expressions still match the shipped `prompt`/`resources` prefixes would pin it.

---

perrozzi:

Since the last review the author fixed the `mkdir -p` to create both `~/.kiro/agents` and `~/.kiro/skills`, dropped the tilde-injecting second `sed` expression, added the Quick re-verification caveat to the README harness section, and added `test_install_md_global_recipe_matches_shipped_agent_paths` to pin the documented rewrite. The `Agent`-tool blocker is still untouched.

### Resolved
- `INSTALL.md` global recipe now creates `~/.kiro/skills` before copying into it.
- The literal-`~`-in-`skill://` rewrite is gone; only the `prompt` depth is rewritten.
- README harness table now carries the "Quick folder-import pending re-verification" caveat.
- The documented `sed` expression is now test-guarded against drift in the shipped `prompt` prefix.

### Still open
- **Blocker:** `agents/aidlc-signal-analyzer.md` and `agents/aidlc-ideation.md` still declare the non-existent `Agent` tool, and `INSTALL.md` still says the orchestrators "delegate to their specialists with the Agent tool" — should be `Task` (or drop `tools`), plus a test assertion.
- `plugin.json` `created_at` / CHANGELOG heading remain future-dated (2026-09-22).
- `skill_frontmatter` remains a hand-rolled parser with a bare `assert`; tests remain `unittest` rather than pytest.
- Please confirm `python3 -m unittest discover tests` is green on this branch (the harness-prose test still scans unmodified prompt files).

### New findings
- **`INSTALL.md` global install now leaves `resources` dangling.** Removing the second `sed` means the copied agents keep `"resources": ["skill://skills/aidlc-discovery/SKILL.md"]`, a workspace-relative path that does not exist under `~/.kiro/agents/`. The new prose asserts "the skill itself is picked up from `~/.kiro/skills/` automatically", but per the CHANGELOG only the repo-root layout was exercised — either verify that global skills resolve without the `skill://` entry or state plainly that the global variant is untested and may lose the skill resource.
- **The new test pins Markdown table escaping, not the command.** `tests/test_documentation_and_packages.py` asserts the literal `s\|file://../../skills/\|file://../skills/\|`, i.e. the pipe-escaped form required inside the table cell; moving the recipe out of the table (a reasonable readability fix) would fail the test for no functional reason. Matching on a normalized string (strip `\|` → `|` before asserting) would be more robust.

---

perrozzi:

Since my last review the author added two "Verified" bullets to `CHANGELOG.md` (live Claude Code delegation, and the Kiro global-install recipe run in an isolated `KIRO_HOME`), extended the `INSTALL.md` Kiro cell with the `skill://` tolerance note, and normalized the pipe-escaping in the new `INSTALL.md` test. That addresses both of my previous new findings and, with evidence, my `Agent`-tool blocker.

### Resolved
- `Agent` as the subagent tool is now evidence-backed (Claude Code 2.1.261 tool inventory lists `Agent`, no `Task`) — withdrawing the blocker.
- The global-install `resources` concern is verified and documented: the skill is inherited from `~/.kiro/skills/` and the stale `skill://` entry is tolerated.
- `tests/test_documentation_and_packages.py` now asserts the normalized command rather than the Markdown table escaping.

### Still open
- `plugin.json` `created_at` and the CHANGELOG `## 1.5.0 — 2026-09-22` heading remain future-dated.
- `skill_frontmatter` is still a hand-rolled parser guarded by a bare `assert`; tests remain `unittest`.

### New findings
- Nit (`agents/aidlc-signal-analyzer.md`, `agents/aidlc-ideation.md`): the `Agent` tool name is now the one loader contract in this PR that isn't test-guarded, even though it is version-sensitive by the author's own finding — a small assertion (`tools` ends in `Agent` for exactly the two orchestrators, matching the `subagent` check already done for Kiro) would catch a future rename, as would simply omitting `tools` to inherit the full set.
- Nit (`CHANGELOG.md`): the two new bullets are separated from the preceding `### Verified` bullets by a blank line, splitting them into a second list block; drop the blank line for consistent rendering.

Otherwise the delta is clean.

---

